PR #27, opened 2020-04-24
Improve block info

turbolent:
Closes dapperlabs/flow-go#3242

## Description

- Add `getBlock(at height: UInt64)` function
- Rename `Block.number` field to `Block.height`
- Add `Block.timestamp` field
- Remove `Block.previousBlock` and `Block.nextBlock` functions


runtime/stdlib/flow_builtin.go
```diff
@@ -323,13 +350,6 @@ func (t *BlockType) GetMember(identifier string, _ ast.Range, _ func(error)) *se
 			},
 		)
 
-	case "previousBlock", "nextBlock":
```
psiemens, line 326: 👍

PR #28, opened 2020-04-24
Include code hash instead of full code in account events

turbolent:
Closes dapperlabs/flow-go#3396

## Description

Instead of including the new code for the account in the account events, just include the SHA3-256 hash of the code.

While at it, fix and clean up the account events' types (addresses have type `Address` instead of `String`, keys are `[UInt8]`)


runtime/interpreter/conversion.go
```diff
@@ -0,0 +1,87 @@
...
+ */
+
+package interpreter
+
+import (
+	"errors"
+	"math"
+)
+
+func ByteArrayValueToByteSlice(value Value) ([]byte, error) {
```
psiemens, line 26: Great idea to move these functions here 👍

PR #29, opened 2020-04-25
Add ToRuntimeValue function to all cadence.Value types

psiemens:
Work towards https://github.com/dapperlabs/flow-go/issues/3286

## Description

This PR adds the `ToRuntimeValue` function to all `cadence.Value` types, allowing users to convert the "public" value types to the internal types usable by the runtime.
______

For contributor use:

### DIFF
--- a/convert_values_test.go
+++ b/convert_values_test.go
@@ -30,209 +30,185 @@ import (
 	"github.com/onflow/cadence/runtime/sema"
 )
 
-func TestConvertVoidValue(t *testing.T) {
-	value := convertValue(interpreter.VoidValue{}, nil)
-
-	assert.Equal(t, NewVoid(), value)
-}
-
-func TestConvertNilValue(t *testing.T) {
-
-	value := convertValue(interpreter.NilValue{}, nil)
-
-	assert.Equal(t, NewOptional(nil), value)
-}
-
-func TestConvertSomeValue(t *testing.T) {
-	t.Run("Nil", func(t *testing.T) {
-		value := convertValue(&interpreter.SomeValue{Value: nil}, nil)
-
-		assert.Equal(t, NewOptional(nil), value)
-	})
-
-	t.Run("Non-nil", func(t *testing.T) {
-		value := convertValue(
-			&interpreter.SomeValue{Value: interpreter.NewIntValueFromInt64(42)},
-			nil,
-		)
-
-		assert.Equal(t, NewOptional(NewInt(42)), value)
-	})
-}
-
-func TestConvertBoolValue(t *testing.T) {
-	t.Run("True", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


values.go
```diff
@@ -65,16 +73,26 @@ func (Optional) Type() Type {
...
 }
 
+func (v Optional) ToRuntimeValue() runtime.Value {
+	value := &interpreter.SomeValue{}
+
+	if v.Value != nil {
+		value.Value = v.Value.ToRuntimeValue().Value
+	}
+
+	return runtime.Value{Value: value}
```
turbolent, line 93: `interpreter.SomeValue{nil}` is not valid, a Cadence `nil` is represented by `interpreter.NilValue`:

```suggestion
	if v.Value == nil {
	    return runtime.Value{interpreter.NilValue{}}
	}
	
	innerValue := v.Value.ToRuntimeValue().Value
	value := interpreter.NewSomeValueOwningNonCopying(innerValue)
	return runtime.Value{Value: value}
```

Maybe also just have a `cadence.Nil` instead of a `cadence.Optional{nil}`?

values.go
```diff
@@ -786,3 +929,19 @@ func (v Event) ToGoValue() interface{} {
...
+func compositeToRuntimeValue(fieldTypes []Field, fieldValues []Value) runtime.Value {
+	fields := make(map[string]interpreter.Value, len(fieldTypes))
+
+	for i := 0; i < len(fieldTypes) && i < len(fieldValues); i++ {
+		fieldType := fieldTypes[i]
+		fieldValue := fieldValues[i]
+		fields[fieldType.Identifier] = fieldValue.ToRuntimeValue().Value
+	}
+
+	return runtime.Value{Value: &interpreter.CompositeValue{Fields: fields}}
```
turbolent, line 946: Just the fields are unfortunately not enough, `Location`, `TypeID`, and `Kind` are needed as well: This information is used e.g. to link in the composite value's functions

convert_values_test.go
```diff
@@ -30,209 +30,185 @@ import (
...
+		label:       "Nil",
+		value:       interpreter.NilValue{},
+		expected:    NewOptional(nil),
+		skipReverse: true,
+	},
+	{
+		label:    "SomeValue nil",
+		value:    &interpreter.SomeValue{Value: nil},
+		expected: NewOptional(nil),
+	},
```
turbolent, line 54: The case "`interpreter.SomeValue` with `nil`" doesn't really exist, it's basically the "Nil" case above, i.e. Cadence `nil` is represented as `interpreter.NilValue`.

Maybe introduce a `cadence.Nil` instead of `cadence.Optional{nil}`, or remove this conversion/test case here and don't skip reverse for the "Nil" case above

psiemens:
Now obsolete, replaced with https://github.com/onflow/cadence/pull/39

PR #30, opened 2020-04-25
Fix type requirements

turbolent:
Closes dapperlabs/flow-go#3177

## Description

Type requirements were broken in several ways. 

Interfaces and composites can be nested for a while now; however the equality check was not updated to consider nesting and location. This lead to implementations of type requirements being able to use more restrictive types or even other types. 

This made the example fungible contract pass, even though it should not. For example, the `deposit` method in the `ExampleToken` implementation specified a more restrictive type `ExampleToken.Vault` instead of the required `FungibleToken.Vault`. 

As failable casting in the interpreter uses the subtyping function of the checker, this also erroneously enabled type confusion: Any type named "Vault" could be casted to e.g. `FungibleToken.Vault` or `ExampleToken.Vault`.

Fixing the equality methods for interface and composite types fixed these bugs, but also uncovered that currently composites conforming to type requirements (e.g. `ExampleToken.Vault` to `FungibleToken.Vault`) were not actually subtypes – the bug above allowed the subtyping to succeed.

Fix the missing subtyping relation by recording composite types' implicit conformances to type requirements in the composite type, just like explicit conformances to interfaces, and add a subtyping rule between composite types that consideres these implicit type requirements conformances.

This is likely hard to be reviewed, but this is a very important bugfix and a review needs to be thorough. I'm more than happy to go over it step by step in a call.

### DIFF
--- a/runtime/sema/check_composite_declaration.go
+++ b/runtime/sema/check_composite_declaration.go
@@ -522,6 +522,29 @@ func (checker *Checker) declareCompositeMembersAndValue(
 				VariableKind:    ast.VariableKindConstant,
 			}
 		}
+
+		// Declare implicit type requirement conformances, if any,
+		// after nested types are declared, and
+		// after explicit conformances are declared.
+		//
+		// For each nested composite type, check if a conformance
+		// declares a nested composite type with the same identifier,
+		// in which case it is a type requirement,
+		// and this nested composite type implicitly conforms to it.
+
+		for nestedTypeIdentifier, nestedType := range compositeType.NestedTypes() {
+			nestedCompositeType, ok := nestedType.(*CompositeType)
+			if !ok {
+				continue
+			}
+
+			for _, compositeTypeConformance := range compositeType.ExplicitInterfaceConformances {
+				conformanceNestedTypes := compositeTypeConformance.NestedTypes()
+				if typeRequirement, ok := conformanceNestedTypes[nestedTypeIdentifier].(*CompositeType); ok {
+					nestedCompositeType.AddImplicitTypeRequirementConformance(typeRequirement)
+				}
+			}
+		}
 	})()
 
 	// Always determine composite constructor type
--- a/runtime/sema/check_composite_declaration.go
+++ b/runtime/sema/check_composite_declaration.go
@@ -135,7 +135,7 @@ func (checker *Checker) visitCompositeDeclaration(declaration *ast.CompositeDecl
 
 	checkMissingMembers := kind != ContainerKindInterface
 
-	for i, interfaceType := range compositeType.Conformances {
+	for i, interfaceType := range compositeType.ExplicitInterfaceConformances {
 		interfaceNominalType := declaration.Conformances[i]
 
 		checker.checkCompositeConformance(
@@ -463,8 +463,8 @@ func (checker *Checker) declareCompositeMembersAndValue(
 
 		// Resolve conformances
 
-		conformances := checker.conformances(declaration, compositeType)
-		compositeType.Conformances = conformances
+		conformances := checker.explicitInterfaceConformances(declaration, compositeType)
+		compositeType.ExplicitInterfaceConformances = conformances
 
 		// NOTE: determine initializer parameter types while nested types are in scope,
 		// and after declaring nested types as the initializer may use nested type in parameters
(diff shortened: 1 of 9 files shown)


runtime/sema/check_composite_declaration.go
```diff
@@ -522,6 +522,29 @@ func (checker *Checker) declareCompositeMembersAndValue(
...
+
+		for nestedTypeIdentifier, nestedType := range compositeType.NestedTypes() {
+			nestedCompositeType, ok := nestedType.(*CompositeType)
+			if !ok {
+				continue
+			}
+
+			for _, compositeTypeConformance := range compositeType.ExplicitInterfaceConformances {
+				conformanceNestedTypes := compositeTypeConformance.NestedTypes()
+				if typeRequirement, ok := conformanceNestedTypes[nestedTypeIdentifier].(*CompositeType); ok {
```
benjaminkvm, line 543: Could there be a case where the `conformanceNestedTypes` do not contain the a type identifier that matches `nestedTypeIdentifier`?

I assume it is it a pre-condition (not Cadence pre-condition heh) that the `compositeTypeConformance` nested types are a subset of the `compositeType` nested types.

runtime/tests/checker/conformance_test.go
```diff
@@ -45,3 +46,200 @@ func TestCheckInvalidEventTypeRequirementConformance(t *testing.T) {
...
+            `,
+			`
+              pub struct S {
+                  fun foo() {}
+              }
+            `,
+			true,
+		},
+		{
+			"Conformance with missing function",
```
benjaminkvm, line 95: Duplicate test case

runtime/tests/checker/conformance_test.go
```diff
@@ -45,3 +46,200 @@ func TestCheckInvalidEventTypeRequirementConformance(t *testing.T) {
...
```
benjaminkvm, line 82: This is the same test case with a different test name, was this intended or just copypasta?

runtime/tests/examples/examples.go
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package examples
+
```
benjaminkvm, line 20: Thanks for breaking these out for broader use!

benjaminkvm:
Read through the PR as thoroughly as I could (the test cases helped a lot), would be great to have you do a quick walkthrough at some point!

runtime/tests/checker/conformance_test.go
turbolent: Copypasta 😄  `foo` needs to be moved up and the test failing

runtime/tests/checker/conformance_test.go
turbolent: Fixed in 6b0124b6d2af5448b9909cbe21a38e0130c7816d

runtime/tests/checker/conformance_test.go
turbolent: Fixed in 6b0124b6d2af5448b9909cbe21a38e0130c7816d

runtime/sema/check_composite_declaration.go
turbolent: Yes, this is the case when the conformance does not have a type requirement, e.g.

```
contract interface CI {}

contract C: CI {
    resource R {}
}
```
 
Here, `declareCompositeMembersAndValue` is called for `C`. It has one nested type `R` (first for loop above), which is a composie. We then look in all explicit interface conformances (`C` specifies it conforms to `CI`, and check if it has a nested type with the same name `R` – it does not, so the inner-most if-statement / line below / `AddImplicitTypeRequirementConformance` isn't called: `R` doesn't implicitly conform to any type requirements.

However, if `CI` would have a type requirement, then `C.R` implicitly conforms to it / `CI.R`, i.e. we call `AddImplicitTypeRequirementConformance` on `C.R` with argument `CI.R`

```
contract interface CI {
    resource R {}
}
```

PR #31, opened 2020-04-25
Reset the resource return info when checking a function

turbolent:
Closes dapperlabs/flow-go#3407
Depends on #30

## Description

The resource invalidation info is global, but contains one flag for the return status of the function. We've set this correctly, but forgot to reset it for each function. This caused functions to influence the return info of other functions, and therefore also the resource invalidation checks, leading to the missing resource loss error.

### DIFF
--- a/runtime/tests/checker/resources_test.go
+++ b/runtime/tests/checker/resources_test.go
@@ -3545,18 +3545,17 @@ func TestCheckInvalidResourceOptionalBindingFailableCastResourceUseAfterInvalida
 
 func TestCheckInvalidResourceOptionalBindingFailableCastResourceLossMissingElse(t *testing.T) {
 
-	_, err := ParseAndCheck(t,
-		`
-         resource interface RI {}
+	_, err := ParseAndCheck(t, `
+      resource interface RI {}
 
-         resource R: RI {}
+      resource R: RI {}
 
-         fun test() {
-             let ri: @AnyResource{RI} <- create R()
-             if let r <- ri as? @R {
-                 destroy r
-             }
-         }
+      fun test() {
+          let ri: @{RI} <- create R()
+          if let r <- ri as? @R {
+              destroy r
+          }
+      }
     `)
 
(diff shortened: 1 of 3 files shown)


runtime/tests/checker/resources_test.go
```diff
@@ -3571,39 +3570,82 @@ func TestCheckInvalidResourceOptionalBindingFailableCastResourceLossMissingElse(
 
 func TestCheckInvalidResourceOptionalBindingFailableCastResourceUseAfterInvalidationAfterThen(t *testing.T) {
```
benjaminkvm, line 3571: Took a moment for me to parse this test case name heh.

runtime/tests/checker/resources_test.go
```diff
@@ -3984,3 +4026,35 @@ func TestCheckInvalidOptionalResourceCoalescingRightSide(t *testing.T) {
...
+          resource R {
+
+              fun foo(r: @R) {
+                  if let r2 <- r as? @R {
+                      destroy r2
+                  }
+              }
+          }
+
+          pub fun bar() {
```
benjaminkvm, line 4050: So previously, this would have lead to `bar()` not passing checking due to a resource loss error, correct?

runtime/tests/checker/resources_test.go
turbolent: Yeah, this is pretty bad. It would be nice to run `t.Run` with descriptive names for more and more tests

runtime/tests/checker/resources_test.go
turbolent: No, actually it's the other way round: `bar` is checked first, and the `return` statement set the `Returns` flag. Then `foo` is checked and the `Returns` flag is already set to true, so the whole of `foo` looks like dead-code to the resource tracking code: the function has already returned, the function is therefore valid – which is incorrect, because the flag was not reset and there is actually a resource loss of `r` in `foo` if the conditional cast fails.

PR #32, opened 2020-04-26
Add an explanation for access modifier errors

turbolent:
## Description

Currently, error messages for incorrect or missing access modifiers do not explain why the modifier is wrong, just that it is wrong.

For example, when using the access modifier `access(contract)` for a nested struct in a contract, the error message is:

> invalid access modifier for structure: `access(contract)`

This PR adds an optional explanation to such errors. 

The above example is now reported as:
  
> invalid access modifier for structure: `access(contract)`. type declarations must be public for now

### DIFF
--- a/runtime/sema/checker.go
+++ b/runtime/sema/checker.go
@@ -1656,6 +1656,8 @@ func (checker *Checker) ResetErrors() {
 	checker.errors = nil
 }
 
+const invalidTypeDeclarationAccessModifierExplanation = "type declarations must be public for now"
+
 func (checker *Checker) checkDeclarationAccessModifier(
 	access ast.Access,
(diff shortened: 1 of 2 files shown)


runtime/sema/checker.go
```diff
@@ -1656,6 +1656,8 @@ func (checker *Checker) ResetErrors() {
 	checker.errors = nil
 }
 
+const invalidTypeDeclarationAccessModifierExplanation = "type declarations must be public for now"
```
joshuahannan, line 1659:
```suggestion
const invalidTypeDeclarationAccessModifierExplanation = "type declarations must be public"
```

I feel like it doesn't really need the `for now` part.

PR #37, opened 2020-04-27
Support caching of program ASTs

benjaminkvm:
Closes https://github.com/dapperlabs/flow-go/issues/3401.

## Description

Provide `runtime.Interface` methods to support caching the program AST of parsed programs. This should expedite loading of previously parsed programs (by skipping the parse phase). An error returned from `GetCachedProgram` (a non-fatal error) will result in the input program being parsed.

### DIFF
--- a/runtime/interface.go
+++ b/runtime/interface.go
@@ -19,8 +19,6 @@
 package runtime
 
 import (
-	"errors"
-
 	"github.com/onflow/cadence/runtime/ast"
 )
 
@@ -67,7 +65,7 @@ func (i *EmptyRuntimeInterface) ResolveImport(location Location) ([]byte, error)
 	return nil, nil
 }
 func (i *EmptyRuntimeInterface) GetCachedProgram(location Location) (*ast.Program, error) {
-	return nil, errors.New("cache not implemented")
+	return nil, nil
 }
 
 func (i *EmptyRuntimeInterface) CacheProgram(location Location, program *ast.Program) error {
--- a/runtime/interface.go
+++ b/runtime/interface.go
@@ -64,6 +64,7 @@ type EmptyRuntimeInterface struct{}
 func (i *EmptyRuntimeInterface) ResolveImport(location Location) ([]byte, error) {
 	return nil, nil
 }
+
 func (i *EmptyRuntimeInterface) GetCachedProgram(location Location) (*ast.Program, error) {
 	return nil, nil
 }
--- a/runtime/interface.go
+++ b/runtime/interface.go
@@ -18,9 +18,19 @@
 
 package runtime
 
+import (
+	"errors"
+
+	"github.com/onflow/cadence/runtime/ast"
+)
+
 type Interface interface {
 	// ResolveImport resolves an import of a program.
 	ResolveImport(Location) ([]byte, error)
+	// GetCachedProgram attempts to get a parsed program from a cache.
+	GetCachedProgram(Location) (*ast.Program, error)
+	// CacheProgram adds a parsed program to a cache.
+	CacheProgram(Location, *ast.Program) error
 	// GetValue gets a value for the given key in the storage, controlled and owned by the given accounts.
 	GetValue(owner, controller, key []byte) (value []byte, err error)
 	// SetValue sets a value for the given key in the storage, controlled and owned by the given accounts.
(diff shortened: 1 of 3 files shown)


runtime/runtime.go
```diff
@@ -277,9 +277,14 @@ func (r *interpreterRuntime) parseAndCheckProgram(
...
+
+	// If the program has not been previously cached, proceed to parse the program.
+	program, err := runtimeInterface.GetCachedProgram(location)
 	if err != nil {
-		return nil, err
+		program, err = r.parse(code)
+		if err != nil {
+			return nil, err
+		}
 	}
```
turbolent, line 288: I'd assume `GetCachedProgram` would return `nil`, `nil` for the case where nothing is cached, but also nothing went wrong during the cache lookup. 

Maybe change this to check program instead, and keep the second error return value for actual errors:

```go
if err != nil {
	return nil, err
}

if program == nil {
	program, err = r.parse(code)
	if err != nil {
		return nil, err
	}
}
```

runtime/runtime_test.go
```diff
@@ -82,6 +85,14 @@ func (i *testRuntimeInterface) ResolveImport(location Location) ([]byte, error)
 	return i.resolveImport(location)
 }
 
+func (i *testRuntimeInterface) GetCachedProgram(location Location) (*ast.Program, error) {
+	return i.getCachedProgram(location)
+}
+
+func (i *testRuntimeInterface) CacheProgram(location Location, program *ast.Program) error {
+	return i.cacheProgram(location, program)
+}
```
turbolent, line 94: Maybe instead of adding empty implementations for `getCachedProgram` and `cacheProgram` to all the tests (e.g. https://github.com/onflow/cadence/pull/37/files#diff-991fa0d56f6a51df18449cd1f05bbbcaR249-R254), maybe just make the two functions optional (and then remove all additions starting with https://github.com/onflow/cadence/pull/37/files#diff-991fa0d56f6a51df18449cd1f05bbbcaR249)

```suggestion
func (i *testRuntimeInterface) GetCachedProgram(location Location) (*ast.Program, error) {
	if i.getCachedProgram == nil {
		return nil, nil
	}
	return i.getCachedProgram(location)
}

func (i *testRuntimeInterface) CacheProgram(location Location, program *ast.Program) error {
	if i.cacheProgram == nil {
		return nil
	}
	return i.cacheProgram(location, program)
}
```

runtime/interface.go
```diff
@@ -56,6 +66,13 @@ type EmptyRuntimeInterface struct{}
 func (i *EmptyRuntimeInterface) ResolveImport(location Location) ([]byte, error) {
 	return nil, nil
 }
+func (i *EmptyRuntimeInterface) GetCachedProgram(location Location) (*ast.Program, error) {
+	return nil, errors.New("cache not implemented")
```
turbolent, line 70: I think this should not return an error: if there's nothing cached, that's not an error:

```suggestion
	return nil, nil
```

runtime/runtime_test.go
```diff
@@ -175,6 +192,49 @@ func TestRuntimeImport(t *testing.T) {
...
+	var cacheHitSuccess bool
+	programCache := map[ast.LocationID]*ast.Program{}
+
+	runtime := NewInterpreterRuntime()
+	runtimeInterface := &testRuntimeInterface{
+		getCachedProgram: func(location ast.Location) (*ast.Program, error) {
+			program, found := programCache[location.ID()]
+			cacheHitSuccess = found
+			if !found {
+				return nil, errors.New("no cache entry")
```
turbolent, line 205:
```suggestion
				return nil, nil
```

runtime/runtime_test.go
```diff
@@ -175,6 +192,49 @@ func TestRuntimeImport(t *testing.T) {
 	assert.Equal(t, interpreter.NewIntValueFromInt64(42), value.Value)
 }
 
+func TestRuntimeCachedProgram(t *testing.T) {
```
turbolent, line 195: Nice! It would be great to extend this test to also import another program and verify that the cache is used for it

runtime/runtime.go
benjaminkvm: Agreed, was kind of playing with what the proper type signature would be for this. Definitely simplifies things 👍

runtime/runtime_test.go
benjaminkvm: Sounds good!

runtime/interface.go
benjaminkvm: Agreed 👍

benjaminkvm: I was originally planning to use that for bubbling up cache errors, going to maintain this signature as per your following comment.

runtime/runtime_test.go
```diff
@@ -175,6 +191,113 @@ func TestRuntimeImport(t *testing.T) {
...
+				return nil, nil
+			}
+			return program, nil
+		},
+		cacheProgram: func(location ast.Location, program *ast.Program) error {
+			programCache[location.ID()] = program
+			return nil
+		},
+		resolveImport: func(location Location) ([]byte, error) {
+			fmt.Println(location.ID())
```
turbolent, line 222:
```suggestion
```

PR #39, opened 2020-04-27
Runtime Values: Add ToRuntimeValue function

psiemens:
**Part 2 of 3**

This PR adds a `runtime.ToRuntimeValue` function that converts a `cadence.Value` type to a `runtime.Value` type.

This PR copies the same logic as this PR, but instead keeps value conversion inside the `runtime` package: https://github.com/onflow/cadence/pull/29

### DIFF
--- a/runtime/convertValues.go
+++ b/runtime/convertValues.go
@@ -44,6 +44,108 @@ func ConvertEvent(event Event) cadence.Event {
 	return cadence.NewEvent(fields).WithType(ConvertType(event.Type).(cadence.EventType))
 }
 
+// ToRuntimeValue converts a Cadence value to a runtime value.
+func ToRuntimeValue(value cadence.Value) Value {
+	switch v := value.(type) {
+	case cadence.Void:
+		return Value{Value: interpreter.VoidValue{}}
+	case cadence.Optional:
+		if v.Value == nil {
+			return Value{Value: interpreter.NilValue{}}
+		}
+
+		innerValue := ToRuntimeValue(v.Value).Value
+		value := interpreter.NewSomeValueOwningNonCopying(innerValue)
+
+		return Value{Value: value}
+	case cadence.Bool:
+		return Value{Value: interpreter.BoolValue(v)}
+	case cadence.String:
+		return Value{Value: interpreter.NewStringValue(string(v))}
+	case cadence.Bytes:
+		return Value{Value: interpreter.ByteSliceToByteArrayValue(v)}
+	case cadence.Address:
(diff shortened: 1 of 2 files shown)


runtime/convertValues_test.go
```diff
@@ -31,19 +31,21 @@ import (
...
 		expected: cadence.NewVoid(),
 	},
 	{
-		label:    "Nil",
-		value:    interpreter.NilValue{},
-		expected: cadence.NewOptional(nil),
+		label:       "Nil",
+		value:       interpreter.NilValue{},
+		expected:    cadence.NewOptional(nil),
+		skipReverse: true,
```
turbolent, line 48: Why is this test case skipped? A Cadence `nil` seems to be represented as `cadence.NewOptional(nil)`, which should convert to `interpreter.NilValue{}`

runtime/convertValues.go
```diff
@@ -44,6 +44,108 @@ func ConvertEvent(event Event) cadence.Event {
...
+func compositeToRuntimeValue(fieldTypes []cadence.Field, fieldValues []cadence.Value) Value {
+	fields := make(map[string]interpreter.Value, len(fieldTypes))
+
+	for i := 0; i < len(fieldTypes) && i < len(fieldValues); i++ {
+		fieldType := fieldTypes[i]
+		fieldValue := fieldValues[i]
+		fields[fieldType.Identifier] = ToRuntimeValue(fieldValue).Value
+	}
+
+	return Value{Value: &interpreter.CompositeValue{Fields: fields}}
```
turbolent, line 146: Just the fields are unfortunately not enough, `Location`, `TypeID`, and `Kind` are needed as well: This information is used e.g. to link in the composite value's functions

runtime/convertValues_test.go
psiemens: I'm going to remove this test case in a follow-up PR. I liked your idea to introduce a `cadence.Nil` value, so I'l add that and then fix the implementation if that sounds good?

runtime/convertValues.go
psiemens: Yeah, I have a proposed solution for this in a follow-up PR: https://github.com/onflow/cadence/pull/45/files#diff-4893a68aef7c913546919a6adf6c0885R142-R144

Let me know what you think!

turbolent: Ah, I see! Sorry, I hadn't seen it yet 👍

runtime/convertValues_test.go
turbolent: Sounds good!

PR #40, opened 2020-04-27
Runtime Values: Remove runtime.Value, replace with cadence.Value

psiemens:
Depends on #39

**Part 3 of 3**

This PR updates the `runtime` package to remove the need for the public `runtime.Value`, `runtime.Event`, and `runtime.Type` types. These types were being used to shield the caller from the internal `interpreter.Value` and `sema.Type` types, but now the runtime simply uses the public `cadence.Value` and `cadence.Event` types directly.

- Remove `runtime.Value` and `runtime.Event` from all public `runtime` interfaces -- use `cadence.Value` and `cadence.Event` instead
- Rename `runtime.Value` to `runtime.exportableValue`, make private (since no longer exposed through public interfaces)
- Rename `runtime.ConvertValue` => `runtime.exportValue`, make private
- Rename `runtime.ConvertEvent` => `runtime.exportEvent`, make private
- Rename `runtime.ConvertType` => `runtime.exportType`, make private


runtime/interpreter/interpreter.go
```diff
@@ -3377,7 +3375,7 @@ func (interpreter *Interpreter) newConverterFunction(converter ValueConverter) F
 // - PublicAccount
 // - Block
 
-func (interpreter *Interpreter) IsSubType(subType DynamicType, superType sema.Type) bool {
```
turbolent, line 3380: Nice!

PR #42, opened 2020-04-28
Improve access modifier checks

### LINKED ISSUE #41: BUG: Contract Interface conformance does not register
### Instructions

When requiring a `access(self)` field in a resource in a contract interface, if the resource in the implementation has the same access control modifier, an error is reported. There should be no error.

### Steps to Reproduce

1. Use this contract:
```
pub contract interface ContractTypeConform {
    pub resource Res {
        access(self) var private: Int
    }
}

pub contract Example: ContractTypeConform {
    pub resource Res {
        access(self) var private: Int

        init () {
            self.private = 0
        }
    }
}
```
2. See that the resource `Res` in `Example` does not match the interface according to the static checker.

### Acceptance Criteria

The resource must conform correctly

### Context

Making nft storage fields private in an NFT collection

turbolent:
Closes #41

## Description

- Report a separate error when an interface member is private, instead of just failing the conformance check. There was no bug in the checking here, this improves UX
- Check the *effective* access when checking conformance, not just the explicit access: If a composite does not specify an access modifier for a member, the access check mode must be used to determine how it will be effectively seen
- Fix error notes
- Report conformance error member mismatches as error notes:
  ```
  error: resource `Example.R` does not conform to resource type requirement `C.R`
   --> test.cdc:8:23
    |
  8 |           pub resource R {
    |                        ^
    |
  9 |               access(self) var x: Int
    |                                - mismatch here
  ```

### DIFF
--- a/runtime/sema/checker.go
+++ b/runtime/sema/checker.go
@@ -2040,3 +2040,39 @@ func (checker *Checker) ValueActivationDepth() int {
 func (checker *Checker) TypeActivationDepth() int {
 	return checker.typeActivations.Depth()
 }
+
+func (checker *Checker) effectiveMemberAccess(access ast.Access, containerKind ContainerKind) ast.Access {
+	switch containerKind {
+	case ContainerKindComposite:
+		return checker.effectiveCompositeMemberAccess(access)
+	case ContainerKindInterface:
+		return checker.effectiveInterfaceMemberAccess(access)
+	default:
+		panic(errors.NewUnreachableError())
+	}
+}
+
+func (checker *Checker) effectiveInterfaceMemberAccess(access ast.Access) ast.Access {
+	if access == ast.AccessNotSpecified {
+		return ast.AccessPublic
+	} else {
+		return access
+	}
+}
+
+func (checker *Checker) effectiveCompositeMemberAccess(access ast.Access) ast.Access {
+	if access != ast.AccessNotSpecified {
+		return access
+	}
+
+	switch checker.accessCheckMode {
+	case AccessCheckModeStrict, AccessCheckModeNotSpecifiedRestricted:
+		return ast.AccessPrivate
+
+	case AccessCheckModeNotSpecifiedUnrestricted, AccessCheckModeNone:
+		return ast.AccessPublic
+
+	default:
+		panic(errors.NewUnreachableError())
+	}
+}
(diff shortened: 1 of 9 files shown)


runtime/tests/checker/access_test.go
```diff
@@ -95,47 +108,112 @@ func TestCheckAccessModifierCompositeFunctionDeclaration(t *testing.T) {
...
+					),
+				)
+
+				if expectSuccess {
+					assert.NoError(t, err)
 				} else {
-					body = "{}"
+					errs := ExpectCheckerErrors(t, err, 1)
+
+					assert.IsType(t, &sema.InvalidAccessModifierError{}, errs[0])
```
psiemens, line 140:
```suggestion
					expectInvalidAccessModifierError(t, err)
```

runtime/tests/checker/access_test.go
```diff
@@ -239,7 +325,15 @@ func TestCheckAccessModifierCompositeVariableFieldDeclaration(t *testing.T) {
 						),
 					)
 
-					assert.NoError(t, err)
+					// private fields in interfaces are invalid
+
+					if isInterface && access == ast.AccessPrivate {
+						errs := ExpectCheckerErrors(t, err, 1)
+
+						assert.IsType(t, &sema.InvalidAccessModifierError{}, errs[0])
```
psiemens, line 333:
```suggestion
						expectInvalidAccessModifierError(t, err)
```

runtime/tests/checker/access_test.go
psiemens: Not sure if this is necessary (it kind of hides details about the test), but figured I'd suggest anyways

runtime/sema/checker.go
```diff
@@ -2040,3 +2040,39 @@ func (checker *Checker) ValueActivationDepth() int {
...
+	case AccessCheckModeStrict, AccessCheckModeNotSpecifiedRestricted:
+		return ast.AccessPrivate
+
+	case AccessCheckModeNotSpecifiedUnrestricted, AccessCheckModeNone:
+		return ast.AccessPublic
+
+	default:
+		panic(errors.NewUnreachableError())
+	}
+}
```
psiemens, line 2078: Nice 👍

runtime/tests/checker/access_test.go
turbolent: Great idea! I've looked for reusable expectation functions and fixed them all up in f347e69a1af1d42c10180f285eb7e85cda200351

runtime/tests/checker/access_test.go
turbolent: Fixed in f347e69a1af1d42c10180f285eb7e85cda200351

turbolent:
Used the existing expectation functions in f347e69

PR #43, opened 2020-04-28
Update interpreter.Invoke functions to accept Value instead of interface{}

psiemens:
Woork towards https://github.com/dapperlabs/flow-go/issues/3286

## Description

In order to support transaction arguments, the interpreter needs to accept all valid value types as inputs. To accomplish this, I updated the `Invoke` functions to accept arguments as `interpreter.Value` rather than `interface{}`. Combined with the new `runtime.importValue` function, we can import and prepare values before passing them into the interpreter.


turbolent:
Great idea! 👍 Thanks for cleaning this up

turbolent:
Argh, I messed up and hit merge before I saw this is dependent on another PR :(

PR #44, opened 2020-04-28
Fix number conversion

turbolent:
Closes dapperlabs/flow-go#2141

## Description

- Infer negative fixed-point literals to `Fix64`, and `UFix64` otherwise
- Check type compatibility of expression statements: 
  For example, an expression statement might be a fixed-point literal, 
  e.g. in the REPL, which might be inferred to `Fix64` or `UFix64`,
  but might exceed the valid range of the type
- Fix the conversion of numbers to `[U]Fix64`: Properly check integer ranges, handle "big" number values (values >`math.Max[u]int64`)
- Move integer and fixed-point interpreter tests to separate files

### DIFF
--- a/runtime/tests/interpreter/fixedpoint_test.go
+++ b/runtime/tests/interpreter/fixedpoint_test.go
@@ -1,0 +1,479 @@
+/*
+ * Cadence - The resource-oriented smart contract programming language
+ *
+ * Copyright 2019-2020 Dapper Labs, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package interpreter_test
+
+import (
+	"fmt"
+	"math/big"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/onflow/cadence/runtime/interpreter"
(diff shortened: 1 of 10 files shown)


runtime/tests/interpreter/fixedpoint_test.go
```diff
@@ -0,0 +1,479 @@
...
+	tests := map[string]interpreter.Value{
+		// Fix*
+		"Fix64": interpreter.Fix64Value(123000000),
+		// UFix*
+		"UFix64": interpreter.UFix64Value(123000000),
+	}
+
+	for _, fixedPointType := range sema.AllFixedPointTypes {
+		if _, ok := tests[fixedPointType.String()]; !ok {
+			panic(fmt.Sprintf("broken test: missing %s", fixedPointType))
```
psiemens, line 56: 👍

runtime/tests/interpreter/fixedpoint_test.go
```diff
@@ -0,0 +1,479 @@
...
+
+		const testedValue = sema.Fix64TypeMaxInt + 1
+		testValueBig := big.NewInt(0).SetUint64(testedValue)
+
+		for _, integerType := range sema.AllIntegerTypes {
+
+			// Only test for integer types that can hold testedValue
+
+			maxInt := integerType.(sema.IntegerRangedType).MaxInt()
+			if maxInt != nil && maxInt.Cmp(testValueBig) < 0 {
```
psiemens, line 416: 👍

runtime/tests/interpreter/integers_test.go
```diff
@@ -0,0 +1,348 @@
...
+		inter.Globals["x"].Value,
+	)
+
+	assert.Equal(t,
+		interpreter.AddressValue{
+			0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x2,
+		},
+		inter.Globals["y"].Value,
+	)
+}
```
psiemens, line 120: Does this test belong in this file?

turbolent: I moved it out into this "integer" test file, because it is basically also just an integer.

PR #50, opened 2020-04-29
Fix license headers

turbolent:
- Add missing license headers
- Check on CI if any Go file is missing a license header

### DIFF
--- a/check-headers.sh
+++ b/check-headers.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+files=$(find . -name \*.go -type f -print0 | xargs -0 egrep -L '(Licensed under the Apache License)|(Code generated from|by)')
+if [ -n "$files" ]; then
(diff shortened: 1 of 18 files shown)


check-headers.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+files=$(find . -name \*.go -type f -print0 | xargs -0 grep -L 'Licensed under the Apache License')
+if [ -n "$files" ]; then
+  echo "Missing license header in:"
+  echo "$files"
+  exit 1
+fi
```
psiemens, line 8: Nice, I'm going to steal this 😄

turbolent: It's not quite useful yet. It doesn't detect generated files, which won't have the license header, I need to add that

turbolent: @psiemens improved the script to ignore generated files

PR #55, opened 2020-05-01
"Virtual" imports

turbolent:
Add support for "virtual" imports, imports that are not backed by Cadence programs, but by Go functions.

This is foundation work for dapperlabs/flow-go#3287, dapperlabs/flow-go#550, and dapperlabs/flow-go#3243.

This enables code like this:

```Swift
import Crypto

fun test(): UInt64 {
    return Crypto.unsafeRandom()
}
```


turbolent:
Brought it up-to-date with `master`, but need to investigate why parsing is duplicated

turbolent:
@benjaminkvm I fixed the duplicate parsing issue I had found/mentioned above in cf4b379, so this should work now and is ready to go

runtime/ast/import.go
```diff
@@ -98,20 +97,25 @@ func LocationsMatch(first, second Location) bool {
...
+		return IdentifierLocation(pieces[1])
+
+	case StringLocationPrefix:
+		return StringLocation(pieces[1])
 
-			return AddressLocation(address)
+	case AddressLocationPrefix:
+		address, err := hex.DecodeString(pieces[1])
+		if err != nil {
+			return nil
```
psiemens, line 114: Do you think we could get rid of location prefixes? I know I added them in the first place, but I've realized they don't make much sense to the end-user and are just confusing overall.

Also, I feel like I'm often conflating two things when I think about locations:
- A place where Cadence code lives (let's say `origin`)
- An _identifier_ for a place where Cadence code lives (let's say `path`)

For example, say we are executing a transaction. Currently we pass in a `TransactionLocation` instance that just tells the runtime "this code originates from a transaction". In order words, the program's `origin` is the transaction. Because locations are also used for imports, we assume that every `origin` has a `path`, but this isn't really the case for transactions and scripts, which can't just be "imported".

What if we allowed origins to be anonymous? Something like this:

```go
type Origin interface {
  String() string // returns ID for the origin, but this isn't necessarily the import path
}

type Importer interface {
  Resolve(loc Location) (Program, Origin)
}

File: runtime/sema/check_import_declaration.go [hunk 2]
```diff
@@ -66,58 +52,68 @@ func (checker *Checker) declareImportDeclaration(declaration *ast.ImportDeclarat
...
+					Pos:            declaration.LocationPos,
+				},
+			)
+			return nil
+		}
+
+		imp = CheckerImport{importChecker}
+
+	} else if checker.importHandler != nil {
+		imp = checker.importHandler(location)
```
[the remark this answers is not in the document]
turbolent: Correct 👍 The import handler may return a "virtual" import here, i.e. one that is not backed by a Cadence program

runtime/ast/import.go
turbolent: The prefixes are used in the location *IDs*, which not meant for end-users, but for serialization: We need a way to differentiate where types are defined, and do so in a way that it can be serialized. The location IDs should never be shown to a user, if we do we should fix that.

Sorry, I'm not quite following with the origins/identifiers/paths, maybe we can chat about this in a call?

psiemens: Yep, sounds good! This isn't a blocker for this PR so we can talk about it separately.

PR #56, opened 2020-05-02
Switch from Gob to CBOR encoding

turbolent:
Closes dapperlabs/flow-go#2814

@benjaminkvm 🙌 

This PR switches encoding of values completely from Gob to CBOR.
A new encoder with support for all value types and static types replaces the old Gob registrations.

We should probably completely remove the XDR encoding after this, as it is not needed anymore and is outdated.


runtime/interpreter/encode.go
```diff
@@ -0,0 +1,899 @@
...
+		cborTagDictionaryStaticType:    encodedDictionaryStaticType{},
+		cborTagRestrictedStaticType:    encodedRestrictedStaticType{},
+		cborTagReferenceStaticType:     encodedReferenceStaticType{},
+	}
+
+	// Register types
+	for tag, encodedType := range types {
+		register(tag, encodedType)
+	}
+}
```
psiemens, line 192: Very nice 👏 @benjaminkvm

runtime/interpreter/encode.go
```diff
@@ -0,0 +1,899 @@
...
+
+	return result, nil
+}
+
+type encodedDictionaryValue struct {
+	Keys    []interface{}          `cbor:"0,keyasint"`
+	Entries map[string]interface{} `cbor:"1,keyasint"`
+}
+
+// TODO: optimize: use CBOR map, but unclear how to preserve ordering
```
psiemens, line 565: I think you might be able to use `SortBytewiseLexical ` from https://github.com/fxamacker/cbor#encoding-options-1

runtime/interpreter/encoding_test.go
```diff
@@ -0,0 +1,1938 @@
...
+	"github.com/onflow/cadence/runtime/tests/utils"
+)
+
+type encodeDecodeTest struct {
+	value   Value
+	encoded []byte
+	invalid bool
+}
+
+func testEncodeDecode(t *testing.T, tests map[string]encodeDecodeTest) {
```
psiemens, line 24: It might be a good idea to repeat each test case in a loop (like a primitive fuzz test) to ensure that the encoding is fully deterministic, as done here: https://github.com/onflow/cadence/blob/master/encoding/xdr/encoding_test.go#L628-L633

runtime/interpreter/primitivestatictype.go
```diff
@@ -0,0 +1,370 @@
...
+	PrimitiveStaticTypeString
+	PrimitiveStaticTypeCharacter
+	_
+	_
+	_
+	_
+	_
+	_
+	_
+	_
```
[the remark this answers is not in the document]
turbolent: No, we don't yet, I just think it's likely we'll add some in the future and we want to group them close to related tags.

runtime/sema/type.go
```diff
@@ -1076,10 +1027,6 @@ type FractionalRangedType interface {
 // IntegerType represents the super-type of all integer types
 type IntegerType struct{}
 
-func init() {
-	gob.Register(&IntegerType{})
```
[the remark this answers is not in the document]
turbolent: 👋

runtime/interpreter/encode.go
turbolent: Ah, I meant preserving the insertion order we keep for the keys of the Cadence dictionary value. 

We could use the byteweise lexical ordering also for the Cadence value 🤔

runtime/interpreter/encoding_test.go
turbolent: Great, idea, adding! 💯

turbolent: Added in 7444398

PR #60, opened 2020-05-04
Expose metrics

turbolent:
Closes dapperlabs/flow-internal#1015

Expose metrics for the following phases and parts:
- Parsing
- Checking
- Interpreting
- Value encoding
- Value decoding

cc @philipstanislaus

### DIFF
--- a/runtime/interface.go
+++ b/runtime/interface.go
@@ -66,6 +66,8 @@ type Metrics interface {
 	ProgramParsed(duration time.Duration)
 	ProgramChecked(duration time.Duration)
 	ProgramInterpreted(duration time.Duration)
+	ValueEncoded(duration time.Duration)
+	ValueDecoded(duration time.Duration)
 }
 
 type EmptyRuntimeInterface struct{}
--- a/runtime/interface.go
+++ b/runtime/interface.go
@@ -19,6 +19,8 @@
 package runtime
 
 import (
+	"time"
+
 	"github.com/onflow/cadence"
 	"github.com/onflow/cadence/runtime/ast"
 )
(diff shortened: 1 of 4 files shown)


runtime/interface.go
```diff
@@ -60,59 +62,67 @@ type Interface interface {
 	DecodeArgument(b []byte, t cadence.Type) (cadence.Value, error)
 }
 
+type Metrics interface {
+	ProgramParsed(duration time.Duration)
+	ProgramChecked(duration time.Duration)
+	ProgramInterpreted(duration time.Duration)
+	ValueEncoded(duration time.Duration)
+	ValueDecoded(duration time.Duration)
+}
```
psiemens, line 71: Could we make this interface report the start and end timestamps rather than the elapsed time? This would allow us to construct traces from these metrics and place them on the larger execution timeline:

Screen Shot 2020-05-05 at 10 43 26 AM

runtime/runtime.go
```diff
@@ -89,6 +89,24 @@ func validTopLevelDeclarations(location ast.Location) []common.DeclarationKind {
...
+		f()
+		return
+	}
+
+	start := time.Now()
+	f()
+	elapsed := time.Since(start)
+
+	report(metrics, elapsed)
+}
```
psiemens, line 108: Nice 👍

runtime/interface.go
turbolent: Great idea! I'll change this quickly

turbolent: Changed in 05c813c8847772df6eb85fa75dae4753c331f74e

PR #63, opened 2020-05-05
Avoid unnecessary writes by tracking modifications

turbolent:
Work towards dapperlabs/flow-go#2884

Track modifications and only write values that are new or modified.
Also, storage references don't need an owner.

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -376,6 +376,8 @@ func (v *ArrayValue) GetOwner() *common.Address {
 }
 
 func (v *ArrayValue) SetOwner(owner *common.Address) {
+	v.modified = true
+
 	if v.Owner == owner {
 		return
 	}
@@ -4467,6 +4469,8 @@ func (v *CompositeValue) GetOwner() *common.Address {
 }
 
 func (v *CompositeValue) SetOwner(owner *common.Address) {
+	v.modified = true
+
 	if v.Owner == owner {
 		return
 	}
@@ -4685,6 +4689,8 @@ func (v *DictionaryValue) GetOwner() *common.Address {
 }
 
 func (v *DictionaryValue) SetOwner(owner *common.Address) {
+	v.modified = true
+
 	if v.Owner == owner {
 		return
 	}
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -322,6 +322,10 @@ func (*StringValue) SetMember(_ *Interpreter, _ LocationRange, _ string, _ Value
 	panic(errors.NewUnreachableError())
 }
 
+func (v *StringValue) SetModified(modified bool) {
+	v.modified = false
+}
+
 // ArrayValue
 
 type ArrayValue struct {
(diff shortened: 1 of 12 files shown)


runtime/interpreter/value.go
```diff
@@ -129,6 +134,10 @@ func (BoolValue) SetOwner(_ *common.Address) {
 	// NO-OP: value cannot be owned
 }
 
+func (BoolValue) Modified() bool {
+	return false
```
psiemens, line 138: Is this false because these values (as well as integers, etc) have copy semantics?

runtime/interpreter/value.go
```diff
@@ -3770,7 +3899,7 @@ func (v *CompositeValue) Destroy(interpreter *Interpreter, locationRange Locatio
 	}
 
 	return tramp.Then(func(_ interface{}) {
-		v.Destroyed = true
+		v.destroyed = true
```
psiemens, line 3902: If it's destroyed is it also modified?

runtime/runtime_storage.go
```diff
@@ -156,38 +171,50 @@ func (s *interpreterRuntimeStorage) writeValue(
...
 }
 
 // writeCached serializes/saves all values in the cache in storage (through the runtime interface).
 //
 func (s *interpreterRuntimeStorage) writeCached() {
 
-	for storageKey, value := range s.cache {
+	for fullKey, entry := range s.cache {
+
+		if !entry.mustWrite && (entry.value == nil || !entry.value.Modified()) {
```
psiemens, line 205: If the cached value is `nil`, is it possible for there to still be a value in storage that must also be set to `nil`?

runtime/interpreter/value.go
turbolent: Good point, yes 👍

turbolent, line 3902:
```suggestion
		v.destroyed = true
		v.modified = true
```

runtime/runtime_storage.go
```diff
@@ -156,38 +171,50 @@ func (s *interpreterRuntimeStorage) writeValue(
...
+	fullKey := storageKey{
 		storageIdentifier: storageIdentifier,
 		key:               key,
 	}
 
 	// Only write the value to the cache.
 	// The Cache is finally written back through the runtime interface in `writeCached`
 
+	entry := s.cache[fullKey]
+	entry.mustWrite = true
```
benjaminkvm, line 183: RE: naming, curious what is better in this case `entry.mustWrite` or something like `entry.modified`/`entry.dirty`. Super minor!

runtime/runtime_storage.go
turbolent: @psiemens As far as I can see, we only have `value == nil` if the value is read and is nil; or it was written by the program and is set to nil, in which case `mustWrite` is set.

Though, I still removed ignoring writes of nil in b479c36 to be on the safe side

turbolent:
@Kay-Zee as discussed I've changed new values to have the modified flag set in f1b0d3b, and also have changed the ownership change functions so that they set the flag in 3c10119

PR #69, opened 2020-05-06
Implement concurrent lexing in the new parser

benjaminkvm:
## Description

Implement concurrent lexing in the new parser, defer closing of token channel on error to a future PR.

All new parser/lexer tests are passing.

### DIFF
--- a/runtime/parser2/parser.go
+++ b/runtime/parser2/parser.go
@@ -26,7 +26,7 @@ import (
 )
 
 type parser struct {
-	tokens  []lexer.Token
+	tokens  chan lexer.Token
 	current lexer.Token
 	pos     int
 	errors  []error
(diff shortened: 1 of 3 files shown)


runtime/parser2/parser.go
```diff
@@ -54,7 +54,8 @@ func (p *parser) report(err error) {
 
 func (p *parser) next() {
 	p.pos++
-	p.current = p.tokens[p.pos]
+	p.current = <-p.tokens
```
turbolent, line 57: Maybe this should use the two-value result form and check if a value was sent.
If the second boolean result value is false, p.current can be set to `TokenEOF`.
Without it, when the channel was closed, this read will return zero-values instantly.

PR #76, opened 2020-05-07
New parser expressions

turbolent:
Already reviewed #75

### DIFF
--- a/runtime/parser2/expression.go
+++ b/runtime/parser2/expression.go
@@ -45,10 +45,23 @@ type infix struct {
 	leftDenotation   infixFunc
 }
 
+type binary struct {
+	tokenType        lexer.TokenType
+	leftBindingPower int
+	rightAssociative bool
+	operation        ast.Operation
+}
+
 type prefix struct {
 	tokenType      lexer.TokenType
 	bindingPower   int
 	nullDenotation prefixFunc
+}
+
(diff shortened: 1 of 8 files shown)


runtime/parser2/expression.go
```diff
@@ -131,65 +170,47 @@ func setLeftDenotation(tokenType lexer.TokenType, leftDenotation leftDenotationF
 
 func init() {
 
-	define(infix{
+	define(binary{
+		tokenType:        lexer.TokenLess,
+		leftBindingPower: 60,
```
benjaminkvm, line 175: Just a thought, would be nice to have the `bindingPower`s defined somewhere + perhaps relative to each other.

benjaminkvm: Something we can do in the future though!

turbolent: Yep, that's a good idea! It would be nice to have something like Swift's precedence group declarations (in Go, not in Cadence 😄):
https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#ID550,
i.e. declare relative and then calculate actual numbers at the end

benjaminkvm: Yeah thats exactly what I was thinking, similar to a constraint solver in a way.

PR #77, opened 2020-05-07
Add locations to metrics

turbolent:
As discussed, add a location parameter for each metric function.

Note that the numbers include dependencies.

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -138,6 +138,7 @@ func (r *interpreterRuntime) ExecuteScript(
 	}
 
 	value, err := r.interpret(
+		location,
 		runtimeInterface,
 		runtimeStorage,
 		checker,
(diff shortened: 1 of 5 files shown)


runtime/runtime.go
```diff
@@ -275,51 +279,19 @@ func (r *interpreterRuntime) ExecuteTransaction(
 	}
 
 	_, err = r.interpret(
+		location,
 		runtimeInterface,
 		runtimeStorage,
 		checker,
 		functions,
 		nil,
-		func(inter *interpreter.Interpreter) (interpreter.Value, error) {
```
turbolent, line 283: Just refactored that into its own function, i.e. https://github.com/onflow/cadence/pull/77/files#diff-6681b213d55a5c65c5d90b211107fd52R306 is identical

runtime/runtime.go
```diff
@@ -379,6 +400,17 @@ func (r *interpreterRuntime) parseAndCheckProgram(
 				sema.WithPredeclaredValues(valueDeclarations),
 				sema.WithPredeclaredTypes(typeDeclarations),
 				sema.WithValidTopLevelDeclarationsHandler(validTopLevelDeclarations),
+				sema.WithCheckHandler(func(location ast.Location, check func()) {
```
benjaminkvm, line 403: 👍

PR #82, opened 2020-05-09
New Parser: If and while statements

turbolent:

### DIFF
--- a/runtime/parser2/expression.go
+++ b/runtime/parser2/expression.go
@@ -232,13 +232,13 @@ func init() {
 		tokenType: lexer.TokenIdentifier,
 		nullDenotation: func(_ *parser, token lexer.Token) ast.Expression {
 			switch token.Value {
-			case "true":
+			case keywordTrue:
 				return &ast.BoolExpression{
 					Value: true,
 					Range: token.Range,
 				}
 
-			case "false":
+			case keywordFalse:
 				return &ast.BoolExpression{
 					Value: false,
 					Range: token.Range,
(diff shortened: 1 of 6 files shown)


runtime/parser2/statement.go
```diff
@@ -84,3 +91,106 @@ func parseReturnStatement(p *parser) *ast.ReturnStatement {
...
+		if thenBlock == nil {
+			panic(fmt.Errorf("expected block for then branch"))
+		}
+
+		var elseBlock *ast.Block
+
+		parseNested := false
+
+		p.skipSpaceAndComments(true)
+		if p.current.IsString(lexer.TokenIdentifier, "else") {
```
benjaminkvm, line 118: Eventually would be nice to use a constant here, `KeywordIf` or something similar!

runtime/parser2/expression.go
```diff
@@ -330,8 +322,6 @@ func defineArrayExpression() {
 }
 
 func defineDictionaryExpression() {
-	leftBindingPowers[lexer.TokenBraceOpen] = 150
```
turbolent, line 333: `{` isn't used as a left denotation, e.g. we don't have something like `foo{bar...`, so this was wrong. 

Left binding powers for `(` and `[` above
(https://github.com/onflow/cadence/pull/82/files#diff-dc3156241c5da3fc88bf5e2220db331aR302 and
https://github.com/onflow/cadence/pull/82/files#diff-dc3156241c5da3fc88bf5e2220db331aR293) are still correct because they are used for function calls (`foo(...)`) and indexing (`foo[...]`), which will be added later

runtime/parser2/statement.go
turbolent: Good point, I'll add constants for all the keywords 👍

turbolent: Added in 46bc216dfda6f3afa4e5ee036d1decb87a2dd10c

PR #84, opened 2020-05-11
New Parser: Some types, more statements, and more expressions  

turbolent:


runtime/parser2/type.go
```diff
@@ -0,0 +1,297 @@
...
+	case prefixType:
+		tokenType := def.tokenType
+		setTypeNullDenotation(
+			tokenType,
+			func(parser *parser, token lexer.Token) ast.Type {
+				right := parseType(parser, def.bindingPower)
+				return def.nullDenotation(right, token.Range)
+			},
+		)
+	case postfixType:
```
benjaminkvm, line 99: Nice work here!

PR #85, opened 2020-05-11
New Parser: Add restricted and dictionary types

turbolent:


runtime/parser2/declaration.go
```diff
@@ -157,32 +164,60 @@ func parseParameterList(p *parser) (parameterList *ast.ParameterList) {
...
 	p.next()
 
 	var endPos ast.Position
 
+	expectParameter := true
+
 	atEnd := false
 	for !atEnd {
 		p.skipSpaceAndComments(true)
 		switch p.current.Type {
```
benjaminkvm, line 184: We will also need to implement `lexer.TokenUnderscore` to handle un-labeled parameters.

benjaminkvm: Handling which will likely live in `parseParameter` I assume?

turbolent: We don't actually need to differentiate between underscore `_` and other identifiers on the lexer level. Identifiers may start with an underscore and may only contain one character, so it already covers the argument label case. See https://github.com/onflow/cadence/blob/505dc9b920954018e75410c4d26c0218bf9e0bc4/runtime/parser2/lexer/state.go#L91-L92

benjaminkvm: Cool wasn't sure if `_` was considered an identifier -- looks like it is 👍

turbolent:
Yeah, this is seems to be so far the worst case in the parser in terms of ambiguity. I'll need to do a pass and document / comment at the end of the MVP stage, probably just copying the BNF like syntax we already have from the ANTLR grammar

PR #86, opened 2020-05-11
Implement invocation expression for new parser

benjaminkvm:

Implement invocation expression, include various test cases.

Forthcoming: labeled parameters, typed arguments.

### DIFF
--- a/runtime/parser2/expression.go
+++ b/runtime/parser2/expression.go
@@ -80,7 +80,7 @@ func defineExpr(def interface{}) {
 
 		rightBindingPower := def.leftBindingPower
 		if def.rightAssociative {
-			rightBindingPower -= 1
+			rightBindingPower--
 		}
 
 		setExprLeftDenotation(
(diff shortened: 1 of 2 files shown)


runtime/parser2/expression.go
```diff
@@ -373,6 +374,42 @@ func parseNumber(token lexer.Token) ast.Expression {
...
+	setExprLeftDenotation(
+		lexer.TokenParenOpen,
+		func(p *parser, token lexer.Token, left ast.Expression) ast.Expression {
+			var arguments []*ast.Argument
+			var endPos ast.Position
+			atEnd := false
+			for !atEnd {
+				p.skipSpaceAndComments(true)
+				switch p.current.Type {
+				case lexer.TokenComma:
```
turbolent, line 387: I realized this allows multiple commas and multiple arguments to follow each other, e.g. `foo, ,, bar` or `foo bar`, so this needs a state flag to check if a comma or an argument is expected.

I fixed this for parameter lists here:
https://github.com/onflow/cadence/pull/85/files#diff-74be485a73d45cdb0cc03c3aa0ecfe09R179-R220

benjaminkvm: Good catch! Will implement!

PR #89, opened 2020-05-12
Add force unwrap expression

benjaminkvm:

Handle force unwrap expression, minding precedence. Includes test cases.

If you can think of other potentially relevant test cases let me know!


benjaminkvm:
Upped the precedence and added a test case, good catch!

PR #90, opened 2020-05-12
Support member exprs in new parser

benjaminkvm:

Handle member expressions, minding precedence. Includes test cases.

If you can think of other potentially relevant test cases let me know! Would be good to test with force unwrap when we have that one merged.

### DIFF
--- a/runtime/parser2/expression.go
+++ b/runtime/parser2/expression.go
@@ -363,6 +363,7 @@ func init() {
 	definePathExpression()
 	defineConditionalExpression()
 	defineReferenceExpression()
+	defineMemberExpression()
 
 	setExprNullDenotation(lexer.TokenEOF, func(parser *parser, token lexer.Token) ast.Expression {
 		panic("expected expression")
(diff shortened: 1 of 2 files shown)


runtime/parser2/expression.go
```diff
@@ -561,6 +562,21 @@ func defineReferenceExpression() {
 	)
 }
 
+func defineMemberExpression() {
+	setExprLeftBindingPower(lexer.TokenDot, 140)
```
turbolent, line 566: Maybe bump this to the same precedence as indexing and function calls:

```suggestion
	setExprLeftBindingPower(lexer.TokenDot, 150)
```

runtime/parser2/expression.go
```diff
@@ -561,6 +562,21 @@ func defineReferenceExpression() {
...
+	setExprLeftDenotation(
+		lexer.TokenDot,
+		func(p *parser, token lexer.Token, left ast.Expression) ast.Expression {
+			p.skipSpaceAndComments(true)
+			identifierToken := p.mustOne(lexer.TokenIdentifier)
+			return &ast.MemberExpression{
+				Expression: left,
+				Identifier: tokenToIdentifier(identifierToken),
+			}
```
turbolent, line 575: The existing `mustIdentifier` helper can be used here:

```suggestion
			identifier := mustIdentifier(p)
			return &ast.MemberExpression{
				Expression: left,
				Identifier: identifier,
			}
```

benjaminkvm:
Fixed up! Thank for taking a look!

PR #91, opened 2020-05-12
New Parser: Add create expression

turbolent:

### DIFF
--- a/runtime/parser2/expression.go
+++ b/runtime/parser2/expression.go
@@ -295,7 +295,7 @@ func init() {
 
 	defineExpr(literalExpr{
 		tokenType: lexer.TokenIdentifier,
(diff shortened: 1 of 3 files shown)


runtime/parser2/expression.go
```diff
@@ -295,7 +295,7 @@ func init() {
 
 	defineExpr(literalExpr{
 		tokenType: lexer.TokenIdentifier,
-		nullDenotation: func(_ *parser, token lexer.Token) ast.Expression {
+		nullDenotation: func(p *parser, token lexer.Token) ast.Expression {
 			switch token.Value {
 			case keywordTrue:
```
benjaminkvm, line 300: I wonder if it would be better to emit a `TokenKeywordTrue` etc. at the lexer level. Also wonder if there is merit to having them as their own tokens versus under the `Identifier` token. Just thinking out loud!

PR #93, opened 2020-05-13
Run CI Makefile rule instead of just tests

turbolent:
Also, clean up the code with recommended improvements

### DIFF
--- a/runtime/parser2/expression.go
+++ b/runtime/parser2/expression.go
@@ -391,6 +391,7 @@ func defineInvocationExpression() {
 			expectArgument := true
 			for !atEnd {
 				p.skipSpaceAndComments(true)
+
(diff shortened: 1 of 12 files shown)


runtime/parser2/expression.go
```diff
@@ -401,13 +402,15 @@ func defineInvocationExpression() {
 					}
 					p.next()
 					expectArgument = true
+
 				case lexer.TokenParenClose:
 					endPos = p.current.EndPos
 					p.next()
 					atEnd = true
-					break
```
benjaminkvm, line 408: Just curious why this was removed?

turbolent: It's not needed, cases don't fall through in Go

PR #95, opened 2020-05-14
New Parser: Implement index expression

benjaminkvm:

Implement index expression in the new parser, include appropriate tests (whitespace, identifier).

I couldn't find any mention of ranges in the index expression in the Cadence docs nor the existing `IndexExpression` parser test. Let me know if that is something we support!


runtime/parser2/expression.go
```diff
@@ -539,6 +540,26 @@ func defineDictionaryExpression() {
 	)
 }
 
+func defineIndexExpression() {
+	setExprLeftBindingPower(lexer.TokenBracketOpen, 150)
+	setExprLeftDenotation(
+		lexer.TokenBracketOpen,
+		func(p *parser, token lexer.Token, left ast.Expression) ast.Expression {
+			fmt.Println("witihin index expr")
```
turbolent, line 548:
```suggestion
```

runtime/parser2/expression.go
```diff
@@ -539,6 +540,26 @@ func defineDictionaryExpression() {
...
+			firstIndexExpr := parseExpression(p, lowestBindingPower)
+			endToken := p.mustOne(lexer.TokenBracketClose)
+			return &ast.IndexExpression{
+				TargetExpression:   left,
+				IndexingExpression: firstIndexExpr,
+				Range: ast.Range{
+					StartPos: left.StartPosition(),
```
turbolent, line 555: I think in the current parser we use the open bracket token's position. 
It's a bit odd, because the target expression occurs before, but we need a way to refer to just the indexing part (`[...]`) in the error messages.

```suggestion
					StartPos: token.StartPos,
```

benjaminkvm: Yeah, was going back and forth on this! Will change / update the tests!

turbolent:
@benjaminkvm Do you have an estimate for the remaining work here?

benjaminkvm:
Cycling back to this -- will push PR changes stat!

runtime/parser2/expression.go
```diff
@@ -697,6 +698,25 @@ func defineDictionaryExpression() {
 	)
 }
 
+func defineIndexExpression() {
+	setExprLeftBindingPower(lexer.TokenBracketOpen, 150)
```
turbolent, line 702: This must be bumped to 160. It's actually a duplicate of the earlier definition here, which must be removed: https://github.com/onflow/cadence/blob/871304251b61fb152baac272bd1a084d6a0cd245/runtime/parser2/expression.go#L686

```suggestion
	setExprLeftBindingPower(lexer.TokenBracketOpen, 160)
```

turbolent:
Merging for now, I'll refactor the binding powers in a separate PR

PR #96, opened 2020-05-15
Defer resource dictionary values

turbolent:
Closes dapperlabs/flow-go#3682

When all values of a dictionary are resources, store the values under separate storage keys. When the dictionary is accessed, lazily load the values from storage.

### DIFF
--- a/runtime/interpreter/decode.go
+++ b/runtime/interpreter/decode.go
@@ -267,9 +267,13 @@ func (d *Decoder) decodeDictionary(v interface{}, path []string) (*DictionaryVal
 	var deferred map[string]string
 	var deferredOwner *common.Address
 
+	// Are the values in the dictionary deferred, i.e. are they encoded
+	// separately and stored in separate storage keys?
+
 	isDeferred := countMismatch && entryCount == 0
 
 	if isDeferred {
+
 		deferred = make(map[string]string, keyCount)
 		entries = map[string]Value{}
 		deferredOwner = d.owner
--- a/runtime/interpreter/decode.go
+++ b/runtime/interpreter/decode.go
@@ -18,21 +18,22 @@ import (
 // A Decoder decodes CBOR-encoded representations of values.
 //
 type Decoder struct {
-	dec *cbor.Decoder
+	decoder *cbor.Decoder
+	owner   *common.Address
 }
 
 // Decode returns a value decoded from its CBOR-encoded representation,
 // for the given owner (can be `nil`).
 //
 func DecodeValue(b []byte, owner *common.Address, path []string) (Value, error) {
-	r := bytes.NewReader(b)
+	reader := bytes.NewReader(b)
 
-	dec, err := NewDecoder(r)
+	decoder, err := NewDecoder(reader, owner)
 	if err != nil {
 		return nil, err
 	}
 
-	v, err := dec.Decode(owner, path)
+	v, err := decoder.Decode(path)
 	if err != nil {
 		return nil, err
 	}
@@ -43,30 +44,33 @@ func DecodeValue(b []byte, owner *common.Address, path []string) (Value, error) 
 // NewDecoder initializes a Decoder that will decode CBOR-encoded bytes from the
 // given io.Reader.
 //
-func NewDecoder(r io.Reader) (*Decoder, error) {
+// It sets the given address as the owner (can be `nil`).
+//
+func NewDecoder(r io.Reader, owner *common.Address) (*Decoder, error) {
 	decMode, err := cbor.DecOptions{}.DecModeWithTags(cborTagSet)
 	if err != nil {
 		return nil, err
 	}
 
-	return &Decoder{decMode.NewDecoder(r)}, nil
+	return &Decoder{
+		decoder: decMode.NewDecoder(r),
+		owner:   owner,
+	}, nil
 }
 
 // Decode reads CBOR-encoded bytes from the io.Reader and decodes them to a value.
 //
-// It sets the given address as the owner (can be `nil`).
-//
-func (d *Decoder) Decode(owner *common.Address, path []string) (Value, error) {
+func (d *Decoder) Decode(path []string) (Value, error) {
 	var v interface{}
-	err := d.dec.Decode(&v)
+	err := d.decoder.Decode(&v)
 	if err != nil {
 		return nil, err
 	}
 
-	return d.decodeValue(v, owner, path)
+	return d.decodeValue(v, path)
 }
 
-func (d *Decoder) decodeValue(v interface{}, owner *common.Address, path []string) (Value, error) {
+func (d *Decoder) decodeValue(v interface{}, path []string) (Value, error) {
 	switch v := v.(type) {
 
 	// CBOR Types
@@ -81,7 +85,7 @@ func (d *Decoder) decodeValue(v interface{}, owner *common.Address, path []strin
 		return NilValue{}, nil
 
 	case []interface{}:
-		return d.decodeArray(v, owner, path)
+		return d.decodeArray(v, path)
 
 	case cbor.Tag:
 		switch v.Number {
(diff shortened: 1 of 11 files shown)


runtime/interpreter/decode.go
```diff
@@ -17,21 +18,22 @@ import (
 // A Decoder decodes CBOR-encoded representations of values.
 //
 type Decoder struct {
-	dec *cbor.Decoder
+	decoder *cbor.Decoder
+	owner   *common.Address
 }
 
 // Decode returns a value decoded from its CBOR-encoded representation,
 // for the given owner (can be `nil`).
```
benjaminkvm, line 26: It would be good to update this comment to describe the `path` parameter.

runtime/interpreter/decode.go
```diff
@@ -211,12 +217,12 @@ func (d *Decoder) decodeArray(v []interface{}, owner *common.Address) (*ArrayVal
 
 	return &ArrayValue{
 		Values:   values,
-		Owner:    owner,
+		Owner:    d.owner,
```
benjaminkvm, line 220: 👍

runtime/interpreter/encode.go
```diff
@@ -543,11 +581,28 @@ func (e *Encoder) prepareString(v *StringValue) string {
...
 }
 
-func (e *Encoder) prepareArray(v *ArrayValue) ([]interface{}, error) {
+// joinPath returns the path for a nested item, for example the index of an array,
+// the key of a dictionary, or the field name of a composite.
+//
+// \x1F = Information Separator One
+//
+func joinPath(elements []string) string {
+	return strings.Join(elements, "\x1F")
```
benjaminkvm, line 590: Would it be possible for someone to pass in a path that contains this hex sequence?

runtime/runtime_storage.go
```diff
@@ -211,41 +212,122 @@ func (s *interpreterRuntimeStorage) writeValue(
...
 		}
 
+		items = append(items, writeItem{
+			storageKey: fullKey,
+			value:      entry.value,
+		})
+	}
+
+	// Don't use a for-range loop, as keys are added while iterating
+	for len(items) > 0 {
```
benjaminkvm, line 235: I wonder about pulling this out into a separate function, might make it easier to unit test down the road -- `determineDefferals` or something perhaps.

runtime/runtime_storage.go
```diff
@@ -211,41 +212,122 @@ func (s *interpreterRuntimeStorage) writeValue(
...
+		},
+	)
+	return
+}
+
+func (s *interpreterRuntimeStorage) move(
+	oldOwner common.Address, oldKey string,
+	newOwner common.Address, newKey string,
+) {
+	// TODO: fix controller
```
benjaminkvm, line 316: Unsure what this comment means, which controller is this referring to?

runtime/interpreter/encode.go
```diff
@@ -562,22 +617,85 @@ type encodedDictionaryValue struct {
...
+				// If the value is deferred, and the deferred value
+				// is stored in another account's storage,
+				// it must be moved.
+
+				deferredOwner := *v.DeferredOwner
+				owner := *v.Owner
+
+				if deferredOwner != owner {
+
+					deferrals.Moves = append(deferrals.Moves,
```
benjaminkvm, line 681: Would be nice to see a couple unit tests around the deferred moves if possible! Specifically around `DeferredOwner` / `NewOwner`.

runtime/interpreter/value.go
```diff
@@ -4874,12 +4920,18 @@ func (v *DictionaryValue) Count() int {
...
 // TODO: unset owner?
-func (v *DictionaryValue) Remove(keyValue Value) (existingValue Value) {
+func (v *DictionaryValue) Remove(inter *Interpreter, keyValue Value) (existingValue Value) {
 	v.modified = true
 
 	key := dictionaryKey(keyValue)
 	existingValue, exists := v.Entries[key]
 
+	if v.prevDeferredKeys != nil {
+		if storageKey, ok := v.prevDeferredKeys[key]; ok {
```
psiemens, line 4930: So if a resource is moved out of the dictionary, we delete its old key in storage, and then rely on resource semantics to make sure it is stored or destroyed later?

runtime/runtime_storage.go
```diff
@@ -211,41 +212,122 @@ func (s *interpreterRuntimeStorage) writeValue(
...
+
+			for _, deferralMove := range deferrals.Moves {
+
+				s.move(
+					deferralMove.DeferredOwner,
+					deferralMove.DeferredStorageKey,
+					deferralMove.NewOwner,
+					deferralMove.NewStorageKey,
+				)
+			}
```
psiemens, line 273: So are these loops just to allow you to skip over deferred values that have not changed? Otherwise all deferred writes are just added to `items` and written, right?

runtime/runtime_storage.go
```diff
@@ -211,41 +212,122 @@ func (s *interpreterRuntimeStorage) writeValue(
...
+
+				items = append(items, writeItem{
+					storageKey: deferredStorageKey,
+					value:      deferredValue,
+				})
+			}
+
+			for _, deferralMove := range deferrals.Moves {
+
+				s.move(
```
psiemens, line 267: 👍

runtime/interpreter/decode.go
turbolent: Documented it for both for the encoding and the decoding function in ec91981bb46499cc7c3ffafeef5596a7837295d2

runtime/interpreter/encode.go
turbolent: Currently it is not possible to, because we only allow `[a-zA-Z0-9_]` in identifiers. But good point, if we would add support for converting arbitrary strings to paths / path identifiers, we will need to reject this character.

runtime/runtime_storage.go
turbolent: This has been unfortunately cargo-culted for a while:

In the early design of the storage architecture there was the idea that fir each piece of stored data there is a "controller" and an "owner": 
The owner would "own" the data in their account and pay storage rent for it. The controller would have read/write permissions to the data.
For example with a non-fungible token, the user would store it, i.e. be the owner, but the controller would be able to modify it, e.g. the account which has the contract for the NFT, e.g. Dapper Labs has the CK contract, and kitties are stored in user accounts, but only the CK contract can change them.

We started with this notion in the storage API (`runtime.Interface`'s `GetValue` and `SetValue` functions have a `controller` and `owner` parameter, but we only use one at the moment.

After some discussion we realized that we're actually enforcing the ownership and control through the statically through the type system, so there is really no need to the separation in the storage API. We should remove/replace it with a single key at some point soon.

runtime/runtime_storage.go
turbolent: I like splitting this up, but the `determineDeferrals` also then encodes values? Maybe `encodeAndDetermineDeferrals` is a better name?

runtime/interpreter/decode.go
```diff
@@ -237,28 +244,63 @@ func (d *Decoder) decodeDictionary(v interface{}, owner *common.Address) (*Dicti
...
+	}
 
-		decodedValue, err := d.decodeValue(value, owner)
-		if err != nil {
-			return nil, fmt.Errorf("invalid dictionary value encoding: %w", err)
+	var entries map[string]Value
+	var deferred map[string]string
+	var deferredOwner *common.Address
+
+	isDeferred := countMismatch && entryCount == 0
```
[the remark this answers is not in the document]
turbolent: Yeah, with "deferred" I mean the deferral of the encoding and storing when writing and deferral of reading the encoded data from separate storage keys. Is there maybe a better word to describe this concept?

I'll add a comment

runtime/interpreter/value.go
turbolent: Correct, well put. I'll add that as a comment :-)

turbolent: Improved documentation a bit in f594c0ff341ce234fd31c4e06012596ec2984c6c

runtime/runtime_storage.go
turbolent: Yeah, the first loop above is the interaction of optimizing writeback by only writing modified values and the new deferral feature: If a value is deferred and was loaded from storage into memory, then we should only write it back if it was modified.

The moves have a different reason: If a dictionary with deferred values is moved from account to another, the deferred values / storage keys have to be moved over as well

PR #97, opened 2020-05-15
Implement functionality to query block info from runtime

philipstanislaus:


runtime/interface.go
```diff
@@ -60,6 +60,10 @@ type Interface interface {
 	GetComputationLimit() uint64
 	// DecodeArgument decodes a transaction argument against the given type.
 	DecodeArgument(b []byte, t cadence.Type) (cadence.Value, error)
+	// GetCurrentBlockHeight returns the current block height.
+	GetCurrentBlockHeight() uint64
+	// GetBlockAtHeight returns the block at the given height.
+	GetBlockAtHeight(height uint64) (hash [32]byte, timestamp int64, exists bool)
```
turbolent, line 66: Maybe we can define a `cadence.Block` type for this instead of returning the parts as individual return values

runtime/runtime.go
```diff
@@ -1131,26 +1129,17 @@ func (r *interpreterRuntime) newLogFunction(runtimeInterface Interface) interpre
...
+	if !ok {
+		return nil
 	}
 
-	encoded := buf.Bytes()
-	var hash [stdlib.BlockIDSize]byte
-	copy(hash[stdlib.BlockIDSize-len(encoded):], encoded)
-
-	block := NewBlockValue(height, hash, time.Unix(int64(height), 0))
+	block := NewBlockValue(height, hash, time.Unix(0, timestamp))
```
turbolent, line 1142: `time.Unix` takes `sec int64, nsec int64`, is the argument order accidentally swapped?

philipstanislaus: Yes, that is intentional. We are using `time.UnixNano()` on the flow side, which returns the time in nanoseconds (not just the remainder). Hence, we need to use 0 as the seconds value.

PR #101, opened 2020-05-15
New Parser: Add optional member access

turbolent:

### DIFF
--- a/runtime/parser2/lexer/state.go
+++ b/runtime/parser2/lexer/state.go
@@ -110,9 +110,14 @@ func rootState(l *lexer) stateFn {
 				l.emitType(TokenSlash)
 			}
 		case '?':
-			if l.acceptOne('?') {
+			r = l.next()
+			switch r {
+			case '?':
 				l.emitType(TokenDoubleQuestionMark)
-			} else {
+			case '.':
+				l.emitType(TokenQuestionMarkDot)
+			default:
+				l.backupOne()
 				l.emitType(TokenQuestionMark)
 			}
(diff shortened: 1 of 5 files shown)


runtime/parser2/lexer/state.go
```diff
@@ -110,9 +110,14 @@ func rootState(l *lexer) stateFn {
 				l.emitType(TokenSlash)
 			}
 		case '?':
-			if l.acceptOne('?') {
+			r = l.next()
+			switch r {
+			case '?':
 				l.emitType(TokenDoubleQuestionMark)
-			} else {
+			case '.':
```
benjaminkvm, line 117: This means we do not permit whitespace between, correct?

turbolent: Yeah, while implementing it I realized that this is an ambiguous case with the conditional expression. 

I checked Swift, and it actually handles this even more restrictive: 
Only if there is no whitespace before, between, and after the `?` and `.` it is considered optional member access, otherwise it is ternary. 

I think in Swift this is so restrictive because it allows implicit member access:

```swift
enum Color {
  case green 
  case blue
}
```

can be used like so:

`let color: Color = something ? .green : .blue`

I think it's fair to require no whitespace between and after, but it's nice to allow it before. This allows us to resolve the ambiguous case easily.
Otherwise we have to check for a colon after the member access

benjaminkvm: Ah makes sense -- was going to say wouldn't think `.ident` would be valid.

PR #102, opened 2020-05-16
Recover external panics

turbolent:
The interpreter recovers all panics except for Go's own errors and returns them as error values.

However, the interpreter is also making calls to "external" components, in particular the `runtime.Interface` functions.

Recover panics from those external calls and re-panic them, so they can be handled by what is calling into the interpreter.


runtime/runtime_test.go
```diff
@@ -3729,3 +3724,35 @@ func TestRuntimeStorageWriteback(t *testing.T) {
...
+
+	assert.PanicsWithValue(t,
+		interpreter.ExternalError{
+			Recovered: logPanic{},
+		},
+		func() {
+			_ = runtime.ExecuteTransaction(script, nil, runtimeInterface, utils.TestLocation)
+		},
+	)
+}
```
philipstanislaus, line 3758: Nice test! 🎉

PR #104, opened 2020-05-16
New Parser: Add assignment statement

turbolent:

- Allow assignment to any expression in the current parser, but reject it in the checker
- Add assignment statement to new parser


runtime/parser/Cadence.g4
```diff
@@ -349,7 +349,7 @@ variableDeclaration
 //   Being unrestritive here allows us to provide better error messages
 //   in the semantic analysis.
 assignment
-    : identifier expressionAccess* transfer expression
+    : target=expression transfer value=expression
```
benjaminkvm, line 352: Was parser performance improvement the intent behind making this more permissive? I suspect this would improve parse times.

turbolent: No, this is actually an UX improvement, as we allow more (potentially wrong) code to be parsed, and then check and report a nice error message when it is invalid, instead of rejecting in the parser and giving a cryptic syntax-based error message.

Also, it simplifies the parser: Instead of having to manually construct the target expression from the identifier and expression access rules, we can just parse and expression and that's it.

PR #111, opened 2020-05-18
New Parser: Number literals

turbolent:


runtime/ast/expression.go
```diff
@@ -173,7 +173,12 @@ func (e *FixedPointExpression) String() string {
 	}
 	builder.WriteString(e.UnsignedInteger.String())
 	builder.WriteRune('.')
-	builder.WriteString(e.Fractional.String())
+	fractional := e.Fractional.String()
+	var i uint = 0
+	for ; i < (e.Scale - uint(len(fractional))); i++ {
```
benjaminkvm, line 178: I believe you could also do:

`for i := uint(0); i < (e.Scale - uint(len(fractional))); i++ {`

as it appears you do not use the `i` outside the scope of the loop.

runtime/parser2/lexer/tokentype.go
```diff
@@ -80,90 +84,98 @@ func (t TokenType) String() string {
...
+		return "hexadecimal integer"
+	case TokenFixedPointLiteral:
+		return "fixed-point number"
 	case TokenIdentifier:
 		return "identifier"
 	case TokenString:
 		return "string"
 	case TokenPlus:
-		return "+"
+		return `'+'`
```
benjaminkvm, line 102: Just curious why you quoted these characters in this way?

runtime/parser2/expression.go
```diff
@@ -876,3 +951,63 @@ func parseHex(r rune) rune {
...
+}
+
+func parseFixedPointLiteral(text string, tokenRange ast.Range) *ast.FixedPointExpression {
+	parts := strings.Split(text, ".")
+	integer, _ := parseFixedPointPart(parts[0])
+	fractional, scale := parseFixedPointPart(parts[1])
+
+	return &ast.FixedPointExpression{
+		Negative:        false,
+		UnsignedInteger: integer,
```
benjaminkvm, line 1008: Just curious why we manage the signed-ness ourselves versus using a `Integer` field in place of `UnsignedInteger` and `Negative` that accepts signed ints.

runtime/ast/expression.go
turbolent: Good point! Simplified in 478ad8c52e89818bb7229ef2f10ce0bcc202856d

runtime/parser2/lexer/tokentype.go
turbolent: The result of this function here is mostly a representation of the token, so I quoted all the ones that are literals (e.g `+`, `"`, etc.), but left the "set" / descriptive ones (e.g. space, comment), so they are properly rendered in error messages, e.g. "expected '+'" but "expected space"

runtime/parser2/expression.go
turbolent: The code initially did that, i.e. it only had an `Integer` field instead of the two `Negative` and `UnsignedInteger` flags.

However, `big.Int` doesn't support negative zeros, so representing e.g. `-0.1`, which has the integer part `-0` wasn't possible.

PR #112, opened 2020-05-19
Add a `toString` function for addresses, integers, and fixed-point numbers

turbolent:
Until we can properly model inheritance we have to duplicate the definitions

### DIFF
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -248,6 +248,28 @@ var toStringFunctionType = &FunctionType{
 	),
 }
 
+var typeBuiltinFunctionTypes = map[TypeID]map[string]*FunctionType{}
+
+func init() {
+	for _, ty := range append(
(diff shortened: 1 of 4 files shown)


runtime/sema/type.go
```diff
@@ -238,6 +238,16 @@ func NewTypeAnnotation(ty Type) *TypeAnnotation {
 	}
 }
 
+// toString
+
+const ToStringFunctionName = "toString"
```
benjaminkvm, line 243: I anticipate in the future we will have additional "built-ins" for all types, at which point it might make sense to have those stored in a `typeBuiltIns map[string][FunctionType]` or something of the like -- for now though this looks good!

turbolent: That's a good idea! I'll refactor this 👍

turbolent: Generalized this in 2c21e53

PR #123, opened 2020-05-21
Don't perform a cache lookup when updating the account code

turbolent:
@rrrkren Noticed that account creation with code works, but creating an account without code and then updating the code does not. Thanks again for the great repro! 👏 

Problem is that the account code update function checked the cache – which contains an entry for the initially empty program when the account was created.

Add a flag to prevent a cache lookup and don't perform the lookup when updating account code.

### DIFF
--- a/runtime/runtime_test.go
+++ b/runtime/runtime_test.go
@@ -267,7 +267,9 @@ func TestRuntimeImport(t *testing.T) {
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


runtime/runtime_test.go
```diff
@@ -380,6 +382,14 @@ func TestRuntimeProgramCache(t *testing.T) {
 	})
 }
 
+func newTransactionLocationGenerator() func() TransactionLocation {
```
psiemens, line 385: 👍

PR #124, opened 2020-05-21
New Parser: Add support for invocation expression argument labels

benjaminkvm:
Closes dapperlabs/flow-go#3764

## Description

Add support for invocation expression argument labels.

Parse the (potential) label AST expression, if a colon follows, attempt to cast the expression to an `IdentifierExpression` and use as the label, continue parsing.

Add additional tests to cover labeled arguments.

### DIFF
--- a/runtime/parser2/expression.go
+++ b/runtime/parser2/expression.go
@@ -629,11 +629,9 @@ func parseArgumentListRemainder(p *parser) (arguments []*ast.Argument, endPos as
 	return
 }
 
-// Invocation Expression Argument Grammar
+// parseArgument parses an argument in an invocation.
 //
-// argument
-// : (identifier ':')? expression
-// ;
+//     argument : (identifier ':' )? expression
 //
 func parseArgument(p *parser) *ast.Argument {
 	var label string
--- a/runtime/parser2/expression.go
+++ b/runtime/parser2/expression.go
@@ -574,10 +574,7 @@ func parseCreateExpressionRemainder(p *parser, token lexer.Token) *ast.CreateExp
 
 // Invocation Expression Grammar:
 //
-// invocation
-// : ('<' ( typeAnnotation (',' typeAnnotation )* )? '>')?
-// '(' ( argument (',' argument)* )? ')'
-// ;
+//     invocation : '(' ( argument ( ',' argument )* )? ')'
 //
 func defineInvocationExpression() {
 	setExprLeftDenotation(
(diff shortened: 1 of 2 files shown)


runtime/parser2/expression.go
```diff
@@ -572,6 +572,13 @@ func parseCreateExpressionRemainder(p *parser, token lexer.Token) *ast.CreateExp
 	}
 }
 
+// Invocation Expression Grammar:
+//
+// invocation
+// : ('<' ( typeAnnotation (',' typeAnnotation )* )? '>')?
+// '(' ( argument (',' argument)* )? ')'
+// ;
+//
```
turbolent, line 581: Good idea to add the grammar! Given that this isn't yet able to parse type arguments, maybe leave them out to avoid confusion:

```suggestion
// Invocation Expression Grammar:
//
//     invocation : '(' ( argument ( ',' argument )* )? ')'
//
```

runtime/parser2/expression.go
```diff
@@ -618,17 +625,56 @@ func parseArgumentListRemainder(p *parser) (arguments []*ast.Argument, endPos as
...
 	}
 	return
 }
 
+// Invocation Expression Argument Grammar
+//
+// argument
+// : (identifier ':')? expression
+// ;
+//
```
turbolent, line 640:
```suggestion
// parseArgument parses an argument in an invocation.
//
//     argument : (identifier ':' )? expression
//
```

runtime/parser2/expression.go
```diff
@@ -618,17 +625,56 @@ func parseArgumentListRemainder(p *parser) (arguments []*ast.Argument, endPos as
...
+func parseArgument(p *parser) *ast.Argument {
+	var label string
+	var labelStartPos, labelEndPos ast.Position
+
+	expr := parseExpression(p, lowestBindingPower)
+	p.skipSpaceAndComments(true)
+
+	// If a colon follows the expression, the expression was our label.
```
turbolent, line 648: 👍

runtime/parser2/expression_test.go
```diff
@@ -1049,7 +1049,7 @@ func TestInvocation(t *testing.T) {
 		result, errs := ParseExpression("f(1)")
 		require.Empty(t, errs)
 
-		assert.Equal(t,
+		utils.AssertEqualWithDiff(t,
```
turbolent, line 1052: Great idea! We should use that everywhere in `runtime/parser2` 👍

PR #129, opened 2020-05-22
Fix resource dictionary value deferral removal

### LINKED ISSUE #128: BUG: Removing objects from a dictionary does not load value correctly.
### Problem

Objects are not being loaded correctly from a dictionary after it has been stored.

### Steps to Reproduce

1. Pull the josh/bugFix branch from the NFT repo:
https://github.com/onflow/flow-nft/tree/josh/bugFix
2. Run `go test -v` in the `nfttests` directory.
3. See that there is an error that says `missing NFT`
4. We use the simple contract `SimpleNFT.cdc`. At deployment time, the contract withdraws and deposits token index 1 from the dictionary to make sure that is working. Then in the `GenerateDestroyScript` in `nft_templates.go`, we load the collection from storage and try to withdraw with causes the `missing NFT` error. We had also gotten the error from borrowing a reference to the collection.

### Acceptance Criteria

This test case must pass.

### Context

This makes it impossible to withdraw an NFT from a collection, so it blocks testing of the topshot contracts on the testnet

turbolent:
Fixes #128

When removing a key from the a dictionary, the value might be deferred, so properly load and return it.

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -5221,7 +5221,7 @@ func dictionaryKey(keyValue Value) string {
 	return hasKeyString.KeyString()
 }
 
-func (v *DictionaryValue) Set(inter *Interpreter, _ LocationRange, keyValue Value, value Value) {
+func (v *DictionaryValue) Set(inter *Interpreter, locationRange LocationRange, keyValue Value, value Value) {
 	v.modified = true
 
 	switch typedValue := value.(type) {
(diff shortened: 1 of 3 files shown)


runtime/interpreter/value.go
```diff
@@ -5334,11 +5323,12 @@ func (v *DictionaryValue) Count() int {
 }
 
 // TODO: unset owner?
-func (v *DictionaryValue) Remove(inter *Interpreter, keyValue Value) (existingValue Value) {
+func (v *DictionaryValue) Remove(inter *Interpreter, locationRange LocationRange, keyValue Value) OptionalValue {
 	v.modified = true
 
+	value := v.Get(inter, locationRange, keyValue)
```
turbolent, line 5329:
```suggestion
    // Don't use `Entries` here: the value might be deferred and needs to be loaded
	value := v.Get(inter, locationRange, keyValue)
```

psiemens, line 5330: 👍

PR #130, opened 2020-05-22
Fix resource dictionary destruction and insertion with deferred values

turbolent:
Related to #128 

When a resource dictionary is destroyed or a key is updated, the value might be deferred. 

Load the deferred values properly instead of only considering loaded values.

Also, simplify/reuse the existing test for ensuring deferred values are properly loaded when removing keys from resource dictionaries.

The diff looks odd, might be best reviewed by looking at the new code in split-diff mode

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -5160,10 +5160,9 @@ func (v *DictionaryValue) Destroy(interpreter *Interpreter, locationRange Locati
 	}
 
 	for _, keyValue := range v.Keys.Values {
+		// Don't use `Entries` here: the value might be deferred and needs to be loaded
+		value := v.Get(interpreter, locationRange, keyValue)
 		maybeDestroy(keyValue)
-	}
-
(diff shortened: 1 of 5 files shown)


runtime/interpreter/value.go
```diff
@@ -5293,18 +5298,14 @@ func (v *DictionaryValue) GetMember(_ *Interpreter, _ LocationRange, name string
...
 
-				existingValue := v.Insert(keyValue, newValue)
+				existingValue := v.Insert(
+					invocation.Interpreter,
+					invocation.LocationRange,
+					keyValue,
+					newValue,
+				)
 
-				var returnValue Value
```
Kay-Zee, line 5298: this part is no longer needed? just making sure we didn't remove it by accident.

turbolent: This was boxing the result from `Insert`, which used to return a `Value` or `nil`, but is now already returning an `OptionalValue`, so it doesn't need to be boxed up here anymore

PR #132, opened 2020-05-23
Fix move of resource dictionary value

turbolent:
We already handled the move/transfer of dictionaries with deferred values by moving the the deferred storage keys.

However, we also have to handle the move/transfer of a dictionary value from one dictionary to another: If the value is not modified, the writeback optimization will ignore the new value in the receiving dictionary, so mark the moved/transferred value as modified.

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -27,7 +27,8 @@ type Value interface {
 	Copy() Value
 	GetOwner() *common.Address
 	SetOwner(address *common.Address)
-	Modified() bool
+	IsModified() bool
+	SetModified(modified bool)
 }
 
 // ValueIndexableValue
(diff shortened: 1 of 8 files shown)


runtime/interpreter/value.go
```diff
@@ -5376,6 +5473,13 @@ func (v *DictionaryValue) Insert(inter *Interpreter, locationRange LocationRange
 
 	value.SetOwner(v.Owner)
 
+	// Mark the inserted value itself modified.
+	// It might have been stored as a deferred value and loaded,
+	// so must be written (potentially as a deferred value again),
+	// and would otherwise be ignored by the writeback optimization.
+
+	value.SetModified(true)
```
Kay-Zee, line 5481: noice

PR #133, opened 2020-05-23
New Parser: Composite declarations

turbolent:

Also:
- Fix the new parser:
  - Fix casting expression: set the parent variable declaration
- Fix the old parser:
  - Fix function type parameter list end position
  - Don't allocated slice for conformances

The changes in `parser_test.go` are all just indentation changes, plus the end position fixes mentioned above, the diff looks quite weird, maybe best viewed by disabling whitespace in the GitHub diff

### DIFF
--- a/runtime/parser2/type.go
+++ b/runtime/parser2/type.go
@@ -464,8 +464,6 @@ func parseNominalTypes(
 	for !atEnd {
 		p.skipSpaceAndComments(true)
 
-		fmt.Printf(">>> %s\n", p.current.Type)
-
 		switch p.current.Type {
 		case lexer.TokenComma:
 			if expectType {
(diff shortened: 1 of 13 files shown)


runtime/parser2/type.go
```diff
@@ -433,71 +433,82 @@ func defineRestrictedOrDictionaryType() {
...
+		},
+	)
+}
 
-			atEnd := false
-			for !atEnd {
-				p.skipSpaceAndComments(true)
+// parseNominalTypes parses zero or more nominal types separated by comma.
+//
+func parseNominalTypes(
```
benjaminkvm, line 455: Might be able to use this for type arguments 👍

PR #136, opened 2020-05-25
CBOR: Up MaxArrayElements, MaxNestedLevels decoding options

benjaminkvm:
## Description

Increase CBOR `DecOptions.MaxArrayElements`, `DecOptions.MaxNestedLevels` decoder options (previously set at defaults 131072 and 32 respectively).

## Bug

While running a large test the following error was encountered while attempting to decode a large array from storage:

```
status: SEALED err: code execution failed: Execution failed:
cbor: exceeded max number of elements 131072 for CBOR array
```

Testcase forthcoming, just wanted to put this out there in case it was urgent.

### DIFF
--- a/runtime/interpreter/decode.go
+++ b/runtime/interpreter/decode.go
@@ -51,7 +51,14 @@ func DecodeValue(b []byte, owner *common.Address, path []string) (Value, error) 
 // It sets the given address as the owner (can be `nil`).
 //
 func NewDecoder(r io.Reader, owner *common.Address) (*Decoder, error) {
-	decMode, err := cbor.DecOptions{}.DecModeWithTags(cborTagSet)
+	decMode, err := cbor.DecOptions{
+		// Maximum nested levels permitted.
+		MaxNestedLevels: 256,
+		// Maximum array elements permitted.
+		MaxArrayElements: 134217728,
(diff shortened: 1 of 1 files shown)


runtime/interpreter/decode.go
```diff
@@ -51,7 +51,14 @@ func DecodeValue(b []byte, owner *common.Address, path []string) (Value, error)
 // It sets the given address as the owner (can be `nil`).
 //
 func NewDecoder(r io.Reader, owner *common.Address) (*Decoder, error) {
-	decMode, err := cbor.DecOptions{}.DecModeWithTags(cborTagSet)
+	decMode, err := cbor.DecOptions{
+		// Maximum nested levels permitted.
+		MaxNestedLevels: 256,
+		// Maximum array elements permitted.
+		MaxArrayElements: 134217728,
```
Kay-Zee, line 58: reasonable to say they'll never hit this limit i think :D

Kay-Zee:
As discussed, would it be okay to only double the previous limits? @benjaminkvm @turbolent

turbolent:
@Kay-Zee Yep! I'll bump, but will do it like @psiemens suggested, as options

turbolent:
Replaced by #370

PR #137, opened 2020-05-25
Add runtime interface method to receive stats

benjaminkvm:
## Description

While profiling Alan's TXs to determine a reasonable gas limit, I needed a way to expose the gas consumed by interpreted programs (we currently do a check against a limit upon each function invocation, loop iteration etc).

I thought I would make this a bit more generic to support future runtime/interpreter statistics we may want to expose to the Cadence "host" (application implementing the `runtime.Interface`).

Note: 

I realize this is very similar to our `Metrics` system, but (perhaps wrongly) considered that more of a profiling interface and that perhaps there might be merit to exposing certain runtime stats (like gas consumed) via the `runtime.Interface`.

Let me know what you think! Happy to implement via `Metrics`.

### DIFF
--- a/runtime/interface.go
+++ b/runtime/interface.go
@@ -50,6 +50,8 @@ type Interface interface {
 	UpdateAccountCode(address Address, code []byte, checkPermission bool) (err error)
 	// GetSigningAccounts returns the signing accounts.
 	GetSigningAccounts() []Address
+	// Statisicts receives statistics about the runtime of a program.
+	Statistics(stats *Statistics)
 	// Log logs a string.
 	Log(string)
 	// EmitEvent is called when an event is emitted by the runtime.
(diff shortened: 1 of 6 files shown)


benjaminkvm:
Linter error, hmm..

```
##[error]`interpretion` is a misspelling of `interpreting` (misspell)
  
  ##[error]issues found
```

benjaminkvm:
"Regardless, I don't think gas consumption should be reported in the same place as performance metrics, which are more of an opt-in debugging feature." 💯 

Agree the `runtime.Interface` is getting a bit overwhelming also. Yeah this could be another optional interface the Cadence-host could implement.

benjaminkvm:
Curious for your thoughts on this @turbolent -- think this might be a good time to start breaking out a few methods from the `runtime.Interface`?

turbolent:
Yeah, I definitely agree with splitting the interface up and making some optional.

benjaminkvm:
Going to close this out, active discussions around refactoring out the `runtime.Interface`! This was useful for profiling Cadence gas consumption in either case!

PR #139, opened 2020-05-26
New Parser: Properly stop lexer gorountines

turbolent:


runtime/parser2/lexer/lexer.go
```diff
@@ -158,7 +169,14 @@ func (l *lexer) emit(ty TokenType, val interface{}, rangeStart ast.Position, con
 			},
 		},
 	}
-	l.tokens <- token
+
+	select {
+	case <-l.ctx.Done():
```
benjaminkvm, line 174: 👍

PR #142, opened 2020-05-27
New Parser: Fixes and cleanup

turbolent:

- Extend the command line parsing program:
  - Benchmark both the old and the new parser
  - Add a comparison mode, which diffs the results of the old and new parser
- Fix left binding power for index expression (150->160)
- Make the function block in composite and interface declarations optional. They also must be optional in composite declarations, because the composite declaration might be a type requirement, which doen't use the `interface` keyword, but is parsed like an interface
- Fix the newline handling in the expression parser
- Don't allow newlines before the left denotation of `[` to prevent a potential array literal expression in the next line to be parsed as an indexing expression
- Define constants for the left binding powers of expressions and types

With these changes, all FT and NFT contracts and transactions are parsed by the new parser like they are parsed by the new parser:

```
$ go run ./cmd/parse -compare flow-{n,}ft/{contracts,transactions}/*
../../flow-nft/contracts/ExampleNFT.cdc:	OK
../../flow-nft/contracts/NonFungibleToken.cdc:	OK
../../flow-nft/transactions/mint_nft.cdc:	OK
../../flow-nft/transactions/read_nft_data.cdc:	OK
../../flow-nft/transactions/setup_account.cdc:	OK
../../flow-nft/transactions/transfer_nft.cdc:	OK
../../flow-ft/contracts/CustodialDeposit.cdc:	OK
../../flow-ft/contracts/FlowToken.cdc:	OK
../../flow-ft/contracts/FungibleToken.cdc:	OK
../../flow-ft/contracts/TokenForwarding.cdc:	OK
../../flow-ft/transactions/burn_tokens.cdc:	OK
../../flow-ft/transactions/create_forwarder.cdc:	OK
../../flow-ft/transactions/create_minter.cdc:	OK
../../flow-ft/transactions/custodial_deposit.cdc:	OK
../../flow-ft/transactions/get_balance.cdc:	OK
../../flow-ft/transactions/get_supply.cdc:	OK
../../flow-ft/transactions/mint_tokens.cdc:	OK
../../flow-ft/transactions/setup_account.cdc:	OK
../../flow-ft/transactions/transfer_tokens.cdc:	OK
```


runtime/parser2/expression.go
```diff
@@ -29,6 +29,24 @@ import (
 	"github.com/onflow/cadence/runtime/parser2/lexer"
 )
 
+const (
```
benjaminkvm, line 32: Love this 👏

runtime/parser2/expression.go
```diff
@@ -997,12 +1012,13 @@ func parseMemberAccess(p *parser, left ast.Expression, optional bool) ast.Expres
 
 func exprLeftDenotationAllowsNewline(tokenType lexer.TokenType) bool {
 
-	// The postfix force unwrap and invocation expressions don't support
-	// newlines before them, as this clashes with a unary negations and
-	// nested expressions on a new line / separate statement
+	// The postfix force unwrap, invocation expressions,
```
benjaminkvm, line 1015: Is this also true for the existing parser?

Something I think we should probably have in the Cadence language reference document (if it isn't there already).

turbolent: Yep, this is the same behavior as in the existing parser: 

- https://github.com/onflow/cadence/blob/master/runtime/parser/Cadence.g4#L459
- https://github.com/onflow/cadence/blob/master/runtime/parser/Cadence.g4#L461
- https://github.com/onflow/cadence/blob/master/runtime/parser/Cadence.g4#L553

PR #143, opened 2020-05-27
Fix order of conformance declaration and usage

turbolent:
Declare nested composite and interface declarations before checking members, so conformances are properly declared when the members are checked, which might use the conformances, e.g. in a restricted type.

### DIFF
--- a/runtime/tests/checker/restriction_test.go
+++ b/runtime/tests/checker/restriction_test.go
@@ -1067,16 +1067,38 @@ func TestCheckRestrictedTypeNoType(t *testing.T) {
 
 func TestCheckRestrictedTypeConformanceOrder(t *testing.T) {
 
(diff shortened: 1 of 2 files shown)


runtime/tests/checker/restriction_test.go
```diff
@@ -1064,3 +1064,19 @@ func TestCheckRestrictedTypeNoType(t *testing.T) {
 		)
 	})
 }
+
+func TestCheckRestrictedTypeConformanceOrder(t *testing.T) {
```
benjaminkvm, line 1068: Might be good to have the invalid case also!

turbolent: Good point, added in eb52a589828981cf1148a8465fddd3e28215fa82

PR #144, opened 2020-05-28
New Parser: More fixes and cleanup

turbolent:

- Parse checker tests with both old and new parser and compare
- Don't allocate an empty slice for restrictions
- Make `prepare`s in transactions optional
- Don't require decimal literal for constant sized type, this is checked later in sema
- Handle ambiguity between function declaration and function expression statement
- Handle multiple unary negation
- Accept all declarations which are statements
- Fix parsing of fixed-point literals
- Unary plus is not supported, remove it
- Identifier import locations are not supported yet, don't produce a nil location

With these changes, all checker tests can now be parsed using the new parser.
Also, go-fuzz has not reported any more crashers so far.

### DIFF
--- a/runtime/parser2/statement.go
+++ b/runtime/parser2/statement.go
@@ -79,12 +79,8 @@ func parseStatement(p *parser) ast.Statement {
 	// it might start with a keyword for a declaration
 
 	declaration := parseDeclaration(p)
-	// TODO: allow more
-	switch declaration := declaration.(type) {
-	case *ast.VariableDeclaration:
-		return declaration
-	case *ast.FunctionDeclaration:
-		return declaration
+	if statement, ok := declaration.(ast.Statement); ok {
+		return statement
 	}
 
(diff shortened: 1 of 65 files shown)


runtime/parser2/statement.go
```diff
@@ -122,6 +122,46 @@ func parseStatement(p *parser) ast.Statement {
 	}
 }
 
+func parseFunctionDeclarationOrFunctionExpressionStatement(p *parser) ast.Statement {
+
+	startPos := p.current.StartPos
+
+	// Skip the `fun` keyword
```
benjaminkvm, line 129: Would be great to make this a convention:

Describe the token that we are advancing with a comment above all `p.next()` calls.

turbolent: Good idea, done in 5be849739ff3adfba996aa39490bbc6c766e83c2

PR #147, opened 2020-05-29
Change error handling to not panic when querying block info from runtime

philipstanislaus:


turbolent:
Agreed, it would be nice if we could figure out a standard for reporting and differentiating between fatal implementation errors (e.g fetching info for an existing block fails) and program errors (e.g. fetching info for a non-existing block).

philipstanislaus:
Agree. Do we have similar logic implemented somewhere already?

PR #148, opened 2020-05-29
Add AuthAccount.unsafeNotInitializingSetCode

turbolent:
Add `AuthAccount.unsafeNotInitializingSetCode`, which, like `setCode`, updates the account's code, but does **not** create the contract object.

This allows updating the contract code and keeping the stored contract object.

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -929,9 +929,7 @@ func (r *interpreterRuntime) newSetCodeFunction(
 				constructorArguments,
 				constructorArgumentTypes,
 				invocation.LocationRange.Range,
-				updateAccountCodeOptions{
-					createContract: options.createContract,
-				},
+				updateAccountCodeOptions(options),
 			)
 
 			codeHashValue := CodeToHashValue(code)
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -214,7 +214,22 @@ func (r *interpreterRuntime) newAuthAccountValue(
 ) interpreter.AuthAccountValue {
 	return interpreter.NewAuthAccountValue(
 		addressValue,
-		r.newSetCodeFunction(addressValue, runtimeInterface, runtimeStorage),
+		r.newSetCodeFunction(
+			addressValue,
+			runtimeInterface,
+			runtimeStorage,
+			setCodeOptions{
+				createContract: true,
+			},
+		),
+		r.newSetCodeFunction(
+			addressValue,
+			runtimeInterface,
+			runtimeStorage,
+			setCodeOptions{
+				createContract: false,
+			},
+		),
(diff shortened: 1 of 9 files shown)


benjaminkvm:
Is there an issue related to this? Just curious why we would want to do this / if this is a workaround  for a current limitation / deployment issue.

joshuahannan:
@benjaminkvm We're doing this right now because the NBA wants to upgrade their smart contract to fix a bug and add some extra functions and we need to be able to do it without re-initializing the whole contract

turbolent:
@benjaminkvm Good question, I should have added some context, given that there is no issue for it. 

We've discussed this on Friday and decided that as a short-term solution until we have code upgrades defined properly, we can add such a function which allows at least functionality update (existing data is not migrated)

runtime/runtime.go
```diff
@@ -975,38 +1003,45 @@ func (r *interpreterRuntime) updateAccountCode(
...
+		address := common.Address(addressValue)
 
-	contractValue.SetOwner(&address)
+		contractValue.SetOwner(&address)
+	}
 
 	// NOTE: only update account code if contract instantiation succeeded
 	wrapPanic(func() {
-		err = runtimeInterface.UpdateAccountCode(addressValue.ToAddress(), code, checkPermission)
+		err = runtimeInterface.UpdateAccountCode(addressValue.ToAddress(), code, options.checkPermission)
```
psiemens, line 1036: As you can see here, the `checkPermission` parameter isn't used by the virtual machine anymore. Can we just remove it? 

https://github.com/dapperlabs/flow-go/blob/a323a57b52300345d617564d45a33c80574b973a/engine/execution/computation/virtualmachine/transactionContext.go#L291-L320

runtime/runtime_test.go
```diff
@@ -4106,3 +4106,154 @@ func TestRuntimeUpdateCodeCaching(t *testing.T) {
...
+          assert(Test.rs.length == 1)
+          assert(Test.rs["r1"]?.name == "1")
+
+          // New function code is executed.
+          // Compare with script1 above, which checked 1.
+
+          assert(Test.rs["r1"]?.hello() == 2)
+          assert(Test.hello() == 2)
+      }
+    `)
```
psiemens, line 4255: Can you just reuse `script1` from above?

runtime/runtime.go
turbolent: Good idea, removed in 491e698c8e780e57c52c874cbbd86f3b23a34485

runtime/runtime_test.go
turbolent: No, they differ in what they check, see the last two asserts in each script

PR #151, opened 2020-06-03
Implement script parameters

benjaminkvm:
Work towards https://github.com/dapperlabs/flow-go/issues/3068.

## Description

Provide support for script parameters, with corresponding test cases.

### DIFF
--- a/runtime/errors.go
+++ b/runtime/errors.go
@@ -107,7 +107,7 @@ type InvalidEntryPointParameterCountError struct {
 
 func (e InvalidEntryPointParameterCountError) Error() string {
 	return fmt.Sprintf(
-		"parameter count mismatch for transaction: expected %d, got %d",
+		"entry point parameter count mismatch: expected %d, got %d",
 		e.Expected,
 		e.Actual,
 	)
--- a/runtime/errors.go
+++ b/runtime/errors.go
@@ -82,11 +82,6 @@ func (e InvalidTransactionCountError) Error() string {
 
 type MissingEntryPointError struct {
 	Expected string
-	Err      error
-}
-
-func (e *MissingEntryPointError) Unwrap() error {
-	return e.Err
 }
 
 func (e *MissingEntryPointError) Error() string {
@@ -97,11 +92,6 @@ func (e *MissingEntryPointError) Error() string {
 
 type InvalidEntryPointTypeError struct {
 	Type sema.Type
-	Err  error
-}
-
-func (e *InvalidEntryPointTypeError) Unwrap() error {
-	return e.Err
 }
 
 func (e *InvalidEntryPointTypeError) Error() string {
@@ -110,12 +100,12 @@ func (e *InvalidEntryPointTypeError) Error() string {
 
 // InvalidTransactionParameterCountError
 
-type InvalidTransactionParameterCountError struct {
+type InvalidEntryPointParameterCountError struct {
 	Expected int
 	Actual   int
 }
 
-func (e InvalidTransactionParameterCountError) Error() string {
+func (e InvalidEntryPointParameterCountError) Error() string {
 	return fmt.Sprintf(
 		"parameter count mismatch for transaction: expected %d, got %d",
 		e.Expected,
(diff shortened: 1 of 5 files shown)


runtime/errors.go
```diff
@@ -78,6 +78,36 @@ func (e InvalidTransactionCountError) Error() string {
...
+// MissingEntryPointError
+
+type MissingEntryPointError struct {
+	Expected string
+	Err      error
+}
+
+func (e *MissingEntryPointError) Unwrap() error {
+	return e.Err
+}
```
turbolent, line 90: What is the purpose of the wrapped error here? It looks like it's only ever created without an inner error: https://github.com/onflow/cadence/pull/151/files#diff-6681b213d55a5c65c5d90b211107fd52R136

Maybe just remove the `Err` field and `Unwrap` method.

runtime/errors.go
```diff
@@ -78,6 +78,36 @@ func (e InvalidTransactionCountError) Error() string {
 	)
 }
 
+// MissingEntryPointError
```
turbolent, line 81: Thanks for making this a proper error 👍

runtime/errors.go
```diff
@@ -78,6 +78,36 @@ func (e InvalidTransactionCountError) Error() string {
...
+// InvalidEntryPointError
+
+type InvalidEntryPointTypeError struct {
+	Type sema.Type
+	Err  error
+}
+
+func (e *InvalidEntryPointTypeError) Unwrap() error {
+	return e.Err
+}
```
turbolent, line 105: Same here, maybe just remove the `Err` field and `Unwrap` method.

runtime/errors.go
```diff
@@ -123,6 +153,21 @@ func (e *InvalidTransactionArgumentError) Error() string {
...
+// InvalidScriptArgumentError
+
+type InvalidScriptArgumentError struct {
+	Index int
+	Err   error
+}
+
+func (e *InvalidScriptArgumentError) Unwrap() error {
+	return e.Err
+}
```
turbolent, line 165: Same here, maybe just remove the `Err` field and `Unwrap` method.

runtime/runtime.go
```diff
@@ -161,6 +167,53 @@ func (r *interpreterRuntime) ExecuteScript(
...
+					Index: i,
+					Err: &InvalidTypeAssignmentError{
+						Value: arg,
+						Type:  parameterType,
+					},
+				}
+			}
+
+			argumentValues[i] = arg
+		}
```
turbolent, line 211: This is the same code as https://github.com/onflow/cadence/blob/a968f040503e86274f7ac9c522110ce66e99b110/runtime/runtime.go#L349-L386, which is correct, but maybe refactor both into a common function to ensure the logic stays the same for both entry points (transactions and scripts)

benjaminkvm:
Still have the `transaction/scriptExecutionFunction` fyi though much smaller as they now call `validateArgumentParameters` to remove duplicate code, let me know if you would prefer those methods as closures in their respective methods.

Had to change the error name to something that would work for both scripts and transactions, let me know what you think!

runtime/errors.go
```diff
@@ -78,14 +78,34 @@ func (e InvalidTransactionCountError) Error() string {
...
-type InvalidTransactionParameterCountError struct {
+type InvalidEntryPointParameterCountError struct {
 	Expected int
 	Actual   int
 }
 
-func (e InvalidTransactionParameterCountError) Error() string {
+func (e InvalidEntryPointParameterCountError) Error() string {
 	return fmt.Sprintf(
 		"parameter count mismatch for transaction: expected %d, got %d",
```
psiemens, line 110:
```suggestion
		"parameter count mismatch for entry point: expected %d, got %d",
```

runtime/runtime.go
```diff
@@ -130,22 +131,27 @@ func (r *interpreterRuntime) ExecuteScript(
 		return nil, newError(err)
 	}
 
-	_, ok := checker.GlobalValues["main"]
+	ep, ok := checker.GlobalValues["main"]
 	if !ok {
-		// TODO: error because no main?
-		return nil, nil
+		return nil, &MissingEntryPointError{Expected: "main"}
```
psiemens, line 136: 💯

runtime/runtime_test.go
```diff
@@ -751,6 +751,276 @@ func TestRuntimeTransactionWithArguments(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.label, func(t *testing.T) {
+			rt := NewInterpreterRuntime()
+
+			var loggedMessages []string
+
+			runtimeInterface := &testRuntimeInterface{
+				// getSigningAccounts: func() []Address { return nil },
```
psiemens, line 993:
```suggestion
```

runtime/runtime_test.go
```diff
@@ -751,6 +751,276 @@ func TestRuntimeTransactionWithArguments(t *testing.T) {
...
+
+			if tt.check != nil {
+				tt.check(t, err)
+			} else {
+				if !assert.NoError(t, err) {
+					for err := err; err != nil; err = errors.Unwrap(err) {
+						t.Log(err)
+					}
+				}
+				fmt.Println(loggedMessages)
```
psiemens, line 1017: Best to use `t.Log` if this is still needed

psiemens:
@turbolent This PR is blocked on a code owner approval, but I'm going to merge as an admin in order to unblock the remaining work in `flow-go` and `flow-emulator` this sprint. Let us know if anything else comes up and we can open up a quick follow-up PR.

psiemens:
Never mind, didn't need to break protocol!

PR #152, opened 2020-06-03
Language Server: Upgrade Cadence and Flow Go SDK to v0.4.0

psiemens:
## Description

This PR updates the Cadence LSP server to use the latest version of the `onflow/cadence` and `onflow/flow-go-sdk`.

- Change from "root account" to "service account"
- Transaction authorizer now required to create accounts


languageserver/server/commands.go
```diff
@@ -452,20 +452,19 @@ func (s *Server) createAccountHelper(conn protocol.Conn) (addr flow.Address, err
 		}
 	}
 
-	if len(txResult.Events) != 1 {
-		return addr, fmt.Errorf("failed to get new account address for tx %s", txID.Hex())
+	for _, event := range txResult.Events {
```
jordanschalm, line 455: 👍

PR #153, opened 2020-06-03
VS Code Extension: Version 0.4.0

psiemens:
## Description

This PR updates the VS Code extension to be compatible with the latest language server (#152).

- Change `root` to `service` in code and settings
- Update account picker to display account labels (see below)

Screen Shot 2020-06-03 at 11 37 03 AM
Screen Shot 2020-06-03 at 11 37 08 AM

### DIFF
--- a/tools/vscode-extension/src/config.ts
+++ b/tools/vscode-extension/src/config.ts
@@ -1,31 +1,43 @@
 import {commands, window, workspace} from "vscode";
-import {shortAddress} from "./address";
+import { addAddressPrefix } from "./address";
 
-export const ROOT_ADDR: string = shortAddress("0000000000000000000000000000000000000001");
+export const SERVICE_ADDR: string = "f8d6e0586b0a20c7";
 
 const CONFIG_FLOW_COMMAND = "flowCommand";
-const CONFIG_ROOT_PRIVATE_KEY = "rootPrivateKey";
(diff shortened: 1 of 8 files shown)


tools/vscode-extension/src/config.ts
```diff
@@ -1,59 +1,87 @@
 import {commands, window, workspace} from "vscode";
-import {shortAddress} from "./address";
+import { addAddressPrefix } from "./address";
 
-export const ROOT_ADDR: string = shortAddress("0000000000000000000000000000000000000001");
+export const SERVICE_ADDR: string = "f8d6e0586b0a20c7";
```
turbolent, line 4: Will this be the same for the emulator and the mainnet?

psiemens: No, mainnet will have a different service address 😢 

I wasn't sure how to best handle this. Do we have any users using this extension with anything other than the emulator? Is that even possible at this point?

I can see _direct-to-mainnet_ or _direct-to-testnet_ being a feature in the future, but until we're ready to figure out how that works (i.e. with keys, accounts, etc), I figured it'd be safe to hardcode the emulator address for now.

turbolent: I was just curious, this is totally fine 👍 

I think the extension is mainly used for development

PR #157, opened 2020-06-07
Resources are temporarily moved into the invoked function

### LINKED ISSUE #73: BUG: Objects shouldn't be able to use themselves as arguments to their functions
### Problem

If a resource or struct has a member function that takes a resource of the same type as an argument, then it should not be able to use itself as an argument.

### Steps to Reproduce

1. Copy this contract:
```
pub contract Nodes {

    pub resource Test {
        pub fun deposit(from: @Test) {
            destroy from
        }
    }

    init() {

        let test <- create Test()

        // this line should fail
        test.deposit(from: <-test)
    }
}
```
2. See that there is no error reported

### Acceptance Criteria

Should invalidate the object that is being invoked so it cannot be used as an argument to itself.

turbolent:
Closes #73
Closes dapperlabs/flow-go#3932

Properly check that an invocation of a function of a resource is not using the resource after an invalidation, e.g. when the resource was invalidated by moving it into the function as an argument.

### DIFF
--- a/runtime/tests/checker/resources_test.go
+++ b/runtime/tests/checker/resources_test.go
@@ -4438,6 +4438,29 @@ func TestCheckResourceMoveMemberInvocation(t *testing.T) {
 		require.NoError(t, err)
 	})
 
+	t.Run("valid use, with argument of same type", func(t *testing.T) {
+
+		t.Parallel()
+
+		_, err := ParseAndCheck(t, `
+
+          resource Test {
+              fun use(_ test: @Test) {
+                  destroy test
+              }
+          }
+
+          fun test() {
+              let test1 <- create Test()
+              let test2 <- create Test()
+              test1.use(<-test2)
+              destroy test1
+          }
+        `)
+
+		require.NoError(t, err)
+	})
+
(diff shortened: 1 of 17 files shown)


runtime/tests/checker/resources_test.go
```diff
@@ -4387,5 +4387,96 @@ func TestCheckInvalidResourceLossInNestedContractResource(t *testing.T) {
...
+              test.use(<-test)
+          }
+        `)
+
+		errs := ExpectCheckerErrors(t, err, 1)
+
+		assert.IsType(t, &sema.ResourceUseAfterInvalidationError{}, errs[0])
+	})
+
+	t.Run("valid use", func(t *testing.T) {
```
benjaminkvm, line 4421: Would perhaps be good to have a test case that includes a Resource method that accepts a Resource of the same type, and is invoked with a non `self` instance.

turbolent: Not quite sure if it's what you suggested, but added another test case in 94c7a3b75351b7b15589b423563e8c768c7fc877

PR #160, opened 2020-06-16
Add `isInstance` function

turbolent:
Closes dapperlabs/flow-go#3269

Add an implicit function `fun isInstance(_ type: Type): Bool` to all values, which returns true if the value has the given type

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -1474,16 +1474,16 @@ func (interpreter *Interpreter) indexExpressionGetterSetter(indexExpression *ast
 func (interpreter *Interpreter) memberExpressionGetterSetter(memberExpression *ast.MemberExpression) Trampoline {
 	return memberExpression.Expression.Accept(interpreter).(Trampoline).
 		FlatMap(func(result interface{}) Trampoline {
-			structure := result.(MemberAccessibleValue)
+			value := result.(Value)
 			locationRange := interpreter.locationRange(memberExpression)
 			identifier := memberExpression.Identifier.Identifier
 			return Done{
 				Result: getterSetter{
 					get: func() Value {
-						return structure.GetMember(interpreter, locationRange, identifier)
+						return interpreter.getMember(value, locationRange, identifier)
 					},
 					set: func(value Value) {
-						structure.SetMember(interpreter, locationRange, identifier, value)
+						interpreter.setMember(value, locationRange, identifier, value)
 					},
 				},
 			}
@@ -1957,9 +1957,9 @@ func (interpreter *Interpreter) VisitMemberExpression(expression *ast.MemberExpr
 				}
 			}
 
-			value := result.(MemberAccessibleValue)
+			value := result.(Value)
 			locationRange := interpreter.locationRange(expression)
-			resultValue := value.GetMember(interpreter, locationRange, expression.Identifier.Identifier)
+			resultValue := interpreter.getMember(value, locationRange, expression.Identifier.Identifier)
 
 			// If the member access is optional chaining, only wrap the result value
 			// in an optional, if it is not already an optional value
@@ -4182,4 +4182,30 @@ func (interpreter *Interpreter) reportFunctionInvocation(pos ast.HasPosition) {
 
 	line := pos.StartPosition().Line
 	interpreter.onFunctionInvocation(interpreter, line)
+}
+
+func (interpreter *Interpreter) getMember(self Value, locationRange LocationRange, identifier string) Value {
+	var result Value
+	if memberAccessibleValue, ok := self.(MemberAccessibleValue); ok {
+		result = memberAccessibleValue.GetMember(interpreter, locationRange, identifier)
+	}
+	if result == nil {
+		switch identifier {
+		case sema.IsInstanceFunctionName:
+			return NewHostFunctionValue(
+				func(invocation Invocation) Trampoline {
+					// NOTE: not invocation.Self, as that is only set for composite values
+					dynamicType := self.DynamicType(interpreter)
+					ty := invocation.Arguments[0].(TypeValue)
+					result := IsSubType(dynamicType, ty.Type)
+					return Done{Result: BoolValue(result)}
+				},
+			)
+		}
+	}
+	return result
+}
+
+func (interpreter *Interpreter) setMember(self Value, locationRange LocationRange, identifier string, value Value) {
+	self.(MemberAccessibleValue).SetMember(interpreter, locationRange, identifier, value)
 }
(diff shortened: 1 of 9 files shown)


runtime/interpreter/interpreter.go
```diff
@@ -4092,3 +4090,32 @@ func (interpreter *Interpreter) reportFunctionInvocation(pos ast.HasPosition) {
 	line := pos.StartPosition().Line
 	interpreter.onFunctionInvocation(interpreter, line)
 }
+
+func (interpreter *Interpreter) getMember(self Value, locationRange LocationRange, identifier string) Value {
```
zhangchiqing, line 4094:
```suggestion
// getMember gets the member value by the given identifier from the given Value depending on its type.
func (interpreter *Interpreter) getMember(self Value, locationRange LocationRange, identifier string) Value {
```

runtime/interpreter/interpreter.go
```diff
@@ -4092,3 +4090,32 @@ func (interpreter *Interpreter) reportFunctionInvocation(pos ast.HasPosition) {
...
+	var result Value
+	if memberAccessibleValue, ok := self.(MemberAccessibleValue); ok {
```
zhangchiqing, line 4096:
```suggestion
	// when a Value has member defined, try to get the member by the identifier
	// for instance, String has a predeclared member "length". So `"12".length` 
	// will match the predeclared member with the identifier "length".
	if memberAccessibleValue, ok := self.(MemberAccessibleValue); ok {
```

runtime/tests/checker/metatype_test.go
```diff
@@ -62,3 +64,77 @@ func TestCheckMetaType(t *testing.T) {
...
+		require.NoError(t, err)
+
+		assert.Equal(t,
+			&sema.BoolType{},
+			checker.GlobalValues["result"].Type,
+		)
+	})
+}
+
+func TestCheckIsInstance_Redeclaration(t *testing.T) {
```
benjaminkvm, line 127: Was going to ask about this case! Nice work! 👍

runtime/interpreter/interpreter.go
```diff
@@ -4092,3 +4090,32 @@ func (interpreter *Interpreter) reportFunctionInvocation(pos ast.HasPosition) {
...
+		result = memberAccessibleValue.GetMember(interpreter, locationRange, identifier)
+	}
+	if result == nil {
+		switch identifier {
```
zhangchiqing, line 4100:
```suggestion
		// if the given Value doesn't have the member predefined,
		// then check if there is a member defined for any type.
		// for instance the `IsInstance` function is a builtin function
		// to return whether the value is of a certain type,
		// like `(1).IsInstance(Type<Int>())`
		switch identifier {
```

runtime/sema/check_member_expression.go
```diff
@@ -233,6 +236,20 @@ func (checker *Checker) visitMember(expression *ast.MemberExpression) (member *M
 	return member, isOptional
 }
 
+func getBuiltinMember(identifier string, ty Type) *Member {
```
zhangchiqing, line 239:
```suggestion
// getBuildInMember tries to get the member of the given type by the identifier.
// currently there is one buildin member defined for all types, which is the 
// `IsInstance` function.
func getBuiltinMember(identifier string, ty Type) *Member {
```

runtime/sema/checker.go
```diff
@@ -1890,6 +1894,14 @@ func (checker *Checker) predeclaredMembers(containerType Type) []*Member {
 
 	if compositeKindedType, ok := containerType.(CompositeKindedType); ok {
 
+		addPredeclaredMember(
```
zhangchiqing, line 1897: Why this can only be defined when `containerType` is a `CompositeKindedType`?

Why not moving this out of the above `if`

runtime/interpreter/interpreter.go
```diff
@@ -2446,7 +2444,7 @@ func (interpreter *Interpreter) declareCompositeValue(
 
 			if declaration.CompositeKind == common.CompositeKindResource {
 				uuid := interpreter.uuidHandler()
-				fields[ResourceUUIDMemberName] = UInt64Value(uuid)
+				fields[sema.UUIDFieldName] = UInt64Value(uuid)
```
benjaminkvm, line 2447: Just curious about the UUID changes, looks like you are adding a predeclared `uuid` member also?

runtime/tests/interpreter/metatype_test.go
```diff
@@ -82,3 +82,84 @@ func TestInterpretMetaType(t *testing.T) {
 		)
 	})
 }
+
+func TestInterpretIsInstance(t *testing.T) {
+
+	t.Parallel()
+
+	cases := []struct {
```
zhangchiqing, line 90: I added more test cases and restructured the test cases a bit so that it's easy to add and review them

runtime/interpreter/interpreter.go
turbolent:
```suggestion
	// When the accessed value has a type that supports the declaration of members or is a built-in type that has members (`MemberAccessibleValue`), 
	// then try to get the member for the given identifier.
	// For example, the built-in type `String` has a member "length",
	// and composite declarations may contain member declarations
	if memberAccessibleValue, ok := self.(MemberAccessibleValue); ok {
```

runtime/sema/check_member_expression.go
turbolent:
```suggestion
// getBuiltinMember tries to get the built-in member with the given identifier.
// For example, the function `isInstance` is a built-in which is defined for all types
func getBuiltinMember(identifier string, ty Type) *Member {
```

runtime/tests/interpreter/metatype_test.go
turbolent: @zhangchiqing Thanks for the additional test cases. 

To make it easier to debug the tests, can you please give them names and run them as sub-tests using `t.Run`. See e.g. https://dave.cheney.net/2019/05/07/prefer-table-driven-tests "Introducing sub tests"

runtime/interpreter/interpreter.go
turbolent: Yes, I just saw that we were redeclaring a constant for the existing `uuid` field name: `ResourceUUIDMemberName`. It's better to have a single source of truth, which is `sema.UUIDFieldName`

runtime/sema/checker.go
turbolent: Good point, I moved it out of the if-statement in 4e4fc03ac891c540b8c88af4b665001c5c088144

turbolent:
@zhangchiqing Ah, I had missed it, sorry! I addressed it in 4e4fc03ac891c540b8c88af4b665001c5c088144

PR #166, opened 2020-06-18
Fix assignment statement parsing and buffering

turbolent:
The parsing of assignments without any space, e.g. `x=1`, was broken.

The reason was that both that the token after the transfer (e.g. `=`) was skipped.

This uncovered another bug in the parser when buffering: When starting buffering, only buffer the current token, if there are no buffered tokens to be read.

Also, fix the position for uninitialized field access errors

### DIFF
--- a/runtime/parser2/parser.go
+++ b/runtime/parser2/parser.go
@@ -264,6 +264,10 @@ func (p *parser) skipSpaceAndComments(skipNewlines bool) (containsNewline bool) 
 
 func (p *parser) startBuffering() {
 	p.buffering = true
+ 
+ 	// starting buffering should only buffer the current token if there’s nothing 
+ 	// to be read from the buffer, otherwise, the current token would be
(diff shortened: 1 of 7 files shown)


runtime/parser2/parser.go
```diff
@@ -264,7 +264,9 @@ func (p *parser) skipSpaceAndComments(skipNewlines bool) (containsNewline bool)
 
 func (p *parser) startBuffering() {
 	p.buffering = true
-	p.bufferedTokens = append(p.bufferedTokens, p.current)
+	if p.bufferPos >= len(p.bufferedTokens) {
```
zhangchiqing, line 267:
```suggestion
 
 	// starting buffering should only buffer the current token if there’s nothing 
 	// to be read from the buffer, otherwise, the current token would be
 	// buffered twice
	if p.bufferPos >= len(p.bufferedTokens) {
```

PR #177, opened 2020-06-25
Report missing return value in return statement

### LINKED ISSUE #176: Missing return value in return statement is not reported
### Problem

If a function has a non-Void return type, the return statements in the function must have a return value.
However, currently no error is reported when the a return statement has no value.

### Steps to Reproduce

The following program is invalid, because the return statement has no return value, but no error is reported:

```swift
fun test(): Int {
    return
}
```	

### Acceptance Criteria

- [ ] An error should be reported

turbolent:
Closes #176 

Properly check return statements and report an error if the enclosing function has a non-Void return type, but the return statement has no value.

### DIFF
--- a/runtime/sema/check_return_statement.go
+++ b/runtime/sema/check_return_statement.go
@@ -30,47 +30,66 @@ func (checker *Checker) VisitReturnStatement(statement *ast.ReturnStatement) ast
 		functionActivation.ReturnInfo.DefinitelyReturned = true
 	}()
 
-	// check value type matches enclosing function's return type
+	returnType := functionActivation.ReturnType
 
 	if statement.Expression == nil {
+
+		// If the return statement has no expression,
+		// and the enclosing function's return type is Void,
+		// then the return statement is missing an expression
+
+		if _, ok := returnType.(*VoidType); !ok {
+			checker.report(
+				&MissingReturnValueError{
+					ExpectedValueType: returnType,
+					Range:             ast.NewRangeFromPositioned(statement),
+				},
+			)
+		}
+
 		return nil
 	}
 
+	// If the return statement has a return value,
+	// check that the value's type matches the enclosing function's return type
+
 	valueType := statement.Expression.Accept(checker).(Type)
-	returnType := functionActivation.ReturnType
 
 	checker.Elaboration.ReturnStatementValueTypes[statement] = valueType
 	checker.Elaboration.ReturnStatementReturnTypes[statement] = returnType
 
-	if valueType == nil {
-		return nil
-	}
+	// If the enclosing function's return type is Void,
+	// then the return statement should not have a return value
 
-	// return statement has expression, but function has Void return type?
 	if _, ok := returnType.(*VoidType); ok {
 		checker.report(
 			&InvalidReturnValueError{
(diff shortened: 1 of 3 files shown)


runtime/sema/check_return_statement.go
```diff
@@ -30,48 +30,67 @@ func (checker *Checker) VisitReturnStatement(statement *ast.ReturnStatement) ast
...
-				},
-			)
-		}
+	// The return statement has a return value,
+	// and the enclosing function has a non-Void return type.
+	// Check that the types are compatible
 
-		checker.checkVariableMove(statement.Expression)
-		checker.checkResourceMoveOperation(statement.Expression, valueType)
+	if !valueType.IsInvalidType() &&
```
zhangchiqing, line 78: Should these `&&` be `||`?
```suggestion
	if !valueType.IsInvalidType() ||
```

zhangchiqing: Could you add a test case where `valueType.IsInvalidType == true` but `valueType.IsInvalidType == false`?

turbolent: No, the and is correct: We want to only check the type compatibility between the value type and return type if both of them are valid, i.e. both of them are not invalid: `!valueType.IsInvalidType() && !returnType.IsInvalidType()`. I'll add some tests

zhangchiqing: I see. 

I wonder if makes sense to call it `IsValidType`, since it's easier to read as `if valueType.IsValidType() && returnType.IsValidType() && ...`

And what would happen if `valueType` and `returnType` are invalid types? Who will handle those cases?

turbolent: Yeah, it could be renamed, but I don't think it's worth the effort.

If any of `valueType` or `returnType` are invalid, then nothing needs to be done, as the invalid type was produced by an earlier check. The whole point of the invalid type is to prevent down stream error reports that are based on earlier errors.

For example, the return type annotation could have could have referred to an unknown type, in which case it is invalid. An error is reported for the return type annotation. As a result, the compatibility check here for the return statement will always fail, and an error would be generated. That's valid, but we attempt to reduce the amount of errors to prevent error fatigue.

turbolent: Added the tests in 8c4bc8709b00121ff5e07ee9c1a5bfeabe4ef892

runtime/sema/check_return_statement.go
```diff
@@ -30,48 +30,67 @@ func (checker *Checker) VisitReturnStatement(statement *ast.ReturnStatement) ast
 		functionActivation.ReturnInfo.DefinitelyReturned = true
 	}()
 
-	// check value type matches enclosing function's return type
+	returnType := functionActivation.ReturnType
 
 	if statement.Expression == nil {
+
+		// If the return statement has no expression,
+		// and the enclosing function's return type is Void,
```
psiemens, line 38:
```suggestion
		// and the enclosing function's return type is non-Void,
```

Is this what you meant?

turbolent: Yes, good catch!

PR #181, opened 2020-06-26
Add a Crypto contract

### LINKED ISSUE #173: Add a Crypto contract
## Context

Contracts may want to verify signatures.

Implement a new `Crypto` module which defines:
- A type for signature algorithms, `SignatureAlgorithm`, and constants for the following signature algorithms:
  - Elliptic Curve Digital Signature Algorithm (ECDSA) on the NIST P-256 curve
  - Elliptic Curve Digital Signature Algorithm (ECDSA) on the secp256k1 curve
- A type for hash algorithms, `HashAlgorithm`, and constants for the following hash algorithms:
  - Secure Hashing Algorithm 2 (SHA-2) with a 256-bit digest
  - Secure Hashing Algorithm 3 (SHA-3) with a 256-bit digest
- A type for public keys, `PublicKey`, wich consists of the key data and signature algorithm
- A type for a collection of public keys, `KeyList`, which allows:
  - Adding entries (`fun add(_ publicKey: PublicKey, hashAlgorithm: HashAlgorithm, weight: UInt32): KeyListEntry`)
  - Getting entries (`fun get(keyIndex: Int): KeyListEntry?`)
  - Revoking entries (`fun revoke(keyIndex: Int)`; entries cannot be deleted)
  - Checking the validity of a set of signatures for some given data 
    (`fun isValid(signatureSet: [KeyListSignature], signedData: [UInt8]): Bool`). 
    As key list entries have a weight, this function supports the multi-sig use-case.
- A type for a key list entry, `KeyListEntry`, which has an index (`keyIndex`), public key (`publicKey`), hash algorithm (`hashAlgorithm`), and has a weight
- A type for signatures, `KeyListSignature`, which consists of the signature data (`signature`), and refers to a key in a key list through a key index (`keyIndex`)

## Proposal

```swift

pub struct interface SignatureVerifier  {

    pub fun verify(
        signature: [UInt8],
        tag: String,
        signedData: [UInt8],
        publicKey: [UInt8],
        signatureAlgorithm: String,
        hashAlgorithm: String
    ): Bool
}

pub contract Crypto {

    pub struct SignatureAlgorithm {
        pub let name: String

        init(name: String) {
            self.name = name
        }
    }

    // ECDSA_P256 is Elliptic Curve Digital Signature Algorithm (ECDSA) on the NIST P-256 curve
    pub let ECDSA_P256: SignatureAlgorithm

    // ECDSA_Secp256k1 is Elliptic Curve Digital Signature Algorithm (ECDSA) on the secp256k1 curve
    pub let ECDSA_Secp256k1: SignatureAlgorithm

    pub struct HashAlgorithm {
        pub let name: String

        init(name: String) {
            self.name = name
        }
    }

    // SHA2_256 is Secure Hashing Algorithm 2 (SHA-2) with a 256-bit digest
    pub let SHA2_256: HashAlgorithm

    // SHA3_256 is Secure Hashing Algorithm 3 (SHA-3) with a 256-bit digest
    pub let SHA3_256: HashAlgorithm

    pub struct PublicKey {
        pub let publicKey: [UInt8]
        pub let signatureAlgorithm: SignatureAlgorithm

        init(publicKey: [UInt8], signatureAlgorithm: SignatureAlgorithm) {
            self.publicKey = publicKey
            self.signatureAlgorithm = signatureAlgorithm
        }
    }

    pub struct KeyListEntry {
        pub let keyIndex: Int
        pub let publicKey: PublicKey
        pub let hashAlgorithm: HashAlgorithm
        pub let weight: UFix64
        pub let isRevoked: Bool

        init(
            keyIndex: Int,
            publicKey: PublicKey,
            hashAlgorithm: HashAlgorithm,
            weight: UFix64,
            isRevoked: Bool
        ) {
            self.keyIndex = keyIndex
            self.publicKey = publicKey
            self.hashAlgorithm = hashAlgorithm
            self.weight = weight
            self.isRevoked = isRevoked
        }
    }

    pub struct KeyList {

        priv let entries: [KeyListEntry]

        init() {
            self.entries = []
        }

        // Adds a new key with the given weight
        pub fun add(
            _ publicKey: PublicKey,
            hashAlgorithm: HashAlgorithm,
            weight: UFix64
        ): KeyListEntry {

            let keyIndex = self.entries.length
            let entry = KeyListEntry(
                keyIndex: keyIndex,
                publicKey: publicKey,
                hashAlgorithm: hashAlgorithm,
                weight: weight,
                isRevoked: false
            )
            self.entries.append(entry)
            return entry
        }

        // Returns the key at the given index, if it exists.
        // Revoked keys are always returned, but they have `isRevoked` field set to true
        pub fun get(keyIndex: Int): KeyListEntry? {
            if keyIndex >= self.entries.length {
                return nil
            }

            return self.entries[keyIndex]
        }

        // Marks the key at the given index revoked, but does not delete it
        pub fun revoke(keyIndex: Int) {
            if keyIndex >= self.entries.length {
                return
            }
            let currentEntry = self.entries[keyIndex]
            self.entries[keyIndex] = KeyListEntry(
                keyIndex: currentEntry.keyIndex,
                publicKey: currentEntry.publicKey,
                hashAlgorithm: currentEntry.hashAlgorithm,
                weight: currentEntry.weight,
                isRevoked: true
            )
        }

        pub fun isValid(
            signatureSet: [KeyListSignature],
            signedData: [UInt8]
        ): Bool {

            var validWeights: UFix64 = 0.0

            let seenKeyIndices: {Int: Bool} = {}

            for signature in signatureSet {

                // Ensure the key index is valid

                if signature.keyIndex >= self.entries.length {
                    return false
                }

                // Ensure this key index has not already been seen

                if seenKeyIndices[signature.keyIndex] ?? false {
                    return false
                }

                // Record the key index was seen

                seenKeyIndices[signature.keyIndex] = true

                // Get the actual key

                let key = self.entries[signature.keyIndex]

                // Ensure the key is not revoked

                if key.isRevoked {
                    return false
                }

                // Ensure the signature is valid

                if !Crypto.signatureVerifier.verify(
                    signature: signature.signature,
                    tag: Crypto.domainSeparationTagFlowUser,
                    signedData: signedData,
                    publicKey: key.publicKey.publicKey,
                    signatureAlgorithm: key.publicKey.signatureAlgorithm.name,
                    hashAlgorithm:key.hashAlgorithm.name
                ) {
                    return false
                }

                validWeights = validWeights + key.weight
            }

            return validWeights >= 1.0
        }
    }

    pub struct KeyListSignature {
        pub let keyIndex: Int
        pub let signature: [UInt8]

        pub init(keyIndex: Int, signature: [UInt8]) {
            self.keyIndex = keyIndex
            self.signature = signature
        }
    }

    priv let domainSeparationTagFlowUser: String

    priv let signatureVerifier: {SignatureVerifier}

    init(signatureVerifier: {SignatureVerifier}) {

        self.signatureVerifier = signatureVerifier

        // Initialize constants

        self.ECDSA_P256 = SignatureAlgorithm(name: "ECDSA_P256")
        self.ECDSA_Secp256k1 = SignatureAlgorithm(name: "ECDSA_Secp256k1")

        self.SHA2_256 = HashAlgorithm(name: "SHA2_256")
        self.SHA3_256 = HashAlgorithm(name: "SHA3_256")

        self.domainSeparationTagFlowUser = "FLOW-V0.0-user"
    }
}
```

## Definition of Done

- [ ] Sema
  - [ ] Add built-in module
  - [ ] Declare types
  - [ ] Declare constants
- [ ] Interpreter
  - [ ] Add built-in module
  - [ ] Implement types
  - [ ] Implement functions
- [ ] Tests
  - [ ] Signature verification
  - [ ] Storage of types

turbolent:
Closes #173

This PR adds a new Crypto contract, which can be imported by programs using `import Crypto`, and provides types for keys and signatures, and a function to verify if a set of signatures is valid.

This is implemented by embedding the contract into the runtime and instantiating it when it is imported.

Currently, this PR adds a new function to the runtime interface, requiring embedders to provide the actual implementation for verifying a signature. Alternatively this could be changed in a way where this function is not added, and the runtime performs the signature verification itself. However, with the current state of open-sourcing code and repository organization, this would mean duplicating the Go crypto library as it exists in https://github.com/onflow/flow-go-sdk, or making it a separate package, as the SDK itself depends on Cadence.


runtime/runtime.go
```diff
@@ -634,21 +662,27 @@ func (r *interpreterRuntime) injectedCompositeFieldsHandler(
 		compositeKind common.CompositeKind,
 	) map[string]interpreter.Value {
 
-		switch compositeKind {
-		case common.CompositeKindContract:
-			var address []byte
+		switch location {
+		case stdlib.CryptoChecker.Location:
```
psiemens, line 666: Do you have ideas for how to standardize these kinds of operations once more standard libraries are added? I'm just a bit nervous about having a special case in all of these switch statements.

runtime/runtime.go
```diff
@@ -508,6 +508,15 @@ func (r *interpreterRuntime) parseAndCheckProgram(
...
 				sema.WithPredeclaredTypes(typeDeclarations),
 				sema.WithValidTopLevelDeclarationsHandler(validTopLevelDeclarations),
+				sema.WithImportHandler(func(location ast.Location) sema.Import {
+					switch location {
+					case stdlib.CryptoChecker.Location:
+						return sema.CheckerImport{
+							Checker: stdlib.CryptoChecker,
+						}
+					}
+					return nil
```
psiemens, line 518: So if this import handler returns `nil`, does the checker call up to the host environment?

turbolent:
Thanks for the review @psiemens! Great questions.

The current architecture is messy: There's currently not a clear separation what is considered "Core Cadence" and what is Flow. Also, some parts that could be considered Flow specific follow a "pull" architecture, where the embedder must implement/provide all functionality (`runtime.Interface`), though it might be optional.

We've been discussing this in the Cadence working session a bit. One idea was to invert the current "pull" architecture and inject everything and leave knowledge about Flow specific parts out of the core Cadence code, but there are some dependency problems here (e.g. the account types are special and knowledge about them is required to e.g. check/run transactions; we want to use the key types in the Crypto contract for the keys of accounts; etc)

I'd really like to refactor this all and clean it up, though I don't have a clear picture yet what the result of the refactor should be. If you have suggestions/ideas, let's please sync, I'd love to hear them!

runtime/runtime.go
turbolent: It would be nice to make this more declarative, similar to / in `FlowBuiltinImpls`, but as noted below I don't have good ideas yet. If you do, please let me know!

runtime/runtime.go
turbolent: It's actually the other way round: The checker first assumes the import is a program. If it is not, the import handler (if any) is called. If the import handler is still returning nil, then the import can not be resolved and an error is reported.

See https://github.com/onflow/cadence/blob/236a6ecdfbdbe7247bc3d6884209324fc0fe1e4d/runtime/sema/check_import_declaration.go#L55-L99

psiemens: Maybe I'm missing something, but doesn't this block contracts from being imported?

psiemens:
> I'd really like to refactor this all and clean it up, though I don't have a clear picture yet what the result of the refactor should be. If you have suggestions/ideas, let's please sync, I'd love to hear them!

I thought of a few possible improvements while I was working through the VM refactor and would love to discuss! None if it is very fleshed out at this point, but in general I really like the idea of being able to inject modular functionality into the runtime.

runtime/runtime.go
turbolent: Hmm, how would it block that?

Is it because there's no call to the runtime interface to get the code for the location?
If so, `ast.Program.ImportedPrograms` is resolved in the runtime before: 

https://github.com/onflow/cadence/blob/a968f040503e86274f7ac9c522110ce66e99b110/runtime/runtime.go#L434-L435

psiemens: Oh I see, that makes sense!

runtime/stdlib/contracts/crypto.cdc
```diff
@@ -0,0 +1,213 @@
...
+                    signatureAlgorithm: key.publicKey.signatureAlgorithm.name,
+                    hashAlgorithm:key.hashAlgorithm.name
+                ) {
+                    return false
+                }
+
+                validWeights = validWeights + key.weight
+            }
+
+            return validWeights >= 1.0
```
tarakby, line 181: 👌

runtime/stdlib/crypto.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/onflow/cadence/runtime/ast"
+	"github.com/onflow/cadence/runtime/common"
+	"github.com/onflow/cadence/runtime/interpreter"
+	"github.com/onflow/cadence/runtime/parser2"
+	"github.com/onflow/cadence/runtime/sema"
+	"github.com/onflow/cadence/runtime/stdlib/internal"
+	"github.com/onflow/cadence/runtime/trampoline"
+)
+
+type CryptoSignatureVerifier interface {
```
tarakby, line 16: 👌

PR #184, opened 2020-06-29
Make type values storable

### LINKED ISSUE #179: Ensure `Type` value is storable
### Context

We've added the `Type` type and value. Make sure it can be saved to and loaded from storage.

### Definition of Done

- [ ] Interpreter: Ensure `TypeValue` is storable

turbolent:
Closes #179


runtime/interpreter/value.go
```diff
@@ -125,18 +125,23 @@ func (v TypeValue) String() string {
...
 	otherMetaType, ok := other.(TypeValue)
 	if !ok {
 		return false
 	}
-	return BoolValue(v.Type.Equal(otherMetaType.Type))
+
+	ty := inter.convertStaticToSemaType(v.Type)
+	otherTy := inter.convertStaticToSemaType(otherMetaType.Type)
+
+	return BoolValue(ty.Equal(otherTy))
```
zhangchiqing, line 137: I understand that now checking the equality of two types need interpreter to convert types first.

I wonder you can think of a case where the result here would be different from the result of `BoolValue(v.Type.Equal(otherMetaType.Type))` (the previous result).

runtime/interpreter/value.go
```diff
@@ -73,7 +73,7 @@ type ConcatenatableValue interface {
 
 type EquatableValue interface {
 	Value
-	Equal(other Value) BoolValue
+	Equal(interpreter *Interpreter, other Value) BoolValue
```
zhangchiqing, line 76: The `interpreter` is only useful for `TypeValue`, and for other types, it's not needed.

Another approach would be to ask if the type of the values need to be converted first. If yes, then convert them using the interpreter, and then call equal with the converted values. So if a value is a TypeValue, its type needs to be converted to `sema.Type`, and `sema.Type` has `Equal` defined.

On a side note, we might need to prevent `convertStaticToSemaType` from returning a `TypeValue`. Because otherwise, the `Equal` will trigger another `convertStaticToSemaType` call, and potentially run into endless loop.

runtime/interpreter/decode.go
```diff
@@ -1161,3 +1164,24 @@ func (d *Decoder) decodeRestrictedStaticType(v interface{}) (StaticType, error)
...
 	}, nil
 }
+
+func (d *Decoder) decodeType(v interface{}) (TypeValue, error) {
+	encoded, ok := v.(map[interface{}]interface{})
+	if !ok {
+		return TypeValue{}, fmt.Errorf("invalid type encoding")
+	}
+
+	decodedStaticType, err := d.decodeStaticType(encoded[uint64(0)])
```
zhangchiqing, line 1174: why using `0`? because [we defined here](https://github.com/onflow/cadence/pull/184/files#diff-b4059fff0bb0bd33abbff410a57cec00R1077)?

runtime/interpreter/value.go
turbolent: It would not be any different, I just did not want to re-implement equality for the `StaticType` type hierarchy

runtime/interpreter/value.go
turbolent: Yeah, maybe conditionally converting might be a good idea, I'll look into it

runtime/interpreter/decode.go
turbolent: Correct. I really don't like that, but so far I haven't found a better way

zhangchiqing: Maybe we could define a const right above where it was used for reference:

```
const (
  EncodedTypeValueType uint64 = 0
)

type encodedTypeValue struct {
	Type interface{} `cbor:"0,keyasint"`
}
```

And then here, we can do 
```
	decodedStaticType, err := d.decodeStaticType(encoded[EncodedTypeValueType])

```

turbolent: I opened #216 to improve this for all keys

runtime/interpreter/value.go
turbolent: As I noted above, the conversion is always performed so that the equality functions of the `sema.Type` hierarchy can be used, and the equality functions don't have to be reimplemented for the `interpreter.StaticType` hierarchy.

I'm not quite following the last paragraph: `convertStaticToSemaType` doesn't return a `TypeValue`, it returns a `sema.Type`.

PR #204, opened 2020-07-03
encoding/json: Fix parsing of unpadded fixed point numbers

### LINKED ISSUE #199: JSON-CDC decoder incorrectly handles unpadded Fix64 and UFix64 values
### Problem

As per the JSON-CDC spec, fixed point numbers do not need to be encoded with trailing zeros after the decimal point: https://github.com/onflow/flow/blob/master/docs/json-cadence-spec.md#fixed-point-numbers

However, the `encoding/json` package only correctly decodes JSON-CDC `Fix64` and `UFix64` values if they contain exactly 8 digits after the decimal point. 

For example, this value:

```json
{
    "type": "Fix64",
    "value": "12.3"
}
```

is incorrectly decoded as `0.00000123`.

Whereas this value:

```json
{
    "type": "Fix64",
    "value": "12.30000000"
}
```

is correctly decoded as `12.3`.

psiemens:
Closes #199 

## Description

This PR fixes a bug in the JSON-CDC decoder that was causing a `Fix64` or `UFix64` value to be decoded incorrectly if it did not contain a fractional that was exactly 8 digits in length.


turbolent:
@psiemens As discussed, refactored the fixed point checking and conversion code into a package. Also extended the checking to handle some more invalid cases

PR #206, opened 2020-07-03
Byte conversion functions for numbers and addresses

### LINKED ISSUE #197: Add value encoding to the standard library
### Issue To Be Solved

With the introduction of the `Crypto` built-in package, there is now a use case for converting arbitrary Cadence values into a serialized byte array, which can then be hashed and verified against a signature.

This encoding format should be secure and well-defined, allowing users to encode values in an external environment (i.e. Go, JavaScript) and expect Cadence to produce the same results.

This encoding format may only need to support Cadence value types (i.e. no resources).

### Possible Solution A

Expose an `Encoding.RLP` package as part of the Cadence standard library.

* Simple API (i.e. `RLP.Encode(value: AnyStruct)`)

### Possible Solution B

Add a `toBytes` method to value types that produces a consistent byte-wise encoding.

- E.g. `let myIntBytes = myInt.toBigEndianBytes()`

turbolent:
Closes #197 

- Add the function `fun toBytes(): [UInt8]` to `Address`. It always returns 8 bytes.
- Add the function `fun toBigEndianBytes(): [UInt8]` to all number types
- Expose the built-in number hierarchy to programs (`Number`, `SignedNumber`, `Integer`, `SignedInteger`, `FixedPoint`, `SignedFixedPoint`)


runtime/interpreter/value.go
```diff
@@ -1364,6 +1447,12 @@ func (Int16Value) SetMember(_ *Interpreter, _ LocationRange, _ string, _ Value)
 	panic(errors.NewUnreachableError())
 }
 
+func (v Int16Value) ToBigEndianBytes() []byte {
+	b := make([]byte, 2)
+	binary.BigEndian.PutUint16(b, uint16(v))
```
psiemens, line 1452: I just realized this, but if we encode arbitrary-length integers as their two's complement, do you think it makes sense to also do it for all integers?

runtime/interpreter/value.go
```diff
@@ -37,6 +37,56 @@ import (
 	"github.com/onflow/cadence/runtime/trampoline"
 )
 
+func SignedBigIntToBigEndianBytes(bigInt *big.Int) []byte {
```
psiemens, line 40: Maybe move these functions further down in the file. Also, do they need to be exported?

runtime/interpreter/value.go
turbolent: Yep, negative built-in integers are already represented in two's complement, so nothing to be done here. See also the tests: https://github.com/onflow/cadence/pull/206/files#diff-b17c665b13703eae54b6cea9fee6aa63R142-R144

runtime/interpreter/value.go
turbolent: I intentionally exported them because I thought we'd also export them from `cadence`? If it's not needed, I'll make them private.

Also, they're used by a bunch of value types, I think I'll just move them to a separate file 👍

runtime/interpreter/value.go
psiemens: Ah I see! Looks good 👍

runtime/interpreter/value.go
psiemens: Oh, that's a great point! In that case I agree that we should export them.

values_test.go
```diff
@@ -0,0 +1,198 @@
...
+		//	NewFix64("42.0"):  {0, 0, 0, 0, 250, 86, 234, 0},
+		//	NewFix64("42.24"): {0, 0, 0, 0, 251, 197, 32, 0},
+		//	NewFix64("-1.0"):  {255, 255, 255, 255, 250, 10, 31, 0},
+		//},
+		//// UFix*
+		//"UFix64": {
+		//	NewFix64("0.0"):   {0, 0, 0, 0, 0, 0, 0, 0},
+		//	NewFix64("42.0"):  {0, 0, 0, 0, 250, 86, 234, 0},
+		//	NewFix64("42.24"): {0, 0, 0, 0, 251, 197, 32, 0},
+		//},
```
psiemens, line 167: Are you saving these for a separate PR?

turbolent: No, this was a work in progress, the PR should be ready as of 9953a94

values.go
```diff
@@ -701,6 +701,19 @@ func NewFix64(s string) (Fix64, error) {
...
+	v, err := fixedpoint.NewFix64(
+		negative,
+		new(big.Int).SetInt64(int64(integer)),
+		new(big.Int).SetInt64(int64(fraction)),
+		fixedpoint.Fix64Scale,
+	)
+	if err != nil {
+		return 0, err
+	}
+	return Fix64(v.Int64()), nil
```
psiemens, line 714: Beautiful! 😍

Issue #210, opened 2020-07-05
JSON Serialization for AST


turbolent:
## Context

To be able to process the AST outside of Go, for example to visualize it in an AST explorer, or use it for pretty-printing, the JSON representation must be improved:

- Type information is necessary, e.g. to distinguish a boolean from a nil expression
- Position information must be eagerly included. The internal representation computes it sometimes from child elements

## Definition of Done

Implement and test `json.MarshalJSON` for the following types, adding a `Type` property where the element is polymorphic (e.g. expressions), and position information as `StartPos` and `EndPos` properties:

- [x] `Access`
- [x] `Argument`
- [x] `Block`
- [x] `FunctionBlock`
- [x] `Condition`
- [x] `CompositeDeclaration`
- [x] `FieldDeclaration`
- [x] `ConditionKind`
- [x] `BoolExpression`
- [x] `NilExpression`
- [x] `StringExpression`
- [x] `IntegerExpression`
- [x] `FixedPointExpression`
- [x] `ArrayExpression`
- [x] `DictionaryExpression`
- [x] `IdentifierExpression`
- [x] `InvocationExpression`
- [x] `MemberExpression`
- [x] `IndexExpression`
- [x] `ConditionalExpression`
- [x] `UnaryExpression`
- [x] `BinaryExpression`
- [x] `FunctionExpression`
- [x] `CastingExpression`
- [x] `CreateExpression`
- [x] `DestroyExpression`
- [x] `ReferenceExpression`
- [x] `ForceExpression`
- [x] `PathExpression`
- [x] `FunctionDeclaration`
- [x] `SpecialFunctionDeclaration`
- [x] `ImportDeclaration`
- [x] `Identifier`
- [x] `IdentifierLocation`
- [x] `StringLocation`
- [x] `AddressLocation`
- [x] `InterfaceDeclaration`
- [x] `Members`
- [x] `Operation`
- [x] `Parameter`
- [x] `ParameterList`
- [x] `Program`
- [x] `ReturnStatement`
- [x] `BreakStatement`
- [x] `ContinueStatement`
- [x] `IfStatement`
- [x] `WhileStatement`
- [x] `ForStatement`
- [x] `EmitStatement`
- [x] `AssignmentStatement`
- [x] `SwapStatement`
- [x] `ExpressionStatement`
- [ ] `TransactionDeclaration`
- [x] `Transfer`
- [x] `TransferOperation`
- [x] `TypeAnnotation`
- [x] `NominalType`
[...]

PR #211, opened 2020-07-05
Improve the JSON serialization for some AST elements

turbolent:
Work towards #210

Start improving the JSON serialization for some AST nodes:

- `Access`
- `Argument`
- `ConditionKind`
- `BoolExpression`
- `NilExpression`

### DIFF
--- a/runtime/ast/expression.go
+++ b/runtime/ast/expression.go
@@ -19,6 +19,7 @@
 package ast
 
 import (
+	"encoding/json"
 	"fmt"
 	"math/big"
 	"strings"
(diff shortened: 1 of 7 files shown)


runtime/ast/expression.go
```diff
@@ -66,10 +67,21 @@ func (e *BoolExpression) String() string {
 	return "false"
 }
 
+func (e *BoolExpression) MarshalJSON() ([]byte, error) {
+	type Alias BoolExpression
```
benjaminkvm, line 71: Curious why you define this type versus just use `*BoolExpression` below?

turbolent: The function tries to add some more properties to the resulting JSON. It could do that by declaring another struct type which has exactly all the same fields like the original, but that would cause a lot of duplication (and maintenance cost). 

Marshaling a wrapper struct which embeds the original won't work: the json.Marshal call would in turn call this very same function again for the embedded original, leading to a stack overflow.
This can be prevented by introducing a type alias, which has the same fields as the original, but the MarshallJSON function is not called on it.

For more, see e.g. http://choly.ca/post/go-json-marshalling/

PR #213, opened 2020-07-06
Change [Int] to [UInt8] where byte array semantics are wanted

### LINKED ISSUE #212: Inconsistent representation of byte arrays
### Problem

The newly added `toBytes` and `toBigEndianBytes` methods return byte arrays typed as `[UInt8]`, whereas the `Crypto` contract accepts public keys, signatures and messages typed as `[Int]`. To support compatibility between these APIs, the types should be consistent.

`[UInt8]` is more correct -- is there any reason not to update all existing byte array representations to be typed this way?

turbolent:
Closes #212


runtime/deferral_test.go
```diff
@@ -91,17 +93,7 @@ func TestRuntimeStorageDeferredResourceDictionaryValues(t *testing.T) {
...
-          transaction {
-
-              prepare(signer: AuthAccount) {
-                  signer.setCode(%s)
-              }
-          }
-        `,
-		ArrayValueFromBytes(contract).String(),
-	))
+	deploy := utils.DeploymentTransaction(contract)
```
psiemens, line 96: 👍

runtime/interpreter/value.go
```diff
@@ -407,6 +395,30 @@ func (v *StringValue) GetMember(_ *Interpreter, _ LocationRange, name string) Va
...
 
+// Length returns the number of characters (grapheme clusters)
+//
+func (v *StringValue) Length() int {
+	return uniseg.GraphemeClusterCount(v.Str)
+}
+
+// DecodeHex hex-decodes this string and returns an array of UInt8 values
+//
+func (v *StringValue) DecodeHex() *ArrayValue {
```
psiemens, line 406: 👍

PR #214, opened 2020-07-06
Don't report an error for a restriction with an invalid type

turbolent:
Just noticed that when a test case was failing

### DIFF
--- a/runtime/sema/checker.go
+++ b/runtime/sema/checker.go
@@ -899,12 +899,14 @@ func (checker *Checker) convertRestrictedType(t *ast.RestrictedType) Type {
 		if !ok || (restrictionCompositeKind != common.CompositeKindResource &&
 			restrictionCompositeKind != common.CompositeKindStructure) {
 
-			checker.report(
-				&InvalidRestrictionTypeError{
-					Type:  restrictionResult,
-					Range: ast.NewRangeFromPositioned(restriction),
-				},
-			)
+			if !restrictionResult.IsInvalidType() {
+				checker.report(
+					&InvalidRestrictionTypeError{
+						Type:  restrictionResult,
+						Range: ast.NewRangeFromPositioned(restriction),
+					},
+				)
+			}
 			continue
 		}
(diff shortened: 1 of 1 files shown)


runtime/sema/checker.go
```diff
@@ -899,12 +899,14 @@ func (checker *Checker) convertRestrictedType(t *ast.RestrictedType) Type {
 		if !ok || (restrictionCompositeKind != common.CompositeKindResource &&
 			restrictionCompositeKind != common.CompositeKindStructure) {
 
-			checker.report(
-				&InvalidRestrictionTypeError{
-					Type:  restrictionResult,
-					Range: ast.NewRangeFromPositioned(restriction),
-				},
-			)
+			if !restrictionResult.IsInvalidType() {
```
benjaminkvm, line 902: Shouldn't this be: `if !restrictionResult.IsInvalidType() {`, if the restriction result is invalid report the error?

turbolent: Assuming you meant `if restrictionResult.IsInvalidType() { ... }`, no, this is the correct logic:

The invalid type is a special type that is returned and propagated when something is incorrect in the program and a real type cannot be determined. For example, the type of an undeclared variable is the invalid type. 

If an invalid type is encountered somewhere, it basically indicates that somewhere else was an error and the current error is a result of the previous error. So we don't report an error if the type is invalid to reduce error fatigue (one error in the program might lead to a whole handful of more errors as a result of it).

benjaminkvm: Thanks for clarifying! Makes sense!

turbolent: Concrete example: If the program contains the restriction type `T{}`, where the restricted type `T` is undeclared, then the checker will report an error that `T` is undeclared and the type of `T` will be `*InvalidType` (`restrictionResult.IsInvalidType()` will be true). 

In that case we don't also want to report another error (`InvalidRestrictionTypeError`)

PR #223, opened 2020-07-09
Refactor activations

turbolent:
While investigating how we can detect when values "on the stack" are freed, I cleaned up the code for  activations by adding a wrapper type and adding a depth field

The implementation detail that an activation is backed by a `hamt.Map` was leaked outside, so I created a type to separate that concern a bit better. Maybe in the future we're changing how the activation is implemented, so this change should make that easier and smoother, because nothing has to change for consumers of this code.

Adding the type also allowed defining the higher-level functions (`FirstRest`, `Find`, `Insert`).
The fact the activation is using a `common.StringEntry` was also leaking out and was cleaned up.

### DIFF
--- a/runtime/interpreter/variable.go
+++ b/runtime/interpreter/variable.go
@@ -19,13 +19,13 @@
 package interpreter
 
 type Variable struct {
-	Value Value
-	Depth int
+	Value           Value
+	ActivationDepth int
 }
 
 func NewVariable(value Value, depth int) *Variable {
 	return &Variable{
-		Value: value,
-		Depth: depth,
+		Value:           value,
+		ActivationDepth: depth,
 	}
 }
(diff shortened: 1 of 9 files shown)


runtime/interpreter/variable.go
```diff
@@ -20,4 +20,12 @@ package interpreter
 
 type Variable struct {
 	Value Value
+	Depth int
+}
+
+func NewVariable(value Value, depth int) *Variable {
+	return &Variable{
+		Value: value,
+		Depth: depth,
```
benjaminkvm, line 29: Minor: Not sure if we would maybe want to call this `ActivationDepth` or `StackDepth` or something a bit more specific.

turbolent: Good idea! I'll rename to `ActivationDepth` 👍

turbolent: Refactored in d0bf87d, including for `sema.Variable`

PR #231, opened 2020-07-10
Prevent a potential stack overflow when checking member storability

turbolent:
Prevent a potential stack overflow when checking member storability by tracking results

### DIFF
--- a/runtime/tests/checker/storable_test.go
+++ b/runtime/tests/checker/storable_test.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/onflow/cadence/runtime/ast"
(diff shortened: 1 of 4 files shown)


runtime/tests/checker/storable_test.go
```diff
@@ -310,3 +311,28 @@ func TestCheckStorable(t *testing.T) {
 		}
 	}
 }
+
+func TestCheckStorableCycle(t *testing.T) {
+
+	t.Parallel()
+
+	_, err := ParseAndCheck(t, `
+      struct S {
```
benjaminkvm, line 320: Would this not be an invalid program (the other errors below) anyways? Is there a case where we have a valid program otherwise with a non-storable member error?

turbolent: Yes, it's definitely invalid, but this lead to a checker crash (stack overflow) before.
Or maybe I'm not quite understanding the question, sorry! 🤔

PR #234, opened 2020-07-14
Add optional high-level write callback

### LINKED ISSUE #218: Add high-level write callback
## Context

Some embedders of Cadence, e.g. the Playground, need to get the high-level representation (i.e. `cadence.Value`, not low-level byte-encoded representation).

Add an additional callback that gets called in addition to the low-level write callback (`SetValue`).

## Definition of Done

- [ ] Add support for an optional function that gets called when all values are written. Pass a `cadence.Value` instead of a `[]byte`.

cc @psiemens

turbolent:
Closes #218

- Add support for exporting contracts
- Add an optional runtime interface which is optionally called when top-level writes occur. The parameters are high-level types (`common.Address` instead of `[]byte` for the address, `string` instead of `[]byte` for the key, `cadence.Value` instead of `[]byte` for the value).

  The main use-case at the moment is observing account storage and rendering it to the user in the Playground.

### DIFF
--- a/runtime/interface.go
+++ b/runtime/interface.go
@@ -23,7 +23,6 @@ import (
 
 	"github.com/onflow/cadence"
 	"github.com/onflow/cadence/runtime/ast"
-	"github.com/onflow/cadence/runtime/common"
 )
 
 const BlockHashLength = 32
@@ -91,7 +90,7 @@ type HighLevelStorage interface {
 	HighLevelStorageEnabled() bool
 
 	// SetCadenceValue sets a value for the given key in the storage, owned by the given account.
-	SetCadenceValue(owner common.Address, key string, value cadence.Value) (err error)
+	SetCadenceValue(owner Address, key string, value cadence.Value) (err error)
 }
 
 type Metrics interface {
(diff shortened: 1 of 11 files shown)


runtime/interface.go
```diff
@@ -81,6 +82,18 @@ type Interface interface {
...
+type HighLevelStorage interface {
+	Interface
+
+	// HighLevelStorageEnabled should return true
+	// if the functions of HighLevelStorage should be called,
+	// e.g. SetCadenceValue
+	HighLevelStorageEnabled() bool
+
+	// SetCadenceValue sets a value for the given key in the storage, owned by the given account.
+	SetCadenceValue(owner common.Address, key string, value cadence.Value) (err error)
```
psiemens, line 94:
```suggestion
	SetCadenceValue(owner Address, key string, value cadence.Value) (err error)
```

`runtime.Address` is just an alias for `common.Address` that allows the caller to avoid depending on `common`, which is kind of internal to the runtime. The other interface methods above use `runtime.Address` as well.

turbolent: Great idea! I'm so used to using `common.Address` :-) 
Changed in 15ee18fc948a26e7d38408646a5ff84727399ead

turbolent:
@psiemens Thank you for already testing this! 👍

Issue #237, opened 2020-07-15
Update documentation


turbolent:
## Context

New features have been added to Cadence in the last few releases.
Update the documentation and describe the new features

## Definition of Done

- [x] Querying block information
  - Get the current block:
    `fun getCurrentBlock(): Block`

  - Get a block at a given height:
    `fun getBlock(at height: UInt64): Block?`

  - The block info is:

    ```
    struct Block {
        let id: [UInt8; 32]
        let height: UInt64
        let timestamp: UFix64
    }
    ```
  - [ ] Also update https://docs.onflow.org/docs/faq#how-do-i-access-block-information-like-timestamp-and-block-number-in-cadence
- [x] Crypto contract
- [x] Typed capabilities
- [x] Run-time Types
- [x] Addresses size (8 instead of 20 bytes)
- [x] Parameters for transactions and scripts
- [x] `fun toString(): String` was added for all number types and addresses
- [x] `fun toBytes(): [UInt8]` was added to `Address`
- [x] `fun toBigEndianBytes(): [UInt8]` was added to all number types.
- [x] Fields which have a non-storable type are now rejected:
  Non-storable types are:
    - Functions (e.g. `((Int): Bool)`)
    - Accounts (`AuthAccount` / `PublicAccount`)
    - Transactions
    - `Void`
- [x] `String.decodeHex(): [Int]` → `String.decodeHex(): [UInt8]`
- [x] `AuthAccount.addPublicKey(publicKey: [Int])` → `AuthAccount.addPublicKey(publicKey: [UInt8])`
- [x] `AuthAccount.setCode(code: [Int])` → `AuthAccount.setCode(code: [UInt8])`
- [x] Conformance check now allows subtype for return type
- [x] Type hierarchy
- [x] `unsafeRandom`
- [x] Account creation

timmymmit:
Add section for Roadmap and Features that are upcoming

PR #245, opened 2020-07-19
Improve conformance check

### LINKED ISSUE #241: Improve conformance check: Allow function subtypes
## Context

Currently conformance requires function implementations to be *equal*, including argument labels.
However, the parameters could be supertypes, and the return type could be a subtype:

https://github.com/onflow/cadence/blob/90cf100822e73355e95499693adeee6ee373d6c7/runtime/sema/check_composite_declaration.go#L885

For example, the following program should be valid, but is not at the moment:

```swift
pub resource interface RI {}

pub resource R: RI {}

pub struct interface SI {
    pub fun get(): @{RI}
}

pub struct S: SI {
    pub fun get(): @R {
        return <- create R()
    }
}
```

## Definition of Done

- [x] Sema: Improve conformance check for functions
- [x] Tests

turbolent:
Closes #241 

Make return types covariant, i.e. allow subtypes for return types of functions to satisfy interface conformance.


runtime/tests/checker/conformance_test.go
```diff
@@ -198,38 +261,156 @@ func TestCheckTypeRequirementConformance(t *testing.T) {
...
+              fun get(): @R {
+                  return <- create R()
+              }
+          }
+        `)
+
+		require.NoError(t, err)
+	})
+
+	t.Run("invalid, return type is supertype", func(t *testing.T) {
```
benjaminkvm, line 296: 👍

benjaminkvm:
If this isn't in the Cadence documentation, would be really great to add it btw. One of those nuanced things that would be great to have written down.

turbolent:
Yeah, we should maybe document it. 

When I implemented this, I checked several other languages, Swift, Rust, Scala, Kotlin, Java, C#, etc. and their behaviour is all different (some treat parameters and return type invariant, some allow covariant return types, etc.) – and it wasn't ever documented anywhere, i.e. I had to try it out.

turbolent:
@benjaminkvm Added this to #237

PR #246, opened 2020-07-20
Parse docstrings for functions

turbolent:
Work towards #247

Add support for docstrings to functions:

- Block comments, which start with `/**`
- Line comments, which start with `///`

### DIFF
--- a/runtime/parser2/declaration.go
+++ b/runtime/parser2/declaration.go
@@ -63,9 +63,7 @@ func parseDeclaration(p *parser, docString string) ast.Declaration {
 	var accessPos *ast.Position
 
 	for {
-		p.parseTrivia(triviaOptions{
-			skipNewlines: true,
-		})
+		p.skipSpaceAndComments(true)
 
 		switch p.current.Type {
 		case lexer.TokenPragma:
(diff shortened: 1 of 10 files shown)


runtime/parser2/declaration.go
```diff
@@ -121,14 +126,18 @@ func parseAccess(p *parser) ast.Access {
 	case keywordPub:
 		// Skip the `pub` keyword
 		p.next()
-		p.skipSpaceAndComments(true)
+		p.parseTrivia(triviaOptions{
```
zhangchiqing, line 129: When I look at this change, it makes sense. Because I have the context, I know you wanted to replace with `parseTrivia` and what it does it to skip space and comments. 

Since it doesn't have the `parseDocStrings: true` option, it's actually doing `skipSpaceAndComments`. The problem with the function name `parseTrivia` is that if I have never seen this PR, I would question what is `Trivia`, why parsing it?

I would suggest to refactor the `skipSpaceAndComments` and use `parseTrivia` to implement `skipSpaceAndComments` if possible, so that we don't need to introduce this change here.

turbolent: I renamed the function, because it was quite verbose, especially when I added the options struct, e.g. code looked like `skipSpaceAndComments(skipSpaceAndCommentsOptions{...`. 

Trivia is the best name I found, as it's also the term used in the Swift AST. I can revert the renaming.

How do you propose to refactor the function?

runtime/parser2/parser.go
```diff
@@ -244,8 +255,22 @@ func (p *parser) replayBuffered() {
 	p.next()
 }
 
-func (p *parser) skipSpaceAndComments(skipNewlines bool) (containsNewline bool) {
```
zhangchiqing, line 247: >How do you propose to refactor the function?

Instead of removing it, we could refactor this function with `parseTrivia`:

```golang
func (p *parser) skipSpaceAndComments(skipNewlines bool) (containsNewline bool) {
  containsNewline, _ := p.parseTrivia(triviaOptions{
    skipNewlines: skipNewlines,
  })
  return containsNewline
}
```

So that we don't need to change `skipSpaceAndComments` everywhere.

turbolent: Great idea 👍

turbolent: Added back in 3360414307707b32fb31232dba804023af9ec56f

PR #250, opened 2020-07-20
Improve equatability check

turbolent:
Refactor the function which determines if a type is equatable.


turbolent:
Thank you for the quick review @benjaminkvm!

PR #253, opened 2020-07-22
Non-composite reference subtyping

### LINKED ISSUE #178: Update subtyping rules for non-composite references
### Problem

Initially we only allowed references to resources, but extended this also to non-resource types. However, the subtyping rules for references only consider composite types. The subtyping rules for non-composite references need to be updated to not only consider composite types, e.g. `&Int <: &AnyStruct`.

### Steps to Reproduce

```swift
let x = 1
let xRef = &x as &Int
let anyRef: &AnyStruct = xRef
```

### Acceptance Criteria
- [x] Subtyping for non-composite types should work correctly

turbolent:
Closes #178

Fix/extend/simplify the subtyping rules:

- `Any` is *not* a subtype of `AnyStruct`
- A reference is a subtype to a reference to `AnyStruct`, if the referenced types are subtypes
  (`&T <: &AnyStruct` iff `T <: AnyStruct`)

Also, extend and parallelize some related tests

### DIFF
--- a/runtime/tests/checker/reference_test.go
+++ b/runtime/tests/checker/reference_test.go
@@ -35,6 +35,8 @@ func TestCheckReferenceTypeOuter(t *testing.T) {
 
 	t.Run("resource", func(t *testing.T) {
 
+		t.Parallel()
+
 		_, err := ParseAndCheck(t, `
           resource R {}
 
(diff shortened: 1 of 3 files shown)


runtime/tests/checker/reference_test.go
```diff
@@ -213,12 +286,59 @@ func TestCheckReferenceExpressionWithInterfaceResultType(t *testing.T) {
...
+		require.NoError(t, err)
+	})
+
+	t.Run("non-composite", func(t *testing.T) {
+
+		t.Parallel()
+
+		_, err := ParseAndCheck(t, `
+          let i = 1
+          let ref = &i as &AnyStruct
```
benjaminkvm, line 327: Hmm, it seems a bit odd to me we can cast a reference to an `Int` to reference to `AnyStruct`

turbolent: `Int` is a subtype of `AnySruct`. I'll try to document this better through #237

PR #254, opened 2020-07-23
Record member access position

turbolent:
Improve parsing of member expressions: 
- Record the position where the access occurs, e.g. the `.` in `foo .bar` or `baz?.`
- Forbid whitespace after the access (parse but report)

### DIFF
--- a/runtime/ast/expression.go
+++ b/runtime/ast/expression.go
@@ -377,6 +377,7 @@ type AccessExpression interface {
 type MemberExpression struct {
 	Expression Expression
 	Optional   bool
+	AccessPos  Position
 	Identifier Identifier
 }
 
@@ -416,7 +417,11 @@ func (e *MemberExpression) StartPosition() Position {
 }
 
 func (e *MemberExpression) EndPosition() Position {
-	return e.Identifier.EndPosition()
+	if e.Identifier.Identifier == "" {
+		return e.AccessPos
+	} else {
+		return e.Identifier.EndPosition()
+	}
 }
 
 // IndexExpression
--- a/runtime/ast/expression.go
+++ b/runtime/ast/expression.go
@@ -377,6 +377,8 @@ type AccessExpression interface {
 type MemberExpression struct {
 	Expression Expression
 	Optional   bool
+	// The position of the token (`.`, `?.`) that separates the accessed expression
+	// and the identifier of the member
 	AccessPos  Position
 	Identifier Identifier
 }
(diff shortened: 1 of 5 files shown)


runtime/parser2/expression.go
```diff
@@ -985,22 +985,35 @@ func defineMemberExpression() {
...
+			return parseMemberAccess(p, token, left, true)
 		},
 	)
 }
 
-func parseMemberAccess(p *parser, left ast.Expression, optional bool) ast.Expression {
+func parseMemberAccess(p *parser, token lexer.Token, left ast.Expression, optional bool) ast.Expression {
 
-	p.skipSpaceAndComments(true)
+	// Whitespace after the '.' (dot token) is not allowed.
```
benjaminkvm, line 1003: Wpuld be good to make a note of this in the docs 👍

runtime/parser2/expression.go
```diff
@@ -985,22 +985,35 @@ func defineMemberExpression() {
...
+	// We parse it anyways and report an error
+
+	if p.current.Is(lexer.TokenSpace) {
+		errorPos := p.current.StartPos
+		p.skipSpaceAndComments(true)
```
benjaminkvm, line 1008: We use `parseTrivia` now for this no?

runtime/parser2/expression.go
```diff
@@ -1094,10 +1124,15 @@ func applyExprMetaLeftDenotation(
 		return left, true
 	}
 
+	skipWhitespace := exprLeftDenotationAllowsWhitespaceAfterToken(p.current.Type)
```
benjaminkvm, line 1127: Would this not be better named `permitWhitespace` or `allowWhitespace`?

benjaminkvm: Actually reading through I follow now! Nvm

runtime/ast/expression.go
```diff
@@ -377,6 +377,7 @@ type AccessExpression interface {
 type MemberExpression struct {
 	Expression Expression
 	Optional   bool
+	AccessPos  Position
```
benjaminkvm, line 380: Minor: Might be good to describe with a comment what this field records!

runtime/parser2/expression.go
turbolent: `skipSpaceAndComments` is just a helper that calls `parseTrivia`, which @zhangchiqing suggested, so all the code can stay as-is and it's shorter than `parseTrivia(parseTriviaOptions{...`

runtime/parser2/expression.go
turbolent: improved in 78b79c8409a6c9545ac0889d8a538653a9cdb05d

runtime/ast/expression.go
turbolent: Added a comment in 76a15454cf02cac6fb9ee3ef3f766cff98d4fc77

PR #256, opened 2020-07-23
Refactor members

turbolent:
Refactor how members are represented: Instead of only allowing to query for a single, known member by name, return all members, but do so in a lazy manner – only return names and resolvers, functions that produce the actual member.

In addition, provide docstrings for all built-in members. (cc @joshuahannan @10thfloor)

Also, record the positions of all member accesses and store the accessed type for each. This enables querying accessed types by positions, just like occurrences

### DIFF
--- a/runtime/sema/occurrences.go
+++ b/runtime/sema/occurrences.go
@@ -78,7 +78,7 @@ func NewOccurrences() *Occurrences {
 	}
 }
 
-func ToPosition(position ast.Position) Position {
+func ASTToSemaPosition(position ast.Position) Position {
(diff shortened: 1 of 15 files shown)


runtime/sema/occurrences.go
```diff
@@ -78,7 +78,7 @@ func NewOccurrences() *Occurrences {
 	}
 }
 
-func ToPosition(position ast.Position) Position {
+func ASTToSemaPosition(position ast.Position) Position {
```
benjaminkvm, line 81: 👍

PR #260, opened 2020-07-23
Export and encode Link and StorageReference values

### LINKED ISSUE #259: Link values are not exportable from the runtime
### Problem

When attempting to emit a storage write of type `interpreter.LinkValue`, the runtime panics because it fails to convert the value to a public `cadence.Value` type.

```
cannot convert value of type interpreter.LinkValue
```

This originates from this function, which does not have a switch case for `interpreter.LinkValue`: https://github.com/onflow/cadence/blob/83952e3867c91ab9fdad2c4c71303b8353970aa9/runtime/convertValues.go#L49-L111

### Steps to Reproduce

1. Execution a transaction with a `runtime.Interface` that implements `runtime.HighLevelStorage`
1. In this transaction, attempt to create link

```
self.account.link<&{NFTReceiver}>(/public/NFTReceiver, target: /storage/NFTCollection)
```
### Acceptance Criteria

The above link should ideally be emitted as type `cadence.Link`. Further more, a `cadence.Link` value should be encodable as JSON-CDC.

Alternatively, we could treat links as un-exportable and ignore them in the storage write handler.

### Context

This is preventing an upgrade of the Playground API, which currently breaks if a user saves a link.

psiemens:
Closes #259 

## Description

This PR introduces the following changes:
- Runtime can now export values of type `LinkValue` and `StorageReferenceValue`
- Added `cadence.Value` definitions for `cadence.Link` and `cadence.StorageReference`
- Add `JSON-CDC` encoding support for `cadence.Link` and `cadence.StorageReference`

### DIFF
--- a/encoding/json/encode.go
+++ b/encoding/json/encode.go
@@ -118,37 +118,43 @@ type jsonCompositeField struct {
 	Value jsonValue `json:"value"`
 }
 
+type jsonLinkValue struct {
+	Target     string `json:"target"`
+	BorrowType string `json:"borrowType"`
+}
+
 const (
-	voidTypeStr       = "Void"
-	optionalTypeStr   = "Optional"
-	boolTypeStr       = "Bool"
-	stringTypeStr     = "String"
-	addressTypeStr    = "Address"
-	intTypeStr        = "Int"
-	int8TypeStr       = "Int8"
-	int16TypeStr      = "Int16"
-	int32TypeStr      = "Int32"
-	int64TypeStr      = "Int64"
-	int128TypeStr     = "Int128"
-	int256TypeStr     = "Int256"
-	uintTypeStr       = "UInt"
-	uint8TypeStr      = "UInt8"
-	uint16TypeStr     = "UInt16"
-	uint32TypeStr     = "UInt32"
-	uint64TypeStr     = "UInt64"
-	uint128TypeStr    = "UInt128"
-	uint256TypeStr    = "UInt256"
-	word8TypeStr      = "Word8"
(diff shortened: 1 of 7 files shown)


encoding/json/encoding_test.go
```diff
@@ -1114,25 +1140,35 @@ func convertValueFromScript(t *testing.T, script string) cadence.Value {
...
+func testAllEncodeAndDecode(t *testing.T, tests ...encodeTest) {
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			testEncode(t, test.val, test.expected)
+			testEncodeAndDecode(t, test.val, test.expected)
 		})
 	}
 }
 
-func testEncode(t *testing.T, val cadence.Value, expectedJSON string) {
```
psiemens, line 1125: Split this function into separate encode and decode steps, renamed to `testEncodeAndDecode`. Links and references are only tested with `testEncode`.

encoding/json/encode.go
```diff
@@ -215,8 +228,12 @@ func (e *Encoder) prepare(v cadence.Value) jsonValue {
 		return e.prepareResource(x)
 	case cadence.Event:
 		return e.prepareEvent(x)
+	case cadence.StorageReference:
+		return e.prepareStorageReference(x)
+	case cadence.Link:
+		return e.prepareLink(x)
 	default:
-		return fmt.Errorf("unsupported value: %T, %v", v, v)
+		panic(fmt.Errorf("unsupported value: %T, %v", v, v))
```
turbolent, line 236: 👍

PR #261, opened 2020-07-24
LSP: Prevent user from using the service account

psiemens:
This PR updates the language server to no longer list the service account as the default active account.

Previously the language server was being initialized with `s.activeAccount = serviceAddress`, which allowed the VS Code extension to deploy as the service account if the user did not explicitly choose a different account from the dropdown. If the user started the emulator, opened a contract and clicked "deploy..." it would deploy to the service account.

This PR initializes the server with an empty active account, and prevents the "submit transaction...", "deploy contract..." code lenses to appear if no active account is set. Once the first user account is created, the active account will be set and the user can begin to submit transactions.


languageserver/integration/initialization.go
```diff
@@ -33,11 +33,9 @@ func (i *FlowIntegration) initialize(initializationOptions interface{}) error {
 	}
 	i.config = conf
 
-	serviceAddress := flow.ServiceAddress(flow.Emulator)
-
 	// add the service account as a usable account
-	i.accounts[serviceAddress] = conf.ServiceAccountKey
-	i.activeAccount = serviceAddress
```
psiemens, line 40: No longer set service account as default active account

PR #267, opened 2020-07-25
Handle parsing ambiguity of return types

### LINKED ISSUE #258: Handle ambiguous parsing of return type / restricted type
(Reported by @MaxStalker in https://forum.onflow.org/t/breaking-changes-in-cadence-parser/485)

### Problem

The following code parses and executes successfully: 

```swift
fun main():Int { return 2 }
```

However, removing the whitespace between the return type and the opening curly brace results in a parsing error:

```swift
pub fun main():Int{ return 2 }
```

```
Parsing failed: unexpected token: got decimal integer, expected ',' or '}'
```

### Acceptance Criteria

The second example should parse like the first.

turbolent:
Closes #258

Handle the parsing ambiguity when parsing a return type which involves a open curly brace

cc @MaxStalker

### DIFF
--- a/runtime/parser2/type.go
+++ b/runtime/parser2/type.go
@@ -456,10 +456,7 @@ func defineRestrictedOrDictionaryType() {
 	// This handles the ambiguous case where a function return type's open brace
 	// may either be a restricted type (if there is no whitespace)
 	// or the start of the function body (if there is whitespace).
-	//
-	// In the fu
 
-	//setTypeLeftBindingPower(lexer.TokenBraceOpen, typeLeftBindingPowerRestriction)
 	setTypeMetaLeftDenotation(
 		lexer.TokenBraceOpen,
 		func(p *parser, rightBindingPower int, left ast.Type) (result ast.Type, done bool) {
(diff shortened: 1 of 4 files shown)


runtime/parser2/type.go
```diff
@@ -430,26 +450,62 @@ func defineRestrictedOrDictionaryType() {
...
+	// This handles the ambiguous case where a function return type's open brace
+	// may either be a restricted type (if there is no whitespace)
+	// or the start of the function body (if there is whitespace).
+	//
+	// In the fu
 
-	setTypeLeftBindingPower(lexer.TokenBraceOpen, typeLeftBindingPowerRestriction)
-	setTypeLeftDenotation(
+	//setTypeLeftBindingPower(lexer.TokenBraceOpen, typeLeftBindingPowerRestriction)
+	setTypeMetaLeftDenotation(
```
psiemens, line 463: Looks like some leftover comments on these lines

turbolent: Good catch, removed in 21fa7ae!

PR #273, opened 2020-07-28
Handle recursive types when exporting, remove XDR encoding

turbolent:
- Properly handle recursive types when exporting 
- Remove the XDR encoding. It is not used and outdated. Improvements to the export would require maintaining it


runtime/convertTypes_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/stretchr/testify/assert"
+
+	"github.com/onflow/cadence"
+	"github.com/onflow/cadence/runtime/ast"
+	"github.com/onflow/cadence/runtime/common"
+	"github.com/onflow/cadence/runtime/sema"
+	"github.com/onflow/cadence/runtime/tests/utils"
+)
+
+func TestExportRecursiveType(t *testing.T) {
```
psiemens, line 33: 💯

runtime/convertTypes.go
```diff
@@ -169,49 +187,65 @@ func exportCompositeType(t *sema.CompositeType) cadence.CompositeType {
...
 			TypeID:     id,
 			Identifier: t.Identifier,
 			Fields:     fields,
 		}
+
+	default:
+		panic(fmt.Sprintf("cannot export composite type %v of unknown kind %v", t, t.Kind))
+	}
+
+	// NOTE: ensure to set the result before recursively export field types
```
psiemens, line 228: 👍

PR #274, opened 2020-07-28
Add RunWithStdio function, move main.go to cmd/languageserver

psiemens:
This PR adds a public `RunWithStdio` that can be used to start the language server from another Go process.

I also deleted the `languageserver/languageserver` binary. Was this needed?


languageserver/languageserver.go
```diff
@@ -1,37 +1,16 @@
-// +build !wasm
```
turbolent, line 1: The license header shouldn't be removed here

PR #277, opened 2020-07-30
Update documentation

### LINKED ISSUE #237: Update documentation

turbolent:
Closes #237

- Document:
  - Run-time types
  - Crypto contract
  - Type hierarchy
  - New address format
  - `toString` function for addresses and numbers
  - `Address.toBytes()` and `Number.toBigEndianBytes()` functions
  - `String.decodeHex()` function
  - Storability requirement for field types
  - Implementation status for field getters and setters
  - `unsafeRandom` function
  - Transaction parameters
  - Account creation
- Update the Crypto contract to use docstrings


docs/language.md
```diff
@@ -6001,6 +6195,21 @@ let counterRef = countCap.borrow()
 let counterRef2 = publicAccount.borrow<&Counter>(/storage/counter)
 ```
 
+### Account Creation
+
+Accounts can be created by calling the `AuthAccount` constructor
+and passing the account that should pay for the account creation for the `payer` parameter.
+
```
joshuahannan, line 6202:
```suggestion
The `payer` must have Flow tokens in their account to be able to create an account. The Flow tokens are automatically withdrawn from their account's token vault.
```

docs/language.md
```diff
@@ -6001,6 +6195,21 @@ let counterRef = countCap.borrow()
...
+For example, to create an account and have the signer of the transaction pay for the account creation:
+
+```cadence,file=account-creation.cdc
+transaction {
+    prepare(signer: AuthAccount) {
+        let account = AuthAccount(payer: signer)
```
joshuahannan, line 6208: Should we include adding keys to an account so it shows the usual way of creating an AuthAccount? or do we want to keep it simple?

docs/language.md
```diff
@@ -6444,26 +6657,46 @@ Although optional, each block serves a specific purpose when executing a transac
...
 
-## Prepare
+### Transaction Parameters
+
+Transactions may declare parameters.
+Transaction parameters are declared like function parameters.
+The arguments for the transaction are passed in the sent transaction.
 
-The `prepare` **block** is used when access to the private `AuthAccount` object of **signing accounts** is required for your transaction.
+Transaction parameters are accessible in all blocks.
```
joshuahannan, line 6666: Sometimes we use the term phase, and sometimes we use the term block when describing the parts of a transaction. I would prefer the term phase, but we at least should be consistent. What do you think, @turbolent ?
```suggestion
Transaction parameters are accessible in all the phases of a transaction.
```

docs/language.md
```diff
@@ -6604,37 +6852,128 @@ TODO
...
+
+```cadence,file=runtime-type-comparison.cdc
+
+Type<Int>() == Type<Int>()
+
+Type<Int>() != Type<String>()
 ```
 
-### `assert`
+The function `fun isInstance(_ type: Type): Bool` can be used to check if a value has a certain type.
```
joshuahannan, line 6887: Can you add more detail about how much `isInstance` checks the type? Will it know that an `@AnyResource` is a `@FlowToken.Vault` or if a `@FlowToken.Vault` is an `@AnyResource`?

docs/language.md
```diff
@@ -6604,37 +6852,128 @@ TODO
...
-fun assert(_ condition: Bool, message: String)
+let collectible <- create Collectible()
+let collectibleType = Type<@Collectible>()
+let result = collectible.isInstance(collectibleType)
+```
+
+For example, this allows implementing a marketplace sale resource:
+
+```cadence,file=runtime-type-simple-sale.cdc
+pub resource SimpleSale {
```
joshuahannan, line 6898: It would probably be good to have some code comments in this file to make sure it is clear

docs/language.md
turbolent: Agreed. It was rewritten to use blocks in most places, but yeah, let's use phase everywhere

docs/language.md
turbolent: Yep, adding that, good point

docs/language.md
turbolent: Good point, added a less-specific variant of this

docs/language.md
turbolent: I checked the document again, and it is actually using block elusively and there is a single explanation using the term phase:

> Then, three optional main phases: Preparation, execution, and postconditions, only in that order.
> Each phase is a block of code that executes sequentially.

So this is already consistent and I'm not sure replacing "block" with "phase" everywhere changes much

docs/language.md
turbolent: Good point, commented it in c69ff0d

docs/language.md
turbolent: Documented that in f3dbda6

docs/language.md
turbolent: Good questions! Tried to improve it in 0b09123 and 13f6a52

turbolent:
@joshuahannan Thanks for the review! Addresses your feedback and questions, could you please have another look at the updated parts? Thank you!

docs/language.md
joshuahannan: I just feel like phase is a more accurate term, because it illustrates that they are groupings of actions that happens in an order. Each phase contains a block of code though, but block doesn't indicate an order, so that is why I like phase more. Does that make sense? It isn't that big of a deal though. If you want to keep it as block, I understand

turbolent: Oh, I hadn't thought of it that way, that's a good argument! I'll replace

turbolent: Improved it in 0c7b5f8

PR #278, opened 2020-07-30
Improve indicies/keys when decoding

### LINKED ISSUE #216: Define constants for CBOR keys
### Context

The key indices used in the CBOR encoder and decoder can easily get out of sync.
Introduce constants that are used on both sides.

### Definition of Done

- [x] Define constants for the keys in the CBOR encoder, e.g.:

  ```
  const (
    EncodedTypeValueType uint64 = 0
  )

  type encodedTypeValue struct {
    Type interface{} `cbor:"0,keyasint"`
  }
  ```

- [x] Use the key constants in the CBOR decoder, e.g.:

  ```
  decodedStaticType, err := d.decodeStaticType(encoded[EncodedTypeValueType])
  ```

turbolent:
Closes #216 

Instead of using magic values for indexing when decoding, define values that are derived from the struct tags, which are the source of truth and used when encoding.

I still don't like how the fields are strings, i.e. it is not checked at compile time, but at least a mismatch will fail at run-time right at startup.

### DIFF
--- a/runtime/interpreter/encode.go
+++ b/runtime/interpreter/encode.go
@@ -214,7 +214,7 @@ func init() {
 	}
 }
 
-func fieldKey(t interface{}, name string) uint64 {
+func cborFieldKey(t interface{}, name string) uint64 {
 	rt := reflect.TypeOf(t)
 
 	field, ok := rt.FieldByName(name)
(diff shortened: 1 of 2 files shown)


runtime/interpreter/encode.go
```diff
@@ -214,6 +214,28 @@ func init() {
 	}
 }
 
+func cborFieldKey(t interface{}, name string) uint64 {
```
benjaminkvm, line 217: Clever!

PR #286, opened 2020-08-04
Add JSON serialization support for more AST elements

turbolent:
Work towards #210

Add JSON marshalling for:

- Block
- FunctionBlock
- Condition
- StringExpression
- IntegerExpression
- FixedPointExpression
- ArrayExpression
- DictionaryExpression
- IdentifierExpression
- PathExpression
- TransferOperation
- VariableKind
- DeclarationKind
- CompositeKind

### DIFF
--- a/runtime/ast/block.go
+++ b/runtime/ast/block.go
@@ -18,6 +18,10 @@
 
 package ast
 
+import (
+	"encoding/json"
+)
+
 type Block struct {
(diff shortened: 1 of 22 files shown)


turbolent:
Yeah, `ast.Range` is embedded for convenience everywhere, refactoring this would be a huge change, so leaving it as is

PR #288, opened 2020-08-04
Recursively check if assignment target expression is valid, don't copy returned values

turbolent:
Instead of just checking if the outer-most expression which is assigned to is assignable, recursively check if it is assignable. Assignable expressions are only identifiers, index expressions, and member access expressions.

This then allows removing the defensive copying of the return statement value. As values are still copied in variable declarations and when passed as arguments, there is no way to return an object and have the caller modify it.

### DIFF
--- a/runtime/sema/check_assignment.go
+++ b/runtime/sema/check_assignment.go
@@ -157,7 +157,7 @@ func (checker *Checker) visitAssignmentValueType(
 	// Check the target is valid (e.g. identifier expression,
 	// indexing expression, or member access expression)
 
-	if _, ok := targetExpression.(ast.TargetExpression); !ok {
+	if !IsValidAssignmentTargetExpression(targetExpression) {
 		checker.report(
 			&InvalidAssignmentTargetError{
 				Range: ast.NewRangeFromPositioned(targetExpression),
@@ -369,4 +369,20 @@ func (checker *Checker) visitMemberExpressionAssignment(
 	}
 
 	return member.TypeAnnotation.Type
+}
+
+func IsValidAssignmentTargetExpression(expression ast.Expression) bool {
+	switch expression := expression.(type) {
+	case *ast.IdentifierExpression:
+		return true
+
+	case *ast.IndexExpression:
+		return IsValidAssignmentTargetExpression(expression.TargetExpression)
+
+	case *ast.MemberExpression:
+		return IsValidAssignmentTargetExpression(expression.Expression)
+
+	default:
+		return false
+	}
 }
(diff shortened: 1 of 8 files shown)


runtime/sema/errors.go
```diff
@@ -2007,7 +2007,7 @@ type InvalidAssignmentTargetError struct {
 }
 
 func (e *InvalidAssignmentTargetError) Error() string {
-	return "cannot assign to expression"
+	return "cannot assign to unsassignable expression"
```
zhangchiqing, line 2010:
```suggestion
	return "cannot assign to unassignable expression"
```

runtime/sema/check_assignment.go
```diff
@@ -370,3 +370,19 @@ func (checker *Checker) visitMemberExpressionAssignment(
...
+func IsValidAssignmentTargetExpression(expression ast.Expression) bool {
+	switch expression := expression.(type) {
+	case *ast.IdentifierExpression:
+		return true
+
+	case *ast.IndexExpression:
+		return IsValidAssignmentTargetExpression(expression.TargetExpression)
+
+	case *ast.MemberExpression:
+		return IsValidAssignmentTargetExpression(expression.Expression)
```
zhangchiqing, line 383: what's the difference between `expression.TargetExpression` and `expression.Expression`?

runtime/tests/checker/assignment_test.go
```diff
@@ -124,15 +124,136 @@ func TestCheckInvalidAssignmentTargetExpression(t *testing.T) {
...
+          fun test() {
+              {"a": 1}["a"] = 2
+          }
+        `)
+
+		errs := ExpectCheckerErrors(t, err, 1)
 
-	assert.IsType(t, &sema.InvalidAssignmentTargetError{}, errs[0])
+		assert.IsType(t, &sema.InvalidAssignmentTargetError{}, errs[0])
+	})
```
zhangchiqing, line 258: Test cases look great. Do we need a test case to verify that if a function returns an internal value, then changes to the returned value won't change the internal value?

runtime/sema/check_assignment.go
turbolent: Not sure I understand the question. `IndexExpression.TargetExpression` is the expression that is indexed into, `MemberExpression.Expression` is the expression that is accessed

runtime/tests/checker/assignment_test.go
turbolent: Good idea, I'll add some

runtime/sema/check_assignment.go
zhangchiqing: nvm, I misunderstood the logic. Looks good to me.

runtime/tests/checker/assignment_test.go
turbolent: Added a test in 69cb18b

zhangchiqing: Looks good to me 👍

PR #291, opened 2020-08-05
Ensure script return type is storable type

### LINKED ISSUE #186: Reject scripts with non-storable return type
### Context

In #169 / #180 we introduced a way to determine if a type can be stored.

The return type of a script must be storable / serializable, so reject functions with a non-storable return type.

### Definition of Done

- [ ] Runtime: Reject scripts with a non-storable return type

benjaminkvm:
Closes #186.

## Description

Fail to execute script if return type is non-storable type. We currently do not have a checker option to determine whether a program is a script or transaction, therefore I did this check within `ExecuteScript`. Let me know if you would prefer a checker option denoting program type + a checker check.

### DIFF
--- a/runtime/errors.go
+++ b/runtime/errors.go
@@ -95,7 +95,7 @@ type InvalidEntryPointTypeError struct {
 }
 
 func (e *InvalidEntryPointTypeError) Error() string {
-	return fmt.Sprintf("invalid entry point type: %s", e.Type.String())
+	return fmt.Sprintf("invalid entry point type: %s", e.Type.QualifiedString())
 }
 
 // InvalidTransactionParameterCountError
@@ -156,7 +156,7 @@ func (e *InvalidTypeAssignmentError) Unwrap() error {
 }
 
 func (e *InvalidTypeAssignmentError) Error() string {
-	return fmt.Sprintf("cannot assign type %s to %s", e.Type, e.Value)
+	return fmt.Sprintf("cannot assign type %s to %s", e.Type.QualifiedString(), e.Value)
 }
 
 // ScriptReturnTypeNotStorableError is an error that is reported for
@@ -172,6 +172,6 @@ type ScriptReturnTypeNotStorableError struct {
 func (e *ScriptReturnTypeNotStorableError) Error() string {
 	return fmt.Sprintf(
 		"return type is non-storable type: %s",
-		e.Type,
+		e.Type.QualifiedString(),
 	)
 }
--- a/runtime/errors.go
+++ b/runtime/errors.go
@@ -95,7 +95,10 @@ type InvalidEntryPointTypeError struct {
 }
 
 func (e *InvalidEntryPointTypeError) Error() string {
-	return fmt.Sprintf("invalid entry point type: %s", e.Type.QualifiedString())
+	return fmt.Sprintf(
+		"invalid entry point type: `%s`",
+		e.Type.QualifiedString(),
+	)
 }
 
 // InvalidTransactionParameterCountError
@@ -156,7 +159,11 @@ func (e *InvalidTypeAssignmentError) Unwrap() error {
 }
 
 func (e *InvalidTypeAssignmentError) Error() string {
-	return fmt.Sprintf("cannot assign type %s to %s", e.Type.QualifiedString(), e.Value)
+	return fmt.Sprintf(
+		"cannot assign type `%s` to %s",
+		e.Type.QualifiedString(),
+		e.Value,
+	)
 }
 
 // ScriptReturnTypeNotStorableError is an error that is reported for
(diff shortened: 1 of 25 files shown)


runtime/errors.go
```diff
@@ -158,3 +158,20 @@ func (e *InvalidTypeAssignmentError) Unwrap() error {
 func (e *InvalidTypeAssignmentError) Error() string {
 	return fmt.Sprintf("cannot assign type %s to %s", e.Type, e.Value)
 }
+
+// ScriptReturnTypeNotStorableError is an error that is reported for
```
turbolent, line 162: 👍

runtime/errors.go
```diff
@@ -158,3 +158,20 @@ func (e *InvalidTypeAssignmentError) Unwrap() error {
...
+// whereas a function type is not.
+
+type ScriptReturnTypeNotStorableError struct {
+	Type sema.Type
+}
+
+func (e *ScriptReturnTypeNotStorableError) Error() string {
+	return fmt.Sprintf(
+		"return type is non-storable type: %s",
+		e.Type,
```
turbolent, line 175: In the error messages we should show the fully-qualified type identifiers to avoid confusion, an we quote them in backticks.

```suggestion
		"return type is non-storable type: `%s`",
		e.Type.QualifiedString(),
```

We already do this everywhere in the sema errors, but not yet here.

While at it, could you please fix the other runtime errors the same way:

- https://github.com/onflow/cadence/blob/benjaminkvm/186-reject-non-storable-return-type/runtime/errors.go#L159
- https://github.com/onflow/cadence/blob/benjaminkvm/186-reject-non-storable-return-type/runtime/errors.go#L174-L175
- https://github.com/onflow/cadence/blob/benjaminkvm/186-reject-non-storable-return-type/runtime/errors.go#L98

runtime/runtime.go
```diff
@@ -150,6 +150,15 @@ func (r *interpreterRuntime) ExecuteScript(
 	}
 	epSignature := invokableType.InvocationFunctionType()
 
+	storableResults := map[*sema.Member]bool{}
+	if !epSignature.ReturnTypeAnnotation.Type.Equal(&sema.VoidType{}) {
```
turbolent, line 154: Could just be a Go type check to avoid the allocation:

```suggestion
	if _, isVoid := epSignature.ReturnTypeAnnotation.Type.(*sema.VoidType); !isVoid {
```

runtime/runtime_test.go
```diff
@@ -1826,6 +1826,32 @@ func TestParseAndCheckProgram(t *testing.T) {
...
 }
 
+func TestScriptReturnTypeNotStorableError(t *testing.T) {
+
+	t.Parallel()
+
+	runtime := NewInterpreterRuntime()
+
+	script := []byte(`
+      pub fun main(): ((): Int) {
```
turbolent, line 1836: 👍

runtime/errors.go
benjaminkvm: Sure thing!

runtime/errors.go
```diff
@@ -95,7 +95,7 @@ type InvalidEntryPointTypeError struct {
 }
 
 func (e *InvalidEntryPointTypeError) Error() string {
-	return fmt.Sprintf("invalid entry point type: %s", e.Type.String())
+	return fmt.Sprintf("invalid entry point type: %s", e.Type.QualifiedString())
```
turbolent, line 98:
```suggestion
	return fmt.Sprintf(
		"invalid entry point type: `%s`", 
		e.Type.QualifiedString(),
	)
```

runtime/errors.go
```diff
@@ -156,5 +156,22 @@ func (e *InvalidTypeAssignmentError) Unwrap() error {
 }
 
 func (e *InvalidTypeAssignmentError) Error() string {
-	return fmt.Sprintf("cannot assign type %s to %s", e.Type, e.Value)
+	return fmt.Sprintf("cannot assign type %s to %s", e.Type.QualifiedString(), e.Value)
```
turbolent, line 159:
```suggestion
	return fmt.Sprintf(
		"cannot assign type `%s` to %s", 
		e.Type.QualifiedString(), 
		e.Value,
	)
```

runtime/errors.go
```diff
@@ -156,5 +156,22 @@ func (e *InvalidTypeAssignmentError) Unwrap() error {
...
+// For example, the type `Int` is a storable type,
+// whereas a function type is not.
+
+type ScriptReturnTypeNotStorableError struct {
+	Type sema.Type
+}
+
+func (e *ScriptReturnTypeNotStorableError) Error() string {
+	return fmt.Sprintf(
+		"return type is non-storable type: %s",
```
turbolent, line 174:
```suggestion
		"return type is non-storable type: `%s`",
```

PR #305, opened 2020-08-08
Reject non-storable transaction/script arguments

### LINKED ISSUE #187: Reject script functions and transactions with non-storable parameter types
### Context

In #169 / #180 we introduced a way to determine if a type can be stored.

The parameter types of script function and transactions must be storable / serializable, so reject script functions and transactions with non-storable parameter types.

### Definition of Done

- [ ] Runtime: Reject script functions and transactions with non-storable parameter types

benjaminkvm:
Closes #187 

## Description

Reject non-storable transaction/script arguments, with relevant test cases.

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -150,11 +150,25 @@ func (r *interpreterRuntime) ExecuteScript(
 	}
 	epSignature := invokableType.InvocationFunctionType()
 
+	// Ensure entrypoint return type is storable
 	storableResults := map[*sema.Member]bool{}
 	if _, isVoid := epSignature.ReturnTypeAnnotation.Type.(*sema.VoidType); !isVoid {
 		if !epSignature.ReturnTypeAnnotation.Type.IsStorable(storableResults) {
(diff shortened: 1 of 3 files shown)


runtime/runtime.go
```diff
@@ -159,6 +160,19 @@ func (r *interpreterRuntime) ExecuteScript(
 		}
 	}
 
+	// Ensure entrypoint parameters type is storable
+	if len(epSignature.Parameters) > 0 {
+		for _, param := range epSignature.Parameters {
+			if _, isVoid := param.TypeAnnotation.Type.(*sema.VoidType); !isVoid {
+				if !param.TypeAnnotation.Type.IsStorable(storableResults) {
+					return nil, &ScriptParameterTypeNotStorableError{
```
turbolent, line 168: Never realized, but `ExecuteTransaction` is using `newError`, and `ExecuteScript` here is not.
@psiemens do you know why the errors are wrapped in one and not the other?

PR #319, opened 2020-08-14
Allow export and import of path values, including JSON encoding

### LINKED ISSUE #318: Allow export of path values
### Problem

The export function does not support values of type `Path`

### Steps to Reproduce

```swift
pub contract Test {
    pub struct R {
        pub let path: Path
        init(path p: Path) {
            self.path = p
        }
    }
    pub resource RSet {
        pub var rsc: {String: R}
        init() {
            self.rsc = {}
        }
    }
	init() {
        var rs <- create RSet()
        rs.rsc["foo"] = R(path: /storage/Foo)
        self.account.save<@RSet>(<-rs, to: /storage/Sets)       
	}
}
```

### Acceptance Criteria

- [x] Path values can be exported, shown in the Playground

turbolent:
Closes #318

### DIFF
--- a/encoding/json/decode.go
+++ b/encoding/json/decode.go
@@ -100,6 +100,8 @@ const (
 	targetKeyKey            = "targetKey"
 	targetPathKey           = "targetPath"
 	borrowTypeKey           = "borrowType"
+	domainKey               = "domain"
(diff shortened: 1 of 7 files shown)


psiemens (approved):
👍

PR #329, opened 2020-08-20
Make recursive types storable

### LINKED ISSUE #328: Execution checking fails on nested types
### Problem

Execution checking fails when trying to store a resource that can recursively stores another copy of itself.

### Steps to Reproduce

Minimum example of bug:
```
pub contract Test7 {
    pub resource Thing {
        pub var nests: @{String: Thing}
        init() {
            self.nests <- {}
        }
        destroy() {
            destroy self.nests
        }
    }
    access(contract) let t: @Thing
    init() {
        self.t <- create Thing()
    }
}
```
The above contract passes static checks against the cadence language server, but when deployed results in
```
DEBU[34773] ️✉️   Transaction submitted
WARN[34773] ❗  Transaction reverted
WARN[34773] ERR [a911e3] Execution failed:
Checking failed:
    field t has non-storable type: Thing 
DEBU[34773] 📦  Block #20 committed
```

The code works fine if `nests` is of any type other than `@Thing`, so somehow the problem is the recursion.

turbolent:
Closes #328 

With #273 merged, it is should now be safe to make recursive types storable.


runtime/sema/type.go
```diff
@@ -5526,12 +5526,12 @@ func (m *Member) IsStorable(results map[*Member]bool) (result bool) {
 		return result
 	}
 
-	// Temporarily assume the member is non-storable while it's type
+	// Temporarily assume the member is storable while it's type
 	// is checked for storability. If a recursive call occurs,
 	// the check for an existing result will prevent infinite recursion
 	// and the result will be correctly false
```
AndrewBurian, line 5532: Can probs drop this last line of the comment, as it no longer guarantees any result on a recursive check.

turbolent, line 5532:
```suggestion
```

turbolent: Good point! Done in bc31767

Issue #221, opened 2020-07-08
Multiple contracts per account


turbolent:
## Context

Currently account code may only declare up to one contract, and multiple contract interfaces.

There is one function, `fun AuthAccount.setCode(_ code: [UInt8], ... initializerArguments)`, which updates the account's code, 
and (if any) instantiates the contract and stores it in the world state.

The function is used both for initial contract deployment and for contract updates. 
For the latter use case, this means that the existing contract instance is overwritten / lost.

Update this API to add support for adding, updating, and removing multiple contracts and contract interfaces, independently.

## Proposal
  
- Add new types:

  - `Contract`:

    A contract that can be deployed to an account.
    The initializer checks the code and aborts if:
    - The code does not contain exactly one contract or contract interface
    - The contract or contract interface name in the code does not match the given name
    - The code/program is invalid

    ```swift
    pub struct Contract {
        pub let name: String
        pub let code: [UInt8]

        init(name: String, code: [UInt8])
    }
    ```

  - `DeployedContract`:
  
    A contract that is deployed at a given account.
  
    ```swift
    pub struct DeployedContract {
        pub let contract: Contract
        pub let address: Address
    }
    ```

  - `AuthAccountContracts`:

    The deployed contracts of an authorized account.
    Allows introspection, and also addition and removal.

    - `fun add(contract: Contract, ... initializerArguments): DeployedContract`
      - Adds the given contract to the account.
        The initializer arguments are passed to the initializer of the contract (if any)
      - Fails if a contract/contract interfaces with the name of the contract/contract interface n the given code already exists in the account

    - `fun update__experimental(contract: Contract): DeployedContract`
      - **Experimental**
      - Updates the code for the contract/contract interface in the given code in the account
      - Does **not** run the initializer of the contract/contract interface in the given code again. The contract instance in the world state stays as is
      - Fails if no contract/contract interface in the account with the name of the contract/contract interface in the given code exist
      - Returns the contract object for the updated contract

    - `fun get(name: String): DeployedContract?`
      - Returns the contract object for the contract/contract interface in the account with the given name, if any
      - Returns nil if no contract/contract interface with the given name exists in the account

    - `fun delete(name: String): DeployedContract?`
      - Deletes the contract/contract interface from the account which has the given name, if any
      - Returns the deleted contract object, if any
      - Returns nil if a contract/contract interface with the given name does not exist in the account

- Add the following member to `AuthAccount`:

  `let contracts: AuthAccountContracts`

[...]

PR #333, opened 2020-08-20
Allow import of multiple programs per address location

turbolent:
Work towards #221

Refactor how programs are imported.

Previously, the code that handles imports assumed that each import location (e.g. an address like `0x1` maps to one program, and the specifically given identifiers in the import statement (if any, otherwise all identifiers) are imported from this single program.

The code that handles imports was previously partially implemented in the AST package and partially in the checker package (`sema`). This PR removes the code in the AST package and implements it fully in the checker package. It is now split into two phases: 

1. Resolution of the import statement. 

    The default case is that one location is resolved to one location (itself), though now an address location may also be resolved into multiple "address contract" locations (naming suggestions welcome).

    For example, an import statement `import a, b from 0x1` specifies the specific identifiers `a` and `b` and the address location `0x1`. It might be resolved into just the location itself, i.e. the address location `0x1`, but can now also resolved into multiple locations, e.g. the address contract location `0x1.a` and the address contract location `0x1.b`.

2. Acquiring the programs for the resolved imports. For each resolved import a separate program can be returned.

The split of the import code into resolution and acquisition required changes to `runtime.Interface`.
Exposing the two phases allows greater flexibility and implementing backwards compatibility for existing deployed contracts in the host environment. 

If we want to keep backwards compatibility of the interface we can revert the changes and hard-code how addresses are resolved, however, I can't quite see how we could keep backwards compatibility for already deployed contracts.

In addition, this refactor also enables repeated imports from the same locations, which was previously not supported, e.g. `import a from 0x1; import b from 0x1`.

While I adjusted the code for the REPL, I added support for importing local files.

### DIFF
--- a/runtime/ast/program.go
+++ b/runtime/ast/program.go
@@ -18,7 +18,9 @@
 
 package ast
 
-import "fmt"
+import (
+	"encoding/json"
+)
 
 type Program struct {
 	// all declarations, in the order they are defined
@@ -35,10 +37,6 @@ type Program struct {
 	_functionDeclarations []*FunctionDeclaration
 	// Use `TransactionDeclarations()` instead
 	_transactionDeclarations []*TransactionDeclaration
-	// Use `ImportedPrograms()` instead
-	_importedPrograms map[LocationID]*Program
-	// Use `ImportedLocations()` instead
-	_importLocations []Location
 }
 
 func (p *Program) StartPosition() Position {
@@ -104,95 +102,6 @@ func (p *Program) TransactionDeclarations() []*TransactionDeclaration {
 	return p._transactionDeclarations
 }
 
-// ImportedPrograms returns the sub-programs imported by this program, indexed by location ID.
-func (p *Program) ImportedPrograms() map[LocationID]*Program {
-	if p._importedPrograms == nil {
-		p._importedPrograms = make(map[LocationID]*Program)
-	}
-
-	return p._importedPrograms
-}
-
-// ImportLocations returns the import locations declared by this program.
-func (p *Program) ImportLocations() []Location {
-	if p._importLocations == nil {
-		p.updateDeclarations()
-	}
-
-	return p._importLocations
-}
-
-type ImportResolver func(location Location) (*Program, error)
-
-func (p *Program) ResolveImports(resolver ImportResolver) error {
-	return p.resolveImports(
-		resolver,
-		map[LocationID]bool{},
-		map[LocationID]*Program{},
-	)
-}
-
-type CyclicImportsError struct {
-	Location Location
-}
-
-func (e CyclicImportsError) Error() string {
-	return fmt.Sprintf("cyclic import of `%s`", e.Location)
-}
-
-func (p *Program) resolveImports(
-	resolver ImportResolver,
-	resolving map[LocationID]bool,
-	resolved map[LocationID]*Program,
-) error {
-	locations := p.ImportLocations()
-
-	for _, location := range locations {
-
-		imported, ok := resolved[location.ID()]
-		if !ok {
-			var err error
-			imported, err = resolver(location)
-			if err != nil {
-				return err
-			}
-			if imported != nil {
-				resolved[location.ID()] = imported
-			}
-		}
-
-		if imported == nil {
-			continue
-		}
-
-		p.setImportedProgram(location.ID(), imported)
-
-		if resolving[location.ID()] {
-			return CyclicImportsError{Location: location}
-		}
-
-		resolving[location.ID()] = true
-
-		err := imported.resolveImports(resolver, resolving, resolved)
-		if err != nil {
-			return err
-		}
-
-		delete(resolving, location.ID())
-	}
-
-	return nil
-}
-
-// setImportedProgram adds an imported program to the set of imports, indexed by location ID.
-func (p *Program) setImportedProgram(locationID LocationID, program *Program) {
-	if p._importedPrograms == nil {
-		p._importedPrograms = make(map[LocationID]*Program)
-	}
-
-	p._importedPrograms[locationID] = program
-}
-
 func (p *Program) updateDeclarations() {
 	// Important: allocate instead of nil
 
@@ -202,7 +111,6 @@ func (p *Program) updateDeclarations() {
 	p._interfaceDeclarations = make([]*InterfaceDeclaration, 0)
 	p._functionDeclarations = make([]*FunctionDeclaration, 0)
 	p._transactionDeclarations = make([]*TransactionDeclaration, 0)
-	p._importLocations = make([]Location, 0)
 
 	for _, declaration := range p.Declarations {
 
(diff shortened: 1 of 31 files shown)


runtime/ast/program.go
```diff
@@ -104,95 +102,6 @@ func (p *Program) TransactionDeclarations() []*TransactionDeclaration {
 	return p._transactionDeclarations
 }
 
-// ImportedPrograms returns the sub-programs imported by this program, indexed by location ID.
```
turbolent, line 107: Remove all code related to imports from the AST package

runtime/cmd/cmd.go
```diff
@@ -84,39 +93,23 @@ func mustClosure(filename string, codes map[string]string) func(error) {
...
-	codes := map[string]string{}
-
+func PrepareProgram(code string, filename string, codes map[string]string) (*ast.Program, func(error)) {
 	must := mustClosure(filename, codes)
 
 	program, err := parser2.ParseProgram(code)
 	codes[filename] = code
 	must(err)
 
-	err = program.ResolveImports(func(location ast.Location) (program *ast.Program, err error) {
```
turbolent, line 105: Now implemented as a checker import handler: https://github.com/onflow/cadence/pull/333/files#diff-040f568b71e2279bc7135a9b78a251b4R134

runtime/cmd/execute/repl.go
```diff
@@ -53,6 +57,37 @@ func RunREPL() {
 
 			println(colorizeResult(value))
 		},
+		[]sema.Option{
+			sema.WithImportHandler(
```
turbolent, line 61: Add support for imports of files (string locations) in the REPL

runtime/interface.go
```diff
@@ -36,9 +37,13 @@ type Block struct {
...
 
+type ResolvedLocation = sema.ResolvedLocation
+
 type Interface interface {
-	// ResolveImport resolves an import of a program.
-	ResolveImport(Location) ([]byte, error)
+	// ResolveLocation resolves an import location.
+	ResolveLocation(identifiers []Identifier, location Location) []ResolvedLocation
+	// GetCode returns the code at a given location
+	GetCode(location Location) ([]byte, error)
```
turbolent, line 46: Split the single import function into two function: resolving the import location, and getting the code for a resolved location

runtime/tests/interpreter/import_test.go
```diff
@@ -115,3 +118,144 @@ func TestInterpretVirtualImport(t *testing.T) {
 		value,
 	)
 }
+
+func TestInterpretImportMultipleProgramsFromLocation(t *testing.T) {
```
turbolent, line 122: This is the most important new test: It demonstrates how two declarations (`a` and `b`) can be imported from the same location (address location `0x1`), are resolved to two locations, which are backed by two programs.

runtime/cmd/cmd.go
```diff
@@ -131,13 +124,42 @@ var typeDeclarations = append(
...
 // and a filename which is used for pretty-printing errors, if any
-func PrepareChecker(program *ast.Program, filename string, must func(error)) (*sema.Checker, func(error)) {
+func PrepareChecker(program *ast.Program, filename string, codes map[string]string, must func(error)) (*sema.Checker, func(error)) {
 	location := runtime.FileLocation(filename)
 	checker, err := sema.NewChecker(
 		program,
 		location,
 		sema.WithPredeclaredValues(valueDeclarations.ToValueDeclarations()),
 		sema.WithPredeclaredTypes(typeDeclarations),
+		sema.WithImportHandler(
```
Kay-Zee, line 134: Nice!

runtime/cmd/cmd.go
```diff
@@ -131,13 +124,42 @@ var typeDeclarations = append(
...
+							fmt.Errorf("cannot import `%s`. only files are supported", location),
+						},
+					}
+				}
+
+				importChecker, err := checker.EnsureLoaded(
+					location,
+					func() *ast.Program {
+						filename := string(stringLocation)
+						imported, _ := PrepareProgramFromFile(filename, codes)
```
Kay-Zee, line 150: Is this an err that's skipped?

runtime/runtime_test.go
```diff
@@ -2337,48 +2350,6 @@ func TestRuntimeTransaction_AddPublicKey(t *testing.T) {
 	}
 }
 
-func TestRuntimeCyclicImport(t *testing.T) {
```
Kay-Zee, line 2340: Is this moved somewhere? Or covered by another test? `TestInterpretImportMultipleProgramsFromLocation` maybe?

runtime/tests/interpreter/import_test.go
```diff
@@ -115,3 +118,150 @@ func TestInterpretVirtualImport(t *testing.T) {
...
+
+	addressLocation := ast.AddressLocation{0x1}
+
+	importedCheckerA, err := checker.ParseAndCheckWithOptions(t,
+		`
+          pub fun a(): Int {
+              return 1
+          }
+
+          pub fun b(): Int {
```
Kay-Zee, line 140: In this scenario, there'd be no way to access this `b()` func?

Kay-Zee (approved):
Sorry this took so long, there was a lot of content to go through.

I spent a majority of my focus on the tests to make sure they make sense to me. Had a few small questions, but otherwise LGTM

runtime/tests/interpreter/import_test.go
turbolent: Good question!

Correct, this function is only declared to ensure that it is **not** imported in the `import a, b from 0x1` statement.

Likewise, [the function `a` below](https://github.com/onflow/cadence/pull/333/files#diff-f759a5f9b49ae7e692b50eb5926c786cR159) should not be imported either. 

I'll add comments to make this test clearer

runtime/cmd/cmd.go
turbolent: It looks like it, but the `PrepareProgramFromFile` function does not return an error as it's second return value, but a `must` function. 

This got introduced at some point and I don't like it, because it's not Go idiomatic and confusing. It would be nice to refactor that at some point.

If you look at `PrepareProgramFromFile`, it already checks the error using the must function, then just returns it so that callers have a convenient way of reporting an error through a single function.

runtime/runtime_test.go
turbolent: This PR initially removed an explicit check for cyclic imports. Cyclic imports where still not supported, but instead of getting an dedicated error for cyclic imports, the checker would still fail, but only through implicit "not declared" errors.

Thinking about it a bit more, this is not very user-friendly: a developer might only see a "not declared" error, see they have a valid declaration, but would not realize the actual cause is a cyclic import. 

I've added the explicit check and error back in [`2b713fc` (#333)](https://github.com/onflow/cadence/pull/333/commits/2b713fcff712cbb54271811a50f9e15c1828503a).

Regarding this test case in particular: This is a runtime check, but all it really tests is that the checker detects the cyclic import, so it shouldn't be a runtime test in the first place, but rather a checker test.

I've improved https://github.com/onflow/cadence/pull/333/files#diff-4fb398062392664e4c2200a9025f9f42R543 and also added another test for a mutually recursive import here:
https://github.com/onflow/cadence/pull/333/files#diff-4fb398062392664e4c2200a9025f9f42R586.
Basically, these two checker tests cover this test and we don't need the runtime test here.

runtime/tests/interpreter/import_test.go
turbolent: Improved this in [`36421c5` (#333)](https://github.com/onflow/cadence/pull/333/commits/36421c5375832e8f4de5d26fdf78726b7c530d01)

PR #346, opened 2020-08-24
Suggest replacing number conversion function calls 

turbolent:
If a number conversion function is called with a number literal, suggest replacing it with a literal and as static cast (if needed):


runtime/sema/checker.go
```diff
@@ -77,6 +77,7 @@ type Checker struct {
 	allCheckers                        map[ast.LocationID]*Checker
 	accessCheckMode                    AccessCheckMode
 	errors                             []error
+	hints                              []Hint
```
benjaminkvm, line 80: 👍

PR #347, opened 2020-08-24
Replace reporting an error for forcing non-optional with a hint suggesting removal

turbolent:
Instead of reporting an error when the force operator is used on a non-optional type, only report a hint suggesting to remove it.

This is pre-work for replacing the return type of functions which currently have an return type which is optional which one that is not. Specifically, this will allow changing the signature of `getCapability` without breaking backwards compatibility.

### DIFF
--- a/runtime/sema/hints.go
+++ b/runtime/sema/hints.go
@@ -45,3 +45,20 @@ func (h *ReplacementHint) Hint() string {
 }
 
 func (*ReplacementHint) isHint() {}
+
+// RemovalHint
+
+type RemovalHint struct {
+	Description string
+	ast.Range
+}
+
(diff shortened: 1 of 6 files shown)


runtime/sema/hints.go
```diff
@@ -45,3 +45,20 @@ func (h *ReplacementHint) Hint() string {
 }
 
 func (*ReplacementHint) isHint() {}
+
+// RemovalHint
+
+type RemovalHint struct {
```
benjaminkvm, line 51: Curious about potentially making this more specific, could happen later on however! `RedundantCodeRemovalHint`, etc.

benjaminkvm (approved):

psiemens (approved):
Nice!

PR #350, opened 2020-08-27
Make FieldTypeNotStorableError a semantic error

turbolent:
The error type `FieldTypeNotStorableError` was not implementing the `SemanticError` interface, so the language server failed to convert and display it. The only indicator was a wrong log message ("Unable to convert non-convertable error to disgnostic")

Reported by @AndrewBurian 🙏

### DIFF
--- a/languageserver/server/server.go
+++ b/languageserver/server/server.go
@@ -1183,7 +1183,7 @@ func getDiagnosticsForParentError(
 		if !ok {
 			conn.LogMessage(&protocol.LogMessageParams{
 				Type:    protocol.Error,
-				Message: fmt.Sprintf("Unable to convert non-convertable error to disgnostic: %s", err.Error()),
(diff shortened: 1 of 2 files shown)


languageserver/server/server.go
```diff
@@ -1183,7 +1183,7 @@ func getDiagnosticsForParentError(
 		if !ok {
 			conn.LogMessage(&protocol.LogMessageParams{
 				Type:    protocol.Error,
-				Message: fmt.Sprintf("Unable to convert non-convertable error to disgnostic: %s", err.Error()),
+				Message: fmt.Sprintf("Unable to convert non-convertable error to disgnostic: %T", childErr),
```
benjaminkvm, line 1186: to *diagnostic* %T"

turbolent, line 1186:
```suggestion
				Message: fmt.Sprintf("Unable to convert non-convertable error to diagnostic: %T", childErr),
```

psiemens (approved):

PR #355, opened 2020-08-28
Copy values on return

turbolent:
I incorrectly removed the copying of return values in a previous PR (https://github.com/onflow/cadence/pull/288/files#diff-a8c2ab8c721531fd42f8f5ecbad5ca28L1092-R1092), but as @joshuahannan pointed out, call-sites could directly access the returned value without assigning it e.g to a variable (which copies), allowing modification of the returned value.


joshuahannan:
Thanks to @alxocity for pointing this out!

PR #356, opened 2020-08-28
Fix composite-type explicit-conformances record timing

### LINKED ISSUE #326: Resource interface conformance issue
### Problem

When specifying a restricted reference to a resource as a return value, the checker reports that `restricted type does not conform to restricting type`. Is this intentional, or is it a bug?

### Steps to Reproduce

1. Copy this code:
```
pub contract Nester {
    pub resource interface Nesting {
        pub fun get(): &Thing{Nesting}
    }
    pub resource Thing: Nesting {
        pub fun get(): &Thing{Nesting}
    }
}
```
2. See the error on the return types: restricted type does not conform to restricting type: Nester.Nesting

turbolent:
Fixes #326

Determine and record explicit composite conformances earlier, before they are used.

### DIFF
--- a/runtime/sema/check_composite_declaration.go
+++ b/runtime/sema/check_composite_declaration.go
@@ -413,6 +413,13 @@ func (checker *Checker) declareCompositeType(declaration *ast.CompositeDeclarati
 		variable,
 	)
 
+	// Resolve conformances
+
+	conformances := checker.explicitInterfaceConformances(declaration, compositeType)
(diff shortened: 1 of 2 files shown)


turbolent:
Thanks @joshuahannan and @AndrewBurian for reporting this 👍

benjaminkvm (approved):
Good catch Josh! LGTM

zhangchiqing (approved):

PR #226, opened 2020-07-09
Start source compatibility suite, improve command line

turbolent:
- Start a source compatibility suite, similar to [Swift's](https://swift.org/source-compatibility/), to prevent regressions and understand what impact changes in the language and implementation have on real-world Cadence projects.

  The suite contains a set of repository descriptions. When the suite is run, the repositories get cloned an checked. The runner can optionally run benchmarks and compare against another commit / branch, producing output in the terminal (pretty), as JSON, or Markdown. 

  In the future it we can integrate this as part of CI, maybe as a periodic job.

  See the example report at the end of the PR description.

- Improve the command line and REPL by providing the Flow-specific value and type declarations (e.g. `getAccount`), though no implementations (the implementations abort with an error message the functionality is not supported)
- Add a new command line program which only checks a given program.


### Example Report


## Parser Errors

| 🙂 Still succeeding                       |
|:-----------------------------------------|
| flow-ft/contracts/ExampleToken.cdc       |
| flow-ft/contracts/FungibleToken.cdc      |
| flow-ft/contracts/TokenForwarding.cdc    |
| flow-ft/transactions/burn_tokens.cdc     |
| flow-ft/transactions/mint_tokens.cdc     |
| flow-ft/transactions/setup_account.cdc   |
| flow-ft/transactions/transfer_tokens.cdc |
| flow-nft/contracts/ExampleNFT.cdc        |
| flow-nft/contracts/NonFungibleToken.cdc  |
| flow-nft/transactions/mint_nft.cdc       |
| flow-nft/transactions/setup_account.cdc  |
| flow-nft/transactions/transfer_nft.cdc   |


## Parser Benchmarks

| 😭 Regressions                            |     OLD |     NEW | DELTA   | RATIO     |
|:-----------------------------------------|--------:|--------:|:--------|:----------|
| flow-ft/transactions/burn_tokens.cdc     | 1081.28 | 1196.89 | -10.69% | **0.90x** |
| flow-ft/transactions/transfer_tokens.cdc | 1079.44 | 1130.04 | -4.69%  | **0.96x** |
| flow-nft/contracts/NonFungibleToken.cdc  | 1114.89 | 1161.6  | -4.19%  | **0.96x** |

| 🎉 Improvements                        |     OLD |     NEW | DELTA   | RATIO     |
|:--------------------------------------|--------:|--------:|:--------|:----------|
| flow-ft/contracts/ExampleToken.cdc    | 1182.53 | 1085.63 | +8.19%  | **1.09x** |
[...]

PR #362, opened 2020-09-01
Improve compat suite

turbolent:
- Add a `requirements.txt` that lists all dependencies
- Add instructions how to run the suite

### DIFF
--- a/compat/README.md
+++ b/compat/README.md
@@ -1,5 +1,11 @@
 # Source Compatibility Suite
 
+The source compatibility suite prevents regressions and helps understand what impact changes in the language and implementation have on real-world Cadence projects.
+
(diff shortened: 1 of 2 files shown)


compat/README.md
```diff
@@ -1,3 +1,15 @@
 # Source Compatibility Suite
 
+## Running
```
zhangchiqing, line 3: Could you briefly explain what the compatibility suite do?

compat/README.md
```diff
@@ -1,3 +1,15 @@
 # Source Compatibility Suite
```
turbolent, line 1:
```suggestion
# Source Compatibility Suite

The source compatibility suite prevents regressions and helps understand what impact changes in the language and implementation have on real-world Cadence projects.

The suite contains a [set of repository descriptions](https://github.com/onflow/cadence/tree/master/compat/suite). When the suite is run, the repositories get cloned and checked. The runner can optionally run benchmarks and compare against another commit / branch, producing output in the terminal (pretty), as JSON, or Markdown.

In the future we can integrate this as part of CI, maybe as a periodic job.
```

compat/README.md
turbolent: Added above, also see #226

zhangchiqing (approved):
👍

PR #365, opened 2020-09-07
Switch statement

### LINKED ISSUE #295: Switch statement
### Context

Add a `switch` statement, which
- Does not fall-through by default
- Only checks for equality for now
- Only supports on value per `case` for now
- Does not check for exhaustiveness yet

In the future this feature can be extended.

Example:
```swift
switch 3 {
case 1:
    log("value is 1")
case 2:
    log("value is 2")
default:
    log("value is larger than 2")
}
```

### Definition of Done

- [x] Add AST element
- [x] Parse
- [x] Check
- [x] Interpret
- [x] Tests
- [ ] Documentation

turbolent:
Closes #295


runtime/tests/interpreter/if_test.go
```diff
@@ -0,0 +1,300 @@
...
+           }
+
+           pub fun testElseIf(): Int {
+               if false {
+                   return 2
+               } else if true {
+                   return 3
+               }
+               return 4
+           }
```
zhangchiqing, line 70:
```suggestion
           }
           
           pub fun testElseIf(): Int {
               if false {
                   return 2
               } else if false {
                   return 3
               } else {
                   return 4
               }
               return 5
           }
```

adding `else if false`

runtime/tests/checker/switch_test.go
```diff
@@ -0,0 +1,293 @@
...
+          }
+      }
+    `)
+
+	errs := ExpectCheckerErrors(t, err, 1)
+
+	assert.IsType(t, &sema.NotDeclaredError{}, errs[0])
+}
+
+func TestCheckInvalidSwitchStatementDefaultPosition(t *testing.T) {
```
zhangchiqing, line 232: 👍

zhangchiqing (approved):
Nice work!

runtime/tests/interpreter/if_test.go
turbolent: This test is not valid, the name is duplicated and the `return 5` is unreachable

runtime/tests/interpreter/if_test.go
```diff
@@ -0,0 +1,311 @@
...
+           pub fun testElseIf(): Int {
+               if false {
+                   return 2
+               } else if false {
+                   return 3
+               } else {
+                   return 4
+               }
+               return 5
+           }
```
turbolent, line 81:
```suggestion
           pub fun testElseIfElse(): Int {
               if false {
                   return 2
               } else if false {
                   return 3
               } else {
                   return 4
               }
           }
```

PR #366, opened 2020-09-09
Add WebAssembly package with WASM binary writer and reader

turbolent:
Start work on experimenting with WebAssembly: Add a new package which allows writing and reading WASM binaries. For now, only enough instructions are implemented to implement an add function, i.e. `local.get`, `i32.add`, and `end`

### DIFF
--- a/runtime/compiler/wasm/buf.go
+++ b/runtime/compiler/wasm/buf.go
@@ -1,0 +1,85 @@
+/*
+ * Cadence - The resource-oriented smart contract programming language
+ *
+ * Copyright 2019-2020 Dapper Labs, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package wasm
+
+import (
+	"io"
+)
+
+type offset int
+
+// buf is a byte buffer, which allows reading and writing.
+//
+type buf struct {
+	data   []byte
+	offset offset
+}
+
+func (buf *buf) WriteByte(b byte) error {
+	if buf.offset < offset(len(buf.data)) {
+		buf.data[buf.offset] = b
+	} else {
+		buf.data = append(buf.data, b)
+	}
+	buf.offset++
+	return nil
+}
+
+func (buf *buf) WriteBytes(data []byte) error {
+	for _, b := range data {
+		err := buf.WriteByte(b)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (buf *buf) Read(data []byte) (int, error) {
+	n := copy(data, buf.data[buf.offset:])
+	if n == 0 && len(data) != 0 {
+		return 0, io.EOF
+	}
+	buf.offset += offset(n)
+	return n, nil
+}
+
+func (buf *buf) ReadByte() (byte, error) {
+	if buf.offset >= offset(len(buf.data)) {
+		return 0, io.EOF
+	}
+	b := buf.data[buf.offset]
+	buf.offset++
+	return b, nil
+}
+
+func (buf *buf) ReadBytesEqual(expected []byte) (bool, error) {
+	off := buf.offset
+	for _, b := range expected {
+		if off >= offset(len(buf.data)) {
+			return false, io.EOF
+		}
+		if buf.data[off] != b {
+			return false, nil
+		}
+		off++
+	}
+	buf.offset = off
+	return true, nil
+}
--- a/runtime/compiler/wasm/leb128_test.go
+++ b/runtime/compiler/wasm/leb128_test.go
@@ -1,0 +1,168 @@
+/*
+ * Cadence - The resource-oriented smart contract programming language
+ *
+ * Copyright 2019-2020 Dapper Labs, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package wasm
+
+import (
+	"math"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestBuf_writeULEB128(t *testing.T) {
+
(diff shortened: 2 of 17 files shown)


runtime/compiler/wasm/reader_test.go
```diff
@@ -0,0 +1,825 @@
...
+func TestWasmReader_readMagicAndVersion(t *testing.T) {
+
+	t.Parallel()
+
+	read := func(data []byte) error {
+		b := buf{data: data}
+		r := WASMReader{buf: &b}
+		return r.readMagicAndVersion()
+	}
+
```
zhangchiqing, line 38: Missing happy cases:  "valid magic" and "valid magic and version"

runtime/compiler/wasm/reader_test.go
```diff
@@ -0,0 +1,825 @@
...
+					Offset:  8,
+					ValType: 0xFF,
+				},
+			},
+			err,
+		)
+		assert.Nil(t, funcTypes)
+	})
+
+	t.Run("invalid, parameter type missing", func(t *testing.T) {
```
zhangchiqing, line 332: 👍 

This is a case when not enough parameter type is provided, I wonder if too many parameter types are provided, what error would it return? I guess the extra parameter type would be parsed as the next item, which is "return value count", which would shows a different error message?

runtime/compiler/wasm/reader_test.go
```diff
@@ -0,0 +1,825 @@
...
+		assert.Nil(t, funcTypes)
+	})
+
+	t.Run("invalid count", func(t *testing.T) {
+
+		t.Parallel()
+
+		funcTypes, err := read([]byte{
+			// section size: 7 (LEB128)
+			0x80, 0x80, 0x80, 0x80, 0x0,
```
zhangchiqing, line 511: first byte for `section size: 7` should be `0x87`, but got `0x80` here

runtime/compiler/wasm/leb128_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+	t.Run("write: max byte count", func(t *testing.T) {
+		var b buf
+		err := b.writeULEB128(math.MaxUint32)
+		require.NoError(t, err)
+		require.GreaterOrEqual(t, max32bitLEB128ByteCount, len(b.data))
+	})
+
+	t.Run("read: max byte count", func(t *testing.T) {
+		b := buf{data: []byte{0x81, 0x82, 0x83, 0x84, 0x85, 0x86, 0x87, 0x88}}
```
zhangchiqing, line 63: Do these bytes represent the biggest value can be read?
isn't biggest uint32 can be represented as `0xFF, 0xFF, 0xFF, 0xFF`?

runtime/compiler/wasm/writer_test.go
```diff
@@ -0,0 +1,288 @@
...
+	require.Equal(t,
+		`(module
+  (type (;0;) (func (param i32 i32) (result i32)))
+  (func (;0;) (type 0) (param i32 i32) (result i32)
+    (local i32)
+    local.get 0
+    local.get 1
+    i32.add))
+`,
+		wasm2wat(b.data),
```
zhangchiqing, line 286: 👍  Nice to compare to the official library

runtime/compiler/wasm/reader_test.go
```diff
@@ -0,0 +1,825 @@
...
+
+	t.Run("valid", func(t *testing.T) {
+
+		t.Parallel()
+
+		codes, err := read([]byte{
+			// section size: 15 (LEB128)
+			0x8f, 0x80, 0x80, 0x80, 0x0,
+			// function count: 1
+			0x1,
```
zhangchiqing, line 569: I was afraid these raw bytes would be unreadable, and useless for review. 
But after reading the code, these bytes and comments are actually quite easier to read!
Nice comments 👍

zhangchiqing (approved):
As always, very clean implementation and great tests!

I had comments and questions, but looks great 👍

turbolent:
@zhangchiqing Thank you for the review! 

I added documentation as a package comment in 40ccbda. Hopefully that answers some of your questions, I'll add more in the other packages later. Let me know if you still have questions regarding the `wasm` package and I can clarify.

runtime/compiler/wasm/buf.go
```diff
@@ -0,0 +1,85 @@
...
+		buf.data = append(buf.data, b)
+	}
+	buf.offset++
+	return nil
+}
+
+func (buf *buf) WriteBytes(data []byte) error {
+	for _, b := range data {
+		err := buf.WriteByte(b)
+		if err != nil {
```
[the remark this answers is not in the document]
turbolent: `WriteByte` implements the `io.ByteWriter` interface, which requires `WriteByte(c byte) error`

runtime/compiler/wasm/magic.go
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package wasm
+
+// wasmMagic is the magic byte sequence that appears at the start of the WASM binary
```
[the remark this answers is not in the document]
turbolent: Good idea, done in 43a4951

runtime/compiler/wasm/reader_test.go
turbolent: Good catch, fixed in ebf8166

runtime/compiler/wasm/section.go
```diff
@@ -0,0 +1,29 @@
...
+package wasm
+
+// sectionID is the ID of a section in the WASM binary
+//
+type sectionID byte
+
+const (
+	sectionIDType     sectionID = 1
+	sectionIDFunction sectionID = 3
+	sectionIDCode     sectionID = 10
```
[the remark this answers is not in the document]
turbolent: Good idea, done in 6246b9e

runtime/compiler/wasm/writer_test.go
turbolent: Note that the tool is not called to "compare": The wasm package produces a binary, and the `wasm2wat` tool is used to check if the written binary (WASM) is valid by parsing it and converting it to the text format (WAT).

runtime/compiler/wasm/reader_test.go
turbolent: I'm glad they're helpful! I think once you have internalized the structure of the binary format, it becomes readable and when something is mis-produced, it helps with debugging.

runtime/compiler/wasm/leb128_test.go
turbolent: No they don't, note it says "max byte count" not "max byte value". This test ensures that only up to the maximum number of bytes are read when reading a LEB128-encoded 32-bit number, which is 5 (see https://github.com/onflow/cadence/blob/ea6b35ba72b9622a814c3d27d718d120a594f06f/runtime/compiler/wasm/leb128.go#L145-L151), i.e. the function checks that only 5 out of the 8 bytes are read, to ensure the LEB128 parser doesn't keep reading infinitely.

I've added comments to clarify this in 4e7e058c9fd16e4884825d3db3cf9b768b70eb3a

runtime/compiler/wasm/reader_test.go
turbolent: See below, https://github.com/onflow/cadence/pull/366/files#diff-0f55d4941b179be22cb3099b9e25051d6d10cf2bf71fd597d8d50d0007bea6ebR99

PR #368, opened 2020-09-10
Add support for imports in WASM writer and reader

turbolent:


runtime/compiler/wasm/writer_test.go
```diff
@@ -115,7 +116,50 @@ func TestWasmWriter_writeTypeSection(t *testing.T) {
...
+		[]byte{
+			// section ID: Import = 2
+			0x2,
+			// section size: 11 (LEB128)
+			0x8b, 0x80, 0x80, 0x80, 0x0,
+			// import count: 1
+			0x1,
+			// module length
+			0x3,
+			// module = "foo"
```
zhangchiqing, line 147: Thanks to these comments, the bytes make a lot sense. 👍

runtime/compiler/wasm/writer_test.go
```diff
@@ -263,21 +366,36 @@ func TestWasmWriter(t *testing.T) {
...
+			0x01, 0x03, 0x65, 0x6e, 0x76, 0x03, 0x61, 0x64,
+			0x64, 0x00, 0x00,
+			// function section
+			0x3,
+			0x82, 0x80, 0x80, 0x80, 0x0,
+			0x1, 0x0,
+			// code section
+			0xa, 0x8f, 0x80, 0x80, 0x80, 0x0,
+			0x1, 0x89, 0x80, 0x80, 0x80, 0x0, 0x1, 0x1,
+			0x7f, 0x20, 0x0, 0x20, 0x1, 0x6a, 0xb,
```
zhangchiqing, line 389: Nice formatting 👍

runtime/compiler/wasm/reader_test.go
```diff
@@ -451,7 +451,247 @@ func TestWasmReader_readTypeSection(t *testing.T) {
...
+		)
+		assert.Nil(t, funcTypes)
+	})
+
+	t.Run("invalid count", func(t *testing.T) {
+
+		t.Parallel()
+
+		funcTypes, err := read([]byte{
+			// section size: 7 (LEB128)
```
zhangchiqing, line 526: the comment seems off

runtime/compiler/wasm/reader_test.go
```diff
@@ -451,7 +451,247 @@ func TestWasmReader_readTypeSection(t *testing.T) {
...
+			0x81, 0x80, 0x80, 0x80, 0x0,
+			// import count
+			0x1,
+		})
+		require.Error(t, err)
+		assert.Equal(t,
+			InvalidImportError{
+				Index: 0,
+				ReadError: InvalidNameLengthError{
```
zhangchiqing, line 554: Should this be `InvalidImportError`, instead of `InvalidNameLengthError`?

runtime/compiler/wasm/reader_test.go
```diff
@@ -451,7 +451,247 @@ func TestWasmReader_readTypeSection(t *testing.T) {
...
+
+		funcTypes, err := read([]byte{
+			// section size: 7 (LEB128)
+			0x81, 0x80, 0x80, 0x80, 0x0,
+			// import count
+			0x1,
+			// module length
+			0x3,
+			// module = "foo"
+			0x66, 0x6f, 0x6f,
```
zhangchiqing, line 576: Why is this invalid?

If we are testing the invalid name, we should construct the following tests:

```
			// name length
			0x3,
			// name = "ba", which only have length 2
			0x62, 0x61,
```

runtime/compiler/wasm/reader_test.go
```diff
@@ -451,7 +451,247 @@ func TestWasmReader_readTypeSection(t *testing.T) {
...
+
+		typeIDs, err := read([]byte{
+			// section size: 11 (LEB128)
+			0x8b, 0x80, 0x80, 0x80, 0x0,
+			// import count: 1
+			0x1,
+			// module length
+			0x3,
+			// module = "foo"
+			0x66, 0x6f, 0x6f,
```
zhangchiqing, line 480: I'm not sure if we have the test case that verifies the module length match, basically we can verify that the following case should fail
```
			// module length
			0x3,
			// module = "fo", the module length, which is 2, is mismatching with the declared module length 3
			0x66, 0x6f
```

runtime/compiler/wasm/reader_test.go
turbolent: Good catch! This was wrong in several places (due to copy and paste), fixed all cases in db04dad5bee706382ba749a7bb8f1f0f02a13305

runtime/compiler/wasm/reader_test.go
turbolent: The errors are wrapped: The outer `InvalidImportError` indicates the i-th import in the import section failed to be read, and the inner error has the concrete cause of the invalid error: the name at the given offset was invalid.

runtime/compiler/wasm/reader_test.go
turbolent: Currently the test tests the case where even the name length (see your comment) is missing.

The example you gave is another invalid case. I don't think we have to re-test invalid names in the import sections, because parsing of names is already tested by `TestWASMReader_readName`.

runtime/compiler/wasm/reader_test.go
turbolent: Same as above: Parsing names is a "building block" for parsing imports (and some other parts), so I don't think we need to re-test parsing of names itself in each larger test that uses it – parsing of names is already tested by the unit test `TestWASMReader_readName`, and in particular, the subtest "invalid name" (https://github.com/onflow/cadence/blob/db04dad5bee706382ba749a7bb8f1f0f02a13305/runtime/compiler/wasm/reader_test.go#L1109) covers exactly this case: The length is larger then the length of the string that follows

zhangchiqing (approved):
Look good 👍

PR #372, opened 2020-09-13
Generate code for instructions

turbolent:
- Add support for encoding and decoding 64-bit LEB128 integers
- Generate the code for the declaration, writing, and reading of instructions.
  This makes it easy to add more instructions, reduces boilerplate, and ensures the code for writing and reading the instructions stays in sync.


runtime/compiler/wasm/instruction.go
```diff
@@ -22,24 +22,6 @@ type Instruction interface {
 	write(*WASMWriter) error
 }
 
-// InstructionLocalGet is the 'local.get' instruction
+// opcode is the byte used to indicate a certain instruction in the WASM binary
 //
-type InstructionLocalGet struct {
```
turbolent, line 27: Moved to `instructions.go`, which is now generated

runtime/compiler/wasm/gen/main.go
```diff
@@ -0,0 +1,355 @@
...
+				Depth:        0,
+				Instructions: instructions,
+			},
+		)
+		if err != nil {
+			panic(err)
+		}
+	}
+
+	declare([]instruction{
```
turbolent, line 333: The main contribution of this PR: decoratively specify the instructions and generate the code for the types, the writing, the reading, and the dispatch/switch for them (including support for opcodes with prefixes)

SupunS (approved):

PR #374, opened 2020-09-14
Allow export and JSON encoding/decoding of type values and capabilities

### LINKED ISSUE #373: Allow saving of types and capabilities into contracts/resources in Playground
### Instructions
NFT or contract could have a field specified on them, which gives a hint on type to cast the reference.

### Problem

Playground doesn't seem to support this at the moment.

### Steps to Reproduce

Here is Playground with test case:
https://play.onflow.org/603a8982-9198-4be7-9f55-0cc91ae29f6f

- Deploy `FungibleToken` contract from 0x01
- Try to deploy `CustomVault` contract from 0x02 - Playground will throw `GraphQL error: internal system error`

turbolent:
Closes #373 

For now, just like for link values, encode the static type as a string.
A future PR may introduce support for exporting static types and encoding/decoding them, but for now that is not needed.

### DIFF
--- a/encoding/json/decode.go
+++ b/encoding/json/decode.go
@@ -103,6 +103,8 @@ const (
 	domainKey               = "domain"
 	identifierKey           = "identifier"
 	staticTypeKey           = "staticType"
+	addressKey              = "address"
+	pathKey                 = "path"
 )
 
 var ErrInvalidJSONCadence = errors.New("invalid JSON Cadence structure")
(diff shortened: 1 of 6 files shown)


tarakby (approved):
Nice, I only checked the syntax though, maybe someone else with Cadence experience should look at the logic.

zhangchiqing (approved):
Looks good!

PR #378, opened 2020-09-17
Fix parsing of right bitwise shift and function invocation where type argument is instantiation

turbolent:
## Bug

Parsing of function invocations which contain a type argument which is a type instantiation is not working, e.g. `f<T<U>>()`.

## Reason

The parser expects a closing '>' token, however the lexer returns a '>>' token.

## Solution

Remove the '>>' token. Replace the twp separate left denotations for '>' and '>>' with a single meta left denotation which  preforms a lookahead for the second '>' token of ">>".

### DIFF
--- a/runtime/parser2/expression.go
+++ b/runtime/parser2/expression.go
@@ -249,17 +249,12 @@ func init() {
 	})
 
 	defineLessThanOrTypeArgumentsExpression()
+	defineGreaterThanOrBitwiseRightShiftExpression()
 
 	defineExpr(binaryExpr{
 		tokenType:        lexer.TokenLessEqual,
 		leftBindingPower: exprLeftBindingPowerComparison,
 		operation:        ast.OperationLessEqual,
-	})
-
-	defineExpr(binaryExpr{
-		tokenType:        lexer.TokenGreater,
-		leftBindingPower: exprLeftBindingPowerComparison,
-		operation:        ast.OperationGreater,
 	})
 
 	defineExpr(binaryExpr{
@@ -309,12 +304,6 @@ func init() {
 		tokenType:        lexer.TokenLessLess,
 		leftBindingPower: exprLeftBindingPowerBitwiseShift,
 		operation:        ast.OperationBitwiseLeftShift,
-	})
-
-	defineExpr(binaryExpr{
-		tokenType:        lexer.TokenGreaterGreater,
-		leftBindingPower: exprLeftBindingPowerBitwiseShift,
-		operation:        ast.OperationBitwiseRightShift,
 	})
 
 	defineExpr(binaryExpr{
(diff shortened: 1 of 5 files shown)


runtime/parser2/expression.go
```diff
@@ -626,6 +615,95 @@ func defineLessThanOrTypeArgumentsExpression() {
 		})
 }
 
+func defineGreaterThanOrBitwiseRightShiftExpression() {
```
zhangchiqing, line 618:
```suggestion
// parse "greater than" (`>`) and "bitwise right shift" (`>>`).
// distinguish them while parsing, since they share a common token (`>`)
// also distinguish the different `>>` between bitwise right shift (`1 >> 3`) and
// generic function call (`f<T<U>>()`)
func defineGreaterThanOrBitwiseRightShiftExpression() {
```

runtime/parser2/expression.go
```diff
@@ -626,6 +615,95 @@ func defineLessThanOrTypeArgumentsExpression() {
...
+
+			if rightBindingPower >= exprLeftBindingPowerBitwiseShift &&
+				rightBindingPower >= exprLeftBindingPowerComparison {
+
+				return left, true
+			}
+
+			// Start buffering before skipping the `>` token,
+			// so it can be replayed in case the right binding power
+			// was higher than the determined left binding power.
```
zhangchiqing, line 635:
```suggestion
			// Since `>` and `>>` share a common token, we need to distinguish them by buffering a `>`
			// and see if the next is another `>`:
			// if yes, it's a `>>`; if no, then it's a `>`.
			//
			// buffering allows us to replay the buffered tokens in case the right
			// binding power was higher than the determined left binding power.
```

runtime/parser2/expression.go
```diff
@@ -626,6 +615,95 @@ func defineLessThanOrTypeArgumentsExpression() {
...
+
+			if isBitwiseShift {
+
+				operation = ast.OperationBitwiseRightShift
+
+				// The expression was determined to be a bitwise shift.
+				// Still, it should have maybe not been parsed if the right binding power
+				// was higher. In that case, replay the buffered tokens and stop.
+
+				if rightBindingPower >= exprLeftBindingPowerBitwiseShift {
```
zhangchiqing, line 658:
```suggestion
				if rightBindingPower >= exprLeftBindingPowerBitwiseShift {
				// for instance, this could happen when parsing `a<T<U>>()`
				// where `()`'s binding power is higher than `>>`.
```

runtime/parser2/expression.go
```diff
@@ -626,6 +615,95 @@ func defineLessThanOrTypeArgumentsExpression() {
...
+
+				nextRightBindingPower = exprLeftBindingPowerComparison
+			}
+
+			p.next()
+			p.skipSpaceAndComments(true)
+
+			right := parseExpression(p, nextRightBindingPower)
+
+			binaryExpression := &ast.BinaryExpression{
```
zhangchiqing, line 697:
```suggestion
			// bitwise right shift (>>) and greater than (>) are both BinaryExpression
			binaryExpression := &ast.BinaryExpression{
```

zhangchiqing (approved):

runtime/parser2/expression.go
turbolent: Great suggestion! Corrected and extended it further

PR #379, opened 2020-09-17
Add a hash function to the crypto contract

turbolent:
Add support for hashing data to the crypto contract.

Add a `hash` function to both the `Crypto` contract and the `HashAlgorithm` struct.

For example, the new functions can be used like this:

```swift
Crypto.SHA3_256.hash("01020304".decodeHex()) ==
Crypto.hash("01020304".decodeHex(), algorithm: Crypto.SHA3_256)
```


joshuahannan (approved):
Looks good to me! Thanks for adding!

psiemens (approved):
Looks great, nice!

PR #385, opened 2020-09-24
Various documentation improvements

andrejtokarcik:
This changeset contains corrections of typos and errors that I happened to stumble upon as I was reading through the language reference and other docs.  Feel free to let me know if I should adapt/reorganize any of the proposed changes.

Note: I fixed the broken link to the VS Code Extension docs in `/README.md` and replaced it by `https://github.com/onflow/flow/blob/master/docs/vscode-extension.md#cadence-visual-studio-code-extension`. However, at docs.onflow.org the left bar's "VS Code Extension" link points to `https://github.com/onflow/vscode-flow#cadence-visual-studio-code-extension` instead. It's not clear which one of these it is preferable to use as an entry point for the VS Code Extension documentation.

PR #392, opened 2020-09-29
Fix GitHub Actions workflows

turbolent:
I contacted GitHub support asking why the CI workflows for PR #385 are not run.

Turns out that the UI doesn't show that the jobs' status is "expected", but the cause is that they are actually never run, because the configuration does not specify that the workflow should be triggered for pull requests – the push event is only triggered for this repository, not for forks.


psiemens (approved):
LGTM

PR #393, opened 2020-09-30
Switch improvements

### LINKED ISSUE #391: Document switch statement
## Problem

In #365 switch statements were added. However, they still need to be documented.

## Definition of Done

- [ ] Switch statements are described in the documentation

turbolent:
Closes #391 

- Document switch statements
- Require switch case bodies to contain at least one statement
- Allow and handle `break` statements in switch statements

### DIFF
--- a/docs/language/control-flow.md
+++ b/docs/language/control-flow.md
@@ -11,7 +11,8 @@ If-statements allow a certain piece of code to be executed only when a given con
 The if-statement starts with the `if` keyword, followed by the condition,
 and the code that should be executed if the condition is true
 inside opening and closing braces.
-The condition must be boolean and the braces are required.
+The condition expression must be boolean.
+The braces are required and not optional.
 Parentheses around the condition are optional.
 
 ```cadence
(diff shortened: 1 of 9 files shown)


runtime/sema/check_while.go
```diff
@@ -118,7 +118,7 @@ func (checker *Checker) VisitBreakStatement(statement *ast.BreakStatement) ast.R
 
 	// check statement is inside loop
 
-	if !checker.inLoop() {
+	if !(checker.inLoop() || checker.inSwitch()) {
```
zhangchiqing, line 121:
```suggestion
	// break is only allowed to be inside of a for-loop or a switch-case,
	// otherwise return error
	if !(checker.inLoop() || checker.inSwitch()) {
```

docs/language/control-flow.md
```diff
@@ -112,6 +113,78 @@ if let number = noNumber {
...
+It ends the execution of the switch statement immediately
+and transfers control to the code after the switch statement
+
+### No Implicit Fallthrough
+
+Unlike switch statements in some other languages,
+switch statements in Cadence do not "fall through":
+execution of the switch statement finishes as soon as the block of code
+associated with the first matching case is completed.
+No explicit `break` statement is required.
```
zhangchiqing, line 174: Do we allow duplicated cases? what would happen for the following case?
```
y = 1
switch x {
case 1:
  y = y + 1
case 1:
  y = y + 2
default:
  y = y + 3
}
```
If we allow, what's the result of `y`?

docs/language/control-flow.md
```diff
@@ -112,6 +113,78 @@ if let number = noNumber {
...
+The block of code associated with a switch case may contain a `break` statement.
+It ends the execution of the switch statement immediately
+and transfers control to the code after the switch statement
+
+### No Implicit Fallthrough
+
+Unlike switch statements in some other languages,
+switch statements in Cadence do not "fall through":
+execution of the switch statement finishes as soon as the block of code
+associated with the first matching case is completed.
```
zhangchiqing, line 173: We could use an example to explain the no-implicit fallthrough

```
// word(2) => "The number is two"
fun word(_ n: Int): String {
    let str = "The number is:"
    switch n {
    case 1:
        str = str + "one"
    case 2:
        str = str + "two"    
    default:
        str = str + "other"
    }
    return str
}
```

Do we have a similar tests for this case?

zhangchiqing (approved):
Nice work! Looks good.

docs/language/control-flow.md
turbolent: Great question, I'll add that

runtime/sema/check_while.go
turbolent: Fixed in a2b0c9e34a5a4dc8fe37001eb828fb851d870312

docs/language/control-flow.md
turbolent: Described in 2b32abf9f1ce10e3ed836b12512c3173f75dae64

docs/language/control-flow.md
turbolent: Added a test and documentation in 8e3688bb5438c4fa32096263db2525f8b3cd92c5

turbolent:
@joshuahannan Could you please have a look at the new documentation?

joshuahannan (approved):
Looks great!

PR #399, opened 2020-10-06
Document iteration over dictionary entries

turbolent:
Developers might want to iterate over a dictionary's entries (keys and values).

However, Cadence does not support iterating over dictionaries with a for-in loop.

Document how this can still be done using existing functionality.

### DIFF
--- a/docs/language/control-flow.md
+++ b/docs/language/control-flow.md
@@ -293,7 +293,8 @@ be executed at all. Otherwise, the code will execute as many times
 as there are elements in the array.
 
 ```cadence
-var array = ["Hello", "World", "Foo", "Bar"]
+let array = ["Hello", "World", "Foo", "Bar"]
+
 for element in array {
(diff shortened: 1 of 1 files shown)


docs/language/control-flow.md
```diff
@@ -303,7 +304,24 @@ for element in array {
 // "World"
 // "Foo"
 // "Bar"
+```
 
+To iterate over a dictionary's entries (keys and values), 
+use a for-in loop over the dictionaries keys and get the value for each key:
```
turbolent, line 310:
```suggestion
use a for-in loop over the dictionary's keys and get the value for each key:
```

janezpodhostnik (approved):
Nice!

turbolent:
@joshuahannan same here, could you please have a look?

joshuahannan (approved):
👌

PR #400, opened 2020-10-06
Allow returning references from scripts

### LINKED ISSUE #358: Allow returning references from scripts
### Issue To Be Solved

It should be possible to return references from scripts, however, such programs are currently rejected.

### Suggested Solution

- [ ] Sema: Allow returning a reference from a script
- [ ] Runtime: return the value that the reference points to

turbolent:
Closes #358

- Check that transaction parameters are storable statically, instead of only at run-time (when the transaction is executed)
- Check that script return types are exportable. Allow references to be exported. Dereference if possible, and handle potential cycles

### DIFF
--- a/runtime/convertValues.go
+++ b/runtime/convertValues.go
@@ -50,6 +50,12 @@ func exportEvent(event exportableEvent) cadence.Event {
 
 type exportResults map[interpreter.Value]cadence.Value
 
+// exportValueWithInterpreter exports the given internal (interpreter) value to an external value.
+//
+// The export is recursive, the results parameter prevents cycles:
+// it is checked at the start of the recursively called function,
+// and pre-set before a recursive call.
+//
 func exportValueWithInterpreter(
 	value interpreter.Value,
 	inter *interpreter.Interpreter,
--- a/runtime/convertValues.go
+++ b/runtime/convertValues.go
@@ -31,117 +31,145 @@ import (
 
 // exportValue converts a runtime value to its native Go representation.
 func exportValue(value exportableValue) cadence.Value {
-	return exportValueWithInterpreter(value.Value, value.Interpreter())
+	return exportValueWithInterpreter(value.Value, value.Interpreter(), exportResults{})
 }
 
 // exportEvent converts a runtime event to its native Go representation.
 func exportEvent(event exportableEvent) cadence.Event {
 	fields := make([]cadence.Value, len(event.Fields))
 
+	results := exportResults{}
+
 	for i, field := range event.Fields {
-		fields[i] = exportValueWithInterpreter(field.Value, field.Interpreter())
+		fields[i] = exportValueWithInterpreter(field.Value, field.Interpreter(), results)
 	}
 
 	eventType := exportType(event.Type, map[sema.TypeID]cadence.Type{}).(*cadence.EventType)
 	return cadence.NewEvent(fields).WithType(eventType)
 }
 
-func exportValueWithInterpreter(value interpreter.Value, inter *interpreter.Interpreter) cadence.Value {
-	switch v := value.(type) {
-	case interpreter.VoidValue:
-		return cadence.NewVoid()
-	case interpreter.NilValue:
-		return cadence.NewOptional(nil)
-	case *interpreter.SomeValue:
-		return exportSomeValue(v, inter)
-	case interpreter.BoolValue:
-		return cadence.NewBool(bool(v))
-	case *interpreter.StringValue:
-		return cadence.NewString(v.Str)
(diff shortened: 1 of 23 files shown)


runtime/convertValues.go
```diff
@@ -31,117 +31,145 @@ import (
...
-		return exportStorageReferenceValue(v)
-	case interpreter.LinkValue:
-		return exportLinkValue(v, inter)
-	case interpreter.PathValue:
-		return exportPathValue(v)
-	case interpreter.TypeValue:
-		return exportTypeValue(v, inter)
-	case interpreter.CapabilityValue:
-		return exportCapabilityValue(v, inter)
+type exportResults map[interpreter.Value]cadence.Value
```
jordanschalm, line 51: What is this type for? I see it written to but not read from.

runtime/sema/errors.go
```diff
@@ -2303,6 +2303,22 @@ func (e *InvalidResourceTransactionParameterError) Error() string {
...
+// InvalidNonStorableTransactionParameterTypeError
+
+type InvalidNonStorableTransactionParameterTypeError struct {
+	Type Type
+	ast.Range
+}
+
+func (e *InvalidNonStorableTransactionParameterTypeError) Error() string {
+	return fmt.Sprintf(
+		"transaction parameter must not be storable: `%s`",
```
jordanschalm, line 2315: `must be storable` or `is not storable` ?

runtime/sema/type.go
```diff
@@ -84,6 +84,9 @@ type Type interface {
 	// e.g. in a field of a composite type.
 	IsStorable(results map[*Member]bool) bool
 
+	// IsExternallyReturnable returns true if a value of this type can be exported
+	IsExternallyReturnable(results map[*Member]bool) bool
```
jordanschalm, line 88: What is the input used for? Does it represent members of the type implementing this interface?

runtime/tests/checker/interface_test.go
```diff
@@ -1363,52 +1363,74 @@ func TestCheckInterfaceWithFieldHavingStructType(t *testing.T) {
...
+				default:
+					errs := ExpectCheckerErrors(t, err, 1)
+
+					assert.IsType(t, &sema.InvalidResourceFieldError{}, errs[0])
+				}
+
+			case common.CompositeKindContract:
+				switch secondKind {
+				case common.CompositeKindResource,
+					common.CompositeKindStructure:
```
jordanschalm, line 1414: This is saying that a contract can be a field in a resource or structure? That doesn't make sense to me -- I thought contracts were always the top-level unit of code.

runtime/convertValues.go
turbolent: The type is used for the "results" parameter of `exportValueWithInterpreter`, and read from at the top of the function call: https://github.com/onflow/cadence/pull/400/files#diff-0f63ad66a8e46ed831efe45b45a08d6aR59.

As the export is recursive, the results parameter prevents cycles.

runtime/sema/errors.go
turbolent: Good catch, this was copy and pasted from above (`InvalidResourceTransactionParameterError`) and I rephrased incorrectly 👍 

```suggestion
		"transaction parameter must be storable: `%s`",
```

runtime/sema/type.go
turbolent: It's the same as for `exportValueWithInterpreter`: The check if the type is externally returnable is recursive, the "results" parameter prevents cycles: it is checked at the start of the recursively called function, and pre-set before a recursive call.

I'll add comments where this pattern is used to clarify.

turbolent: Added comments in 04c162c757f2a6ea4de1d515d36c092252ec45a7

runtime/tests/checker/interface_test.go
turbolent: Contracts are just objects, just like structs, and resources. 
They can be stored in variables, passed to functions, and also put into a struct or resource as a field.

Now, this may seem odd, but could be useful when using interfaces.
For example, a function could return a result object which has a field that has a contract interface type, indicating to the caller what contract should be called next.

Note that this has nothing to do with declaration nesting: contracts can still not be declared inside resources or structs, this is purely about fields inside containers, i.e. your statement " "contracts [are] always the top-level unit of code" is still true.

runtime/sema/type.go
jordanschalm: Thanks for adding the comments 🙏

jordanschalm (approved):
Nice work 👌

Issue #369, opened 2020-09-10
Typed Paths


turbolent:
### Issue To Be Solved

Currently, all paths have the type `Path` and the `getCapability` functions accepts an argument of type `Path` and returns an optional capability if the argument has the correct domain,

However, this run-time check is confusing and can be moved to a static check by typing paths based on domain, and changing the parameter type of the `getCapability` functions.

### Definition of Done

- Sema:
  - [ ] Introduce `StoragePath`, `CapabilityPath`, `PublicPath`, and `PrivatePath`
  - [ ] Make `PublicPath`, and `PrivatePath` a subtype of `CapabilityPath`
  - [ ] Change the type of path literals, based on the domain
  - [ ] Change the `getCapability` functions to accepts the appropriate path type
  - [ ] Change the `getCapability` functions to return a non-optional type
- [ ] Interpreter: Change the `getCapability` functions to return a non-optional value
- [ ] Tests
- [ ] Documentation

PR #401, opened 2020-10-07
Refactor types

turbolent:
Work towards #369 

Adding new types in the checker currently involves a lot of boilerplate. 
Before adding more types, introduce a simple "nominal type" type, and refactor some types to using it:
- Void
- Never
- Path
- Storable

In addition to reducing boilerplate code and making the future addition of more types easier, this also  reduces allocations and simplifies type checks (Go equality checks instead of Go type checks).

### DIFF
--- a/runtime/sema/path_type.go
+++ b/runtime/sema/path_type.go
@@ -24,6 +24,7 @@ var PathType = &NominalType{
 	Name:          "Path",
 	QualifiedName: "Path",
 	TypeID:        "Path",
+	IsInvalid:     false,
 	IsResource:    false,
 	Storable:      true,
 	// TODO: implement support for equating paths in the future
--- a/runtime/sema/path_type.go
+++ b/runtime/sema/path_type.go
@@ -18,25 +18,16 @@
 
 package sema
 
-import (
-	"github.com/onflow/cadence/runtime/ast"
-	"github.com/onflow/cadence/runtime/common"
-)
+// PathType
 
-func (checker *Checker) VisitPathExpression(expression *ast.PathExpression) ast.Repr {
-
-	// Check that the domain is valid
-
-	domain := expression.Domain
-
-	if _, ok := common.AllPathDomainsByIdentifier[domain.Identifier]; !ok {
-		checker.report(
-			&InvalidPathDomainError{
-				ActualDomain: domain.Identifier,
-				Range:        ast.NewRangeFromPositioned(domain),
-			},
-		)
-	}
-
-	return PathType
+var PathType = &NominalType{
+	Name:          "Path",
+	QualifiedName: "Path",
+	TypeID:        "Path",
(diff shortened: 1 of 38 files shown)


runtime/sema/storable_type.go
```diff
@@ -0,0 +1,45 @@
...
+	Name:          "Storable",
+	QualifiedName: "Storable",
+	TypeID:        "Storable",
+	// NOTE: Subtypes may be either resource types or not.
+	//
+	// Returning false here is safe, because this type is
+	// only used as e.g. a type bound, but is not accessible
+	// to user programs, i.e. can't be used in type annotations
+	// for e.g. parameters, return types, fields, etc.
+	IsResource: false,
```
jordanschalm, line 37: `IsInvalid` field is missing here, better to include and be explicit

runtime/sema/path_type.go
```diff
@@ -0,0 +1,32 @@
...
+
+var PathType = &NominalType{
+	Name:          "Path",
+	QualifiedName: "Path",
+	TypeID:        "Path",
+	IsResource:    false,
+	Storable:      true,
+	// TODO: implement support for equating paths in the future
+	Equatable:     false,
+	IsSuperTypeOf: nil,
```
jordanschalm, line 31: `IsInvalid` field is missing here, better to include and be explicit

jordanschalm (approved):
👍 Good idea

runtime/sema/path_type.go
```diff
@@ -0,0 +1,32 @@
...
+
+package sema
+
+// PathType
+
+var PathType = &NominalType{
+	Name:          "Path",
+	QualifiedName: "Path",
+	TypeID:        "Path",
+	IsResource:    false,
```
turbolent, line 27:
```suggestion
	IsInvalid:     false,
	IsResource:    false,
```

runtime/sema/storable_type.go
```diff
@@ -0,0 +1,45 @@
...
+// StorableType is the supertype of all types which are storable.
+//
+// It is only used as e.g. a type bound, but is not accessible
+// to user programs, i.e. can't be used in type annotations
+// for e.g. parameters, return types, fields, etc.
+//
+var StorableType = &NominalType{
+	Name:          "Storable",
+	QualifiedName: "Storable",
+	TypeID:        "Storable",
```
turbolent, line 30:
```suggestion
	TypeID:        "Storable",
	IsInvalid:     false,
```

runtime/sema/storable_type.go
turbolent: Good point, added in 68c4340

runtime/sema/path_type.go
turbolent: Good point, added in 68c4340

PR #404, opened 2020-10-09
Switch back to upstream CBOR library

turbolent:
fxamacker/cbor#249 was merged, switch back to it instead of using a fork


jordanschalm (approved):
🎸

PR #406, opened 2020-10-10
Add language server command to get entry point parameters

### LINKED ISSUE #405: Language Server: Add command to get transaction and script parameters
### Issue To Be Solved
For Playground and VS Code we need to be able to get list parameters that we want to pass to transactions and scripts.
It's possible to make up a solution that will extract those with a bunch of regexps, but since language server parses the code anyway - we can utilize it.

### (Optional): Extra
Would be also great if we can extract parameters for `init` method of the contract 😊

turbolent:
Closes #405 

- Refactor the existing entry point analysis (transaction declaration or script function declaration) from the language server to the AST and sema. It could be useful in other tools, and having it in a central place ensures the assumptions won't get out of sync.
- Add a function to sema which returns the parameters of the entry point 
- Add a command to the language server which returns the parameters of the entry point, leveraging the function in sema. This allow e.g. the VS Code extension and Playground, which use the language server, to request this information
- Add a test suite for the language server
- Add a test for the new command


languageserver/integration/diagnostics.go
```diff
@@ -36,47 +37,68 @@ func (i *FlowIntegration) diagnostics(
...
+
 		isFirst := true
-		for declaration := range elaboration.TransactionDeclarationTypes {
+
+		for _, declaration := range transactionDeclarations {
 			// Skip the first declaration
 			if isFirst {
 				isFirst = false
 				continue
 			}
```
jordanschalm, line 53:
```suggestion
		for _, declaration := range transactionDeclarations[1:] {
```

runtime/sema/entrypoint.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+	}
+
+	functionType := invokableType.InvocationFunctionType()
+	return functionType, nil
+}
+
+// FunctionEntryPointDeclaration returns the entry point function declaration, if any.
+//
+// Returns nil if there are multiple function declarations with the same function entry point name.
```
jordanschalm, line 53: (or any other top-level declarations)

jordanschalm (approved):
👏 Nice work

runtime/sema/entrypoint.go
turbolent, line 53:
```suggestion
// Returns nil if there are multiple function declarations with the same function entry point name, or any other top-level declarations.
```

languageserver/integration/diagnostics.go
```diff
@@ -36,47 +37,63 @@ func (i *FlowIntegration) diagnostics(
...
+		compositeAndInterfaceDeclarations = append(compositeAndInterfaceDeclarations, compositeDeclaration)
+	}
+
+	for _, interfaceDeclaration := range checker.Program.InterfaceDeclarations() {
+		compositeAndInterfaceDeclarations = append(compositeAndInterfaceDeclarations, interfaceDeclaration)
+	}
+
+	if len(compositeAndInterfaceDeclarations) > 1 {
+
 		isFirst := true
```
turbolent, line 78:
```suggestion
```

PR #408, opened 2020-10-12
Interpreter: Always find the declared variable

turbolent:
Minor: There is no need to declare a variable, always just find the variable

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -547,14 +547,6 @@ func (interpreter *Interpreter) findVariable(name string) *Variable {
 	return result.(*Variable)
 }
 
-func (interpreter *Interpreter) findOrDeclareVariable(name string) *Variable {
-	variable := interpreter.findVariable(name)
-	if variable == nil {
-		variable = interpreter.declareVariable(name, nil)
-	}
-	return variable
-}
-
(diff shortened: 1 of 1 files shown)


runtime/interpreter/interpreter.go
```diff
@@ -881,7 +873,10 @@ func (interpreter *Interpreter) VisitFunctionDeclaration(declaration *ast.Functi
 
 	functionType := interpreter.Checker.Elaboration.FunctionDeclarationFunctionTypes[declaration]
 
-	variable := interpreter.findOrDeclareVariable(identifier)
+	variable := interpreter.findVariable(identifier)
+	if variable == nil {
+		panic(errors.NewUnreachableError())
+	}
```
SupunS, line 879: There seems to be other places where the `interpreter.findVariable()` is called and the value returned is directly used to access members etc.

Would it be a better option to move the panic to the `interpreter.findVariable()` itself?

turbolent: I think the other cases, e.g. https://github.com/onflow/cadence/blob/c46d71cf798dfeae3fef86a68aa69b10106175bb/runtime/interpreter/interpreter.go#L1555, should be  fine, as in the interpreter we can assume the reference to a variable/identifier will always succeed, because the checker rejected invalid references before

SupunS (approved):

PR #409, opened 2020-10-12
Start of a compiler, with IR and WASM code generator

turbolent:
- Add IR package
- Add compiler which can translate AST + elaboration to IR
- Add code generator which can translate IR to WASM binary
- Add CLI command to parse, check, and compile a program to a WASM binary

Demo:

```shell
$ echo "pub fun inc(a: Int): Int { let b = 1; return a + b }" | \
  go run ./runtime/cmd/compile /dev/stdin | \
  wasmtime /dev/stdin --invoke inc 1

2
```


SupunS (approved):
Looks great!

PR #411, opened 2020-10-13
Transaction argument parser 

psiemens:
## Description

This PR demonstrates a simple Cadence value parser than can be used to parser standalone Cadence arguments (e.g. in the CLI, language server or Playground API).


turbolent:
This is great, thank you for opening this PR @psiemens! 

In #406 I introduced `Checker.EntryPointParameters`, so that could be used instead of `transactionParameterTypes`.

Also, I think it would be great to move the `new*` functions, which convert AST literals to values, into e.g. the interpreter or a new package, and merge it with the existing code in the interpreter that already does this.

Again, we could expose this as a language server command and call it in the Playground and VS Code extension 👍

psiemens:
Those are both great ideas! I probably won't have time to make those changes this week, but please feel free to modify this PR in any way.

turbolent:
@psiemens Thanks again for opening this PR and the inspiration! I've opened #417, which should implement the same functionality, reusing as much existing code as possible.

psiemens:
Thanks @turbolent! I'll close this PR.

PR #415, opened 2020-10-15
Prefix NPM package names with onflow organization

turbolent:
The packages are going to be published under the NPM organization `onflow`, so prefix all package names and adjust the imports


psiemens (approved):
LGTM

PR #417, opened 2020-10-16
Add a function to parse a literal with a given target type

### LINKED ISSUE #416: Allow parsing literals to values
### Issue To Be Solved

Tools have the need to accept user input and convert it into values, for example to provide script or transaction arguments.

### Suggest A Solution

Add a new function which takes code and an expected type, and which returns a Cadence value, e.g `func ParseLiteral(literal string, ty Type) (Value, error)`

turbolent:
Closes #416 

Also, add a language server command that parses entry point arguments and returns them as values.
This allows them to be included as transaction arguments.

cc @MaxStalker

### DIFF
--- a/runtime/literal_test.go
+++ b/runtime/literal_test.go
@@ -1,0 +1,534 @@
+/*
+ * Cadence - The resource-oriented smart contract programming language
+ *
+ * Copyright 2019-2020 Dapper Labs, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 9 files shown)


runtime/literal_test.go
```diff
@@ -0,0 +1,534 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/onflow/cadence"
+	"github.com/onflow/cadence/runtime/sema"
+)
+
+func TestLiteralValue(t *testing.T) {
```
jordanschalm, line 31: 👏 Great work with these

jordanschalm (approved):
LGTM

PR #422, opened 2020-10-20
Export component making composite types IDs as separate fields

### LINKED ISSUE #420: Add `EventType.TypeID` elements
### Issue To Be Solved

Event typeIDs is constructed from elements like account address, contract name and other. They are transformed into one string. It would be much easier to filter and query events if those elements were emitted as separate elements,

###  Suggest A Solution

Either create new structure containing those fields or just add them beside (or instead) `TypeID` to `EventType` - the current `TypeID` can be recomputed from elements,

m4ksio:
Closes #420

## Description

Adds separate components making composite type ID exported.
For now kept original `TypeID` for compatibility.

______

For contributor use:

### DIFF
--- a/encoding/json/decode.go
+++ b/encoding/json/decode.go
@@ -89,19 +89,22 @@ func (d *Decoder) Decode() (value cadence.Value, err error) {
 }
 
 const (
-	typeKey       = "type"
-	valueKey      = "value"
-	keyKey        = "key"
-	nameKey       = "name"
-	fieldsKey     = "fields"
-	idKey         = "id"
-	targetPathKey = "targetPath"
-	borrowTypeKey = "borrowType"
-	domainKey     = "domain"
-	identifierKey = "identifier"
-	staticTypeKey = "staticType"
-	addressKey    = "address"
-	pathKey       = "path"
+	typeKey                    = "type"
+	valueKey                   = "value"
+	keyKey                     = "key"
+	nameKey                    = "name"
+	fieldsKey                  = "fields"
+	idKey                      = "id"
+	compositeTypeKey           = "compositeTypeID"
+	compositeTypeIdentifierKey = "identifier"
+	compositeTypeLocationKey   = "location"
+	targetPathKey              = "targetPath"
+	borrowTypeKey              = "borrowType"
+	domainKey                  = "domain"
+	identifierKey              = "identifier"
+	staticTypeKey              = "staticType"
+	addressKey                 = "address"
+	pathKey                    = "path"
 )
 
 var ErrInvalidJSONCadence = errors.New("invalid JSON Cadence structure")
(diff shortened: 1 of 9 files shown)


turbolent (commented):
Do we have to add a new structured type ID to all types? Would it be enough to just add a function that decodes the information from the existing type ID?

m4ksio:
> Do we have to add a new structured type ID to all types? Would it be enough to just add a function that decodes the information from the existing type ID?

It's probably better from architectural point of view to have separate components in a defined structure, but I'm not sure about compatibility implications though

turbolent:
@m4ksio does the structured type identifier be added to the JSON representation? I’m a bit worried that will blow up the size of values quite a bit

@psiemens @orodio I guess changing the unstructured type identifier to a structured identifier would not be backwards compatible, but we could add a function/getter that concatenates the components of the structured type identifier and returns the unstructured one, like before?

turbolent:
@psiemens @orodio Talked to @m4ksio again, this is blocking other issues. The plan is to change the type ID from the old unstructured from (string) to the structured form (object), instead of adding/increasing size. This is a breaking change and will effect the SDKs

encoding/json/decode.go
```diff
@@ -691,6 +701,18 @@ func toObject(valueJSON interface{}) jsonObject {
 	return v
 }
 
+func compositeTypeIDFromObject(valueJSON interface{}) cadence.CompositeTypeID {
+	object := toObject(valueJSON)
+
+	id := compositeTypeIdentifierKey
+	object.GetString(id)
```
psiemens, line 708: Is this to force a panic?

types.go
```diff
@@ -559,10 +570,15 @@ func (t *ResourceType) CompositeInitializers() [][]Parameter {
 	return t.Initializers
 }
 
-// EventType
+type CompositeTypeID struct {
+	Location   string
+	Identifier string
+}
```
psiemens, line 576: Do you think it would make sense to have a separate `LocationKind` field that is one of `script`, `transaction`, `account`? This would replace the `S.`, `T.`, `A.` prefixes.

If we're splitting the type into pieces, why not split it all the way?

encoding/json/decode.go
m4ksio: Not sure I get what you mean?

types.go
m4ksio: That's very good idea! I started with fields I needed for other tasks, counting on feedback like this from people with more knowledge.

encoding/json/decode.go
psiemens: The value returned from `object.GetString(id)` is never used, so at first I thought these two lines might be redundant, but then I thought you might be checking for the existence of `compositeTypeIdentifierKey`. `GetString` will panic if it's not found

types.go
turbolent: @psiemens @m4ksio do you have any preference for modelling this either as 
- a "tagged struct": define a struct `TypeID`, which has a `LocationKind` field, and a constant for each location kind
- or as an interface `TypeID`, and dedicated structs for each location kind, e.g. `ScriptTypeID`, `AccountTypeID`, etc. (how e.g. `ast.Location` is declared at the moment)

psiemens: @turbolent I think I prefer a simple struct in this case, therefore no reflection required.

m4ksio: Agree, I don't think we need more polymorphism here

turbolent:
I've opened #477, OK if I close this one?

turbolent:
Closing this as #477 is now merged

PR #424, opened 2020-10-22
Add position information to interpreter errors

turbolent:
Work towards dapperlabs/flow-go#2426

Improve fatal interpreter errors by wrapping them with position information.

- Before:

  Screen Shot 2020-10-22 at 11 39 57 AM

- After:

  Screen Shot 2020-10-22 at 11 37 15 AM

### DIFF
--- a/runtime/cmd/cmd.go
+++ b/runtime/cmd/cmd.go
@@ -45,38 +45,45 @@ func PrettyPrintError(writer io.Writer, err error, filename string, codes map[st
 		i++
 	}
 
-	if parserError, ok := err.(parser2.Error); ok {
-		for _, err := range parserError.Errors {
+	switch typedErr := err.(type) {
+	case parser2.Error:
+		for _, err := range typedErr.Errors {
 			printErr(err, filename)
 		}
-	} else if checkerError, ok := err.(*sema.CheckerError); ok {
-		for _, err := range checkerError.Errors {
+	case *sema.CheckerError:
+		for _, err := range typedErr.Errors {
 			printErr(err, filename)
 			if err, ok := err.(*sema.ImportedProgramError); ok {
-				var filename string
-				switch importLocation := err.ImportLocation.(type) {
-				case ast.StringLocation:
-					filename = string(importLocation)
-				case ast.AddressLocation:
-					filename = importLocation.ToAddress().ShortHexWithPrefix()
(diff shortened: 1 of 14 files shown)


runtime/tests/interpreter/capability_test.go
```diff
@@ -319,6 +343,12 @@ func TestInterpretCapability_borrow(t *testing.T) {
 			_, err := inter.Invoke("loop")
 			require.Error(t, err)
 
+			require.IsType(t,
+				interpreter.Error{},
+				err,
+			)
+			err = err.(interpreter.Error).Unwrap()
+
 			require.IsType(t, interpreter.CyclicLinkError{}, err)
```
jordanschalm, line 352: I think you can use `errors.As` to avoid the explicit unwrapping step in all these checks. `As` recursively unwraps the error and checks whether it can be assigned to a given type which should work here. Something like:
```go
require.True(t, errors.As(err, &interpreter.CyclicLinkError{})
```

jordanschalm (approved):

runtime/cmd/cmd.go
```diff
@@ -45,40 +45,47 @@ func PrettyPrintError(writer io.Writer, err error, filename string, codes map[st
...
+			filename = importLocationFileName(location)
 		}
 		printErr(err, filename)
-	} else {
+	default:
 		printErr(err, filename)
 	}
 }
 
+func importLocationFileName(importLocation ast.Location) string {
```
psiemens, line 75: Should `ast.Location` just declare a `FileName()` method?

psiemens (approved):
👏

runtime/tests/interpreter/capability_test.go
turbolent: Great suggestions! I wanted to have a better way, this is great 👌 Refactored as many cases as I could in 98771cb.

I checked tesitify and they'll have a `ErrorAs` assertion function soon (https://github.com/stretchr/testify/commit/95a9d909e98735cd8211dfc5cbbb6b8b0b665915), so for now I added a helper that does the same and added a TODO to replace it once testify has released their function.

runtime/cmd/cmd.go
turbolent: Not all locations have a filename so I didn't opt for that

PR #430, opened 2020-10-28
WebAssembly: Add support for start section

turbolent:
Also, only write sections if they have any content


runtime/compiler/wasm/reader.go
```diff
@@ -158,6 +158,16 @@ func (r *WASMReader) readSection() error {
 			return err
 		}
 
+	case sectionIDStart:
+		if r.Module.StartFunctionIndex != nil {
```
zhangchiqing, line 162: Why not moving this check to `readStartSection`?

So that this case will become easier to test.

zhangchiqing (approved):
Looks good. Suggested some idea to verify results

runtime/compiler/wasm/writer_test.go
```diff
@@ -594,21 +633,26 @@ func TestWASMWriterReader(t *testing.T) {
 		0x1, 0x0, 0x0, 0x0,
 		// type section
 		0x1,
-		0x87, 0x80, 0x80, 0x80, 0x0,
-		0x1, 0x60, 0x2, 0x7f, 0x7f, 0x1, 0x7f,
+		0x8a, 0x80, 0x80, 0x80, 0x0,
```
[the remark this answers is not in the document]
turbolent: > I've been seen some tests that include these human-unreadable form of results as the expected result to compare with the actual result in a test case.

Right, the tests for the sections have annotations to help debug test failures.

> I feel the process of generating them was simply taking the output the actual result here. I guess it's not from you handwriting these bytes.

Only in this particular test (`TestWASMWriterReader`), which is an integration test.
If this test fails, there must already be a test failure in the unit test for the section.

> I think the tests created from this process is useful as a regression tests. But it's not easy for the reviewer or even the author to know if they are really correct.
>
> So the question is if we are testing a function f, which outputs raw bytes, how to verify the output is correct?

Yes, the main purpose of this test is avoiding regressions. The byte output is not supposed to be manually reviewed, that is why I added the WAT format, which is the human-readable representation and use [Wabt](https://github.com/WebAssembly/wabt)'s `wasm2wat` to validate the binary. Given that the tool also validates the binary while translating from WASM (binary format) to WAT (text format), a reviewer of the test can ensure the byte array is valid.

> Compare the outputs with each other

Do you have an example for this? I'm not quite sure I follow

> Integration tests
> The raw bytes is not readable, but if we have a function that can read it so that we can verify by creating a tests to check assert.Equal(data, read(write(data))) without looking at the output.
> Or we could some external tool/library that is able to load the output bytes and see if things work as expected. This also indirectly verifies the output.

This is basically what this function does: An example module is prepared, then written, then the output read, and the read result is compared to the original input, the prepared module.
In addition, like mentioned above, the function uses `wasm2wat` to verify that the written output  is valid, to make sure that not both the writer and reader are broken in a way that still makes the test pass, even though the written output it invalid.

runtime/compiler/wasm/reader.go
turbolent: I structured the code in a way that each section reader function only handle reading a section, independent of the whole module structure, and then have this function ensure there are no duplicates.

PR #434, opened 2020-11-01
Implement fmt.Stringer for cadence.Value

### LINKED ISSUE #433: Printing Fix and UFix or optional values of them are confusing
## Instructions

Right now if you create an UFix64(64.01) is it printed as 6401000000 and not 64.010000. 

### Issue To Be Solved

Print the Ufix/Fis numbers in a human readable way.

Would it be possible to use the encode method here https://github.com/onflow/cadence/blob/master/encoding/json/encode.go for the string value of a UFix/Fix value?

### Why
I am creating some code to pretty print an cadence.Event to send it to a discord webhook.

bjartek:
Closes #433 

## Description

Started on adding Stringer methods

### DIFF
--- a/values.go
+++ b/values.go
@@ -165,7 +165,7 @@ func (v Bytes) ToGoValue() interface{} {
 }
 
 func (v Bytes) String() string {
-	return string(v[:])
+	return format.Bytes(v)
 }
 
 // Address
--- a/values.go
+++ b/values.go
@@ -1172,8 +1172,7 @@ func (Path) ToGoValue() interface{} {
 }
 
 func (v Path) String() string {
-	return fmt.Sprintf(
-		"/%s/%s",
+	return format.Path(
 		v.Domain,
 		v.Identifier,
 	)
--- a/values.go
+++ b/values.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 
 	"github.com/onflow/cadence/fixedpoint"
+	"github.com/onflow/cadence/runtime/common"
 	"github.com/onflow/cadence/runtime/format"
 	"github.com/onflow/cadence/runtime/interpreter"
 )
@@ -191,7 +192,7 @@ func (v Address) Bytes() []byte {
 }
 
 func (v Address) String() string {
-	return v.Hex()
+	return format.Address(common.Address(v))
 }
 
 func (v Address) Hex() string {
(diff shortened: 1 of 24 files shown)


bjartek:
Most of these are copied from interpeter like @turbolent suggested in Discord.

values.go
```diff
@@ -33,6 +36,7 @@ type Value interface {
 	isValue()
 	Type() Type
 	ToGoValue() interface{}
+	String() string //not really sure if this is needed?
```
turbolent, line 39: We can re-use the existing standard interface `fmt.Stringer` here:

```suggestion
	fmt.Stringer
```

turbolent (commented):
@bjartek This looks good, thank you for adding this! Thank you also for adding tests	

Now looking at the changes, I've realized it might be a good idea to actually create a package for the formatting of values (maybe `runtime/format`?), and using it both for `interpreter.Value`s and here for `cadence.Value`s. This would prevent future inconsistencies, where e.g. a fix is only applied on of the format functions, but not thee other related one.

I'm happy to do this if you want, let me know what you think.

values_test.go
```diff
@@ -9,6 +9,123 @@ import (
...
+		"address":       StringTestCase{value: NewAddress([8]byte{0, 0, 0, 0, 0, 0, 0, 1}), expected: "0000000000000001"},
+		"struct":        StringTestCase{value: fooStruct, expected: "Foo:{y: bar}"},
+		"resource":      StringTestCase{value: resource, expected: "FooResource:{bar: 1}"},
+		"event":         StringTestCase{value: event, expected: "FooEvent:{a: 1, b: foo}"},
+		"link":          StringTestCase{value: link, expected: "Link(type: Bar, targetPath: /storage/foo)"},
+		"path":          StringTestCase{value: path, expected: "/storage/foo"},
+		"type":          StringTestCase{value: TypeValue{StaticType: "Int"}, expected: "Type<Int>"},
+		"capability":    StringTestCase{value: cap, expected: "Capability<Int>(/0000000102030405/storage/foo)"},
+	}
+
```
turbolent, line 119: It would be good to iterate over all names in `sema.baseTypes` and checking that there exists a test case for it. This would prevent the case in the future where a new type/value is added, but the author forgets to add a test case for it here. This might require refactoring the tests, e.g. grouping them by type.

bjartek: Good call.

bjartek: Can you have a look at how i tried solving this in the two latest commits?

bjartek:
Feel free to move the code/tests to runtime/format.

turbolent:
Can you please allow changes to the PR? 
See https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

That would allow me to commit the remaining work. Thank you!

bjartek:
Allow edit by maintainers should be on now.

turbolent:
@bjartek Thanks again for your work! I refactored the formatting code to common functions in a separate new package `format`, and changed the `cadence.Value`s and `interpreter.Value`s to use them. I also duplicated the new tests for the interpreter values.

values_test.go
turbolent: When working on the refactoring I realized that the check actually isn't really possible, because not all base types have corresponding values in the `cadence` package. I removed the check, it's OK if we manually ensure tests for now

bjartek:
Great refactoring!

janezpodhostnik (approved):
Very nice addition! LGTM.

turbolent (approved):
🚀

PR #446, opened 2020-11-06
Don't use the cache when deploying or updating account code

### LINKED ISSUE #445: AST cache is used when updating account code
### Problem

When updating the account code, the program cache is read, but it shouldn't.

### Steps to Reproduce

- Update contract code with a program cache
- Attempt to use the updated code
- Observe how the old code is returned, the code was not updated

### Acceptance Criteria

- Code updates work correctly when a cache is active

### Context

I introduced a regression while implementing multiple contracts per account: We previously had the issue where the cache was used when checking the program update, though it should not: https://github.com/onflow/cadence/commit/63044233f9fb799d1218d4f103bac9db80b1bf96. 

When I implemented multiple contracts per account I used the exported function again which uses the cache: https://github.com/onflow/cadence/blob/d114710e6759326b0d7bbeaf41f6c70ddefdef62/runtime/runtime.go#L1386

The problem is that the tests in Cadence don't use the cache. I'll push a fix and will try to run the code management tests also with a cache enabled. I'll release v0.10.1 of Cadence once merged.

turbolent:
Closes #445 

This is a regression that was introduced in v0.10.0, so I'll release this as v0.10.1.
This will allow me to continue work on https://github.com/onflow/flow-emulator/pull/2#issuecomment-722715987.

I'll open another PR to apply this to master.

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -1404,7 +1404,14 @@ func (r *interpreterRuntime) newAuthAccountContractsChangeFunction(
 				AddressLocation: addressValue[:],
 				Name:            nameArgument,
 			}
-			checker, err := r.ParseAndCheckProgram(code, runtimeInterface, location)
+
+			// NOTE: do NOT use the cache!
+
+			const useCache = false
+
+			functions := r.standardLibraryFunctions(runtimeInterface, runtimeStorage)
+
+			checker, err := r.parseAndCheckProgram(code, runtimeInterface, location, functions, nil, useCache)
 			if err != nil {
 				panic(fmt.Errorf("invalid contract: %w", err))
 			}
--- a/runtime/tests/utils/utils.go
+++ b/runtime/tests/utils/utils.go
@@ -94,3 +94,18 @@ func DeploymentTransaction(name string, contract []byte) []byte {
 		hex.EncodeToString(contract),
 	))
 }
+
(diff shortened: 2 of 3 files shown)


runtime/runtime.go
```diff
@@ -1404,7 +1404,14 @@ func (r *interpreterRuntime) newAuthAccountContractsChangeFunction(
 				AddressLocation: addressValue[:],
 				Name:            nameArgument,
 			}
-			checker, err := r.ParseAndCheckProgram(code, runtimeInterface, location)
+
+			// NOTE: do NOT use the cache!
+
+			const useCache = false
```
psiemens, line 1410: If only Go had argument labels 😄

runtime/tests/utils/utils.go
```diff
@@ -94,3 +94,18 @@ func DeploymentTransaction(name string, contract []byte) []byte {
 		hex.EncodeToString(contract),
 	))
 }
+
+func UpdateTransaction(name string, contract []byte) []byte {
```
psiemens, line 98:
```suggestion
func ContractUpdateTransaction(name string, contract []byte) []byte {
```

Just a small suggestion, feel free to ignore :)

psiemens (approved):
Great tests 👍 💯

janezpodhostnik (approved):

PR #448, opened 2020-11-09
WebAssembly: Use reference types

turbolent:
- Import runtime functions for constructing `Int` and `String` values, and adding two number values.
  Re-use interpreter values
- Allocate constants in memory, export so host environment can access it
- Use [wasmtime](https://github.com/bytecodealliance/wasmtime) to execute
- Compile string literals


turbolent:
@SupunS could you please have a look at this last compiler PR? @janezpodhostnik also mentioned he had interest in understanding, if you want we can have a quick call and I can go over it

SupunS:
@turbolent Sure, I will have a look. A walk-through of the code would be great.

vm/vm.go
```diff
@@ -0,0 +1,141 @@
...
+		return leftNumber.Plus(rightNumber), nil
+	})
+
+	instance, err := wasmtime.NewInstance(
+		store,
+		module,
+		[]*wasmtime.Extern{
+			intFunc.AsExtern(),
+			stringFunc.AsExtern(),
+			addFunc.AsExtern(),
```
SupunS, line 129: out of curiosity: when the imports are added to these extern functions (referring to [addRuntimeImports](https://github.com/onflow/cadence/pull/448/files#diff-1f6a4a5ec36bcf2c32f34ef4341b81af60a52fe5ebb2eb06981d5ed3bf2fbd4eR209-R213)), how does it map that import to the relevant function definition? I guess its by the order defined here and  [there](https://github.com/onflow/cadence/pull/448/files#diff-1f6a4a5ec36bcf2c32f34ef4341b81af60a52fe5ebb2eb06981d5ed3bf2fbd4eR209-R213)? (and both places needs to ensure the same order?)

SupunS (approved):

vm/vm.go
turbolent: The mapping is by index, which is not great. You're right, the code generator declares imports, and wasmtime expects them to be passed in in the same order. 

I found this odd, because inside the WebAssembly module the imports are also named, and other WebAssembly APIs, [like the JavaScript one available in browsers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/instantiate), provide means to pass imports by name.

Maybe wasmtime will also provide support for passing imports by name and matching them with the module's imports at some point

PR #449, opened 2020-11-11
Add a visitor for interpreter values

turbolent:
Work towards #84

Add a visitor which allows traversing a whole interpreter object tree. This can e.g. be used for data migration.

### DIFF
--- a/runtime/ast/import_test.go
+++ b/runtime/ast/import_test.go
@@ -92,6 +92,30 @@ func TestAddressLocation_MarshalJSON(t *testing.T) {
 	)
 }
 
+func TestAddressContractLocation_MarshalJSON(t *testing.T) {
+
+	t.Parallel()
+
+	loc := AddressContractLocation{
+		AddressLocation: AddressLocation([]byte{1}),
+		Name:            "A",
+	}
+
+	actual, err := json.Marshal(loc)
+	require.NoError(t, err)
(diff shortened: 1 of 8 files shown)


runtime/ast/import_test.go
```diff
@@ -86,6 +86,30 @@ func TestAddressLocation_MarshalJSON(t *testing.T) {
 	)
 }
 
+func TestAddressContractLocation_MarshalJSON(t *testing.T) {
```
janezpodhostnik, line 89: This isn't related to this PR right?

janezpodhostnik (approved):
LGTM 🚀

runtime/ast/import_test.go
turbolent: Good catch! No, this test doesn't belong to this PR, I accidentally committed it. I'll remove

runtime/ast/import_test.go
```diff
@@ -86,6 +86,30 @@ func TestAddressLocation_MarshalJSON(t *testing.T) {
...
+        {
+            "Type": "AddressContractLocation",
+            "Address": "0x1",
+            "Name": "A"
+        }
+        `,
+		string(actual),
+	)
+}
+
```
turbolent, line 112:
```suggestion
```

PR #451, opened 2020-11-12
Extended debug

m4ksio:
Part of changes to support dapperlabs/flow-go#1291

It's not the most elegant solution, but I assume this code is temporary anyway.
It supports returning extended information on case of execution errors so they can be recorded by the caller - currently if TopShot contract fails.
It was started from v0.9.2 branch as it was the latest one we use in Flow-go - I plan to update this PR accordingly once more mature version of Cadence is used in Flow-go

### DIFF
--- a/runtime/errors.go
+++ b/runtime/errors.go
@@ -26,7 +26,6 @@ import (
 	"github.com/onflow/cadence/runtime/errors"
 	"github.com/onflow/cadence/runtime/interpreter"
 	"github.com/onflow/cadence/runtime/sema"
-	"github.com/onflow/cadence/runtime/stdlib"
 )
 
 // Error is the containing type for all errors produced by the runtime.
@@ -223,15 +222,14 @@ func (e *TransactionParameterTypeNotStorableError) Error() string {
 // when a parsing or a checking error occurred
 //
 type ParsingCheckingError struct {
-	Err            error
-	RuntimeStorage *InterpreterRuntimeStorage
-	Functions      stdlib.StandardLibraryFunctions
-	Code           []byte
-	Location       Location
-	Options        []sema.Option
-	UseCache       bool
-	Program        *ast.Program
-	Checker        *sema.Checker
+	Err          error
+	StorageCache Cache
+	Code         []byte
+	Location     Location
+	Options      []sema.Option
+	UseCache     bool
+	Program      *ast.Program
+	Checker      *sema.Checker
 }
 
 func (e *ParsingCheckingError) ChildErrors() []error {
--- a/runtime/errors.go
+++ b/runtime/errors.go
@@ -218,10 +218,10 @@ func (e *TransactionParameterTypeNotStorableError) Error() string {
 	)
 }
 
-// ExtendedParsingCheckingError is a special error which aids in debugging checking problems
-// by providing extra information about the state of the environment
-// Separate package to prevent cyclic imports with checker tests
-type ExtendedParsingCheckingError struct {
+// ParsingCheckingError provides extra information about the state of the environment
+// when a parsing or a checking error occurred
+//
+type ParsingCheckingError struct {
 	Err            error
 	RuntimeStorage *InterpreterRuntimeStorage
 	Functions      stdlib.StandardLibraryFunctions
@@ -232,14 +232,14 @@ type ExtendedParsingCheckingError struct {
 	Checker        *sema.Checker
 }
 
-func (e *ExtendedParsingCheckingError) ChildErrors() []error {
+func (e *ParsingCheckingError) ChildErrors() []error {
 	return []error{e.Err}
 }
 
-func (e *ExtendedParsingCheckingError) Error() string {
+func (e *ParsingCheckingError) Error() string {
 	return e.Err.Error()
 }
 
-func (e ExtendedParsingCheckingError) Unwrap() error {
+func (e ParsingCheckingError) Unwrap() error {
 	return e.Err
 }
--- a/runtime/errors.go
+++ b/runtime/errors.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/onflow/cadence/runtime/ast"
 	"github.com/onflow/cadence/runtime/errors"
 	"github.com/onflow/cadence/runtime/interpreter"
 	"github.com/onflow/cadence/runtime/sema"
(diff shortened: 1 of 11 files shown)


turbolent:
@m4ksio is this supposed to be released in the v0.9 release line? Why/how are the first three commits by me part of this?

runtime/runtime_test.go
```diff
@@ -2377,7 +2376,14 @@ func TestRuntimeCyclicImport(t *testing.T) {
...
 	require.IsType(t, Error{}, err)
-	assert.IsType(t, ast.CyclicImportsError{}, err.(Error).Unwrap())
+
+	// Temporary to help catch checking error
+	ee := err.(Error).Unwrap()
+
+	require.IsType(t, &ExtendedParsingCheckingError{}, ee)
+	eee := ee.(*ExtendedParsingCheckingError)
+
+	assert.IsType(t, ast.CyclicImportsError{}, eee.Unwrap())
```
turbolent, line 2386: The old could should have used `utils.RequireErrorAs` here, and not manual unwrap. This would have prevented you to have to add the additional manually unwrapping step.

```suggestion
	utils.RequireErrorAs(t, err, &ast.CyclicImportsError{})
```

Then again, `ast.CyclicImportsError` is gone in `master` (it is now `sema.CyclicImportsError`), are you sure you are working off of `master`?

runtime/errors.go
```diff
@@ -216,3 +217,29 @@ func (e *TransactionParameterTypeNotStorableError) Error() string {
 		e.Type.QualifiedString(),
 	)
 }
+
+// ExtendedParsingCheckingError is a special error which aids in debugging checking problems
+// by providing extra information about the state of the environment
+// Separate package to prevent cyclic imports with checker tests
```
turbolent, line 223: Not quite sure I understand this line, can you elaborate?

runtime/errors.go
```diff
@@ -216,3 +217,29 @@ func (e *TransactionParameterTypeNotStorableError) Error() string {
 		e.Type.QualifiedString(),
 	)
 }
+
+// ExtendedParsingCheckingError is a special error which aids in debugging checking problems
```
turbolent, line 221: I think it's fine to just name this `ParsingCheckingError`, it's useful in the general case.

runtime/tests/checker/utils.go
```diff
@@ -92,6 +93,12 @@ func ExpectCheckerErrors(t *testing.T, err error, count int) []error {
 
 	require.Error(t, err)
 
+	// Temporary to help catch checking error
+	ee, is := err.(*runtime.ExtendedParsingCheckingError)
+	if is {
+		err = ee.Unwrap()
+	}
+
 	assert.IsType(t, &sema.CheckerError{}, err)
```
turbolent, line 102: Same as above, this should use `utils.RequireErrorAs` instead of manually unwrapping

m4ksio:
> @m4ksio is this supposed to be released in the v0.9 release line? Why/how are the first three commits by me part of this?

I branched off `v0.9.2` tag, as this is what current `flow-go` repo is using, to minimize changes. I would assume some commits from this tag are not part of master, hence they show up here.

turbolent:
@m4ksio Can you please create a v0.9.3 branch and target this PR against it?

runtime/runtime_test.go
m4ksio: As discussed, it was branched off `0.9.2` so resolving comment.

runtime/tests/checker/utils.go
m4ksio: `RequireErrorAs` was added in after `0.9` so resolving

turbolent: Ah, right 👍

runtime/errors.go
m4ksio: Does this clarify cyclic imports enough? https://github.com/onflow/cadence/pull/451/files#diff-74ffa81b6938856f695ba112a721503dfc5ff2e852bc65d79d15ee4868f4bb97R4784-R4791

runtime/errors.go
```diff
@@ -216,3 +217,29 @@ func (e *TransactionParameterTypeNotStorableError) Error() string {
...
+// by providing extra information about the state of the environment
+// Separate package to prevent cyclic imports with checker tests
+type ExtendedParsingCheckingError struct {
+	Err            error
+	RuntimeStorage *InterpreterRuntimeStorage
+	Functions      stdlib.StandardLibraryFunctions
```
turbolent, line 227: How are these used for debugging? The fields in `interpreterRuntimeStorage` are also not exported, so is exposing the storage here (and exporting `interpreterRuntimeStorage` as `InterpreterRuntimeStorage` below) useful?

runtime/runtime.go
```diff
@@ -298,6 +298,13 @@ func (r *interpreterRuntime) ExecuteTransaction(
 
 	checker, err := r.parseAndCheckProgram(script, runtimeInterface, location, functions, nil, false)
 	if err != nil {
+		ee, is := err.(*ExtendedParsingCheckingError)
+		if is {
+			ee.RuntimeStorage = runtimeStorage
+			ee.Functions = functions
+			//return newError(ee.Err)
```
turbolent, line 305:
```suggestion
```

turbolent:
@m4ksio Pushed some improvements, feel free to convert the draft into a proper PR

runtime/errors.go
m4ksio: > How are these used for debugging?

They are dumped to a file for further usage. I didn't analyze what was particularly useful, just grabbed all there is

turbolent: Given that the runtime storage value doesn't provide any info, OK if I remove it?

Also, the functions that are injected are always the same, so IDK how useful it it is to include them in the error.

m4ksio: > Given that the runtime storage value doesn't provide any info, OK if I remove it?
Feel free if you think it doesn't provide any value, and add what you think it's best - you are the best person to know what internal elements might be helpful

turbolent:
@m4ksio What does the linked issue (1291) refer to? The issue in dapperlabs/flow-go seems unrelated

m4ksio:
> @m4ksio What does the linked issue (1291) refer to? The issue in dapperlabs/flow-go seems unrelated

https://github.com/dapperlabs/flow-internal/issues/1291

PR #455, opened 2020-11-17
Improving the error message when there is constructor arguments.

zhangchiqing:
## Description

When I'm deploying a contract with missing arguments, it shows the following error. I think it worth to print what was missing, which could help debugging the cause

Before:
```
invalid argument count: expected 1, got 0
```

After:
```
invalid argument count, too few arguments: expected 1, got 0, next missing argument: `AuthAccount
```

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -1162,9 +1162,21 @@ func (r *interpreterRuntime) instantiateContract(
 	argumentCount := len(argumentTypes)
 	parameterCount := len(parameterTypes)
 
-	if argumentCount != parameterCount {
-		return nil, fmt.Errorf("invalid argument count: expected %d, got %d", parameterCount, argumentCount)
+	if argumentCount < parameterCount {
+		return nil, fmt.Errorf(
+			"invalid argument count, too few arguments: expected %d, got %d, next missing argument: `%s`",
+			parameterCount, argumentCount,
+			parameterTypes[argumentCount],
+		)
+	} else if argumentCount > parameterCount {
+		return nil, fmt.Errorf(
+			"invalid argument count, too many arguments: expected %d, got %d",
+			parameterCount,
+			argumentCount,
+		)
 	}
+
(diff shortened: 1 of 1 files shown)


runtime/runtime.go
```diff
@@ -1163,7 +1163,14 @@ func (r *interpreterRuntime) instantiateContract(
...
 
 	if argumentCount != parameterCount {
-		return nil, fmt.Errorf("invalid argument count: expected %d, got %d", parameterCount, argumentCount)
+		if argumentCount < parameterCount {
+			return nil, fmt.Errorf("invalid argument count, too few arguments: expected %d, got %d, next missing argument: `%s`",
+				parameterCount, argumentCount,
+				parameterTypes[argumentCount])
+		} else if argumentCount > parameterCount {
+			return nil, fmt.Errorf("invalid argument count, too many arguments: expected %d, got %d",
+				parameterCount, argumentCount)
```
turbolent, line 1172: Chopping down the arguments could make this a bit more readable:

```suggestion
			return nil, fmt.Errorf(
				"invalid argument count, too few arguments: expected %d, got %d, next missing argument: `%s`",
				parameterCount, 
				argumentCount,
				parameterTypes[argumentCount],
			)
		} else if argumentCount > parameterCount {
			return nil, fmt.Errorf(
				"invalid argument count, too many arguments: expected %d, got %d",
				parameterCount, 
				argumentCount,
			)
```

turbolent (approved):
Great idea, nice!

runtime/runtime.go
```diff
@@ -1162,10 +1162,22 @@ func (r *interpreterRuntime) instantiateContract(
 	argumentCount := len(argumentTypes)
 	parameterCount := len(parameterTypes)
 
-	if argumentCount != parameterCount {
```
zhangchiqing, line 1165: replaced the check here with the `if...else if`

PR #457, opened 2020-11-17
Refactor address locations, make composite decoding backwards-compatible

turbolent:
Closes onflow/flow#84

[Cadence v0.10](https://github.com/onflow/cadence/releases/tag/v0.10.0) added support for multiple contracts per account.

Along with the changes to the contract management API, this also introduced a breaking change to how values are stored: Composite values are stored with a location and type identifier, which allows the interpreter to load the code related to the type. 

Previously, the location was an address location, which has for example an ID like `A.0x1`, which indicates the code for the type is the code deployed in account `0x1` (the whole code of the account, as there was only one code associated with an account).

In v0.10, address contract locations were introduced: Each contract of an account has a distinct identifier, e.g `AC.0x1.SomeContract`, which indicates the code for the type is specifically the code of the "SomeContract" contract deployed to account `0x1`. This was a breaking change, as we did not add support for decoding values stored with plain address locations. Instead, we planned to migrate all existing data of the existing networks to the new format (see #84). This would have worked, but would have been a breaking change for users: Many developers have started using type identifiers (e.g. directly comparison to expected values), for example to detect certain events (events are also composite values).

This PR reverts the breaking change, i.e. the introduction of address contract locations, by merging them with the existing address location, and adding backwards-compatibility to the composite value decoding: When a composite value is decoded and it is stored in the old format, i.e. it only has a plain address location, then infer the contract name from the type identifier, to reconstruct an address location with name, just like it is generated now by new code.

Once this PR is merged to master, I'll open another PR against v0.10.1, then release v0.10.2, and update all downstream dependencies.

### DIFF
--- a/runtime/ast/import.go
+++ b/runtime/ast/import.go
@@ -103,6 +103,10 @@ func (d *ImportDeclaration) MarshalJSON() ([]byte, error) {
 type Location interface {
 	// ID returns the canonical ID for this import location.
 	ID() LocationID
+	// TypeID returns a type ID for the given qualified identifier
+	TypeID(qualifiedIdentifier string) TypeID
+	// QualifiedIdentifier returns the qualified identifier for the given type ID
+	QualifiedIdentifier(typeID TypeID) string
 }
 
 func LocationsMatch(first, second Location) bool {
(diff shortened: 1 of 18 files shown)


janezpodhostnik:
So I understand that if I currently have a contract that is checking types (not 100% on how that currently looks like), I would be checking if the type is: `A.0x1.TestStruct`. However after the switch to multiple contracts per address this same object will look like `A.0x1.TestContract.TestStruct`. Did I read this correctly?

turbolent:
@janezpodhostnik Almost: A type `TestStruct` nested in a contract `TestContract` and deployed to address 0x1 used to have type ID `A.0x1.TestContract.TestStruct` (location `A.0x1` + qualified identifier `TestContract.TestStruct`).
 
After the changes of v0.10, it became `AC.0x1.TestContract.TestContract.TestStruct` (location `AC.0x1.TestContract` + qualified identifier `TestContract.TestStruct`).

This PR reverts to the old type ID `A.0x1.TestContract.TestStruct`, i.e. reverts the breaking change of changing the type ID completely.

janezpodhostnik:
Ah ok missed that part!

Everything looks good to me.

janezpodhostnik (approved):

PR #460, opened 2020-11-19
Refactor address locations, make composite decoding backwards-compatible

turbolent:
Same as #457, backported to v0.10 (for more info, see the description in the referenced PR)


janezpodhostnik (approved):
LGTM!

PR #462, opened 2020-11-19
Adding error to emit event

ramtinms:
## Description

This PR updates the interface to return errors for emitEvent method, this functionality is needed for the environment to enforce event limit sizes.
______

For contributor use:

- [ ] Linked to Github issue with discussion and accepted design OR link to the spec that describes this work

### DIFF
--- a/runtime/interface.go
+++ b/runtime/interface.go
@@ -67,9 +67,9 @@ type Interface interface {
 	// GetSigningAccounts returns the signing accounts.
 	GetSigningAccounts() []Address
 	// Log logs a string.
-	Log(string)
+	Log(string) error
 	// EmitEvent is called when an event is emitted by the runtime.
-	EmitEvent(cadence.Event)
+	EmitEvent(cadence.Event) (err error)
 	// ValueExists returns true if the given key exists in the storage, owned by the given account.
 	ValueExists(owner, key []byte) (exists bool, err error)
 	// GenerateUUID is called to generate a UUID.
@@ -194,7 +194,7 @@ func (i *EmptyRuntimeInterface) GetSigningAccounts() []Address {
 
 func (i *EmptyRuntimeInterface) Log(_ string) {}
 
-func (i *EmptyRuntimeInterface) EmitEvent(_ cadence.Event) {}
+func (i *EmptyRuntimeInterface) EmitEvent(_ cadence.Event) error {}
 
 func (i *EmptyRuntimeInterface) GenerateUUID() uint64 {
 	return 0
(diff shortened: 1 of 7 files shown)


runtime/runtime.go
```diff
@@ -927,10 +931,14 @@ func (r *interpreterRuntime) emitAccountEvent(
...
 
+	var err error
 	exportedEvent := exportEvent(eventValue)
 	wrapPanic(func() {
-		runtimeInterface.EmitEvent(exportedEvent)
+		err = runtimeInterface.EmitEvent(exportedEvent)
 	})
+	if err != nil {
+		panic(err)
+	}
```
ramtinms, line 941: Is this the right way to revert the transaction?

turbolent:
@ramtinms Currently the runtime interface isn't very consistent, some functions have an error return value, some don't, the embedder/host environment implementing the interface sometimes reports errors through the error return value, other times through a panic.

Both variants are handled in the same way and there is no difference: When an interface function panics, the panic is caught (`wrapPanic`) and "bubbled up" by "rethrowing" the panic. If the interface functions happens to return an error, again, Cadence will just panic with it, as execution can not continue if the function failed.

So ideally the interface should be cleaned up and all functions should have the same way of reporting an error. Given that all function calls are already wrapped and their panics are caught, I think we can just remove the error return value from all functions, instead of adding one to the ones that don't have one yet. That way we don't have to handle two error paths, the function returns an error + the function panics, but only the latter, because at the end of the day, both error paths are handled the same way: the error is just bubbled up through a panic.

TLDR: I don't think this PR is needed, you can just panic in the implementation of `EmitEvent` in e.g. FVM, and in fact, we should probably just remove the error return value for the other functions.

Happy to sync further on this, just ping me!

turbolent:
@ramtinms thanks for syncing again! I missed that the intend here is to be able to indicate a user error, looks good!

runtime/interface.go
```diff
@@ -194,7 +194,7 @@ func (i *EmptyRuntimeInterface) GetSigningAccounts() []Address {
 
 func (i *EmptyRuntimeInterface) Log(_ string) {}
 
-func (i *EmptyRuntimeInterface) EmitEvent(_ cadence.Event) {}
+func (i *EmptyRuntimeInterface) EmitEvent(_ cadence.Event) error {return nil}
```
turbolent, line 197:
```suggestion
func (i *EmptyRuntimeInterface) EmitEvent(_ cadence.Event) error {
	return nil
}
```

runtime/runtime.go
turbolent: @ramtinms yep, that looks good!

turbolent (approved):

PR #463, opened 2020-11-19
Extended debug

turbolent:
Same as #451, against v0.10:

Wrap parsing and checking errors, and include useful information for e.g. debugging purposes.


m4ksio (approved):
Looks good!

PR #465, opened 2020-11-21
Add support for coverage reports

turbolent:
When running tests it is useful to know what parts of the program where covered. Add support for a coverage to the interpreter. Start with measuring line hits. This can be extended in the future.

### DIFF
--- a/runtime/coverage.go
+++ b/runtime/coverage.go
@@ -1,0 +1,61 @@
+/*
+ * Cadence - The resource-oriented smart contract programming language
+ *
+ * Copyright 2019-2020 Dapper Labs, Inc.
+ *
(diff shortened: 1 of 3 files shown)


janezpodhostnik (approved):
This will be amazing for a lot of things, not just coverage reports. Debugging for instance. 

@turbolent do you thing we should put a coverage report collector on the fvm and just always track line hits on the testnet?

turbolent:
@janezpodhostnik Yeah, I think that would be great! This PR is quit low-level, but the FVM package, or even higher up the emulator, could generate reports (JSON, HTML, etc.) from this. I guess it could be just an `fvm.Option` and then a configuration flag in the emulator.

PR #467, opened 2020-11-23
Permit storage path expression to contain dots

### LINKED ISSUE #419: Allow a Path to contain dots
## Instructions

Please fill out the template below to the best of your ability.

### Issue To Be Solved

Allow a Path to contain dots.
Example:
```
/storage/this.path.contains.dots
/public/A.0x01.ContractName.ResourceName
```

(Replace this text: A concise description of your preferred solution. Things to address include:

* Details of the technical implementation
* Tradeoffs made in design decisions
* Caveats and considerations for the future

If there are multiple solutions, please present each one separately. Save comparisons for the very end.)

### (Optional): Context

(Replace this text:
What are you currently working on that this is blocking?)

benjaminkvm:
Closes #419.

## Description

Hey Bastian! 👋 Hope things are going well in Cadence-land! 🎶 Quick little PR:

Parse the `TokenDot` as a valid character within a `PathExpression`. Update relevant tests and add an additional test.

I exposed the sub-identifiers (identifiers delimited by `TokenDot`) to the interpreter, however on second thought I could consolidate them at the parser level if that is preferred -- let me know if you would prefer that and I can revise!

This approach probably makes less sense if dots have no significance but I thought it would be good to get your take before I simplify!
______

For contributor use:

### DIFF
--- a/docs/language/accounts.md
+++ b/docs/language/accounts.md
@@ -122,7 +122,7 @@ Objects are stored under paths.
 Paths consist of a domain and an identifier.
 
 Paths start with the character `/`, followed by the domain, the path separator `/`,
-and finally the identifier.
+and finally the identifier. Path identifiers can consist of alphanumeric characters in addition to the `_` and `.` characters.
 For example, the path `/storage/test` has the domain `storage` and the identifier `test`.
 
 There are only three valid domains: `storage`, `private`, and `public`.
--- a/runtime/ast/expression.go
+++ b/runtime/ast/expression.go
@@ -1007,9 +1007,9 @@ func (e *ForceExpression) MarshalJSON() ([]byte, error) {
 // PathExpression
 
 type PathExpression struct {
-	StartPos   Position `json:"-"`
-	Domain     Identifier
-	Identifier Identifier
+	StartPos    Position `json:"-"`
+	Domain      Identifier
+	Identifiers []Identifier
 }
 
 func (*PathExpression) isExpression() {}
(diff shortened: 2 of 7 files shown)


benjaminkvm:
Hmm, just considered the case where a storage path is started/terminated with a `TokenDot`, or there are multiple `TokenDot`s in a row, etc.

Those dot lexemes should be added to the `Identifiers` and then the entire slice concat'd from within `Identifier()` (provided again we want those individual identifiers up to the interpreter level).

Can fix that!

turbolent (approved):

turbolent:
Hi Ben! Things are well, we're improving things steadily 😄 Hope you're well, thank you for continuing to contribute! 👏 

The code looks good! I think it's fine to allow just `identifier ('.' identifier)*` for now. Good point with the other cases (leading dot, consecutive dots, etc.), but we can improve that in a following PR if needed. 

Nice work also updating the documentation 👍

turbolent:
I think I need to fix CI (again) because somehow I seem to have broken it for PR of forks

turbolent:
@benjaminkvm Sorry for not getting back to this for such a long time!

When I initially [saw the issue and commented on it](https://github.com/onflow/cadence/issues/419#issuecomment-710272503), I hadn't thought this feature request fully through, but the last weeks and going over the parser code base with @SupunS, I realized that this change is actually a bit problematic: 

Currently, member access expressions use the syntax `'.' identifier`, e.g. `foo.bar`,
and so allowing dots in paths would make it impossible to add fields or functions to paths in the future, as the path literal parser would eagerly parse as much as possible, including an intended member access. For example, in the future we might want to add a `length` field, but with this change `/foo/bar.length` would be a path with identifier `bar.length`. I guess we could still merge this change, and developers would need to perform member access by using parentheses (`(/foo/bar).length`).

turbolent:
Closing this for now. We'll need to discuss the design first

PR #468, opened 2020-11-23
Infer address location name from type ID for static types

turbolent:
Just like inferring the name for address locations of composite values, also infer the name for address locations in static composite types and static interface types.

Kudos to @janezpodhostnik for finding and fixing this bug! I only refactored and added tests


runtime/interpreter/encoding_test.go
```diff
@@ -3387,6 +3432,47 @@ func TestEncodeDecodeLinkValue(t *testing.T) {
 		)
 	})
 
+	t.Run("interface, struct, address location without name", func(t *testing.T) {
```
janezpodhostnik, line 3435: Nice tests!

janezpodhostnik (approved):
👍

PR #470, opened 2020-11-25
Updating interface methods

ramtinms:
## Description
This PR 
  - updates several methods of the Interface and appends error return values, which enables the environment to return non-fatal errors (e.g. method not supported) as an outcome of calling the method. 
  - introduces a new interface method (SetComputationUsed) which will be used by the runtime to report the computation used values
______

For contributor use:


runtime/runtime.go
```diff
@@ -540,7 +542,14 @@ func (r *interpreterRuntime) parseAndCheckProgram(
 				sema.WithPredeclaredValues(valueDeclarations),
 				sema.WithPredeclaredTypes(typeDeclarations),
 				sema.WithValidTopLevelDeclarationsHandler(validTopLevelDeclarations),
-				sema.WithLocationHandler(runtimeInterface.ResolveLocation),
+				sema.WithLocationHandler(func(identifiers []Identifier, location Location) (res []ResolvedLocation) {
+					var err error
+					res, err = runtimeInterface.ResolveLocation(identifiers, location)
```
turbolent, line 547: Just realizing that the call to the runtime interface should be wrapped, just like in all other cases:

```suggestion
					wrapPanic(func() {
						res, err = runtimeInterface.ResolveLocation(identifiers, location)
					})
```

runtime/runtime.go
```diff
@@ -633,7 +642,11 @@ func (r *interpreterRuntime) newInterpreter(
 		),
 		interpreter.WithUUIDHandler(func() (uuid uint64) {
 			wrapPanic(func() {
-				uuid = runtimeInterface.GenerateUUID()
+				var err error
+				uuid, err = runtimeInterface.GenerateUUID()
+				if err != nil {
+					panic(err)
+				}
```
turbolent, line 649: The `wrapPanic` should be only around the interface call:

```go
var err error
wrapPanic(func() {
	uuid, err = runtimeInterface.GenerateUUID()
})
if err != nil {
	panic(err)
}
```

runtime/runtime.go
```diff
@@ -781,6 +794,10 @@ func (r *interpreterRuntime) meteringInterpreterOptions(runtimeInterface Interfa
 			return
 		}
 
+		err := runtimeInterface.SetComputationUsed(used)
```
turbolent, line 797: The call to the runtime interface should be wrapped, just like in all other cases:

```suggestion
		var err error
		wrapPanic(func() {
			err = runtimeInterface.SetComputationUsed(used)
		})
```

runtime/runtime.go
```diff
@@ -1305,17 +1322,24 @@ func (r *interpreterRuntime) newLogFunction(runtimeInterface Interface) interpre
 	return func(invocation interpreter.Invocation) trampoline.Trampoline {
 		message := fmt.Sprint(invocation.Arguments[0])
 		wrapPanic(func() {
-			runtimeInterface.Log(message)
+			err := runtimeInterface.Log(message)
+			if err != nil {
+				panic(err)
+			}
 		})
```
turbolent, line 1329: The `wrapPanic` should be only around the interface call:

```go
var err error
wrapPanic(func() {
	err = runtimeInterface.Log(message)	
})
if err != nil {
	panic(err)
}
```

runtime/runtime.go
```diff
@@ -1305,17 +1322,24 @@ func (r *interpreterRuntime) newLogFunction(runtimeInterface Interface) interpre
...
 
-func (r *interpreterRuntime) getCurrentBlockHeight(runtimeInterface Interface) (currentBlockHeight uint64) {
+func (r *interpreterRuntime) getCurrentBlockHeight(runtimeInterface Interface) (currentBlockHeight uint64, err error) {
+	var er error
 	wrapPanic(func() {
-		currentBlockHeight = runtimeInterface.GetCurrentBlockHeight()
+		currentBlockHeight, er = runtimeInterface.GetCurrentBlockHeight()
 	})
+	if er != nil {
+		err = newError(er)
```
turbolent, line 1341:
```suggestion
		return 0, newError(er)
```

runtime/runtime.go
```diff
@@ -1343,7 +1367,10 @@ func (r *interpreterRuntime) getBlockAtHeight(height uint64, runtimeInterface In
 
 func (r *interpreterRuntime) newGetCurrentBlockFunction(runtimeInterface Interface) interpreter.HostFunction {
 	return func(invocation interpreter.Invocation) trampoline.Trampoline {
-		height := r.getCurrentBlockHeight(runtimeInterface)
+		height, err := r.getCurrentBlockHeight(runtimeInterface)
```
turbolent, line 1370: The call to the runtime interface should be wrapped, just like in all other cases:

```suggestion
		var height uint64
		var err error
		wrapPanic(func() {
			height, err = r.getCurrentBlockHeight(runtimeInterface)
		})
```

runtime/stdlib/crypto.go
```diff
@@ -107,13 +107,16 @@ func newCryptoContractVerifySignatureFunction(signatureVerifier CryptoSignatureV
...
-			isValid := signatureVerifier.VerifySignature(signature,
+			isValid, err := signatureVerifier.VerifySignature(signature,
 				tag,
 				signedData,
 				publicKey,
 				signatureAlgorithm,
 				hashAlgorithm,
 			)
+			if err != nil {
+				panic(fmt.Errorf("VerifySignature: failed to verify the signature: %w", err))
```
turbolent, line 118: Just panic with the error

```suggestion
				panic(err)
```

runtime/stdlib/crypto.go
```diff
@@ -153,7 +156,11 @@ func newCryptoContractHashFunction(hasher CryptoHasher) interpreter.FunctionValu
 			}
 			hashAlgorithm := hashAlgorithmStringValue.Str
 
-			digest := hasher.Hash(data, hashAlgorithm)
+			digest, err := hasher.Hash(data, hashAlgorithm)
+			if err != nil {
+				panic(fmt.Errorf("hash: hashing failed: %w", err))
```
turbolent, line 161: Just panic with the error

```suggestion
				panic(err)
```

turbolent (commented):
Looks good, thanks for cleaning this up @ramtinms! Just a few comments

runtime/runtime.go
ramtinms: applied at 
19d823e

runtime/runtime.go
ramtinms: applied at 
19d823e

turbolent (approved):
Nice, looks good!

turbolent:
@ramtinms OK if I rebase on master and merge?

runtime/runtime.go
```diff
@@ -540,7 +542,16 @@ func (r *interpreterRuntime) parseAndCheckProgram(
 				sema.WithPredeclaredValues(valueDeclarations),
 				sema.WithPredeclaredTypes(typeDeclarations),
 				sema.WithValidTopLevelDeclarationsHandler(validTopLevelDeclarations),
-				sema.WithLocationHandler(runtimeInterface.ResolveLocation),
+				sema.WithLocationHandler(func(identifiers []Identifier, location Location) (res []ResolvedLocation) {
```
janezpodhostnik, line 545: Should this error be handled here or on the location where it is called?

runtime/runtime.go
```diff
@@ -632,9 +643,13 @@ func (r *interpreterRuntime) newInterpreter(
 			r.injectedCompositeFieldsHandler(runtimeInterface, runtimeStorage),
 		),
 		interpreter.WithUUIDHandler(func() (uuid uint64) {
+			var err error
```
janezpodhostnik, line 646: Should this error be handled here or on the location where it is called?

janezpodhostnik (approved):
👍

runtime/runtime.go
turbolent: @janezpodhostnik Do you mean `Interpreter.UUIDHandler` should be have an error return value and this handler should just pass it down?

runtime/runtime.go
turbolent: @janezpodhostnik Do you mean `Interpreter.LocationHandler` should have an error return value and this handler should just pass it down?

runtime/runtime.go
janezpodhostnik: Yep, not sure which is better here? Passing it down or handling it here.

turbolent: Refactored it in d37c9f3

runtime/runtime.go
turbolent: Refactored it in d37c9f3

PR #472, opened 2020-11-27
Allow host environment to predeclare values, predicated on location

### LINKED ISSUE #471: Selective injection of value declarations
### Issue To Be Solved

When executing a transaction or script, the host environment may want to provide additional value declarations, and make them selectively available in certain locations.

This will allow the FVM package to provide a function to the service account which allows it to set account storage capacities.

### Suggested Solution

- [x] Make value declarations only available in certain locations based on a predicate 
- [x] Allow host environment to provide value declarations when executing transactions and scripts
- [x] Tests

turbolent:
Closes #471

- Refactor/bundle the parameters of the runtime interface main functions for script and transaction execution, introduce `Script` and `Context`
- Allow the host environment to provide predeclared values. The value declarations are passed to all  checkers and interpreters.
- However, value declarations may have a predicate function, which determines if the value should be declared in the given location

### DIFF
--- a/runtime/predeclaredvalues_test.go
+++ b/runtime/predeclaredvalues_test.go
@@ -1,0 +1,150 @@
+/*
+ * Cadence - The resource-oriented smart contract programming language
+ *
+ * Copyright 2019-2020 Dapper Labs, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package runtime
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/onflow/cadence/runtime/ast"
+	"github.com/onflow/cadence/runtime/common"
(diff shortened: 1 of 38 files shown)


runtime/tests/interpreter/equality_test.go
```diff
@@ -37,6 +37,22 @@ func TestInterpretEquality(t *testing.T) {
...
+				Identifier: "something",
+			},
+		}
+
+		capabilityValueDeclarartion := stdlib.StandardLibraryValue{
+			Name:       "cap",
+			Type:       &sema.CapabilityType{},
+			Value:      capabilityValue,
+			Kind:       common.DeclarationKindConstant,
+			IsConstant: true,
```
janezpodhostnik, line 53:
```
			Kind:       common.DeclarationKindConstant,
			IsConstant: true,
```
Any way to get rid of this redundancy?

runtime/predeclaredvalues_test.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/onflow/cadence/runtime/ast"
+	"github.com/onflow/cadence/runtime/common"
+	"github.com/onflow/cadence/runtime/sema"
+	"github.com/onflow/cadence/runtime/tests/checker"
+	"github.com/onflow/cadence/runtime/tests/utils"
+)
+
+func TestRuntimePredeclaredValues(t *testing.T) {
```
janezpodhostnik, line 34: Nice! This test explains a lot about usage.

janezpodhostnik (approved):
Nice! This decouple cadence and flow go by a lot!

runtime/tests/interpreter/equality_test.go
turbolent: Good idea! I've removed it in 3518c6d61be4f69ebf0e865d2aaad656d3e39c89 and replaced it with being inferred from the declaration kind

PR #476, opened 2020-11-30
Update the diagram illustrating the architcture of the runtime

turbolent:
The diagram was outdated: The parser is not generated by ANTLR anymore.


zhangchiqing (approved):
nice diagram

PR #477, opened 2020-12-01
Structured type identifiers

### LINKED ISSUE #420: Add `EventType.TypeID` elements

### LINKED ISSUE #442: Decompose composite types ID into field
Some work has been started under https://github.com/onflow/cadence/issues/420 but further decomposition is planned, as discussed.

turbolent:
Closes #420 
Closes #442
Closes dapperlabs/flow-go#5033 (not needed anymore)
Replaces #422

This PR

- Refactors all location related code to the `common` package. The code was partially in the `ast` package and the `runtime` package, and didn't properly handle decoding for all kinds of locations. Maybe the code should be in a separate `location` package?
- Refactors the single type ID decoding function into separate functions, and adds tests for all of them. Custom type ID decoders can be registered from other packages (e.g. for Flow locations)
- Changes `cadence.CompositeType`s and `cadence.InterfaceType`s to so the location and qualified identifier can be accessed without having to decode the type ID.
- Unlike #422 this PR does not change the JSON encoding, to stay backwards-compatible and avoid breakage in SDKs; i.e. this PR is only a breaking change for users of the `cadence` package.

Sorry this PR is so large. Most of the changes are package name changes in tests


runtime/common/addresslocation.go
```diff
@@ -0,0 +1,158 @@
...
+
+	parts := strings.SplitN(typeID, ".", 4)
+
+	pieceCount := len(parts)
+	switch pieceCount {
+	case 1:
+		return newError("missing location")
+	case 2:
+		return newError("missing qualified identifier")
+	}
```
jordanschalm, line 123: Can we be sure which piece is missing based only on the number of pieces? What about an input like `<location>.<qualifiedIdentifier>`

runtime/common/addresslocation.go
```diff
@@ -0,0 +1,158 @@
...
+			err,
+		)
+	}
+
+	name := parts[2]
+	qualifiedIdentifier := name
+
+	if pieceCount > 3 {
+		qualifiedIdentifier = fmt.Sprintf("%s.%s", qualifiedIdentifier, parts[3])
+	}
```
jordanschalm, line 150: A comment explaining these two cases would be handy. I'm guessing one case is `A.<address>.<variableName>` and the other is `A.<address>.<contractName>.<variableName>`? 

Or maybe we can name the different parts in each case like:
```suggestion
	var qualifiedIdentifier
	switch pieceCount {
	case 3:
		varName := parts[2]
		qualifiedIdentifier = varName
	case 4:
		contractName := parts[2]
		varName := parts[3]
		qualifiedIdentifier = fmt.Sprintf("%s.%s", contractName, varName)
	default:
	// return error?
	}

```

runtime/common/location.go
```diff
@@ -0,0 +1,99 @@
...
+
+// TypeID
+//
+type TypeID string
+
+func NewTypeID(parts ...string) TypeID {
+	return TypeID(strings.Join(parts, "."))
+}
+
+type TypeIDDecoder = func(typeID string) (location Location, qualifiedIdentifier string, err error)
```
jordanschalm, line 67:
```suggestion
type TypeIDDecoder func(typeID string) (location Location, qualifiedIdentifier string, err error)
```

jordanschalm (approved):
Nice work 👏

runtime/common/addresslocation.go
turbolent: The type with an address location must have the format `<prefix>.<location>.<qualifiedIdentifier>`, where `<qualifiedIdentifier>` itself is one or more identifiers separated by a dot. `<prefix>` must be `A`. `<location>` must be a hex string. The first part of `qualifiedIdentifier` is also the contract name.

So we split by at most 4 components (we don't need to split `qualifiedIdentifier` completely, just the first part for the name). 

We don't need to check `pieceCount` for 0, as `strings.SplitN` will always return at minimum one item. So `pieceCount` must be >=3, and we have errors for >0 && <3.

runtime/common/addresslocation.go
turbolent: Good point, I tried to explain the decoding by adding comments and making cases more explicit in 1cf82ef.

PR #480, opened 2020-12-01
Flush the cache of the storage before querying the used storage amount

### LINKED ISSUE #479: Flush  runtime storage cache before storageUsed call
### Issue To Be Solved

It should be possible to add a snippet to the end of any transaction that does the following:
- Check accounts storage used.
- If it has grown over accounts storage capacity,
- buy more storage for it.

For this to work `accont.storageUsed` should return the current storage used.

```
let u1 = accont.storageUsed
// store something new on the account
let u2 = accont.storageUsed 
if(u1 != u2) // should be true
```

This currently doesn't happen because storing something on the account only puts it into the runtime storage, while storageUsed is being calculated on the delta level.

P.S.: This same issue also hold for storageCapacity

### Suggest A Solution

On call to storage used `accont.storageUsed` flush the cache first.

### Context

storage fees FLIP: https://github.com/onflow/flow/pull/99

turbolent:
Closes #479

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -6525,7 +6525,7 @@ type AccountValue interface {
 // AuthAccountValue
 type AuthAccountValue struct {
 	Address                 AddressValue
-	storageUsedGet          func() UInt64Value
+	storageUsedGet          func(interpreter *Interpreter) UInt64Value
 	storageCapacityGet      func() UInt64Value
 	addPublicKeyFunction    FunctionValue
 	removePublicKeyFunction FunctionValue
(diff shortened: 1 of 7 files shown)


runtime/interpreter/value.go
```diff
@@ -6525,7 +6525,7 @@ type AccountValue interface {
 // AuthAccountValue
 type AuthAccountValue struct {
 	Address                 AddressValue
-	storageUsedGet          func() UInt64Value
+	storageUsedGet          func(interpreter *Interpreter) UInt64Value
 	storageCapacityGet      func() UInt64Value
```
janezpodhostnik, line 6529: storageCapacityGet will possibly need the same thing, but we can wait with changing that until I confirm that is the case.

janezpodhostnik (approved):
LGTM!

PR #481, opened 2020-12-03
Improve error pretty printing

### LINKED ISSUE #437: Improve information in execution error
### Issue To Be Solved

Execution errors are currently not very useful when debugging issues in Cadence programs.
For example:

```
Execution failed:
    Checking failed:
        checking of imported program `0b2a3299cc857e29` failed:
            checking of imported program `1d7e57aa55817448` failed:
                cannot find type in this scope: `X`. not found in this scope
            cannot find type in this scope: `NonFungibleToken`. not found in this scope
...
```

There is no position information and it is unclear what the imported programs on-chain are.

### Solution

- Include position information
- Maybe use error pretty printer
- Maybe include full source code

turbolent:
Closes #437

This PR
- Refactors error printing into one place: it was split up, with one part only in the command line,  and was tightly coupled to parser and checker error. Make it generic so that any error can be pretty printed, based on the `HasImportLocation` and `ParentErrors` interfaces
- Pretty prints runtime errors (which may be parsing, checking, or execution errors). Support imports bu keep track of all imported programs (codes)
- Fixes the location info in interpreter errors: We need to use the interpreter of the executed statement to get the position information, not the interpreter that runs the trampoline

### DIFF
--- a/runtime/cmd/check/main.go
+++ b/runtime/cmd/check/main.go
@@ -33,6 +33,7 @@ import (
 
 	"github.com/onflow/cadence/runtime/ast"
 	"github.com/onflow/cadence/runtime/cmd"
+	"github.com/onflow/cadence/runtime/common"
 	"github.com/onflow/cadence/runtime/pretty"
 	"github.com/onflow/cadence/runtime/sema"
 )
@@ -177,7 +178,9 @@ func runPath(path string, bench bool, useColor bool) (res result, succeeded bool
 	var program *ast.Program
 	var must func(error)
 
-	codes := map[string]string{}
+	codes := map[common.Location]string{}
+
+	location := common.StringLocation(path)
 
 	func() {
 		defer func() {
@@ -187,15 +190,15 @@ func runPath(path string, bench bool, useColor bool) (res result, succeeded bool
 			}
 		}()
 
-		program, must = cmd.PrepareProgram(code, path, codes)
+		program, must = cmd.PrepareProgram(code, location, codes)
 
-		checker, _ = cmd.PrepareChecker(program, path, codes, must)
+		checker, _ = cmd.PrepareChecker(program, location, codes, must)
 
 		err = checker.Check()
 		if err != nil {
 			var builder strings.Builder
 			printErr := pretty.NewErrorPrettyPrinter(&builder, useColor).
-				PrettyPrintError(err, path, codes)
+				PrettyPrintError(err, location, codes)
 			if printErr != nil {
 				panic(printErr)
 			}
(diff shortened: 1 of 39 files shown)


runtime/tests/interpreter/interpreter_test.go
```diff
@@ -68,7 +69,14 @@ func parseCheckAndInterpretWithOptions(
 		options.HandleCheckerError(err)
 	} else {
```
janezpodhostnik, line 70: unnecessary nesting. Join into else if.

runtime/tests/checker/utils.go
```diff
@@ -47,16 +48,23 @@ func ParseAndCheckWithOptions(
 	options ParseAndCheckOptions,
 ) (*sema.Checker, error) {
 
+	if options.Location == nil {
+		options.Location = utils.TestLocation
+	}
+
 	program, err := parser2.ParseProgram(code)
 	if !options.IgnoreParseError && !assert.NoError(t, err) {
```
janezpodhostnik, line 56: Double negatives are confusing, could this be replaced by `assert.Error(t, err)`.

runtime/cmd/check/main.go
```diff
@@ -184,14 +190,18 @@ func runPath(path string, bench bool) (res result, succeeded bool) {
...
-		checker, _ = cmd.PrepareChecker(program, path, codes, must)
+		checker, _ = cmd.PrepareChecker(program, location, codes, must)
 
 		err = checker.Check()
 		if err != nil {
 			var builder strings.Builder
-			cmd.PrettyPrintError(&builder, err, path, codes)
+			printErr := pretty.NewErrorPrettyPrinter(&builder, useColor).
+				PrettyPrintError(err, location, codes)
+			if printErr != nil {
```
janezpodhostnik, line 202: all currant usages of `PrettyPrintError` panic if the function returns an error... should that perhaps be the responsibility of `PrettyPrintError`?

runtime/pretty/print.go
```diff
@@ -1,32 +1,17 @@
-/*
- * Cadence - The resource-oriented smart contract programming language
```
janezpodhostnik, line 2: Just curious why this file doesn't get the copyright header?

janezpodhostnik (approved):
Nice improvement!

I left some comments that are not directly about the changes I just happened to notice it while going through the changes.

runtime/tests/checker/utils.go
turbolent: Agreed, but here we want to make sure that the assertion succeeds that there is no error – if it does not succeed, then an unexpected error occurred and we report it. 
Asserting there is an error is not the same.

runtime/tests/interpreter/interpreter_test.go
turbolent: Merged in 00892183404433aeff0678f368bd400556d343bf

runtime/cmd/check/main.go
turbolent: I think letting callers decide if a pretty printing error is fatal or not is better than always making it fatal. You're right that in all current cases where the pretty printer is used, an error is fatal, but it doesn't need to be in future uses.

runtime/pretty/print.go
turbolent: Good catch! I copied the code over to a new file and Git detected the move. While copying I forgot to copy the license header. There was a bug in the license header check script, I fixed it in 
2459ac3 and added the missing license headers in 7ea2ea4

PR #487, opened 2020-12-04
Fix the export of static types

turbolent:
Static types are currently only exported as a string. However, only the qualified identifier is used (e.g. any type `S` in `C` has qualified identifier `C.S`), which is not unique. 
Use the full type ID instead (e.g. type `S` in `C`, deployed to account 0x1, has ID `A.0x1.C.S`).

Also, add more tests (export of static types which are not primitive)

### DIFF
--- a/encoding/json/encoding_test.go
+++ b/encoding/json/encoding_test.go
@@ -678,15 +678,15 @@ func TestEncodeResource(t *testing.T) {
 		t.Parallel()
 
 		script := `
-			access(all) resource Foo {
-				access(all) let bar: Int
+			pub resource Foo {
+				pub let bar: Int
 	
 				init(bar: Int) {
 					self.bar = bar
 				}
 			}
 	
-			access(all) fun main(): @Foo {
+			pub fun main(): @Foo {
 				return <- create Foo(bar: 42)
 			}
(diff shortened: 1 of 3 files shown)


encoding/json/encoding_test.go
```diff
@@ -677,15 +677,15 @@ func TestEncodeResource(t *testing.T) {
 		t.Parallel()
 
 		script := `
-			access(all) resource Foo {
-				access(all) let bar: Int
+			pub resource Foo {
```
janezpodhostnik, line 680: I thought access(all) and pub were equivalent. Why the change? just for clarity?

janezpodhostnik (approved):
I think I understand the change and why, but I'm missing a decoding of static types change... or did I misunderstand?

encoding/json/encoding_test.go
turbolent: Yeah, all tests use `pub` and I just replaced it for consistency reasons.

psiemens (approved):

turbolent:
@janezpodhostnik this PR only fixes the export of the static types in links, types, and capabilities, because the import of those values is not supported yet

PR #488, opened 2020-12-05
Optimize composite values

turbolent:
Only store the qualified identifier of a composite value, not the full type ID.
The composite value already stores the location, so it was duplicated in the type ID.

### DIFF
--- a/runtime/common/addresslocation_test.go
+++ b/runtime/common/addresslocation_test.go
@@ -74,7 +74,7 @@ func TestDecodeAddressLocationTypeID(t *testing.T) {
 
 		t.Parallel()
 
-		_, _, err := decodeAddressLocationTypeID("A.01")
+		_, _, err := decodeAddressLocationTypeID("A.0000000000000001")
 		require.EqualError(t, err, "invalid address location type ID: missing qualified identifier")
 	})
 
@@ -82,7 +82,7 @@ func TestDecodeAddressLocationTypeID(t *testing.T) {
 
 		t.Parallel()
 
-		_, _, err := decodeAddressLocationTypeID("X.01.T")
+		_, _, err := decodeAddressLocationTypeID("X.0000000000000001.T")
 		require.EqualError(t, err, "invalid address location type ID: invalid prefix: expected \"A\", got \"X\"")
 	})
 
@@ -90,7 +90,7 @@ func TestDecodeAddressLocationTypeID(t *testing.T) {
 
 		t.Parallel()
 
-		location, qualifiedIdentifier, err := decodeAddressLocationTypeID("A.01.T")
+		location, qualifiedIdentifier, err := decodeAddressLocationTypeID("A.0000000000000001.T")
 		require.NoError(t, err)
 
 		assert.Equal(t,
@@ -107,7 +107,7 @@ func TestDecodeAddressLocationTypeID(t *testing.T) {
 
 		t.Parallel()
 
-		location, qualifiedIdentifier, err := decodeAddressLocationTypeID("A.01.T.U")
+		location, qualifiedIdentifier, err := decodeAddressLocationTypeID("A.0000000000000001.T.U")
 		require.NoError(t, err)
 
 		assert.Equal(t,
--- a/runtime/interpreter/decode.go
+++ b/runtime/interpreter/decode.go
@@ -227,11 +227,20 @@ func (d *Decoder) decodeValue(v interface{}, path []string) (Value, error) {
 			return d.decodeType(v.Content)
 
 		default:
-			return nil, fmt.Errorf("unsupported decoded tag: %d, %v", v.Number, v.Content)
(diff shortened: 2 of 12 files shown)


runtime/common/addresslocation_test.go
```diff
@@ -74,23 +74,23 @@ func TestDecodeAddressLocationTypeID(t *testing.T) {
 
 		t.Parallel()
 
-		_, _, err := decodeAddressLocationTypeID("A.01")
```
turbolent, line 77: The type identifiers here were not representing actual values, so I fixed them

runtime/interpreter/decode.go
```diff
@@ -227,11 +227,20 @@ func (d *Decoder) decodeValue(v interface{}, path []string) (Value, error) {
 			return d.decodeType(v.Content)
 
 		default:
-			return nil, fmt.Errorf("unsupported decoded tag: %d, %v", v.Number, v.Content)
+			return nil, fmt.Errorf(
+				"unsupported decoded tag (@ %s): %d, %v",
+				strings.Join(path, "."),
```
turbolent, line 232: I've made the error messages more descriptive by including the object graph path, which helps with debugging.

runtime/interpreter/encoding_test.go
```diff
@@ -358,6 +358,51 @@ func TestEncodeDecodeComposite(t *testing.T) {
 		testEncodeDecode(t,
 			encodeDecodeTest{
 				value: expected,
+				encoded: []byte{
```
turbolent, line 361: The diff is inferred incorrectly here, the actual change is duplicating the existing test and making it decode-only (to test backwards-compatibility), and adjusting the existing test to store the qualified identifier instead of the type identifier

runtime/interpreter/encoding_test.go
```diff
@@ -425,12 +470,15 @@ func TestEncodeDecodeComposite(t *testing.T) {
...
+					0x78, 0x1F,
+					// A.0000000000000001.SimpleStruct
 					0x41,
-					0x2E, 0x30, 0x78, 0x31,
-					0x2E, 0x53, 0x69, 0x6d, 0x70, 0x6c, 0x65, 0x53, 0x74, 0x72, 0x75, 0x63, 0x74,
+					0x2E,
+					0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30,
+					0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x31,
+					0x2E,
+					0x53, 0x69, 0x6d, 0x70, 0x6c, 0x65, 0x53, 0x74, 0x72, 0x75, 0x63, 0x74,
```
turbolent, line 481: The type identifier here was incorrect, so I fixed it (the type ID is `A.0000000000000001.SimpleStruct`, not `A.0x1.SimpleStruct`)

runtime/interpreter/encoding_test.go
```diff
@@ -463,6 +511,71 @@ func TestEncodeDecodeComposite(t *testing.T) {
 		testEncodeDecode(t,
 			encodeDecodeTest{
 				value: expected,
+				encoded: []byte{
```
turbolent, line 514: Same as above, the diff is inferred incorrectly, the actual change is duplicating the existing test and making it decode-only (to test backwards-compatibility), and adjusting the existing test to store the qualified identifier instead of the type identifier

runtime/interpreter/encoding_test.go
```diff
@@ -640,12 +757,6 @@ func TestEncodeDecodeComposite(t *testing.T) {
 					0x6a,
 					0x54, 0x65, 0x73, 0x74, 0x53, 0x74, 0x72, 0x75,
 					0x63, 0x74,
-					// key 1
```
turbolent, line 643: As you can see, the composite no longer stores key 1 in the CBOR map, which was the type identifier, but now stores key 4, the qualified identifier (see next hunk)

runtime/interpreter/encoding_test.go
```diff
@@ -3434,12 +3551,15 @@ func TestEncodeDecodeLinkValue(t *testing.T) {
...
+					// UTF-8 string, length 31
+					0x78, 0x1F,
+					// A.0000000000000001.SimpleStruct
 					0x41,
-					0x2E, 0x30, 0x78, 0x31,
-					0x2E, 0x53, 0x69, 0x6d, 0x70, 0x6c, 0x65, 0x53, 0x74, 0x72, 0x75, 0x63, 0x74,
+					0x2E,
+					0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30,
+					0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x31,
+					0x2E,
```
turbolent, line 3561: Like above, the type identifier was wrong so I fixed the test case

jordanschalm (approved):

PR #490, opened 2020-12-08
Improve fixed-point multiplication and division

turbolent:
Use `big.Int` to improve fixed-point multiplication and division.

See https://axiomzen.slack.com/archives/CG0B7CJAJ/p1607115540200500.


tarakby (approved):
Nice, thanks @turbolent 👌

dete (changes requested):
Looks good! Do you think we can move `Fix64MaxIntDividend` and `UFix64MaxIntDividend` into the testing files since they are only used there?

turbolent:
@dete Refactored in d8228adffeb8b6b9377e5b53cf5a884aa3cbe9f8

dete (approved):

PR #493, opened 2020-12-09
Add getType function 

### LINKED ISSUE #195: Get the run-time type of an object
### Issue To Be Solved

Using interfaces we can't check what type is really used at run-time, sometimes it's required to check a specific type. Need to know the full name of a type I'm working with.

I know you are working to add a new type `Type`, so maybe you can add something like `fun getType(): Type` and also a `fun toString(): String` to the `Type` if it doesn't exist now. 
I'd expect to call it in the following way: `obj.getType().toString()`. 
The format of a name of a type can be for instance: `T.{account_address}.{contract_name}.{type_name}`

turbolent:
Closes #195 

Add a function `fun getType(): Type` to all values. Implement it for values where a static type is known – for some values it is not currently possible to return a static type (e.g. arrays, dictionaries)

1. Make the static type in type value optional to support the case where a value has no known static type
2. Refactor the `Block` type to `sema`, and the value to `interpreter`. Add the missing dynamic type. This is only temporary – we should move it out of Cadence and inject it in the FVM package.
3. Add the missing static types (`AuthAccount`, `PublicAccount`, `DeployedContract`, etc.)

### DIFF
--- a/encoding/json/decode.go
+++ b/encoding/json/decode.go
@@ -598,8 +598,15 @@ func decodePath(valueJSON interface{}) cadence.Path {
 func decodeTypeValue(valueJSON interface{}) cadence.TypeValue {
 	obj := toObject(valueJSON)
 
+	var staticType string
+
+	staticTypeProperty, ok := obj[staticTypeKey]
+	if ok && staticTypeProperty != nil {
+		staticType = toString(staticTypeProperty)
+	}
+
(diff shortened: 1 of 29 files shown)


janezpodhostnik (approved):
Nice! The tests really helped me to understand the changes here.

ramtinms (approved):
LGTM

PR #503, opened 2020-12-14
Backport #440 to v0.10

turbolent:


runtime/ast/programindices.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (i *programIndices) initializer(declarations []Declaration) func() {
+	return func() {
+		i.init(declarations)
+	}
+}
+
+func (i *programIndices) init(declarations []Declaration) {
+
+	// Important: allocate instead of nil
```
jordanschalm, line 81: Why is it important?

runtime/ast/programindices.go
```diff
@@ -0,0 +1,112 @@
...
+	_interfaceDeclarations []*InterfaceDeclaration
+	// Use `compositeDeclarations` instead
+	_compositeDeclarations []*CompositeDeclaration
+	// Use `functionDeclarations()` instead
+	_functionDeclarations []*FunctionDeclaration
+	// Use `transactionDeclarations()` instead
+	_transactionDeclarations []*TransactionDeclaration
+}
+
+func (i *programIndices) pragmaDeclarations(declarations []Declaration) []*PragmaDeclaration {
```
jordanschalm, line 43: We are assuming here that these methods are always called with the same `[]Declarations`, otherwise the declarations cached by the `initializer` function will be inconsistent, which makes me a bit nervous. This method is called from the `Program` method like `p.indices.transactionDeclarations(p.Declarations)`, and `Declarations` is a public field of `Program`. Any code with access to a `Program` instance could change `Declarations` and make all these methods incorrect.

runtime/ast/programindices.go
turbolent: It used to be important before the refactor, because nil was used as the marker to know if the index had been created or not. The comment is now not necessary anymore.

runtime/ast/programindices.go
turbolent: This is correct, and I don't like it, though I'm not sure how to improve this in Go.
Basically all I want is lazy fields (for the indices) and having the `Declarations` field immutable. If you have any ideas, please let me know.

jordanschalm (approved):

runtime/ast/programindices.go
jordanschalm: 👍 Maybe making `Declarations` private and exposing a getter publicly would help? At least then it's not as easy to modify the field outside the package.

PR #506, opened 2020-12-16
Add presentation about implementation

turbolent:
Add a presentation which gives an overview of the past, present, and future implementation of Cadence


joshuahannan (approved):
The presentation looks good! Should we add something to the readme or the docs site to point to this presentation if people want to check it out?

turbolent:
@joshuahannan Good idea! I'll add it to the README and contribution guide

janezpodhostnik (approved):
Great presentation slides!

PR #540, opened 2021-01-19
Add support for multiple contracts per account to the language server 

### LINKED ISSUE #517: Contract import from account not recognized
### Problem

The VS code extension is not able to access the contract and resource definitions of an account, leading to errors being reported in transactions when they are none and the transactions executed correctly.

The editor will report three kind of errors related to this issue:

- on the report in question, it complains about not being able to find the declaration;
- on calling a function from the imported contract, it will report invalid type;
- on loading a type defined in the imported contract, it will report invalid type and that it can't infer type parameter.

### Steps to Reproduce

All you need to do is to copy `HelloWorld.cdc` and `Transaction1.cdc` into a project in VScode, start the emulator, deploy the `HelloWorld` contract to the first account and open the first transaction. Just make sure you use the correct accounts for deploying and transacting (it still will show the errors).

turbolent:
Closes #517

## Description

The language server was not supporting multiple contracts properly. The language server was still fetching account information and the deprecated `Code` field, which used to contain all code of an account, but is now empty.

Properly implement the two stage process of first resolving address locations + optional identifiers of the import statement to  multiple address locations. If the developer didn't specify any identifiers in the import statements (e.g. `import 0x1`), fetch all deployed contract names from the network. Finally, use the new `Contracts` field of the account instead of the deprecated `Code` field.

Also fix a crasher where the deployment codelens assumed that a contract declaration always exists, which is not the case (e.g. an empty program does not).

______

For contributor use:

### DIFF
--- a/languageserver/integration/codelenses.go
+++ b/languageserver/integration/codelenses.go
@@ -100,6 +100,10 @@ func (i *FlowIntegration) showDeployContractAction(
 
 	contract := program.SoleContractDeclaration()
 
+	if contract == nil {
+		return nil
+	}
+
 	name := contract.Identifier.Identifier
 
 	position := contract.StartPosition()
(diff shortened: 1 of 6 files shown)


janezpodhostnik (approved):
LGTM!

PR #541, opened 2021-01-20
[WIP] refactor runtime interface

ramtinms:
Closes #???

## Description

______

For contributor use:


runtime/interface.go
```diff
@@ -41,93 +43,110 @@ type Block struct {
...
-type HighLevelStorage interface {
-	Interface
+	ValueExists(address Address, key []byte, caller Location) (exists bool, err error)
 
-	// HighLevelStorageEnabled should return true
-	// if the functions of HighLevelStorage should be called,
-	// e.g. SetCadenceValue
-	HighLevelStorageEnabled() bool
+	// StoredKeys returns list of keys and their sizes owned by this account
+	StoredKeys(address Address, caller Location) (keys [][]byte, sizes []uint64, err error)
```
janezpodhostnik, line 78: I like this one a lot!

runtime/interface.go
```diff
@@ -41,93 +43,110 @@ type Block struct {
 type ResolvedLocation = sema.ResolvedLocation
 type Identifier = ast.Identifier
 type Location = common.Location
+type AddressLocation = common.AddressLocation
 
-type Interface interface {
-	// ResolveLocation resolves an import location.
-	ResolveLocation(identifiers []Identifier, location Location) ([]ResolvedLocation, error)
```
janezpodhostnik, line 47: What about this one?

runtime/runtime.go
```diff
@@ -75,25 +77,48 @@ func (c *Context) InitializeCodesAndPrograms() {
 	}
 }
 
+// Runnable can be run by a runner (including transactions, system operations, or read only scripts)
+// includes all the read-only data provided for the execution
```
janezpodhostnik, line 81: The data provided is readonly, but since this could be a transaction there could be changes to the state right?

runtime/interface.go
```diff
@@ -41,93 +43,110 @@ type Block struct {
...
+	// AccountContractCode returns the code associated with an account contract.
+	ContractCode(address AddressLocation) (code []byte, err error)
 	// UpdateAccountContractCode updates the code associated with an account contract.
-	UpdateAccountContractCode(address Address, name string, code []byte) (err error)
-	// GetAccountContractCode returns the code associated with an account contract.
-	GetAccountContractCode(address Address, name string) (code []byte, err error)
-	// RemoveAccountContractCode removes the code associated with an account contract.
-	RemoveAccountContractCode(address Address, name string) (err error)
-	// GetSigningAccounts returns the signing accounts.
-	GetSigningAccounts() ([]Address, error)
```
janezpodhostnik, line 71: What happened to this one?

runtime/interface.go
ramtinms: yeah was not sure where this should live, probably break into two parts, part inside the accounts, part inside the runtime

runtime/interface.go
ramtinms: it has been moved to Runnable (Authorizers() []Address) should provide the same

runtime/runtime.go
ramtinms: yeah, probably my note is misleading, this doesn't mean the operation is read-only, it means the values provided by runnable won't change in the middle of execution (such as source, arguments, authorizers)

runtime/interface.go
```diff
@@ -41,93 +43,110 @@ type Block struct {
...
 
-type Interface interface {
-	// ResolveLocation resolves an import location.
-	ResolveLocation(identifiers []Identifier, location Location) ([]ResolvedLocation, error)
-	// GetCode returns the code at a given location
-	GetCode(location Location) ([]byte, error)
-	// GetCachedProgram attempts to get a parsed program from a cache.
-	GetCachedProgram(Location) (*ast.Program, error)
-	// CacheProgram adds a parsed program to a cache.
-	CacheProgram(Location, *ast.Program) error
```
ramtinms, line 53: I removed these two

runtime/interface.go
```diff
@@ -41,93 +43,110 @@ type Block struct {
 type ResolvedLocation = sema.ResolvedLocation
 type Identifier = ast.Identifier
 type Location = common.Location
+type AddressLocation = common.AddressLocation
 
-type Interface interface {
-	// ResolveLocation resolves an import location.
-	ResolveLocation(identifiers []Identifier, location Location) ([]ResolvedLocation, error)
-	// GetCode returns the code at a given location
-	GetCode(location Location) ([]byte, error)
```
ramtinms, line 49: this returns the contract code right?

runtime/interface.go
ramtinms: Yeah, should remove the need for reading registers to compute storage changes

runtime/interface.go
ramtinms: we might be able to do it through Contracts()

turbolent: This is an abstract method that allows the host environment choose how to resolve locations used by the Cadence program. For exampl, a Cadence declaration like `import ServiceAccount` specifies an `IdentifierLocation`, which the host environment through this call might resolve to an `AddressLocation` for a service account.

A function that lists contract names for an address doesn't fully cover that functionality.

runtime/interface.go
turbolent: We have to figure out how to do caching properly. We don't want to remove it, but make it work and extend it to other objects, e.g. elaboration results from the checker.

runtime/interface.go
turbolent: Correct. What is the replacement for this?

runtime/interface.go
```diff
@@ -41,93 +43,124 @@ type Block struct {
...
+// fatal errors (e.g. Storage failure) are called by panic inside the methods
+// methods with power of edits have caller as an argument which can be used
+// to limit access to these method calls from smart contracts only, transaction only, or
+// specific smart contracts (e.g. service account).
+type Accounts interface {
+
+	// NewAccount creates a new account address and set the exists flag for this account
+	NewAccount(caller Location) (address Address, err error)
+	// Exists returns true if the account exists
+	Exists(address Address) (exists bool, err error)
```
turbolent, line 60: Maybe keep naming consistent and include `Account` like in the other method names.

```suggestion
	AccountExists(address Address) (exists bool, err error)
```

runtime/interface.go
```diff
@@ -41,93 +43,124 @@ type Block struct {
...
+	// NumberOfAccounts returns the number of accounts
+	NumberOfAccounts(caller Address) (count uint64, err error)
+
+	// SuspendAccount suspends an account (set suspend flag to true)
+	SuspendAccount(address Address, caller Location) error
+	// UnsuspendAccount unsuspend an account (set suspend flag to false)
+	UnsuspendAccount(address Address, caller Location) error
+	// returns true if account is suspended
+	IsSuspended(address Address) (isSuspended bool, err error)
```
turbolent, line 69: Maybe keep naming consistent and include `Account` like in the other method names.

```suggestion
	IsAccountSuspended(address Address) (isSuspended bool, err error)
```

runtime/interface.go
```diff
@@ -41,93 +43,124 @@ type Block struct {
...
-	GetSigningAccounts() ([]Address, error)
-	// Log logs a string.
-	Log(string) error
-	// EmitEvent is called when an event is emitted by the runtime.
-	EmitEvent(cadence.Event) error
+	UpdateContractCode(address AddressLocation, code []byte, caller Location) (err error)
+	// RemoveContractCode removes the code associated with an account contract.
+	RemoveContractCode(address AddressLocation, caller Location) (err error)
+	// Contracts returns a list of contract names under this account
+	Contracts(address AddressLocation, caller Location) (Name []string, err error)
```
turbolent, line 79: What is this for?

runtime/interface.go
turbolent: This will be very useful, nice!

The might get very large, do we maybe want to have this be some sort of iterator?

runtime/interface.go
```diff
@@ -41,93 +43,124 @@ type Block struct {
...
+
+	// NewAccount creates a new account address and set the exists flag for this account
+	NewAccount(caller Location) (address Address, err error)
+	// Exists returns true if the account exists
+	Exists(address Address) (exists bool, err error)
+	// NumberOfAccounts returns the number of accounts
+	NumberOfAccounts(caller Address) (count uint64, err error)
+
+	// SuspendAccount suspends an account (set suspend flag to true)
+	SuspendAccount(address Address, caller Location) error
```
turbolent, line 65: As discussed, it is not clear what a `caller Location` is, so maybe remove it here and below

runtime/interface.go
```diff
@@ -41,93 +43,124 @@ type Block struct {
...
-	SetCadenceValue(owner Address, key string, value cadence.Value) (err error)
+	// AddAccountKey appends a key to an account.
+	AddAccountKey(address Address, publicKey []byte, caller Location) error
+	// RemoveAccountKey removes a key from an account by index.
+	RevokeAccountKey(address Address, index uint, caller Location) error
+	// AccountPublicKey returns the account key for the given index
+	AccountPublicKey(address Address, index uint, caller Location) (publicKey []byte, err error)
+	// VerifyAccountSignature verifies a signature for the given address and index
+	// TODO RAMTIN do I need the tag here?
+	VerifyAccountSignature(address Address, index uint, signature []byte, tag string, signedData []byte, caller Location) (isValid bool, err error)
```
turbolent, line 102: It's useful to have this for account keys, but we still need a method for non-account keys too

runtime/interface.go
```diff
@@ -41,93 +43,124 @@ type Block struct {
...
+//
+// Results won't be directly callable by users
+// and will be used by the runtime to capture outputs
+type Results interface {
+	// AppendLog appends a log to the log collection
+	AppendLog(string) error
+	// Logs returns all the logs
+	Logs() ([]string, error)
+	// returns log i of the log collection
+	Log(i uint) (string, error)
```
turbolent, line 116: This probably only makes sense if you have a function that returns the number of logs. Once that exists, `Logs()` would just be a convenience and could be removed.

runtime/interface.go
```diff
@@ -41,93 +43,124 @@ type Block struct {
...
+
+	// AppendEvent appends an event to the event collection
+	AppendEvent(cadence.Event) error
+	// Events returns all the events
+	Events() ([]cadence.Event, error)
+	// returns event i of the event collection
+	Event(i uint) (cadence.Event, error)
```
turbolent, line 123: Same as above, maybe replace `Events() ([]cadence.Event, error)` with an `EventCount() (int, error)` function

runtime/interface.go
```diff
@@ -41,93 +43,124 @@ type Block struct {
...
+
+	// AppendError appends a non-fatal error
+	AppendError(error)
+	// return all the errors
+	Errors() multierror.Error
```
turbolent, line 128: Same as above, maybe keep the API consistent with events and logs, and replace this wuth `Error(i)` and `ErrorCount()` functions

runtime/interface.go
```diff
@@ -41,93 +43,124 @@ type Block struct {
...
+type AddressLocation = common.AddressLocation
+
+// Accounts manages changes to the accounts, account keys and account storage
+//
+// errors returns by methods are non-fatal ones (e.g. accountNotExist),
+// fatal errors (e.g. Storage failure) are called by panic inside the methods
+// methods with power of edits have caller as an argument which can be used
+// to limit access to these method calls from smart contracts only, transaction only, or
+// specific smart contracts (e.g. service account).
+type Accounts interface {
```
turbolent, line 55: Would be nice to split the storage functions into a separate interface.

runtime/interface.go
```diff
@@ -41,93 +43,124 @@ type Block struct {
...
+	// ValueEncoded captures the time spent on encoding a value
+	// TODO: maybe add type
 	ValueEncoded(duration time.Duration)
+	// ValueDecoded capture the time spent on decoding an encoded value
+	// TODO: maybe add type
 	ValueDecoded(duration time.Duration)
 }
 
+type HighLevelAccounts interface {
+	Accounts
```
turbolent, line 153: Would be nice to just have this as `HighLevelStorage` and not embed any other interface, it's stand-alone

runtime/runtime.go
```diff
@@ -75,25 +77,49 @@ func (c *Context) InitializeCodesAndPrograms() {
...
+	// Arguments returns arguments for this script
+	Arguments() [][]byte
+	// Authorizers returns a list address who has authorized this script
+	Authorizers() []Address
+	// ComputationLimit returns the max computation limit allowed while running
+	// Ramtin: (we might not need this to be passed and just be enforced in the Results)
+	ComputationLimit() uint64
+	// TODO: get rid of this decoding argument and merge it with arguments()
+	// DecodeArgument decodes a transaction argument against the given type.
+	DecodeArgument(argument []byte, argumentType cadence.Type) (cadence.Value, error)
```
turbolent, line 95: As discussed, maybe we can just pass decoded arguments directly, i.e. remove `DecodeArgument` and change `Arguments()` to `[]cadence.Value`

runtime/runtime.go
```diff
@@ -75,25 +77,49 @@ func (c *Context) InitializeCodesAndPrograms() {
...
+
+// Runner runs a "Runnable" and stores result into "Results".
+// It returns fatal errors
+// A runner augments the Cadence runtime with the host functionality.
+//
+// note that non-fatal runtime errors are captured inside the result
+// and error should only be used for returning fatal errors.
+type Runner interface {
+	// Run runs transactions
+	Run(Runnable, Context) error
```
turbolent, line 106:
```suggestion
	RunTransaction(Runnable, Context) error
```

runtime/runtime.go
```diff
@@ -75,25 +77,49 @@ func (c *Context) InitializeCodesAndPrograms() {
...
-	// or if the execution fails.
-	ExecuteTransaction(Script, Context) error
+	Runner
 
 	// ParseAndCheckProgram parses and checks the given code without executing the program.
 	//
 	// This function returns an error if the program contains any syntax or semantic errors.
 	ParseAndCheckProgram(source []byte, context Context) (*sema.Checker, error)
 
+	// RAMTIN: this can be moved to construction and doesn't have to be a method I guess,
```
turbolent, line 121: 👍

runtime/interface.go
ramtinms: for non-account keys we inject the verify method.

runtime/interface.go
ramtinms: I still feel this should be on the cadence side, the only part of the current implementation on the FVM that needs the env is [these two lines](https://github.com/onflow/flow-go/blob/master/fvm/env.go#L213-L214), the rest is operations can stay on the cadence side.

runtime/interface.go
ramtinms: Yeah, I tried that unfortunately there is a large dependency on the FVM side on the methods in this handler and since we would like to move to make keys to be resources, I think we can keep them here and long term remove them from this interface. what do you think ?

runtime/interface.go
ramtinms: done

runtime/interface.go
ramtinms: done

runtime/interface.go
ramtinms: what if we call the caller first location on top of stack calls?

runtime/interface.go
ramtinms: I have some ideas, I'll create a separate handler for caching for now until we discuss it.

runtime/interface.go
turbolent: Different host environments might handle this differently and hard-coding this logic has made this hard to extend in the past: Initially this function didn't exists, and we had added it for multiple contracts per account. Having it allows us for example in the future to rewrite certain imports easily, e.g. rewrite something like `import FungibleToken from ServiceAccount` to `import 0x123`

runtime/interface.go
turbolent: We can yes. I'm worried that this will get used in the wrong way. We're explicitly not adding such functionality to Cadence itself.

runtime/interface.go
turbolent: Sounds good! Can you please add a comment for this. I bet in the future someone (e.g. me, haha) will ask/wonder how Crypto contract verification is handled, looking for a `Verify` function here

runtime/interface.go
ramtinms: returns the list of contracts stored on this account, I saw an internal method call inside ResolveLocation thought might be useful for otherplaces as well.

runtime/interface.go
ramtinms: or maybe the smart contract address for smart contract calls and nil for transaction calls

runtime/interface.go
ramtinms: I see, I put it back then

runtime/interface.go
ramtinms: applied

runtime/interface.go
ramtinms: applied

runtime/interface.go
ramtinms: applied

runtime/interface.go
ramtinms: Just split them into smaller ones, you're right it is more helpful on the cadence side to keep them separate, even if the same account provider on the fvm side provides all functionalities.

runtime/interface.go
ramtinms: ContractCode(address AddressLocation)?

runtime/interface.go
ramtinms: changed it to an iterator

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
-	// GetCurrentBlockHeight returns the current block height.
-	GetCurrentBlockHeight() (uint64, error)
-	// GetBlockAtHeight returns the block at the given height.
-	GetBlockAtHeight(height uint64) (block Block, exists bool, err error)
-	// UnsafeRandom returns a random uint64, where the process of random number derivation is not cryptographically
-	// secure.
-	UnsafeRandom() (uint64, error)
+}
+
+type CryptoProvider interface {
```
ramtinms, line 169: I put this one back in for now, I thought we might actually need hash functionality for some cadence operations and it was a great deal of dependency that we can figure out later.

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
+	ValueExists(key StorageKey, caller Location) (exists bool, err error)
+	// StoredKeys returns list of keys and their sizes owned by this account
+	StoredKeys(address Address, caller Location) (iter StorageKeyIterator, err error)
+	// StorageUsed gets storage used in bytes by the address at the moment of the function call.
+	StorageUsed(address Address, caller Location) (value uint64, err error)
+	// Note: StorageCapacity has been moved to injected methods (similar to get balance)
+}
+
+// AccountKeys manages account keys
+type AccountKeys interface {
```
turbolent, line 99: Might be nice to add a count function? This would allow Cadence programs to iterate over all keys of an account.

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
+
+type LocationResolver interface {
 	// GetCode returns the code at a given location
 	GetCode(location Location) ([]byte, error)
+	// ResolveLocation resolves an import location.
+	ResolveLocation(identifiers []Identifier, location Location) ([]ResolvedLocation, error)
+}
+
+// Results are responsible to capture artifacts generated
+// when running the runnable
```
turbolent, line 116: "Runnable" isn't mentioned anywhere

```suggestion
// Results is an interface that is used to capture artifacts generated by executing a Cadence program. 
```

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
+//
+// Results won't be directly callable by users
+// and will be used by the runtime to capture outputs
```
turbolent, line 119:
```suggestion
// These functions won't be directly callable by users, but through Cadence language constructs,
// such as the `emit` statement to emit an event.
```

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
+	// Note: StorageCapacity has been moved to injected methods (similar to get balance)
+}
+
+// AccountKeys manages account keys
+type AccountKeys interface {
+	// AddAccountKey appends a key to an account.
+	AddAccountKey(address Address, publicKey []byte, caller Location) error
+	// RemoveAccountKey removes a key from an account by index.
+	RevokeAccountKey(address Address, index int, caller Location) (publicKey []byte, err error)
+	// AccountPublicKey returns the account key for the given index
```
turbolent, line 104:
```suggestion
	// AccountPublicKey returns the account key for the given index.
```

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
+	ResolveLocation(identifiers []Identifier, location Location) ([]ResolvedLocation, error)
+}
+
+// Results are responsible to capture artifacts generated
+// when running the runnable
+//
+// Results won't be directly callable by users
+// and will be used by the runtime to capture outputs
+type Results interface {
+	// AppendLog appends a log to the log collection
```
turbolent, line 121:
```suggestion
	// AppendLog appends a log to the log collection.
```

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
+	AppendLog(string) error
+	// Logs returns all the logs
```
turbolent, line 123:
```suggestion
	// Logs returns all logs that have been appended so far.
```

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
+	Logs() ([]string, error)
+	// returns log i of the log collection
+	LogAt(i uint) (string, error)
```
turbolent, line 126:
```suggestion
	// LogAt returns the log at the given index of the log collection.
	LogAt(index uint) (string, error)
```

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
+	// returns number of logs
+	LogCount() uint
```
turbolent, line 128:
```suggestion
	// LogCount returns number of logs in the log collection.
	LogCount() uint
```

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
+
+	// AppendEvent appends an event to the event collection
+	AppendEvent(cadence.Event) error
+	// Events returns all the events
+	Events() ([]cadence.Event, error)
+	// returns event i of the event collection
+	EventAt(i uint) (cadence.Event, error)
+	// returns number of events
+	EventCount() uint
```
turbolent, line 137:
```suggestion
	// AppendEvent appends an event to the event collection.
	AppendEvent(cadence.Event) error
	// Events returns all the events in the event collection.
	Events() ([]cadence.Event, error)
	// EventAt returns the event at the given index of the event collection
	EventAt(index uint) (cadence.Event, error)
	// EventCount returns number of events in the event collection.
	EventCount() uint
```

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
+
+	// AppendError appends a non-fatal error
+	AppendError(error) error
+	// return all the errors
+	Errors() multierror.Error
+	// returns event i of the event collection, the first error is the actual runtime error, the second error
+	// returns if there is any error while fetching the error i
+	ErrorAt(i uint) (Error, error)
+	// returns number of errors
+	ErrorCount() uint
```
turbolent, line 147: What is calling this append error function? Do you have an example use-case? 
Also, why is it a collection, wouldn't the program abort on the first error, as it does right now?

turbolent, line 147:
```suggestion
	// AppendError appends a non-fatal error to the error collection.
	AppendError(error) error
	// Errors returns all errors in the error collection.
	Errors() multierror.Error
	// ErrorAt returns the error at the given index of the error collection.
	// The first return value is the actual non-fatal error, 
	// the second return value indicates if there was an error while getting the error.
	ErrorAt(index uint) (Error, error)
	// ErrorCount returns number of errors in the error collection.
	ErrorCount() uint
```

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
+
+	// AddComputationUsed adds a new uint64 value to the computationUsed (computation accumulator)
+	AddComputationUsed(uint64) error
```
turbolent, line 150:
```suggestion
	// AddComputationUsed increases the computation usage accumulator by the given amount.
	AddComputationUsed(uint64) error
```

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
+	// ComputationSpent returns the total amount of computation spent during the execution
+	ComputationSpent() uint64
```
turbolent, line 152:
```suggestion
	// ComputationSpent returns the total amount of computation spent during the execution.
	ComputationSpent() uint64
```

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
+	// ComputationLimit returns the max computation limit allowed while running
+	// Ramtin: (we might not need this to be passed and just be enforced in the Results)
```
turbolent, line 154: Yes, this could also be enforced by `AddComputationUsed` returning an error when a certain limit is reached.

However, if we have this function, then we can expose it to Cadence programs. Dete had the idea to allow programs to potentially increase the limit and pay for it on the fly.

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
+	// returns if there is any error while fetching the error i
+	ErrorAt(i uint) (Error, error)
+	// returns number of errors
+	ErrorCount() uint
+
+	// AddComputationUsed adds a new uint64 value to the computationUsed (computation accumulator)
+	AddComputationUsed(uint64) error
+	// ComputationSpent returns the total amount of computation spent during the execution
+	ComputationSpent() uint64
+	// ComputationLimit returns the max computation limit allowed while running
```
turbolent, line 153:
```suggestion
	// ComputationLimit returns the computation limit, the maximum computation that may be used during execution.  
```

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
+	// Ramtin: (we might not need this to be passed and just be enforced in the Results)
+	ComputationLimit() uint64
+}
+
+// CacheProvider provides caching functionality to the cadence runtime
+//
+// cache should be tx failure aware (rollback changes)
+// cache should not break verification (register touches when returns)
+type CacheProvider interface {
```
turbolent, line 162:
```suggestion
// ProgramCache provides caching functionality for Cadence programs (ASTs).
//
type ProgramCache interface {
```

runtime/interface.go
turbolent: Yes, I think let's keep this in for now and then remove it in a subsequent PR by replacing them with injected functions.

runtime/interface.go
```diff
@@ -102,138 +179,131 @@ type Interface interface {
 	) (bool, error)
 	// Hash returns the digest of hashing the given data with using the given hash algorithm
```
turbolent, line 180:
```suggestion
	// Hash returns the digest of hashing the given data with using the given hash algorithm.
```

runtime/interface.go
```diff
@@ -102,138 +179,131 @@ type Interface interface {
...
-	SetCadenceValue(owner Address, key string, value cadence.Value) (err error)
 }
 
 type Metrics interface {
+	// ProgramParsed captures the time spent on parsing the program
 	ProgramParsed(location common.Location, duration time.Duration)
+	// ProgramChecked captures the time spent on checking the parsed program
 	ProgramChecked(location common.Location, duration time.Duration)
+	// ProgramInterpreted captures the time spent on interpreting the parsed and checked program
 	ProgramInterpreted(location common.Location, duration time.Duration)
```
turbolent, line 190:
```suggestion
	// ProgramParsed captures the time spent on parsing the program.
	ProgramParsed(location common.Location, duration time.Duration)
	// ProgramChecked captures the time spent on checking the parsed program.
	ProgramChecked(location common.Location, duration time.Duration)
	// ProgramInterpreted captures the time spent on interpreting the parsed and checked program.
	ProgramInterpreted(location common.Location, duration time.Duration)
```

runtime/interface.go
```diff
@@ -102,138 +179,131 @@ type Interface interface {
...
+
+	// ValueEncoded captures the time spent on encoding a value
+	// TODO: maybe add type
 	ValueEncoded(duration time.Duration)
+	// ValueDecoded capture the time spent on decoding an encoded value
+	// TODO: maybe add type
```
turbolent, line 196:
```suggestion
	// ValueEncoded captures the time spent on encoding a value.
	// TODO: maybe add type
	ValueEncoded(duration time.Duration)
	// ValueDecoded capture the time spent on decoding an encoded value.
	// TODO: maybe add type
```

runtime/interface.go
```diff
@@ -102,138 +179,131 @@ type Interface interface {
...
 type Metrics interface {
+	// ProgramParsed captures the time spent on parsing the program
 	ProgramParsed(location common.Location, duration time.Duration)
+	// ProgramChecked captures the time spent on checking the parsed program
 	ProgramChecked(location common.Location, duration time.Duration)
+	// ProgramInterpreted captures the time spent on interpreting the parsed and checked program
 	ProgramInterpreted(location common.Location, duration time.Duration)
+
+	// ValueEncoded captures the time spent on encoding a value
+	// TODO: maybe add type
```
turbolent, line 193: What "type" does the TODO refer to? The type of the encoded value?

runtime/interface.go
```diff
@@ -102,138 +179,131 @@ type Interface interface {
...
 
-func (i *EmptyRuntimeInterface) ResolveLocation(identifiers []Identifier, location Location) ([]ResolvedLocation, error) {
-	return []ResolvedLocation{
-		{
-			Location:    location,
-			Identifiers: identifiers,
-		},
-	}, nil
+	// SetCadenceValue sets a value for the given key in the storage, owned by the given account.
+	SetCadenceValue(owner Address, key string, value cadence.Value) (err error)
```
turbolent, line 207: Maybe keep consistency with the "low-level" storage interface: Maybe replace the `owner` and `key` parameters with a storage key parameter, like you did for the other storage functions above.

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
+// methods with power of edits have caller as an argument which can be used
+// to limit access to these method calls from smart contracts only, transaction only, or
+// specific smart contracts (e.g. service account).
+type Accounts interface {
+	// NewAccount creates a new account address and set the exists flag for this account
+	NewAccount(caller Location) (address Address, err error)
+	// AccountExists returns true if the account exists
+	AccountExists(address Address) (exists bool, err error)
+	// NumberOfAccounts returns the number of accounts
+	NumberOfAccounts(caller Location) (count uint64, err error)
```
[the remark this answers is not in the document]
ramtinms: there are two use case right now for this
- Dapper needs a way to figure out the number of accounts to be able to track them for fraud stuff.
- we need them for governance voting on smart contracts in the future.

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
 type Location = common.Location
+type AddressLocation = common.AddressLocation
+
+// Accounts manages changes to the accounts
+//
+// errors returns by methods are non-fatal ones (e.g. accountNotExist),
+// fatal errors (e.g. Storage failure) are called by panic inside the methods
+// methods with power of edits have caller as an argument which can be used
+// to limit access to these method calls from smart contracts only, transaction only, or
+// specific smart contracts (e.g. service account).
```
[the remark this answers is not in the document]
ramtinms: yeah caller is not working well and I'll take it out soon, we still need a way to limit access but we can figure it out later

runtime/interface.go
```diff
@@ -41,55 +43,130 @@ type Block struct {
...
+	AccountExists(address Address) (exists bool, err error)
+	// NumberOfAccounts returns the number of accounts
+	NumberOfAccounts(caller Location) (count uint64, err error)
+
+	// SuspendAccount suspends an account (set suspend flag to true)
+	SuspendAccount(address Address, caller Location) error
+	// UnsuspendAccount unsuspend an account (set suspend flag to false)
+	UnsuspendAccount(address Address, caller Location) error
+	// returns true if account is suspended
+	IsAccountSuspended(address Address) (isSuspended bool, err error)
```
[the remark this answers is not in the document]
ramtinms: let's sync on this

runtime/interface.go
ramtinms: yeah, just keeping some room for the future in case we want to return all the errors. but if it will always be one error I can change it.

runtime/interface.go
ramtinms: yeah something like that, more info on what was encoded or decoded. so we can have some info on what is costly right now on chain.

j1010001:
closing after quick check with @ramtinms

PR #549, opened 2021-01-22
Fix nested enum declarations

### LINKED ISSUE #548: BUG: Cannot use enums properly when declared in a contract
### Problem

When trying to use enums in a contract, the parser does not recognize the type of the enum assignment properly.

### Steps to Reproduce

Use this contract:
```

pub contract EnumTest {

    pub enum Phase: UInt8 {
        pub case disabled
        pub case enabled
    }

    access(account) var currentPhase: Phase

    init() {

        self.currentPhase = Phase.disabled
        self.currentPhase = Phase(rawValue: 0)
    
    }
}
```
2. See that the assignments in `init` don't work with these error messages:
First assignment:
```
value of type `((): EnumTest.Phase)` has no member `disabled`. unknown member
```
Second Assignment:
```
incorrect number of arguments. expected 0, got 1
```

### Acceptance Criteria

assignments should work properly

### Context

Blocking using enums in the Flow Epoch smart contracts

turbolent:
Closes #548 

## Description

When re-declaring the constructor for a nested composite declaration (e.g. an enum in a contract), then declare the constructor function appropriate for the composite kind.

______

For contributor use:

### DIFF
--- a/runtime/tests/interpreter/enum_test.go
+++ b/runtime/tests/interpreter/enum_test.go
@@ -202,8 +202,16 @@ func TestInterpretEnumInContract(t *testing.T) {
 		},
 	)
 
+	c := inter.Globals["C"].Value
+	require.IsType(t, &interpreter.CompositeValue{}, c)
+	contract := c.(*interpreter.CompositeValue)
+
+	e := contract.Fields["e"]
+	require.IsType(t, &interpreter.CompositeValue{}, c)
+	enumCase := e.(*interpreter.CompositeValue)
+
 	require.Equal(t,
-		nil,
-		inter.Globals["C"].Value,
+		interpreter.UInt8Value(0),
+		enumCase.Fields["rawValue"],
 	)
 }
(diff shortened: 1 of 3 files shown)


runtime/tests/interpreter/enum_test.go
```diff
@@ -175,3 +175,43 @@ func TestInterpretEnumInstance(t *testing.T) {
...
+			},
+		},
+	)
+
+	c := inter.Globals["C"].Value
+	require.IsType(t, &interpreter.CompositeValue{}, c)
+	contract := c.(*interpreter.CompositeValue)
+
+	e := contract.Fields["e"]
+	require.IsType(t, &interpreter.CompositeValue{}, c)
```
turbolent, line 210:
```suggestion
	require.IsType(t, &interpreter.CompositeValue{}, e)
```

janezpodhostnik (approved):
LGTM! I like the test cases, they explain the problem well.

Issue #436, opened 2020-11-02
Ensure iteration order is defined


turbolent:
### Issue To Be Solved

To make the checking process deterministic, non-deterministic constructs in GO must be avoided. 
One such construct in Go is the iteration over maps, which has no defined order.

Ensure there is no iteration over maps.

### Solution

- [x] Refactor iteration over maps to a defined iteration order
- [x] Find or write an analysis pass that prohibits iteration over maps and integrate it into CI 
  - [ ] Merge PR #450

PR #554, opened 2021-01-26
Optionally run checker tests concurrently

turbolent:
## Description

Add a flag to the checker tests which runs each check of a test multiple times, concurrently.
This allows the detection of data races (e.g. in the AST or elaboration) and non-determinism (order of checker errors).

______

For contributor use:

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -14,7 +14,7 @@
 
 This will automatically recompile the language server every time it is started.
 
-## How can data races in the checker be detected?
+## How is it possible to detect non-determinism and data races in the checker?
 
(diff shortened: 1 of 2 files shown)


janezpodhostnik (approved):
Does this speed up the tests? should this be the default?

turbolent:
No, this flag slows down the tests a bit, as with this flag on, checking is run 10 times instead of 1 time. The checks are run in parallel, so it's not 10x the time, but it's still noticeably slower.

I'd like to eventually turn this on by default, but this is blocked by #436: Currently some errors are reported in different order, so some tests randomly fail with this flag on.

PR #562, opened 2021-01-28
Use elaborations instead of checkers

turbolent:
Work towards dapperlabs/flow-go#5229

## Description

- Remove all sub-checkers from checker. Always query the import handler, leave potential caching up to the handler
- Remove checker input from the interpreter, it now only depends on the AST + elaboration
- Change interpreter imports: Remove all sub-checkers from the interpreter, 
  Instead of importing just a program, import other interpreters (`ProgramImport` -> `InterpreterImport`)
- Add elaboration caching to runtime interface
- Refactor runtime to new checker and interpreter import handlers
- Replace the notion of optional caching programs and elaborations with the *requirement* to store prepared programs and returning them when needed (like they were stored)

All tests are passing again, so it seems to work. 
However, I still want to clean this up a bit before opening it up as a proper PR.

NOTE: The diff is at times pretty weird, maybe view in split mode / only look at the new code.

______

For contributor use:

### DIFF
--- a/runtime/cmd/cmd.go
+++ b/runtime/cmd/cmd.go
@@ -79,6 +79,8 @@ var typeDeclarations = append(
 	stdlib.BuiltinTypes...,
 ).ToTypeDeclarations()
 
+var checkers = map[common.LocationID]*sema.Checker{}
+
 // PrepareChecker prepares and initializes a checker with a given code as a string,
 // and a filename which is used for pretty-printing errors, if any
 func PrepareChecker(
@@ -93,7 +95,7 @@ func PrepareChecker(
 		sema.WithPredeclaredValues(valueDeclarations.ToSemaValueDeclarations()),
 		sema.WithPredeclaredTypes(typeDeclarations),
 		sema.WithImportHandler(
-			func(checker *sema.Checker, importedLocation common.Location) (sema.Import, *sema.CheckerError) {
+			func(checker *sema.Checker, importedLocation common.Location) (sema.Import, error) {
 				stringLocation, ok := importedLocation.(common.StringLocation)
 
 				if !ok {
@@ -106,19 +108,15 @@ func PrepareChecker(
 					}
 				}
 
-				importChecker, err := checker.EnsureLoaded(
-					importedLocation,
-					func() *ast.Program {
-						imported, _ := PrepareProgramFromFile(stringLocation, codes)
-						return imported
-					},
-				)
-				if err != nil {
-					return nil, err
+				importedChecker, ok := checkers[importedLocation.ID()]
+				if !ok {
+					importedProgram, _ := PrepareProgramFromFile(stringLocation, codes)
+					importedChecker, _ = PrepareChecker(importedProgram, importedLocation, codes, must)
+					checkers[importedLocation.ID()] = importedChecker
 				}
 
-				return sema.CheckerImport{
-					Checker: importChecker,
+				return sema.ElaborationImport{
+					Elaboration: importedChecker.Elaboration,
 				}, nil
 			},
 		),
@@ -143,7 +141,8 @@ func PrepareInterpreter(filename string) (*interpreter.Interpreter, *sema.Checke
 	var uuid uint64
 
 	inter, err := interpreter.NewInterpreter(
-		checker,
+		interpreter.ProgramFromChecker(checker),
+		checker.Location,
 		interpreter.WithPredeclaredValues(valueDeclarations.ToInterpreterValueDeclarations()),
 		interpreter.WithUUIDHandler(func() (uint64, error) {
 			defer func() { uuid++ }()
--- a/runtime/cmd/execute/repl.go
+++ b/runtime/cmd/execute/repl.go
@@ -28,12 +28,9 @@ import (
 
 	"github.com/onflow/cadence"
 	"github.com/onflow/cadence/runtime"
-	"github.com/onflow/cadence/runtime/ast"
-	"github.com/onflow/cadence/runtime/cmd"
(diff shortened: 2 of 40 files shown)


runtime/cmd/execute/repl.go
```diff
@@ -64,38 +56,7 @@ func RunREPL() {
 
 			println(colorizeResult(value))
 		},
-		[]sema.Option{
```
turbolent, line 67: Moved to the REPL itself

runtime/context.go
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+package runtime
+
+import (
+	"github.com/onflow/cadence/runtime/ast"
+	"github.com/onflow/cadence/runtime/common"
+)
+
+type Context struct {
```
turbolent, line 26: Moved from runtime.go

runtime/runtime.go
```diff
@@ -43,38 +43,6 @@ type Script struct {
 	Arguments [][]byte
 }
 
-type Context struct {
```
turbolent, line 46: Moved to separate file

runtime/sema/entrypoint.go
```diff
@@ -24,30 +24,6 @@ import (
 
 const FunctionEntryPointName = "main"
 
-// FunctionEntryPointType returns the type of the entry point function declaration, if any.
-//
-// Returns an error if no valid entry point function declaration exists.
-//
-func (checker *Checker) FunctionEntryPointType() (*FunctionType, error) {
```
turbolent, line 31: Moved to elaboration, as this function only needs the elaboration, not the full checker

runtime/cmd/cmd.go
```diff
@@ -106,19 +108,15 @@ func PrepareChecker(
...
-						imported, _ := PrepareProgramFromFile(stringLocation, codes)
-						return imported
-					},
-				)
-				if err != nil {
-					return nil, err
+				importedChecker, ok := checkers[importedLocation.ID()]
+				if !ok {
+					importedProgram, _ := PrepareProgramFromFile(stringLocation, codes)
+					importedChecker, _ = PrepareChecker(importedProgram, importedLocation, codes, must)
```
ramtinms, line 114: should we return errors here ?

runtime/runtime.go
```diff
@@ -92,7 +60,7 @@ type Runtime interface {
 	// ParseAndCheckProgram parses and checks the given code without executing the program.
 	//
 	// This function returns an error if the program contains any syntax or semantic errors.
-	ParseAndCheckProgram(source []byte, context Context) (*sema.Checker, error)
+	ParseAndCheckProgram(source []byte, context Context) (*interpreter.Program, error)
```
ramtinms, line 63: We might need to update consensus nodes for this change

runtime/cmd/cmd.go
turbolent: Short answer: No, the functions panic on their own, the second return value is not an error.

These two functions, and some others, are written in an unidiomatic-Go style (not by me): The second return value is not an error – these functions already panic on their own if they fail, the second return value is a "must" function that can be used to panic in a way that has access to the code of all programs (`codes`), so that a pretty error can be printed to the terminal

runtime/runtime.go
turbolent: This function is neither used anywhere in Cadence, not anywhere in Flow Go. I think it used to be used maybe at some point, or just exists for convenience reasons.

runtime/interface.go
```diff
@@ -47,10 +47,17 @@ type Interface interface {
...
 	// GetCachedProgram attempts to get a parsed program from a cache.
 	GetCachedProgram(Location) (*ast.Program, error)
 	// CacheProgram adds a parsed program to a cache.
 	CacheProgram(Location, *ast.Program) error
+
+	// GetCachedElaboration attempts to get an elaboration from a cache.
+	GetCachedElaboration(Location) (*sema.Elaboration, error)
+	// CacheElaboration adds an elaboration to a cache.
+	CacheElaboration(Location, *sema.Elaboration) error
+
```
ramtinms: 👍

runtime/runtime.go
```diff
@@ -644,20 +716,73 @@ func (r *interpreterRuntime) parseAndCheckProgram(
...
+		wrapPanic(func() {
+			elaboration, err = context.Interface.GetCachedElaboration(context.Location)
+		})
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if elaboration == nil {
+		elaboration, err = r.check(program, context, functions, options, useCache)
```
ramtinms: just a minor comment that might be more readable if the cacheElaboration is here than inside the check, thoughts ?

ramtinms (approved):
Looks good to me, but I don't have that much context on the interaction between the program  and the checker.

runtime/runtime.go
turbolent: I think you commented on older code, this function is gone in the last commit. Maybe refresh to make sure you're looking at the latest code

janezpodhostnik (approved):

SupunS (approved):

turbolent:
@ramtinms @SupunS @janezpodhostnik @Kay-Zee Thank you for your reviews! 

To make it easier to review this I've opened three separate PRs, split by components. I'd appreciate if you could have another look there:

- Part 1: Checker #572
- Part 2: Interpreter #573
- Part 3: Runtime & rest #574 

Hopefully it's easier to review and it makes more sense. I'd close this PR in favor of these PRs if that sounds good to you.

Thanks again!

turbolent:
Closing in favor of #576

PR #566, opened 2021-01-30
Update to hamt 37930cf9f7d8, use ForEach instead of FirstRest

turbolent:
## Description

Switch from the inefficient `FirstRest` functions to the more efficient `ForEach`

______

For contributor use:

- [ ] Targeted PR against `master` branch – will open another PR once this one is merged

### DIFF
--- a/runtime/sema/variable_activations.go
+++ b/runtime/sema/variable_activations.go
@@ -161,23 +161,17 @@ func (a *VariableActivations) VariablesDeclaredInAndBelow(depth int) map[string]
 
 	activation := a.activations.CurrentOrNew()
 
-	var name string
-	var value interface{}
-
-	for {
-		name, value, activation = activation.FirstRest()
-		if name == "" {
-			break
-		}
-
+	_ = activation.ForEach(func(name string, value interface{}) error {
 		variable := value.(*Variable)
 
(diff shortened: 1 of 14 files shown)


runtime/sema/variable_activations.go
```diff
@@ -161,23 +161,17 @@ func (a *VariableActivations) VariablesDeclaredInAndBelow(depth int) map[string]
...
-		if name == "" {
-			break
-		}
-
+	_ = activation.ForEach(func(name string, value interface{}) error {
 		variable := value.(*Variable)
 
 		if variable.ActivationDepth < depth {
-			continue
+			return nil
```
ramtinms, line 168: just double-checking this is the intention here?

ramtinms (approved):

turbolent (commented):
@ramtinms Yes. Previously the `continue` statement caused the rest of the loop iteration to be skipped, and continued iteration.

The `return` without an error has the same effect now: The loop iteration is now in an anonymous function, so returning skips the rest of the "loop iteration", and returning `nil` will continue the iteration (an error would abort it, i.e behave like a `break`)

Kay-Zee (approved):
:shipit:

PR #567, opened 2021-01-30
Make checker occurrences and origins optional

turbolent:
## Description

Occurrences and origins are only needed by the language server. Disable them by default.

______

For contributor use:

- [ ] Targeted PR against `master` branch – I'll open another PR against master once this is merged

### DIFF
--- a/languageserver/server/server.go
+++ b/languageserver/server/server.go
@@ -1120,6 +1120,7 @@ func (s *Server) getDiagnostics(
 		runtime.FileLocation(uri),
 		sema.WithPredeclaredValues(valueDeclarations),
 		sema.WithPredeclaredTypes(typeDeclarations),
+		sema.WithOriginsAndOccurrencesEnabled(true),
 		sema.WithImportHandler(func(checker *sema.Checker, location ast.Location) (sema.Import, *sema.CheckerError) {
 			switch location {
 			case stdlib.CryptoChecker.Location:
--- a/runtime/sema/checker.go
+++ b/runtime/sema/checker.go
@@ -96,10 +96,11 @@ type Checker struct {
 	GlobalTypes                        map[string]*Variable
 	TransactionTypes                   []*TransactionType
 	inCondition                        bool
+	originsAndOccurrencesEnabled       bool
 	Occurrences                        *Occurrences
-	MemberAccesses                     *MemberAccesses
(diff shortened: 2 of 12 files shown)


runtime/sema/checker.go
```diff
@@ -1399,6 +1422,10 @@ func (checker *Checker) recordFunctionDeclarationOrigin(
 	function *ast.FunctionDeclaration,
 	functionType *FunctionType,
 ) *Origin {
+	if !checker.originsAndOccurrencesEnabled {
```
Kay-Zee, line 1425: Noticed that we actually did a check before using this function, but also do the check inside the function. Is that for extra safety?

languageserver/server/server.go
```diff
@@ -1120,6 +1120,7 @@ func (s *Server) getDiagnostics(
 		runtime.FileLocation(uri),
 		sema.WithPredeclaredValues(valueDeclarations),
 		sema.WithPredeclaredTypes(typeDeclarations),
+		sema.WithOriginsAndOccurrencesEnabled(true),
```
Kay-Zee, line 1123: It's default off, and only turned on for language server?

runtime/tests/checker/interface_test.go
```diff
@@ -1834,6 +1836,68 @@ func TestCheckContractInterfaceFungibleTokenConformance(t *testing.T) {
 	}
 }
 
+func BenchmarkContractInterfaceFungibleToken(b *testing.B) {
```
Kay-Zee, line 1839: Nice, did these show good results from the change?

Kay-Zee (approved):

runtime/sema/checker.go
turbolent: Yeah, this is just to be defensive

languageserver/server/server.go
turbolent: Correct, it's disabled by default, because it's not needed to determine all semantic errors.
The occurrences and origin functionality is only used by the language server to e.g. query type information for a specific location in the source code.

runtime/tests/checker/interface_test.go
turbolent: A bit, as there's less work done now, but not much.

psiemens (commented):
LGTM, nice find!

PR #571, opened 2021-02-01
Optimize activations

turbolent:
## Description

Replace the use of HAMT maps in activations with a plain map and a parent chain.
______

For contributor use:

### DIFF
--- a/runtime/activations/activations.go
+++ b/runtime/activations/activations.go
@@ -21,44 +21,63 @@
 //
 package activations
 
-import (
-	"github.com/raviqqe/hamt"
-
-	"github.com/onflow/cadence/runtime/common"
-)
-
-// An Activation is an immutable map of strings to arbitrary values.
+// An Activation is a map of strings to arbitrary values.
 // It can be used to represent an active scope in a program,
 // i.e. it can be used as a symbol table during semantic analysis,
 // or as an activation record during interpretation.
 //
-type Activation hamt.Map
+type Activation struct {
+	entries map[string]interface{}
+	Depth   int
+	Parent  *Activation
+}
 
-func NewActivation() Activation {
-	return Activation(hamt.NewMap())
+func NewActivation(parent *Activation) *Activation {
+	var depth int
+	if parent != nil {
+		depth = parent.Depth + 1
+	}
+	return &Activation{
+		entries: map[string]interface{}{},
+		Depth:   depth,
+		Parent:  parent,
+	}
 }
 
 // Find returns the value for a given key in the activation.
 // It returns nil if no value is found.
 //
-func (a Activation) Find(name string) interface{} {
-	return hamt.Map(a).Find(common.StringEntry(name))
+func (a *Activation) Find(name string) interface{} {
+	result, ok := a.entries[name]
+	if ok {
+		return result
+	}
+
+	if a.Parent != nil {
+		return a.Parent.Find(name)
+	}
+
+	return nil
 }
 
-// Insert inserts the given key-value pair into the activation.
+// Set sets the given key-value pair in the activation.
 //
-func (a Activation) Insert(name string, value interface{}) Activation {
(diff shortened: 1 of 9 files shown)


SupunS (approved):
LGTM!

runtime/activations/activations.go
```diff
@@ -21,44 +21,63 @@
...
 
 // ForEach calls the given function for each entry (key-value pair) in the activation.
 // It can be used to iterate over all entries of the activation.
 //
-func (a Activation) ForEach(f func(string, interface{}) error) error {
-	return hamt.Map(a).ForEach(func(entry hamt.Entry, v interface{}) error {
-		name := string(entry.(common.StringEntry))
-		return f(name, v)
-	})
+func (a *Activation) ForEach(cb func(string, interface{}) error) error {
```
dete, line 72: This doesn't look like it iterates over the entries in the parent. This will be tricky, because I assume you don't want to iterate over any entires in a parent that are "shadowed" in a descendent. Not sure how to do this efficiently, to be honest!

runtime/activations/activations.go
```diff
@@ -73,71 +92,58 @@ func (a Activation) ForEach(f func(string, interface{}) error) error {
...
 //
 func (a *Activations) Set(name string, value interface{}) {
-	current := a.current()
+	current := a.Current()
 	// create the first scope if there is no scope
 	if current == nil {
-		a.PushCurrent()
-		current = &a.activations[0]
+		current = NewActivation(nil)
+		a.Push(current)
```
dete, line 129: Not sure this is an improvement, but I would have put `PushNew()` here.

runtime/activations/activations.go
```diff
@@ -158,13 +164,13 @@ func (a *Activations) Pop() {
 // CurrentOrNew returns the current activation,
 // or if it does not exists, a new activation
 //
-func (a *Activations) CurrentOrNew() Activation {
-	current := a.current()
+func (a *Activations) CurrentOrNew() *Activation {
+	current := a.Current()
 	if current == nil {
-		return NewActivation()
+		return NewActivation(nil)
```
dete, line 170: It's surprising to me that if there is no current activation, that you don't create one _and make it the current_. Clearly this logic hasn't changed with this PR, but maybe a comment is warranted here?

runtime/interpreter/interpreter.go
```diff
@@ -2235,7 +2234,7 @@ func (interpreter *Interpreter) invokeInterpretedFunction(
 	// Start a new activation record.
 	// Lexical scope: use the function declaration's activation record,
 	// not the current one (which would be dynamic scope)
-	interpreter.activations.Push(function.Activation)
+	interpreter.activations.Push(activations.NewActivation(function.Activation))
```
dete, line 2237: This pattern repeats a lot. Could it be replaced with `PushNewWithParent(function.Activation)`?

runtime/sema/variable_activations.go
```diff
@@ -156,22 +156,21 @@ func (a *VariableActivations) DeclareImplicitConstant(
 	)
 }
 
-func (a *VariableActivations) VariablesDeclaredInAndBelow(depth int) map[string]*Variable {
-	variables := map[string]*Variable{}
+func (a *VariableActivations) ForEachVariablesDeclaredInAndBelow(depth int, f func(name string, value *Variable)) {
```
dete, line 159: Oh! I like this... Iterate in place instead of building a list and iterating over the list. 👍

runtime/activations/activations.go
turbolent: Refactored in acf5410

runtime/activations/activations.go
turbolent: Refactored in [`56e851c` (#571)](https://github.com/onflow/cadence/pull/571/commits/56e851c1ecfd496bc815e5fa6e21d3362936b5a8)

PR #572, opened 2021-02-01
Use elaborations instead of checkers – Part 1: Checker

turbolent:
## Description

This PR is the first smaller part of #562 which only contains the changes to the checker.

### DIFF
--- a/runtime/sema/checker.go
+++ b/runtime/sema/checker.go
@@ -267,6 +267,20 @@ func NewChecker(program *ast.Program, location common.Location, options ...Optio
 	return checker, nil
 }
 
+func (checker *Checker) SubChecker(program *ast.Program, location common.Location) (*Checker, error) {
+	return NewChecker(
+		program,
+		location,
+		WithPredeclaredValues(checker.PredeclaredValues),
+		WithPredeclaredTypes(checker.PredeclaredTypes),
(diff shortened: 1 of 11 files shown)


psiemens (approved):
👍 Looks good

runtime/sema/checker.go
```diff
@@ -283,13 +267,18 @@ func NewChecker(program *ast.Program, location common.Location, options ...Optio
...
 }
 
-// SetAllCheckers sets the given map of checkers as the map of all checkers.
-//
-func (checker *Checker) SetAllCheckers(allCheckers map[common.LocationID]*Checker) {
-	checker.AllCheckers = allCheckers
-
-	// Register self
-	checker.AllCheckers[checker.Location.ID()] = checker
+func (checker *Checker) SubChecker(program *ast.Program, location common.Location) (*Checker, error) {
```
SupunS, line 270: Is this function only used for tests?

SupunS (approved):

runtime/sema/checker.go
turbolent: It could be, I'm not sure. It's pretty much a convenience function to easily create a checker for a program which has the same options as the given checker. 

I guess the naming could be improved, though can't think of a better name.

PR #573, opened 2021-02-01
Use elaborations instead of checkers – Part 2: Interpreter

turbolent:
⚠️ Depends on #572

## Description

This PR is the second part of #562, which only contains the changes to the interpreter.


psiemens (approved):
Looks good to me. Nice job decoupling the interpreter from the checker!

SupunS (approved):

PR #574, opened 2021-02-01
Use elaborations instead of checkers – Part 3: Runtime & rest

turbolent:
## Description

This PR is the second part of #562, which only contains the changes to the runtime and rest of the PR.

This PR is still large, but not sure how to split up further – the many added lines are mostly in the tests.

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -181,7 +181,7 @@ func (r *interpreterRuntime) ExecuteScript(script Script, context Context) (cade
 		return nil, newError(err, context)
 	}
 
-	functionEntryPointType, err := checker.FunctionEntryPointType()
+	functionEntryPointType, err := checker.Elaboration.FunctionEntryPointType()
 	if err != nil {
 		return nil, newError(err, context)
 	}
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -60,7 +60,7 @@ type Runtime interface {
 	// ParseAndCheckProgram parses and checks the given code without executing the program.
 	//
 	// This function returns an error if the program contains any syntax or semantic errors.
-	ParseAndCheckProgram(source []byte, context Context) (*sema.Checker, error)
+	ParseAndCheckProgram(source []byte, context Context) (*interpreter.Program, error)
 
 	// SetCoverageReport activates reporting coverage in the given report.
 	// Passing nil disables coverage reporting (default).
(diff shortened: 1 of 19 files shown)


runtime/runtime_test.go
```diff
@@ -373,51 +379,153 @@ func TestRuntimeImport(t *testing.T) {
...
+			require.NoError(t, err)
 
-	assert.Equal(t, cadence.NewInt(42), value)
+			assert.Equal(t, cadence.NewInt(42), value)
+		}()
+	}
+	wg.Wait()
+
+	//TODO:
+	//require.Equal(t, concurrency+1, checkCount)
```
psiemens, line 499: What's the blocker here?

runtime/runtime.go
```diff
@@ -174,14 +142,27 @@ func (r *interpreterRuntime) ExecuteScript(script Script, context Context) (cade
 
 	runtimeStorage := newRuntimeStorage(context.Interface)
 
-	functions := r.standardLibraryFunctions(context, runtimeStorage)
+	var checkerOptions []sema.Option
+	var interpreterOptions []interpreter.Option
```
psiemens, line 146: Are these meant to be empty here?

psiemens (commented):
Looks good

SupunS (approved):

runtime/runtime_test.go
```diff
@@ -373,51 +379,153 @@ func TestRuntimeImport(t *testing.T) {
...
+		programChecked: func(location common.Location, duration time.Duration) {
+			checkCount += 1
+		},
+		setProgram: func(location Location, program *interpreter.Program) error {
+			programs[location.ID()] = program
+			return nil
+		},
+		getProgram: func(location Location) (*interpreter.Program, error) {
+			return programs[location.ID()], nil
+		},
```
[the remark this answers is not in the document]
turbolent: My thinking was that the functions are required to be implemented, and to be implemented correctly, favoring explicitness over implicitness, spending extra time to add it everywhere.

I seen now it's a lot of boilerplate, I've added a default implementation and removed all the new, repeated code in b9f7c73dffd4b35e0f02c6e0509b8cd07d471d0a

runtime/runtime_test.go
```diff
@@ -373,51 +379,153 @@ func TestRuntimeImport(t *testing.T) {
...
 	}
 
-	nextTransactionLocation := newTransactionLocationGenerator()
+	const transactionCount = 10
+	for i := 0; i < transactionCount; i++ {
 
-	value, err := runtime.ExecuteScript(
-		Script{
-			Source: script,
+		nextTransactionLocation := newTransactionLocationGenerator()
```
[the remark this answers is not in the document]
turbolent: Yes, good catch! Fixed in 7e9ab654634b64c97887794cc6ecd72852b2fbcd

runtime/runtime.go
turbolent: Yes, I declared those and gave them descriptive names, instead of just passing `nil` literals as arguments, to make it explicit what the arguments are / mean.

runtime/runtime_test.go
turbolent: Ideally we would expect the imported program only be checked once (`concurrency` transactions + 1 for the imported program), however, currently the imported program gets re-checked if it is currently being checked. 

I think this can be optimized by synchronizing the checking of a program using `sync`, but I felt like that was a future optimization that could be done in another PR.

I've added this explanation as a comment in 762985cf5693181aa6cfdde9461a67e0c09e869b

PR #577, opened 2021-02-03
Fix parsing 'from' keyword in imported-identifiers

### LINKED ISSUE #534: Cannot use 'from' as an identifier in multiple imports
### Steps to Reproduce

 Cannot use  'from' as a follow-up identifier in the import statement. Can use as the first identifier though.

```go
import from from 0x01cf0e2f2f715450    // works

import from, hello from 0x01cf0e2f2f715450    //  works

import hello, from from 0x01cf0e2f2f715450  // syntax error: `expected identifier, got keyword "from" `
```

SupunS:
Closes #534

## Description

With this change, if the  `from` keyword is found while parsing the identifiers in an import declaration, it will look-ahead to see whether this `from` token is part of the imported-identifiers list, or is it the `from` keyword that precede the location.

______

For contributor use:

### DIFF
--- a/runtime/parser2/declaration.go
+++ b/runtime/parser2/declaration.go
@@ -444,6 +444,14 @@ func parseImportDeclaration(p *parser) *ast.ImportDeclaration {
 				if p.current.Value == keywordFrom {
 
 					if !expectCommaOrFrom {
+						if isCurrentTokenAnImportedIdentifier(p) {
+							identifier := tokenToIdentifier(p.current)
+							identifiers = append(identifiers, identifier)
+
+							expectCommaOrFrom = true
+							break
+						}
+
 						panic(fmt.Errorf(
 							"expected %s, got keyword %q",
 							lexer.TokenIdentifier,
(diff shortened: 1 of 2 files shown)


runtime/parser2/declaration.go
```diff
@@ -441,30 +441,38 @@ func parseImportDeclaration(p *parser) *ast.ImportDeclaration {
 
 			case lexer.TokenIdentifier:
 
-				if p.current.Value == keywordFrom {
+				if p.current.Value != keywordFrom {
```
SupunS, line 444: Refactored this snippet to reduce the nested levels

runtime/parser2/declaration.go
```diff
@@ -562,6 +570,31 @@ func parseImportDeclaration(p *parser) *ast.ImportDeclaration {
...
+func isCurrentTokenAnImportedIdentifier(p *parser) bool {
+	p.startBuffering()
+
+	// skip the 'from' keyword.
+	p.next()
+	p.skipSpaceAndComments(true)
+
+	defer func() {
+		p.replayBuffered()
+	}()
```
turbolent, line 584: Should be possible to simplify this to

```suggestion
	defer p.replayBuffered()
```

turbolent: Also, maybe move this to the top, right after `p.startBuffering()`, so the start and stop are grouped

runtime/parser2/declaration.go
```diff
@@ -441,30 +441,38 @@ func parseImportDeclaration(p *parser) *ast.ImportDeclaration {
...
+						"expected %s, got keyword %q",
+						lexer.TokenIdentifier,
+						p.current.Value,
+					))
 				}
 
+				identifier := tokenToIdentifier(p.current)
+				identifiers = append(identifiers, identifier)
+
+				expectCommaOrFrom = true
```
turbolent, line 474: Looks like this code is now duplicated from above (the code in the `if p.current.Value != keywordFrom {` branch)

runtime/parser2/declaration.go
```diff
@@ -562,6 +570,31 @@ func parseImportDeclaration(p *parser) *ast.ImportDeclaration {
 	}
 }
 
+// isCurrentTokenAnImportedIdentifier check whether the current token
+// needs to be processed as an imported identifier or not.
+func isCurrentTokenAnImportedIdentifier(p *parser) bool {
```
turbolent, line 575: Given that this function is considering the next token (it skips the current), maybe rename this function

turbolent (commented):
Nice!

turbolent (approved):
Nice!

PR #581, opened 2021-02-04
Make CompositeType.Members field an ordered map

SupunS:
## Description

Work towards #436
- Introduces a new ordered map with `string` keys and `*Member` values
- Change the type of the `CompositeType.Members` field to the generated ordered map.

______

For contributor use:


turbolent:
@SupunS can you please rebase on `master`?

SupunS:
@turbolent Rebased to master.
I force-pushed some commits, that must have made the conflicts :/

runtime/convertTypes.go
```diff
@@ -191,9 +191,9 @@ func exportCompositeType(t *sema.CompositeType, results map[sema.TypeID]cadence.
 	fieldMembers := make([]*sema.Member, 0, len(t.Fields))
 
 	for _, identifier := range t.Fields {
-		member := t.Members[identifier]
+		member, ok := t.Members.Get(identifier)
 
-		if member.IgnoreInSerialization {
+		if !ok || member.IgnoreInSerialization {
```
turbolent, line 196: Is there ever a case where the lookup would fail?

runtime/sema/checker.go
```diff
@@ -998,13 +998,18 @@ func (checker *Checker) convertRestrictedType(t *ast.RestrictedType) Type {
...
 
 				// If there is an overlap in members, ensure the members have the same type
 
-				memberType := restrictionInterfaceType.Members[name].TypeAnnotation.Type
-				previousMemberType := previousDeclaringInterfaceType.Members[name].TypeAnnotation.Type
+				memberType := member.TypeAnnotation.Type
+
+				prevMemberType, ok := previousDeclaringInterfaceType.Members.Get(name)
+				if !ok {
+					return
```
turbolent, line 1010: AFAICS the member lookup should never fail, otherwise that's an implementation error, i.e. `previousDeclaringInterfaceType.Members` wasn't populated properly. Maybe `panic` with an unreachable error

runtime/stdlib/flow.go
```diff
@@ -275,14 +275,13 @@ func newFlowEventType(identifier string, parameters ...*sema.Parameter) *sema.Co
...
-				// TODO: add docstring support for parameters
-				"",
-			)
+		eventType.Members.Set(parameter.Identifier, sema.NewPublicConstantFieldMember(
+			eventType,
+			parameter.Identifier,
+			parameter.TypeAnnotation.Type,
+			// TODO: add docstring support for parameters
+			"",
+		))
```
turbolent, line 284:
```suggestion
		eventType.Members.Set(
			parameter.Identifier, 
			sema.NewPublicConstantFieldMember(
				eventType,
				parameter.Identifier,
				parameter.TypeAnnotation.Type,
				// TODO: add docstring support for parameters
				"",
			),
		)
```

runtime/tests/checker/composite_test.go
```diff
@@ -1744,8 +1744,8 @@ func TestCheckCompositeConstructorUseInInitializerAndFunction(t *testing.T) {
 				structureType.Identifier,
 			)
 
-			testFunctionMember := structureType.Members["test"]
-
+			testFunctionMember, ok := structureType.Members.Get("test")
+			require.True(t, ok)
```
turbolent, line 1748: 👍

turbolent (commented):
Nice! Just a couple questions / suggestions

turbolent:
@SupunS thanks for rebasing, looks good

runtime/convertTypes.go
SupunS: AFAIU `Fields` are a subset of `Members`. So the lookup should always succeed. 

Maybe a panic would be a good idea.

runtime/sema/checker.go
SupunS: Yeah agree. Added `panic` for all such similar places.

runtime/sema/check_composite_declaration.go
```diff
@@ -86,7 +86,11 @@ func (checker *Checker) visitCompositeDeclaration(declaration *ast.CompositeDecl
 
 		for _, field := range declaration.Members.Fields() {
 			fieldName := field.Identifier.Identifier
-			member := compositeType.Members[fieldName]
+			member, ok := compositeType.Members.Get(fieldName)
+			if !ok {
+				continue
```
turbolent, line 91: Maybe catch this case with 

```suggestion
				panic(errors.NewUnreachableError())
```

turbolent (approved):
Nice!

runtime/sema/check_composite_declaration.go
SupunS: I think we need to continue in this case, since some declarations can have fields that are not part of the type. (in negative scenarios)

eg:
```rust
enum Foo: Int {
    let a: Int
}
```
This fails with a semantic error in the downstream-code. But at this point, the filed `a` is part of the declaration, but not part of the type. Hence panicking here can crash the compiler.

turbolent: Oh, interesting, good catch!

PR #583, opened 2021-02-04
Return a dedicated error for encoding an unsupported value

turbolent:
## Description

Make it easier to handle this case.

______

For contributor use:

### DIFF
--- a/runtime/interpreter/encode.go
+++ b/runtime/interpreter/encode.go
@@ -417,6 +417,7 @@ func (e *Encoder) prepare(
 
 	default:
 		return nil, EncodingUnsupportedValueError{
+			Path:  path,
 			Value: v,
 		}
 	}
--- a/runtime/interpreter/encode.go
+++ b/runtime/interpreter/encode.go
@@ -416,7 +416,9 @@ func (e *Encoder) prepare(
 		return e.prepareTypeValue(v)
 
 	default:
-		return nil, fmt.Errorf("unsupported value: %[1]T, %[1]v", v)
+		return nil, EncodingUnsupportedValueError{
+			Value: v,
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


Kay-Zee (approved):

runtime/interpreter/errors.go
```diff
@@ -357,11 +357,21 @@ func (e UUIDUnavailableError) Error() string {
...
 	return fmt.Sprintf("failed to load type: %s", e.TypeID)
 }
+
+// EncodingUnsupportedValueError
+//
+type EncodingUnsupportedValueError struct {
+	Value Value
+}
+
+func (e EncodingUnsupportedValueError) Error() string {
```
m4ksio, line 375: Doesn't it need to be of FVM error (so implement `Code() uint32` as well) - so it will be passed as Tx error?

runtime/interpreter/encode.go
```diff
@@ -416,7 +416,9 @@ func (e *Encoder) prepare(
 		return e.prepareTypeValue(v)
 
 	default:
-		return nil, fmt.Errorf("unsupported value: %[1]T, %[1]v", v)
+		return nil, EncodingUnsupportedValueError{
```
janezpodhostnik, line 419: Should we add the path to the error as well?

turbolent: Good idea, doesn't hurt. Added in 9a02b9514b5f822db9560d461377de848f544d4e

janezpodhostnik (approved):

runtime/interpreter/errors.go
turbolent: This is Cadence, which doesn't know anything about FVM. This PR only makes it easier to look for such an error, without having to resort to looking at the error message.

turbolent: Once this PR is merged I'll update FVM and look for this error

PR #586, opened 2021-02-05
Revert dictionary key string format for address values

turbolent:
## Description

Revert the dictionary key string format back to [the format that was used as of v0.10.4](https://github.com/onflow/cadence/blob/v0.10.4-hamt-update/runtime/interpreter/value.go#L6322).

Ensure dictionaries which were temporarily creates using the different format can be decoded.

The migration does **not** support migrating resource dictionaries, i.e. dictionaries that have resource values, which are stored in separate storage keys. Error in this case for now. 
If we detect this error we need to also migrate deferred storage keys.

______

For contributor use:


runtime/interpreter/errors.go
```diff
@@ -375,7 +375,7 @@ type EncodingUnsupportedValueError struct {
 
 func (e EncodingUnsupportedValueError) Error() string {
 	return fmt.Sprintf(
-		"unsupported value to path [%s]: %[1]T, %[1]v",
+		"encoding unsupported value to path [%s]: %[2]T, %[2]v",
```
ramtinms, line 378: 👍

runtime/interpreter/decode.go
```diff
@@ -318,6 +319,58 @@ func (d *Decoder) decodeDictionary(v interface{}, path []string) (*DictionaryVal
...
+		for _, keyValue := range keys.Values {
+			keyAddressValue, ok := keyValue.(AddressValue)
+			if !ok {
+				continue
+			}
+
+			currentKeyString := keyAddressValue.KeyString()
+			wrongKeyString := hex.EncodeToString(keyAddressValue[:])
+
+			// Is the value for this key stored using the wrong key string format for address values?
```
ramtinms, line 345: we might add an extra condition to return early if currentKeyString exists

ramtinms (approved):

runtime/interpreter/decode.go
turbolent: Good point, refactored in 7ceb1ad

Kay-Zee (approved):

PR #589, opened 2021-02-08
Optimize member set by using plain Go maps and parent-child chaining

turbolent:
## Description

Refactor the member set from using HAMT sets to Go map and parent-child chaining, just like in activations.

This improves performance of `BenchmarkCheckContractInterfaceFungibleTokenConformance` by ~8%.
______

For contributor use:

### DIFF
--- a/runtime/interface.go
+++ b/runtime/interface.go
@@ -202,10 +202,6 @@ func (i *emptyRuntimeInterface) RemoveAccountKey(_ Address, _ int) (publicKey []
 	return nil, nil
 }
 
-func (i *emptyRuntimeInterface) UpdateAccountCode(_ Address, _ []byte) error {
-	return nil
-}
-
 func (i *emptyRuntimeInterface) UpdateAccountContractCode(_ Address, _ string, _ []byte) (err error) {
 	return nil
 }
--- a/runtime/interface.go
+++ b/runtime/interface.go
@@ -202,6 +202,10 @@ func (i *emptyRuntimeInterface) RemoveAccountKey(_ Address, _ int) (publicKey []
 	return nil, nil
 }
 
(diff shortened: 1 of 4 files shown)


runtime/interface.go
```diff
@@ -202,10 +202,6 @@ func (i *emptyRuntimeInterface) RemoveAccountKey(_ Address, _ int) (publicKey []
 	return nil, nil
 }
 
-func (i *emptyRuntimeInterface) UpdateAccountCode(_ Address, _ []byte) error {
```
janezpodhostnik, line 205: Not sure how this change connects to the rest of the changes.

turbolent: Oh, this is unrelated. I saw it was old, unused code, then accidentally committed the change as part of this PR.

turbolent: Reverted in 3ccad68

runtime/sema/member_set.go
```diff
@@ -18,60 +18,94 @@
...
+}
+
+// Intersection returns a new set containing all members that exist in this and the given set.
+//
+func (ms *MemberSet) Intersection(otherMS *MemberSet) *MemberSet {
+
+	result := NewMemberSet(nil)
+
+	_ = ms.ForEach(func(member *Member) error {
+		if !otherMS.Contains(member) {
```
SupunS, line 94: Just a minor suggestion: inverting the condition could reduce the number of return statements to  one.

SupunS (approved):

runtime/sema/member_set.go
turbolent: I usually try to structure code with early exits which avoids deep nesting. This also seems to be idiomatic / often seen in Go code.

PR #592, opened 2021-02-10
Optimize resource tracking

### LINKED ISSUE #590: Optimize resource tracking
### Issue To Be Solved

Resource tracking is currently implemented in an inefficient way.

### Suggested Solution

Optimize resource tracking by refactoring the data structures from using HAMT maps and sets to plain Go maps and parent-child chains.

turbolent:
Closes #590 

## Description

Optimize the resource tracking data structure by replacing the use of HAMT maps and sets with plain Go maps and implementing using parent-child chaining.

This removes the last traces of the HAMT library, which can now be removed.

I've added a diagram to visualize the behavior of cloning the resources map and how the child-sets are cloned as part of it.

These changes result locally in ~72% faster runs of `BenchmarkCheckContractInterfaceFungibleTokenConformance`.

______

For contributor use:


runtime/sema/resource_invalidations.go
```diff
@@ -37,33 +76,57 @@ func (ris ResourceInvalidations) All() (result []ResourceInvalidation) {
...
-	entry := ResourceInvalidationEntry{invalidation}
-	newInvalidations := ris.invalidations.Insert(entry)
-	return ResourceInvalidations{newInvalidations}
+// Delete removes the given resource invalidation from this set.
+//
+func (ris *ResourceInvalidations) Delete(invalidation ResourceInvalidation) {
+	if ris.invalidations == nil {
+		return
+	}
+	delete(ris.invalidations, invalidation)
```
SupunS, line 97: should it need to check and delete in the `ris.Parent` also?

There is a chance that `ris.Delete(x)` followed by a `ris.Contains(x)` might still return `true`

turbolent: Good catch/question! The semantics are a bit weird here.

In general, all functions should only ever operate on the current map/set (i.e. child), never the parent. The idea is that operations in a child environment, e.g. branches of an if-statement, don't effect the parent environment.

To properly implement this for deletion, we could mark the entry as deleted in the child (sort of like a "tombstone").

However, the delete function here only exists and is only used `Resources.RemoveTemporaryInvalidation`, which is called after a temporary invalidation was added to the resources map before. So all that is needed for this function is the current semantics (it's OK if the element was temporarily added before in the parent, then temporarily added in the child, then removed from the child, ... and now contains still returns true).

I don't think we have to implement proper deletion as described above, but rather, I'll rename the function and make it clear that it only applies to this child, and no deletion is performed for the parent.

SupunS (approved):

janezpodhostnik (approved):

runtime/sema/resources.txt
```diff
@@ -0,0 +1,35 @@
+
```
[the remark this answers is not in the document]
turbolent: No, I drew it in Monodraw and exported to SVG, PNG, and TXT

PR #593, opened 2021-02-10
Validate contract updates

### LINKED ISSUE #588: Detect invalid contract updates
### Issue To Be Solved

Currently a transaction may change a contract in a way that Cadence does not support.
We should detect and reject these cases.

### Suggested Solution

- [x] Diff ASTs, detect the following illegal cases
	- [x] Field added
	- [x] Field type changed
- [x] When a contract is updated, perform check, abort if illegal change is performed

Some valid cases:
  - Adding type
  - Removing a field
  - Changing a function (body or signature)

### Context

- `interpreterRuntime.newAuthAccountContractsChangeFunction`

SupunS:
Work towards #588

## Description

This PR introduces a validation step before updating a contract, to avoid any data inconsistencies.

:heavy_check_mark: This validate and prevents:
- Changing the type of a contract field.
- Adding a new contract field.
- Changing type of a field / adding a field, of a nested declaration
- Converting/updating a contract to a contract-interface, and vise-versa. Otherwise one can change from `contract` -> `contract-interface` -> `contract`, that may again cause field type incompatibility (i.e: indirect update to an incompatible state)
- Doing any of above for contract-interfaces (may be we can relax this later?).

:heavy_check_mark: Things that are allowed:
- Removing a field.
- Updating composite types that are not directly/indirectly referred by a contract field.
- changing functions (signature and content).

______

For contributor use:


runtime/contract_update_validation.go
```diff
@@ -0,0 +1,434 @@
...
+
+	// Then, check whether the fields of the two composite declarations
+	// referred by this nominal type, are also compatible.
+
+	var compositeDeclName string
+	if isQualifiedName(foundNominalType) {
+		compositeDeclName = foundNominalType.NestedIdentifiers[0].Identifier
+	} else {
+		compositeDeclName = foundNominalType.Identifier.Identifier
+	}
```
turbolent, line 189: This check assumes that nesting will always be one level deep. Maybe we can generalize this in case we allow other nesting of declarations in the future?

runtime/contract_update_validation.go
```diff
@@ -0,0 +1,434 @@
...
+		}
+	}
+
+	if validator.newNestedCompositeDecls == nil {
+		validator.newNestedCompositeDecls = map[string]*ast.CompositeDeclaration{}
+		for _, nestedComposite := range getDeclarationMembers(validator.newDeclaration).Composites() {
+			validator.newNestedCompositeDecls[nestedComposite.Identifier.Identifier] = nestedComposite
+		}
+	}
+}
```
turbolent, line 326: Maybe replace this function with another index and getter function on `memberIndices`, just like it has an index for functions by identifier, e.g `CompositesByIdentifier()`

runtime/contract_update_validation.go
```diff
@@ -0,0 +1,434 @@
...
+	}
+
+	newCompositeDecl := validator.newNestedCompositeDecls[compositeDeclName]
+	if newCompositeDecl == nil {
+		// If the declaration is not available, that means this is an imported contract.
+		// Thus, no need to validate anymore. Ideally shouldn't reach here.
+		return nil
+	}
+
+	return validator.checkDeclarationUpdatability(oldCompositeDecl, newCompositeDecl)
```
turbolent, line 207: I don't think this check is needed here. `CheckNominalTypeEquality` should just check the type equality.

The checking of nested composites declarations should recurse down into checking nested declaration.

runtime/contract_update_validation.go
```diff
@@ -0,0 +1,434 @@
...
+	foundConstSizedType, ok := found.(*ast.ConstantSizedType)
+	if !ok {
+		return validator.getTypeMismatchError(expected, found)
+	}
+
+	// Check size
+	if foundConstSizedType.Size.Value.Cmp(expected.Size.Value) != 0 ||
+		foundConstSizedType.Size.Base != expected.Size.Base {
+		return validator.getTypeMismatchError(expected, found)
+	}
```
turbolent, line 238: 👍

runtime/ast/type.go
```diff
@@ -64,6 +64,7 @@ type Type interface {
 	HasPosition
 	fmt.Stringer
 	isType()
+	Equal(other Type, checker TypeEqualityChecker) error
```
turbolent, line 67: Maybe rename this function to `CheckEqual`. `Equal` sounds like it just returns a boolean

runtime/contract_update_validation.go
```diff
@@ -0,0 +1,434 @@
...
+			return validator.getTypeMismatchError(expected, found)
+		}
+	}
+
+	return nil
+}
+
+func (validator *ContractUpdateValidator) CheckFunctionTypeEquality(_ *ast.FunctionType, _ ast.Type) error {
+	// Non storable type
+	panic(errors.NewUnreachableError())
```
turbolent, line 304: Don't panic but return an error

runtime/contract_update_validation.go
```diff
@@ -0,0 +1,434 @@
...
+}
+
+func (validator *ContractUpdateValidator) CheckReferenceTypeEquality(_ *ast.ReferenceType, _ ast.Type) error {
+	// Non storable type
+	panic(errors.NewUnreachableError())
```
turbolent, line 309: Don't panic but return an error

runtime/contract_update_validation.go
```diff
@@ -0,0 +1,434 @@
...
+		if expected[index] != element {
+			return false
+		}
+	}
+	return true
+}
+
+func isQualifiedName(nominalType *ast.NominalType) bool {
+	return len(nominalType.NestedIdentifiers) > 0
+}
```
turbolent, line 434: Maybe just add this as a function to `ast.NominalType` itself

runtime/contract_update_validation.go
```diff
@@ -0,0 +1,434 @@
...
+	case *ast.CompositeDeclaration:
+		return decl.Members
+	case *ast.InterfaceDeclaration:
+		return decl.Members
+	default:
+		panic(errors.NewUnreachableError())
+	}
+}
+
+func checkSliceEquality(expected []ast.Identifier, found []ast.Identifier) bool {
```
turbolent, line 419: Maybe rename `checkSliceEquality` to `identifiersEqual`

runtime/contract_update_validation.go
```diff
@@ -0,0 +1,434 @@
...
+		expectedType,
+		foundType)
+}
+
+// isInNestedDecl checks whether the currently visiting declaration is a nested one.
+func (validator *ContractUpdateValidator) isInNestedDecl() bool {
+	return validator.entryPointName != validator.currentDeclName
+}
+
+func getDeclarationName(declaration ast.Declaration) string {
```
turbolent, line 397: As mentioned above `getDeclarationName` can be removed and call-sites can use `ast.Declaration's DeclarationIdentifier()` instead

runtime/runtime.go
```diff
@@ -1732,6 +1732,19 @@ func (r *interpreterRuntime) newAuthAccountContractsChangeFunction(
...
 
+			if isUpdate {
+				validator := NewContractUpdateValidator(existingCode, program.Program)
+				err := validator.Validate()
+				if err != nil {
+					panic(&InvalidContractDeploymentError{
+						Err: fmt.Errorf("cannot update contract `%s` in account %s: %s",
+							nameArgument,
+							address.ShortHexWithPrefix(),
+							err.Error()),
```
turbolent, line 1743: `fmt.Errorf` has a special formatting directive for nested errors, `%w`:

```suggestion
						Err: fmt.Errorf("cannot update contract `%s` in account %s: %w",
							nameArgument,
							address.ShortHexWithPrefix(),
							err,
						),
```

runtime/runtime.go
```diff
@@ -1732,6 +1732,19 @@ func (r *interpreterRuntime) newAuthAccountContractsChangeFunction(
 				))
 			}
 
+			if isUpdate {
+				validator := NewContractUpdateValidator(existingCode, program.Program)
+				err := validator.Validate()
```
turbolent, line 1737: 👏

runtime/runtime_test.go
```diff
@@ -6122,3 +6122,643 @@ func TestPanics(t *testing.T) {
...
+				Location:  nextTransactionLocation(),
+			},
+		)
+		return err
+	}
+
+	t.Run("change field type", func(t *testing.T) {
+		const oldCode = `
+      		pub contract Test1 {
+          		pub var a: String
```
turbolent, line 6179: Indentation is mixed spaces/tabs here, normalizing to one or the other makes this more readable

runtime/runtime_test.go
```diff
@@ -6122,3 +6122,643 @@ func TestPanics(t *testing.T) {
 	)
 	assert.Error(t, err)
 }
+
+func TestContractUpdateValidation(t *testing.T) {
```
turbolent, line 6126: Great tests! 👏

turbolent (commented):
Nice work! Code looks overall great, just a few comments on re-using existing code and code organization.

runtime/contract_update_validation.go
SupunS: Yeah, I think we should recursively visit nested declarations as well. That will solve [this problem](https://github.com/onflow/cadence/issues/588#issuecomment-776628175) as well.

runtime/contract_update_validation.go
```diff
@@ -0,0 +1,434 @@
...
+	case *ast.CompositeDeclaration:
+		return decl.Identifier.Identifier
+	case *ast.InterfaceDeclaration:
+		return decl.Identifier.Identifier
+	default:
+		panic(errors.NewUnreachableError())
+	}
+}
+
+func getDeclarationMembers(declaration ast.Declaration) *ast.Members {
```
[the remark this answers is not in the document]
SupunS: I din't add it to the interface because there are many other implementations (like FieldDeclaration, ImportDeclaration, etc) that actually don't have members. 

We can still add it and return nil for those who don't have members. But feels that might misinterpret the interface/API.

runtime/contract_update_validation.go
```diff
@@ -0,0 +1,434 @@
...
+
+		return fmt.Errorf("too many fields. expected %d, found %d",
+			len(oldFields), len(newFields))
+	}
+
+	// Put the old field-types against their field-names to a map, for faster lookup.
+	oldFiledTypes := map[string]ast.Type{}
+	for _, field := range oldFields {
+		oldFiledTypes[field.Identifier.Identifier] = field.TypeAnnotation.Type
+	}
```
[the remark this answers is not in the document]
SupunS: nice!

runtime/contract_update_validation.go
```diff
@@ -0,0 +1,434 @@
...
+	// Updated contract has to have at-most the same number of field as the old contract.
+	// Any additional field may cause crashes/garbage-values when deserializing the
+	// already-stored data. However, having less number of fields is fine for now. It will
+	// only leave some unused-data, and will not do any harm to the programs that are running.
+
+	// This is a fail-fast check.
+	if len(oldFields) < len(newFields) {
+		if validator.isInNestedDecl() {
+			return fmt.Errorf("too many fields in `%s`. expected %d, found %d",
+				validator.currentDeclName, len(oldFields), len(newFields))
```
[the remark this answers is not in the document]
SupunS: Added dedicated error types for validation errors.

Will send a separate PR for reporting all errors at once.

runtime/contract_update_validation.go
SupunS: Ideally we shouldn't reach here, because the checker  already validates this, before reaching here.

Anyway updated to return an error to be on the safe side.

runtime/contract_update_validation.go
SupunS: added

runtime/contract_update_validation.go
SupunS: this snippet is no longer used now.

SupunS:
@turbolent Thanks for the suggestions! Add the required changes.

runtime/contract_update_validation.go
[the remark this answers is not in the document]
SupunS: Opened a PR #597 (still WIP) for improving error reporting. 

It tackles reporting all the validation errors at once, and also improve errors in overall.

runtime/contract_update_validation.go
```diff
@@ -0,0 +1,456 @@
...
+	err = validator.checkNestedDeclarations(oldDeclaration, newDeclaration)
+	if err != nil {
+		return err
+	}
+
+	switch newDecl := newDeclaration.(type) {
+	case *ast.CompositeDeclaration:
+		oldDecl := oldDeclaration.(*ast.CompositeDeclaration)
+		return validator.checkConformances(oldDecl, newDecl)
+	}
```
turbolent, line 104: The type-switch with the single case can be shortened to an if-statement.
Also, you maybe want to also do the conditional cast for the old declaration.
Finally, maybe only return if an error occurs instead of always:
There might be code added in the future after this check, which would be skipped 

```suggestion
	if  newDecl, ok := newDeclaration.(*ast.CompositeDeclaration); ok {
		if oldDecl, ok := oldDeclaration.(*ast.CompositeDeclaration); ok {
			err := validator.checkConformances(oldDecl, newDecl)
			if err != nil {
				return err
			}
		}
	}
```

runtime/ast/declaration.go
```diff
@@ -26,4 +26,5 @@ type Declaration interface {
 	DeclarationIdentifier() *Identifier
 	DeclarationKind() common.DeclarationKind
 	DeclarationAccess() Access
+	DeclarationMembers() *Members
```
turbolent, line 29: 👍

runtime/contract_update_validation.go
```diff
@@ -0,0 +1,456 @@
...
+
+func (validator *ContractUpdateValidator) checkFields(
+	oldDeclaration ast.Declaration,
+	newDeclaration ast.Declaration,
+) error {
+
+	oldFields := oldDeclaration.DeclarationMembers().FieldsByIdentifier()
+	newFields := newDeclaration.DeclarationMembers().Fields()
+
+	// Updated contract has to have at-most the same number of field as the old contract.
```
turbolent, line 117: 👍

runtime/contract_update_validation.go
```diff
@@ -0,0 +1,456 @@
...
+
+func (validator *ContractUpdateValidator) checkNestedDeclarations(
+	oldDeclaration ast.Declaration,
+	newDeclaration ast.Declaration,
+) error {
+
+	oldNestedCompositeDecls := oldDeclaration.DeclarationMembers().CompositesByIdentifier()
+	oldNestedInterfaceDecls := oldDeclaration.DeclarationMembers().InterfacesByIdentifier()
+
+	getCompositeOrInterfaceDecl := func(name string) (ast.Declaration, bool) {
```
turbolent, line 179: Maybe indicate that this function looks up the composite or interface declaration in the old nested declarations, e.g. rename to `getOldCompositeOrInterfaceDecl`

runtime/contract_update_validation.go
```diff
@@ -0,0 +1,456 @@
...
+	}
+
+	return nil
+}
+
+func getTypeMismatchError(expectedType ast.Type, foundType ast.Type) *TypeMismatchError {
+	return &TypeMismatchError{
+		expectedType: expectedType,
+		foundType:    foundType,
+		Range:        ast.NewRangeFromPositioned(expectedType),
```
turbolent, line 425: The error position should be for the new type, not the old one:

```suggestion
		Range:        ast.NewRangeFromPositioned(foundType),
```

runtime/contract_update_validation.go
```diff
@@ -0,0 +1,456 @@
...
+			Range:    ast.NewRangeFromPositioned(newEnum.Identifier),
+		}
+	}
+
+	for index, conformance := range oldConformances {
+		err := conformance.CheckEqual(newConformances[index], validator)
+		if err != nil {
+			return &ConformanceMismatchError{
+				err:   err,
+				Range: ast.NewRangeFromPositioned(conformance),
```
turbolent, line 413: The range should be for the new conformance. Maybe declare `newConformance := newConformances[index]` and use it in `CheckEqual` above and here instead of `conformance`

runtime/contract_update_validation.go
```diff
@@ -0,0 +1,456 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (validator *ContractUpdateValidator) checkField(
+	newField *ast.FieldDeclaration,
+	oldField *ast.FieldDeclaration,
```
turbolent, line 154: All other check function have parameters ordered old-new instead of new-old. Maybe swap

turbolent (commented):
Great refactor! Just a couple last comments and after that it should be ready to go

turbolent (approved):
Very nice! 👏

PR #594, opened 2021-02-10
Minor additions to the docs

janezpodhostnik:
## Description

While using Cadence I notice there are some things I couldn't find in the docs:
- what is the type of a resource dictionary/resource
- if uuid is available in the `init()` method

______

For contributor use:

### DIFF
--- a/docs/language/composite-types.mdx
+++ b/docs/language/composite-types.mdx
@@ -1125,8 +1125,9 @@ resource R {}
 
 // Declare a constant for an array of resources.
 // Create two resources and move them into the array.
+// `resources` has type `@[R]`
 //
-let resources: @[R] <- [
+let resources <- [
     <-create R(),
     <-create R()
 ]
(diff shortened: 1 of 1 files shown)


docs/language/composite-types.mdx
```diff
@@ -1126,7 +1126,7 @@ resource R {}
 // Declare a constant for an array of resources.
 // Create two resources and move them into the array.
 //
-let resources <- [
+let resources: @[R] <- [
```
turbolent, line 1129: The type annotation isn't necessary here, the type of the declaration can be inferred from the value.
In the other examples I used something like this below the declaration: 

```
// `resources` has type `@[R]`
```

docs/language/composite-types.mdx
```diff
@@ -1301,7 +1301,7 @@ destroy resources
 Resources have an implicit unique identifier associated with them,
 implemented by a predeclared public field `let uuid: UInt64` on each resource.
 
-This identifier will be automatically set when the resource is created,
+This identifier will be automatically set when the resource is created (before the resources `init` function is called),
```
turbolent, line 1304:
```suggestion
This identifier will be automatically set when the resource is created, before the resource's initializer is called
(i.e. the identifier can be used in the initializer),
```

turbolent (commented):
Nice, thank you!

turbolent (approved):

PR #603, opened 2021-02-16
Make resource tracking deterministic 

turbolent:
Work towards #436

## Description

- Use an ordered map

______

For contributor use:


runtime/sema/resources.go
```diff
@@ -111,19 +111,20 @@ func (ris *Resources) String() string {
...
+	info, _ := ris.resources.Get(resource)
+	return info
 }
 
 // AddInvalidation adds the given invalidation to the set of invalidations for the given resource.
 // If the invalidation is not temporary, marks the resource to be definitely invalidated.
 //
 func (ris *Resources) AddInvalidation(resource interface{}, invalidation ResourceInvalidation) {
-	info := ris.resources[resource]
+	info, _ := ris.resources.Get(resource)
```
SupunS, line 122: can the `resource` be unavailable at this point, causing a NPD?

turbolent: Yes, and in that case `Get` will return the default value for the value type, in this case an empty `ResourceInfo`

PR #616, opened 2021-02-22
Support nested file imports in LSP

psiemens:
## Description

I noticed a small bug that occurs when using nested imports in the VS Code extension (e.g. import a file that imports another file). 

Here's an example:

```
contracts/
  MyContractA.cdc
  MyContractB.cdc
transactions/
  myTransaction.cdc
```
```
// transactions/myTransaction.cdc

import MyContractA from "../contracts/MyContractA.cdc"
```

```
// contracts/MyContractA.cdc

import MyContractB from "./MyContractB.cdc"
```

In this case, when resolving the import of `MyContractB`, the `mainPath` value is still `transactions/myTransaction.cdc`, so the computed absolute file path (I call this path normalization) is `transactions/MyContractB.cdc`, which fails.

This PR updates the import logic to normalize paths _before_ setting the location on the sub-checker for the imported program. This way, each value of `checker.Location` points to the absolute path for that file. Then, any further nested imports are normalized against `checker.Location`, allowing for arbitrary depth of imports.

Disclaimer: I have only tested this code with file imports and have not yet tested extensively enough to ensure I didn't introduce a regression with address-based imports. I'll try to find time for that today.
______

For contributor use:


languageserver/server/server.go
```diff
@@ -1429,6 +1438,50 @@ func (s *Server) resolveImport(
...
+
+	return string(stringLocation)
+}
+
+func uriToLocation(uri protocol.DocumentUri) common.StringLocation {
+	s := string(uri)
+
+	if strings.HasPrefix(s, filePrefix) {
+		return common.StringLocation(s[len(filePrefix):])
+	}
```
turbolent, line 1480: Can use `strings.TrimPrefix` here

runtime/sema/errors.go
```diff
@@ -1121,8 +1121,9 @@ type ImportedProgramError struct {
 
 func (e *ImportedProgramError) Error() string {
 	return fmt.Sprintf(
-		"checking of imported program `%s` failed",
+		"checking of imported program `%s` failed: %s",
 		e.Location,
+		e.CheckerError,
```
turbolent, line 1126: I'm not quite sure this is OK. Normally the errors only contain the error message of themselves, and nested errors are unrolled separately by e.g. the error pretty printer. 

Including the nested error directly in the error message removes control over how the nested errors are reported (separately, just like the other errors of this current location).

turbolent (commented):
Looks good, thank you for fixing the local file resolution, I hadn't realized it wasn't handling relative paths properly 👍 

Just one concern, and there seem to be build errors

runtime/sema/errors.go
psiemens: Ah sorry! I had included this line for my own debugging, didn't meant to commit it 😬 

Completely agree -- I'll remove

turbolent (approved):
Very nice! Thank you for fixing this 🙏

turbolent (approved):

PR #624, opened 2021-02-24
Refactor AuthAccount, PublicAccount, and Block type to singleton

turbolent:


runtime/sema/nominal_type.go
```diff
@@ -109,3 +111,11 @@ func (t *NominalType) initializeMembers() {
 		t.members = withBuiltinMembers(t, members)
 	})
 }
+
+func (t *NominalType) isContainerType() bool {
+	return t.NestedTypes != nil
```
SupunS, line 116: Should this be as:
```suggestion
	return t.IsContainerType
```

runtime/sema/authaccount_type.go
```diff
@@ -0,0 +1,575 @@
...
+//
+var AuthAccountType = &NominalType{
+	Name:                 "AuthAccount",
+	QualifiedName:        "AuthAccount",
+	TypeID:               "AuthAccount",
+	IsInvalid:            false,
+	IsResource:           false,
+	Storable:             false,
+	Equatable:            false,
+	ExternallyReturnable: false,
```
SupunS, line 38: Probably make `IsContainerType : true` as well?

runtime/sema/nominal_type.go
SupunS: Or perhaps, we don't need `IsContainerType` at all, can simply rely on the `t.NestedTypes`. But then we might have to check both nil and emptiness.

turbolent: Oh, right, I think I had added `IsContainerType` at first, then changed my mind mid-way and thought we could just rely on `NestedTypes != nil` and didn't remove `IsContainerType`. 

Good catch!

turbolent: Removed the unused field in 67f13e7d1723c914c3e15a13fec63ad4e1e5d44a

runtime/sema/authaccount_type.go
turbolent: Removed the unused field in 67f13e7d1723c914c3e15a13fec63ad4e1e5d44a

SupunS (approved):

Issue #629, opened 2021-02-25
Add documentation for the new Account Key API


SupunS:
This is a sub-task of #371

## Instructions

Add Documentation for the new Account Key API.

Issue #582, opened 2021-02-04
New crypto features support


tarakby:
### Issue To Be Solved

- add new features to the [crypto contract]( https://github.com/onflow/cadence/blob/master/runtime/stdlib/contracts/crypto.cdc), as required by the upcoming Flow protocol epoch features:
   - validation of public keys
   - support the signature scheme BLS and its corresponding hashing algorithm.
   
### Suggest A Solution

This is modified version of the file `crypto.cdc`. Many functions are copied from the original code but I pasted the entire file for a better clarity, as there a few changes here and there. I've added some comments in the code to help understand the changes. 

This code is a starting point to discuss and improve, I'm open to all suggestions making the APIs easier to use while keeping all the features needed for the epoch switch implementation in Flow protocol.

```
// the functions outside the crypto contract are meant to be implemented by the fvm 
// in Go. I can implement/support implementing these functions.

// the first 2 functions could be part of a same interface 
//(as they are both related to signature algos)

// verify is a low level signature verification function,
// that uses a signature scheme to verify a signature under hash(tag, data) and a public key.
pub fun verify(
    signature: [UInt8],
    tag: String,
    signedData: [UInt8],
    publicKey: [UInt8],
    signatureAlgorithm: String,
    hashAlgorithm: String
): Bool

// validatePublicKey is a low level function that checks if the input array 
// is a valid serialization of a public key of the given signature scheme.
pub fun validatePublicKey(
    publicKey: [UInt8],
    algorithm: String
): Bool

// low level hash function that computes hash(tag, data). 
// tag is mixed with data in a way dependent of the hash algo, this can be detailed in a doc later. 
// for instance, sha2(tag, data) = sha2(tag || data)
pub fun hash(
        data: [UInt8],
        tag: String,
        algorithm: String
    ): [UInt8] 


pub contract Crypto {

    // SignatureAlgorithm 
    pub struct SignatureAlgorithm {
        pub let name: String
        init(name: String) {       }
        
        // a method to validate a public key serialization
        pub fun validatePublicKey(publicKey: [UInt8]): Bool {
            return validatePublicKey(publicKey: publicKey, algorithm:self.name)
        }
        
        // a method to verify a signature
        pub fun verify(        
            signature: [UInt8],
            tag: String,
            signedData: [UInt8],
            publicKey: [UInt8],
            hashAlgorithm: HashAlgorithm ): Bool {

            return verify( 
                signature: signature, 
                tag: tag, 
                signedData: signedData, 
[...]

PR #633, opened 2021-02-26
Introduce a new Account Key API

### LINKED ISSUE #371: Add native AccountKey type and update account key methods
## Context

Currently account public keys are encoded together with their signature algorithm, hashing algorithm and weight before being passed into Cadence as byte strings. This solution is a relic of the early days of Cadence, before we had support for more sophisticated composite types.

We'd like to move towards a more user-friendly solution, perhaps something like this:

```
transaction {
    let key = AccountKey(
      publicKey: pubKeyBytes, 
      sigAlgo: crypto.ECDSA_P256, 
      hashAlgo: crypto.SHA3_256, 
      weight: 1000,
    )
    signer.keys.add(key)
  }
}
```

## Definition of Done

- [ ] Add an `AccountKey` type: 

  ```
  struct AccountKey {
        let keyIndex: Int
        let publicKey: Crypto.PublicKey
        let hashAlgorithm: Crypto.HashAlgorithm
        let weight: UFix64
        let isRevoked: Bool
  }
  ```

- [ ] Add an `AuthAccount.Keys` type with the following functions:

  - [ ] Add a `AuthAccount.Keys.add` function:

    ```
    /// Adds a new key with the given weight
    fun add(
        _ publicKey: PublicKey,
        hashAlgorithm: HashAlgorithm,
        weight: UFix64
    ): AccountKey
    ```

  - [ ] Add a `AuthAccount.Keys.get` function:

    ```
    /// Returns the key at the given index, if it exists.
    /// Revoked keys are always returned, but they have `isRevoked` field set to true
    fun get(keyIndex: Int): AccountKey?
    ```

  - [ ] Add a `AuthAccount.Keys.revoke` function:

    ```
    /// Marks the key at the given index revoked, but does not delete it
    fun revoke(keyIndex: Int)
    ```

- [ ] Add an `AuthAccount.keys` field with type `AuthAccount.Keys` 

- [ ] Add an `PublicAccount.Keys` type with the following functions:

  - [ ] Add a `PublicAccount.Keys.get` function:

    ```
    /// Returns the key at the given index, if it exists.
    /// Revoked keys are always returned, but they have `isRevoked` field set to true
    fun get(keyIndex: Int): AccountKey?
    ```

- [ ] Add an `PublicAccount.keys` field with type `PublicAccount.Keys` 

- [ ] Tests
- [ ] Documentation

SupunS:
Closes #371

## Description
This PR introduces a new API for account key management.

A sample code:
```rust
transaction {
		let pubKey = PublicKey(
			publicKey: "010203".decodeHex(),
			signAlgo: SignatureAlgorithm.ECDSA_P256
		)

		// Add a key
		var addedKey: AccountKey = signer.keys.add(
			publicKey: pubKey,
			hashAlgo: HashAlgorithm.SHA3_256,
			weight: 100.0
		)


		// Retrieve a key
		var key: AccountKey? = signer.keys.get(keyIndex: addedKey.keyIndex)


		// Revoke a key
		key = signer.keys.revoke(keyIndex:  addedKey.keyIndex)
	}
}
```

### Implementation Details:
- Uses `CompositeType` and `CompositeValue` to represent any native type/value.
  - This avoids having to define new types/values to each and every new native type.
  - Makes it easier to add new native types in future
- `PublicKey`, `AccountKey`, `AuthAccountKeys` and `PublicAccount.Keys` are defined as native structs.
- `HashAlgorithm` and `SignatureAlgorithm` are defined as native enums.

### Backward Compatibility:
- The `AuthAccount` now supports `authAccount.keys.add()` and `authAccount.keys.revoke()` methods to add and revoke keys.
    - But, to maintain the backward compatibility, It also still supports the old way of adding and revoking key: `authAccount.addPublicKey()` and `authAccount.removePublicKey()` methods.

### Breaking Changes:
- Crypto contract now uses the native types for `PublicKey`, `HashAlogirthm` and `SignatureAlgorithm`. This means:
  - Types that was there inside the Crypto contract with the same names are removed.
  - `Crypto.SHA2_256.hash(data)` is no longer supported.
      - `Crypto.hash(data, HashAlgorithm.SHA2_256)` can be used instead.
- Some of methods in the `Interface` API are updated to use the newly introduced types and values. 

**Remaining tasks:**
- Need to add documentation (WIP). Created a separate issue #629 for this.
______

For contributor use:


runtime/account_keys_test.go
```diff
@@ -0,0 +1,741 @@
...
+		var err error
+		args[i], err = json.Encode(arg)
+		if err != nil {
+			panic(fmt.Errorf("broken test: invalid argument: %w", err))
+		}
+	}
+	return args
+}
+
+type TestCase struct {
```
turbolent, line 722: This type declaration is available in the whole package, but is only used by / useful for account key tests. Maybe rename to indicate that.

runtime/account_keys_test.go
```diff
@@ -0,0 +1,741 @@
...
+		events: make([]cadence.Event, 0),
+		keys:   make([]AccountKey, 0),
+	}
+}
+
+type Storage struct {
+	events      []cadence.Event
+	keys        []AccountKey
+	returnedKey AccountKey
+}
```
turbolent, line 741: Same here, this storage is only used for the account key tests. Maybe rename to avoid confusion in other tests (e.g. this "storage" type doesn't allow storage reads/writes)

runtime/common/nativelocation.go
```diff
@@ -0,0 +1,108 @@
...
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+const NativeLocationPrefix = "N"
+
+// NativeLocation
+//
+type NativeLocation struct{}
```
turbolent, line 31: I don't think we have to introduce a new location. The other "native" / built-in types don't have a location, e.g. `AuthAccount`'s type ID is just `AuthAccount

runtime/convertValues.go
```diff
@@ -423,6 +448,22 @@ func importCompositeValue(
...
+	switch qualifiedIdentifier {
+	case sema.PublicKeyType.QualifiedString():
+		builtinType = sema.PublicKeyType
+	case sema.AccountKeyType.QualifiedString():
+		builtinType = sema.AccountKeyType
+	case sema.HashAlgorithmType.QualifiedString():
+		builtinType = sema.HashAlgorithmType
+	case sema.SignatureAlgorithmType.QualifiedString():
+		builtinType = sema.SignatureAlgorithmType
+	}
```
turbolent, line 461: Maybe just declare them in a map and perform the lookup there?

Also, I think this switch should only be used if the location is native? (or once removed, I guess is nil). Currently this code would result in type confusion: if a value is imported that happens to have the same name as a built-in type, but is actually in e.g. an address location (from a deployed contract), then the built-in type is incorrectly returned.

runtime/interface.go
```diff
@@ -43,6 +43,39 @@ type ResolvedLocation = sema.ResolvedLocation
...
+	PublicKey *PublicKey
+	HashAlgo  HashAlgorithm
+	Weight    int
+	IsRevoked bool
+}
+
+type PublicKey struct {
+	PublicKey []byte
+	SignAlgo  SignatureAlgorithm
+}
```
turbolent, line 77: Maybe move these declarations to a separate file

runtime/interpreter/value.go
```diff
@@ -5617,13 +5617,17 @@ func (v *CompositeValue) SetMember(_ *Interpreter, locationRange LocationRange,
 }
 
 func (v *CompositeValue) String() string {
-	fields := make([]struct {
+	return formatComposite(string(v.TypeID()), v.Fields)
+}
+
+func formatComposite(typeId string, fields *StringValueOrderedMap) string {
```
turbolent, line 5623: 👍

runtime/interpreter/value.go
```diff
@@ -6791,12 +6792,6 @@ func (v AuthAccountValue) GetMember(inter *Interpreter, _ LocationRange, name st
 	case "storageCapacity":
 		return v.storageCapacityGet()
 
-	case "addPublicKey":
-		return v.addPublicKeyFunction
-
-	case "removePublicKey":
-		return v.removePublicKeyFunction
-
```
turbolent, line 6799: We should try to keep those for backwards compatibility

runtime/interpreter/value.go
```diff
@@ -7147,3 +7149,130 @@ func (v LinkValue) String() string {
...
+func (v *BuiltinCompositeValue) DynamicType(_ *Interpreter) DynamicType {
+	return BuiltinCompositeDynamicType{v.structType}
+}
+
+func (v *BuiltinCompositeValue) StaticType() StaticType {
+	return v.staticType
+}
+
+func (v *BuiltinCompositeValue) Copy() Value {
+	return v
```
turbolent, line 7187: This should return a copy, not a pointer to the same value. For example, all the fields need to be copied. See `CompositeValue.Copy`.

runtime/interpreter/value.go
```diff
@@ -7147,3 +7149,130 @@ func (v LinkValue) String() string {
 		v.TargetPath.String(),
 	)
 }
+
+// BuiltinCompositeValue
+type BuiltinCompositeValue struct {
+	Fields     *StringValueOrderedMap
+	staticType StaticType
+	structType *sema.BuiltinCompositeType
```
turbolent, line 7157: Maybe rename `structType` to `semaType`? Or why "struct"?

runtime/interpreter/value.go
```diff
@@ -7147,3 +7149,130 @@ func (v LinkValue) String() string {
...
+	// value is never owned
+	return nil
+}
+
+func (*BuiltinCompositeValue) SetOwner(_ *common.Address) {
+	// NO-OP: value cannot be owned
+}
+
+func (*BuiltinCompositeValue) IsModified() bool {
+	return false
```
turbolent, line 7200: I think this needs to check the fields. See e.g. `CompositeValue.IsModified`

runtime/interpreter/value.go
```diff
@@ -7147,3 +7149,130 @@ func (v LinkValue) String() string {
...
+}
+
+func (*BuiltinCompositeValue) SetModified(_ bool) {
+	// NO-OP
+}
+
+func (v *BuiltinCompositeValue) Destroy(_ *Interpreter, _ LocationRange) trampoline.Trampoline {
+	return trampoline.Done{}
```
turbolent, line 7208: This assumes that the value itself is not a resource type, and i.e. none of the fields are, otherwise all fields' destructor must be called here.

Maybe put an assertion in `NewBuiltinCompositeValue` that checks that the given type is not a resource type.

runtime/interpreter/value.go
```diff
@@ -7147,3 +7149,130 @@ func (v LinkValue) String() string {
...
+	value, _ := v.Fields.Get(name)
+	return value
+}
+
+func (*BuiltinCompositeValue) SetMember(_ *Interpreter, _ LocationRange, _ string, _ Value) {
+	panic(errors.NewUnreachableError())
+}
+
+// NewAccountKeyValue constructs an AccountKey value.
+func NewAccountKeyValue(
```
turbolent, line 7225: Nice! Can you please open an issue for refactoring the other built-in values to this new built-in composite value, too?

runtime/sema/authaccount_type.go
```diff
@@ -71,28 +71,6 @@ var AuthAccountType = &NominalType{
...
-				Kind: common.DeclarationKindFunction,
-				Resolve: func(identifier string, _ ast.Range, _ func(error)) *Member {
-					return NewPublicFunctionMember(
-						t,
-						identifier,
-						authAccountTypeRemovePublicKeyFunctionType,
-						authAccountTypeRemovePublicKeyFunctionDocString,
-					)
-				},
-			},
```
turbolent, line 95: I think we should keep the existing functionality for backwards-compatibility

runtime/sema/type.go
```diff
@@ -6436,3 +6469,227 @@ func (t *CapabilityType) GetMembers() map[string]MemberResolver {
 		},
 	})
 }
+
+type BuiltinCompositeType struct {
```
turbolent, line 6473: I had basically at the same time also introduced the same as `NominalType`. Could you please switch over to using that one instead and remove this new one? Sorry, I should have coordinated that better

runtime/stdlib/builtin.go
```diff
@@ -138,3 +140,113 @@ var LogFunction = NewStandardLibraryFunction(
...
+			var result interpreter.Value = interpreter.NilValue{}
+
+			if rawValueArgument >= 0 && rawValueArgument < caseCount {
+				caseValue := caseValues[rawValueArgument]
+				result = interpreter.NewSomeValueOwningNonCopying(caseValue)
+			}
+
+			return trampoline.Done{Result: result}
+		},
+	)
```
turbolent, line 247: It would be great if we would avoid duplicating the code from `declareEnumConstructor` here. 
Can you please try to refactor this constructor function out and use it here in `getEnumValue` and in `declareEnumConstructor`?

runtime/stdlib/builtin.go
```diff
@@ -138,3 +140,113 @@ var LogFunction = NewStandardLibraryFunction(
...
+				},
+			),
+		},
+		Members: sema.GetMembersAsMap(members),
+	}
+
+	return constructorType
+}
+
+func getEnumValue(enumType *sema.BuiltinCompositeType, enumCases []sema.BuiltinEnumCase) (value interpreter.Value) {
```
turbolent, line 223: Maybe make it clearer that this function only applies to built-in composite types with built-in enum cases

runtime/stdlib/crypto.go
```diff
@@ -246,3 +238,43 @@ func NewCryptoContract(
...
+		panic("cannot find hash algorithm raw value")
+	}
+
+	hashAlgoRawValue, ok := rawValue.(interpreter.IntValue)
+	if !ok {
+		panic("hash algorithm raw value needs to be subtype of integer")
+	}
+
+	hashAlgo := HashAlgorithm(hashAlgoRawValue.ToInt())
+	return hashAlgo
```
turbolent, line 259:
```suggestion
	return HashAlgorithm(hashAlgoRawValue.ToInt())
```

runtime/stdlib/flow.go
```diff
@@ -347,6 +347,12 @@ var AccountKeyAddedEventType = newFlowEventType(
 	AccountEventPublicKeyParameter,
 )
 
+var AccountKeyGetEventType = newFlowEventType(
+	"AccountKeyRetrieved",
+	AccountEventAddressParameter,
+	AccountEventPublicKeyParameter,
+)
```
turbolent, line 354: I'm not sure it's a good idea to emit an even each time a key is read, maybe remove this

runtime/stdlib/value.go
```diff
@@ -74,8 +74,8 @@ type StandardLibraryValues []StandardLibraryValue
 
 func (values StandardLibraryValues) ToSemaValueDeclarations() []sema.ValueDeclaration {
 	valueDeclarations := make([]sema.ValueDeclaration, len(values))
-	for i, function := range values {
-		valueDeclarations[i] = function
+	for i, value := range values {
```
turbolent, line 77: 👍

turbolent (commented):
Great job implementing this! 👏 Left a few comments.

In general I think we should try to keep backwards-compatibility for the account key API, as it is heavily used right now.

Keeping backwards-compatbility for the Crypto contract is not really possible (we don't ave type aliases), so the backward-incompatible changes are OK IMHO.

turbolent:
@tarakby Could you please also have a look at these chhanges, as they affect the Crypto contract?

runtime/common/nativelocation.go
SupunS: Yeah, I agree. Had to introduce this to make the exporting type/value work. Currently, the built-in types/values also get exported like any other struct, so that we can give the same user-experience when exporting/importing (also would be consistent in the JSON-Cadence data format as well). 

`cadence.StructType` relies on location. But we can go with a nil location, if that's ok. Thought it would be a bit dangerous for the downstream code to return nil..

runtime/sema/type.go
SupunS: That's a really good point. I also had the same thought at one point. But the catch was, there are a couple of places that we switch on the type `BuiltinCompositeType` to go in different paths. So it's sort of useful to have a separate type for built-in composites. (However, this `BuiltinCompositeType` might well be temporary, because of the below reason.)

Also, I realized that, as we migrate the rest of the types (e.g: auth account, public account, etc), and maybe for making future proof (leaving space to add builtin interfaces, contracts, etc. in the future), we could actually use the existing `CompositeType` and `CompositeValue` to represent BuiltinComposite types as well, eventually. So the checker and runtime only know one composite type/value - it doesn't matter where it comes from. 

Then we could also remove some of the fields (like `Members`, `nestedTypes`, etc) from the `NominalType` and use it purely for truly-nominal types (like `Int`, `String`, etc.).  
WDYT?

runtime/stdlib/builtin.go
SupunS: It causes a circular dependency, that's why had to duplicate it. 
But I'll see if I can move to a common-place and get rid of the recursion.

runtime/convertValues.go
SupunS: I think both issues will go away if we do the unification I mentioned here: https://github.com/onflow/cadence/pull/633#discussion_r586100296

runtime/interpreter/value.go
SupunS: Created https://github.com/onflow/cadence/issues/650

runtime/common/nativelocation.go
SupunS: just realized location is also needed for composite value encoding/decoding to/from CBOR. 
(I have missed adding that part here - one more reason to unify the two composite types :smile: ).

Perhaps would be good to have chat on what we can do here?

turbolent:
As discussed this morning, it would be great to refactor the code to

- [ ] Use `CompositeType` instead of introducing `BuiltinType`
- [ ] Use `CompositeValue` instead of introducing `BuiltinCompositeValue`

turbolent:
Really nice to see the great progress! 👏

orodio:
I was telling @psiemens about a week ago how much i wanted something like this. <3 <3 Great work @SupunS

runtime/common/nativelocation.go
SupunS: Removed in https://github.com/onflow/cadence/pull/633/commits/a5abb9669d4770a3a99e3d637581abb6567eae9f.

SupunS:
@turbolent Refactored to use `CompositeType` and `CompositeValue` to represent the native types and values.

runtime/sema/crypto_algorithm_types.go
```diff
@@ -0,0 +1,190 @@
...
+	panic(errors.NewUnreachableError())
+}
+
+func (algo SignatureAlgorithm) RawValue() int {
+	return int(algo)
+}
+
+func (algo SignatureAlgorithm) DocString() string {
+	switch algo {
+	case SignatureAlgorithmECDSA_P256:
```
tarakby, line 74: Just a general comment about the signature algorithm naming in the PR. We could use a more consistent rule `<algoName>_<curveName>` or `<algoName><curveName>`. 
In our case `ECDSA` and `BLS` are the algo names, while `P256`, `Secp256k1` and `BLS12381` are the curve names.

tarakby: So we could choose between: 
 - `ECDSA_P256` , `BLS_BLS123281` ..
 - `ECDSAP256` , `BLSBLS123281` ..

runtime/sema/crypto_algorithm_types.go
```diff
@@ -0,0 +1,190 @@
...
+	panic(errors.NewUnreachableError())
+}
+
+var HashAlgorithmType = newNativeEnumType(HashAlgorithmTypeName, &UInt8Type{})
+
+type HashAlgorithm int
+
+const (
+	// Supported hashing algorithms
+	HashAlgorithmSHA2_256 HashAlgorithm = iota
```
tarakby, line 91: I don't think there is a need to support the `384` versions in Cadence, unless there is another reason I'm not aware of. 
Should we remove them? cc @turbolent @psiemens

tarakby:
Nice work @SupunS 👏 
I've looked at some parts of the PR. Just a heads-up that the crypto contract might have non backward-compatible changes again soon. New features will need to be supported (my [proposal here](https://github.com/onflow/cadence/issues/582)), and I wanted to make sure the current PR won't make the future changes unfeasible.

runtime/sema/crypto_algorithm_types.go
SupunS: I only add them because the FVM would probably support those algorithms (https://github.com/onflow/flow-go/blob/master/crypto/hash/types.go#L8:L16).

If are not planning to support them, I am all-in to remove them.

runtime/sema/crypto_algorithm_types.go
SupunS: Good catch! I like the suggestion for `<algoName>_<curveName>`, looks more readable. Also consistent with your new crypto proposal.

Updated in https://github.com/onflow/cadence/pull/633/commits/a6f332cb95a402df32b6cadc46490734149ff194

runtime/account_keys_test.go
```diff
@@ -0,0 +1,937 @@
...
+func TestImportInvalidType(t *testing.T) {
+	t.Parallel()
+
+	runtime := NewInterpreterRuntime()
+
+	// encoded with an invalid type: 'N.PublicKey'
+	encodedArgs := []byte(`{
+		"type":"Struct",
+		"value":{
+			"id":"N.PublicKey",
```
turbolent, line 241: Is it invalid because of the unknown location prefix?

runtime/interface.go
```diff
@@ -69,10 +52,16 @@ type Interface interface {
...
+	// RemoveEncodedAccountKey removes a key from an account by index, add returns the encoded key.
+	RemoveEncodedAccountKey(address Address, index int) (publicKey []byte, err error)
 	// AddAccountKey appends a key to an account.
-	AddAccountKey(address Address, publicKey []byte) error
+	AddAccountKey(address Address, publicKey *PublicKey, hashAlgo HashAlgorithm, weight int) (*AccountKey, error)
+	// GetAccountKey retrieves a key from an account by index.
+	GetAccountKey(address Address, index int) (*AccountKey, error)
 	// RemoveAccountKey removes a key from an account by index.
-	RemoveAccountKey(address Address, index int) (publicKey []byte, err error)
+	RemoveAccountKey(address Address, index int) (*AccountKey, error)
```
turbolent, line 64: Given that the key is not removed, but revoked, maybe rename this to `RevokeAccountKey` (The Cadence function is also named `revoke`)

runtime/interface.go
```diff
@@ -69,10 +52,16 @@ type Interface interface {
 	SetValue(owner, key, value []byte) (err error)
 	// CreateAccount creates a new account.
 	CreateAccount(payer Address) (address Address, err error)
+	// AddEncodedAccountKey appends an encoded key to an account.
+	AddEncodedAccountKey(address Address, publicKey []byte) error
+	// RemoveEncodedAccountKey removes a key from an account by index, add returns the encoded key.
+	RemoveEncodedAccountKey(address Address, index int) (publicKey []byte, err error)
```
turbolent, line 58: Given that the key is not removed, but revoked, maybe rename this to `RevokeEncodedAccountKey` (The Cadence function is also named `revoke`). It should have been named "revoke" in the first place.

runtime/interpreter/interpreter.go
```diff
@@ -2718,14 +2717,21 @@ func (interpreter *Interpreter) declareEnumConstructor(
 		constructorMembers.Set(enumCase.Identifier.Identifier, caseValue)
 	}
 
+	value = EnumConstructorFunction(caseValues, constructorMembers)
+	variable.Value = value
+
+	return lexicalScope, value
+}
+
+func EnumConstructorFunction(caseValues []*CompositeValue, members *StringValueOrderedMap) HostFunctionValue {
```
turbolent, line 2726: 👍

runtime/interpreter/interpreter.go
```diff
@@ -4428,6 +4430,17 @@ func (interpreter *Interpreter) getElaboration(location common.Location) *sema.E
 }
 
 func (interpreter *Interpreter) getCompositeType(location common.Location, qualifiedIdentifier string) *sema.CompositeType {
+	if location == nil {
+		ty := sema.NativeCompositeTypes[qualifiedIdentifier]
```
turbolent, line 4434: 👍

runtime/interpreter/visitor.go
```diff
@@ -110,6 +110,8 @@ type EmptyVisitor struct {
 	DeployedContractValueVisitor     func(interpreter *Interpreter, value DeployedContractValue)
 }
 
+var _ Visitor = &EmptyVisitor{}
```
turbolent, line 113: 👍

runtime/stdlib/builtin.go
```diff
@@ -138,3 +140,97 @@ var LogFunction = NewStandardLibraryFunction(
...
+				TypeAnnotation: sema.NewTypeAnnotation(&sema.VariableSizedType{Type: &sema.UInt8Type{}}),
+			},
+			{
+				Label:          sema.PublicKeySignAlgoField,
+				Identifier:     sema.PublicKeySignAlgoField,
+				TypeAnnotation: sema.NewTypeAnnotation(sema.SignatureAlgorithmType),
+			},
+		},
+		ReturnTypeAnnotation:  sema.NewTypeAnnotation(sema.PublicKeyType),
+		RequiredArgumentCount: sema.RequiredArgumentCount(2),
```
turbolent, line 160: This function shouldn't have any optional (variable) arguments, right? If so, this line should be removed

runtime/stdlib/builtin.go
```diff
@@ -138,3 +140,97 @@ var LogFunction = NewStandardLibraryFunction(
...
+// BuiltinValues
+
+var BuiltinValues = StandardLibraryValues{
+	SignatureAlgorithmValue,
+	HashAlgorithmValue,
+}
+
+var SignatureAlgorithmValue = StandardLibraryValue{
+	Name:  sema.SignatureAlgorithmTypeName,
+	Type:  nativeEnumType(sema.SignatureAlgorithmType, sema.SignatureAlgorithms),
```
turbolent, line 181: Maybe refactor the common subexpression `nativeEnumType(sema.SignatureAlgorithmType, sema.SignatureAlgorithms)` into a variable declaration

runtime/stdlib/builtin.go
```diff
@@ -138,3 +140,97 @@ var LogFunction = NewStandardLibraryFunction(
...
+				},
+			),
+		},
+		Members: sema.GetMembersAsMap(members),
+	}
+
+	return constructorType
+}
+
+func nativeEnumValue(enumType *sema.CompositeType, enumCases []sema.NativeEnumCase) (value interpreter.Value) {
```
turbolent, line 223: 👍

runtime/runtime.go
```diff
@@ -2145,3 +2184,207 @@ func NewBlockValue(block Block) interpreter.BlockValue {
...
 }
+
+func (r *interpreterRuntime) newAccountKeysAddFunction(
+	addressValue interpreter.AddressValue,
+	runtimeInterface Interface,
+) interpreter.HostFunctionValue {
+	return interpreter.NewHostFunctionValue(
+		func(invocation interpreter.Invocation) trampoline.Trampoline {
+			publicKeyValue := invocation.Arguments[0].(*interpreter.CompositeValue)
+			if publicKeyValue.QualifiedIdentifier != sema.PublicKeyTypeName {
```
turbolent, line 2195: I think we can remove this check here and for the other arguments below, because the static check will already ensure that the value has the correct type.

runtime/runtime.go
```diff
@@ -2145,3 +2184,207 @@ func NewBlockValue(block Block) interpreter.BlockValue {
...
+					sema.UFix64Type{},
+				))
+			}
+
+			var err error
+			var accountKey *AccountKey
+			wrapPanic(func() {
+				publicKey := NewPublicKeyFromValue(publicKeyValue)
+				accountKey, err = runtimeInterface.AddAccountKey(addressValue.ToAddress(), publicKey, hashAlgo, weight.ToInt())
+			})
```
turbolent, line 2217: Only perform the runtime interface call in `wrapPanic`, so that only potential panics by the runtime interface call are treated as an external error

```suggestion
			publicKey := NewPublicKeyFromValue(publicKeyValue)
			address := addressValue.ToAddress()
			rawWeight := weight.ToInt()

			var err error
			var accountKey *AccountKey
			wrapPanic(func() {
				accountKey, err = runtimeInterface.AddAccountKey(address, publicKey, hashAlgo, rawWeight)
			})
```

runtime/runtime.go
```diff
@@ -2145,3 +2184,207 @@ func NewBlockValue(block Block) interpreter.BlockValue {
...
+			index, ok := invocation.Arguments[0].(interpreter.IntValue)
+			if !ok {
+				panic("get method requires the first argument to be an integer")
+			}
+
+			var err error
+			var accountKey *AccountKey
+			wrapPanic(func() {
+				accountKey, err = runtimeInterface.GetAccountKey(addressValue.ToAddress(), index.ToInt())
+			})
```
turbolent, line 2252:
```suggestion
			address := addressValue.ToAddress()
			keyIndex := index.ToInt()

			var err error
			var accountKey *AccountKey
			wrapPanic(func() {
				accountKey, err = runtimeInterface.GetAccountKey(address, keyIndex)
			})
```

runtime/runtime.go
```diff
@@ -2145,3 +2184,207 @@ func NewBlockValue(block Block) interpreter.BlockValue {
...
+			var accountKey *AccountKey
+			wrapPanic(func() {
+				accountKey, err = runtimeInterface.RemoveAccountKey(addressValue.ToAddress(), index.ToInt())
+			})
+			if err != nil {
+				panic(err)
+			}
+
+			r.emitAccountEvent(
+				stdlib.AccountKeyAddedEventType,
```
turbolent, line 2292:
```suggestion
				stdlib.AccountKeyRevokedEventType,
```

runtime/runtime.go
```diff
@@ -2145,3 +2184,207 @@ func NewBlockValue(block Block) interpreter.BlockValue {
...
+			index, ok := invocation.Arguments[0].(interpreter.IntValue)
+			if !ok {
+				panic("revoke method requires the first argument to be an integer")
+			}
+
+			var err error
+			var accountKey *AccountKey
+			wrapPanic(func() {
+				accountKey, err = runtimeInterface.RemoveAccountKey(addressValue.ToAddress(), index.ToInt())
+			})
```
turbolent, line 2286:
```suggestion
			address := addressValue.ToAddress()
			keyIndex := index.ToInt()

			var err error
			var accountKey *AccountKey
			wrapPanic(func() {
				accountKey, err = runtimeInterface.RemoveAccountKey(address, keyIndex)
			})
```

runtime/runtime.go
```diff
@@ -2145,3 +2184,207 @@ func NewBlockValue(block Block) interpreter.BlockValue {
...
+		NewPublicKeyValue(accountKey.PublicKey),
+		interpreter.NewNativeEnumCaseValue(sema.HashAlgorithmType, int(accountKey.HashAlgo)),
+		interpreter.NewUFix64ValueWithInteger(uint64(accountKey.Weight)),
+		interpreter.BoolValue(accountKey.IsRevoked),
+	)
+}
+
+func NewHashAlgorithmFromValue(value interpreter.Value) HashAlgorithm {
+	hashAlgoValue, ok := value.(*interpreter.CompositeValue)
+	if !ok || hashAlgoValue.QualifiedIdentifier != sema.HashAlgorithmTypeName {
```
turbolent, line 2374: Maybe also assert location is nil

runtime/runtime_test.go
```diff
@@ -2710,125 +2725,6 @@ func TestRuntimeTransaction_CreateAccount(t *testing.T) {
 	assert.EqualValues(t, stdlib.AccountCreatedEventType.ID(), events[0].Type().ID())
 }
 
-func TestRuntimeTransaction_AddPublicKey(t *testing.T) {
```
turbolent, line 2713: Thank you for moving these tests to a separate file 👍

runtime/sema/authaccount_type.go
```diff
@@ -192,11 +192,23 @@ var AuthAccountType = &SimpleType{
...
+						AuthAccountKeysType,
+						accountTypeKeysFieldDocString,
+					)
+				},
+			},
 		}
 	},
 	NestedTypes: func() *StringTypeOrderedMap {
 		nestedTypes := NewStringTypeOrderedMap()
 		nestedTypes.Set("Contracts", AuthAccountContractsType)
```
turbolent, line 210: Good idea to define a constant for the type name below. 
Could you please also add a constant for this name?

types.go
```diff
@@ -516,6 +516,10 @@ type StructType struct {
 func (*StructType) isType() {}
 
 func (t *StructType) ID() string {
+	if t.Location == nil {
```
turbolent, line 519: Maybe also add this for the other kinds, i.e. resources and contracts

turbolent (commented):
Great work! Just a couple things left, then it should be good to go 👏

runtime/sema/type.go
```diff
@@ -6480,3 +6514,138 @@ func (t *CapabilityType) GetMembers() map[string]MemberResolver {
...
+	}
+
+	accountKeyType.Members = GetMembersAsMap(members)
+	accountKeyType.Fields = getFieldNames(members)
+	return accountKeyType
+}()
+
+const PublicKeyTypeName = "PublicKey"
+const PublicKeyPublicKeyField = "publicKey"
+const PublicKeySignAlgoField = "signAlgo"
```
turbolent, line 6596: Let's keep the name as `signatureAlgorithm`

runtime/sema/type.go
```diff
@@ -6480,3 +6514,138 @@ func (t *CapabilityType) GetMembers() map[string]MemberResolver {
...
+
+	for _, semaType := range types {
+		NativeCompositeTypes[semaType.QualifiedIdentifier()] = semaType
+	}
+}
+
+const AccountKeyTypeName = "AccountKey"
+const AccountKeyKeyIndexField = "keyIndex"
+const AccountKeyPublicKeyField = "publicKey"
+const AccountKeyHashAlgoField = "hashAlgo"
```
turbolent, line 6538: Let's keep the name as `hashAlgorithm`

runtime/account_keys_test.go
SupunS: yeah. 
Maybe this is not the best place for the test case though.

runtime/stdlib/builtin.go
SupunS: Yeah, it shouldn't. I thought its a dynamic validation for any parameter to avoid inconsistencies. Will remove it!

runtime/stdlib/builtin.go
SupunS: `nativeEnumType(sema.SignatureAlgorithmType, sema.SignatureAlgorithms)` is only used here. In the other place, arguments are different.

runtime/runtime.go
SupunS: Hmm.. added this check because I figured we could actually send corrupted data from outside (by importing as script/function arguments). Because we are not doing a dynamic type checking for imports. 

But yeah, this is not the place to handle that. I will create a separate issue for that and tackle it separately.

runtime/sema/authaccount_type.go
SupunS: Migrated these in a separate PR: https://github.com/onflow/cadence/pull/668/files#diff-b5688033a56b955357a3fc86e260200fe8e341d094cfed5f9bf870cc0ee2f87aR25-R40

runtime/runtime.go
SupunS: Similar to https://github.com/onflow/cadence/pull/633#discussion_r592774043, these should have been validated at the time of importing, and all the other cases should have been statically validated. So removed these validations from here as well.

runtime/account_keys_test.go
SupunS: Moved under json encoding/decoding test: f436f41

runtime/runtime.go
SupunS: Created https://github.com/onflow/cadence/issues/689

runtime/stdlib/builtin.go
turbolent: Ah, I missed that! Makes sense 👍

turbolent (approved):
Awesome work @SupunS!

PR #634, opened 2021-02-27
Fix code examples and improve documentation in language reference

jeroenlm:
## Description

- Fix code example in composite-types part
- Remove false statement in composite-types part
- Remove typo in the functions part
- Fix code example in restricted-types part

PS: this is my very first PR. Feel free to give any feedback!
______

For contributor use:

### DIFF
--- a/docs/language/composite-types.mdx
+++ b/docs/language/composite-types.mdx
@@ -79,7 +79,7 @@ of the composite type requires them.
 
 Composite types can only be declared within [contracts](../contracts)
 and not locally in functions.
-They can also not be nested.
+
 
 Resource must be created (instantiated) by using the `create` keyword
 and calling the type like a function.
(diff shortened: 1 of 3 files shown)


docs/language/composite-types.mdx
```diff
@@ -79,7 +79,7 @@ of the composite type requires them.
 
 Composite types can only be declared within [contracts](../contracts)
 and not locally in functions.
-They can also not be nested.
```
turbolent, line 82: Was this line removed on purpose? If it doesn't make sense, I think we need to extend it / improve the wording

turbolent (commented):
@jeroenlm Thank you for the fixes! 

Looks good, I just have one question

docs/language/composite-types.mdx
jeroenlm: I removed this line on purpose indeed. When you scroll up a little bit on that page (composite-types), there is written: "Nesting of resources is only allowed within other resource types, or in data structures like arrays and dictionaries, but not in structures, as that would allow resources to be copied." Since resources are composite types, this means that composite types can be nested in some cases. That's why I removed that line. Does it make sense? Or maybe am I misunderstanding the writer of that sentence?

turbolent: Yeah, I guess we can remove that sentence 👍

jeroenlm:
Great! Very happy with my first contribution :)

PR #638, opened 2021-03-01
Optimize deferred dictionary keys

turbolent:
## Description

Only construct storage keys for deferred dictionary keys when needed, instead of creating all of them when decoding.

______

For contributor use:


runtime/interpreter/decode.go
```diff
@@ -411,8 +411,9 @@ func (d *Decoder) decodeDictionary(v interface{}, path []string) (*DictionaryVal
 
 	entries := NewStringValueOrderedMap()
 
-	var deferred *orderedmap.StringStringOrderedMap
+	var deferred *orderedmap.StringStructOrderedMap
```
SupunS, line 414: curious to know, whether using a string-interface map with nil values be more efficient, because we are not writing the values anyway?

SupunS (approved):

runtime/interpreter/decode.go
turbolent: Great concern, @dete had asked the same.

In Go, an `interface{}` is 8/16 bytes, whereas an `struct{}` is literally zero size, see e.g. https://stackoverflow.com/questions/22770114/any-difference-in-using-an-empty-interface-or-an-empty-struct-as-a-maps-value

PR #646, opened 2021-03-03
Make contract update validation optional and disable it by default

turbolent:

### DIFF
--- a/runtime/contract_update_validation_test.go
+++ b/runtime/contract_update_validation_test.go
@@ -36,7 +36,9 @@ func TestContractUpdateValidation(t *testing.T) {
 
 	t.Parallel()
 
-	runtime := NewInterpreterRuntime()
+	runtime := NewInterpreterRuntime(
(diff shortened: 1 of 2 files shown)


janezpodhostnik (approved):
LGTM. 

How come the default is that it is disabled? Isn't that dangerous?

turbolent:
@janezpodhostnik Ah, sorry, I should have added more context: This is port of #647 to master.

The contract update validation is a new feature of v0.13 and was tested on Testnet, but there were two problems:
- Some nodes were running an older version (intentionally), and even though we assumed there are no changes that would lead to a different execution result, the contract validation may actually reject a contract update now and reject such a transaction, whereas the nodes with the older version would accept and succeed
- We discovered #648 

So for now we disable the new feature and keep the current behaviour of the networks where there is no contract update validation, and plan to re-enable it once #648 is fixed, in a proper spork (we can't have it just enabled on some of the nodes)

PR #657, opened 2021-03-04
Adds contentType tags. 

10thfloor:
### Tagging all markdown with the content type.

Possible content types: 
`HOWTO` (Guide)
`TUT` (Tutorial)
`INTRO` (index /intro page)
`EXP` (Explainer)
`REF` (Reference)
`FAQ` (Answers)

* all `README` are considered `INTRO` and all `CHANGELOG` are considered `REF`

As outlined here: https://www.notion.so/dapperlabs/Types-of-Docs-033e722018f34036811ff38dac03eded

PR also formats whitespace.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -1,6 +1,6 @@
 ---
 title: FAQ
-type: EXP
+type: FAQ
 ---
 
 # FAQ
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -1,3 +1,8 @@
+---
+title: FAQ
+type: EXP
+---
+
 # FAQ
 
 ## Is there a formal grammar (e.g. in BNF) for Cadence?
--- a/docs/anti-patterns.mdx
+++ b/docs/anti-patterns.mdx
@@ -1,5 +1,6 @@
 ---
 title: Cadence Anti-Patterns
+type: EXP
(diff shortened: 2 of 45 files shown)


docs/FAQ.md
```diff
@@ -1,3 +1,8 @@
+---
```
turbolent, line 1: Oh, it just occurred to me: Are now all files in `docs` pulled into the docs website? 

Previously the content here was developer documentation, i.e. docs for contributing to Cadence, not for using Cadence. Maybe we should move the content somewhere else?

SupunS: I think that would be a good idea. At some point in the future, we would have to tag the documentation with a (major/minor) version of Cadence. When we do that. having separate repos, leaves us the freedom to do individual release cycles for docs and Cadence. i.e: We can keep improving/fix (patching) documentation without doing a release on the Cadence.

10thfloor: ok, I'm open to suggestions. A manual deploy of the docs site is still required fro changes from these files to show up. 
Maybe we could create a guides folder to hold the other docs that don;t need to be at the root of this folder. 

Also, I believe some of these docs may be obsolete?

turbolent:
@10thfloor what's the status here?

Issue #650, opened 2021-03-03
Refactor existing native composite values to use CompositeValue


SupunS:
## Context

Currently native composite values (e.g. `AuthAccount`, `AuthAccount.Contracts`, `PublicAccount`, etc.) are defined through separate Go types implementing the `Value` interface (e.g. `AuthAccountValue`).

Refactor those existing value type to use `CompositeValue` instead.

Pros:
  - No need to declare new Value type, reduce boilerplate
  - No special conversion functions in internal encoding/decoding and import/export

## Definition of Done

Non-exhaustive list:
- [ ] `AuthAccountValue`
- [ ] `PublicAccountValue`
- [ ] `AuthAccountContractsValue`

Not good candidates:
- `PathValue`
- `LinkValue`
- `StringValue`

PR #664, opened 2021-03-05
Add computed fields to the composite value

### LINKED ISSUE #660: Add support for computed fields in `CompositeValue`
### Issue To Be Solved

`CompositeValue` currently only supports stored fields. 

Some built-in ("native") composite values have computed fields.

Add support for computed fields to `CompositeValue`, so it can be used to represent built-in values.

### Suggested Solution

- [ ] Add a field `computedFields` to `CompositeValue`, which has an ordered map of `string` to `func() Value` entries
- [ ] On member access, also look up the requested member name in `computedFields`
- [ ] Tests

SupunS:
Closes #660 

Working towards #650

## Description

A computed field is a `func(*Interpreter) Value` function pointer. Invoking this function computes the value on the fly and returns it. 

A computed value takes an `*Interpreter` as an argument, as computation of the value may rely on the interpreter.
e.g: [auth account fields](https://github.com/onflow/cadence/blob/master/runtime/interpreter/value.go#L6800:L6820)
______

For contributor use:


runtime/interpreter/value.go
```diff
@@ -5341,6 +5342,8 @@ type CompositeValue struct {
 	modified            bool
 }
 
+type ComputedField func(*Interpreter) Value
```
turbolent, line 5345: Good idea to introduce a type for this 👍

turbolent (approved):
Nice!

PR #669, opened 2021-03-09
Add support for exporting enums

### LINKED ISSUE #632: Add Enum cadence.Value
### Issue To Be Solved

Currently, enums are not supported as `cadence.Value`. Need to add support for this so the values can be parsed by the SDK.

### Suggest A Solution

internally resources, structs, contracts, and enums are all represented using one Go type, `CompositeValue`, but the external representation, i.e Value in the cadence package has them all separate

* need to copy and paste and create `cadence.EnumType`
* copy paste eg. `cadence.Struct` to a  `cadence.Enum` , and then in `exportCompositeValue` add a case

SupunS:
Closes #632

## Description

Currently exporting an enum is not supported. Runtime crashes with a panic.

______

For contributor use:

### DIFF
--- a/runtime/convertTypes.go
+++ b/runtime/convertTypes.go
@@ -236,6 +236,14 @@ func exportCompositeType(t *sema.CompositeType, results map[sema.TypeID]cadence.
 			Fields:              fields,
 		}
 
+	case common.CompositeKindEnum:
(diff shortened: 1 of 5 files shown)


turbolent (approved):
Nice, thank you for adding this! 👍

PR #674, opened 2021-03-10
Optimize checker activations

turbolent:
## Description

Reduce allocation in the checker by reusing activations through an object pool.

```
before: BenchmarkCheckContractInterfaceFungibleTokenConformance-12    	    8824	    115880 ns/op	   80832 B/op	    1381 allocs/op
after:  BenchmarkCheckContractInterfaceFungibleTokenConformance-12    	   10022	    110549 ns/op	   68713 B/op	    1193 allocs/op
```

______

For contributor use:

### DIFF
--- a/runtime/common/orderedmap/orderedmap.go
+++ b/runtime/common/orderedmap/orderedmap.go
@@ -48,7 +48,8 @@ func NewKeyTypeValueTypeOrderedMap() *KeyTypeValueTypeOrderedMap {
 // Clear removes all entries from this ordered map.
 func (om *KeyTypeValueTypeOrderedMap) Clear() {
 	om.list.Init()
-	for key := range om.pairs {
+	// NOTE: Range over map is safe, as it is only used to delete entries
+	for key := range om.pairs { //nolint:maprangecheck
 		delete(om.pairs, key)
 	}
 }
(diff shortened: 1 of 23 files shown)


runtime/common/orderedmap/orderedmap.go
```diff
@@ -45,6 +45,15 @@ func NewKeyTypeValueTypeOrderedMap() *KeyTypeValueTypeOrderedMap {
 	}
 }
 
+// Clear removes all entries from this ordered map.
+func (om *KeyTypeValueTypeOrderedMap) Clear() {
+	om.list.Init()
+	// NOTE: Range over map is safe, as it is only used to delete entries
+	for key := range om.pairs { //nolint:maprangecheck
+		delete(om.pairs, key)
+	}
```
SupunS, line 54: would it be faster if we re-initialized the map, and let the old one be GC'ed?
```suggestion
	om.pairs = make(map[KeyType]*KeyTypeValueTypePair)
```

turbolent: The idea here is to reuse the map and avoid allocating more memory and producing garbage.
I benchmarked this suggestion and it's both less memory efficient, but also slower:

```
clearing by deleting:	7682	    147602 ns/op	   68704 B/op	    1193 allocs/op
clearing by new map:	7294	    153780 ns/op	   78146 B/op	    1283 allocs/op
```

Good idea though, I was curious here how Go performs

SupunS: oh nice, interesting results!

SupunS (approved):

Issue #678, opened 2021-03-10
Improve interpreter performance


turbolent:
### Issue To Be Solved

The interpreter currently uses trampolining to evaluate programs.

This is useful for implementing features such as metering and e.g. debugging, but has a significant performance impact.

### Suggested Solution

Increase the performance by switching to direct evaluation.

PR #679, opened 2021-03-10
Optimize interpreter: Evaluate binary expressions directly

turbolent:
Work towards #678

## Description

Instead of using trampolines to evaluate binary expressions, evaluate sub-expressions directly.

______

For contributor use:

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -239,6 +239,7 @@ type Interpreter struct {
 	importLocationHandler          ImportLocationHandlerFunc
 	uuidHandler                    UUIDHandlerFunc
 	interpreted                    bool
+	statement                      *Statement
 }
 
 type Option func(*Interpreter) error
@@ -611,8 +612,6 @@ func (interpreter *Interpreter) runUntilNextStatement(t Trampoline) (interface{}
 // When there is a statement, it calls the onStatement callback, and then continues the execution.
 func (interpreter *Interpreter) runAllStatements(t Trampoline) interface{} {
 
-	var statement *Statement
-
 	// Wrap errors if needed
 
 	defer recoverErrors(func(internalErr error) {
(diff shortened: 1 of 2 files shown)


runtime/interpreter/interpreter.go
```diff
@@ -1614,181 +1616,144 @@ func (interpreter *Interpreter) VisitIdentifierExpression(expression *ast.Identi
...
-			left := tuple.left.(NumberValue)
-			right := tuple.right.(NumberValue)
-			return f(left, right)
-		})
+func (interpreter *Interpreter) visitBinaryOperation(
+	expr *ast.BinaryExpression,
+	f func(left, right Value) Value,
+) Trampoline {
+	left := interpreter.runAllStatements(expr.Left.Accept(interpreter).(Trampoline)).(Value)
+	right := interpreter.runAllStatements(expr.Right.Accept(interpreter).(Trampoline)).(Value)
```
SupunS, line 1624: Since we will be writing this line a lot going forward, maybe its worth extracting this logic to a method. 
```go
func (interpreter *Interpreter) evaluate(element ast.Element) Value {
    return interpreter.runAllStatements(element.Accept(interpreter).(Trampoline)).(Value)
}
```

Then these lines simplify to:
```go
left := interpreter.evaluate(expr.Left)
right := interpreter.evaluate(expr.Right)
```

runtime/interpreter/interpreter.go
```diff
@@ -1614,181 +1616,144 @@ func (interpreter *Interpreter) VisitIdentifierExpression(expression *ast.Identi
...
+	return Done{Result: f(left, right)}
 }
 
 func (interpreter *Interpreter) VisitBinaryExpression(expression *ast.BinaryExpression) ast.Repr {
 	switch expression.Operation {
 	case ast.OperationPlus:
-		return interpreter.visitNumberBinaryOperation(
+		return interpreter.visitBinaryOperation(
```
SupunS, line 1631: I am thinking maybe we could do another minor optimization (also reduces code) here, by avoiding the creation of function pointers and avoiding the delegation.
```go
func (interpreter *Interpreter) VisitBinaryExpression(expression *ast.BinaryExpression) ast.Repr {
	left := interpreter.evaluate(expression.Left)
	right := interpreter.evaluate(expression.Right)

	var result Value

	switch expression.Operation {
	case ast.OperationPlus:
		result = left.(NumberValue).Plus(right.(NumberValue))
	case ast.OperationMinus:
		result = left.(NumberValue).Minus(right.(NumberValue))
	case ast.OperationMod:
		result = left.(NumberValue).Mod(right.(NumberValue))
	}

	return Done{Result: result}
}

```

we could early exit the ones that are not migrated yet.

SupunS (approved):

runtime/interpreter/interpreter.go
turbolent: Yep 👍  I refactored that in the next PR: https://github.com/onflow/cadence/pull/681/files#diff-a36f730b18ca31316909bd74125003c1fbe785a13f926b4a0631e7c0cc6a51e4R1552

runtime/interpreter/interpreter.go
turbolent: Yep, in the final PR the code looks similar to this 👍

PR #695, opened 2021-03-15
Fix nested error pretty printing

turbolent:
## Description

Don't let nested error's location override outer errors' locations: 
When pretty printing errors of a program that has errors, but also imports another program that has errors, the location of the nested errors was accidentally used for the outer errors.

While debugging I also noticed that the "invalid restricted type" error (`InvalidRestrictedTypeError`) was generated for invalid types, but should not. (Invalid types are produced when a previous error occurred, and indicates that the error was caused due to the preceding error).

______

### DIFF
--- a/runtime/pretty/print.go
+++ b/runtime/pretty/print.go
@@ -161,16 +161,19 @@ func (p ErrorPrettyPrinter) PrettyPrintError(err error, location common.Location
 		}
 
 		if err, ok := err.(errors.ParentError); ok {
+
 			for _, childErr := range err.ChildErrors() {
+
+				childLocation := location
 
 				if childErr, ok := childErr.(common.HasImportLocation); ok {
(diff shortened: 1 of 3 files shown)


jordanschalm (approved):

runtime/pretty/print.go
```diff
@@ -161,16 +161,19 @@ func (p ErrorPrettyPrinter) PrettyPrintError(err error, location common.Location
...
+
 			for _, childErr := range err.ChildErrors() {
 
+				childLocation := location
+
 				if childErr, ok := childErr.(common.HasImportLocation); ok {
 					importLocation := childErr.ImportLocation()
 					if importLocation != nil {
-						location = importLocation
+						childLocation = importLocation
```
turbolent, line 172: This nested loop was accidentally overriding the outer variable

PR #710, opened 2021-03-18
Add support for access(account) and multiple contracts per account

### LINKED ISSUE #702: Contract access level does not work for multiple contracts per account
### Problem

The access modifier `access(account)` should allow access to code from the same account.
However, it currently does not support access between contracts in the same account.

### Steps to Reproduce

- Use access modifier in e.g. address location `0x1.A`
- Access code `0x1.A` from `0x1.B`
- Access is prohibited, but should be allowed

### Acceptance Criteria

- Access modifier works for multiple contracts per account

### Context

When support for multiple contracts per account was added, the access modifier check was not updated and still checks locations match by comparing IDs, but should ignore the name for address locations: https://github.com/onflow/cadence/blob/9272737ac0a0861a587d9db0fef2edb5a40fe92c/runtime/sema/check_member_expression.go#L285-L293

turbolent:
Closes #702

## Description

When checking `access(account)`, check address locations' addresses are equal, not their full IDs are equal.

______

For contributor use:

### DIFF
--- a/runtime/common/location.go
+++ b/runtime/common/location.go
@@ -40,11 +40,12 @@ type Location interface {
 // LocationsMatch returns true if both locations are nil or their IDs are the same.
 //
 func LocationsMatch(first, second Location) bool {
-	if first == nil && second == nil {
-		return true
+
+	if first == nil {
+		return second == nil
 	}
 
-	if (first == nil && second != nil) || (first != nil && second == nil) {
+	if second == nil {
 		return false
 	}
 
@@ -56,11 +57,12 @@ func LocationsMatch(first, second Location) bool {
 // or otherwise if their IDs are the same.
 //
 func LocationsInSameAccount(first, second Location) bool {
-	if first == nil && second == nil {
-		return true
+
+	if first == nil {
+		return second == nil
 	}
 
-	if (first == nil && second != nil) || (first != nil && second == nil) {
+	if second == nil {
 		return false
 	}
 
--- a/runtime/common/location.go
+++ b/runtime/common/location.go
@@ -37,6 +37,8 @@ type Location interface {
 	QualifiedIdentifier(typeID TypeID) string
 }
 
+// LocationsMatch returns true if both locations are nil or their IDs are the same.
+//
 func LocationsMatch(first, second Location) bool {
 	if first == nil && second == nil {
 		return true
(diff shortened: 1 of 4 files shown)


runtime/tests/checker/access_test.go
```diff
@@ -2286,3 +2286,119 @@ func TestCheckInvalidRestrictiveAccessModifier(t *testing.T) {
...
+
+	for _, variableKind := range ast.VariableKinds {
+
+		t.Run(variableKind.Name(), func(t *testing.T) {
+
+			importedChecker, err := ParseAndCheckWithOptions(t,
+				fmt.Sprintf(
+					`
+			          pub contract A {
+                          access(account) %s a: Int
```
SupunS, line 2355: seems there are some tabs along with the whitespaces at this line

runtime/tests/checker/access_test.go
```diff
@@ -2286,3 +2286,119 @@ func TestCheckInvalidRestrictiveAccessModifier(t *testing.T) {
...
+	}
+
+	location2B := common.AddressLocation{
+		// NOTE: different address from A
+		Address: common.BytesToAddress([]byte{0x2}),
+		Name:    "B",
+	}
+
+	const importingCode = `
+	  import A from 0x1
```
SupunS, line 2312: probably a similar case here

runtime/common/location.go
```diff
@@ -49,6 +51,33 @@ func LocationsMatch(first, second Location) bool {
...
+// or otherwise if their IDs are the same.
+//
+func LocationsInSameAccount(first, second Location) bool {
+	if first == nil && second == nil {
+		return true
+	}
+
+	if (first == nil && second != nil) || (first != nil && second == nil) {
+		return false
+	}
```
SupunS, line 65: i think we could simplify a bit:
```suggestion
	if first == nil {
		return second == nil
	}

	if second == nil {
		return first == nil   
		// or simply `return false`. But above is more verbose.
	}
```

SupunS (approved):

runtime/common/location.go
turbolent: Good idea, simplified in ac4ae8b

runtime/tests/checker/access_test.go
turbolent: Right, cleaned up in 8a1a6b0

runtime/tests/checker/access_test.go
turbolent: Cleaned up in 8a1a6b0

PR #713, opened 2021-03-19
Make native composite types non-storable

turbolent:
## Description

As discussed, make native composite types like `AccountKey` and `PublicKey` non-storable for now, so that we can make future changes without breaking stored values. In a future version of Cadence we could make some types storable and guarantee backwards-compatibility.

Unfortunately, this change also makes those types unusable as entry point arguments (script arguments, transaction arguments), as the runtime is currently checking that parameter types are storable (so they can be imported). Ideally we would introduce the concept of "importable types", just like we have the concept of exportable types, so that those types could stay non-storable, but could be passed in arguments.

While adjusting the account key tests, also improve them (e.g use script and transaction locations, improve naming, etc.)

______

For contributor use:

### DIFF
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -4366,6 +4366,11 @@ func (t *CompositeType) IsStorable(results map[*Member]bool) bool {
 		return false
 	}
 
+	// Native/built-in types are not storable for now
+	if t.Location == nil {
+		return false
+	}
+
 	// If this composite type has a member which is non-storable,
 	// then the composite type is not storable.
 
@@ -6630,6 +6635,7 @@ var PublicKeyType = func() *CompositeType {
 
 	accountKeyType.Members = GetMembersAsMap(members)
 	accountKeyType.Fields = getFieldNames(members)
+
 	return accountKeyType
 }()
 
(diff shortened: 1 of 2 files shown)


runtime/sema/type.go
```diff
@@ -4366,6 +4366,11 @@ func (t *CompositeType) IsStorable(results map[*Member]bool) bool {
 		return false
 	}
 
+	// Native/built-in types are not storable for now
+	if t.Location == nil {
```
SupunS, line 4370: since making it non-storable makes it non-importable too, should we also make it non-exportable (non externally returnable), just to be consistent?

SupunS (approved):

runtime/sema/type.go
turbolent: That's a good point! I'll try to make the change and see 👍

SupunS: Oh, I see. Yeah, that's a fair point! Maybe it is better to leave it this way then.
I guess we will be anyway allowing them to be passed as arguments, and this limitation is only temporary.

Thanks for checking on that!

turbolent: Right, the non-importability should ideally only be temporary. I've opened #717 to make the types importable again

PR #715, opened 2021-03-20
Lazily load contract values

turbolent:
Often contracts are only imported to refer to types, not to use them as values.

Instead of eagerly loading the contract value when a contract declaration is interpreted, lazily load it once it is used. 

Add support for interpreter variables to have an optional getter function, which is called once to initialize the value of the variable.

Also refactor nested values to nested variables, so they too can be lazily initialized.

______


runtime/interpreter/function.go
```diff
@@ -170,8 +170,10 @@ func (f HostFunctionValue) Invoke(invocation Invocation) Value {
 }
 
 func (f HostFunctionValue) GetMember(_ *Interpreter, _ func() LocationRange, name string) Value {
-	value, _ := f.Members.Get(name)
-	return value
+	if variable, ok := f.NestedVariables.Get(name); ok {
```
SupunS, line 173: should we check whether `f.NestedVarable` is nil, since [NewHostFunctionValue](https://github.com/onflow/cadence/pull/715/files#diff-bd8f574d953b36c96ac20b3a19aee38b0189d4711eb452622caabb6465bcfb24L122) constructor-alike function only acecpts the `HostFunction`.

runtime/interpreter/variable.go
```diff
@@ -19,11 +19,26 @@
...
+	}
+	return v.value
+}
+
+func (v *Variable) SetValue(value Value) {
+	v.getter = nil
+	v.value = value
+}
+
+func NewVariable(value Value, getter func() Value) *Variable {
```
SupunS, line 39: The signature implies it needs both the getter and the value (probably a personal taste :smile: ), but I guess it always needs only one of the two at a given time? Maybe having two constructors is better? (I miss overloading!)

runtime/interpreter/variable.go
```diff
@@ -19,11 +19,26 @@
...
 type Variable struct {
-	Value Value
+	value  Value
+	getter func() Value
 }
 
-func NewVariable(value Value) *Variable {
+func (v *Variable) GetValue() Value {
+	if v.getter != nil {
+		v.value = v.getter()
```
SupunS, line 28: I have two questions:

When would the value be loaded for a contract?

Will this be thread-safe? As in, will there be a situation where two or more programs trying to load a 'contract' at the same time (not sure if that's a valid use-case though)?

runtime/interpreter/variable.go
janezpodhostnik: The alternative that comes to mind is to have a `func (v *Variable) WithValueGetter(getter func() Value) *Variable` receiver  instead of the extra constructor parameter (that is often null).

janezpodhostnik (approved):

runtime/interpreter/function.go
turbolent: Good point! Fixed in 501f6563e6fa825e40a6296e24f8a4a50bd939a7

runtime/interpreter/variable.go
turbolent: Good idea! I'll add two different constructor functions, and yes, I do miss overloading, default arguments for parameters, etc. too :D

turbolent: Refactored in 1443910e43560dac8daf41d365df32b788e14f45

runtime/interpreter/variable.go
turbolent: The interpreter used to eagerly load the contract value right when the interpreter interpreted a contract declaration: it would load the value and declare a variable with the value.

Now, when the interpreter interprets a variable declaration, the interpreter only declares a variable with a getter function, and the first time the variable's value is accessed, the getter function is called, and the contract value is loaded, and placed in the variable.

If the contract variable is never accessed, the contract value is never loaded. This is very often the case, so we can safe the unnecessary load.

There is no need to be thread-safe in the interpreter, it is single-threaded and so there is no potential for other parts of interpretation trying to load it at the same time. If external code attempts to e.g. access `Globals` and a variable's value in a concurrent way, then that's undefined behaviour.

Issue #712, opened 2021-03-19
Allow import of Capability values (cadence.Capability)


SupunS:
### Issue To Be Solved

It should be possible to import Capability values (cadence.Capability), e.g. when sending them as an argument in a transaction or script.

### Suggested Solution

- [x] Add support for importing a **public** `cadence.Capability`
  - [x] Convert `cadence.Capability` in `importValue` to `interpreter.CapabilityValue`
  - [x] Check that the domain is public, fail if not
  - [x] Check that the type exists, fail if not
- [x] Update argument validation (if necessary)
  - [x] https://github.com/onflow/cadence/pull/724/files#diff-863660c79fae3d9b381c2dff3025cd88dc46eb62571b754e7695acbfeab4fbfeR1211
  - [x] https://github.com/onflow/cadence/pull/724/files#diff-863660c79fae3d9b381c2dff3025cd88dc46eb62571b754e7695acbfeab4fbfeR1226
- [x] Tests

#### Related issues:
- Allow import of type values #491

turbolent:
@SupunS If we want to allow that, we must ensure the capability is not private. I updated the description with that constraint and added a TODO item in the proposed solution to check it.

turbolent:
This is done 🎉

PR #724, opened 2021-03-23
Validate arguments passed into cadence programs

### LINKED ISSUE #689: Reject malformed values from being imported
### Problem
Consider the below program, which accepts a `Foo` struct that have a `String` typed `bar` field.
```cadence
pub fun main(f: Foo): Foo {
    return f
}

pub struct Foo {
    pub var bar: String

    init() {
        self.bar = "Hello"
    }
}
```

It is possible to send a malformed encoded value, as long as the type name is a valid one:
```json
{
   "type":"Struct",
   "value":{
      "id":"S.test.Foo",
      "fields":[
         {
            "name":"nonExistingField",
            "value":{
               "type":"Int",
               "value":"3"
            }
         }
      ]
   }
}
```

The program would accept this value but eventually crashes internally when tries to perform different operations.

### Acceptance Criteria
- [ ] Validate the value against the type before importing.
    - [ ] For simple types: Value should comply with the specified type
    - [ ] For composite types: The name and the value (recursively) of the fields should comply with the fields of the specified type
    - [ ] For other structural types (arrays, dictionaries, etc): The member values should comply with the specified member type
- [ ] Test cases

SupunS:
Closes #689

## Description

This PR adds validation for the arguments passed into a cadence program, which checks the value against the dynamic type of the value.

______

For contributor use:


runtime/convertValues.go
```diff
@@ -378,7 +379,7 @@ func importValue(value cadence.Value) interpreter.Value {
 		}
 	case cadence.Enum:
 		return importCompositeValue(
-			common.CompositeKindStructure,
```
turbolent, line 381: 👍

runtime/runtime.go
```diff
@@ -529,8 +529,19 @@ func validateArgumentParams(
 
 		arg := importValue(value)
 
+		// Check whether the decoded value conforms to the type associated with the value.
+		dynamicType := arg.DynamicType(inter)
+		if !valueConformsToDynamicType(arg, dynamicType) {
```
turbolent, line 534: 👍

runtime/argument_validation.go
```diff
@@ -0,0 +1,354 @@
...
+// Utility functions for validating the arguments passed into a script or a transaction.
+
+// valueConformsToDynamicType checks whether the shape of the value is same as the shape
+// defined by the type associated with the value. Returns `true` if the value and the
+// type matches, and `false` otherwise.
+//
+// e.g: A value with the type as 'Foo', but having additional fields than what is
+// specified in 'Foo', would return `false`.
+//
+func valueConformsToDynamicType(value interpreter.Value, dynamicType interpreter.DynamicType) (ok bool) {
```
turbolent, line 36: I think it would be good to move this to the interpreter, as it's only working with interpreter values and dynamic types

runtime/argument_validation.go
```diff
@@ -0,0 +1,354 @@
...
+
+	// Currently the type-vs-value mismatch could only happen for composite types, as
+	// that's the only place where the type is loaded from the runtime, based on the name,
+	// rather than based on the value structure of the user input.
+	// Hence the possible set of malformed values are:
+	//     - Composite types.
+	//     - Arrays/Dictionaries/Optionals of composite types.
+
+	switch typ := dynamicType.(type) {
```
turbolent, line 45: Maybe instead of having this large type switch and we forget to update it when adding more types in the future, we could add a function to the `interpreter.DynamicType` interface, e.g. `ValueConforms(value Value) bool`, just like / analogous to the `interpreter.Value.DynamicType()` function which returns the dynamic type for a value?

runtime/argument_validation.go
```diff
@@ -0,0 +1,354 @@
...
+	case interpreter.FunctionDynamicType:
+		_, ok = value.(interpreter.FunctionValue)
+	case interpreter.BlockDynamicType:
+		_, ok = value.(interpreter.BlockValue)
+	case interpreter.DeployedContractDynamicType:
+		_, ok = value.(interpreter.DeployedContractValue)
+	case interpreter.StorageReferenceDynamicType:
+		// Currently this only checks whether the the value is `interpreter.StorageReferenceValue`.
+		// It doesn't check whether the 'referenced' value conforms to the innerType.
+		// TODO: add support for checking the referenced value conformance (if importing is supported).
```
turbolent, line 87: This is actually OK, this comment/TODO can be removed: References are checked when they are used, so we don't have to do it here, and even if we would, it would not be enough, because the referenced value could be changed to a value of a different type

runtime/argument_validation.go
```diff
@@ -0,0 +1,354 @@
...
+		_, ok = value.(*interpreter.StorageReferenceValue)
+	case interpreter.EphemeralReferenceDynamicType:
+		ok = valueConformsToEphemeralReferenceDynamicType(value, typ)
+	case interpreter.VoidDynamicType:
+		// Void type cannot have a value.
+		ok = false
```
turbolent, line 93: There actually is a value, `VoidValue`. It's a bit counter-intuitive, but it is e.g returned from a function that returns `Void`, and could be passed to a function with a `Void` parameter type

runtime/argument_validation.go
```diff
@@ -0,0 +1,354 @@
...
+	return true
+}
+
+func valueConformsToDictionaryDynamicType(
+	value interpreter.Value,
+	dictionaryType interpreter.DictionaryDynamicType,
+) bool {
+
+	dictionaryValue, ok := value.(*interpreter.DictionaryValue)
+	if !ok || len(dictionaryValue.Keys.Values) != len(dictionaryType.EntryTypes) {
```
turbolent, line 120: Could be shortened, but OK as-is:

```suggestion
	if !ok || dictionaryValue.Count() != len(dictionaryType.EntryTypes) {
```

runtime/argument_validation.go
```diff
@@ -0,0 +1,354 @@
...
+
+	for index, entryKey := range dictionaryValue.Keys.Values {
+		entryType := dictionaryType.EntryTypes[index]
+
+		// Check the key
+		if !valueConformsToDynamicType(entryKey, entryType.KeyType) {
+			return false
+		}
+
+		// Check the value. Here it is assumed an imported value can only have
```
turbolent, line 132: This is a good assumption / restriction to make 👍

runtime/argument_validation.go
```diff
@@ -0,0 +1,354 @@
...
+		return false
+	}
+
+	return valueConformsToDynamicType(*referenceValue.ReferencedValue(), refType.InnerType())
+}
+
+// valueConformsToSemaType checks whether a value conforms to a given semantic type.
+// Returns `true` if the value and the type matches, and `false` otherwise.
+//
+func valueConformsToSemaType(value interpreter.Value, semaType sema.Type) (ok bool) {
```
turbolent, line 169: Just as mentioned above this could be maybe a function, potentially inverted, i.e on `interpreter.Value` and taking a `sema.Type` parameter, instead of the other way round (on `sema.Type`) because that would lead to cyclic imports of packages

runtime/argument_validation.go
```diff
@@ -0,0 +1,354 @@
...
+	case interpreter.NumberDynamicType:
+		ok = valueConformsToSemaType(value, typ.StaticType)
+	case interpreter.NilDynamicType:
+		_, ok = value.(interpreter.NilValue)
+	case interpreter.AddressDynamicType:
+		_, ok = value.(interpreter.AddressValue)
+	case interpreter.MetaTypeDynamicType:
+		_, ok = value.(interpreter.TypeValue)
+	case interpreter.CapabilityDynamicType:
+		_, ok = value.(interpreter.CapabilityValue)
```
turbolent, line 71: Needs to consider the (optional) `BorrowType`

runtime/argument_validation.go
```diff
@@ -0,0 +1,354 @@
...
+	case *sema.VariableSizedType:
+		ok = valueConformsToVariableSizedArrayType(value, typ)
+	case *sema.ConstantSizedType:
+		ok = valueConformsToConstantSizedArrayType(value, typ)
+	case *sema.DictionaryType:
+		ok = valueConformsToDictionaryType(value, typ)
+	case *sema.CompositeType:
+		ok = valueConformsToCompositeType(value, typ)
+	case *sema.CapabilityType:
+		_, ok = value.(*interpreter.CapabilityValue)
```
turbolent, line 229: Needs to consider the (optional) `BorrowType`

runtime/argument_validation.go
```diff
@@ -0,0 +1,354 @@
...
+	case sema.BoolType:
+		_, ok = value.(interpreter.BoolValue)
+	case sema.PublicPathType:
+		ok = valueConformsToPathType(value, common.PathDomainPublic)
+	case sema.PrivatePathType:
+		ok = valueConformsToPathType(value, common.PathDomainPrivate)
+	case sema.StoragePathType:
+		ok = valueConformsToPathType(value, common.PathDomainStorage)
+	case sema.PathType, sema.CapabilityPathType:
+		_, ok = value.(interpreter.PathValue)
```
turbolent, line 285: for `CapabilityPathType` should check domain is either public or private

runtime/argument_validation.go
```diff
@@ -0,0 +1,354 @@
...
+	case sema.MetaType:
+		_, ok = value.(interpreter.TypeValue)
+	case sema.VoidType:
+		// false
```
turbolent, line 289: Should allow `VoidValue`

runtime/argument_validation.go
```diff
@@ -0,0 +1,354 @@
...
+	case *sema.FunctionType:
+		_, ok = value.(interpreter.FunctionValue)
+	case *sema.TransactionType:
+		// false
+	}
+
+	return
+}
+
+func valueConformsToCompositeType(arg interpreter.Value, compositeType *sema.CompositeType) bool {
```
turbolent, line 241: Maybe already implicitly checked due to the dynamic check performed by the runtime after-wards, but given that this function should be independent on other usage, also check location and qualified identifier match

turbolent (commented):
This is great @SupunS, thank you for noticing this and implementing this all! 👏

runtime/argument_validation.go
SupunS: Yeah, that's a good idea. Initially, I didn't want to pollute the existing interface and the structs. But yeah, it is easy to miss updating this switch-case.

runtime/argument_validation.go
SupunS: Currently, capabilities are not importable: https://github.com/onflow/cadence/issues/712

I was planning the add complete support together with that.

runtime/argument_validation.go
SupunS: Similarly, will add this with https://github.com/onflow/cadence/issues/712

SupunS:
@turbolent Thanks for the review! 

As suggested, I moved the conformance checking to the `interpreter.Value` by introducing the following two methods:
- `ConformsToDynamicType(dynamicType DynamicType) bool`
- `conformsToSemaType(semaType sema.Type) bool`  - This is a package-private method because we should discourage using it directly, because:
   - It doesn't check for subtyping. It is handled by a wrapper method (`ValueConformsToSemaType`) to avoid repetitive subtyping checks. Anyone who wishes to use this functionality should use `ValueConformsToSemaType` function instead.
    - It is currently only implemented for the values that are importable. For the rest, e.g: `FunctionValue`,  it doesn't do a deep value equality (we can gradually add these, but doesn't require for now).

Can please have a look at the new changes too?

runtime/argument_validation.go
SupunS: added with 36327c9

runtime/argument_validation.go
SupunS: added with 36327c9

runtime/convertValues_test.go
```diff
@@ -1012,8 +1013,716 @@ func importAndExportValuesFromScript(t *testing.T, script string, arg cadence.Va
...
+
+			if !test.skipExport {
+				returnSignature = fmt.Sprintf(": %[1]s", test.typeSignature)
+				returnStmt = "return arg"
+			}
+
+			script := fmt.Sprintf(
+				`pub fun main(arg: %[1]s)%[2]s {
+
+					if !arg.isInstance(Type<%[1]s>()) {
```
turbolent, line 1254: Good idea to check both argument validation (value conforms to static type) and also check dynamic type conformance 👍

runtime/convertValues_test.go
```diff
@@ -1012,8 +1013,716 @@ func importAndExportValuesFromScript(t *testing.T, script string, arg cadence.Va
...
+	typeSignature string
+	exportedValue cadence.Value
+	skipExport    bool
+}
+
+func TestArgumentPassing(t *testing.T) {
+
+	t.Parallel()
+
+	var argumentPassingTests = []argumentPassingTest{
```
turbolent, line 1029: 👍

runtime/interpreter/value.go
```diff
@@ -5649,6 +5910,47 @@ func (v *CompositeValue) TypeID() common.TypeID {
...
+		v.Kind != compositeType.Kind ||
+		v.QualifiedIdentifier != compositeType.QualifiedIdentifier() ||
+		v.Location.ID() != compositeType.Location.ID() {
+
+		return false
+	}
+
+	// Here it is assumed that imported values can only have static fields values,
+	// but not computed field values.
+	if v.Fields.Len() != len(compositeType.Fields) {
```
turbolent, line 5931: 👍

runtime/interpreter/value.go
```diff
@@ -785,6 +827,45 @@ func (v *ArrayValue) Count() int {
...
+			return false
+		}
+		memberType = arrayType.Type
+	default:
+		return false
+	}
+
+	for _, member := range v.Values {
+		if !ValueConformsToSemaType(member, memberType) {
+			return false
```
turbolent, line 862: Minor: I was confused about "members" in arrays at first (we normally only use this term to refer to fields and functions in composites), so maybe use "element" for arrays here:

```suggestion
	var elementType sema.Type

	switch arrayType := semaType.(type) {
	case *sema.VariableSizedType:
		elementType = arrayType.Type
	case *sema.ConstantSizedType:
		if int64(len(v.Values)) != arrayType.Size {
			return false
		}
		elementType = arrayType.Type
	default:
		return false
	}

	for _, element := range v.Values {
		if !ValueConformsToSemaType(element, elementType) {
			return false
```

runtime/interpreter/value.go
```diff
@@ -5649,6 +5910,47 @@ func (v *CompositeValue) TypeID() common.TypeID {
 	return v.Location.TypeID(v.QualifiedIdentifier)
 }
 
+func (v *CompositeValue) ConformsToDynamicType(dynamicType DynamicType) bool {
+	compositeType, ok := dynamicType.(CompositeDynamicType)
+	return ok && ValueConformsToSemaType(v, compositeType.StaticType)
```
turbolent, line 5915: I wonder if we should include composites' fields' dynamic types in the composite dynamic type, just like we include the dynamic types of arrays' elements, and then check them here?

runtime/interpreter/value.go
```diff
@@ -6549,6 +6931,34 @@ func (v *EphemeralReferenceValue) Equal(_ *Interpreter, other Value) BoolValue {
...
+
+	referencedValue := v.ReferencedValue()
+	if referencedValue == nil {
+		return false
+	}
+
+	return (*referencedValue).ConformsToDynamicType(refType.InnerType())
+}
+
+func (v *EphemeralReferenceValue) conformsToSemaType(semaType sema.Type) bool {
```
turbolent, line 6948: I realized in another PR that ephemeral references are a special case, in that they allow recursive object graphs, e.g. a reference could have an implicit reference to itself.

We need to handle this potential recursion here. I already implemented handling recursion for the `DynamicType()` function in https://github.com/dapperlabs/cadence-internal/commit/ffb32bcb3db13c4c8a3fc974203af0b183476ff6. 

We shouldn't do this in this PR, so it doesn't grow too large, but could you please create a follow-up issue to handle recursion in `ConformsToDynamicType` and `conformsToSemaType` (as both use `ReferencedValue`)?

runtime/interpreter/value.go
```diff
@@ -7019,6 +7457,42 @@ func (v PathValue) KeyString() string {
...
+	case PrivatePathDynamicType:
+		return v.Domain == common.PathDomainPrivate
+	case StoragePathDynamicType:
+		return v.Domain == common.PathDomainStorage
+	default:
+		return false
+	}
+}
+
+func (v PathValue) conformsToSemaType(semaType sema.Type) bool {
```
turbolent, line 7473: 👍

turbolent (commented):
Thank you for refactoring this @SupunS!

I thought about this a bit more and have the feeling this could be simplified a lot: 
Currently adding this feature requires a lot of code to be added, but we already have most of the same checks in `interpreter.IsSubType`, the function that validates if a dynamic type (which should have all necessary information about a value) to a static type. 

OK If I'll try to see if I can maybe reduce the code a bit and rely on what we already have? I'd open another PR against this one

SupunS:
> Thank you for refactoring this @SupunS!
> 
> I thought about this a bit more and have the feeling this could be simplified a lot:
> Currently adding this feature requires a lot of code to be added, but we already have most of the same checks in `interpreter.IsSubType`, the function that validates if a dynamic type (which should have all necessary information about a value) to a static type.
> 
> OK If I'll try to see if I can maybe reduce the code a bit and rely on what we already have? I'd open another PR against this one

Sure, please do! 

I also had a look at the `isSubType()` method - why I couldn't use it because it mostly does a nominal type check, whereas here we actually do sort of a structural type check. But maybe worth looking from a fresh pair of eyes!

Also maybe its worth merging https://github.com/onflow/cadence/pull/732 prior to this, as it also removes a few (a lot, infact) changes added by this PR (pretty much all the numeric types get reduced to just two)

turbolent:
> I also had a look at the isSubType() method - why I couldn't use it because it mostly does a nominal type check, whereas here we actually do sort of a structural type check. But maybe worth looking from a fresh pair of eyes!

Right, I was mostly thinking the same, but then realized how much overlap / duplication there is and it might be good to re-consider.

> Also maybe its worth merging #732 prior to this,

Please do @SupunS!

runtime/interpreter/value.go
SupunS: Added in 4fd3be9

runtime/interpreter/value.go
SupunS: Yup, that would be ideal.

If we have that, we can also get rid of, all the `value.conformsToSemaType()` methods, because this is the only entry point to that method.

SupunS: oh wait, that could be a problem when someone changes the value of a field. because we don't update the dynamic type upon the field updates. Wonder how it works for the arrays.

turbolent:
@SupunS Opened #776 against this PR to reduce the code, let me know what you think!

runtime/interpreter/value.go
SupunS: Created https://github.com/onflow/cadence/issues/777

SupunS:
Thanks, @turbolent for the cleanup! I merged your PR, and synched/rebased this branch with the master.

Opened https://github.com/onflow/cadence/issues/777 to handle the potential cyclic references.

runtime/convertValues.go
```diff
@@ -196,6 +196,9 @@ func exportCompositeValue(v *interpreter.CompositeValue, inter *interpreter.Inte
 		fields[i] = exportValueWithInterpreter(fieldValue, inter, results)
 	}
 
+	// NOTE: when modifying the cases below,
+	// also update error message below!
```
turbolent, line 200:
```suggestion
	// also update the error message below!
```

turbolent (approved):
Looks good, great work! 👍

PR #727, opened 2021-03-25
Add array appendAll function

### LINKED ISSUE #631: Add a function to Array to append another array
### Issue To Be Solved

It is currently costly and inefficient to append the elements of one array to another, as there is only an `append` function for adding one element to an array, which has to be called in a loop.

### Suggested Solution

- [ ] Add a function to `Array`, which has a parameter of an equally typed array, and that appends all elements of the given array to the array on which the function is called. 
  Potential names: `appendAll`, `appendContents`, `extend`.

### Context

Querying the IDs of a shared Top Shot moment collection is currently inefficient:
https://github.com/dapperlabs/nba-smart-contracts/blob/75a5858873b65121e3a3bd6e2254e1c818df95a6/contracts/TopShotShardedCollection.cdc#L133-L135

lkadian:
Closes #631

## Description

Chose "extend" for the name, let me know what the preferred name is!

______

For contributor use:

### DIFF
--- a/runtime/tests/checker/arrays_dictionaries_test.go
+++ b/runtime/tests/checker/arrays_dictionaries_test.go
@@ -369,6 +369,70 @@ func TestCheckInvalidArrayAppendToConstantSize(t *testing.T) {
 	assert.IsType(t, &sema.NotDeclaredMemberError{}, errs[0])
 }
 
+func TestCheckArrayAppendAll(t *testing.T) {
+
+	t.Parallel()
+
+	_, err := ParseAndCheck(t, `
+	  fun test(): [Int] {
+	 	  let a = [1, 2]
+		  let b = [3, 4]
+		  a.appendAll(b)
+		  return a
+      }
+    `)
+
+	require.NoError(t, err)
+}
+
+func TestCheckInvalidArrayAppendAll(t *testing.T) {
+
+	t.Parallel()
+
+	_, err := ParseAndCheck(t, `
+	  fun test(): [Int] {
+	 	  let a = [1, 2]
+		  let b = ["a", "b"]
+		  a.appendAll(b)
+		  return a
+      }
+    `)
+
+	errs := ExpectCheckerErrors(t, err, 1)
+
+	assert.IsType(t, &sema.TypeMismatchError{}, errs[0])
+
+	_, err = ParseAndCheck(t, `
+	  fun test(): [Int] {
+	 	  let a = [1, 2]
+		  let b = 3
+		  a.appendAll(b)
+		  return a
+      }
+    `)
+
+	errs = ExpectCheckerErrors(t, err, 1)
+
+	assert.IsType(t, &sema.TypeMismatchError{}, errs[0])
+}
+
+func TestCheckInvalidArrayAppendAllOnConstantSize(t *testing.T) {
+
+	t.Parallel()
+
+	_, err := ParseAndCheck(t, `
+      fun test(): [Int; 3] {
+          let x: [Int; 3] = [1, 2, 3]
+          x.appendAll([4, 5])
+          return x
+      }
+    `)
+
+	errs := ExpectCheckerErrors(t, err, 1)
+
+	assert.IsType(t, &sema.NotDeclaredMemberError{}, errs[0])
+}
+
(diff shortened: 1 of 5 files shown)


joshuahannan:
Isn't this the same as concact? https://docs.onflow.org/cadence/language/values-and-types/#array-fields-and-functions

lkadian:
> Isn't this the same as concact? https://docs.onflow.org/cadence/language/values-and-types/#array-fields-and-functions

I was 

> Isn't this the same as concact? https://docs.onflow.org/cadence/language/values-and-types/#array-fields-and-functions

I had the same concern, probably should've asked before comitting! The difference is that .extend(...) modifies the existing array, while concat returns a new array while leaving the original untouched.

runtime/tests/checker/arrays_dictionaries_test.go
```diff
@@ -369,6 +369,70 @@ func TestCheckInvalidArrayAppendToConstantSize(t *testing.T) {
...
+	assert.IsType(t, &sema.TypeMismatchError{}, errs[0])
+}
+
+func TestCheckInvalidArrayExtendOnConstantSize(t *testing.T) {
+
+	t.Parallel()
+
+	_, err := ParseAndCheck(t, `
+      fun test(): [Int; 3] {
+          let x: [Int; 3] = [1, 2, 3]
```
turbolent, line 425: Nice work only adding this function for variable-sized arrays and testing it is not supported for fixed-size arrays 👍

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -4666,6 +4719,30 @@ func TestInterpretArrayConcatBound(t *testing.T) {
 	)
 }
 
+func TestInterpretArrayConcatDoesNotModifyOriginalArray(t *testing.T) {
```
turbolent, line 4722: 👌

turbolent (approved):
Nice, thank you for contributing this Laurent!

Regarding naming:

- Both Rust and Python call this function `extend` 
- Swift calls it `append(contentsOf:)`, an overload of `append` for multiple elements. The single value function is just `append`. In Cadence we don't have argument label-based overloading (yet)
- Kotlin calls the function `addAll`, the single value function `add`
- Java and Scala call the function `appendAll`

`extend` sounds good to me. Given that the existing single function is called `append`, we could also go with `appendAll`.

Any last thoughts on this @lkadian @SupunS @joshuahannan ?

SupunS:
> Nice, thank you for contributing this Laurent!
> 
> Regarding naming:
> 
> * Both Rust and Python call this function `extend`
> * Swift calls it `append(contentsOf:)`, an overload of `append` for multiple elements. The single value function is just `append`. In Cadence we don't have argument label-based overloading (yet)
> * Kotlin calls the function `addAll`, the single value function `add`
> * Java and Scala call the function `appendAll`
> 
> `extend` sounds good to me. Given that the existing single function is called `append`, we could also go with `appendAll`.
> 
> Any last thoughts on this @lkadian @SupunS @joshuahannan ?
 
Both look fine for me. Maybe `appendAll` might feel juuust a bit more natural for someone who is familiar with `append` (already existing one).

lkadian:
> Nice, thank you for contributing this Laurent!
> 
> Regarding naming:
> 
> * Both Rust and Python call this function `extend`
> * Swift calls it `append(contentsOf:)`, an overload of `append` for multiple elements. The single value function is just `append`. In Cadence we don't have argument label-based overloading (yet)
> * Kotlin calls the function `addAll`, the single value function `add`
> * Java and Scala call the function `appendAll`
> 
> `extend` sounds good to me. Given that the existing single function is called `append`, we could also go with `appendAll`.
> 
> Any last thoughts on this @lkadian @SupunS @joshuahannan ?

No strong opinion either way for me

turbolent:
I asked on Discord and it seems like `appendAll` is the favourite. 

@lkadian Could you please rename it? Thank you!

lkadian:
> I asked on Discord and it seems like `appendAll` is the favourite.
> 
> @lkadian Could you please rename it? Thank you!

done!

turbolent (approved):
Looks good!

PR #736, opened 2021-03-28
Add address field to capabilities

### LINKED ISSUE #721: Getting the `Address` from a `Capability`
Currently there's no way to get the `Address` from a `Capability`.

My use case was that I had a function which accepts a `Capability`, and I wanted to emit an event that records the address, like this:

```
pub registerCapability(c: Capability) {
    emit CapabilityRegistered(address: c.getAddress()) // this isn't possible
}
```

ceelo777:
Closes #721

## Description
This PR should add the ability to grab the address field from a capability.
______

For contributor use:

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -7070,6 +7070,7 @@ func (v CapabilityValue) GetMember(inter *Interpreter, _ func() LocationRange, n
 		}
 		return inter.capabilityCheckFunction(v.Address, v.Path, borrowType)
 
+
 	case "address":
 		return v.Address
 	}
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -7069,6 +7069,7 @@ func (v CapabilityValue) GetMember(inter *Interpreter, _ func() LocationRange, n
 			borrowType = inter.ConvertStaticToSemaType(v.BorrowType).(*sema.ReferenceType)
 		}
 		return inter.capabilityCheckFunction(v.Address, v.Path, borrowType)
+
 	case "address":
 		return v.Address
 	}
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -7069,7 +7069,8 @@ func (v CapabilityValue) GetMember(inter *Interpreter, _ func() LocationRange, n
 			borrowType = inter.ConvertStaticToSemaType(v.BorrowType).(*sema.ReferenceType)
 		}
 		return inter.capabilityCheckFunction(v.Address, v.Path, borrowType)
-
+	case "address":
+		return v.Address
 	}
 
 	return nil
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -6527,6 +6527,10 @@ const capabilityTypeCheckFunctionDocString = `
 Returns true if the capability currently targets an object that satisfies the given type, i.e. could be borrowed using the given type
 `
 
+const addressTypeCheckFunctionDocString = `
+Returns the address for a capability
+`
+
 func (t *CapabilityType) GetMembers() map[string]MemberResolver {
 	return withBuiltinMembers(t, map[string]MemberResolver{
 		"borrow": {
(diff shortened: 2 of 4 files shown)


ceelo777:
So this isn't finished @turbolent - I'm having a weird issue with the test I wrote -

`func TestCheckCapability_address(t *testing.T) {

	t.Parallel()
	t.Run("check address", func(t *testing.T) {
		checker, err := ParseAndCheckWithPanic(t,
			`		  			
				let capability: Capability = panic("")
				let addr = capability.address
			`,
		)
		require.NoError(t, err)

		addrType := RequireGlobalType(t, checker.Elaboration, "addr")
		require.Equal(t, sema.AddressType{}, addrType)
	})
}
`

I'm getting this error - "global type 'addr' missing" even though it seems that I'm initializing it? Not sure exactly why this is happening.

runtime/tests/checker/capability_test.go
```diff
@@ -434,3 +434,20 @@ func TestCheckCapability_check(t *testing.T) {
...
+	t.Run("check address", func(t *testing.T) {
+		checker, err := ParseAndCheckWithPanic(t,
+			`		  			
+				let capability: Capability = panic("")
+				let addr = capability.address
+			`,
+		)
+		require.NoError(t, err)
+
+		addrType := RequireGlobalType(t, checker.Elaboration, "addr")
```
SupunS, line 450: I think the issue is, `addr` is a variable but not a type. Using `RequireGlobalValue` would solve the test failure.  
```suggestion
		addrType := RequireGlobalValue(t, checker.Elaboration, "addr")
```

runtime/tests/checker/capability_test.go
```diff
@@ -434,3 +434,20 @@ func TestCheckCapability_check(t *testing.T) {
...
+		require.Equal(t, sema.AddressType{}, addrType)
```
turbolent, line 451:
```suggestion
		require.Equal(t, &sema.AddressType{}, addrType)
```

turbolent (commented):
Thank you for contributing this!

Looks good, just the test needs to be fixed and then it can be merged

runtime/tests/checker/capability_test.go
turbolent: @ceelo777 like Supun pointed out here, you want to get the type of the variable declaration, which is a value. 

`RequireGlobalType` is used to the the type introduced by a type declaration, e.g. a struct or resource declaration.

runtime/tests/checker/capability_test.go
ceelo777: Thanks! It worked, updating the PR now.

turbolent:
@ceelo777 Looks good, glad the  test passes now. Just one last thing: Could you please add a test for the interpreter?

ceelo777:
To be honest, I'm not sure if I did this interpreter test correctly but it passed - anything else I should add to it?

runtime/tests/interpreter/capability_test.go
```diff
@@ -657,3 +657,46 @@ func TestInterpretCapability_check(t *testing.T) {
...
+
+				fun nonExistent() : Address {
+					return account.getCapability(/public/nonExistent).address
+				}				
+			`,
+		)
+
+		t.Run("single", func(t *testing.T) {
+			_, err := inter.Invoke("single")
+			require.NoError(t, err)
```
turbolent, line 689: Here and in the tests below, assert that the result value, that is currently being ignored, is in fact of type `interpreter.AddressValue`:

```suggestion
			value, err := inter.Invoke("single")
			require.NoError(t, err)
			
			require.IsType(t, interpreter.AddressValue{}, value)
```

turbolent (commented):
The test looks good! 👍 

Good idea to use the `testAccount` helper.
One last thing would be to check the result values, then we can merge

runtime/tests/interpreter/capability_test.go
turbolent: Resolved in 11655c8

runtime/interpreter/value.go
```diff
@@ -7069,7 +7069,8 @@ func (v CapabilityValue) GetMember(inter *Interpreter, _ func() LocationRange, n
 			borrowType = inter.ConvertStaticToSemaType(v.BorrowType).(*sema.ReferenceType)
 		}
 		return inter.capabilityCheckFunction(v.Address, v.Path, borrowType)
-
+	case "address":
```
turbolent, line 7072:
```suggestion

	case "address":
```

turbolent, line 7073:
```suggestion

	case "address":
```

runtime/sema/type.go
```diff
@@ -6527,6 +6527,10 @@ const capabilityTypeCheckFunctionDocString = `
 Returns true if the capability currently targets an object that satisfies the given type, i.e. could be borrowed using the given type
 `
 
+const addressTypeCheckFunctionDocString = `
+Returns the address for a capability
```
turbolent, line 6531:
```suggestion
The address of the capability
```

turbolent (approved):
Nice work, thanks again for adding this @ceelo777!

PR #749, opened 2021-03-31
Allow iteration over resource arrays.

### LINKED ISSUE #704: Enable iteration over resource arrays
### Issue To Be Solved

Currently the for-in loop is only supported for struct arrays, not resource arrays.

Enable for-in loops to iterate over arrays of resources, binding an authorized reference:

```kotlin
pub resource R {
    pub let name: String

    init(name: String) {
        self.name = name
    }
}

let rs <- [
    <-create R(name: "bar"), 
    <-create R(name: "foo")
]
for ref in rs {
	// ref is `auth &R`
    log(ref.name)
}

destroy rs
```

### Definition of Done

- [ ] Checker: Bind authorized references when looping over resource arrays
- [ ] Interpreter: Bind authorized references when looping over resource arrays
- [ ] Tests
- [ ] Documentation

ceelo777:
Closes #704

## Description
Remove resource type blocker.
Check for resource loss.
______

For contributor use:

- [Y] Targeted PR against `master` branch
- [Y] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [Y] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [Y] Updated relevant documentation 
- [Y] Re-reviewed `Files changed` in the Github PR explorer
- [Y] Added appropriate labels

### DIFF
--- a/runtime/sema/check_for.go
+++ b/runtime/sema/check_for.go
@@ -35,19 +35,9 @@ func (checker *Checker) VisitForStatement(statement *ast.ForStatement) ast.Repr 
 
 	if !valueType.IsInvalidType() {
 
-		// Only get the element type if the array is not a resource array.
-		// Otherwise, in addition to the `UnsupportedResourceForLoopError`,
-		// the loop variable will be declared with the resource-typed element type,
-		// leading to an additional `ResourceLossError`.
-
-		if valueType.IsResourceType() {
-			checker.report(
-				&UnsupportedResourceForLoopError{
-					Range: ast.NewRangeFromPositioned(valueExpression),
-				},
-			)
-		} else if arrayType, ok := valueType.(ArrayType); ok {
+		if arrayType, ok := valueType.(ArrayType); ok {
 			elementType = arrayType.ElementType(false)
+			checker.checkUnusedExpressionResourceLoss(elementType, valueExpression)
 		} else {
 			checker.report(
 				&TypeMismatchWithDescriptionError{
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -7069,7 +7069,8 @@ func (v CapabilityValue) GetMember(inter *Interpreter, _ func() LocationRange, n
 			borrowType = inter.ConvertStaticToSemaType(v.BorrowType).(*sema.ReferenceType)
 		}
 		return inter.capabilityCheckFunction(v.Address, v.Path, borrowType)
-
+	case "address":
+		return v.Address
 	}
 
 	return nil
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -6527,6 +6527,10 @@ const capabilityTypeCheckFunctionDocString = `
 Returns true if the capability currently targets an object that satisfies the given type, i.e. could be borrowed using the given type
 `
 
+const addressTypeCheckFunctionDocString = `
+Returns the address for a capability
+`
+
 func (t *CapabilityType) GetMembers() map[string]MemberResolver {
 	return withBuiltinMembers(t, map[string]MemberResolver{
 		"borrow": {
@@ -6548,6 +6552,17 @@ func (t *CapabilityType) GetMembers() map[string]MemberResolver {
 					identifier,
 					capabilityTypeCheckFunctionType(t.BorrowType),
 					capabilityTypeCheckFunctionDocString,
+				)
+			},
+		},
+		"address": {
+			Kind: common.DeclarationKindField,
+			Resolve: func(identifier string, _ ast.Range, _ func(error)) *Member {
+				return NewPublicConstantFieldMember(
+					t,
+					identifier,
+					&AddressType{},
+					addressTypeCheckFunctionDocString,
 				)
 			},
 		},
--- a/runtime/tests/checker/capability_test.go
+++ b/runtime/tests/checker/capability_test.go
@@ -434,3 +434,20 @@ func TestCheckCapability_check(t *testing.T) {
 		})
 	})
 }
+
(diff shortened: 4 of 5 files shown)


ceelo777:
This isn't done yet (once again) - just want to make sure I'm on the right path @turbolent. I called the function for resource loss and removed the if statement prohibiting iteration over resource arrays.

I'm struggling with how to go about declaring the iteration variable as the element type for the checker and setting the iteration value to each element of the array. Two questions:

1) What exactly is an authorized reference of an element type? 
2) What is the authorized ephemeral reference to an element?

Any references to other code snippets or general answers would help, thanks!

turbolent:
@ceelo777 No worries, happy to guide you through this, no rush!

I'll have a look at the current changes.

As to your question, have a look at the language reference: https://docs.onflow.org/cadence/language/references/. Basically, normal references cannot be downcasted, while authorized references can. 

Unauthorized references (`&T`) are usually granted and used by third-parties to interact with objects in storage. Here, it is usually desired to hand out a reference that only allows interacting with the referenced value by using the functionality of the given type. However, knowledge about the underlying concrete type of the value shouldn't allow someone to use more functionality then they were granted by downcasting the reference. 

For example, if I give you a reference to the `FlowToken.Vault` that is in my storage, I can give you a `&FlowToken.Vault{FungibleToken.Balance}`, which is a restricted type specifying that only the functionality of `FungibleToken.Balance` may be used on the value of type `FlowToken.Vault`. So even though you have now a reference to the whole vault, you can only access the balance, but not downcast the reference to a `&FlowToken.Vault`, which grants more permissions, e.g. withdrawal.

Authorized references (`auth &T`) are usually used when interacting with data that the user of the reference also owns. For example, iterating over the resources requires owning them, so it is safe to make the reference authorized. 

A good exemplary use case is iterating over an array of NFTs: The array stores multiple different `NonFungibleToken.NFT`, but each element has a type that is a concrete implementation of the `NonFungibleToken.NFT` interface. Making the iteration variable and authorized reference instead of an unauthorized one allows downcasting the reference of type `auth &NonFungibleToken.NFT` to the concrete type, e.g. `&TopShot.NFT` (if that is what the element's type is):

```kotlin
let nfts: @[NonFungibleToken.NFT] <- // ...

for ref in nfts {
    // `ref` is `auth &NonFungibleToken.NFT`

    // attempt to failably downcast
    if momentRef = ref as? &TopShot.NFT { 

        // `momentRef` is `&TopShot.NFT`
    }
}

```

runtime/sema/check_for.go
```diff
@@ -35,19 +35,9 @@ func (checker *Checker) VisitForStatement(statement *ast.ForStatement) ast.Repr
...
-		if valueType.IsResourceType() {
-			checker.report(
-				&UnsupportedResourceForLoopError{
-					Range: ast.NewRangeFromPositioned(valueExpression),
-				},
-			)
-		} else if arrayType, ok := valueType.(ArrayType); ok {
+		if arrayType, ok := valueType.(ArrayType); ok {
 			elementType = arrayType.ElementType(false)
+			checker.checkUnusedExpressionResourceLoss(elementType, valueExpression)
```
turbolent, line 40: The first argument type is wrong here, the type of the `valueExpression` is expected:

```suggestion
			checker.checkUnusedExpressionResourceLoss(valueType, valueExpression)
```

Also, this line is not conditional on the value being an array, the resource loss should always be checked, i.e. move it above the array type check

ceelo777:
> @ceelo777 No worries, happy to guide you through this, no rush!
> 
> I'll have a look at the current changes.
> 
> As to your question, have a look at the language reference: https://docs.onflow.org/cadence/language/references/. Basically, normal references cannot be downcasted, while authorized references can.
> 
> Unauthorized references (`&T`) are usually granted and used by third-parties to interact with objects in storage. Here, it is usually desired to hand out a reference that only allows interacting with the referenced value by using the functionality of the given type. However, knowledge about the underlying concrete type of the value shouldn't allow someone to use more functionality then they were granted by downcasting the reference.
> 
> For example, if I give you a reference to the `FlowToken.Vault` that is in my storage, I can give you a `&FlowToken.Vault{FungibleToken.Balance}`, which is a restricted type specifying that only the functionality of `FungibleToken.Balance` may be used on the value of type `FlowToken.Vault`. So even though you have now a reference to the whole vault, you can only access the balance, but not downcast the reference to a `&FlowToken.Vault`, which grants more permissions, e.g. withdrawal.
> 
> Authorized references (`auth &T`) are usually used when interacting with data that the user of the reference also owns. For example, iterating over the resources requires owning them, so it is safe to make the reference authorized.
> 
> A good exemplary use case is iterating over an array of NFTs: The array stores multiple different `NonFungibleToken.NFT`, but each element has a type that is a concrete implementation of the `NonFungibleToken.NFT` interface. Making the iteration variable and authorized reference instead of an unauthorized one allows downcasting the reference of type `auth &NonFungibleToken.NFT` to the concrete type, e.g. `&TopShot.NFT` (if that is what the element's type is):
> 
> ```kotlin
> let nfts: @[NonFungibleToken.NFT] <- // ...
> 
> for ref in nfts {
>     // `ref` is `auth &NonFungibleToken.NFT`
> 
>     // attempt to failably downcast
>     if momentRef = ref as? &TopShot.NFT { 
> 
>         // `momentRef` is `&TopShot.NFT`
>     }
> }
> ```

This makes a lot more sense now - thanks for explaining it so well!

ceelo777:
One more question - how do I go about setting the elements to be authorized references or authorized ephemeral references?

turbolent:
No progress in a long time, we may revisit this after Stable Cadence

Issue #738, opened 2021-03-29
Optimize CBOR representation of data in Cadence storage layer


fxamacker:
### Issue To Be Solved

Cadence storage layer's use of CBOR can be optimized to improve speed, memory use, and stored data size.

### Suggestion

Performance gains are possible by tweaking the CBOR representation of Cadence `Value` types.  For example, replacing CBOR map with CBOR array when feasible.

In addition to performance gains, replacing CBOR map with CBOR array preserves order (e.g. `DictionaryValue.Entries`).

UPDATE:  Additional performance gains (especially for decoding) are possible by removing backwards compatible decoding (thanks @turbolent) and not saving key strings twice (thanks @SupunS).  

No changes to [fxamacker/cbor](https://github.com/fxamacker/cbor) are required to support these optimizations.

The current CBOR representation in Cadence is very nice — so this optimization is kind of like taking a nicely normalized RDBMS schema and denormalizing some of it to gain speed.

### Performance Comparisons

A preliminary comparison (from PR #752) with non-production data showed potential performance improvements. Although faster execution speed was the primary objective, memory use and stored data size also improved.

Comparisons using production data (in PR #761) shows performance gains are cumulative.  On a very large value (776,155 bytes), optimization of composite + dictionary types showed:
 - encoding is 44.96% faster, 41.99% less memory, 46.02% fewer allocs
 - decoding is 34.48% faster, 29.67% less memory, 18.56% fewer allocs

Additional performance gains are possible by optimizing remaining `Value` types.  Smaller data will probably begin to catch up in percent improved as more remaining types are optimized.  Larger data with big composites and dictionaries will show more modest gains with remaining types.

Cumulative performance gains from PR #752, PR #761, PR #778, and PR #788 confirmed expectations about small data catching up in performance gains.

For a 167-byte LinkValue (extracted from a mainnet snapshot): 
 - encoding is 50.54% faster, 57.81% less kB per alloc, 54.46% fewer allocs/op
 - decoding is 31.43% faster, 47.60% less kB per alloc, same allocs/op

For a 776155-byte CompositeValue (extracted from a mainnet snapshot):
 - encoding is 45.23% faster, 48.25% less kB per alloc, 51.57% fewer allocs/op
 - decoding is 35.83% faster, 31.23% less kB per alloc, 18.56% fewer allocs/op

Additional performance gains are possible by:
 - optimizing code in the Cadence storage layer
 - adding higher-performance API to fxamacker/cbor
[...]

PR #761, opened 2021-04-04
Optimize encoding of composites

### LINKED ISSUE #744: Optimize encoding of composites
### Issue To Be Solved

The encoding of composites to CBOR values can be optimized.

###  Suggestion

- Switch from CBOR map to CBOR array

### Definition of Done

- [ ] Optimize composite encoding
- [ ] Change decoding to only support this current format (i.e. remove old backwards-compatibility decoding code)
- [ ] Remove unnecessary field sorting when decoding
- [ ] Tests
  - [ ] Encode new format
  - [ ] Decode new format
  - [ ] Decode old format, incl. round-trip, i.e. encoding should match new encoding

fxamacker:
Closes #744
Updates #738

### Changes
 - Switch from CBOR map to CBOR array to improve speed and preserve ordering.
 - Remove unnecessary field sorting when decoding.
 - Remove old backwards-compatibility decoding code (decoding type ID).
 - Add tests, including round-trip for decoding old format and encoding new format.

### Benchmark comparisons

Encoding speed improved by 44.35% and decoding speed improved by 34.56% for a 776,155 byte value from production data that contains both composites and dictionaries.  Memory use also improved.  Encoded/decoded data includes Cadence `Value` types that have not yet been optimized.

Speed gains were only 15-27% for encoding and 16-24% for decoding for smaller production data that did not include dictionaries.  However, these will gain more from optimizing remaining `Value` types (outside the scope of this PR).

__Click to expand:__

<details>
  <summary> Big Value with Composites and Dictionaries 🚀🚀 </summary><p>
  
#### Encoding Comparisons
```
name                                           old time/op    new time/op    delta
EncodeComposite/870comp_837dict_776155bytes-4    35.2ms ± 1%    19.6ms ± 1%  -44.35%  (p=0.000 n=10+10)

name                                           old alloc/op   new alloc/op   delta
EncodeComposite/870comp_837dict_776155bytes-4    7.50MB ± 5%    4.35MB ± 0%  -41.99%  (p=0.000 n=10+10)

name                                           old allocs/op  new allocs/op  delta
EncodeComposite/870comp_837dict_776155bytes-4      141k ± 0%       76k ± 0%  -46.02%  (p=0.000 n=10+10)
```

#### Decoding Comparisons ####
```
name                                           old time/op    new time/op    delta
DecodeComposite/870comp_837dict_776155bytes-4    40.3ms ± 1%    26.3ms ± 1%  -34.65%  (p=0.000 n=10+10)

name                                           old alloc/op   new alloc/op   delta
DecodeComposite/870comp_837dict_776155bytes-4    15.0MB ± 0%    10.5MB ± 0%  -29.67%  (p=0.000 n=10+10)

name                                           old allocs/op  new allocs/op  delta
DecodeComposite/870comp_837dict_776155bytes-4      265k ± 0%      216k ± 0%  -18.56%  (p=0.000 n=10+10)
```

</details>

<details>
  <summary> Values with Composites and No Dictionaries  🚀 </summary><p>

#### Encoding Comparisons
```
name                                           old time/op    new time/op    delta
EncodeComposite/1comp_0dict_139bytes-4           16.0µs ± 1%    13.6µs ± 1%  -15.27%  (p=0.000 n=10+10)
EncodeComposite/2comp_0dict_160bytes-4           20.9µs ± 1%    15.7µs ± 1%  -25.16%  (p=0.000 n=10+10)
EncodeComposite/104comp_0dict_14850bytes-4        910µs ± 1%     657µs ± 1%  -27.78%  (p=0.000 n=9+10)

name                                           old alloc/op   new alloc/op   delta
EncodeComposite/1comp_0dict_139bytes-4           3.79kB ± 1%    2.98kB ± 2%  -21.32%  (p=0.000 n=10+10)
EncodeComposite/2comp_0dict_160bytes-4           4.85kB ± 2%    3.36kB ± 0%  -30.74%  (p=0.000 n=9+10)
EncodeComposite/104comp_0dict_14850bytes-4        247kB ± 1%     188kB ± 0%  -24.16%  (p=0.000 n=8+10)

name                                           old allocs/op  new allocs/op  delta
EncodeComposite/1comp_0dict_139bytes-4             76.0 ± 0%      62.0 ± 0%  -18.42%  (p=0.000 n=10+9)
EncodeComposite/2comp_0dict_160bytes-4              103 ± 0%        73 ± 0%  -29.13%  (p=0.000 n=7+10)
EncodeComposite/104comp_0dict_14850bytes-4        5.03k ± 0%     3.45k ± 0%  -31.28%  (p=0.000 n=8+10)
```

#### Decoding Comparisons ####
```
name                                           old time/op    new time/op    delta
DecodeComposite/1comp_0dict_139bytes-4           11.6µs ± 1%     9.7µs ± 1%  -16.38%  (p=0.000 n=9+10)
DecodeComposite/2comp_0dict_160bytes-4           16.9µs ± 1%    12.7µs ± 1%  -24.47%  (p=0.000 n=10+9)
DecodeComposite/104comp_0dict_14850bytes-4        757µs ± 1%     569µs ± 1%  -24.90%  (p=0.000 n=9+10)

name                                           old alloc/op   new alloc/op   delta
DecodeComposite/1comp_0dict_139bytes-4           3.90kB ± 0%    3.35kB ± 0%  -13.96%  (p=0.000 n=10+10)
DecodeComposite/2comp_0dict_160bytes-4           5.54kB ± 0%    4.38kB ± 0%  -20.92%  (p=0.000 n=10+10)
DecodeComposite/104comp_0dict_14850bytes-4        296kB ± 0%     238kB ± 0%  -19.61%  (p=0.000 n=10+10)

name                                           old allocs/op  new allocs/op  delta
DecodeComposite/1comp_0dict_139bytes-4             56.0 ± 0%      53.0 ± 0%   -5.36%  (p=0.000 n=10+10)
DecodeComposite/2comp_0dict_160bytes-4             97.0 ± 0%      89.0 ± 0%   -8.25%  (p=0.000 n=10+10)
DecodeComposite/104comp_0dict_14850bytes-4        5.05k ± 0%     4.65k ± 0%   -7.98%  (p=0.000 n=10+10)
```

</details>

Benchmarks used Go 1.15.10 on linux_amd64 using production data.

Special thanks to @turbolent for the state decoding tool in #759 and @ramtinms for production data in jsonl!
______

For contributor use:

- [x] Targeted PR against `feature/storage-optimizations` branch

### DIFF
--- a/runtime/interpreter/decode.go
+++ b/runtime/interpreter/decode.go
@@ -25,7 +25,6 @@ import (
 	"io"
 	"math"
 	"math/big"
-	"sort"
 	"strconv"
 	"strings"
 
@@ -528,13 +527,21 @@ func (d *DecoderV4) decodeAddressLocation(v interface{}) (common.Location, error
 }
 
 func (d *DecoderV4) decodeComposite(v interface{}, path []string) (*CompositeValue, error) {
-
-	encoded, ok := v.(map[interface{}]interface{})
+	encoded, ok := v.(cborArray)
 	if !ok {
 		return nil, fmt.Errorf(
 			"invalid composite encoding (@ %s): %T",
 			strings.Join(path, "."),
 			v,
+		)
+	}
+
+	if len(encoded) != encodedCompositeValueLength {
+		return nil, fmt.Errorf(
+			"invalid composite encoding (@ %s): invalid size: expected %d, got %d",
+			strings.Join(path, "."),
+			encodedCompositeValueLength,
+			len(encoded),
 		)
 	}
 
@@ -549,55 +556,16 @@ func (d *DecoderV4) decodeComposite(v interface{}, path []string) (*CompositeVal
 		)
 	}
 
-	// Qualified identifier or Type ID.
-	//
-	// An earlier version of the format stored the whole type ID.
-	// However, the composite already stores the location,
-	// so the current version of the format only stores the qualified identifier.
-
-	var qualifiedIdentifier string
+	// Qualified identifier
 
 	qualifiedIdentifierField := encoded[encodedCompositeValueQualifiedIdentifierFieldKey]
-	if qualifiedIdentifierField != nil {
-		qualifiedIdentifier, ok = qualifiedIdentifierField.(string)
-		if !ok {
-			return nil, fmt.Errorf(
-				"invalid composite qualified identifier encoding (@ %s): %T",
-				strings.Join(path, "."),
-				qualifiedIdentifierField,
-			)
-		}
-	} else {
-		typeIDField := encoded[encodedCompositeValueTypeIDFieldKey]
-		if typeIDField != nil {
-
-			encodedTypeID, ok := typeIDField.(string)
-			if !ok {
-				return nil, fmt.Errorf(
-					"invalid composite type ID encoding (@ %s): %T",
-					strings.Join(path, "."),
-					typeIDField,
-				)
-			}
-
-			_, qualifiedIdentifier, err = common.DecodeTypeID(encodedTypeID)
-			if err != nil {
-				return nil, fmt.Errorf(
-					"invalid composite type ID (@ %s): %w",
-					strings.Join(path, "."),
-					err,
-				)
-			}
-
-			// Special case: The decoded location might be an address location which has no name
-
-			location = d.inferAddressLocationName(location, qualifiedIdentifier)
-		} else {
-			return nil, fmt.Errorf(
-				"missing composite qualified identifier or type ID (@ %s)",
-				strings.Join(path, "."),
-			)
-		}
+	qualifiedIdentifier, ok := qualifiedIdentifierField.(string)
+	if !ok {
+		return nil, fmt.Errorf(
+			"invalid composite qualified identifier encoding (@ %s): %T",
+			strings.Join(path, "."),
+			qualifiedIdentifierField,
+		)
 	}
 
 	// Kind
@@ -616,7 +584,7 @@ func (d *DecoderV4) decodeComposite(v interface{}, path []string) (*CompositeVal
 	// Fields
 
 	fieldsField := encoded[encodedCompositeValueFieldsFieldKey]
-	encodedFields, ok := fieldsField.(map[interface{}]interface{})
+	encodedFields, ok := fieldsField.(cborArray)
 	if !ok {
 		return nil, fmt.Errorf(
 			"invalid composite fields encoding (@ %s): %T",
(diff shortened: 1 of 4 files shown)


runtime/interpreter/encode.go
```diff
@@ -801,19 +808,34 @@ func (e *Encoder) prepareCompositeValue(
...
+		index++
+	}
 
-	for pair := v.Fields.Oldest(); pair != nil; pair = pair.Next() {
-		fieldName := pair.Key
-		value := pair.Value
+	sort.Strings(fieldNames)
+
+	// Create fields (key and value) array, sorted by field name.
+	fields := make(cborArray, fieldLen*2)
```
turbolent, line 826: I think we can just keep the existing iteration over the ordered map, because that already results in a deterministic encoding. Sorting the fields also works, but is not necessary I think

runtime/interpreter/encode.go
```diff
@@ -801,19 +808,34 @@ func (e *Encoder) prepareCompositeValue(
...
 
 		valuePath := append(path[:], fieldName)
 
 		prepared, err := e.prepare(value, valuePath, deferrals)
 		if err != nil {
 			return nil, err
 		}
-		fields[fieldName] = prepared
+
+		fields[i*2], fields[i*2+1] = fieldName, prepared
```
turbolent, line 838: 👌

runtime/interpreter/decode.go
```diff
@@ -625,37 +593,31 @@ func (d *DecoderV4) decodeComposite(v interface{}, path []string) (*CompositeVal
...
 		}
 
-		fieldNames = append(fieldNames, nameString)
-
-		index++
-	}
-
-	// Decode all fields in lexicographic order
-
-	sort.Strings(fieldNames)
```
turbolent, line 652: It's great we don't need this anymore 👍

turbolent (commented):
Nice! Just one comment regarding the encoding of fields

runtime/interpreter/encode.go
fxamacker: I tried this approach (it's faster and cleaner) but ran into a test error.  It causes this test to fail:

- `TestInterpretCompositeValueFieldEncodingOrder`  in interpreter_test.go.  

I strongly prefer your suggestion and am hoping the test might be obsolete or unneeded.

This is the code block incorporating the faster and cleaner code you suggested:
```go
	fields := make(cborArray, v.Fields.Len()*2)

	i := 0
	for pair := v.Fields.Oldest(); pair != nil; pair = pair.Next() {
		fieldName := pair.Key
		value := pair.Value

		valuePath := append(path[:], fieldName)

		prepared, err := e.prepare(value, valuePath, deferrals)
		if err != nil {
			return nil, err
		}

		fields[i], fields[i+1] = fieldName, prepared
		i += 2
	}

```

I suspect `TestInterpretCompositeValueFieldEncodingOrder` tests if encoding composites with entries of different orderings is deterministic.  

Test expects: 
- CBOR data ["a", 1, "b", 2, "c", 3] by encoding composite with entries `["self.a = 1", "self.b = 2", "self.c = 3"]`.
- CBOR data ["a", 1, "b", 2, "c", 3] by encoding composite with entries `["self.b = 2", "self.c = 3", "self.a = 1", ]`. 

Without sorting, this implementation produces:
- CBOR data ["a", 1, "b", 2, "c", 3] by encoding composite with entries `["self.a = 1", "self.b = 2", "self.c = 3"]`.
- CBOR data ["b", 2, "c", 3, "a", 1] by encoding composite with entries `["self.b = 2", "self.c = 3", "self.a = 1", ]`.

Any thoughts or suggestions?

turbolent: This test checks that no matter what the order of initialization is, the encoding is always the same. That's not required though, we only have to ensure the encoding order is always deterministic, and the ordered map for the fields ensures that.

I've adjusted the test case in f2a268b to prepare the expected encodings for all permutations of initializing the fields

turbolent (approved):
I've switched sorting to iterating over the ordered map of fields in f2a268b33c315ef2fad849207428bd55aedd2302.

I enabled CI for feature branches, including from forks, CI ran, and everything is green 👏 

⛵

PR #762, opened 2021-04-05
Add validation for enum cases during contract updates

### LINKED ISSUE #757: Ensure contract update validator handles enum cases correctly
## Definition of Done

- Ensure enum cases are handled correctly by the contract update validator
  - What are we storing for enum cases?
    - If we only store the raw value:
      - [ ] Forbid name changes (either change or removal which leads to name change)
      - [ ] Tests
    - If we store the name, then there is no chance for type/value cofusion

SupunS:
Closes #757

## Description

As in #757, the enum-case values are represented using a composite value, having a field called `rawValue`  which stores the enum-case raw value. Hence serializing/deserializing, only the raw value gets serialized/deserialized, but not the actual enum-case name. So changing enum-cases can lead to type/value confusions.

This PR adds a validation to avoid such scenarios. With these changes, update to an enum need to make sure that:

1. Order of the updated enum-cases must be the same as the existing enum.
2. Updated enum must have at-least the same number of enum-cases as the existing one.
 
 
The above restrictions would mean:
  - Adding enum-cases at the end is allowed
  - Adding enum cases to middle/top is not allowed (violates 1)
  - Swapping enum-cases is not allowed (violates 1)
  - Removing new enum cases is not allowed (violates 2)

______

For contributor use:

### DIFF
--- a/runtime/contract_update_validation.go
+++ b/runtime/contract_update_validation.go
@@ -193,6 +193,7 @@ func (validator *ContractUpdateValidator) checkNestedDeclarations(
 		return nil, false
 	}
 
+	// Check nested structs, enums, etc.
 	newNestedCompositeDecls := newDeclaration.DeclarationMembers().Composites()
 	for _, newNestedDecl := range newNestedCompositeDecls {
 		oldNestedDecl, found := getOldCompositeOrInterfaceDecl(newNestedDecl.Identifier.Identifier)
@@ -204,6 +205,7 @@ func (validator *ContractUpdateValidator) checkNestedDeclarations(
 		validator.checkDeclarationUpdatability(oldNestedDecl, newNestedDecl)
 	}
 
+	// Check nested interfaces.
 	newNestedInterfaces := newDeclaration.DeclarationMembers().Interfaces()
 	for _, newNestedDecl := range newNestedInterfaces {
 		oldNestedDecl, found := getOldCompositeOrInterfaceDecl(newNestedDecl.Identifier.Identifier)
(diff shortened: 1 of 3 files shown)


runtime/contract_update_validation.go
```diff
@@ -214,6 +216,43 @@ func (validator *ContractUpdateValidator) checkNestedDeclarations(
 
 		validator.checkDeclarationUpdatability(oldNestedDecl, newNestedDecl)
 	}
+
+	// Check enum-cases, if theres any.
+	validator.checkEnumCases(oldDeclaration, newDeclaration)
+}
+
+// checkEnumCases validates updating enum cases. Updated enum must:
+//   - Have at-most the same number of enum-cases as the old enum (Removing is allowed, but no additions).
```
turbolent, line 225: I think it's actually the other way round: 

Adding new cases is OK, because there are no stored values with the associated raw value. Accessing the new case on the enum type (e.g. `E.newCase`) won't be invalid like adding a new field to a composite, because the enum type value (`E`) is never stored, and is only created when the enum declaration is interpreted.

Removing a case is then actually not valid, because this still lead to confusion in a two-step process: First, remove a case in the first update, then add a new case in another update.

turbolent (commented):
Thank you for adding this @SupunS!

runtime/contract_update_validation.go
SupunS: oh yeah! we could do the other way around, that would also be less painful to the users.

SupunS: added in https://github.com/onflow/cadence/pull/762/commits/d056627a32781deb439492df48a9c8df3ff6003c

runtime/errors.go
```diff
@@ -386,3 +386,21 @@ func (e *EnumCaseMismatchError) Error() string {
 		e.foundName,
 	)
 }
+
+// MissingEnumCasesError is reported during an enum update, if any enum cases are removed
+// from an existing enum.
+type MissingEnumCasesError struct {
+	declName string
+	expected int
+	found    int
```
turbolent, line 395: All other errors export the fields so maybe do the same here

runtime/contract_update_validation_test.go
```diff
@@ -1404,6 +1504,38 @@ func assertConformanceMismatchError(
...
+	enumMismatchError := err.(*EnumCaseMismatchError)
+
+	assert.Equal(
+		t,
+		fmt.Sprintf(
+			"mismatching enum case: expected `%s`, found `%s`",
+			expectedEnumCase,
+			foundEnumCase,
+		),
+		enumMismatchError.Error(),
```
turbolent, line 1519: Maybe instead of comparing error messages literally, just check declName, expectedEnumCase, and foundEnumCase of the error directly, so changing the error message doesn't break the test

runtime/contract_update_validation_test.go
```diff
@@ -1404,6 +1504,38 @@ func assertConformanceMismatchError(
...
+	assert.Equal(
+		t,
+		fmt.Sprintf(
+			"missing cases in enum `%s`: expected %d or more, found %d",
+			declName,
+			expectedCaes,
+			foundCases,
+		),
+		extraFieldError.Error(),
+	)
```
turbolent, line 1536: Maybe instead of comparing error messages literally, just check declName, expectedCases, and foundCases of the error directly, so changing the error message doesn't break the test

turbolent, line 1536:
```suggestion
func assertMissingEnumCasesError(t *testing.T, err error, declName string, expectedCases int, foundCases int) {
	require.Error(t, err)
	require.IsType(t, &MissingEnumCasesError{}, err)
	extraFieldError := err.(*MissingEnumCasesError)
	assert.Equal(
		t,
		fmt.Sprintf(
			"missing cases in enum `%s`: expected %d or more, found %d",
			declName,
			expectedCases,
			foundCases,
		),
		extraFieldError.Error(),
	)
```

runtime/contract_update_validation.go
```diff
@@ -214,6 +216,51 @@ func (validator *ContractUpdateValidator) checkNestedDeclarations(
...
+			found:    newEnumCaseCount,
+			Range:    ast.NewRangeFromPositioned(newDeclaration.DeclarationIdentifier()),
+		})
+
+		// If some enum cases are removed, trying to match each enum case
+		// may result in too many regression errors. hence return.
+		return
+	}
+
+	// Validate the the new enum cases.
```
turbolent, line 247:
```suggestion
	// Validate the enum cases match the old enum cases.
```

turbolent (approved):
Looks good! just a couple minor suggestions

runtime/contract_update_validation.go
turbolent: @SupunS Wow, so fast! 👏

runtime/errors.go
SupunS: Added in 51bb585

runtime/contract_update_validation_test.go
SupunS: Also wanted to assert the error message, to make sure they are consistent, and the to-string works with nested errors, etc.

runtime/contract_update_validation.go
SupunS: added in 51bb585

runtime/contract_update_validation_test.go
SupunS: added in 51bb585

PR #770, opened 2021-04-06
Get resource owner from host environment

### LINKED ISSUE #754: Owner is not fully initialized for composite values
### Problem

In `OwnerValue()` method of `CompositeValue` owner is not properly initialized. Only the address field is set. This may cause programs to fail at the runtime, if any of the uninitialized fields/methods are accessed.

https://github.com/onflow/cadence/blob/3bbc8c8a1a85f729a6a7ae1537769f6cb8de4964/runtime/interpreter/value.go#L5572

### Acceptance Criteria

- [ ] Properly initialized all the members of the `PublicAccountValue`
- [ ] Add tests

SupunS:
Closes #754

## Description

Currently, when the owner field of a resource is accessed, the interpreter creates a new `PublicAccount` using the owner address. However, since there's no access to the host environment at that point, some of the fields and member functions are not properly initialized.

### Solution:
Load the owner account from the host environment using the address.

______

For contributor use:

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -5619,7 +5619,7 @@ func (v *CompositeValue) GetMember(interpreter *Interpreter, getLocationRange fu
 	if v.Kind == common.CompositeKindResource &&
 		name == sema.ResourceOwnerFieldName {
 
-		return v.OwnerValue()
+		return v.OwnerValue(interpreter)
 	}
 
 	value, ok := v.Fields.Get(name)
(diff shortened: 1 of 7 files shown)


runtime/interpreter/value.go
```diff
@@ -5562,31 +5562,33 @@ func (v *CompositeValue) InitializeFunctions(interpreter *Interpreter) {
...
 	}
 
+	// There's no direct access to the host environment at this point.
+	// Therefore, load and invoke the `getAccount` host function.
+	getAccountValue, ok := interpreter.effectivePredeclaredValues["getAccount"]
+	if !ok {
+		panic(errors.NewUnreachableError())
+	}
+
+	getAccountHostFunc := getAccountValue.ValueDeclarationValue().(HostFunctionValue)
```
turbolent, line 5577: This works, but is brittle, as we will break this accidentally if we e.g. change the name or signature in the future.

Also, currently the `getAccount` function is only injected by the runtime, i.e. outside of the interpreter, so we shouldn't introduce an (implicit) cyclic dependency here.

Given that the `owner` field is baked into the language, and both the `sema` and `interpreter` packages are aware of the notion of a resource having an owner account, I think the proper way to fix this is to make `getAccount` a built-in function too, and make the implementation of it an interpreter option, i.e a handler function that returns a public account value for an address (just like the runtime already does in `interpreterRuntime.newGetAccountFunction`).

I can add this if you want to

turbolent (commented):
Thank you for fixing this!

The changes and tests look great, my only concern is how we get the public account value for an address

runtime/interpreter/value.go
SupunS: oh wow, I was thinking all day on a way to eliminate that dependency loop. Only managed to push it to runtime. Didn't know this is even possible! Thank you so much!

Added the interpreter option in https://github.com/onflow/cadence/pull/770/commits/6978995a52f76ea14e98aa79b642f38053c60438

runtime/stdlib/flow.go
```diff
@@ -47,7 +47,7 @@ var accountFunctionType = &sema.FunctionType{
 	),
 }
 
-var getAccountFunctionType = &sema.FunctionType{
+var GetAccountFunctionType = &sema.FunctionType{
```
turbolent, line 50: Looks like this change is not necessary anymore

runtime/runtime_test.go
```diff
@@ -4261,6 +4267,19 @@ func TestInterpretResourceOwnerFieldUseComposite(t *testing.T) {
...
 			loggedMessages = append(loggedMessages, message)
 		},
+		getStorageUsed: func(_ Address) (uint64, error) {
+			var amount uint64 = 0
+
+			for _, data := range storage.storedValues {
+				amount += uint64(len(data))
+			}
+
+			return amount, nil
```
turbolent, line 4277: This is a good mock implementation, but makes the use and assertions below (https://github.com/onflow/cadence/pull/770/files#diff-74ffa81b6938856f695ba112a721503dfc5ff2e852bc65d79d15ee4868f4bb97R4333) make this test dependent on the storage format, the size of the encoded data might change, breaking this test unnecessarily. 

Maybe just return a unique constant value like in `getStorageCapacity` below

runtime/interpreter/value.go
turbolent: Nice, looks great!

runtime/stdlib/flow.go
SupunS: yeah right, missed reverting this. good catch!

PR #772, opened 2021-04-06
Reject indirect incompatible updates to composite declarations

### LINKED ISSUE #771: Reject indirect changes to composite types during contract updates
### Problem

Currently, it is not allowed to do incompatible changes to a composite type (like changing field types, adding new fields, etc). However, it is possible to remove an existing composite type, and add another composite type with the same name, but with a different structure. This has the same effect as doing an incompatible change to the composite type.

### Acceptance Criteria
- [ ] Reject indirect changes to a struct. Restrict removing any existing composite type declarations, but allow adding new ones
- [ ] Tests

SupunS:
Closes #771

## Description

Currently, doing incompatible changes to a composite type (like changing field types, adding new fields, etc) is not allowed. However, it is possible to remove an existing composite type, and add another composite type with the same name, but with a different structure, which has the same effect as doing an incompatible change.

This PR rejects such indirect changes to a struct. With this,
  - Removing any existing composite type declaration is not allowed
  - Adding new composite type declarations is allowed

______

For contributor use:

### DIFF
--- a/runtime/errors.go
+++ b/runtime/errors.go
@@ -408,13 +408,13 @@ func (e *MissingEnumCasesError) Error() string {
 // MissingCompositeDeclarationError is reported during an contract update, if an existing
 // composite declaration (struct or struct interface) is removed.
 type MissingCompositeDeclarationError struct {
-	DeclName string
+	Name string
 	ast.Range
 }
 
 func (e *MissingCompositeDeclarationError) Error() string {
 	return fmt.Sprintf(
 		"missing composite declaration `%s`",
-		e.DeclName,
+		e.Name,
 	)
 }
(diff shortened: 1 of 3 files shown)


runtime/errors.go
```diff
@@ -404,3 +404,17 @@ func (e *MissingEnumCasesError) Error() string {
...
+// composite declaration (struct or struct interface) is removed.
+type MissingCompositeDeclarationError struct {
+	DeclName string
+	ast.Range
+}
+
+func (e *MissingCompositeDeclarationError) Error() string {
+	return fmt.Sprintf(
+		"missing composite declaration `%s`",
+		e.DeclName,
```
turbolent, line 418:
```suggestion
	Name string
	ast.Range
}

func (e *MissingCompositeDeclarationError) Error() string {
	return fmt.Sprintf(
		"missing composite declaration `%s`",
		e.Name,
```

turbolent (approved):
This is great! Good catch 👍

PR #776, opened 2021-04-07
Improve value type conformance check

turbolent:
⚠️ Targets #724

## Description

Improve the check that an argument value conforms to the dynamic type it claims to be.
Remove `ValueConformsToSemaType` and `conformsToSemaType` and the code duplication it entailed.

Instead, when checking if a composite value conforms to a composite dynamic type (in `CompositeValue.ConformsToDynamicType`), use the same combination of dynamic type check (`interpreter.IsSubType`) and check that the value conforms to the dynamic type it claims to be (`Value.ConformsToDynamicType`) as is already performed by the runtime when validating the arguments (click "Load diff" for `value.go` when following this link):

[`supun/improve-import-export-3...bastian/improve-import-export?expand=1`#diff-424575c293](https://github.com/onflow/cadence/compare/supun/improve-import-export-3...bastian/improve-import-export?expand=1#diff-424575c293dc2bc3a28aab845829d7656fa83a7be4e026de8906935e23bbd52eR5841-R5847)

Also:
- Improve the name of the error types, and fix the error message of `InvalidValueTypeError`.
- Be more conservative in `ConformsToDynamicType` for function values (`interpreter.IsSubType` is also as conservative)
- Fix the test cases (optional values were not boxed correctly, this is now detected)
  - https://github.com/onflow/cadence/pull/776/files?diff=unified&w=1#diff-863660c79fae3d9b381c2dff3025cd88dc46eb62571b754e7695acbfeab4fbfeR1468
  - https://github.com/onflow/cadence/pull/776/files?diff=unified&w=1#diff-863660c79fae3d9b381c2dff3025cd88dc46eb62571b754e7695acbfeab4fbfeR1343

______

For contributor use:

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -50,11 +50,7 @@ type Value interface {
 	IsModified() bool
 	SetModified(modified bool)
 	StaticType() StaticType
-	ConformsToDynamicType(dynamicType DynamicType) bool
-
-	// This method is for internal use only.
-	// Use `ValueConformsToSemaType` to check the type conformance of a value.
-	conformsToSemaType(semaType sema.Type) bool
+	ConformsToDynamicType(interpreter *Interpreter, dynamicType DynamicType) bool
 }
 
 // ValueIndexableValue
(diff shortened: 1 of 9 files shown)


runtime/interpreter/value.go
```diff
@@ -1133,13 +1089,9 @@ func (v IntValue) ToBigEndianBytes() []byte {
...
 
-func (v IntValue) ConformsToDynamicType(dynamicType DynamicType) bool {
+func (v IntValue) ConformsToDynamicType(_ *Interpreter, dynamicType DynamicType) bool {
 	numberType, ok := dynamicType.(NumberDynamicType)
-	return ok && ValueConformsToSemaType(v, numberType.StaticType)
-}
-
-func (v IntValue) conformsToSemaType(semaType sema.Type) bool {
-	return (sema.IntType).Equal(semaType)
+	return ok && numberType.StaticType == sema.IntType
```
SupunS, line 1094: Maybe we could delegate the static type check to `equal` method
```suggestion
	return ok && sema.IntType.equal(numberType.StaticType)
```

I can add this change.

SupunS (approved):
Looks great! Thank you for adding this!

I'll merge this and update the PR.

PR #783, opened 2021-04-08
Benchmark CBOR encoding/decoding using mainnet data

### LINKED ISSUE #765: Benchmark CBOR encoding/decoding using production data
### Issue To Be Solved

Benchmarks for encoding and decoding CBOR can use data that is more similar to production data.

### Suggestion

Using Cadence-specific production data (or research data) can be useful for optimizing and detecting performance regressions.

It can include a variety of data, for example:
- smallest total size
- largest total size
- mean and median total size
- most frequently used mix of types and sizes
- data known to be a bottleneck
- etc.

### Context

These benchmarks would be useful during development, especially optimization.  

@fxamacker  gathered and used a handful of data to benchmark #761 and thinks it would be more useful to have a separate PR to provide benchmarks for PR #752, PR #761, and #764 (Optimize remaining Value types still using CBOR map).

Updates: #738

fxamacker:
Closes #765
Updates #738 

## Description

Create benchmarks that use testdata/*.cbor.gz files.

Add initial 6 *.cbor.gz files using data from a mainnet snapshot.
 - Filenames currently include v3 or v2 to indicate version (v4 is default but can be included in the name).  
 - The 16 character hex portion is a truncated SHA256 of the .cbor.gz file.

For safety:
- limit number of files processed (currently 100).
- use io.LimitReader to limit the max bytes uncompressed from each file (currently 16 MiB).

#### Caveats

The SHA256 isn't verified by the benchmark.  We can do that or we can do it manually using sha256sum command and an extra file if desired.

I chose SHA256 because it's included in Go.
______

For contributor use:

- [x] Targeted PR against `feature/storage-optimizations` branch


turbolent (approved):
Very nice!

PR #784, opened 2021-04-08
Extend defensive run-time checking to all value transfers

turbolent:
Port of dapperlabs/cadence-internal#3

## Description

- Extend the defensive run-time check that argument's dynamic type are a subtype to all value transfers, e.g. assignments. So e.g. when we copy what we believe to be a struct in e.g. an assignment (`=`), we also include a sanity check that we aren’t actually copying a resource.

  From Dete:
  > we should be liberal in including checks like this; we can optimize them out later But we should always focus on safety over performance … until we see performance problems, and then we can optimize the situations that come up while being mindful of safety."

- This detected a few bugs:
  - `SignedNumber` was accidentally a subtype of `SignedFixedPoint`. This was a "copy-and-paste typo" and should be the type itself, `SignedFixedPoint`
  - Some tests / host functions returned values of other types
  - Getting the dynamic type of an ephemeral reference didn't handle potential recursive reference

______

### DIFF
--- a/runtime/tests/interpreter/builtinfunctions_test.go
+++ b/runtime/tests/interpreter/builtinfunctions_test.go
@@ -70,7 +70,7 @@ func TestInterpretToString(t *testing.T) {
 
 			var literal string
 
-			isSigned := sema.IsSubType(ty, &sema.SignedFixedPointType{})
+			isSigned := sema.IsSubType(ty, sema.SignedFixedPointType)
 
 			if isSigned {
 				literal = "-12.34"
--- a/runtime/tests/interpreter/builtinfunctions_test.go
+++ b/runtime/tests/interpreter/builtinfunctions_test.go
@@ -69,13 +69,16 @@ func TestInterpretToString(t *testing.T) {
 		t.Run(ty.String(), func(t *testing.T) {
 
 			var literal string
-
+ 			var expected interpreter.Value
+ 			
 			isSigned := sema.IsSubType(ty, sema.SignedFixedPointType)
 
 			if isSigned {
 				literal = "-12.34"
+				expected = interpreter.NewStringValue("-12.34000000")
 			} else {
 				literal = "12.34"
+				expected = interpreter.NewStringValue("12.34000000")
 			}
 
 			inter := parseCheckAndInterpret(t,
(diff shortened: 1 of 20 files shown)


runtime/tests/interpreter/builtinfunctions_test.go
```diff
@@ -68,18 +68,36 @@ func TestInterpretToString(t *testing.T) {
...
+					literal,
 				),
 			)
 
+			var expected interpreter.Value
+			if isSigned {
+				expected = interpreter.NewStringValue("-12.34000000")
+			} else {
+				expected = interpreter.NewStringValue("12.34000000")
+			}
```
SupunS, line 97: Just a minor thing, we could merge this if-condition with the if-condition above (in L#75).

SupunS (approved):

runtime/tests/interpreter/builtinfunctions_test.go
turbolent: Good idea, I hand't noticed 👍

runtime/tests/interpreter/builtinfunctions_test.go
```diff
@@ -68,18 +68,36 @@ func TestInterpretToString(t *testing.T) {
...
 
+			var literal string
+
+			isSigned := sema.IsSubType(ty, sema.SignedFixedPointType)
+
+			if isSigned {
+				literal = "-12.34"
+			} else {
+				literal = "12.34"
+			}
```
turbolent, line 79:
```suggestion
			var literal string
 			var expected interpreter.Value
 			
			isSigned := sema.IsSubType(ty, sema.SignedFixedPointType)

			if isSigned {
				literal = "-12.34"
				expected = interpreter.NewStringValue("-12.34000000")
			} else {
				literal = "12.34"
				expected = interpreter.NewStringValue("12.34000000")
			}
```

runtime/tests/interpreter/builtinfunctions_test.go
```diff
@@ -68,18 +68,36 @@ func TestInterpretToString(t *testing.T) {
...
 				),
 			)
 
+			var expected interpreter.Value
+			if isSigned {
+				expected = interpreter.NewStringValue("-12.34000000")
+			} else {
+				expected = interpreter.NewStringValue("12.34000000")
+			}
+
```
turbolent, line 98:
```suggestion
```

runtime/tests/interpreter/builtinfunctions_test.go
turbolent: Refactored in 7463495

PR #787, opened 2021-04-09
Handle cyclic references in dynamic type conformance check

### LINKED ISSUE #777: Handle cycles in dynamic type checking for ephemeral references
### Issue To Be Solved

Ephemeral values can have potential self-references, resulting in cyclic graphs. This should be properly handled in `ConformsToDynamicType` of `EphemeralReferenceValue` to avoid getting stuck in an infinite recursion.

More context here: https://github.com/onflow/cadence/pull/724#discussion_r607249381

### Acceptance Criteria
- [ ] Maintain a result set that holds already-checked/currently-checking value.
- [ ] Check the result set before checking a new value
  - [ ]  Return if the value is already there in the results
  - [ ] Continue to check, otherwise
- [ ] Test cases

SupunS:
Closes #777

## Description

Ephemeral values can have cyclic references. e.g:
```
pub fun main() {
    var foo = Foo()
    var foo_ref = &foo as &Foo

    // Creates cyclic reference
     foo_ref.bar = foo_ref
}

pub struct Foo {

    pub(set) var bar: &Foo?

    init() {
        self.bar = nil
    }
}
```
This could cause the `ConformsToDynamicType` check to get stuck in an infinite recursion.

Even though  it is not possible to pass reference values as arguments to programs at the moment (and `ConformsToDynamicType` is only used in that path for now), this could be a potential problem in the future.

______

For contributor use:

### DIFF
--- a/runtime/interpreter/block.go
+++ b/runtime/interpreter/block.go
@@ -110,7 +110,7 @@ func (v BlockValue) String() string {
 	)
 }
 
-func (v BlockValue) ConformsToDynamicType(_ *Interpreter, dynamicType DynamicType) bool {
+func (v BlockValue) ConformsToDynamicType(_ *Interpreter, dynamicType DynamicType, _ TypeConformanceResults) bool {
 	_, ok := dynamicType.(BlockDynamicType)
 	return ok
 }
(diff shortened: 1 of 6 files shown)


turbolent (approved):
Nice, looks good!

PR #792, opened 2021-04-12
Deny removing contracts if contract update validation is enabled

### LINKED ISSUE #791: Removing contracts can lead to storage inconsistencies
### Issue To Be Solved

Consider a situation where a contract `Foo` with a struct declaration is deployed.
```cadence
pub contract Foo {

    pub struct SomeStruct {
        pub let a: Int

        init() {
            self.a = 123
        }
    }
}
```

Another contract `Bar` has a field of type `Foo.SomeStruct`
```cadence
import Foo from 0x42

pub contract Bar {
    pub let a: Foo.SomeStruct

    init() {
      self.x = Foo.SomeStruct()
    }
}
```

Then it is possible to remove the `Foo` contract and add a new contract with the same name, but the changes the field type.

```cadence
pub contract Foo {

    pub struct SomeStruct {
        pub let a: Struct

        init() {
            self.a = "hello"
        }
    }
}
```

But, the `Bar` contract still uses the `Foo.SomeStruct`, and the structure/member-types of the `SomeStruct` is different from what it was. This could lead to issues when decoding the already stored data of the `Bar` contract.

### Suggest A Solution
Deny removing already deployed contracts, similar to the validation restriction imposed for other composite types (i.e: https://github.com/onflow/cadence/issues/771).

SupunS:
Closes #791

## Description

$subject to avoid inconsistencies at the storage when decoding already stored data. 

______

For contributor use:

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -2166,8 +2166,17 @@ func (r *interpreterRuntime) newAuthAccountContractsRemoveFunction(
 				// NOTE: *DO NOT* call SetProgram – the program removal
 				// should not be effective during the execution, only after
 
-				// Deny removing a contract, if the contract validation is enabled.
-				if r.contractUpdateValidationEnabled {
+				// Deny removing a contract, if the contract validation is enabled, and
+				// the existing code contains enums.
+				existingProgram, err := parser2.ParseProgram(string(code))
+				if err != nil {
+					panic(&ContractRemovalError{
+						Name:          nameArgument,
+						LocationRange: invocation.GetLocationRange(),
+					})
+				}
+
+				if r.contractUpdateValidationEnabled && containsEnumsInProgram(existingProgram) {
 					panic(&ContractRemovalError{
 						Name:          nameArgument,
 						LocationRange: invocation.GetLocationRange(),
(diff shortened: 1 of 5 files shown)


turbolent:
@SupunS Good thinking here, but I think we should discuss this: As far as I remember, in our discussion regarding the remaining checks for contract updates we determined that this was not needed, as just like in other languages / environments, we cannot and should not consider the effect on downstream dependencies, but only on the effects on the stored data itself.

turbolent:
@SupunS Given our discussion, for now, maybe this PR could be modified to only reject contract removals if there are no enum declarations in the contract?

SupunS:
@turbolent Updated the PR with the discussed approach. Removing a contract is now rejected only if there are enum declarations inside them.

runtime/runtime.go
```diff
@@ -2166,6 +2166,22 @@ func (r *interpreterRuntime) newAuthAccountContractsRemoveFunction(
...
 
+				// Deny removing a contract, if the contract validation is enabled, and
+				// the existing code contains enums.
+				if r.contractUpdateValidationEnabled {
+
+					existingProgram, err := parser2.ParseProgram(string(code))
+
+					// If the existing code is not parsable (i.e: `err != nil`), that shouldn't be a reason to
+					// fail the contract removal. Therefore, validate only if the code is a valid one.
+					if err == nil && containsEnumsInProgram(existingProgram) {
```
turbolent, line 2177: 👍

turbolent (approved):

PR #793, opened 2021-04-13
Document contract update validation criteria

### LINKED ISSUE #756: Document contract update checker
## Context

Contract updates are now checked and invalid updates are rejected.

Document current rules and design decisions.

## Definition of Done

- [ ] Documented current rules and design decisions
- [ ] Examples for most frequent use-cases

## Additional Info

The contract update checker ensures that:

1. Stored data doesn't change its meaning when a contract is updated
2. Decoding and using stored data does not lead to crashes.
   For example, it is invalid to add a field because existing stored data won't have the new field.
   A static check of the access of the field would be valid, but then a crash would occur in the interpreter when accessing the field, because the field is missing in the decoded value (wrong shape)

Some rules:

- Importing contracts don't matter. If a change breaks an importing contract, that's valid.
   This ensures that changes / fixes can be applied at the root of the dependency chain.

- It is not valid to change a field type to a subtype, because existing stored data might be of a supertype.

- Remove conformance of composite to interface is OK, the stored data only stores concrete type identifier

- Changing the body of a function is valid, as long as the signature stays the same

SupunS:
Closes #756

## Description

Document the rules and the design decisions for the contract update validation.
______

For contributor use:

### DIFF
--- a/docs/contract-updatability.md
+++ b/docs/contract-updatability.md
@@ -1,0 +1,340 @@
+---
+title: Contract Updatability
+---
+
+## Introduction
+A [contract](https://docs.onflow.org/cadence/language/contracts/) in Cadence is a collection of data (its state) and
+code (its functions) that lives in the contract storage area of an account.
+When a contract is updated, it is important to make sure that the changes introduced does not lead to runtime
+inconsistencies for already persisted data.
+Cadence maintains this state consistency by validating the contracts and all their components before an update.
+
+## Validation Goals
+The contract update validation ensures that:
+
+- Stored data doesn't change its meaning when a contract is updated.
+- Decoding and using stored data does not lead to runtime crashes.
+  - For example, it is invalid to add a field because existing stored data won't have the new field.
+  - Loading the existing data will result in garbage/missing values for such fields.
+  - A static check of the access of the field would be valid, but the interpreter would crash when accessing the field,
+    because the field has a missing/garbage value.
+
+## Updating a Contract
+
+A contract may consist of fields and other declarations such as composite types, functions, constructors, etc.
+When an existing contract is updated, all its inner declarations are also validated.
+
+### Contract Fields
+When a contract is deployed, the fields of the contract are stored in an account's contract storage.
+Changing the fields of a contract only changes the way the program treats the data, but does not change the already
+stored data itself, which could potentially result in runtime inconsistencies as mentioned in the previous section.
+
+See the [Fields](#fields) section for the possible updates that can be done to the fields, and the restrictions
+imposed on changing fields of a contract.
+
+### Nested Declarations
+Contracts can have nested composite type declarations such as structs, resources, interfaces, and enums.
+When a contract is updated, its nested declarations are checked, because:
+ - They can be used as type annotation for the fields of the same contract, directly or indirectly.
+ - Any third-party contract can import the types defined in this contract and use them as type annotations.
+ - Hence, changing the type definition is the same as changing the type annotation of such a field (which is also invalid,
+   as in the [Fields](#fields) section).
+
+Changes that can be done to the nested declarations, and the update restrictions are described in following sections:
+ - [Structs, resources and interface](#structs-resources-and-interfaces)
+ - [Enums](#enums)
+ - [Functions](#functions)
+ - [Constructors](#constructors)
+
+## Fields
+A field may belong to a contract, struct, resource, or interface.
+
+#### Valid Changes:
+- Removing a field is valid
+  ```cadence
+  // Existing contract
+
+  pub contract Foo {
+      var a: String
+      var b: Int
+  }
+
+
+  // Updated contract
+
+  pub contract Foo {
+      var a: String
+  }
+  ```
+  - It leaves data for the removed field unused at the storage, as it is no longer accessible.
+  - However, it does not cause any runtime crashes.
+
+- Changing the order of fields is valid.
+  ```cadence
+  // Existing contract
+
+  pub contract Foo {
+      var a: String
+      var b: Int
+  }
+
+
+  // Updated contract
+
+  pub contract Foo {
+      var b: Int
+      var a: String
+  }
+  ```
+
+#### Invalid Changes
+- Adding a new field is not valid.
+  ```cadence
+  // Existing contract
+
+  pub contract Foo {
+      var a: String
+  }
+
+
+  // Updated contract
+
+  pub contract Foo {
+      var a: String
+      var b: Int      // Invalid new field
+  }
+  ```
+    - Existing stored data won't have the new field.
+    - Decoding stored data will result in garbage or missing values for such fields.
+
+- Changing the type of existing field is not valid.
+  ```cadence
+  // Existing contract
+  
+  pub contract Foo {
+      var a: String
+  }
+
+
+  // Updated contract
+
+  pub contract Foo {
+      var a: Int      // Invalid type change
+  }
+  ```
+    - In an already stored contract, the field `a` would have a value of type `String`.
+    - Changing the type of the field `a` to `Int`, would make the runtime read the already stored `String`
+      value as an `Int`, which will result in deserialization errors.
+    - Changing the field type to a subtype/supertype of the existing type is also not valid, as it would also
+      potentially cause issues while decoding/encoding.
+      e.g:
+      - Changing an `Int64` field to `Int8` - Stored value could be number `624` which exceeds the value space
+        for `Int8`.
+
+## Structs, Resources and Interfaces
+
+#### Valid Changes:
+- Adding a new struct, resource, or interface is valid.
+- Adding, removing, or changing interface conformance of a struct/resource is valid, since the stored data only
+  stores concrete type/value, but doesn't store the conformance info.
+  ```cadence
+  // Existing struct
+  
+  pub struct Foo: T {
+  }
+
+
+  // Upated struct
+
+  pub struct Foo: R {
+  }
+  ```
+
+#### Invalid Changes:
+- Removing an existing declaration is not valid.
+  - Removing a declaration allows adding a new declaration with the same name, but with a different structure.
+  - Any program that uses that declaration would face inconsistencies in the stored data.
+- Renaming a declaration is not valid. It can have the same effect as removing an existing declaration and adding
+  a new one.
+- Changing the type of declaration is not valid. i.e: Changing from a struct to interface, and vise versa.
+  ```cadence
+  // Existing struct
+  
+  pub struct Foo {
+  }
+
+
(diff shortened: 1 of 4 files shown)


docs/contract-updatability.md
```diff
@@ -0,0 +1,340 @@
...
+  }
+  ```
+  - Raw value of an enum is implicit, and corresponds to the defined order.
+  - Changing the order of enum-cases has the same effect as changing the raw-value, which could cause the storage
+    inconsistencies and type-confusions as described earlier.
+
+## Functions
+Updating a function definition is always valid, as function definitions are never stored as data. 
+i.e: Function definition is a part of the code, but not data.
+- Changing a function signature (parameters, return types) is valid.
```
SupunS, line 327: Changing function signature is allowed at the moment. So I documented the current behavior.

docs/contract-updatability.md
```diff
@@ -0,0 +1,340 @@
...
+
+  // Updated enum
+
+  pub enum Color: Int {
+    pub case RED
+
+    // invalid removal of `case BLUE`
+  }
+  ```
+- Changing the order or enum-cases is permitted
```
bluesign, line 303: this should be "not permitted" right?

docs/contract-updatability.md
```diff
@@ -0,0 +1,340 @@
...
+When a contract is deployed, the fields of the contract are stored in an account's contract storage.
+Changing the fields of a contract only changes the way the program treats the data, but does not change the already
+stored data itself, which could potentially result in runtime inconsistencies as mentioned in the previous section.
+
+See the [Fields](#fields) section for the possible updates that can be done to the fields, and the restrictions
+imposed on changing fields of a contract.
+
+### Nested Declarations
+Contracts can have nested composite type declarations such as structs, resources, interfaces, and enums.
+When a contract is updated, its nested declarations are also being checked, because:
```
joshuahannan, line 37:
```suggestion
When a contract is updated, its nested declarations are checked, because:
```

docs/contract-updatability.md
bluesign: when you say signature, it also covers access modifiers? 

can I change access(all) to access(self) ?

docs/contract-updatability.md
```diff
@@ -0,0 +1,340 @@
...
+  ```
+    - In an already stored contract, the field `a` would have a value of type `String`.
+    - Changing the type of the field `a` to `Int`, would make the runtime read the already stored `String`
+      value as an `Int`, which will result in deserialization errors.
+    - Changing the field type to a subtype/supertype of the existing type is also not valid, as it would also
+      potentially cause issues while decoding/encoding.
+      e.g:
+      - Changing an `Int64` field to `Int8` - Stored value could be number `624` which exceeds the value space
+        for `Int8`.
+
```
joshuahannan, line 133: It would be good to add a section about changing the access modifier for a field

docs/contract-updatability.md
```diff
@@ -0,0 +1,340 @@
...
+  ```cadence
+  // Existing struct
+  
+  pub struct Foo {
+  }
+
+
+  // Changed to a struct interface
+
+  pub struct interface Foo {
```
joshuahannan, line 169:
```suggestion
  pub struct interface Foo {  // Invalid type declaration change
```

docs/contract-updatability.md
```diff
@@ -0,0 +1,340 @@
...
+
+  pub enum Color: Int {
+    pub case RED
+    pub case BLUE
+  }
+
+
+  // Updated enum with `UInt8` raw type
+
+  pub enum Color: UInt8 {
```
joshuahannan, line 208:
```suggestion
  pub enum Color: UInt8 {  // Invalid change of raw type
```

joshuahannan (approved):
Looks good!

docs/contract-updatability.md
SupunS: oh yes, good catch!

docs/contract-updatability.md
SupunS: That's a good idea! will ad it too.

SupunS: Added in 0e08133

docs/contract-updatability.md
SupunS: Yes. Added a note in 0e08133

docs/contract-updatability.md
```diff
@@ -0,0 +1,357 @@
+---
+title: Contract Updatability
+---
+
+## Introduction
+A [contract](https://docs.onflow.org/cadence/language/contracts/) in Cadence is a collection of data (its state) and
```
turbolent, line 6:
```suggestion
A [contract](../contracts) in Cadence is a collection of data (its state) and
```

docs/contract-updatability.md
```diff
@@ -0,0 +1,357 @@
...
+code (its functions) that lives in the contract storage area of an account.
+When a contract is updated, it is important to make sure that the changes introduced does not lead to runtime
```
turbolent, line 8:
```suggestion
When a contract is updated, it is important to make sure that the changes introduced do not lead to runtime
```

docs/contract-updatability.md
```diff
@@ -0,0 +1,357 @@
...
+inconsistencies for already persisted data.
```
turbolent, line 9: We normally use the terms "storage" and "stored", instead of "persistence" and "persisted"

```suggestion
inconsistencies for already stored data.
```

docs/contract-updatability.md
```diff
@@ -0,0 +1,357 @@
...
+
+A contract may consist of fields and other declarations such as composite types, functions, constructors, etc.
+When an existing contract is updated, all its inner declarations are also validated.
+
+### Contract Fields
+When a contract is deployed, the fields of the contract are stored in an account's contract storage.
+Changing the fields of a contract only changes the way the program treats the data, but does not change the already
+stored data itself, which could potentially result in runtime inconsistencies as mentioned in the previous section.
+
+See the [Fields](#fields) section for the possible updates that can be done to the fields, and the restrictions
```
turbolent, line 32:
```suggestion
See the [section about fields below](#fields) for the possible updates that can be done to the fields, and the restrictions
```

docs/contract-updatability.md
```diff
@@ -0,0 +1,357 @@
...
+imposed on changing fields of a contract.
+
+### Nested Declarations
+Contracts can have nested composite type declarations such as structs, resources, interfaces, and enums.
+When a contract is updated, its nested declarations are checked, because:
+ - They can be used as type annotation for the fields of the same contract, directly or indirectly.
+ - Any third-party contract can import the types defined in this contract and use them as type annotations.
+ - Hence, changing the type definition is the same as changing the type annotation of such a field (which is also invalid,
+   as in the [Fields](#fields) section).
```
turbolent, line 41:
```suggestion
   as described in the [section about fields fields](#fields) below).
```

docs/contract-updatability.md
```diff
@@ -0,0 +1,357 @@
...
+
+  // Updated contract
+
+  pub contract Foo {
+      pub var b: Int
+      pub var a: String
+  }
+  ```
+
+- Changing the access modifier of a field is valid.
```
turbolent, line 90: Maybe add a note below explaining that the validation doesn't ensure that other code depending on this code will stay valid, only that the code and the data it stores itself stays valid.

docs/contract-updatability.md
```diff
@@ -0,0 +1,357 @@
...
+
+
+  // Updated contract
+
+  pub contract Foo {
+      pub var a: String
+      pub var b: Int      // Invalid new field
+  }
+  ```
+    - Existing stored data won't have the new field.
```
turbolent, line 123: Maybe mention here that the initializer of a contract is only run once, when the contract is deployed, and it is not rerun when the contract is updated, so any field initialization there would not be executed.

Maybe also mention that this current restriction may be remedied in a future version by providing means to initialize added fields.

docs/contract-updatability.md
```diff
@@ -0,0 +1,357 @@
...
+  ```
+    - In an already stored contract, the field `a` would have a value of type `String`.
+    - Changing the type of the field `a` to `Int`, would make the runtime read the already stored `String`
+      value as an `Int`, which will result in deserialization errors.
+    - Changing the field type to a subtype/supertype of the existing type is also not valid, as it would also
+      potentially cause issues while decoding/encoding.
+      e.g:
+      - Changing an `Int64` field to `Int8` - Stored value could be number `624` which exceeds the value space
+        for `Int8`.
+
```
turbolent, line 149: Maybe mention here that this current limitation may be remedied in a future version by providing means to migrate existing fields.

docs/contract-updatability.md
```diff
@@ -0,0 +1,357 @@
...
+    inconsistencies and type-confusions as described earlier.
+
+## Functions
+Updating a function definition is always valid, as function definitions are never stored as data. 
+i.e: Function definition is a part of the code, but not data.
+- Changing a function signature (parameters, return types) is valid.
+- Changing a function body is also valid.
+- Changing the access modifier is valid.
+
+However, changing a `function type` may or may not be valid, depending on where it is used.
```
turbolent, line 347:
```suggestion
However, changing a *function type* may or may not be valid, depending on where it is used.
```

docs/language/contracts.md
```diff
@@ -306,10 +306,9 @@ signer.contracts.add(
 
 > 🚧 Status: Updating contracts is **experimental**.
 >
-> This function only works for updates of function bodies, conditions, access modifiers, and comments.
->
-> This function does **not** support other changes, e.g. changes of function signatures (parameters, return type);
-> adding, removing, or updating the name or type of fields; updating type declarations, etc.
+> Updating contracts currently have limited support to maintain data consistency.
+> Read more details on possible changes and restrictions imposed on updating contracts in the
+> [contract updatability](../contract-updatability.md) section.
```
turbolent, line 311:
```suggestion
> Updating contracts is currently limited to maintain data consistency.
> Read more details on valid changes and restrictions imposed on updating contracts in the
> [contract updatability](../contract-updatability) section.
```

turbolent (approved):
Nice write-up! 👏 

Just a couple minor suggestions and a few link fixes, then it should be ready to be merged

docs/contract-updatability.md
SupunS: That's a really good idea! 
We could actually add it in general. e.g: under the "goals" section it currently says what are the goals are, but it doesn't say what is **not a goal**.

SupunS:
Thanks for the reviews!  Updated the PR with suggestions.

turbolent:
Looks good! Feel free to merge

PR #798, opened 2021-04-13
Improve state decoding tool

### LINKED ISSUE #780: Ensure Mainnet state is migratable
### Issue To Be Solved

While optimizing the storage format #738, make sure we keep being able to migrate all of the Mainnet state once we are done with the format changes.

### Suggested Solution

Add functionality to the state decoding tool to use the old decoder (v3), then encode and decode using the new encoder and new decoder (v4), and finally compare the final value to the initially decoded value.

This requires storable values to be comparable.

### Definition of Done

- [ ] Extend state decoding tool with 
   - [ ] Support for decoding using pre-v4 format
   - [ ] Support for re-encoding and decoding initially decoded data
   - [ ] Comparsion of initially decoded value and values that was re-encoded and decoded
- [ ] Implement equality for storable values

turbolent:
Closes #780 

## Description

Add two new flags:
- `old`: 
  Decode the data using the old decoder (v3)
- `roundtrip`: 
  After decoding, encode the data using the new encoder (v4), and decode it again. 
  Ensure the equality of the originally decoded value, and the re-decoded value

______


runtime/cmd/decode-state-values/main.go
```diff
@@ -56,6 +60,11 @@ func worker(jobs <-chan entry, wg *sync.WaitGroup, decoded *uint64) {
 	var err error
 	var data []byte
 
+	decodeFunction := interpreter.DecodeValue
+	if *oldFlag {
+		decodeFunction = interpreter.DecodeValueV3
+	}
```
fxamacker, line 66: Just an idea for consideration (not a proposal) -- what do you think about using version returned by `StripMagic` inside the loop that follows this?  The version returned can be used to pick the decode function.

turbolent: Good idea! I replaced it in 390c477898e9b4b295a1e830c93e0266b0c60c71

turbolent:
Also added a gzip flag for onflow/flow-go#627

fxamacker (approved):
Looks good!

turbolent:
Thank you for having a look @fxamacker!

PR #802, opened 2021-04-14
Simplify function types

turbolent:
## Description

- Remove `CheckedFunctionType` wrapper, move functionality to `FunctionType`
- Rename `SpecialFunctionTypes` to the more descriptive `ConstructorFunctionType`
- Move the functionality of special function types to have members to `FunctionType`

______

### DIFF
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -1913,6 +1913,7 @@ type FunctionType struct {
 	ReturnTypeAnnotation     *TypeAnnotation
 	RequiredArgumentCount    *int
 	ArgumentExpressionsCheck ArgumentExpressionsCheck
+	Members                  *StringMemberOrderedMap
 }
 
 func RequiredArgumentCount(count int) *int {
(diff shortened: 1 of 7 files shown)


runtime/sema/type.go
```diff
@@ -2332,6 +2329,12 @@ func (t *SpecialFunctionType) GetMembers() map[string]MemberResolver {
 	return withBuiltinMembers(t, members)
 }
 
+// ConstructorFunctionType is the the type representing a constructor function
+
+type ConstructorFunctionType struct {
+	*FunctionType
+}
```
SupunS, line 2336: Could we use a type alias?

SupunS (approved):

runtime/sema/type.go
turbolent: The separate type is used to check if a function type is a constructor. 

If it's just an alias (`type ConstructorFunctionType = FunctionType`), then there is no way to distinguish the two.

If it's a type definition (`type ConstructorFunctionType FunctionType`), then that type is not implementing `Type`.

An alternative would be to e.g. just have a boolean field `IsConstructor` on `FunctionType`

PR #822, opened 2021-04-21
Rename signature algorithm ECDSA_Secp256k1 to ECDSA_secp256k1

### LINKED ISSUE #760: Inconsistent SigningAlgorithm identifiers between flow-go and cadence repositories.
### Instructions

### Problem

On the currently running testnet, when creating a PublicKey that uses `secp256k1` I get an `unkown error` with a pointer to the algorithm identifier `SignatureAlgorithm.ECDSA_secp256k1`.

```
let publicKey = PublicKey(
    publicKey: "1234".decodeHex(),
    signatureAlgorithm: SignatureAlgorithm.ECDSA_secp256k1
)
```

### Steps to Reproduce

Send this transaction with a valid secp256k1 public key to testnet.

```
transaction(publicKey: String) {
    let account = AuthAccount(payer: signer);
    let key = PublicKey(
      publicKey: publicKey.decodeHex(),
      signatureAlgorithm: SignatureAlgorithm.ECDSA_secp256k1
    );
    account.keys.add(
      publicKey: key,
      hashAlgorithm: HashAlgorithm.SHA3_256,
      weight: 1.0
    );
  }
}
```

### Acceptance Criteria

Not sure.

### Context

In the `flow-go` repository (flow-go/crypto/types.go) you have the following string identifiers.

```go
func (f SigningAlgorithm) String() string {
	return [...]string{"UNKNOWN", "BLS_BLS12381", "ECDSA_P256", "ECDSA_secp256k1"}[f]
}
```

My assumption, when building the transaction initially, was that these identifiers will be understood by Cadence. But for some reason Cadence only accepts `SignatureAlgorithm.ECDSA_Secp256k1`, emphasis on the uppercase **S**.

SupunS:
Closes #760

## Description

Work towards #582

______

For contributor use:

### DIFF
--- a/runtime/types.go
+++ b/runtime/types.go
@@ -48,7 +48,7 @@ const (
 	// Supported signing algorithms
 	SignatureAlgorithmUnknown         = sema.SignatureAlgorithmUnknown
 	SignatureAlgorithmECDSA_P256      = sema.SignatureAlgorithmECDSA_P256
-	SignatureAlgorithmECDSA_Secp256k1 = sema.SignatureAlgorithmECDSA_Secp256k1
+	SignatureAlgorithmECDSA_secp256k1 = sema.SignatureAlgorithmECDSA_secp256k1
 )
 
 type HashAlgorithm = sema.HashAlgorithm
--- a/docs/language/crypto.md
+++ b/docs/language/crypto.md
@@ -35,8 +35,8 @@ pub enum SignatureAlgorithm: UInt8 {
     /// ECDSA_P256 is Elliptic Curve Digital Signature Algorithm (ECDSA) on the NIST P-256 curve.
     pub case ECDSA_P256 = 1
 
-    /// ECDSA_Secp256k1 is Elliptic Curve Digital Signature Algorithm (ECDSA) on the secp256k1 curve.
-    pub case ECDSA_Secp256k1 = 2
+    /// ECDSA_secp256k1 is Elliptic Curve Digital Signature Algorithm (ECDSA) on the secp256k1 curve.
+    pub case ECDSA_secp256k1 = 2
 }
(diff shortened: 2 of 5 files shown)


runtime/types.go
```diff
@@ -48,7 +48,7 @@ const (
 	// Supported signing algorithms
 	SignatureAlgorithmUnknown         = sema.SignatureAlgorithmUnknown
 	SignatureAlgorithmECDSA_P256      = sema.SignatureAlgorithmECDSA_P256
-	SignatureAlgorithmECDSA_Secp256k1 = sema.SignatureAlgorithmECDSA_Secp256k1
+	SignatureAlgorithmECDSA_Secp256k1 = sema.SignatureAlgorithmECDSA_secp256k1
```
SupunS, line 51: I didn't change the interface type (this constant) so that this change won't affect FVM.

tarakby (approved):
👏

runtime/types.go
turbolent: We should change it everywhere to be consistent here, it's also upper-case on the FVM side

turbolent (commented):
Nice! 

This is a breaking change, we need to make sure that contracts on Testnet and Mainnet will be upgrable / upgraded

SupunS:
That's a good point on updatability! 
Technically it should be OK since `SignatureAlgorithm.ECDSA_secp256k1` can never be used as a type (can only use as a value), so no risk of anyone using it as a field of a contract. 

At the storage level, only the raw value along with the type name `SignatureAlgorithm` is stored, and the case-name `ECDSA_secp256k1` is never stored. So that should be fine as well.

But yeah, we need to make sure they are properly updated.

PR #823, opened 2021-04-21
Include wrapped error in InvalidEntryPointArgumentError message

mikeylemmon:
While trying to debug issues passing struct args to scripts and transactions using fcl (documented here: https://github.com/onflow/flow-js-sdk/issues/607), I discovered that the root cause of invalid arguments has been obscured and difficult to debug because `runtime.InvalidEntryPointArgumentError` wasn't including wrapped errors in its Error() string. This PR adds the wrapped error to the Error() string so that the underlying cause is passed downstream (e.g. to the emulator output).

Emulator output without this change:
```
WARN[23896] ❗  Script reverted                            scriptID=0d00027[...]
WARN[23896] ERR [0d0002] [Error Code: 1101] cadence runtime error Execution failed:
error: invalid argument at index 0
--> 0d00027[...]
```

Emulator output with this change:
```
WARN[0007] ❗  Script reverted                            scriptID=0d00027[...]
WARN[0007] ERR [0d0002] [Error Code: 1101] cadence runtime error Execution failed:
error: invalid argument at index 0: decodeing argument failed: [Error Code: 1052]
transaction arguments are invalid: (argument is not json decodable: failed to decode
value: invalid JSON Cadence structure. invalid type ID: `0xf8d6e0586b0a20c7.Rawr.Foo`)
--> 0d00027[...]
```

______

For contributor use:

### DIFF
--- a/runtime/errors.go
+++ b/runtime/errors.go
@@ -122,7 +122,7 @@ func (e InvalidTransactionAuthorizerCountError) Error() string {
 }
 
 // InvalidEntryPointArgumentError
-
+//
 type InvalidEntryPointArgumentError struct {
 	Index int
 	Err   error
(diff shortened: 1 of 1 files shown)


turbolent:
@mikeylemmon Good catch! Thank you for your contribution 👍 

I'll also apply the same fix to `InvalidTypeAssignmentError`

turbolent:
Turns out the nested error in `InvalidTypeAssignmentError` was never provided, so I removed it

turbolent (approved):

PR #842, opened 2021-04-26
Add hashWithTag function to Crypto contract

SupunS:
## Description

Work towards #582

______

For contributor use:


docs/language/crypto.md
```diff
@@ -107,7 +107,10 @@ The API of the Crypto contract is:
 pub contract Crypto {
 
     // Hash the data using the given hashing algorithm and returns the hashed data.
-    pub fun hash(_ data: [UInt8], algorithm: HashAlgorithm): [UInt8]
+    pub fun hash(data: [UInt8], algorithm: HashAlgorithm): [UInt8]
```
turbolent, line 110: Any reason for adding the argument label? Usually the main argument that the function operates on doesn't need it.

Adding it would be a breaking change, so I'd recommend leaving it away / as-is, and also removing it from `hashWithTag`

turbolent (approved):
Nice! Just one comment regarding backwards-compatibility

tarakby (approved):
I reviewed the general APs, but I am not sure I understand how the runtime exactly works :) 

I guess we will need to sync before we merge the feature branch and review the general APIs together.

docs/language/crypto.md
SupunS: The new `hashWithTag` has more parameters. I was thinking that having parameters with labels as well as without labels, would be a bit confusing to the users. So this change was for consistency with the other method.

>  Usually the main argument that the function operates on doesn't need it.

But if that's a best practice we follow for Cadence, then yeah we can keep it as is.

SupunS: Reverte it in 0633d07

PR #852, opened 2021-04-28
Add new crypto features

### LINKED ISSUE #582: New crypto features support

SupunS:
Closes #582

## Description

This PR introduces the Cadence changes needed for the new crypto features. Includes the following changes:

#### Hash Algorithms
- New KMAC128_BLS_BLS12_381 hashing algorithm
  ```go
  var algo: HashAlgorithm = HashAlgorithm.KMAC128_BLS_BLS12_381
  ```

#### Signature Algorithms
- New BLS_BLS12_381 signature algorithm
  ```go
  var algo: SignatureAlgorithm = SignatureAlgorithm.BLS_BLS12_381
  ```
- Rename `ECDSA_Secp256k1` signature algorithm to `ECDSA_secp256k1`

#### PublicKey Native Type
- New `isValid` field
  - eg:
    ```go
    let publicKey = PublicKey(publicKey: [], signatureAlgorithm: SignatureAlgorithm.ECDSA_P256)

    let valid = publicKey.isValid
    ```
  - Get initialized during the construction of a public key, to the value returned by the `ValidatePublicKey` method on the host env.
- New `verify` method to validate a signature 
  - signature: `pub fun verify(signature: [UInt8], signedData: [UInt8], domainSeparationTag: String, hashAlgorithm: HashAlgorithm): Bool`
  - e.g:
    ```go
    let valid = publicKey.verify(
        signature: [],
        signedData: [],
        domainSeparationTag: "something",
        hashAlgorithm: HashAlgorithm.SHA2_256
    )
    ```
#### Crypto Contract
- New `hashWithTag` function
  - Signature: `pub fun hashWithTag(_ data: [UInt8], tag: string, algorithm: HashAlgorithm): [UInt8]` 
  - e.g:
    ```
    Crypto.hashWithTag(
        "01020304".decodeHex(),
        tag: "some-tag",
        algorithm: HashAlgorithm.SHA3_256
     )
    ``` 

### Changes to the FVM Interface
- New `ValidatePublicKey(key *PublicKey) (bool, error)` method
- Additional parameter ` tag string` in the `Hash` method.
  - New signature: `Hash(data []byte, tag string, hashAlgorithm HashAlgorithm) ([]byte, error)` 
- `PublicKey` has two new fields: `IsValid` and `Validated`
______

For contributor use:


turbolent (approved):
Looks great, nice work! 

I added a breaking changes label because of the naming change from `ECDSA_Secp256k1` to `ECDSA_secp256k1`, which we need to communicate

docs/language/crypto.md
```diff
@@ -99,13 +106,12 @@ The API of the Crypto contract is:
...
 
-    pub struct SignatureAlgorithm {
-        pub let name: String
-    }
-
     // Hash the data using the given hashing algorithm and returns the hashed data.
     pub fun hash(_ data: [UInt8], algorithm: HashAlgorithm): [UInt8]
 
+    // Hash the data using the given hashing algorithm and the tag. Returns the hashed data.
+    pub fun hashWithTag(_ data: [UInt8], tag: string, algorithm: HashAlgorithm): [UInt8]
```
tarakby, line 113: I have suggested `hashWithTag` as I wasn't sure if `PublicKey.Verify` would be implemented entirely by the fvm or would be partly implemented in Cadence code. In the case of the former, I don't think `hashWithTag` is needed in Cadence, we would only need it as an fvm function.

docs/language/accounts.mdx
```diff
@@ -167,6 +167,16 @@ struct AccountKey {
 struct PublicKey {
```
tarakby, line 167: Can `PublicKey` fields (data, signature algo, isValid..) be updated after the creation of the object?

docs/language/accounts.mdx
```diff
@@ -167,6 +167,16 @@ struct AccountKey {
...
     let publicKey: [UInt8]
     let signatureAlgorithm: SignatureAlgorithm
+    let isValid: Bool
+
+    /// Verifies a signature. Checks whether the signature was produced by signing
+    /// the given tag and data, using this public key and the given hash algorithm
+    pub fun verify(
```
tarakby, line 174: As @turbolent mentioned about breaking changes, I think the change from `isValid` to `verify` is a breaking one.

docs/language/accounts.mdx
SupunS: No, they are defined as constant fields (i.e. read-only).

docs/language/accounts.mdx
```diff
@@ -167,6 +167,16 @@ struct AccountKey {
 struct PublicKey {
     let publicKey: [UInt8]
     let signatureAlgorithm: SignatureAlgorithm
+    let isValid: Bool
+
+    /// Verifies a signature. Checks whether the signature was produced by signing
+    /// the given tag and data, using this public key and the given hash algorithm
```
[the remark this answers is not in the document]
SupunS: I think it should be fine to add it here, or we can also add it as a note outside of the code block.

docs/language/accounts.mdx
turbolent: Hmm, both were added, so this isn't a breaking change

tarakby: ah my bad, I somehow though `isValid` already existed.

PR #858, opened 2021-04-29
Pre-allocate and reuse value-path for encoding

### LINKED ISSUE #742: Pre-allocate and reuse value-path during preparation for encoding
### Issue To Be Solved

Currently, when preparing arrays for encoding, a new slice is created for the `value path` for each element in the array.

https://github.com/onflow/cadence/blob/1e61ffa65cd977f1baab73a36dd639ac6bcd4bf2/runtime/interpreter/encode.go#L649-L650

This is slightly expensive both in terms of time and memory, as it allocates new memory for each iteration.

### Suggest A Solution

We could pre-allocate memory for the slice, and then re-use the common component of the path.
```go
// pre-allocate memory
valuePath := append(path, "")

lastIndex := len(path)

for i, value := range v.Values {

    // re-use valuePath, only change the last potion of the path (current index)
    valuePath[lastIndex] = strconv.Itoa(i)
    ...
}
```
I believe it is safe to re-use because preparing the members of a collection is not parallelized.

This could be extended to all collections (dictionaries, composite values, etc.) as well.

fxamacker:
Closes #742

## Description

Pre-allocate and reuse value-path during encoding to prevent reslicing.

Thanks @SupunS  for suggesting this optimization!

### Benchmark Comparisons

Comparison to commit affa4a197b4a59e3e74bc6ae8841fa138d37c4fd shows speed improved by 1.5% to over 27% depending on data encoded.

Memory allocs/op improved by 7-48%.

```
name                                               old time/op    new time/op    delta
EncodeCBOR/comp_v2_96d7e06eaf4b3fcf_14850bytes-4      152µs ± 3%     110µs ± 1%  -27.69%  (p=0.000 n=10+10)
EncodeCBOR/comp_v3_99dc360eee32dcec_160bytes-4       4.34µs ± 0%    3.98µs ± 1%   -8.39%  (p=0.000 n=10+10)
EncodeCBOR/comp_v3_b52a33b7e56868f6_139bytes-4       3.22µs ± 0%    3.17µs ± 1%   -1.65%  (p=0.000 n=10+10)
EncodeCBOR/comp_v3_d99d7e6b4dad41e1_776155bytes-4    7.02ms ± 0%    6.38ms ± 1%   -9.03%  (p=0.000 n=9+10)

name                                               old alloc/op   new alloc/op   delta
EncodeCBOR/comp_v2_96d7e06eaf4b3fcf_14850bytes-4     75.4kB ± 0%    59.4kB ± 0%  -21.21%  (p=0.000 n=10+10)
EncodeCBOR/comp_v3_99dc360eee32dcec_160bytes-4       1.26kB ± 0%    1.17kB ± 0%   -7.59%  (p=0.000 n=10+10)
EncodeCBOR/comp_v3_b52a33b7e56868f6_139bytes-4         864B ± 0%      848B ± 0%   -1.85%  (p=0.000 n=10+10)
EncodeCBOR/comp_v3_d99d7e6b4dad41e1_776155bytes-4    1.96MB ± 0%    1.78MB ± 0%   -9.57%  (p=0.000 n=10+9)

name                                               old allocs/op  new allocs/op  delta
EncodeCBOR/comp_v2_96d7e06eaf4b3fcf_14850bytes-4        620 ± 0%       320 ± 0%  -48.39%  (p=0.000 n=10+10)
EncodeCBOR/comp_v3_99dc360eee32dcec_160bytes-4         19.0 ± 0%      15.0 ± 0%  -21.05%  (p=0.000 n=10+10)
EncodeCBOR/comp_v3_b52a33b7e56868f6_139bytes-4         14.0 ± 0%      13.0 ± 0%   -7.14%  (p=0.000 n=10+10)
EncodeCBOR/comp_v3_d99d7e6b4dad41e1_776155bytes-4     5.95k ± 0%     3.24k ± 0%  -45.56%  (p=0.000 n=10+10)

```
______

For contributor use:

### DIFF
--- a/runtime/interpreter/encode.go
+++ b/runtime/interpreter/encode.go
@@ -793,8 +793,13 @@ func (e *Encoder) encodeArray(
 		return err
 	}
 
+	// Pre-allocate and reuse valuePath.
+	valuePath := append(path, "")
+
+	lastIndex := len(path)
+
 	for i, value := range v.Values {
-		valuePath := append(path[:], strconv.Itoa(i))
+		valuePath[lastIndex] = strconv.Itoa(i)
 		err := e.Encode(value, valuePath, deferrals)
 		if err != nil {
 			return err
@@ -843,7 +848,7 @@ func (e *Encoder) encodeDictionaryValue(
 		return err
 	}
 
-	keysPath := append(path[:], dictionaryKeyPathPrefix)
+	keysPath := append(path, dictionaryKeyPathPrefix)
 
 	// Encode keys (as array) at array index encodedDictionaryValueKeysFieldKey
 	err = e.encodeArray(v.Keys, keysPath, deferrals)
@@ -882,10 +887,15 @@ func (e *Encoder) encodeDictionaryValue(
 		return err
 	}
 
+	// Pre-allocate and reuse valuePath.
+	valuePath := append(path, dictionaryValuePathPrefix, "")
+
+	lastIndex := len(path) + 1
+
 	for _, keyValue := range v.Keys.Values {
 		key := dictionaryKey(keyValue)
 		entryValue, _ := v.Entries.Get(key)
-		valuePath := append(path[:], dictionaryValuePathPrefix, key)
+		valuePath[lastIndex] = key
 
 		if deferred {
 
@@ -1006,6 +1016,11 @@ func (e *Encoder) encodeCompositeValue(
 		return err
 	}
 
+	// Pre-allocate and reuse valuePath.
+	valuePath := append(path, "")
+
+	lastIndex := len(path)
+
 	for pair := v.Fields.Oldest(); pair != nil; pair = pair.Next() {
 		fieldName := pair.Key
 
@@ -1017,7 +1032,7 @@ func (e *Encoder) encodeCompositeValue(
 
 		value := pair.Value
 
-		valuePath := append(path[:], fieldName)
+		valuePath[lastIndex] = fieldName
 
 		// Encode value as fields array element
 		err = e.Encode(value, valuePath, deferrals)


runtime/interpreter/encode.go
```diff
@@ -793,8 +793,13 @@ func (e *Encoder) encodeArray(
 		return err
 	}
 
+	// Pre-allocate and reuse valuePath.
+	valuePath := append(path, "")
```
turbolent, line 797: `append` mutates the argument, so this would effect the calling code. The source argument should be a copy:

```suggestion
	valuePath := append(path[:], "")
```

This is still better than the original, which copied for each element

runtime/interpreter/encode.go
```diff
@@ -1006,6 +1016,11 @@ func (e *Encoder) encodeCompositeValue(
...
```
turbolent, line 1020: Same as above: `append` mutates the argument, so this would effect the calling code. The source argument should be a copy:

```suggestion
	valuePath := append(path[:], "")
```

This is still better than the original, which copied for each element

runtime/interpreter/encode.go
```diff
@@ -843,7 +848,7 @@ func (e *Encoder) encodeDictionaryValue(
 		return err
 	}
 
-	keysPath := append(path[:], dictionaryKeyPathPrefix)
+	keysPath := append(path, dictionaryKeyPathPrefix)
```
turbolent, line 851: Same as above: `append` mutates the argument, so this would effect the calling code. The source argument should be a copy:

```suggestion
	keysPath := append(path[:], dictionaryKeyPathPrefix)
```

runtime/interpreter/encode.go
```diff
@@ -882,10 +887,15 @@ func (e *Encoder) encodeDictionaryValue(
 		return err
 	}
 
+	// Pre-allocate and reuse valuePath.
+	valuePath := append(path, dictionaryValuePathPrefix, "")
+
+	lastIndex := len(path) + 1
+
```
turbolent, line 894: Same as above: append mutates the argument, so this would effect the calling code. The source argument should be a copy.

However, we already did the copy above for `keyPath`, so we can re-use it here to avoid the repeated copy of `path`:

```suggestion
 	// Pre-allocate and reuse valuePath.
	valuePath := keysPath
	valuePath[len(path)] = dictionaryValuePathPrefix
	valuePath = append(valuePath, "")

	lastIndex := len(valuePath)
```

turbolent (commented):
Nice work Faye!

We need to be careful when using `append`. I was overly cautious and copied on every use of `append`, and we should optimize the repeated copies when we e.g. encode arrays, but I don't think we can completely avoid copying in each encoding function, because that would effect the calling function's path. 

We could maybe look into using a persistent list we can cheaply append to without having to copy the prefix, or maybe even a rope: At the leaf functions, the path slice is actually concatenated together, and so there is also some allocation overhead. I think we can avoid the repeated string concatenation by just appending the formatted storage keys with separators directly (but in an immutable way). 

Let me know if you want to sync about the concerns using append, and further about the thi potential optimizations I mentioned above.

fxamacker:
@turbolent I agree we should avoid side effects.  There's a good chance I may have misunderstood the code review.  Please let me know.

If I understand you correctly, calling function's `path` must not be modified by the called function.

Unless I'm mistaken, the calling function's `path` will still have the same element values and number of elements, because the called functions only modify new elements not visible to the calling function when the called function returns.

For example, in Go 1.15.11:
```Go
func m1(path []string) {
	path1 := append(path, "m1")
	fmt.Printf("m1()   path1 %v, len %d, cap %d\n", path1, len(path1), cap(path1))
	path1[1] = "xx"
	fmt.Printf("m1()   path1 %v, len %d, cap %d\n", path1, len(path1), cap(path1))
}

func main() {
	path := []string{"m0"}
	fmt.Printf("main() path  %v, len %d, cap %d\n", path, len(path), cap(path))
	m1(path)
	fmt.Printf("main() path  %v, len %d, cap %d\n", path, len(path), cap(path))
}
```

Produces:
```
main() path  [m0], len 1, cap 1
m1()   path1 [m0 m1], len 2, cap 2
m1()   path1 [m0 xx], len 2, cap 2
main() path  [m0], len 1, cap 1
```
https://play.golang.com/p/RDCcSBs7GVh

runtime/interpreter/encode.go
fxamacker: Unless I'm mistaken, `valuePath := append(path,"")` followed by `valutePath[len(path)] = "abc"` won't mutate `path` in our particular case.  More details at https://github.com/onflow/cadence/pull/858#issuecomment-829652459

turbolent:
Yes, that's a good summary of the two points!

> 1. Further performance gains are possible by using a different data structure

Yes, we should definitely get this PR in first, I was only thinking out loud and brought up this idea, it was not meant to be addressed in this PR 👍 

> 2. calls to append() in this PR mutate the path argument and would affect the calling code

My main point was that `append` modifies the slice's underlying array, so we have to be careful with this side-effect, e.g. see
- https://blog.golang.org/slices#TOC_4.
- https://www.tugberkugurlu.com/archive/working-with-slices-in-go-golang-understanding-how-append-copy-and-slicing-syntax-work#calling-append-on-a-sliced-slice-may-modify-the-original-slice

For example (https://play.golang.org/p/SsapANm9Q-p):

```go
func use (b []int) {
	fmt.Printf("b: %v, cap: %d\n", b, cap(b))
	b = append(b, 20)
	fmt.Printf("b: %v, cap: %d\n", b, cap(b))
}

func main() {
	a := []int{1, 2, 3, 4, 5}
	fmt.Printf("a: %v, cap: %d\n", a, cap(a))

	use(a[2:4])
	fmt.Printf("a: %v, cap: %d\n", a, cap(a))
}
```

```
a: [1 2 3 4 5], cap: 5
b: [3 4], cap: 3
b: [3 4 20], cap: 3
a: [1 2 3 4 20], cap: 5
```

> this PR won't cause `v=append(path,"")` followed by `v[len(path)]="abc"` to mutate path. 

It does, though, I think in our particular case here, this is not a problem: 
Even though a function call deeper in the call stack modifies the slice by extending it (`append`ing one or more elements) and writing to the newly appended element/index, functions up the call stack won't observe it because the slice header's length won't change, and deeper functions only ever *add*, they always pass the full slice.

For example, if function A calls B, passing the path slice of length 3, and B appends one element and modifies it, then A's slice header still has a length of 3, though the capacity of the slice is now >3 and there's an additional element in the array pointed to by the slice header (not visible in A).

I then think it's fine to keep removing the defensive copies in the "deeper" encoding functions, but we must be very careful that we keep the invariant that all encoding functions only append, not modify parts in the passed path, because those changes are visible to the parent function)

turbolent:
@fxamacker Let me resolve my first comments, and go over the PR one more time to make sure we are safe

runtime/interpreter/encode.go
turbolent: Resolving as per discussion below

runtime/interpreter/encode.go
turbolent: Resolving as per discussion below

runtime/interpreter/encode.go
turbolent: Resolving as per discussion below

runtime/interpreter/encode.go
turbolent: Resolving as per discussion below

turbolent (approved):
Nice work! 👍 

I re-reviewed the changes again after thinking about it more, and like I mentioned in the comment, I think these changes should be safe, as the deeper functions only ever add elements, and functions pass the whole slice, not just a part, so mutations of the path slice in deeper functions won't cause observable side-effects in the functions up the call-stack.

Do we already have an issue for applying the same changes to the decoder?

fxamacker:
> Do we already have an issue for applying the same changes to the decoder?

Great question!  I don't think we do.  We should probably open an issue to assess whether the same changes can be done.

Issue #845, opened 2021-04-27
Add deferred decoding for composite values


SupunS:
### Issue To Be Solved

Details: #844

PR #861, opened 2021-04-30
Add deferred decoding support for composite values

SupunS:
Work towards #845

## Description

Composite values have four storable properties:
- location
- qualified-Identifier
- kind
- fields

When reading a value from storage, only these four properties are decoded. Once the decoding is delayed, these properties may not have been properly initialized for a value returned from the decoder. Hence accessing them directly is unsafe.

This PR introduces a getter for each of the four properties. The getter function ensures the property is decoded and properly loaded.

### Overhead
- There are two overheads:
  - Now we have to make a function call instead of directly using a field.
  - Function call always does a nil check.
    - Alternative is to store the getter as a local-field (function pointer) in the composite value, and replace it with a direct-field-returning function after the first invocation.
    - But the benchmarking showed that invoking a nil-checking static function is 5x faster than invoking a function pointer.
- However, both of these turned out to be very **trivial** at `go` level. It doesn't impact the overall performance of the runtime.

### Alternavies
An alternative approach is to introduce a new implementation for `DeferredCompositeValue`. Then make it act as a delegation wrapper for the `CompositeValue`. But the downsides are:
- The `CompositeValue` is used in numerous places in the runtime. We have to add the new `DeferredCompositeValue` also for all such places and duplicate the logic.
- Cannot do the optimization by separating 'meta-content' vs 'fields-content'.

### TODO:
Add the decoding logic once the stream-decoder API is available.

______
For contributor use:

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -6052,7 +6052,7 @@ type CompositeValue struct {
 	Location            common.Location
 	QualifiedIdentifier string
 	Kind                common.CompositeKind
-	Fields              *StringValueOrderedMap
+	fields              *StringValueOrderedMap
 	InjectedFields      *StringValueOrderedMap
 	ComputedFields      *StringComputedFieldOrderedMap
 	NestedVariables     *StringVariableOrderedMap
@@ -6062,6 +6062,28 @@ type CompositeValue struct {
 	destroyed           bool
 	modified            bool
 	stringer            func() string
+	rawContent          []byte
+}
+
+func (v *CompositeValue) Fields() *StringValueOrderedMap {
+	v.ensureLoaded()
+	return v.fields
+}
+
+func (v *CompositeValue) WithFields(fields *StringValueOrderedMap) *CompositeValue {
+	v.fields = fields
+	return v
+}
+
+func (v *CompositeValue) ensureLoaded() {
+	if v.rawContent == nil {
+		return
+	}
+
+	// TODO: otherwise decode
+
+	// Clear the byte cache
+	v.rawContent = nil
 }
 
 type ComputedField func(*Interpreter) Value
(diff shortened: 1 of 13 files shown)


fxamacker:
```Go
	// Raw-content cache for decoded values. Maintains the meta info and field content separately.
	metaContent   []byte
	fieldsContent []byte
```

I love the idea of separating fields from metacontent  (location, qualified-identifier, and kind).  It's more future proof in case we need to add more metacontent.

In the v4 storage format, composite value's content (not tag) is encoded as CBOR array with 5 elements in this sequence:

1.  location
2.  type id (obsolete)
3.  kind
4.  fields
5.  qualified identifier

This isn't optimal because fields is in-between kind and qualified identifier.

With v4 storage format, we can use one of the following approaches:

A.  we probably need to create a new buffer on the heap for `metacontent` containing location, kind, and qualified identifier,  
B.  we can use separate raw CBOR bytes for location, kind, and qualified identifier which avoids new buffer allocation,
C.  we can directly decode location, kind, and qualified identifier without deferring (only defer decoding of fields).

I think A will use more memory than B and C.

If we create storage format v5, we can use an array to store metacontent, which would:

- make it easier to extract as raw CBOR bytes.
- make it easier to add new metacontent because it's just another element.

@SupunS @turbolent what do you think about A, B, C or v5?  Any thoughts?

runtime/interpreter/value.go
```diff
@@ -6049,19 +6049,25 @@ func (v UFix64Value) ConformsToDynamicType(_ *Interpreter, dynamicType DynamicTy
...
+	Functions       map[string]FunctionValue
+	Destructor      FunctionValue
+	Owner           *common.Address
+	destroyed       bool
+	modified        bool
+	stringer        func() string
+
+	// Raw-content cache for decoded values. Maintains the meta info and field content separately.
+	metaContent   []byte
+	fieldsContent []byte
```
turbolent, line 6070: Would be good to have comments here describing what is considered part of each of these two parts

turbolent (commented):
The approach and the code so far looks good 👍

SupunS:
@fxamacker Thank you for your thoughts and suggestions. I managed to get the deferred decoding for both meta-info and fields 
 using the current implementation. The only drawback is loading fields would also load the meta info (but not the other way around), which I assume is fine for now.

SupunS:
Closing this PR in favour of https://github.com/onflow/cadence/pull/896

PR #873, opened 2021-05-05
Remove field names from composite value encoding

### LINKED ISSUE #834: Storage format optimization: Don't store composite field names
### Issue To Be Solved

Currently the encoding of composite values includes field names, which is often a significant overhead.

Optimize the storage format by removing the field names.

### Suggested A Solution

- [ ] Store field values in a CBOR array
- [ ] Forbid the removal of fields in contract updates, as this would lead to stored data be re-interpreted in a different way
- [ ] Look up field names from elaboration

turbolent:
Closes #834

## Description

- Update the storage format: Only encode composites' field values, and not the field names. Bump the version, but keep support for decoding the previous version.
- Remove `Fields` field from composite type and interface types. This was not really needed, and actually error prone, because it was supposed to be initialized properly, to match the information in `Members`. I replaced it for one specific other use, a slice of the stored field members
- Remove the state decoding tool. Decoding now requires the field names, which are not known given just a state dump

TODO:
- [ ] Forbid the removal of fields in contract update
______

For contributor use:


turbolent:
This is outdated, we will get back to this based on the new storage layer

PR #875, opened 2021-05-05
Add non-local type inference for expressions

### LINKED ISSUE #512: Non-local type inference
### Problem

Type inference of array literals does not take into account the variable declaration's type annotation.

### Steps to Reproduce

```kotlin
let someInts: [UInt32] = [2, 3]
```

```
Execution failed: Checking failed: mismatched types. expected `[UInt32]`, got `[Int]`
```

SupunS:
Work towards #512

## Description

Simple idea: 
- Currently, when an expression is visited, the type is first inferred by looking at the expression, and then validates it against the type annotations (or the contextually expected type)
- With these new changes, the checker takes the contextually expected type down the tree when visiting an expression, and uses that to infer the type and validate.
  - Must use `checker.VisitExpression(...)` method when visiting an expression (instead of `expression.Accept(checker)`)
  - All type compatibility validations now happens in a single place (`VisitExpression` method)
  - Type inferring logics are distributed to the respective visitor method.

### Next Steps:
Add type inferring for function invocation arguments.
  - With generics, it is a bit unclear how to support this. Need to figure that out.
______

For contributor use:


runtime/sema/check_swap.go
```diff
@@ -25,59 +25,26 @@ import (
...
-		}
-	}
+	lhsChecked := checker.checkSwapStatementExpression(swap.Left, leftType, common.OperandSideLeft)
+	rhsChecked := checker.checkSwapStatementExpression(swap.Right, rightType, common.OperandSideRight)
 
-	if !IsValidAssignmentTargetExpression(swap.Right) {
+	// The types of both sides must be subtypes of each other,
+	// so that assignment can be performed in both directions.
+	// i.e: The two types have to be equal.
+	if lhsChecked && rhsChecked && !leftType.Equal(rightType) {
```
SupunS, line 40: check for equality because `A⊆B and B⊆A ⟺ A=B`

runtime/sema/checker.go
```diff
@@ -546,6 +547,9 @@ func (checker *Checker) checkTransfer(transfer *ast.Transfer, valueType Type) {
 	}
 }
 
+// This method is only used for checking invocation-expression.
+// This is also temporary, until the type inferring support is added for func arguments
+// TODO: Remove this method
 func (checker *Checker) checkTypeCompatibility(expression ast.Expression, valueType Type, targetType Type) bool {
```
SupunS, line 553: This function is no longer used, except in function-invocation expression checking.

The functionality inside this method is now distributed to the respective expressions.
e.g:  
the logic inside the `case *ast.IntegerExpression:` of the switch-case in this method, is now moved to `func (checker *Checker) VisitIntegerExpression()` method.

runtime/sema/check_expression.go
```diff
@@ -148,22 +148,6 @@ func (checker *Checker) VisitExpressionStatement(statement *ast.ExpressionStatem
...
-	//
-	// This check is already performed for e.g. variable declarations
-	// and function function arguments.
-	//
-	// It might seem odd that the target type is the value type,
-	// but this is exactly the case for an expression that is a separate statement.
-
-	if !ty.IsInvalidType() {
-		checker.checkTypeCompatibility(expression, ty, ty)
-	}
```
SupunS, line 165: No longer needed. These validations now happen at the expression level

runtime/tests/checker/type_inference_test.go
```diff
@@ -0,0 +1,293 @@
...
+			foo(a: [1, 2, 3])
+		}
+
+		fun foo(a: [Int8]) {
+		}
+	`)
+
+	// Type inferring for function arguments is not supported yet.
+	require.Error(t, err)
+	assert.IsType(t, &sema.CheckerError{}, err)
```
turbolent, line 184: Could you please make assertion more specific here, i.e. there is one error, a type mismatch?

I verified this is also the current behaviour 👍

runtime/tests/checker/type_inference_test.go
```diff
@@ -0,0 +1,293 @@
...
+
+		require.Error(t, err)
+		require.IsType(t, &sema.CheckerError{}, err)
+
+		checkerErr := err.(*sema.CheckerError)
+
+		// Should only give a single error
+		require.Len(t, checkerErr.Errors, 1)
+
+		assert.IsType(t, &sema.InvalidUnaryOperandError{}, checkerErr.Errors[0])
```
turbolent, line 231:
```suggestion
		errs := ExpectCheckerErrors(t, err, 1)

		assert.IsType(t, &sema.InvalidUnaryOperandError{}, errs[0])
```

runtime/tests/checker/type_inference_test.go
```diff
@@ -0,0 +1,293 @@
...
+}
+
+func TestForceExpressionTypeInference(t *testing.T) {
+
+	t.Parallel()
+
+	t.Run("array forced", func(t *testing.T) {
+		_, err := ParseAndCheck(t, `
+			fun test() {
+				var x: [Int8] = [5, 7, 2]!
```
turbolent, line 242: Nice, this answers the question I had above regarding force-unwrapping non-optionals 👍

runtime/tests/checker/type_inference_test.go
```diff
@@ -0,0 +1,293 @@
...
+				var x = [1, 764] as [Int8]
+			}
+		`)
+
+		require.Error(t, err)
+		require.IsType(t, &sema.CheckerError{}, err)
+		checkerErr := err.(*sema.CheckerError)
+
+		require.Len(t, checkerErr.Errors, 1)
+		assert.IsType(t, &sema.InvalidIntegerLiteralRangeError{}, checkerErr.Errors[0])
```
turbolent, line 276:
```suggestion
		errs := ExpectCheckerErrors(t, err, 1)
		
		assert.IsType(t, &sema.InvalidIntegerLiteralRangeError{}, errs[0])
```

runtime/tests/checker/type_inference_test.go
```diff
@@ -0,0 +1,293 @@
...
+				var x = [1, "hello"] as [Int8]
+			}
+		`)
+
+		require.Error(t, err)
+		require.IsType(t, &sema.CheckerError{}, err)
+		checkerErr := err.(*sema.CheckerError)
+
+		require.Len(t, checkerErr.Errors, 1)
+		assert.IsType(t, &sema.TypeMismatchError{}, checkerErr.Errors[0])
```
turbolent, line 291:
```suggestion
		errs := ExpectCheckerErrors(t, err, 1)
		
		assert.IsType(t, &sema.TypeMismatchError{}, errs[0])
```

runtime/sema/check_expression.go
```diff
@@ -264,23 +297,7 @@ func (checker *Checker) visitValueIndexingExpression(
...
-		!IsSubType(indexingType, indexedType.IndexingType()) {
+	checker.VisitExpression(indexingExpression, indexedType.IndexingType())
 
-		checker.report(
-			&NotIndexingTypeError{
-				Type:  indexingType,
-				Range: ast.NewRangeFromPositioned(indexingExpression),
-			},
-		)
-	}
```
turbolent, line 283: I guess `NotIndexingTypeError` isn't really useful, as a type mismatch error is also reported and much more useful to fix the problem

runtime/sema/check_reference_expression.go
```diff
@@ -27,65 +27,12 @@ import (
...
-
-	if _, ok := referencedType.(*OptionalType); ok {
-
-		checker.report(
-			&OptionalTypeReferenceError{
-				ActualType: referencedType,
-				Range:      getReferencedExpressionRange(),
-			},
-		)
-	}
```
turbolent, line 82: I think we should keep this error, because it is a special case of referencing and the type mismatch error on its own might not be very helpful

runtime/sema/check_array_expression.go
```diff
@@ -40,28 +74,23 @@ func (checker *Checker) VisitArrayExpression(expression *ast.ArrayExpression) as
...
-		} else if !valueType.IsInvalidType() &&
-			!IsSubType(valueType, elementType) {
-
-			checker.report(
-				&TypeMismatchError{
-					ExpectedType: elementType,
-					ActualType:   valueType,
-					Range:        ast.NewRangeFromPositioned(value),
-				},
-			)
```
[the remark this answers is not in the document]
SupunS: This check now happens when we visit the element, i.e: `checker.VisitExpression(value, elementType)`. There, we pass the expected element type, so the `VisitExpression` validates the actual type against this expected-type, and logs the type-mismatching error. That way, now all the type mismatch checks are done and the errors are reported in a single place (which is `VisitExpression`).

If the expected type for array-literal is not available, then the type of the first element is taken as the  'expected element type', just like before.

runtime/sema/check_assignment.go
```diff
@@ -207,27 +203,11 @@ func (checker *Checker) visitIdentifierExpressionAssignment(
...
-		!checker.checkTypeCompatibility(valueExpression, valueType, variable.Type) {
-
-		checker.report(
-			&TypeMismatchError{
-				ExpectedType: variable.Type,
-				ActualType:   valueType,
-				Range:        ast.NewRangeFromPositioned(valueExpression),
-			},
-		)
-	}
```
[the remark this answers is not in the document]
SupunS: Same as above (see https://github.com/onflow/cadence/pull/875#discussion_r629809002).

runtime/sema/checker.go
```diff
@@ -2297,3 +2301,63 @@ func (checker *Checker) convertInstantiationType(t *ast.InstantiationType) Type
...
+	// forceTypes - flag indicating whether to use the expected type as a hard requirement (forceTypes = true)
+	// or whether to use the expected type for type inferring only (forceTypes = false)
+
+	// Cache the current contextually expected type, and set the `expectedType`
+	// as the new contextually expected type.
+	prevExpectedType := checker.expectedType
+
+	// If the expected type is invalid or void, treat it in the same manner as
+	// expected type is unknown.
+	if expectedType == VoidType ||
```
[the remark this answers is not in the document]
SupunS: We could also pass `Void` type as an expected type, but that will cause a duplicate error message.

The reason is, currently, the expected type can be `Void` only for function returns. But that is already validated at the return-statement checking, and a more meaningful error message is logged there (`InvalidReturnValueError`). If we allow `Void` here, that will also log an additional error message saying `expected Void, found X`, which is sort of duplicate.

runtime/sema/check_reference_expression.go
SupunS: This is actually a bit tricky now. 
Since there is a type-mismatch error at the expression level, what we get back as the `referencedType` (i.e: the type of the referenced expr) is also the expected-type, not the actual type of the expression.

Even if we add that check/error back, we can only validate a scenario like:
```
let x: Int? = 1
let ref = &x as &Int?    // Note &Int?
```

But we can't validate this:
```
let x: Int? = 1
let ref = &x as &Int    // Note &Int
```

runtime/sema/checker.go
```diff
@@ -2297,3 +2301,63 @@ func (checker *Checker) convertInstantiationType(t *ast.InstantiationType) Type
...
+
+		// If there are type mismatch errors, return the expected type as the type of the expression.
+		// This is done to avoid the same error getting delegated up.
+		return expectedType
+	}
+
+	return actualType
+}
+
+func getPosition(expression ast.Expression) ast.Range {
```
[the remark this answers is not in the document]
SupunS: :+1: Renamed to `expressionRange()`

runtime/tests/checker/type_inference_test.go
```diff
@@ -0,0 +1,293 @@
...
+
+	t.Run("invalid", func(t *testing.T) {
+		_, err := ParseAndCheck(t, `
+			fun test() {
+				var x: {Int8:Int64} = {23423:6, 1:5}
+			}
+		`)
+
+		require.Error(t, err)
+		assert.IsType(t, &sema.CheckerError{}, err)
```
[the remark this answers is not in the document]
SupunS: Added more detailed asserts for all the error-returned scenarios.

runtime/sema/check_unary_expression.go
```diff
@@ -43,13 +43,15 @@ func (checker *Checker) VisitUnaryExpression(expression *ast.UnaryExpression) as
 		expectedType := BoolType
 		if !IsSubType(valueType, expectedType) {
 			reportInvalidUnaryOperator(expectedType)
+			return InvalidType
```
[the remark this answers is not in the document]
turbolent: Nice!

runtime/sema/checker.go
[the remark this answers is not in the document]
turbolent: `Void` can also be used for type annotations other than function's return value.

Maybe then we should just remove `InvalidReturnValueError` and have the error be reported here?

runtime/sema/check_array_expression.go
```diff
@@ -22,14 +22,48 @@ import "github.com/onflow/cadence/runtime/ast"
...
 
+	expectedType := UnwrapOptionalType(checker.expectedType)
+
 	var elementType Type
+	if expectedType != nil {
+
+		switch typ := expectedType.(type) {
+
+		case *ConstantSizedType:
+			elementType = typ.ElementType(true)
```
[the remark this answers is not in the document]
turbolent: Yeah, maybe `ElementType` should just be split into two functions, one for getting the element type when it's read/is not an assignment, one when its set/is an assignment.

runtime/sema/check_force_expression.go
```diff
@@ -24,7 +24,14 @@ import (
 
 func (checker *Checker) VisitForceExpression(expression *ast.ForceExpression) ast.Repr {
 
-	valueType := expression.Expression.Accept(checker).(Type)
+	// Expected type of the `expression.Expression` is the optional of expected type of current context.
+	// i.e: if `x!` is `string`, then `x` is expected to be `string?`.
+	expectedType := UnwrapOptionalType(checker.expectedType)
+	if expectedType != nil {
+		expectedType = &OptionalType{Type: expectedType}
+	}
```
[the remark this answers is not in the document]
turbolent: Ah, good catch with the repeated unwrapping! If there aren't any yet, could you please add test cases for e.g. `y: Int??; y!!`, `y: Int?; y!!` and `y: Int; y!!` ?

runtime/sema/checker.go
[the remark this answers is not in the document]
SupunS: Oh, I wasn't aware of that. Thanks for pointing that one out!

Yeah, then it makes sense to do as you suggested. That would also make it simple :open_hands:

runtime/sema/check_force_expression.go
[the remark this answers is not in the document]
SupunS: The first one was there. Added a couple of tests for the other two cases

runtime/sema/checker.go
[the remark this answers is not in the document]
SupunS: Updated in a9a0f91

runtime/sema/check_reference_expression.go
SupunS: We wouldn't still get the 'actual-type' of the referenced-expr.

I think we could solve this by making the `VisitExpression()` method to return two values (two types). 
- One is the 'actual-type' of the expression - This is not returned now.
- The type that others should see as the actual-type (like a 'visible-type') - This is what the method currently returns. 

For this particular case, we could use the 'actual-type'. Every other case would use the 'visible-type' for type checking - because it avoids the errors being delegated up.

SupunS: Added with abf2b1c

runtime/tests/checker/type_inference_test.go
```diff
@@ -324,10 +367,32 @@ func TestCastExpressionTypeInference(t *testing.T) {
...
+
+		errs := ExpectCheckerErrors(t, err, 1)
+
+		require.IsType(t, &sema.TypeMismatchError{}, errs[0])
 		typeMismatchErr := errs[0].(*sema.TypeMismatchError)
-		assert.Equal(t, typeMismatchErr.ExpectedType, sema.Int8Type)
-		assert.Equal(t, typeMismatchErr.ActualType, sema.StringType)
+
+		assert.Equal(t, sema.VoidType, typeMismatchErr.ExpectedType)
+		assert.Equal(t, sema.IntType, typeMismatchErr.ActualType)
```
turbolent: 👍

runtime/sema/check_reference_expression.go
turbolent: Great idea, the commit looks good! 👍

turbolent (approved):
Looks great, awesome work! 👏 ⛵

SupunS:
Thanks, @turbolent for the reviews! :bow:

PR #877, opened 2021-05-05
Remove the high-level storage API

### LINKED ISSUE #876: Remove High-Level storage API
### Issue To Be Solved

In addition to a low-level storage API (`runtime.Interface`'s `GetValue`/`SetValue`), which is used for storage reads and writes during execution, Cadence also provides a high-level API to inform the hist environment about storage changes on a "high-level", i.e. in-memory objects rather than serialized bytes.

This high-level storage API is only used in the Playground API, but currently not working properly, because contract updates (deployments, updates, etc.) are delayed by the host-environment to the end of the execution of the Cadence program – however, the high-level export occurs before the contract changes are committed, which results in a failure to export newly stored values during a contract deployment.

### Suggest A Solution

Remove this API, as it currently cannot be implemented properly. The feature of inspecting account storage to obtain Cadence values rather than just serialized data can be replicated by using Cadence's new `ReadStored` method.

### Context

- onflow/flow-playground-api#8

turbolent:
Closes #876 

See the issue description for the reason

______

For contributor use:


janezpodhostnik (approved):

runtime/runtime_storage.go
```diff
@@ -313,31 +291,6 @@ func (s *runtimeStorage) writeCached(inter *interpreter.Interpreter) {
 
 	// Write cache entries in order
```
janezpodhostnik, line 292: is this comment still relevant?

turbolent: The comment applied to both parts below, writing the items in order using the high-level storage API, and then again using the low-level API, so yes, it's still relevant

Issue #872, opened 2021-05-04
Adjust Crypto contract


turbolent:
### Context

A discussion about PR #852 identified some required changes.

### Definition of Done

- [x] Document the `PublicKey` initializer in `docs/language/accounts.mdx`
- [ ] Change the new `isValid` field to be computed. The `publicKey` field could be changed after initialization.
  - [ ] Call the runtime verification function on each call (this might get cached at some point later).
  - [ ] Remove the internal `validated` flag  
- [ ] Check that the implementation of `VerifySignature` in FVM rejects invalid public keys, i.e. returns an error when decoding fails

### Future work / open questions

- How could we avoid the repeated isValid? 
  - Keys use 64/96 bytes. We could keep a copy of the last validated key
  - We could re-verify when the raw public key changes. This would require adding a callback to arrays

- The raw public key is passed to the runtime interface function `VerifySignature`, which leads to the implementation of the function in FVM having to re-decode the raw public key on each call. This could be optimized by  e.g. using a shared data structure that contains the potentially decoded public key

SupunS:
There is documentation for `PublicKey` constructor here: https://docs.onflow.org/cadence/language/accounts/#account-keys

turbolent:
@SupunS Ah, I see. I just had talked to Tarak again and we looked for it, and couldn't find it. Could you please add a reference to it to the Crypto page? It refers to it / uses it too. I think that will prevent confusion

turbolent:
As for the second item in the description, that task has changed to making the key immutable, which is now done through #879, correct? I guess this issue can be closed now?

SupunS:
Added a reference to the public key in the crypt9 contract in: https://github.com/onflow/cadence/pull/912
We can close the issue with that.

I think the unification of the public-key representation across the Cadence and FVM has to/can be done as a separate issue.

PR #879, opened 2021-05-07
Make PublicKey value immutable

SupunS:
Work towards #872

## Description

Makes the raw public-key field a computed field, so that it returns a copy always.

______

For contributor use:


runtime/runtime.go
```diff
@@ -2481,11 +2481,15 @@ func (r *interpreterRuntime) newPublicAccountKeys(addressValue interpreter.Addre
...
 	// publicKey field
-	key, ok := publicKey.Fields.Get(sema.PublicKeyPublicKeyField)
+	publicKeyFieldGetter, ok := publicKey.ComputedFields.Get(sema.PublicKeyPublicKeyField)
 	if !ok {
 		panic("public key value is not set")
 	}
 
+	// Interpreter is not available at this point.
+	// For now its safe to pass a nil interpreter here, as the publicKey field doesn't use it.
+	key := publicKeyFieldGetter(nil).(*interpreter.ArrayValue)
```
SupunS, line 2491: Not sure is there a better way to do this. The interpreter is not available at this point.

turbolent (approved):
Looks good!

PR #880, opened 2021-05-08
Record variable declaration ranges

turbolent:
## Description

Record the ranges in which variable declarations are available, if position information is enabled (off by default).

This feature can be used to provide completion items in the language server.

Also, remove the unused `Hash` function for `sema.Position` and the dependency on `fasthash`.

______

For contributor use:


runtime/ast/position.go
```diff
@@ -40,12 +40,13 @@ func (position Position) Shifted(length int) Position {
 	}
 }
 
-func (position Position) Hash() (result uint32) {
```
janezpodhostnik, line 43: Was this not used?

janezpodhostnik (approved):
LGTM

runtime/ast/position.go
turbolent: It had been previously used / was previously needed when we used a persistent map/set library that used this function for the key. We replaced it and removed the dependency a while ago, and this was a left-over

SupunS (approved):
Looks good!

PR #886, opened 2021-05-11
Handle case with no arguments

MaxStalker:
## Description
Language server is crashing if there is no arguments in script. This PR shall fix this behaviour.

______

For contributor use:

### DIFF
--- a/languageserver/integration/codelenses.go
+++ b/languageserver/integration/codelenses.go
@@ -145,9 +145,11 @@ func (i *FlowIntegration) entryPointActions(
 		return codeLenses, nil
 	}
 
+	emptyEntryPoints := len(entryPointInfo.parameters) == 0
+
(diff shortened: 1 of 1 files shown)


languageserver/integration/codelenses.go
```diff
@@ -145,9 +145,11 @@ func (i *FlowIntegration) entryPointActions(
 		return codeLenses, nil
 	}
 
+	emptyEntryPoints := len(entryPointInfo.parameters) == 0
```
turbolent, line 148: Maybe rename this to e.g. `noEntryPointParameters` or just `noParameters`, as the current name looks like it implies there are no entry points, but the test is for the parameters

MaxStalker: Good point :)

psiemens (approved):

PR #895, opened 2021-05-13
capture computation used for all transactions

ramtinms:
## Description

Seems we were only capturing computation used if they pass the limit and not for all transactions.

______

For contributor use:

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -171,6 +171,9 @@ type SignatureValidationHandlerFunc func(
 	key *CompositeValue,
 ) BoolValue
 
+// ExitHandlerFunc is a function that is called at the end of execution
+type ExitHandlerFunc func() error
+
 // CompositeTypeCode contains the the "prepared" / "callable" "code"
 // for the functions and the destructor of a composite
 // (contract, struct, resource, event).
@@ -248,6 +251,7 @@ type Interpreter struct {
 	uuidHandler                    UUIDHandlerFunc
 	PublicKeyValidationHandler     PublicKeyValidationHandlerFunc
 	SignatureValidationHandler     SignatureValidationHandlerFunc
+	ExitHandler                    ExitHandlerFunc
 	interpreted                    bool
 	statement                      ast.Statement
 }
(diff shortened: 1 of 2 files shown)


turbolent:
I see, I didn't know we wanted to always capture the used computation, sounds good.

The proposed change will be quite expensive, as the limit check function is called for each statement, loop iteration, and function call. 

I think we should keep the current code as-is, so we can still abort the program when the limit is reached, and add code that also calls `SetComputationUsed` at the end of the successful execution, i.e. at the end of `ExecuteTransaction` (and also maybe `ExecuteScript`?)

ramtinms:
@turbolent how about changes  I made at 0f1e46b ?

runtime/interpreter/interpreter.go
```diff
@@ -171,6 +171,9 @@ type SignatureValidationHandlerFunc func(
 	key *CompositeValue,
 ) BoolValue
 
+// ExitHandlerFunc is a function that is called at the end of execution
+type ExitHandlerFunc func() error
```
turbolent, line 175: Looks good! 👍

runtime/runtime.go
```diff
@@ -1061,13 +1062,6 @@ func (r *interpreterRuntime) meteringInterpreterOptions(runtimeInterface Interfa
 			return
 		}
 
-		var err error
-		wrapPanic(func() {
-			err = runtimeInterface.SetComputationUsed(used)
-		})
-		if err != nil {
-			panic(err)
-		}
```
turbolent, line 1070: Are we sure we want to remove this? The exit handle won't be called when the computation limit exceeded panic below occurs, right?

I guess it's fine, if FVM still performs the logic of `SetComputationUsed` when it encounters a `ComputationLimitExceededError`, but maybe keep the explicit `SetComputationUsed` in both cases (success and limit exceeded)?

runtime/runtime.go
```diff
@@ -320,7 +320,8 @@ func (r *interpreterRuntime) interpret(
 		exportedValue = newExportableValue(result, inter)
 	}
 
-	return exportedValue, inter, nil
+	err = inter.ExitHandler()
```
turbolent, line 323: CI is failing because this handler is currently always called, but it is optional:
```suggestion
	if inter.ExitHandler != nil {
		err = inter.ExitHandler()
	}
```

runtime/runtime.go
ramtinms: hmm, how can we change it to make sure it is always called before returning the panic (other errors as well) ?

ramtinms: I put it back.

turbolent (approved):
Looks good!

PR #896, opened 2021-05-13
Add deferred decoding support for composite values

### LINKED ISSUE #845: Add deferred decoding for composite values

SupunS:
Closes #845

## Description

Composite values have four storable properties:
- location
- qualified-Identifier
- kind
- fields

When reading a value from storage, only these four properties are decoded. Once the decoding is delayed, these properties may not have been properly initialized for a value returned from the decoder. Hence accessing them directly is unsafe.

This PR introduces a getter for each of the four properties. The getter function ensures the property is decoded and properly loaded.

### Approach:
- Composite value has two new extra properties:
  - `content` - Holds the raw content. Contains meta-info +  fields-content.
  - `fieldsContent` - Initially `nil`. Gets populated when the meta info is loaded. Hold the raw byte content related to composite value's fields
- When one of the meta info fields is accessed, all the meta info is loaded. At the same time, fields-content is also extracted out from the `content` and cached in `fieldsContent`.
- One drawback is accessing a field will also load the meta-info. This is a limitation in the storage format.

### Overhead
- There are two overheads:
  - Now we have to make a function call instead of directly using a field.
  - Function call always does a nil check.
    - Alternative is to store the getter as a local-field (function pointer) in the composite value, and replace it with a direct-field-returning function after the first invocation.
    - But the benchmarking showed that invoking a nil-checking static function is 5x faster than invoking a function pointer.
- However, both of these turned out to be very **trivial** at `go` level. It doesn't impact the overall performance of the runtime.

### Performance
For a composite value of `Foo` type, where :
```kotlin
struct Foo {
  var fname:   String
  var lname:   String
  var age:     Int
  var status:  String
  var address: Address
} 

struct Address {
  var street:  String
  var city:    String
  var state:   String
  var country: String
} 
```
Decoding perf numbers are:
```
name                                            old time/op    new time/op    delta
CompositeDeferredDecoding/Simply_decode-12        5.33µs ± 6%    1.45µs ± 5%  -72.86%  (p=0.008 n=5+5)
CompositeDeferredDecoding/Access_identifier-12    5.69µs ± 7%    2.23µs ±15%  -60.87%  (p=0.008 n=5+5)
CompositeDeferredDecoding/Access_field-12         5.91µs ± 2%    5.44µs ± 4%   -7.91%  (p=0.016 n=4+5)
CompositeDeferredDecoding/Re-encode_decoded-12    2.43µs ±26%    1.76µs ±13%  -27.68%  (p=0.008 n=5+5)

name                                            old alloc/op   new alloc/op   delta
CompositeDeferredDecoding/Simply_decode-12        2.61kB ± 0%    0.55kB ± 0%  -78.83%  (p=0.008 n=5+5)
CompositeDeferredDecoding/Access_identifier-12    2.61kB ± 0%    0.86kB ± 0%  -66.87%  (p=0.008 n=5+5)
CompositeDeferredDecoding/Access_field-12         2.82kB ± 0%    2.42kB ± 0%  -14.20%  (p=0.008 n=5+5)
CompositeDeferredDecoding/Re-encode_decoded-12    1.16kB ± 0%    0.96kB ± 0%  -17.24%  (p=0.008 n=5+5)

name                                            old allocs/op  new allocs/op  delta
CompositeDeferredDecoding/Simply_decode-12          67.0 ± 0%       6.0 ± 0%  -91.04%  (p=0.008 n=5+5)
CompositeDeferredDecoding/Access_identifier-12      67.0 ± 0%      13.0 ± 0%  -80.60%  (p=0.008 n=5+5)
CompositeDeferredDecoding/Access_field-12           69.0 ± 0%      48.0 ± 0%  -30.43%  (p=0.008 n=5+5)
CompositeDeferredDecoding/Re-encode_decoded-12      14.0 ± 0%      10.0 ± 0%  -28.57%  (p=0.008 n=5+5)
```
_*'Simply_decode' - Only decode the value. Doesn't access the content of the value, like fields, type, etc._
______

For contributor use:

### DIFF
--- a/runtime/interpreter/decode.go
+++ b/runtime/interpreter/decode.go
@@ -1713,8 +1713,6 @@ func decodeCompositeMetaInfo(v *CompositeValue, content []byte) error {
 		return err
 	}
 
-	v.fieldsContent = fieldsContent
-
 	// Qualified identifier
 
 	// Decode qualified identifier at array index encodedCompositeValueQualifiedIdentifierFieldKey
@@ -1733,6 +1731,7 @@ func decodeCompositeMetaInfo(v *CompositeValue, content []byte) error {
 	v.location = location
 	v.qualifiedIdentifier = qualifiedIdentifier
 	v.kind = kind
+	v.fieldsContent = fieldsContent
 
 	return nil
 }
--- a/runtime/interpreter/decode.go
+++ b/runtime/interpreter/decode.go
@@ -1777,6 +1777,7 @@ func decodeCompositeFields(v *CompositeValue, content []byte) error {
 	fields := NewStringValueOrderedMap()
 
 	// Pre-allocate and reuse valuePath.
+	//nolint:gocritic
 	valuePath := append(v.valuePath, "")
 
 	lastValuePathIndex := len(v.valuePath)
--- a/runtime/interpreter/decode.go
+++ b/runtime/interpreter/decode.go
@@ -628,9 +628,7 @@ func (d *DecoderV4) decodeComposite(path []string) (*CompositeValue, error) {
 	valuePath := make([]string, len(path))
 	copy(valuePath, path)
 
-	compositeValue := NewDeferredCompositeValue(valuePath, content, d.owner)
-	compositeValue.modified = false
-	return compositeValue, nil
+	return NewDeferredCompositeValue(valuePath, content, d.owner, d.decodeCallback), nil
 }
 
 var bigOne = big.NewInt(1)
@@ -1640,7 +1638,7 @@ func (d *DecoderV4) decodeCapabilityStaticType() (StaticType, error) {
 // This also extracts out the fields raw content and cache it separately inside the value.
 //
 func decodeCompositeMetaInfo(v *CompositeValue, content []byte) error {
-	d, err := NewByteDecoder(content, v.Owner, CurrentEncodingVersion, nil)
+	d, err := NewByteDecoder(content, v.Owner, CurrentEncodingVersion, v.decodeCallback)
 	if err != nil {
 		return err
 	}
@@ -1743,7 +1741,7 @@ func decodeCompositeMetaInfo(v *CompositeValue, content []byte) error {
 // decodeCompositeFields decodes fields from the byte content and updates the composite value.
 //
 func decodeCompositeFields(v *CompositeValue, content []byte) error {
-	d, err := NewByteDecoder(content, v.Owner, CurrentEncodingVersion, nil)
+	d, err := NewByteDecoder(content, v.Owner, CurrentEncodingVersion, v.decodeCallback)
 	if err != nil {
 		return err
 	}
--- a/runtime/interpreter/decode.go
+++ b/runtime/interpreter/decode.go
@@ -624,7 +624,11 @@ func (d *DecoderV4) decodeComposite(path []string) (*CompositeValue, error) {
 		return nil, err
 	}
 
-	compositeValue := NewDeferredCompositeValue(path, content, d.owner)
+	// Make a copy so that the path will not be affected by any modification at upper levels.
+	valuePath := make([]string, len(path))
+	copy(valuePath, path)
+
+	compositeValue := NewDeferredCompositeValue(valuePath, content, d.owner)
 	compositeValue.modified = false
 	return compositeValue, nil
 }
@@ -1648,7 +1652,7 @@ func decodeCompositeMetaInfo(v *CompositeValue, content []byte) error {
 	if err != nil {
 		if e, ok := err.(*cbor.WrongTypeError); ok {
 			return fmt.Errorf("invalid composite encoding (@ %s): expected [%d]interface{}, got %s",
-				strings.Join(v.decodePath, "."),
+				strings.Join(v.valuePath, "."),
 				expectedLength,
 				e.ActualType.String(),
 			)
@@ -1658,7 +1662,7 @@ func decodeCompositeMetaInfo(v *CompositeValue, content []byte) error {
 
 	if size != expectedLength {
 		return fmt.Errorf("invalid composite encoding (@ %s): expected [%d]interface{}, got [%d]interface{}",
-			strings.Join(v.decodePath, "."),
+			strings.Join(v.valuePath, "."),
 			expectedLength,
 			size,
 		)
@@ -1671,7 +1675,7 @@ func decodeCompositeMetaInfo(v *CompositeValue, content []byte) error {
 	if err != nil {
 		return fmt.Errorf(
 			"invalid composite location encoding (@ %s): %w",
-			strings.Join(v.decodePath, "."),
+			strings.Join(v.valuePath, "."),
 			err,
 		)
 	}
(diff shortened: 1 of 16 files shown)


fxamacker (commented):
Nice work!  The performance improvements look great!  :rocket:

runtime/interpreter/value.go
```diff
@@ -6487,6 +6544,72 @@ func (v *CompositeValue) ConformsToDynamicType(
...
+	if v.content == nil {
+		return
+	}
+
+	err := decodeCompositeMetaInfo(v, v.content)
+	if err != nil {
+		panic(err)
+	}
+
+	// Raw content is no longer needed. Clear the cache and free-up the memory.
```
fxamacker, line 6586: What do you think about postponing the release of v.content until it is modified?  That way, we don't have to re-encode them if they are stored without changes.

runtime/interpreter/value.go
```diff
@@ -6487,6 +6544,72 @@ func (v *CompositeValue) ConformsToDynamicType(
...
+	if v.fieldsContent == nil {
+		return
+	}
+
+	err := decodeCompositeFields(v, v.fieldsContent)
+	if err != nil {
+		panic(err)
+	}
+
+	// Path and the fields-content are no longer needed.
```
fxamacker, line 6607: What do you think about postponing the release of v.fieldsContent until it is modified?  That way, we don't have to re-encode it if it's stored without changes.

runtime/interpreter/decode.go
```diff
@@ -1736,3 +1627,186 @@ func (d *DecoderV4) decodeCapabilityStaticType() (StaticType, error) {
...
+			)
+		}
+		return err
+	}
+
+	kind := common.CompositeKind(encodedKind)
+
+	// Fields
+
+	fieldsContent, err := d.decoder.DecodeRawBytes()
```
fxamacker, line 1705: What do you think about using `d.decoder.DecodeRawBytesZeroCopy` here?

fxamacker (commented):
Nice work!  Performance improvements look great!

runtime/interpreter/decode.go
SupunS: Oh, yes, I used it at the other place, but missed out using it here. Good catch!

runtime/interpreter/value.go
SupunS: That's a good idea. However, the catch is, fields also could get indirectly modified. e.g: Updating a field of a nested composite value. For the outer value, it would only see it as a field-retrieval. i.e: `foo.bar.someField = "updated"`
Maybe if we had something like a 'update callback', that would have helped.

runtime/interpreter/encoding_test.go
```diff
@@ -85,6 +85,11 @@ func testEncodeDecode(t *testing.T, test encodeDecodeTest) {
 	} else {
 		require.NoError(t, err)
 
+		// Make sure the content is built.
+		if composite, ok := decoded.(*CompositeValue); ok {
+			composite.ensureFieldsLoaded()
```
turbolent, line 90: Maybe just call the public `Fields()` function instead, which also calls this private function

runtime/interpreter/decode.go
```diff
@@ -1736,3 +1625,186 @@ func (d *DecoderV4) decodeCapabilityStaticType() (StaticType, error) {
...
+	v.kind = kind
+	v.fieldsContent = fieldsContent
+
+	return nil
+}
+
+// decodeCompositeFields decodes fields from the byte content and updates the composite value.
+//
+func decodeCompositeFields(v *CompositeValue, content []byte) error {
+	d, err := NewByteDecoder(content, v.Owner, CurrentEncodingVersion, v.decodeCallback)
```
turbolent, line 1741: The version of the data might not be the current version. The version of the encoded data should be saved in the deferred composite and then used here

runtime/interpreter/decode.go
```diff
@@ -1736,3 +1625,186 @@ func (d *DecoderV4) decodeCapabilityStaticType() (StaticType, error) {
...
+// decodeCompositeMetaInfo decodes the meta info from the byte content and updates the composite value.
+// Meta info includes:
+//    - location
+//    - qualifiedIdentifier
+//    - kind
+//
+// This also extracts out the fields raw content and cache it separately inside the value.
+//
+func decodeCompositeMetaInfo(v *CompositeValue, content []byte) error {
+	d, err := NewByteDecoder(content, v.Owner, CurrentEncodingVersion, v.decodeCallback)
```
turbolent, line 1638: The version of the data might not be the current version. The version of the encoded data should be saved in the deferred composite and then used here

turbolent (commented):
Awesome work! 👏

runtime/interpreter/decode.go
```diff
@@ -1736,3 +1627,186 @@ func (d *DecoderV4) decodeCapabilityStaticType() (StaticType, error) {
...
+	v.kind = kind
+	v.fieldsContent = fieldsContent
+
+	return nil
+}
+
+// decodeCompositeFields decodes fields from the byte content and updates the composite value.
+//
+func decodeCompositeFields(v *CompositeValue, content []byte) error {
+	d, err := NewByteDecoder(content, v.Owner, CurrentEncodingVersion, nil)
```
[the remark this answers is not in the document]
SupunS: Right, that's a good catch!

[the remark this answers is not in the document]
SupunS: Added with 16fd84f

runtime/interpreter/encode.go
```diff
@@ -978,6 +978,27 @@ func (e *Encoder) encodeCompositeValue(
...
 	deferrals *EncodingDeferrals,
 ) error {
+
+	// If the value is not loaded, dump the raw content as it is.
+	if v.content != nil {
+		// Encode CBOR tag number
+		err := e.enc.EncodeRawBytes([]byte{
+			// tag number
+			0xd8, cborTagCompositeValue,
+		})
```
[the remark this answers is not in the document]
SupunS: Added in 5a08ea0

runtime/interpreter/encode.go
```diff
@@ -1002,44 +1021,54 @@ func (e *Encoder) encodeCompositeValue(
...
-		fieldName := pair.Key
-
-		// Encode field name as fields array element
-		err := e.enc.EncodeString(fieldName)
+	// If the fields are not loaded, dump the raw fields content as it is.
+	if v.content != nil {
+		err := e.enc.EncodeRawBytes(v.fieldsContent)
+		if err != nil {
+			return err
+		}
```
fxamacker, line 1036: Should `v.content` be changed to `v.fieldsContent`?

SupunS: Yes, that was a silly mistake. And a test case was missing. Fixed it in c4c2c8b.

fxamacker (approved):
Nice work!  The benchmark results are great!  👍 🚀

turbolent (approved):
Awesome work! 👏 ⛵

PR #903, opened 2021-05-14
PublicKeyValue stringer fix

janezpodhostnik:
ref: https://github.com/dapperlabs/flow-internal/issues/1515

## Description

After the `PublicKeyValue.publicKey` field was changed to a computed field (https://github.com/onflow/cadence/pull/879), it no longer appeared if logged (`.String()`). It is desired that the public key is still a part of  `.String()` output.

There is also a test suite in the FVM that relies on this: https://github.com/onflow/flow-go/blob/4c6b6f6916d8f9d3ce347f946e6db1b84ed77c89/fvm/accounts_test.go#L1007

I added a custom `PublicKeyValue` stringer that includes the `publicKey` field.

Because stringers were not copied when a composite was copied I also included that change.

______


runtime/interpreter/value.go
```diff
@@ -6173,6 +6173,7 @@ func (v *CompositeValue) Copy() Value {
 		// NOTE: new value has no owner
 		Owner:    nil,
 		modified: true,
+		stringer: v.stringer,
```
SupunS, line 6176: Thanks for catching this!

runtime/interpreter/value.go
```diff
@@ -8179,6 +8180,17 @@ func NewPublicKeyValue(
 	// Validate the public key, and initialize 'isValid' field.
 	publicKeyValue.Fields.Set(sema.PublicKeyIsValidField, validationFunction(publicKeyValue))
 
+	// Public key value to string should include the key even though it is a computed field
+	stringerFields := NewStringValueOrderedMap()
+	stringerFields.Set(sema.PublicKeyPublicKeyField, publicKey.Copy())
+	publicKeyValue.Fields.Foreach(func(key string, value Value) {
+		stringerFields.Set(key, value)
+	})
```
SupunS, line 8188: :+1:

SupunS (approved):
Looks good! Thanks, @janezpodhostnik for the fix!

runtime/interpreter/value.go
```diff
@@ -8179,6 +8180,17 @@ func NewPublicKeyValue(
...
+	stringer := func() string {
+		return formatComposite(string(publicKeyValue.TypeID()), stringerFields)
+	}
+	publicKeyValue.stringer = stringer
```
turbolent, line 8192: Could we maybe delay the construction of the map required for the formatting until it is actually is needed? Also, I think the `ForEach` can be inlined. Maybe something like this:

```suggestion
	// Public key value to string should include the key even though it is a computed field
	var *stringerFields *StringValueOrderedMap
	publicKeyValue.stringer = func() string {
		if stringerFields == nil {
			stringerFields = NewStringValueOrderedMap()
			stringerFields.Set(sema.PublicKeyPublicKeyField, publicKey.Copy())
			publicKeyValue.Fields.Foreach(stringerFields.Set)
		}
		return formatComposite(string(publicKeyValue.TypeID()), stringerFields)
	}
```

runtime/interpreter/value_test.go
```diff
@@ -366,13 +367,17 @@ func TestSetOwnerCompositeCopy(t *testing.T) {
...
+		return "random string"
+	}
 
 	compositeCopy := composite.Copy().(*CompositeValue)
 	valueCopy, _ := compositeCopy.Fields.Get(fieldName)
 
 	assert.Nil(t, compositeCopy.GetOwner())
 	assert.Nil(t, valueCopy.GetOwner())
 	assert.Equal(t, &oldOwner, value.GetOwner())
+	assert.Equal(t, composite.stringer(), compositeCopy.stringer())
```
turbolent, line 380: Good catch, thanks for fixing and testing this 👍

turbolent (approved):

turbolent:
Thank you for adding this @janezpodhostnik 👍

runtime/interpreter/value.go
janezpodhostnik: stringerFields.Set is a `func (key string, value Value) (oldValue Value, present bool)` Foreach expects `func (key string, value Value)` so inlining like that wont work

turbolent: Ah, right! 👍

runtime/interpreter/value.go
```diff
@@ -6241,6 +6241,7 @@ func (v *CompositeValue) Copy() Value {
 		// NOTE: new value has no owner
 		Owner:           nil,
 		modified:        true,
+    stringer: v.stringer,
```
turbolent, line 6244:
```suggestion
		stringer: v.stringer,
```

runtime/interpreter/value.go
```diff
@@ -6241,6 +6241,7 @@ func (v *CompositeValue) Copy() Value {
 		// NOTE: new value has no owner
 		Owner:           nil,
 		modified:        true,
+		stringer: v.stringer,
```
turbolent, line 6244:
```suggestion
		stringer:        v.stringer,
```

turbolent (approved):

turbolent:
@janezpodhostnik resolved the conflicts, ready to be merged

runtime/interpreter/value_test.go
```diff
@@ -2089,3 +2097,46 @@ func TestNumberValue_Equal(t *testing.T) {
...
+
+		sigAlgo := func() *CompositeValue {
+			fields := NewStringValueOrderedMap()
+			fields.Set(sema.EnumRawValueFieldName, UInt8Value(sema.SignatureAlgorithmECDSA_secp256k1.RawValue()))
+
+			return &CompositeValue{
+				qualifiedIdentifier: sema.SignatureAlgorithmType.QualifiedIdentifier(),
+				kind:                sema.SignatureAlgorithmType.Kind,
+				fields:              fields,
+			}
```
turbolent, line 2125: @SupunS is this construction of a non-deferred composite value here correct?

turbolent (approved):

runtime/interpreter/value_test.go
SupunS: This looks OK. We could also use the constructor, but that takes additional two params (owner and location).

SupunS (approved):

PR #907, opened 2021-05-16
Track origin's occurrences

turbolent:
## Description

Record occurrences in origins, so that it is possible to get all occurrences from one occurrence.

For example, in the following program:

```kotlin
struct Counter {
    var n: Int

    init() {
        self.n = 0
    }

    fun inc() {
        self.n = self.n + 1
    }
}
```

It is possible to get from one of the occurrences of field `n` in functions `inc` to the origin, the field declaration on line 2, and from there get all occurrences (the one in the initializer, and the other two in the `inc` function.

Also:
- Fix the origin range for field declarations
- Add a function to get all occurrences for a position (multiple checker passes (re)declare occurrences)

______

### DIFF
--- a/languageserver/server/server.go
+++ b/languageserver/server/server.go
@@ -873,7 +873,7 @@ var containerCompletionItems = []*protocol.CompletionItem{
 // Completion is called to compute completion items at a given cursor position.
 //
 func (s *Server) Completion(
-	conn protocol.Conn,
+	_ protocol.Conn,
 	params *protocol.CompletionParams,
 ) (
 	items []*protocol.CompletionItem,
--- a/languageserver/server/server.go
+++ b/languageserver/server/server.go
@@ -873,7 +873,7 @@ var containerCompletionItems = []*protocol.CompletionItem{
 // Completion is called to compute completion items at a given cursor position.
 //
 func (s *Server) Completion(
-	_ protocol.Conn,
+	conn protocol.Conn,
 	params *protocol.CompletionParams,
 ) (
 	items []*protocol.CompletionItem,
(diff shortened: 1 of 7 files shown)


SupunS (approved):

runtime/sema/occurrences.go
```diff
@@ -96,6 +97,13 @@ func (o *Occurrences) Put(startPos, endPos ast.Position, origin *Origin) {
 		occurrence.EndPos,
 	)
 	o.tree.Put(interval, occurrence)
+	origin.Occurrences = append(
+		origin.Occurrences,
+		ast.Range{
+			StartPos: startPos,
+			EndPos:   endPos,
+		},
+	)
```
turbolent, line 106: The origin may be nil, e.g. for built-in members:

```suggestion
	if origin != nil {
		origin.Occurrences = append(
			origin.Occurrences,
			ast.Range{
				StartPos: startPos,
				EndPos:   endPos,
			},
		)
	}
```

languageserver/server/server.go
```diff
@@ -873,7 +873,7 @@ var containerCompletionItems = []*protocol.CompletionItem{
 // Completion is called to compute completion items at a given cursor position.
 //
 func (s *Server) Completion(
-	_ protocol.Conn,
+	conn protocol.Conn,
```
SupunS, line 876: Probably doesn't need this change?

SupunS (approved):

languageserver/server/server.go
turbolent: Good catch! I used it to log something client-side, but yeah, not needed anymore 👍

PR #912, opened 2021-05-17
Add a reference to PublicKey in the crypto docs

### LINKED ISSUE #872: Adjust Crypto contract

SupunS:
Closes #872

______

### DIFF
--- a/docs/language/crypto.md
+++ b/docs/language/crypto.md
@@ -100,6 +100,7 @@ pub fun test main() {
     )
 }
 ```
+Refer the [PublicKey](../accounts#publickey) section for more details on the creation and the validity of public keys.
 
 The API of the Crypto contract is:
 


turbolent (approved):

turbolent:
Thanks for adding this 👍

PR #915, opened 2021-05-17
Check storability when value is written (before it is encoded)

turbolent:
## Description

Values which are saved directly using the storage API (using the `save` function) or indirectly by mutating a stored value, may be non-storable, as the types `AnyStruct` and `AnyResource` are currently storable. The static check, the argument of the type parameter for the function `save` to be a subtype of the type `Storable` only ensures that the types are definitely not-storable – we still need a run-time check that the stored value is storable.

Currently the storage of non-storable values is already indirectly done by the encoding/preparation function, which returns an error when a value cannot be encoded:

https://github.com/onflow/cadence/blob/ca1e6400714f4b412688add73ddaf72b945c9562/runtime/interpreter/encode.go#L409-L413

However, this eventually is treated further up as an implementation error (and correctly so: the error indicates e.g. we forgot to implement encoding for an actually storable value), so leads to a crash.

Implement this check properly / handle this gracefully by adding a function `IsStorable` for all types conforming to `interpreter.Value`, and using it when writing all stored values.

This is a port of the fix in the internal repo:
- dapperlabs/cadence-internal#7
- Part of dapperlabs/cadence-internal#8

In addition, this PR contains follow-up fixes, a port of cadence-internal#9:

- 1adf652: An error return value was added to `runtimeStorage.writeCached`.
  Check the error in the tests

- 3af7759: The function `Value.IsStorable` was added. It was implemented to return false for the types `AuthAccountValue`, `PublicAccountValue` and `AuthAccountContractsValue`.
  In Cadence v0.15 these Go types got replaced and are now all backed by `CompositeValue`.
  Re-apply the notion of these and other built-in composite values not being non-storable

______

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -6851,6 +6851,14 @@ func (v *CompositeValue) ConformsToDynamicType(
 
 func (v *CompositeValue) IsStorable() bool {
 
+	// Composite value's of native/built-in types are not storable for now
+	if v.Location() == nil {
+		return false
+	}
+
+	// If this composite value has a field which is non-storable,
+	// then the composite value is not storable.
+
 	// TODO: only check decoded fields
 	//   and assume still encoded fields are storable?
 
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -62,6 +62,7 @@ type Value interface {
 	StaticType() StaticType
 	ConformsToDynamicType(interpreter *Interpreter, dynamicType DynamicType, results TypeConformanceResults) bool
 	String(results StringResults) string
+	IsStorable() bool
 }
 
 // ValueIndexableValue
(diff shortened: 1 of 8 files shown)


runtime/interpreter/value.go
```diff
@@ -6668,6 +6779,28 @@ func (v *CompositeValue) ConformsToDynamicType(
...
 
+func (v *CompositeValue) IsStorable() bool {
+
+	// Composite value's of native/built-in types are not storable for now
+	if v.Location() == nil {
+		return false
+	}
+
+	// If this composite value has a field which is non-storable,
+	// then the composite value is not storable.
```
SupunS, line 6790: We check for the existence of computed fields when checking for storability of a [CompositeType](https://github.com/onflow/cadence/blob/master/runtime/sema/type.go#L3172:L3174) at the checker. Do we need to do the same here?

SupunS (approved):
Looks good! Just a question on composite fields.

runtime/interpreter/value.go
turbolent: Oh, we shouldn't consider computed fields when determining storability, because computed fields are not stored, right?

SupunS: Idea was, at the time of decoding, we wouldn't know how to compute the field. 
We could still allow it I guess, but then we'll have to special case the types with computed fields to, and specifically construct a value of that type. e.g: if the type is `PublicKey`, use the `NewPublicKey` constructor, etc.

I think this has no effect for the moment, since only built-in types can have computed fields, and all built-in types are not storable anyway.

turbolent: Makes sense! I'll leave it as-is for now, as we don't readily have access to `hasComputedMembers` of the type here.

Still, I had a look at the type's storability function, and noticed it was checking the composite kind. I've added that in 85109787507fe559c51d8a74f7a72a078ef1eee8 to the value (see the comment why contracts are included)

SupunS: oh right, that's a good catch!

For the `hasComputedMembers`, instead maybe we could check the existence of the `ComputedFields` in the composite value. The `hasComputedMembers` is a de-facto for the same. I added that flag only because there is no separation between computed vs normal fields at the type level, like in the value.

PR #921, opened 2021-05-18
Improve reference values' dynamic type, static type, copy, equal, and conformance functions

### LINKED ISSUE #836: Improve EphemeralReferenceValue and StorageReferenceValue

turbolent:
Closes #836 

______

### DIFF
--- a/runtime/tests/interpreter/string_test.go
+++ b/runtime/tests/interpreter/string_test.go
@@ -42,13 +42,13 @@ func TestInterpretRecursiveValueString(t *testing.T) {
 	require.NoError(t, err)
 
 	require.Equal(t,
-		`{"mapRef": {"mapRef": ...}}`,
+		`{"mapRef": {"mapRef": {"mapRef": ...}}}`,
 		mapValue.String(interpreter.StringResults{}),
 	)
 
 	require.IsType(t, &interpreter.DictionaryValue{}, mapValue)
 	require.Equal(t,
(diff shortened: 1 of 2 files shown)


runtime/tests/interpreter/string_test.go
```diff
@@ -42,13 +42,13 @@ func TestInterpretRecursiveValueString(t *testing.T) {
...
 	require.Equal(t,
-		`{"mapRef": {"mapRef": ...}}`,
+		`{"mapRef": {"mapRef": {"mapRef": ...}}}`,
 		mapValue.String(interpreter.StringResults{}),
 	)
 
 	require.IsType(t, &interpreter.DictionaryValue{}, mapValue)
 	require.Equal(t,
-		`{"mapRef": ...}`,
+		`{"mapRef": {"mapRef": ...}}`,
```
SupunS, line 51: Just curious as to what causes the additional iteration when converting to string? :thinking:

SupunS (approved):
Nice!

runtime/tests/interpreter/string_test.go
turbolent: It was late last evening, but I also was curious why this changed, so I quickly looked into it: 
This is caused by the change of the `Copy` function of the `EphemeralReferenceValue`, which used to return the same pointer, but now allocated a new one. I changed it to be more defensive, though technically it's not needed, as there are no mutating functions

PR #925, opened 2021-05-19
Add deferred decoding support for dictionary values

### LINKED ISSUE #844: Add deferred decoding support for Cadence values
### Issue To Be Solved

When the cadence values are read from the storage, all the raw values are decoded to go-types. However, sometimes not all these values are used. e.g: 
- In composite types, some fields may never be used.
- In array types, not all members be accessed.

So always decoding all the values is a bit of extra work, and could be a waste of resources if they are never used.

### Suggest A Solution
The idea is to delay the decoding until a value actually used.

This could be done by:
- When encoding a value, write an additional property `content-length`, before writing the content.
  e.g:` <type-tag>  <content-length>  <content>`
- When reading back from the storage, read the `content-length` and load the raw-bytes for the `content`. Do not try to decode.
- So the cadence-value will internally hold the raw bytes.
- Decode the raw bytes on-demand, if the values  actually used in some way (e.g: access a field, logging,  etc.). This again is a shallow loading. Only loaded up to a one-level down.
- Once decoded, cache the decoded values.

This could be extended to:
- [x] Composite values
- [x] Array values
- [ ] Map values

### LINKED ISSUE #924: Add deferred decoding for dictionary values
### Issue To Be Solved

Details: #844

SupunS:
Closes #844
Closes #924

## Description

$subject.

### Performance
Performance for a dictionary with `String` keys and `String` values (`{String:String}`) of size 100:
```
name                                             old time/op    new time/op    delta
DictionaryDeferredDecoding/Simply_decode-12        57.4µs ±19%     3.2µs ± 1%  -94.48%  (p=0.016 n=5+4)
DictionaryDeferredDecoding/Get_value-12            50.2µs ± 9%    56.1µs ±29%     ~     (p=0.222 n=5+5)
DictionaryDeferredDecoding/Re-encode_decoded-12    14.0µs ±10%     1.8µs ± 7%  -87.20%  (p=0.008 n=5+5)

name                                             old alloc/op   new alloc/op   delta
DictionaryDeferredDecoding/Simply_decode-12        27.9kB ± 0%     0.5kB ± 0%  -98.19%  (p=0.008 n=5+5)
DictionaryDeferredDecoding/Get_value-12            27.9kB ± 0%    28.1kB ± 0%   +0.75%  (p=0.008 n=5+5)
DictionaryDeferredDecoding/Re-encode_decoded-12    6.47kB ± 0%    3.46kB ± 0%  -46.54%  (p=0.008 n=5+5)

name                                             old allocs/op  new allocs/op  delta
DictionaryDeferredDecoding/Simply_decode-12           622 ± 0%         6 ± 0%  -99.04%  (p=0.008 n=5+5)
DictionaryDeferredDecoding/Get_value-12               623 ± 0%       626 ± 0%   +0.48%  (p=0.008 n=5+5)
DictionaryDeferredDecoding/Re-encode_decoded-12      17.0 ± 0%      10.0 ± 0%  -41.18%  (p=0.008 n=5+5)
```
______

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -19,6 +19,7 @@
 package interpreter
 
 import (
+	"bytes"
 	"encoding/binary"
 	"encoding/hex"
 	"fmt"
@@ -6956,8 +6957,8 @@ func (v *CompositeValue) ensureFieldsLoaded() {
 // DictionaryValue
 
 type DictionaryValue struct {
-	Keys     *ArrayValue
-	Entries  *StringValueOrderedMap
+	keys     *ArrayValue
+	entries  *StringValueOrderedMap
 	Owner    *common.Address
 	modified bool
 	// Deferral of values:
@@ -7002,8 +7003,8 @@ func NewDictionaryValueUnownedNonCopying(keysAndValues ...Value) *DictionaryValu
 	}
 
 	result := &DictionaryValue{
-		Keys:    NewArrayValueUnownedNonCopying(),
-		Entries: NewStringValueOrderedMap(),
+		keys:    NewArrayValueUnownedNonCopying(),
+		entries: NewStringValueOrderedMap(),
 		// NOTE: new value has no owner
 		Owner:                  nil,
 		modified:               true,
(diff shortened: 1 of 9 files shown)


turbolent (approved):
Looks good! Nice work 👍

fxamacker (commented):
Nice work! :+1:  Amazing benchmarks again!  :rocket:

runtime/interpreter/value.go
```diff
@@ -7090,11 +7177,13 @@ func (v *DictionaryValue) IsModified() bool {
...
+
+	if v.keys.IsModified() {
 		return true
 	}
 
-	for pair := v.Entries.Oldest(); pair != nil; pair = pair.Next() {
+	for pair := v.entries.Oldest(); pair != nil; pair = pair.Next() {
 		value := pair.Value
 		if value.IsModified() {
 			return true
```
fxamacker, line 7189: Because of deferred decoding, is it possible for `v.keys.IsModified()` or `value.IsModified()` return true without DictionaryValue being decoded?

SupunS: Right, returning `isModified` as false, if the value is not loaded at all, makes sense.
```go
if v.content != nil {
    return false
}
```
We could do the same for composites and arrays as well.

Good suggestion! :1st_place_medal:

SupunS:
@fxamacker @turbolent Can you please have a look at the latest changes as well?

runtime/interpreter/value.go
```diff
@@ -715,7 +715,12 @@ func (v *ArrayValue) IsModified() bool {
 		return true
 	}
 
-	for _, value := range v.Elements() {
+	// If the elements are not loaded, that implies they are not modified.
+	if v.content != nil {
+		return false
+	}
```
turbolent, line 721: 👌

runtime/interpreter/value.go
```diff
@@ -6852,11 +6868,13 @@ func (v *CompositeValue) ConformsToDynamicType(
 
 func (v *CompositeValue) IsStorable() bool {
 
+	v.ensureMetaInfoLoaded()
```
turbolent, line 6871: 👍

turbolent (approved):
@SupunS your last commit looks good! nice optimization and fix 👍

PR #931, opened 2021-05-21
Test contract member intialization

### LINKED ISSUE #312: Qualifying a contract-level struct errors
### Instructions

### Problem

Qualifying a contract-level struct with its contract name gives an error. 

Example:

```
contract A {
   pub struct B {}
}
```

I'd expect to be able to do `A.B()`. I have a playground to share.

### Steps to Reproduce

https://play.onflow.org/1e0a4413-21f8-4302-b7d4-6705bd78dc5d

Starting at line 245, if you quality `HealthStats` as `HealthCrossingCore.HealthStat`, Cadence returns an error.

### Acceptance Criteria

You should be able to qualify the contract struct.

### Context

This is a non-blocking issue because I don't necessarily have to qualify the struct with the contract name.

turbolent:
Closes #312

## Description

Accessing fields before they are initialized does not cause a crash anymore since #768.

Add additional tests for this particular use-case, accessing contract fields before they are initialized.

When adding the tests I realized that in tests, the initializer of a contract is not called.
This is due to an optimization we added a while ago to only lazily load a contract value.
As most tests rely on contract declarations and thus contract initializers being evaluated eagerly, I added this to the parseCheckAndInterpret helper function that is used by all interpreter tests.

______

### DIFF
--- a/runtime/tests/interpreter/contract_test.go
+++ b/runtime/tests/interpreter/contract_test.go
@@ -33,7 +33,7 @@ func TestInterpretContractUseBeforeInitializationComplete(t *testing.T) {
 
 		t.Parallel()
 
-		_, _ = parseCheckAndInterpretWithOptions(t,
+		_, err := parseCheckAndInterpretWithOptions(t,
 			`
               contract C {
 
@@ -57,6 +57,7 @@ func TestInterpretContractUseBeforeInitializationComplete(t *testing.T) {
 				},
 			},
 		)
+		require.NoError(t, err)
 	})
 
 	t.Run("constructor of nested type, first qualified ", func(t *testing.T) {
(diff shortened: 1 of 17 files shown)


runtime/tests/interpreter/contract_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+func TestInterpretContractUseBeforeInitializationComplete(t *testing.T) {
+
+	t.Parallel()
+
+	t.Run("constructor of nested type, always qualified ", func(t *testing.T) {
+
+		t.Parallel()
+
+		_, _ = parseCheckAndInterpretWithOptions(t,
```
SupunS, line 36: I guess the return error must be asserted for 'no error'?

SupunS (approved):

runtime/tests/interpreter/contract_test.go
turbolent: Good catch! I had added the error return value and then forgot to actually check it for this test case 👍 

Fixed in e02872921db62d2eac8816f7b9aee291f0cd46cc

PR #935, opened 2021-05-23
Fix parsing of negative fractional fixed-point strings

### LINKED ISSUE #934: Negative fractional fixed-point string values (e.g. "-0.1") are parsed as positive
### Problem

Strings representing negative Fix64 values are parsed as positive if the value is between -1 and 0

### Steps to Reproduce

When using the flow-cli or Flow Playground to pass the value "-0.1" to the following script, the return value is "0.1":
```
pub fun main(ff: Fix64): Fix64 {
  return ff
}
```

https://play.onflow.org/59e4485e-ce00-4b85-9e09-b542b6109315?type=script&id=6cf9432e-d219-4537-8b9d-2aeabbc0f07b

### Acceptance Criteria

All negative fixed-point values are parsed as negative

mikeylemmon:
Closes #934 

## Description

A bug in `parseFixedPoint` was causing the function to return positive values for strings like "-0.1" because it was only checking if the integer component is less than zero. This PR fixes the bug by ensuring the return value is negative if the string begins with '-'.

______

### DIFF
--- a/fixedpoint/parse.go
+++ b/fixedpoint/parse.go
@@ -104,7 +104,7 @@ func parseFixedPoint(v string) (
 
 	scale = uint(len(fractionalStr))
 
-	negative = false
+	negative = len(v) > 0 && v[0] == '-'
 
 	integer, ok := new(big.Int).SetString(integerStr, 10)
 	if !ok {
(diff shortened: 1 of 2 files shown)


Kay-Zee:
Repro for the problem. https://play.golang.org/p/URyXIX-HTU_S

turbolent:
@mikeylemmon Thank you for reporting this bug and also fixing it! 👍 

I'll add some tests.

turbolent (approved):

PR #939, opened 2021-05-24
Add an AST walking function

turbolent:
## Description

Add a function to walk the AST, similar to Go's function: https://golang.org/src/go/ast/walk.go

The implementation of Walk for the AST elements currently only visits child elements. A future version could also visit more information, such as identifiers, types, etc.

The current implementation is enough to determine the parent element of each element.

______

### DIFF
--- a/runtime/ast/inspect.go
+++ b/runtime/ast/inspect.go
@@ -1,0 +1,33 @@
+/*
+ * Cadence - The resource-oriented smart contract programming language
+ *
+ * Copyright 2019-2021 Dapper Labs, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ast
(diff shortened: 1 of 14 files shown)


runtime/ast/inspect.go
```diff
@@ -0,0 +1,33 @@
...
+	if f(element) {
+		return f
+	}
+
+	return nil
+}
+
+func Inspect(element Element, f func(Element) bool) {
+	Walk(inspector(f), element)
+}
```
SupunS, line 33:
```suggestion
func Inspect(element Element, f inspector) {
	Walk(f, element)
}
```

runtime/ast/walk.go
```diff
@@ -0,0 +1,62 @@
...
+// It starts by calling walker.Walk(element);
+// If the returned walker is nil,
+// child elements are not walked.
+// If the returned walker is not-nil,
+// then Walk is invoked recursively on this returned walker
+// for each of the non-nil children of the element,
+// followed by a call of Walk(nil) on the returned walker.
+//
+func Walk(walker Walker, element Element) {
+	if walker = walker.Walk(element); walker == nil {
```
SupunS, line 35: Could the initial `walker` parameter be nil?

SupunS (approved):
Nice addition to the visitors!

runtime/ast/inspect.go
turbolent: The `inspector` type is private, so I think it's better to keep it as is (just like in Go's codebase)

runtime/ast/walk.go
turbolent: Good question! I'll document that it can't, only the walker returned from a walker

turbolent: Added in 7aaf2da8e8e458c415d718651ddd88dda30871fd

PR #948, opened 2021-05-26
Optimize qualified identifier creation

SupunS:
## Description

A micro-optimization:
- Simply return the identifier, if there are no containers.
- Avoid appending by creating the identifier-array at the lowest level of the chain (when the size is known).
- Calculate the buffer size on the fly, and grow the string-builder buffer at once.

```
name                                                  old time/op    new time/op    delta
QualifiedIdentifierCreation/One_level-12              24.6ns ± 4%     2.0ns ± 1%   -91.96%  (p=0.008 n=5+5)
QualifiedIdentifierCreation/Three_levels-12            197ns ± 4%     105ns ± 7%   -46.93%  (p=0.008 n=5+5)

name                                                  old alloc/op   new alloc/op   delta
QualifiedIdentifierCreation/One_level-12               8.00B ± 0%     0.00B       -100.00%  (p=0.008 n=5+5)
QualifiedIdentifierCreation/Three_levels-12             120B ± 0%       64B ± 0%   -46.67%  (p=0.008 n=5+5)

name                                                  old allocs/op  new allocs/op  delta
QualifiedIdentifierCreation/One_level-12                1.00 ± 0%      0.00       -100.00%  (p=0.008 n=5+5)
QualifiedIdentifierCreation/Three_levels-12             4.00 ± 0%      2.00 ± 0%   -50.00%  (p=0.008 n=5+5)

```

______

### DIFF
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -79,16 +79,7 @@ func containerTypeNames(typ Type, level int) (typeNames []string, bufSize int) {
 		typeName = typedContainerType.Identifier
 		containerType = typedContainerType.ContainerType
 	default:
-		switch typ {
-		case PublicAccountType:
-			typeName = string(typedContainerType.ID())
-			containerType = nil
-		case AuthAccountType:
-			typeName = string(typedContainerType.ID())
-			containerType = nil
-		default:
-			panic(errors.NewUnreachableError())
-		}
+		panic(errors.NewUnreachableError())
 	}
 
 	typeNames, bufSize = containerTypeNames(containerType, level+1)
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -32,36 +32,26 @@ import (
 )
 
 func qualifiedIdentifier(identifier string, containerType Type) string {
-
-	// Gather all identifiers: this, parent, grand-parent, etc.
-
-	identifiers := []string{identifier}
-
-	for containerType != nil {
-		switch typedContainerType := containerType.(type) {
-		case *InterfaceType:
-			identifiers = append(identifiers, typedContainerType.Identifier)
-			containerType = typedContainerType.ContainerType
-		case *CompositeType:
-			identifiers = append(identifiers, typedContainerType.Identifier)
-			containerType = typedContainerType.ContainerType
-		default:
-			switch containerType {
-			case PublicAccountType:
-				identifiers = append(identifiers, string(typedContainerType.ID()))
-				containerType = nil
-			case AuthAccountType:
-				identifiers = append(identifiers, string(typedContainerType.ID()))
-				containerType = nil
-			default:
-				panic(errors.NewUnreachableError())
-			}
-		}
+	if containerType == nil {
+		return identifier
 	}
 
-	// Append all identifiers, in reverse order
+	// Gather all identifiers: this, parent, grand-parent, etc.
+	const level = 0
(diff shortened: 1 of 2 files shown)


runtime/interpreter/interpreter.go
```diff
@@ -1963,18 +1963,16 @@ func (interpreter *Interpreter) declareTypeRequirement(
 	// Evaluate nested declarations in a new scope, so values
 	// of nested declarations won't be visible after the containing declaration
 
-	(func() {
```
turbolent, line 1966: I guess this refactor is OK, but I'm not 100% sure: Are we sure that the part after the initial end of this nested function (https://github.com/onflow/cadence/pull/948/files#diff-a36f730b18ca31316909bd74125003c1fbe785a13f926b4a0631e7c0cc6a51e4L1977) will not be affected by the nested declarations being in scope?

runtime/sema/type.go
```diff
@@ -32,37 +32,27 @@ import (
 )
 
 func qualifiedIdentifier(identifier string, containerType Type) string {
+	if containerType == nil {
+		return identifier
+	}
```
turbolent, line 37: 💯

runtime/sema/type.go
```diff
@@ -73,6 +63,42 @@ func qualifiedIdentifier(identifier string, containerType Type) string {
...
+		typeName = typedContainerType.Identifier
+		containerType = typedContainerType.ContainerType
+	default:
+		switch typ {
+		case PublicAccountType:
+			typeName = string(typedContainerType.ID())
+			containerType = nil
+		case AuthAccountType:
+			typeName = string(typedContainerType.ID())
+			containerType = nil
```
turbolent, line 88: Do we have to special case these specific types here? It would be great if we could avoid it

turbolent: (I see now that this was already in the original code, https://github.com/onflow/cadence/pull/948/files#diff-c01765841e64b85306258f5ee4daff2bdc5b9e9e107167f55a6dba4162e8bdb8L50-L55)

runtime/sema/type_test.go
```diff
@@ -529,3 +529,71 @@ func TestBeforeType_Strings(t *testing.T) {
 		beforeType.QualifiedString(),
 	)
 }
+
+func TestQualifiedIdentifierCreation(t *testing.T) {
```
turbolent, line 533: Thank you for adding a unit-test 👍 

Could you please add another case where the container type is nil? (for coverage)

turbolent:
This is a great optimization! This code is called often, so optimizing it makes a lot of sense 👍

runtime/interpreter/interpreter.go
SupunS: yeah, probably we can leave it as it is then. This change is anyway not related to the identifier change.

runtime/sema/type.go
SupunS: Yeah, it's simply a move from the original code.

But that's a good catch. Those special cases are baggage coming from pre-refactoring days. Now, those also go under the `CompositeType` path, and no longer needed to be special-cased. Cleaned it up in 1222de9

turbolent: Great, thank you for cleaning this up! 👏

turbolent (approved):
Nice work!

PR #950, opened 2021-05-27
Cache type ID for composite types

SupunS:
## Description

Cache type ID for composite types, so that it doesn't have to recalculate again and again. This was showing up on the profiler.

Benchmarking shows it could gain up to 25% improvement for certain methods. e.g: declareNonEnumCompositeValue

_Note: These benchmark results are on top of #948. Improvement against the direct master-branch is even higher._

```
name                           old time/op    new time/op    delta
DeclareNonEnumCompositeValue   58.2µs ±40%    43.2µs ±29%  -25.77%  (p=0.040 n=10+5)

name                           old alloc/op   new alloc/op   delta
DeclareNonEnumCompositeValue    30.0kB ± 0%    27.3kB ± 0%   -9.01%  (p=0.000 n=9+5)

name                           old allocs/op  new allocs/op  delta
DeclareNonEnumCompositeValue       576 ± 0%       485 ± 0%  -15.80%  (p=0.000 n=10+5)
```

______

### DIFF
--- a/runtime/sema/authaccount_contracts.go
+++ b/runtime/sema/authaccount_contracts.go
@@ -71,7 +71,7 @@ var AuthAccountContractsType = func() *CompositeType {
 
 func init() {
 	// Set the container type after initializing the `AuthAccountContractsType`, to avoid initializing loop.
-	AuthAccountContractsType.ContainerType = AuthAccountType
+	AuthAccountContractsType.SetContainerType(AuthAccountType)
 }
 
 const authAccountContractsTypeAddFunctionDocString = `
--- a/runtime/sema/authaccount_type.go
+++ b/runtime/sema/authaccount_type.go
@@ -600,7 +600,7 @@ var authAccountKeysTypeRevokeFunctionType = &FunctionType{
 
 func init() {
 	// Set the container type after initializing the AccountKeysTypes, to avoid initializing loop.
-	AuthAccountKeysType.ContainerType = AuthAccountType
+	AuthAccountKeysType.SetContainerType(AuthAccountType)
 }
 
 const AccountKeysTypeName = "Keys"
--- a/runtime/sema/check_composite_declaration.go
+++ b/runtime/sema/check_composite_declaration.go
@@ -853,7 +853,7 @@ func (checker *Checker) explicitInterfaceConformances(
 ) []*InterfaceType {
 
 	var interfaceTypes []*InterfaceType
-	seenConformances := map[TypeID]bool{}
+	seenConformances := map[*InterfaceType]bool{}
 
 	for _, conformance := range declaration.Conformances {
 		convertedType := checker.ConvertType(conformance)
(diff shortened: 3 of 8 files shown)


turbolent:
Thank you for looking into optimizing this Supun!

We had tried this before, in #658, but then saw issues on Canary and reverted it in #670. Do we know now what caused the problems and can be sure we won't repeat the same issue? 

Also, like in #658, we need to make this thread-safe, as the types are potentially accessed concurrently in the host environment (program caching)

SupunS:
Oh, I didn't know it was reverted :thinking: 
I'll look a bit more into that and see if I find something. Thanks for the heads-up!

turbolent:
@SupunS here is the whole thread discussing the problem and the revert: 
https://axiomzen.slack.com/archives/C01E28D03FV/p1615274743014800?thread_ts=1615273189.014700&cid=C01E28D03FV

turbolent:
@SupunS We only quickly reverted it to get that release deployed, but I just want to say that it is worth it and we should still try to get this in. 

I think the fix would be to invalidate the cached ID when the container type changes. However, there might be a whole chain of container types, so if a parent's container type changes, also its children's IDs need to be updated (recursively). We currently only have a pointer from the child to the container, I think we should make the `ContainerType` field private and add a setter function (e.g. `SetContainerType`), which also informs the children (recursively) about the change, so they can update their IDs.

This isn't really a problem right now / does not exist at the moment, as only one level of containment is possible: non-contract types may be contained in a contract. However, in the future we might change this, so we shouldn't rely

SupunS:
Right, that makes sense.

I think I found the issue: Seems like it's happening for nested interfaces. 
Here at [check_composite_declaration.go#L864](https://github.com/onflow/cadence/blob/master/runtime/sema/check_composite_declaration.go#L864) `interfaceType.ID()` is called and there it caches the typeID. And later on at [check_interface_declaration.go#L262](https://github.com/onflow/cadence/blob/master/runtime/sema/check_interface_declaration.go#L262), `ContainerType` of the interface is updated. 

I will add a setter for `ContainerType` and invalidate the cache upon updates, like you suggested. Thanks for the tips!

turbolent:
@SupunS Nice work, thank you for finding the spot where this happens. I agree, in this case it should be OK. I'm still a bit worried that there are potentially other cases or future cases where it is not OK. Could we potential mark the read of the ID in the case where it is OK (because it is only used temporarily) as "safe if invalidated later", and other cases would be "panic if the ID changes"? This could be just a boolean on the cache which is checked by the setter. Happy to sync on details

SupunS:
Yeah, that's a good safety measure.

It seems this is the only place where the ID is cached before setting the `containerType`. We can avoid that as well, by using a `map[*InterfaceType]bool{}` assuming types are singleton?

Then we can simply panic if the ID is cached always. Added this change with 2e22c7d

In the future, if we have a valid use-case where the ID is needed, then maybe we can add a method to retrieve the ID without caching. But I feel we don't need to add it at this point, so that such usages are discouraged automatically. WDYT?

turbolent:
I guess when I wrote the code and used the type ID my thinking was that the conformances might be type aliases at some point, but I guess we can use `map[*InterfaceType]bool{}` for now, yes.

2e22c7d looks good!

turbolent (approved):
Nice work!

PR #951, opened 2021-05-27
Always check arguments and record the argument types in the elaboration

turbolent:
## Description

When the invoked expression has an invalid type or is not a function type, then the arguments were not checked and their types were not recorded in the elaboration.

This is sufficient, as an error is reported, but it is better to report all errors in the program and not reveal them gradually. Also, this information is useful in the language server to propose code actions to fix the error (e.g. declare a missing function).

______


SupunS (approved):
Looks good!

PR #957, opened 2021-05-28
Improve type inferrence for binary expressions

turbolent:
## Description

I wrote a Base64 encoding function for Cadence here:
https://forum.onflow.org/t/base64-encode-in-cadence/1915/7

This currently requires a lot of static type annotations and casts and could be much simplified through the following observations and changes.

- Currently the contextually expected type (if any) is only used for arithmetic binary expressions.
  
  However, bitwise binary expressions have the same property that the resulting type must be compatible with the operands.

- Currently the left expression and the right expression are type checked individually, and each side is provided with the contextually expected type (if any).
  
  However, the both sides of arithmetic, logical, etc. expressions must have the same type, so if no contextually expected type is present, use the left type as the expected type when checking the right side.

- The contextually expected type might be less specific. For example, the indexing type for arrays and strings is `Integer`. This causes an expression like `let x = 1 as UInt8; "abc"[x + 1]` to fail to type check.

With the following changes, this portion of the example program can be significantly simplified:

  - Before:

    ```kotlin
    let n = (UInt32(data[i]) << 16 as UInt32)
        + (UInt32(data[i + 1 as Int]) << 8 as UInt32)
        + UInt32(data[i + 2 as Int])

    let n1 = (n >> 18 as UInt32) & 63 as UInt32
    let n2 = (n >> 12 as UInt32) & 63 as UInt32
    let n3 = (n >> 6 as UInt32) & 63 as UInt32
    let n4 = n & 63 as UInt32
    ```

  - After:
  
    ```kotlin
    let n = UInt32(data[i]) << 16
        + UInt32(data[i + 1]) << 8
        + UInt32(data[i + 2])

    let n1 = (n >> 18) & 63
    let n2 = (n >> 12) & 63
    let n3 = (n >> 6) & 63
    let n4 = n & 63
    ```

Also, improve the tests by
- Removing the wrapper functions
- Adding assertions for the resulting types

______

### DIFF
--- a/runtime/sema/check_binary_expression.go
+++ b/runtime/sema/check_binary_expression.go
@@ -105,7 +105,9 @@ func (checker *Checker) VisitBinaryExpression(expression *ast.BinaryExpression) 
 		// let character = string[index + 1]
 		// ```
 
-		if expectedType == nil || IsProperSubType(leftType, expectedType) {
+		if expectedType == nil ||
+			(leftType != expectedType && IsProperSubType(leftType, expectedType)) {
+
 			expectedType = leftType
 		}
 
--- a/runtime/sema/check_binary_expression.go
+++ b/runtime/sema/check_binary_expression.go
@@ -46,7 +46,9 @@ func (checker *Checker) VisitBinaryExpression(expression *ast.BinaryExpression) 
 	// is done based on the operation kind.
 
 	var expectedType Type
-	if operationKind == BinaryOperationKindArithmetic {
+	switch operationKind {
+	case BinaryOperationKindArithmetic,
+		BinaryOperationKindBitwise:
 		expectedType = UnwrapOptionalType(checker.expectedType)
 	}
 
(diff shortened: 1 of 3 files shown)


runtime/sema/check_binary_expression.go
```diff
@@ -73,9 +75,42 @@ func (checker *Checker) VisitBinaryExpression(expression *ast.BinaryExpression)
...
 
-		// Visit the expression, with contextually expected type. Use the expected type
-		// only for inferring wherever possible, but do not check for compatibility.
+		// Visit the expression, with contextually expected type.
+		// Use the expected type only for inferring wherever possible,
+		// but do not check for compatibility.
 		// Compatibility is checked separately for each operand kind.
+		//
+		// If there is no contextually expected type,
+		// then expect the right type to have the type of the left side.
```
SupunS, line 84: This enforcement seems sensible :+1:

runtime/sema/check_binary_expression.go
```diff
@@ -73,9 +75,42 @@ func (checker *Checker) VisitBinaryExpression(expression *ast.BinaryExpression)
...
+		// ```
+		//
+		// ```
+		// let string = "this is a test"
+		// let index = 1 as UInt8
+		// let character = string[index + 1]
+		// ```
+
+		if expectedType == nil ||
+			(IsSubType(leftType, expectedType) && !leftType.Equal(expectedType)) {
```
SupunS, line 109: It might be possible to get the `leftType` as `nil` for invalid codes. Maybe can add a nil check here, or can make sure to return an `invalidType` if the returning type is nil at `(checker *Checker) visitExpressionWithForceType`?

SupunS: What do you think of swapping the sides of the second condition and doing the `!leftType.Equal(expectedType)` check first, as it could be the more prominent case and also the less expensive one?

runtime/sema/check_binary_expression.go
```diff
@@ -73,9 +75,42 @@ func (checker *Checker) VisitBinaryExpression(expression *ast.BinaryExpression)
...
+		//
+		// ```
+		// let x = 1 as UInt8
+		// let y = x + 1
+		// ```
+		//
+		// Also, if there is a contextually expected type,
+		// but the left type is a subtype and more specific (i.e not the same),
+		// then use it instead as the expected type for the right type.
+		// For example, this allows declarations like the following to type-check:
```
SupunS, line 95: :+1:

SupunS (approved):

runtime/sema/check_binary_expression.go
SupunS: and maybe an additional check of `leftType != expectedType` for an early exit?

turbolent: Subtype and equality testing should be roughly equal in complexity, they are very "similar" (rescursing down into the type structure and comparing).

However, I looked into this and realized that the equality check is needed, as subtyping also consideres equal types subtypes. So what we want here is the proper subtype. I've refactored this in d089fd0939ec20f1dd9f68b99613cab35d33ef1f, so we only have to perform one equality check instead of two.

I also added your exit early suggestion in 9b9f1234a904629bf9d62bd988ca7478aefd1d89: Good observation that two pointer-equal types are never proper subtypes.

PR #906, opened 2021-05-15
Handle recursive values (due to references) in string function

turbolent:
Closes #349

______

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #969, opened 2021-06-02
Bring back fmt.Stringer implementation for interpreter.Value

turbolent:
## Description

In #906 we refactored the string formatting of interpreter values so it supports recursive structures (due to references). This PR then removed support for `fmt.Stringer`. 

This is problematic, because many parts of the code, most prominently errors, rely on values implementing `String()` implicitly by using `fmt` functions with the `%s` format directive.
Now that support for `String()` was removed from values, error messages were rendered incorrectly.

Most notably, address values are just byte slices, and Go's default formatting, which will be used when the value does not implement `fmt.Stringer`, will encode the bytes directly as-is. For example, the byte slice `[]byte{0x1, 0x2}` results in the Go string `"\x01\x02"` – which is not valid UTF-8.
This results in problems in downstream dependencies like flow-go, which requires strings to be valid UTF-8 (e.g. for Protocol Buffer encoding).

Bringing back support for `fmt.Stringer` thus also fixes many error messages that are currently incorrectly formatted.

______

### DIFF
--- a/runtime/interpreter/errors_test.go
+++ b/runtime/interpreter/errors_test.go
@@ -1,7 +1,7 @@
 /*
  * Cadence - The resource-oriented smart contract programming language
  *
- * Copyright 2021 Dapper Labs, Inc.
+ * Copyright 2019-2021 Dapper Labs, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/runtime/interpreter/errors_test.go
+++ b/runtime/interpreter/errors_test.go
@@ -1,0 +1,40 @@
+/*
+ * Cadence - The resource-oriented smart contract programming language
+ *
+ * Copyright 2021 Dapper Labs, Inc.
+ *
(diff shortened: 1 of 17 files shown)


runtime/interpreter/value.go
```diff
@@ -6751,7 +6858,7 @@ func formatComposite(typeId string, fields *StringValueOrderedMap, results Strin
 				Value string
 			}{
 				Name:  fieldName,
-				Value: value.String(results),
+				Value: value.RecursiveString(results),
```
SupunS, line 6861: nice idea! :+1:

runtime/interpreter/errors_test.go
```diff
@@ -0,0 +1,40 @@
+/*
+ * Cadence - The resource-oriented smart contract programming language
+ *
+ * Copyright 2021 Dapper Labs, Inc.
```
SupunS, line 4:
```suggestion
 * Copyright 2019-2021 Dapper Labs, Inc.
```

SupunS (approved):
Looks good! :clap:

PR #972, opened 2021-06-03
Validate UTF-8 compatibility in string value constructor

### LINKED ISSUE #966: Validate strings to be UTF-8
### Issue To Be Solved

Strings in Cadence are assumed to be valid UTF-8. This is the case when string literals are parsed, but there are other means of constructing a string value which also need to be ensured to be valid.

### Suggest A Solution

- Validate input is UTF-8 in `cadence.NewString`, so that the preparation of a value client-side already fails there

### Context

https://axiomzen.slack.com/archives/CEXKUH74G/p1622587881117800

SupunS:
Closes #966

______


turbolent (approved):
Looks good!

Issue #717, opened 2021-03-22
Allow import of native composite types


turbolent:
### Issue To Be Solved

PR #713 made native composite types, such as `PublicKey` and `SigningAlgorithm` non-storable.
However, this also made them non-importable.

Allow import of native composite types.

### Suggested Solution

- [ ] Add the notion of "importable types", just like the notion of "exportable types", and allow native composite types (ones without a location) to be imported
- [ ] Ensure native composite types can be imported:
  - [ ] `PublicKey`
  - [ ] `SigningAlgorithm`
  - [ ] `HashAlgorithm`
- [ ] Switch the entry point argument checks (script arguments, transaction arguments) from checking storability to importability

turbolent:
@SupunS I had initially written that all types should be importable, which is obviously wrong (e.g. `AuthAccount` shouldn't), I adjusted the description

PR #973, opened 2021-06-03
Allow native composite types to be passed as script arguments

SupunS:
Work towards #717

## Description

- Introduces an `IsImportable` method to check the importability as script arguments.
- Add a pre-determined `importable` flag for native composite types
  - Some native composite types need to be constructed in a special manner, because their methods are coming from the host-enviornment.
  - It requires some additional work to allow them to be constructed via script arguments (not included in this PR)

______


runtime/sema/type.go
```diff
@@ -3354,6 +3389,35 @@ func (t *CompositeType) IsStorable(results map[*Member]bool) bool {
...
+	// Use the pre-determined flag for native types
+	if t.Location == nil {
+		return t.importable
+	}
+
+	// Only structures, resources, and enums can be imported
+
+	switch t.Kind {
+	case common.CompositeKindStructure,
+		common.CompositeKindResource,
```
turbolent, line 3402: Resources may not be imported!

runtime/sema/type.go
```diff
@@ -5440,6 +5560,10 @@ func (*CapabilityType) IsExternallyReturnable(_ map[*Member]bool) bool {
 	return true
 }
 
+func (t *CapabilityType) IsImportable(_ map[*Member]bool) bool {
+	return true
```
turbolent, line 5564: Capabilities may not be imported! 

```suggestion
	return false
```

After this PR is merged, #712 should become easier to implement though 👍

runtime/sema/anyresource_type.go
```diff
@@ -31,4 +31,6 @@ var AnyResourceType = &SimpleType{
 	Equatable: false,
 	// The actual returnability of a value is checked at run-time
 	ExternallyReturnable: true,
+	// The actual importability is checked at runtime
+	Importable: true,
```
turbolent, line 35: Resources may not be imported!

```suggestion
	Importable: true,
```

runtime/sema/type.go
SupunS: I kept that as is, because it was also the current behavior. But yeah makes sense to restrict it.

Added with 9381434

runtime/sema/type.go
```diff
@@ -3354,6 +3389,34 @@ func (t *CompositeType) IsStorable(results map[*Member]bool) bool {
 	return true
 }
 
+func (t *CompositeType) IsImportable(results map[*Member]bool) bool {
+	// Use the pre-determined flag for native types
+	if t.Location == nil {
+		return t.importable
+	}
+
+	// Only structures, resources, and enums can be imported
```
turbolent, line 3398:
```suggestion
	// Only structures and enums can be imported
```

turbolent (approved):
Nice work!

PR #981, opened 2021-06-04
Auto update flow-go github action

janezpodhostnik:
Closes https://github.com/dapperlabs/flow-internal/issues/1517

## Description

A github action that creates a PR in flow-go to update cadence version every time a PR is merged into master in cadence. The PRs in flow-go are based on the master branch or the latest `auto-cadence-upgrade/*` branch if it exists

I tested this on private repos, but it might still need some tweaking on the `go get && make tidy` part.
______

### DIFF
--- a/.github/workflows/sync-flow-go.yml
+++ b/.github/workflows/sync-flow-go.yml
@@ -61,7 +61,7 @@ jobs:
         uses: peter-evans/create-pull-request@v3
         with:
           token: ${{ secrets.REMOTE_REPO_PAT }}
-          commit-message: update cadence to commit ${{ github.event.pull_request.head.sha }}
+          commit-message: update Cadence to commit ${{ github.event.pull_request.head.sha }}
           title: Auto-Update ${{ env.LOCAL_REPO_NAME }} - ${{ github.event.pull_request.title }}
           body: |
               Auto generated PR to update Cadence version.
--- a/.github/workflows/sync-flow-go.yml
+++ b/.github/workflows/sync-flow-go.yml
@@ -1,0 +1,70 @@
+# Open a PR in flow-go when a PR in cadence is merged
+# The PR in flow-go is based on the last opened `auto-cadence-upgrade/*` if it exists, otherwise on the master branch
+
+
+# Only run on pull requests merged to master
+on:
+  pull_request:
+    branches:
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-flow-go.yml
```diff
@@ -0,0 +1,70 @@
...
+  pull_request:
+    branches:
+      - master
+    types: [closed]
+
+jobs:
+  sync-flow-go:
+    runs-on: ubuntu-latest
+    # the PR could have been closed otherwise. Only run if it has actually ben merged
+    if: github.event.pull_request.merged == true
```
turbolent, line 16: Maybe this can be simplified to just run on each push to master?

But then I guess the events at the bottom (`github.event.pull_request`) aren't available / would need to change?


```suggestion
  push:
    branches:
      - master
      
jobs:
  sync-flow-go:
    runs-on: ubuntu-latest
```

.github/workflows/sync-flow-go.yml
```diff
@@ -0,0 +1,70 @@
...
+      
+      - name: Make tidy
+        run: make tidy
+
+      # create the pull request
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
+        with:
+          token: ${{ secrets.REMOTE_REPO_PAT }}
+          commit-message: update cadence to commit ${{ github.event.pull_request.head.sha }}
```
turbolent, line 64:
```suggestion
          commit-message: update Cadence to commit ${{ github.event.pull_request.head.sha }}
```

turbolent (approved):
Great work! Looking forward to using this 👍

.github/workflows/sync-flow-go.yml
janezpodhostnik: The convenience of having a reference of a Cadence PR in the generated flow-go PR would be lost... Not sure that would be the best approach.

Are there pushes to master that happen without PRs?

janezpodhostnik:
@turbolent when you have some time, could you add the `secrets.REMOTE_REPO_PAT` (or we could rename it to `FLOW_GO_PAT`) to the repo secrets?

turbolent:
Added the secret, should be good to go now!

PR #995, opened 2021-06-10
Make `PublicKey` type importable

### LINKED ISSUE #717: Allow import of native composite types

SupunS:
Closes #717

______

### DIFF
--- a/runtime/account_keys_test.go
+++ b/runtime/account_keys_test.go
@@ -193,43 +193,6 @@ func TestRuntimeAccountKeyConstructor(t *testing.T) {
 	assert.Contains(t, err.Error(), "cannot find variable in this scope: `AccountKey`")
 }
 
-func TestRuntimeImportAccountAPITypes(t *testing.T) {
-
-	t.Parallel()
-
-	nextTransactionLocation := newTransactionLocationGenerator()
-
-	for _, ty := range []sema.Type{
-		sema.AccountKeyType,
-		sema.PublicKeyType,
-	} {
-
-		rt := NewInterpreterRuntime()
-
-		script := []byte(fmt.Sprintf(`
-			pub fun main(key: %s) {}
-		`, ty.String()))
-
-		runtimeInterface := &testRuntimeInterface{}
-
-		_, err := rt.ExecuteScript(
-			Script{
-				Source: script,
-			},
-			Context{
-				Interface: runtimeInterface,
-				Location:  nextTransactionLocation(),
-			},
-		)
-
-		require.Error(t, err)
-		assert.IsType(t, Error{}, err)
-		runtimeErr := err.(Error)
-
-		assert.IsType(t, &ScriptParameterTypeNotImportableError{}, runtimeErr.Err)
-	}
-}
-
 func TestRuntimeStoreAccountAPITypes(t *testing.T) {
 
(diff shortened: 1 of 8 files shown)


runtime/account_keys_test.go
```diff
@@ -193,43 +193,6 @@ func TestRuntimeAccountKeyConstructor(t *testing.T) {
 	assert.Contains(t, err.Error(), "cannot find variable in this scope: `AccountKey`")
 }
 
-func TestRuntimeImportAccountAPITypes(t *testing.T) {
```
SupunS, line 196: Same scenario is already covered in `TestScriptParameterTypeValidation`. So this is a redundant test case.

runtime/convertValues.go
```diff
@@ -465,3 +500,78 @@ func importCompositeValue(
...
+			if !ok {
+				panic(fmt.Errorf(
+					"cannot import value of type '%s'. invalid value for field '%s': %v",
+					sema.PublicKeyType,
+					fieldName,
+					value,
+				))
+			}
+
+			publicKeyValue = arrayValue
```
turbolent, line 526: Do we have to validate the array value itself here too, or is this done before / after?

runtime/convertValues.go
```diff
@@ -465,3 +500,78 @@ func importCompositeValue(
...
+			if !ok {
+				panic(fmt.Errorf(
+					"cannot import value of type '%s'. invalid value for field '%s': %v",
+					sema.PublicKeyType,
+					fieldName,
+					value,
+				))
+			}
+
+			signAlgoValue = compositeValue
```
turbolent, line 539: Do we have to validate the composite value itself here too, or is this done before / after?

runtime/convertValues.go
```diff
@@ -465,3 +500,78 @@ func importCompositeValue(
...
+
+		case sema.PublicKeyIsValidField:
+			// 'isValid' field set by the user must be ignored.
+			// This is calculated when creating the public key.
+			return
```
turbolent, line 544: 👍

turbolent (commented):
Nice! Just have one question regarding the validation of the field values

runtime/convertValues.go
SupunS: Yes, that gets validated in downstream, when validating the public-key also through the `ConformsToDynamicType` check.

Added a test here: f047877. Also improved public-key validation to gracefully handle errors.

runtime/convertValues.go
SupunS: same for the signature algorithm as well.

turbolent (approved):
Looks good! I have a follow-up PR in #1002

PR #1004, opened 2021-06-13
Embed markdown templates to docgen tool at compile time

SupunS:
## Description

Embed the template content in binary form during compile time.

This is required when exporting and using the tool in a non-Go environment. e.g: WASM. The previous package aware path does not work for such cases.

Uses go-embed, hence need to use go-1.16.*

Also make the docgen tool accept a generic "Template Provider", so that the tool is not hard-bound to the output format. This makes it is easy to customize the output by pluging-in different template-providers in the future (e.g: html, etc)
______


tools/docgen/gen/templates/markdown-templates.go
```diff
@@ -0,0 +1,45 @@
...
+
+package templates
+
+import (
+	"embed"
+	"path"
+)
+
+//go:embed markdown
+var mdTemplateFiles embed.FS
```
turbolent, line 27: 👍

turbolent (approved):

PR #1007, opened 2021-06-15
Add dynamic type importability check for arguments

### LINKED ISSUE #979: Restricted types allow non-importable types to be imported as script arguments
### Instructions

It is possible to send a value of a non-importable composite type `T` as a script argument, if the script parameter type is a restricted type `{R}`, and `T` implements `R`.

### Steps to Reproduce

```go
// argument is restricted to be an implementation of Foo interface.
pub fun main(arg: {Foo}) {
}

pub struct interface Foo {
}

// FooImpl conforms to the interface, but is not an importable type due to its field.
//
pub struct interface FooImpl: Foo {
    pub var funcTypedField: (():Void)

    init() {
        self.funcTypedField = fun() {}
    }
}
```

### Suggested Solution
A similar use-case is: having the parameter type as `AnyStruct` and sending a struct value that is non-importable. But this is currently restricted at:
https://github.com/onflow/cadence/blob/3386e0fb952f4f81b8118842dd9a27c0731de09a/runtime/interpreter/interpreter.go#L760-L768

A simple solution would be to add a similar validation for restricted types as well.

However, that may restrict certain use-cases from being achieved. A less-restrictive/more-user-friendly approach would be to validate using the dynamic type of the value.

SupunS:
Closes #979
______


runtime/interpreter/dynamictype.go
```diff
@@ -40,24 +41,40 @@ type MetaTypeDynamicType struct{}
 
 func (MetaTypeDynamicType) IsDynamicType() {}
 
+func (MetaTypeDynamicType) IsImportable() bool {
+	return false
```
turbolent, line 45: Should be true, maybe just refer to use the same value as in `sema`, here and in the other implementations, i.e.:

```suggestion
	return sema.MetaType.Importable
```

runtime/convertValues.go
```diff
@@ -488,7 +488,9 @@ func importCompositeValue(
 		case sema.SignatureAlgorithmType, sema.HashAlgorithmType:
 			// continue in the normal path
 		default:
-			panic(fmt.Errorf("cannot import value of type %s", qualifiedIdentifier))
+			// Continue and return a dummy value.
+			// It will be validated later through the 'IsStorable' dynamic check.
+			// TODO: return an error
```
turbolent, line 493: I have another PR that does refactor this to return an error, for now, can we just keep the panic and not silently accept everything?

turbolent (commented):
Nice! Just a couple things.

Would be nice to get this into the next release / Mainnet spork

turbolent (approved):

PR #1013, opened 2021-06-16
Don't check missing program

### LINKED ISSUE #996: Cannot start Cadence Server: Got error in github.com/onflow/cadence/languageserver@v0.15.2/server/server.go:507
### Problem

While trying to start Cadence server in VS Code, I am getting error as follows:
```
github.com/onflow/cadence/languageserver@v0.15.2/server/server.go:1363 +0x3c5
github.com/onflow/cadence/languageserver/server.(*Server).checkAndPublishDiagnostics(0xc000207170, 0x168a8f8, 0xc00052e5d0, 0xc00004a640, 0x4d, 0xc00026e000, 0x3f0e, 0x3ff0000000000000)
	github.com/onflow/cadence/languageserver@v0.15.2/server/server.go:507 +0xa5
[Error - 4:41:36 pm] Connection to server is erroring. Shutting down server.

```
**Probably, the error has occured in github.com/onflow/cadence/languageserver@v0.15.2/server/server.go:507**

### Context

I am writing smart contracts in Cadence and I am unable to figure out the syntax errors. Also debugging has become very cumbersome as the Cadence server is not running.

turbolent:
Closes #996

## Description

https://github.com/Bhawana15/Kitty-Auction-House/blob/main/cadence/contracts/Auction.cdc is a good stress test, it uncovered a few issues:

- If an import cannot be resolved in the language server, the empty program was still checked, leading to a crash
- The AST walking implementations for function declarations and return statements didn't handle nil-cases properly, leading to a crash

After applying these changes, the language server is able to check the program and does not crash anymore.

______

 
### DIFF
--- a/runtime/ast/function_declaration.go
+++ b/runtime/ast/function_declaration.go
@@ -55,7 +55,9 @@ func (d *FunctionDeclaration) Accept(visitor Visitor) Repr {
 func (d *FunctionDeclaration) Walk(walkChild func(Element)) {
 	// TODO: walk parameters
 	// TODO: walk return type
-	walkChild(d.FunctionBlock)
+	if d.FunctionBlock != nil {
+		walkChild(d.FunctionBlock)
+	}
 }
 
 func (*FunctionDeclaration) isDeclaration() {}
--- a/languageserver/server/server.go
+++ b/languageserver/server/server.go
@@ -1794,6 +1794,11 @@ func (s *Server) getDiagnostics(
 						if err != nil {
 							return nil, err
 						}
+						if importedProgram == nil {
+							return nil, &sema.CheckerError{
+								Errors: []error{fmt.Errorf("cannot import %s", importedLocation)},
+							}
+						}
 
 						importedChecker, err = checker.SubChecker(importedProgram, importedLocation)
 						if err != nil {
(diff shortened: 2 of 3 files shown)


runtime/ast/function_declaration.go
```diff
@@ -55,7 +55,9 @@ func (d *FunctionDeclaration) Accept(visitor Visitor) Repr {
 func (d *FunctionDeclaration) Walk(walkChild func(Element)) {
 	// TODO: walk parameters
 	// TODO: walk return type
-	walkChild(d.FunctionBlock)
+	if d.FunctionBlock != nil {
```
SupunS, line 58: is it possible to move the nil checks to a common location (e.g: [ast/walk.go#L42](https://github.com/onflow/cadence/blob/master/runtime/ast/walk.go#L42)), just to be safe?

SupunS (approved):

runtime/ast/function_declaration.go
turbolent: That's a great idea and I had actually tried this, then spent a good half hour wondering why it isn't working: as the walk function parameter has type `func(Element)`, calling it with a concrete element type, e.g. `*FunctionBlock` and which is `nil`, won't be visible/detectable in the implementation of the function that is passed, in `ast.Walk`. This is because it is now a *typed nil*, and not the untyped nil used in the equality check (e.g. `== nil`). 

Here's a Go Playground that demonstrates the issue ("Go feature" ...): https://play.golang.org/p/m-dElD3ckXE

This is fairly common Go gotcha, and I wish Go didn't have this "feature"

PR #1022, opened 2021-06-18
Directly declare base values in activation without interpreter

turbolent:
## Description

The declaration of the base values does not depend on an interpreter, so refactor the declaration of the base values to be directly performed on the base activation.

______


SupunS (approved):
Nice!
I was tempted to do the same :smile: , but I wasn't sure whether the interpreter would be useful later on.

PR #1027, opened 2021-06-21
Use common super-type for expression type inferring

### LINKED ISSUE #61: Find join of types, improve inference
## Context

Implement a function to return the least common supertype.

This improves type checking for:
- Conditionals: instead of just returning the left type (type of true branch) and requiring the right type (else branch) to be the same
- Array literals
- Dictionary literals

See ["Types and Programming Languages, 16.3 Joins and Meets"](https://books.google.ca/books?id=ti6zoAC9Ph8C&lpg=PA219&pg=PA218#v=onepage&q&f=false)

In the following example `y` should be inferred to `String?`:

```
let x: String? = ""
let y = true ? nil : x 
```

## Definition of Done

- Sema: 
  - [ ] Add function to find join of two or more types
  - [ ] Use join function for conditionals
  - [ ] Use join function for array literals
  - [ ] Use join function for dictionary literals
- [ ] Tests
- [ ] Documentation

### LINKED ISSUE #1239: Inferred type for ternary operator with nil leads to unexpected `mismatched types` error
### Problem

This program is semantically correct:
```
let cond = false
let message = ""
let messageOptional : String? =
  cond ? nil : message
```

However, it fails with Cadence's type inference on the fourth line on the `message` token, giving a `mismatched types` error, reporting that `expected 'Never?', got 'String'`.

Presumably this is because the type of the right-hand side of the initialization expression of variable declaration `messageOptional` is inferred on its own, with respect to `nil`, to `Never?`. Changing the order of the ternary operator does not affect the error, so presumably the type inference is performed solely on the strength of the `nil` token.

Perhaps the type of an expression on the right-hand side should be synthesized along with type information from the left-hand side, if available (bidirectional type checking).

As a workaround, forcing the type of the `nil` branch, like so, manages to convince Cadence's type checker.
```
let cond = false
let message = ""
let messageOptional/* : String?*/ =
  cond ? (nil as String?) : message
```

### Steps to Reproduce

See above.

SupunS:
Closes #61
Closes #1239

## Description

This PR adds support for inferring the common supertype for:
- Array literals
- Dictionary literals
- Conditional expression

## Breaking Changes
With these changes, array literals and dictionary literals no longer use the first element to infer the type for the literal. Thus, if it requires a literal to have a certain specific type, then explicit type annotation or casting must be used.
e.g:
Old way
```kotlin
var x = [1 as UInt8, 6, 35]
```

New Way
```kotlin
var x = [1, 6, 35] as [UInt8]
```

______


runtime/sema/check_variable_declaration.go
```diff
@@ -233,41 +231,6 @@ func (checker *Checker) recordVariableDeclarationRange(
 	)
 }
 
-func (checker *Checker) checkVariableDeclarationUsability(declaration *ast.VariableDeclaration) {
```
SupunS, line 236: This check now happens at the expression level for arrays/dictionaries. Because:
- we need to infer the type even in non-variable-delectation scenarios. e.g: the expression `[].getType()` should have a type inferred for the empty array.
- Even if the array is not empty, there could be non-inferrable scenarios
  e.g: `[<some_struct>, <some_resource>]`

runtime/sema/check_array_expression.go
```diff
@@ -70,22 +74,26 @@ func (checker *Checker) VisitArrayExpression(expression *ast.ArrayExpression) as
...
-		elementType = NeverType
-	}
+		// Contextually expected type is not available.
+		// Therefore, find the least common supertype of the elements.
+		elementType = LeastCommonSuperType(argumentTypes...)
+
+		if elementType == InvalidType {
+			checker.report(
+				&TypeAnnotationRequiredError{
+					Cause: "cannot infer type from array literal: ",
```
turbolent, line 89:
```suggestion
					Cause: "cannot infer type from array literal",
```

Or was the idea to add something after the colon?

runtime/sema/check_array_expression.go
```diff
@@ -70,22 +74,26 @@ func (checker *Checker) VisitArrayExpression(expression *ast.ArrayExpression) as
 
 		checker.checkVariableMove(value)
 		checker.checkResourceMoveOperation(value, valueType)
-
-		// infer element type from first element
-		// TODO: find common super type?
-		if elementType == nil {
-			elementType = valueType
-		}
```
turbolent, line 78: 👌

runtime/sema/check_dictionary_expression.go
```diff
@@ -56,25 +59,27 @@ func (checker *Checker) VisitDictionaryExpression(expression *ast.DictionaryExpr
...
-		keyType = NeverType
-	}
-
-	if valueType == nil {
-		valueType = NeverType
+	if keyType == nil && valueType == nil {
+		// Contextually expected type is not available.
+		// Therefore, find the least common supertype of the keys and values.
+		keyType = LeastCommonSuperType(keyTypes...)
+		valueType = LeastCommonSuperType(valueTypes...)
```
turbolent, line 70: 👌

runtime/sema/check_conditional.go
```diff
@@ -62,26 +62,32 @@ func (checker *Checker) VisitIfStatement(statement *ast.IfStatement) ast.Repr {
...
-				ActualType:   elseType,
-				Range:        ast.NewRangeFromPositioned(expression.Else),
-			},
-		)
+	if thenType.Equal(elseType) {
+		return thenType
 	}
 
-	return resultType
+	return LeastCommonSuperType(thenType, elseType)
```
turbolent, line 90: 👌

runtime/sema/check_dictionary_expression.go
```diff
@@ -106,7 +111,7 @@ func IsValidDictionaryKeyType(keyType Type) bool {
 		return keyType.Kind == common.CompositeKindEnum
 	default:
 		switch keyType {
-		case NeverType, BoolType, CharacterType, StringType, MetaType:
+		case NeverType, BoolType, CharacterType, StringType, MetaType, InvalidType:
```
turbolent, line 114: Why is this needed here? Is the idea here that we allow the invalid type to avoid the additional error that the type is not a valid key type, as there was already an error report for the invalid type?

runtime/sema/check_variable_declaration.go
turbolent: Nice!

runtime/sema/type_tags.go
```diff
@@ -376,7 +376,16 @@ func LeastCommonSuperType(types ...Type) Type {
...
 
-	return findCommonSuperType(join, types...)
+	supertype := findCommonSuperType(join, types...)
+
+	// 'Any' is a valid type, since it's the supertype of all types.
+	// However, in the context of checker, 'Any' is not a possible
+	// type for expressions. Hence, return 'InvalidType'.
+	if supertype == AnyType {
+		return InvalidType
+	}
```
turbolent, line 386: 👍 

Maybe flag this as being important, and also describe why `Any` is not a valid type:

```suggestion
	// NOTE: Important:
	// 'Any' is a valid inferred type, since it's the supertype of all types.
	// However, in the context of checker, 'Any' is not a possible type for expressions,
	// as it mixes value-kinded and resource-kinded types, leading to undefined behaviour.
	// Hence, return 'InvalidType'.
	if supertype == AnyType {
		return InvalidType
	}
```

runtime/tests/checker/arrays_dictionaries_test.go
```diff
@@ -137,20 +137,7 @@ func TestCheckInvalidDictionaryKeys(t *testing.T) {
 
 	errs := ExpectCheckerErrors(t, err, 1)
 
-	assert.IsType(t, &sema.TypeMismatchError{}, errs[0])
-}
-
-func TestCheckInvalidDictionaryValues(t *testing.T) {
```
turbolent, line 143: Maybe don't remove this test case but turn it into a valid one, now that it is 🎉

runtime/tests/checker/arrays_dictionaries_test.go
```diff
@@ -950,19 +937,6 @@ func TestCheckInvalidDictionarySubtyping(t *testing.T) {
 	assert.IsType(t, &sema.TypeMismatchError{}, errs[0])
 }
 
-func TestCheckInvalidArrayElements(t *testing.T) {
```
turbolent, line 953: Maybe don't remove this test case but turn it into a valid one

runtime/tests/checker/conditional_test.go
```diff
@@ -59,45 +59,90 @@ func TestCheckInvalidConditionalExpressionElse(t *testing.T) {
...
+			},
+			xType,
+		)
+	})
+
+	t.Run("chained optional", func(t *testing.T) {
+		t.Parallel()
+
+		checker, err := ParseAndCheck(t, `
+            let x = true ? Int8(1) : (false ? Int(5) : nil)
```
turbolent, line 134: Nice! 👌

runtime/tests/checker/for_test.go
```diff
@@ -68,7 +68,9 @@ func TestCheckForEmpty(t *testing.T) {
       }
     `)
 
-	assert.NoError(t, err)
+	assert.Error(t, err)
+	errs := ExpectCheckerErrors(t, err, 1)
+	assert.IsType(t, &sema.TypeAnnotationRequiredError{}, errs[0])
```
turbolent, line 73: Why is this producing an error now? If the problem is that there is no contextual information: We could expect this to be `[AnyStruct]`, as iteration in this construct is currently always over arrays and iteration over resources is not supported.

Then this change here and in https://github.com/onflow/cadence/pull/1027/files#diff-ec994e4f2ba53bd43232ba59cf3f9f4c83f36bed61c097893e6d2b5c89eb25b3L202 could be undone.

runtime/tests/checker/string_test.go
```diff
@@ -281,7 +281,7 @@ func TestCheckStringEncodeHex(t *testing.T) {
 	t.Parallel()
 
 	checker, err := ParseAndCheck(t, `
-        let x = String.encodeHex([1 as UInt8, 2, 3, 0xCA, 0xDE])
+        let x = String.encodeHex([1, 2, 3, 0xCA, 0xDE] as [UInt8])
```
turbolent, line 284: Could this now be even further simplified by removing the explicit type annotation?

```suggestion
        let x = String.encodeHex([1, 2, 3, 0xCA, 0xDE])
```

runtime/tests/checker/type_inference_test.go
```diff
@@ -771,3 +772,350 @@ func TestCheckInferenceWithCheckerErrors(t *testing.T) {
...
+			},
+			{
+				name: "values with nil",
+				code: `let x = ["hello", nil, nil, nil]`,
+				expectedElementType: &sema.OptionalType{
+					Type: sema.StringType,
+				},
+			},
+			{
+				name: "common interfaced values",
```
turbolent, line 816: Nice! 👌

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -9435,3 +9435,58 @@ func TestHostFunctionStaticType(t *testing.T) {
...
+
+		value, err := inter.Invoke("test")
+		require.NoError(t, err)
+
+		AssertValuesEqual(
+			t,
+			inter,
+			interpreter.TypeValue{
+				Type: interpreter.VariableSizedStaticType{
+					Type: interpreter.PrimitiveStaticTypeAnyStruct,
```
turbolent, line 9486: Totally fine and not a problem, but this behaviour is now different from the current one, where the type of the first element was used, so would have been `Int` here

runtime/tests/interpreter/metatype_test.go
```diff
@@ -657,11 +657,11 @@ func TestInterpretGetType(t *testing.T) {
 		{
 			name: "array",
 			code: `
-              let result = [].getType()
```
turbolent, line 660: Maybe leave the existing case as-is and add the new code as a new test case

runtime/tests/interpreter/string_test.go
```diff
@@ -115,7 +115,7 @@ func TestInterpretStringEncodeHex(t *testing.T) {
 
 	inter := parseCheckAndInterpret(t, `
       fun test(): String {
-          return String.encodeHex([1 as UInt8, 2, 3, 0xCA, 0xDE])
+          return String.encodeHex([1, 2, 3, 0xCA, 0xDE] as [UInt8])
```
turbolent, line 118: Like above in the checker test, maybe simplify this by removing the explicit type annotation, now that it is not necessary anymore:

```suggestion
          return String.encodeHex([1, 2, 3, 0xCA, 0xDE])
```

turbolent (commented):
Very nice work! 👏

runtime/tests/interpreter/interpreter_test.go
SupunS: Yeah, I also don't have a strong case against the old behavior. Only thing was that the old behavior is not consistent with `let x: AnyStruct = []`, where the array is empty and the type cannot be inferred by itself.

I guess it's fine to handle the two cases separately. WIll revert back to the old behavior

runtime/tests/checker/string_test.go
SupunS: oh right, I implemented this PR before adding type inference to function arguments 😄  So this wasn't working back then. But should be working now. Good catch!

runtime/sema/check_array_expression.go
SupunS: yes, this is to get a more readable error message, since `TypeAnnotationRequiredError` appends some more texts to the cause.

https://github.com/onflow/cadence/blob/81c63bfef73cd4874d05b6274dbd2402a0a0b788/runtime/sema/errors.go#L2883-L2888

runtime/sema/check_dictionary_expression.go
SupunS: ah, I had added this initially, but later moved this check to the `VisitDictionaryExpression` method itself. This is a leftover that I should have cleaned.

runtime/tests/checker/arrays_dictionaries_test.go
SupunS: Added a set of tests under `TestCheckDictionarySupertypeInference` that covers all cases. So removed this since it's a duplicate now.

runtime/tests/checker/arrays_dictionaries_test.go
SupunS: Added a set of tests under `TestCheckArraySupertypeInference` that covers all cases. So removed this since it's a duplicate now.

runtime/tests/interpreter/metatype_test.go
SupunS: This particular test checks the runtime behavior. But now that empty arrays without contextual type are giving semantic errors, this test-case is no longer valid.
I added a separate test to check the semantic error: https://github.com/onflow/cadence/blob/935fa5c9ba493e8d86c6d1a79ba8798b7e5e82a8/runtime/tests/checker/type_inference_test.go#L915-L924

runtime/tests/interpreter/interpreter_test.go
SupunS: Updated in 67e3268522d210d752cc49c0a34613fda665826b

runtime/sema/check_dictionary_expression.go
SupunS: Cleaned it up in 67e3268522d210d752cc49c0a34613fda665826b

runtime/tests/checker/for_test.go
SupunS: Yeah, the catch is we can't always pass `[AnyStruct]` as the expected type, because then it would cause other errors (i.e: imprecise inferring for arrays, etc. fixed-length arrays, etc). So we need to specially check for an empty array, and then pass `[AnyStruct]` as the expected type. Didn't really like special-casing there, but maybe it makes sense to do so 👍

SupunS: Updated in a961aae0e157156a17984b134d434240e84e67eb

turbolent: Ah, I see. Yeah, special casing is a bit unfortunate, but I think the cost is OK for keeping backwards compatibility.

runtime/tests/interpreter/interpreter_test.go
turbolent: Great, thank you!

runtime/tests/interpreter/metatype_test.go
turbolent: Ah, makes sense! 👍

turbolent (approved):
Great work! 👏

PR #1032, opened 2021-06-25
Replace `IsSubType` check with `IsSameTypeKind` check where necessary

### LINKED ISSUE #1031: 'Never' typed expressions get skipped from certain type-checks
### Problem

`Never` is the subtype of all types. Hence it gets passed through the `IsSubType` check for contexts where the `Never` type shouldn't be allowed.

### Steps to Reproduce
e.g:

(1)
```kotlin
var a: Never = "c"    // no errors reported
```

(2)
```kotlin
pub fun test(a: Never, b: Never) {
    var x: Int = a + b  // no errors reported
}
```

### Acceptance Criteria

- [ ] Differentiate checking `IsSubType` Vs checking whether a type belongs to a particular 'kind' of type (e.g: is `T` of type `Integer`?). Can introduce a separate method for the later.
- [ ] Update the usages of `IsSubType` to use the proper method out of the two, depending on the context.

SupunS:
Closes #1031 

## Description

- Introduce a method to check whether a given type 'belongs' to the same kind as a given type. 
  - Same as IsSubType, but doesn't treat `Never` as 'not' belongs to the same kind.
- Replace the usages where strictly 'same kind types' check is needed.

______

### DIFF
--- a/runtime/sema/checker.go
+++ b/runtime/sema/checker.go
@@ -601,9 +601,8 @@ func (checker *Checker) checkTypeCompatibility(expression ast.Expression, valueT
 	case *ast.FixedPointExpression:
 		unwrappedTargetType := UnwrapOptionalType(targetType)
 
-		valueTypeOK := CheckFixedPointLiteral(typedExpression, valueType, checker.report)
-
 		if IsSameTypeKind(unwrappedTargetType, FixedPointType) {
+			valueTypeOK := CheckFixedPointLiteral(typedExpression, valueType, checker.report)
 			if valueTypeOK {
 				CheckFixedPointLiteral(typedExpression, unwrappedTargetType, checker.report)
 			}
--- a/runtime/sema/checker.go
+++ b/runtime/sema/checker.go
@@ -587,19 +587,12 @@ func (checker *Checker) checkTypeCompatibility(expression ast.Expression, valueT
 	case *ast.IntegerExpression:
 		unwrappedTargetType := UnwrapOptionalType(targetType)
 
-		// If the target type is `Never`, the checks below will be performed
-		// (as `Never` is the subtype of all types), but the checks are not valid
-
-		if IsSubType(unwrappedTargetType, NeverType) {
-			break
-		}
-
-		if IsSubType(unwrappedTargetType, IntegerType) {
+		if IsSameTypeKind(unwrappedTargetType, IntegerType) {
 			CheckIntegerLiteral(typedExpression, unwrappedTargetType, checker.report)
 
 			return true
 
-		} else if IsSubType(unwrappedTargetType, &AddressType{}) {
+		} else if IsSameTypeKind(unwrappedTargetType, &AddressType{}) {
 			CheckAddressLiteral(typedExpression, checker.report)
 
 			return true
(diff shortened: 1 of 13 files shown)


runtime/sema/checker.go
```diff
@@ -608,16 +601,9 @@ func (checker *Checker) checkTypeCompatibility(expression ast.Expression, valueT
...
 		unwrappedTargetType := UnwrapOptionalType(targetType)
 
-		// If the target type is `Never`, the checks below will be performed
-		// (as `Never` is the subtype of all types), but the checks are not valid
-
-		if IsSubType(unwrappedTargetType, NeverType) {
-			break
-		}
-
 		valueTypeOK := CheckFixedPointLiteral(typedExpression, valueType, checker.report)
```
turbolent, line 604: The check above which was removed made this line conditional. Should this line now maybe be moved below new line 606, into the body of the if-statement?

turbolent (commented):
Great work!

Do we have a way of auditing the rest of the code for cases where `IsSameTypeKind` should be used instead of `IsSubType`? Also, maybe we can document a guideline for future code?

runtime/sema/checker.go
SupunS: It wouldn't change the checker results, but yeah, can avoid additional computation if the condition fails. 
Refactored: e6be2687902bd3ea6cedefa01e6a36b260f4e6ae

SupunS:
> Do we have a way of auditing the rest of the code for cases where IsSameTypeKind should be used instead of IsSubType?

I searched all the usages and replaced them where necessary. Unfortunately, this has to be done manually, can't think of a way to automate this/ audit with a test case. The general rule is:
- `IsSameType()`  - To check the assignability. e.g: Is argument type `T` is a sub-type of parameter type `R`. This is the more frequent use-case.
- `IsSAmeYTypeKind()` - To check if a type strictly belongs to a certain category. e.g: Is the expression type `T` is any of the integer types, but nothing else. Another way to check is, asking the question "if the `subType` is `Never`, should the check still pass?"

> Also, maybe we can document a guideline for future code?

I had  already added some comments to the newly added mehtod. Also added a note on the old `IsSameType()` method for the usage: 1b324ec1b79e3ce5306788eac0e97ad4420273a9

turbolent:
@SupunS Nice! Could we keep this comment/guideline somewhere in the codebase? Maybe as a comment somewhere, or in `docs`?

turbolent (approved):
Great work! 👏

runtime/sema/type.go
```diff
@@ -4708,9 +4708,17 @@ func (t *AddressType) GetMembers() map[string]MemberResolver {
...
+// using `IsSameTypeKind` method. e.g: "Is type `T` an Integer type?". Using this method
+// for the later use-case may produce incorrect results.
+//   * IsSubType()      - To check the assignability. e.g: Is argument type T is a sub-type
+//                        of parameter type R. This is the more frequent use-case.
+//   * IsSameTypeKind() - To check if a type strictly belongs to a certain category. e.g: Is the
+//                        expression type T is any of the integer types, but nothing else.
+//                        Another way to check is, asking the question of "if the subType is Never,
+//                        should the check still pass?". A common code-smell for potential incorrect
+//                        usage is, using IsSubType() method with a constant/pre-defined superType.
+//                        e.g: IsSubType(<<someType>>, FixedPointType)
```
turbolent, line 4721: ✨ 👏

PR #1033, opened 2021-06-25
Add type inferring for func-args when there are no generics

### LINKED ISSUE #512: Non-local type inference

SupunS:
Work towards #512

## Description

With this PR, the only scenario where the non-local type inferring not supported is function arguments with generics.

______

### DIFF
--- a/runtime/sema/check_invocation_expression.go
+++ b/runtime/sema/check_invocation_expression.go
@@ -518,35 +518,44 @@ func (checker *Checker) checkInvocationRequiredArgument(
 ) {
 	argument := arguments[argumentIndex]
 
-	// TODO: pass the expected type to support type inferring for parameters
-	argumentType := checker.VisitExpression(argument.Expression, nil)
+	parameter := functionType.Parameters[argumentIndex]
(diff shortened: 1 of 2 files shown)


runtime/sema/check_invocation_expression.go
```diff
@@ -518,35 +518,44 @@ func (checker *Checker) checkInvocationRequiredArgument(
...
+			}
 		}
+
+		// Check that the type of the argument matches the type of the parameter.
+
+		// TODO: remove this once type inferring support for parameters is added
+		checker.checkInvocationArgumentParameterTypeCompatibility(
+			argument.Expression,
+			argumentType,
+			parameterType,
```
SupunS, line 552: This is the same old code, only moved under the 'else' branch.

Kay-Zee (approved):

Issue #870, opened 2021-05-04
Implement `StaticType()` for arrays and dictionaries


SupunS:
### Issue To Be Solved

`getType()` function on containers (arrays. dictionaries), returns no type information.
e.g:
```kotlin
var array : [Int] = [1, 3]

log(array.getType())
```

Actual output: `Type()`

Expected Output: `Type([Int])`

### Solution

- [x] Checker: Record static types for array expressions and dictionary expressions in elaboration 
      @turbolent 
      #1034
- [x] Interpreter: 
    @turbolent 
    #1034
  - [x] Add static type field to array values and dictionary values
  - [x] Implement `StaticType()` for arrays and dictionaries
  - [x] Add static type information when evaluation array expressions and dictionary expressions (use elaboration)
- [x] Add function to walk value
  @turbolent 
  #1037
- [ ] Encoding and decoding
  - This requires a storage format change
  - [x] Add new encoder/decoder version
    @SupunS
    #1039
  - [x] Store static types 
    @SupunS
    #1035
  - [x] Read static types
    @SupunS 
    #1036
  - [x] Ensure existing stored data can be decoded using the old decoder and re-encoding using the new encoder: Arrays and dictionaries are first deferred and need to be forced to be decoded, and the new decoding functions for deferred arrays and dictionaries need to delegate to the old deferred decoding functions. + Tests
    @SupunS
    #1042
    #1046
  - [ ] Storage migration. Infer static types from stored values, by walking decoded value object graph (see item "Add function to walk value" above)
    @turbolent 
    - Similar to https://github.com/onflow/flow-go/blob/master/cmd/util/ledger/migrations/storage_v4.go
    - [x] Handle deferred values
    - [x] Migrate values even when type information is unavailable
    - [ ] Delete values that don't have type information
    - [ ] Generate report for deleted data
- [x] Import and export
  - [x] Add array type to imported array value and dictionary value
    @turbolent 
    #1034
  - [x] Add static type to imported array values and dictionary values using target parameter type
     @turbolent 
     #1038
  - [x] Infer type from the value when parameter/target type is not specific enough (e.g: target type has `AnyStruct` and the value is `[3, 5, 2]`). Could make use of: #1025, by getting static type of each element -> convert to sema type -> find super type -> convert back to static type.
   @SupunS
   Short-term solution: #1052
   Long-term solution: #1049
- [x] Dynamic types: Add static types to array values and dictionary values
  @turbolent 
  #1043
[...]

PR #1035, opened 2021-06-29
Add encoding/decoding for array value static type

SupunS:
Work towards #870

______


turbolent (commented):
Looks great!

runtime/interpreter/encoding_test.go
```diff
@@ -5061,7 +5171,10 @@ func TestEncodeDecodeDictionaryDeferred(t *testing.T) {
 
 		expected := NewDictionaryValueUnownedNonCopying(
 			// TODO: type
```
turbolent, line 5173:
```suggestion
```

runtime/interpreter/encoding_test.go
```diff
@@ -4956,7 +5055,10 @@ func TestEncodeDecodeDictionaryDeferred(t *testing.T) {
...
```
turbolent, line 5057:
```suggestion
```

runtime/interpreter/decode.go
```diff
@@ -1755,6 +1762,77 @@ func (d *DecoderV4) decodeArrayElements(path []string) ([]Value, error) {
 	return values, nil
 }
 
+func (d *DecoderV4) decodeArrayValueStaticType(valuePath []string) (StaticType, error) {
+	const expectedLength = encodedArrayValueLength
```
turbolent, line 1766: I was a bit confused by the function name and the variable declaration here, but I think I understand it now: This function decodes the start of an array value, in particular only the static type. Maybe I also misunderstood.

Maybe rename the function to make this more obvious (maybe `decodeArrayValueHead` or `decodeArrayValueStart`?), and add a comment describing the behaviour.

runtime/interpreter/decode.go
```diff
@@ -1755,6 +1762,77 @@ func (d *DecoderV4) decodeArrayElements(path []string) ([]Value, error) {
...
+func decodeArrayMetaInfo(array *ArrayValue, content []byte) error {
+	d, err := NewByteDecoder(content, array.Owner, array.encodingVersion, array.decodeCallback)
+	if err != nil {
+		return err
+	}
+
+	// TODO: store array type
+	//   Option 1: convert to sema type. - Don't have the interpreter
+	//   Option 2: Store static type in array
+	_, err = d.decodeArrayValueStaticType(array.valuePath)
```
turbolent, line 1804: Like discussed, I think it might be enough to go with Option 2, i.e. only have the static type in the array value, and then convert it to the sema type when needed / where the interpreter is available. That way we don't have to add the interpreter to the decoder here.

turbolent:
@SupunS Before making changes to the decoder and encoder, i.e. before merging this PR: 
We still need the current decoder and encoder later for the storage migration. 

Could you please open another PR which copies the current `decoder.go`/`encoder.go` to `decoder_v4.go`, then renames `DecoderV4`/`EncoderV4` in `decoder.go`/`encoder.go` to `DecoderV5`/`EncoderV5`, and also replaces the use from `DecoderV4`/`EncoderV4` everywhere to `DecoderV5`/`EncoderV5`?

At the same time, also `CurrentEncodingVersion` needs to be updated to 5.

Thank you!

turbolent:
@fxamacker Could you please also have a look at this PR? It changes the storage format to add static type information for arrays, the work is similar to your earlier work on optimizing encoding and decoding. Thank you!

runtime/interpreter/decode.go
SupunS: yeah, this was a side effect of trying to re-use the same code. I'll see if I can split it into two to make it more intuitive, or else rename as you've suggestedd.

SupunS: Added in 9f54635

SupunS:
@turbolent Added the ned encoder/decoder in https://github.com/onflow/cadence/pull/1039

runtime/interpreter/decode.go
```diff
@@ -1571,10 +1609,6 @@ func (d *DecoderV5) decodeCapabilityStaticType() (StaticType, error) {
 // This also extracts out the fields raw content and cache it separately inside the value.
 //
 func decodeCompositeMetaInfo(v *CompositeValue, content []byte) error {
-	if v.encodingVersion < 5 {
-		return decodeCompositeMetaInfoV4(v, content)
-	}
```
turbolent, line 1576: I removed the delegation here and in similar places, as we don't need to be backwards-compatible in the new encoder and decoder, because we are going to migrate all of the state

PR #1036, opened 2021-06-29
Add encoding/decoding dictionary static type info

SupunS:
Work towards #870
______


turbolent (commented):
Looks great!

runtime/interpreter/decode.go
```diff
@@ -1877,6 +1877,70 @@ func decodeDictionaryEntries(v *DictionaryValue, content []byte) error {
...
+		}
+		return err
+	}
+
+	// Decode type
+	// TODO: store dictionary type
+	//   Option 1: convert to sema type. - Don't have the interpreter
+	//   Option 2: Store static type in array
+	_, err = d.decodeStaticType()
+	if err != nil {
```
turbolent, line 1927: I think Option 2 might work, just like I commented in the PR for arrays: https://github.com/onflow/cadence/pull/1035#discussion_r660868483

runtime/interpreter/decode.go
```diff
@@ -1877,6 +1877,70 @@ func decodeDictionaryEntries(v *DictionaryValue, content []byte) error {
...
+		return err
+	}
+
+	//nolint:gocritic
+	v.entriesContent = append(keysContent, valuesContent...)
```
turbolent, line 1932: Is the linter complaining because the target variable isn't the same as the first argument?
If so, maybe replace with:

```suggestion
	keysContent = append(keysContent, valuesContent...)
	v.entriesContent = keysContent
```

runtime/interpreter/encode.go
```diff
@@ -855,12 +855,13 @@ func (e *Encoder) encodeArray(
 const (
 	encodedDictionaryValueKeysFieldKey    uint64 = 0
 	encodedDictionaryValueEntriesFieldKey uint64 = 1
+	encodedDictionaryValueTypeFieldKey    uint64 = 2
```
turbolent, line 858: It's a bigger breaking change, but given that we have to perform a storage migration anyways, maybe we want to place the type info at the start, just to be consistent with arrays?

If that's too much work, then let's just leave it as-is

turbolent:
@fxamacker Could you please also have a look at this PR? It changes the storage format to add static type information for dictionaries, the work is similar to your earlier work on optimizing encoding and decoding. Thank you!

runtime/interpreter/encode.go
SupunS: I also like that. Shouldn't be a big change :+1:

runtime/interpreter/encode.go
```diff
@@ -910,7 +912,13 @@ func (e *EncoderV5) encodeDictionaryValue(
 	//nolint:gocritic
 	keysPath := append(path, dictionaryKeyPathPrefix)
 
-	// Encode keys (as array) at array index encodedDictionaryValueKeysFieldKey
+	// (1) Encode dictionary static type at array index encodedDictionaryValueTypeFieldKey
```
turbolent, line 915: 👌

turbolent (approved):
Nice work! 👏

PR #1042, opened 2021-06-30
Finish storage format changes

turbolent:
Work towards #870 

## Description

- Fix the encoding of deferred array values: Always write the CBOR value tag
- Improve the re-encoding tests and assert the encoded and re-encoded contents are the same
- Remove the tests for old storage formats, they are / will not be used anymore
- Remove the old V3 decoder
- I had incorrectly removed the delegation from the current decoder to the previous one, i.e. 567d20e85521641669a735536152e7ef258de841 and f913717a80ed137cc8766c018dfb9b81b9f06214.

  I reverted these changes, as I now understand what they do and why it is important:
  When we are using the previous decoder, deferred values are created – we also need to decode the deferred data with the same version of the decoder.

  @SupunS aren't we missing a check like at the top of  `decodeArrayMetaInfo` also in `decodeDictionaryMetaInfo`?

  We'll need this scenario of decoding using the old decoder and re-encoding, where the arrays and dictionaries are first deferred and need to be forced to be decoded, for the state migration.
  That I was able to break this by reverting these commits and CI not failing is a sign that we're missing tests for it.

  @SupunS Could you please add tests for this ^? I've added this as a TODO to #870

______


runtime/interpreter/encode.go
```diff
@@ -799,6 +799,15 @@ func (e *EncoderV5) encodeArray(
 	deferrals *EncodingDeferrals,
 ) error {
 
+	// Encode tag number and array head
+	err := e.enc.EncodeRawBytes([]byte{
```
SupunS, line 803: good catch!

SupunS (approved):
Looks good!

Strange thing with the missing delegations for  `decodeDictionaryMetaInfo`. It was added in f00489b and seems to be still there in my local branch. Probably Git overwrote it with another commit somehow during merging?

I will add it back, and some tests too. Thanks for catching that!

SupunS:
Added the missing check and some tests: 28d2173.

PR #1038, opened 2021-06-30
Infer array and dictionary static types from expected type during import

turbolent:
⚠️ Depends on #1042

Work towards #870 

## Description

When importing values, provide the expected type (e.g. the parameter type for a script argument) to the import function, and use it to determine the static type for array values and dictionary values.

______

- [ ] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #1043, opened 2021-06-30
Include and consider static types of arrays and dictionaries

turbolent:
⚠️ Depends on #1038

Work towards #870 

## Description

- Consider static types of arrays and dictionaries in equality tests
- Include static types in dynamic types for arrays and dictionaries
- Consider array / dictionary dynamic type's static type in dynamic type check
______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -1769,7 +1769,7 @@ func (interpreter *Interpreter) checkValueTransferTargetType(value Value, target
 	dynamicTypeResults := DynamicTypeResults{}
 
 	valueDynamicType := value.DynamicType(interpreter, dynamicTypeResults)
-	if IsSubType(valueDynamicType, targetType) {
+	if interpreter.IsSubType(valueDynamicType, targetType) {
 		return true
 	}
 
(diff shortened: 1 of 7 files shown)


runtime/interpreter/interpreter.go
```diff
@@ -2684,9 +2684,19 @@ func IsSubType(subType DynamicType, superType sema.Type) bool {
 		case *sema.VariableSizedType:
 			superTypeElementType = typedSuperType.Type
 
+			subTypeStaticType := interpreter.ConvertStaticToSemaType(typedSubType.StaticType)
+			if !sema.IsSubType(subTypeStaticType, typedSuperType) {
+				return false
+			}
```
SupunS, line 2690: Once the runtime mutability checks (checking the type of array/dictionary before insertions) are added, I think we can rely only on this check, and maybe don't need to check both the static type and the dynamic type. Because that 'mutability checks on insertions' ensures that any value within the container is a subtype of the static type.

SupunS (approved):
Looks good!

I added a couple of suggestions, but not blockers for this PR. So I'll go ahead and merge this :+1:

PR #1048, opened 2021-07-05
Add static type sanity check for imported values

SupunS:
Work towards #870

## Description

Add a check to ensure the inferred type for arrays/ditionaries is always present and is valid. 

Ideally, this check should never fail. But this is more of a second-line of defense, to make sure no value is passed into a program without type info / with incorrect type info.

______

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -715,10 +715,47 @@ func validateArgumentParams(
 			}
 		}
 
+		// Ensure static type info is available for all values
+		interpreter.InspectValue(arg, func(value interpreter.Value) bool {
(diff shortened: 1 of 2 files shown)


runtime/runtime.go
```diff
@@ -715,12 +715,49 @@ func validateArgumentParams(
 			}
 		}
 
+		// Ensure static type info is available for all values
+		interpreter.InspectValue(arg, func(value interpreter.Value) bool {
```
turbolent, line 719: 👌

turbolent (approved):
Nice!

PR #1049, opened 2021-07-05
Infer static type from the value itself for imported arrays/dictionaries

SupunS:
Work towards #870

## Description

Infer the static type from the array/dictionary value, if the expected type is not specific enough/ not available.

Depends on #1025 and #1048
______


SupunS:
The last two commits 1d26160 and f24e710 have the actual code changes. The rest is from the dependent PRs.

turbolent:
@SupunS Do you think that we could hold off with this for now and get all the supertype inference PRs in later, i.e. the next release / spork? Do you think it's possible to do something simple for now, like e.g. infer `[AnyStruct]` for an array value where the expected static type is `AnyStruct`? (same for dictionaries)

SupunS:
@turbolent Yeah, that should be fine. I added this more as a way to test the concept/usage.
I think arrays are rather simple. We can assume it to be `[AnyStruct]`
Same can be done for dictionary values. 

But for dictionary keys, `AnyStruct` might be a problem because it is not a valid type for keys. And might become a problem somewhere down the line?

runtime/convertValues.go
```diff
@@ -663,8 +663,23 @@ func importArrayValue(
...
+			typ, err := inter.ConvertStaticToSemaType(value.StaticType())
+			if err != nil {
+				return nil, err
+			}
+			types[i] = typ
+		}
+
+		elementSuperType := sema.LeastCommonSuperType(types...)
+		if elementSuperType == sema.InvalidType {
+			return nil, fmt.Errorf("cannot import array: elements does not belong to the same type")
```
turbolent, line 678:
```suggestion
			return nil, fmt.Errorf("cannot import array: elements do not belong to the same type")
```

turbolent (approved):
Nice!

PR #1051, opened 2021-07-05
Test and implement json.Unmarshal() support code

jsatdapr:
## Description

In attempt to prove that the existing JSON marshaling is complete, we refactor AST tests to "roundtrip" the marshaling process. Where we normally call `json.Marshal(x)` in an AST test, we now call `json.Marshal(json.Unmarshal(json.Marshal(x))`

See `runtime/ast/unmarshal.go`
______

- `n/a` Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- `n/a` Updated relevant documentation 
- [?] Added appropriate labels

### DIFF
--- a/runtime/ast/unmarshal.go
+++ b/runtime/ast/unmarshal.go
@@ -1,0 +1,9 @@
+package ast
+
+import (
+	"encoding/json"
+)
+
+func jsonMarshalAndVerify(i interface{}) ([]byte, error) {
+	return json.Marshal(i)
+}
--- a/runtime/ast/unmarshal.go
+++ b/runtime/ast/unmarshal.go
@@ -2,8 +2,28 @@ package ast
 
 import (
 	"encoding/json"
+	"fmt"
+	"reflect"
 )
 
 func jsonMarshalAndVerify(i interface{}) ([]byte, error) {
-	return json.Marshal(i)
+	b, err := json.Marshal(i)
+	if err != nil { return b, err }
+
+	expected := string(b)
+
+	unmarshalled := reflect.New(reflect.TypeOf(i)).Interface()
+	err = json.Unmarshal(b, &unmarshalled)
+	if err != nil { return b, err }
+
+	b2, err2 := json.Marshal(unmarshalled)
+	if err2 != nil { return b2, err2 }
+
+	actual := string(b2)
+
+	if expected != actual {
+		return nil, fmt.Errorf("un/marshal failed:\n%s\n%s\n", expected, actual)
+	}
+
+	return b, err
 }
--- a/runtime/ast/unmarshal.go
+++ b/runtime/ast/unmarshal.go
@@ -3,27 +3,899 @@ package ast
 import (
 	"encoding/json"
 	"fmt"
+	"math/big"
 	"reflect"
+	"strconv"
+	"strings"
+
+	"github.com/onflow/cadence/runtime/common"
 )
 
-func jsonMarshalAndVerify(i interface{}) ([]byte, error) {
-	b, err := json.Marshal(i)
-	if err != nil { return b, err }
-
-	expected := string(b)
-
-	unmarshalled := reflect.New(reflect.TypeOf(i)).Interface()
-	err = json.Unmarshal(b, &unmarshalled)
-	if err != nil { return b, err }
-
-	b2, err2 := json.Marshal(unmarshalled)
-	if err2 != nil { return b2, err2 }
-
-	actual := string(b2)
-
-	if expected != actual {
-		return nil, fmt.Errorf("un/marshal failed:\n%s\n%s\n", expected, actual)
+func jsonMarshalAndVerify(i interface{}) (b []byte, err error) {
+	if b, err = json.Marshal(i); err == nil {
+		expected := string(b)
+		roundtripped := reflect.New(reflect.TypeOf(i)).Interface()
+		if err = json.Unmarshal(b, &roundtripped); err == nil {
+			if b, err = json.Marshal(roundtripped); err == nil {
+				actual := string(b)
+				if expected != actual {
+					return nil, fmt.Errorf("un/marshal failed:\n%s\n%s\n", expected, actual)
+				}
+			}
+		}
 	}
+	return
+}
 
-	return b, err
+//////////////////////////////////////////////////////////////////////////////
+
+var accessKindMap = map[string]Access{
+	"AccessNotSpecified":   AccessNotSpecified,
+	"AccessPrivate":        AccessPrivate,
+	"AccessContract":       AccessContract,
+	"AccessAccount":        AccessAccount,
+	"AccessPublic":         AccessPublic,
+	"AccessPublicSettable": AccessPublicSettable,
+}
+
+var conditionKindMap = map[string]ConditionKind{
+	"ConditionKindUnknown": ConditionKindUnknown,
+	"ConditionKindPre":     ConditionKindPre,
+	"ConditionKindPost":    ConditionKindPost,
+}
+
+var expressionTypeMap = map[string]func() Expression{
+	"ArrayExpression":       func() Expression { return &ArrayExpression{} },
+	"BinaryExpression":      func() Expression { return &BinaryExpression{} },
+	"BoolExpression":        func() Expression { return &BoolExpression{} },
+	"CastingExpression":     func() Expression { return &CastingExpression{} },
+	"ConditionalExpression": func() Expression { return &ConditionalExpression{} },
+	"CreateExpression":      func() Expression { return &CreateExpression{} },
+	"DestroyExpression":     func() Expression { return &DestroyExpression{} },
+	"DictionaryExpression":  func() Expression { return &DictionaryExpression{} },
+	"FixedPointExpression":  func() Expression { return &FixedPointExpression{} },
+	"ForceExpression":       func() Expression { return &ForceExpression{} },
+	"FunctionExpression":    func() Expression { return &FunctionExpression{} },
+	"IdentifierExpression":  func() Expression { return &IdentifierExpression{} },
+	"IndexExpression":       func() Expression { return &IndexExpression{} },
+	"IntegerExpression":     func() Expression { return &IntegerExpression{} },
+	"InvocationExpression":  func() Expression { return &InvocationExpression{} },
+	"MemberExpression":      func() Expression { return &MemberExpression{} },
+	"NilExpression":         func() Expression { return &NilExpression{} },
+	"PathExpression":        func() Expression { return &PathExpression{} },
+	"ReferenceExpression":   func() Expression { return &ReferenceExpression{} },
+	"StringExpression":      func() Expression { return &StringExpression{} },
+	"UnaryExpression":       func() Expression { return &UnaryExpression{} },
+}
+
+var locationKindMap = map[string]func() common.Location{
+	"AddressLocation":     func() common.Location { return &common.AddressLocation{} },
+	"REPLLocation":        func() common.Location { return &common.REPLLocation{} },
+	"ScriptLocation":      func() common.Location { return &common.ScriptLocation{} },
+	"TransactionLocation": func() common.Location { return &common.TransactionLocation{} },
+	"StringLocation":      func() common.Location { return nil },
+	"FlowLocation":        func() common.Location { return nil },
+	"IdentifierLocation":  func() common.Location { return nil },
+}
+
+var operationMap = map[string]Operation{
+	"OperationUnknown":           OperationUnknown,
+	"OperationOr":                OperationOr,
+	"OperationAnd":               OperationAnd,
+	"OperationEqual":             OperationEqual,
+	"OperationNotEqual":          OperationNotEqual,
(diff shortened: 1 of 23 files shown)


runtime/common/declarationkind.go
```diff
@@ -190,3 +192,41 @@ func (k DeclarationKind) Keywords() string {
...
+	"DeclarationKindPrepare":            DeclarationKindPrepare,
+	"DeclarationKindExecute":            DeclarationKindExecute,
+	"DeclarationKindTypeParameter":      DeclarationKindTypeParameter,
+	"DeclarationKindPragma":             DeclarationKindPragma,
+	"DeclarationKindEnum":               DeclarationKindEnum,
+	"DeclarationKindEnumCase":           DeclarationKindEnumCase,
+}
+
+func (a *DeclarationKind) UnmarshalJSON(b []byte) error {
+	if c, ok := declarationKindMap[strings.Trim(string(b), `"`)]; ok {
```
turbolent, line 227: Instead of trimming the serialized string's quotes manually, potentially leaving quoted content, unmarshal as a string:

```go
var s string
if err := json.Unmarshal(b, &s); err != nil {
	return err
}
```

runtime/common/compositekind.go
```diff
@@ -180,3 +182,20 @@ func (k CompositeKind) SupportsInterfaces() bool {
...
+	"CompositeKindUnknown":   CompositeKindUnknown,
+	"CompositeKindStructure": CompositeKindStructure,
+	"CompositeKindResource":  CompositeKindResource,
+	"CompositeKindContract":  CompositeKindContract,
+	"CompositeKindEvent":     CompositeKindEvent,
+	"CompositeKindEnum":      CompositeKindEnum,
+}
+
+func (a *CompositeKind) UnmarshalJSON(b []byte) error {
+	if c, ok := compositeKindMap[strings.Trim(string(b), `"`)]; ok {
```
turbolent, line 196: Instead of trimming the serialized string's quotes manually, potentially leaving quoted content, unmarshal as a string:

```go
var s string
if err := json.Unmarshal(b, &s); err != nil {
	return err
}
```

runtime/ast/unmarshal.go
```diff
@@ -0,0 +1,901 @@
...
+		roundtripped := reflect.New(reflect.TypeOf(i)).Interface()
+		if err = json.Unmarshal(b, &roundtripped); err == nil {
+			if b, err = json.Marshal(roundtripped); err == nil {
+				actual := string(b)
+				if expected != actual {
+					return nil, fmt.Errorf("un/marshal failed:\n%s\n%s\n", expected, actual)
+				}
+			}
+		}
+	}
```
turbolent, line 26: Instead of nesting the code deeper and deeper, consider exiting early to keep the flow flat, e.g
```suggestion
	if b, err = json.Marshal(i); err != nil {
		return nil, err
	}
	expected := string(b)
	roundtripped := reflect.New(reflect.TypeOf(i)).Interface()
	if err = json.Unmarshal(b, &roundtripped); err != nil {
		return nil, err
	}
	if b, err = json.Marshal(roundtripped); err != nil {
		return nil, err
	}
	actual := string(b)
	if expected != actual {
		return nil, fmt.Errorf("un/marshal failed:\n%s\n%s\n", expected, actual)
	}
}
```

Also, given that this is a test-only function, move it to the test files.

runtime/ast/unmarshal.go
```diff
@@ -0,0 +1,901 @@
...
+	return
+}
+
+//////////////////////////////////////////////////////////////////////////////
+
+var accessKindMap = map[string]Access{
```
turbolent, line 32: Just like with `declarationKindMap`, consider moving these maps here next to their respective types. That way, we won't forget updating these maps when the enums change.

runtime/ast/unmarshal.go
```diff
@@ -0,0 +1,901 @@
...
+}
+
+var conditionKindMap = map[string]ConditionKind{
+	"ConditionKindUnknown": ConditionKindUnknown,
+	"ConditionKindPre":     ConditionKindPre,
+	"ConditionKindPost":    ConditionKindPost,
+}
+
+var expressionTypeMap = map[string]func() Expression{
+	"ArrayExpression":       func() Expression { return &ArrayExpression{} },
```
turbolent, line 48: What is the reason for the value being a function here? Can the value type of the map just be `Expression`?

runtime/ast/unmarshal.go
```diff
@@ -0,0 +1,901 @@
...
+	"MemberExpression":      func() Expression { return &MemberExpression{} },
+	"NilExpression":         func() Expression { return &NilExpression{} },
+	"PathExpression":        func() Expression { return &PathExpression{} },
+	"ReferenceExpression":   func() Expression { return &ReferenceExpression{} },
+	"StringExpression":      func() Expression { return &StringExpression{} },
+	"UnaryExpression":       func() Expression { return &UnaryExpression{} },
+}
+
+var locationKindMap = map[string]func() common.Location{
+	"AddressLocation":     func() common.Location { return &common.AddressLocation{} },
```
turbolent, line 72: What is the reason for the value being a function here? Can the value type of the map just be Expression?

runtime/ast/unmarshal.go
```diff
@@ -0,0 +1,901 @@
...
+	"OperationForceCast":         OperationForceCast,
+	"OperationBitwiseOr":         OperationBitwiseOr,
+	"OperationBitwiseXor":        OperationBitwiseXor,
+	"OperationBitwiseAnd":        OperationBitwiseAnd,
+	"OperationBitwiseLeftShift":  OperationBitwiseLeftShift,
+	"OperationBitwiseRightShift": OperationBitwiseRightShift,
+}
+
+var statementKindMap = map[string]func() Statement{
+	"AssignmentStatement":        func() Statement { return &AssignmentStatement{} },
```
turbolent, line 110: What is the reason for the value being a function here? Can the value type of the map just be Expression?

runtime/ast/unmarshal.go
```diff
@@ -0,0 +1,901 @@
...
+
+var transferOperationMap = map[string]TransferOperation{
+	"TransferOperationUnknown":    TransferOperationUnknown,
+	"TransferOperationCopy":       TransferOperationCopy,
+	"TransferOperationMove":       TransferOperationMove,
+	"TransferOperationMoveForced": TransferOperationMoveForced,
+}
+
+var typeNameMap = map[string]func() Type{
+	"ConstantSizedType": func() Type { return &ConstantSizedType{} },
```
turbolent, line 139: What is the reason for the value being a function here? Can the value type of the map just be Expression?

runtime/ast/unmarshal.go
```diff
@@ -0,0 +1,901 @@
...
+
+var variableKindMap = map[string]VariableKind{
+	"VariableKindNotSpecified": VariableKindNotSpecified,
+	"VariableKindVariable":     VariableKindVariable,
+	"VariableKindConstant":     VariableKindConstant,
+}
+
+//////////////////////////////////////////////////////////////////////////////
+
+type BigIntUnmarshaler struct{ Int **big.Int }
```
turbolent, line 158: Just like with the maps above: Consider moving these unmarshaling types and functions next to their corresponding type / next to the type's JSON marshaling function. 

That way we won't forget to update the unmarshaling code when we update the type (i.e. keep type, marshaling, and unmarshaling code all in sync).

Also, consider putting the fields on separate lines to make the code more readable.

runtime/ast/unmarshal.go
```diff
@@ -0,0 +1,901 @@
...
+	} else if s == "null" {
+		*a.Int = nil
+		return nil
+	}
+	if z, ok := (&big.Int{}).SetString(s, 10); !ok {
+		return fmt.Errorf("invalid BigInt %s", s)
+	} else {
+		*a.Int = z
+		return nil
+	}
```
turbolent, line 179:
```suggestion
	bigInt, ok := (&big.Int{}).SetString(s, 10)
	if !ok {
		return fmt.Errorf("invalid BigInt %s", s)
	}
	*a.Int = bigInt
	return nil
```

runtime/ast/unmarshal.go
```diff
@@ -0,0 +1,901 @@
...
+		*a.Type = f()
+		return json.Unmarshal(b, &a.Type)
+	}
+	return fmt.Errorf("unknown Type")
+}
+
+//////////////////////////////////////////////////////////////////////////////
+
+func (a *Access) UnmarshalJSON(b []byte) error {
+	if c, ok := accessKindMap[strings.Trim(string(b), `"`)]; ok {
```
turbolent, line 276: Instead of trimming the serialized string's quotes manually, potentially leaving quoted content, unmarshal as a string:

```go
var s string
if err := json.Unmarshal(b, &s); err != nil {
	return err
}
```

runtime/ast/unmarshal.go
```diff
@@ -0,0 +1,901 @@
...
+	}{
+		Alias:   (*Alias)(a),
+		Test:    ExpressionUnmarshaler{&a.Test},
+		Message: ExpressionUnmarshaler{&a.Message},
+	})
+	return err
+}
+
+func (a *ConditionKind) UnmarshalJSON(b []byte) error {
+	if c, ok := conditionKindMap[strings.Trim(string(b), `"`)]; ok {
```
turbolent, line 353: Instead of trimming the serialized string's quotes manually, potentially leaving quoted content, unmarshal as a string:

```go
var s string
if err := json.Unmarshal(b, &s); err != nil {
	return err
}
```

runtime/ast/unmarshal.go
```diff
@@ -0,0 +1,901 @@
...
+func (a *NominalType) UnmarshalJSON(b []byte) error {
+	type Alias NominalType
+	return json.Unmarshal(b, &struct {
+		*Alias
+	}{
+		Alias: (*Alias)(a)})
+}
+
+func (a *Operation) UnmarshalJSON(b []byte) error {
+	if op, ok := operationMap[strings.Trim(string(b), `"`)]; ok {
```
turbolent, line 499: Instead of trimming the serialized string's quotes manually, potentially leaving quoted content, unmarshal as a string:

```go
var s string
if err := json.Unmarshal(b, &s); err != nil {
	return err
}
```

runtime/ast/unmarshal.go
```diff
@@ -0,0 +1,901 @@
...
+		*Alias
+		StartPos *Position
+	}{
+		Alias:    (*Alias)(a),
+		StartPos: &a.Pos,
+	})
+}
+
+func (a *TransferOperation) UnmarshalJSON(b []byte) error {
+	if op, ok := transferOperationMap[strings.Trim(string(b), `"`)]; ok {
```
turbolent, line 612: Instead of trimming the serialized string's quotes manually, potentially leaving quoted content, unmarshal as a string:

```go
var s string
if err := json.Unmarshal(b, &s); err != nil {
	return err
}
```

runtime/ast/unmarshal.go
```diff
@@ -0,0 +1,901 @@
...
+	}{
+		Alias:       (*Alias)(a),
+		Value:       ExpressionUnmarshaler{&a.Value},
+		SecondValue: ExpressionUnmarshaler{&a.SecondValue},
+		StartPos:    &a.StartPos,
+	})
+}
+
+func (a *VariableKind) UnmarshalJSON(b []byte) error {
+	if vk, ok := variableKindMap[strings.Trim(string(b), `"`)]; ok {
```
turbolent, line 648: Instead of trimming the serialized string's quotes manually, potentially leaving quoted content, unmarshal as a string:

```go
var s string
if err := json.Unmarshal(b, &s); err != nil {
	return err
}
```

turbolent (commented):
Nice work, thank you for adding this! 👍

turbolent:
@jsatdapr Linting is failing. You can run `make lint` locally to see the errors.

jsatdapr:
This does not verify that `json.Unmarshal()` is "complete", only "sufficient" because it relies on the completeness of the test suite.

The test suite apparently does not test recursive AST bits like `ParentIfStatement` and `ParentVariableDeclaration` because those are *not* umarshalled, I don't think.  I don't see how they could be.  The JSON would need to be golang-specific with some way to "point" at other values elsewhere in the JSON, or the marshalling would infinitely recurse.  Indeed, those two fields are `json:"-"`d out of the marshalling.

`ParentIfStatement` and `ParentVariableDeclaration` are only referenced in `check_casting_expression.go`:
```
if expression.Operation == ast.OperationFailableCast {
	if expression.ParentVariableDeclaration == nil ||
		expression.ParentVariableDeclaration.ParentIfStatement == nil {
		checker.report(
			&InvalidFailableResourceDowncastOutsideOptionalBindingError{
				Range: ast.NewRangeFromPositioned(expression),
			},
		)
	}
```

These details suggest that this logic should probably be enforced by the parser, before it gets to the checker.

It also suggests obliquely that those fields should be removed from the AST, and a ban on recursive elements in the AST should be introduced.  That ban can be checked with a test case via reflection on the AST types, starting at ast.Program or somesuch.

PR #1052, opened 2021-07-06
 Infer static-type from imported array/dictionary values

SupunS:
Work towards #870

## Description

This is more of a short-term solution for #1049. Types are inferred in the following manner:
- Infer the array static type as `[AnyStruct]`
- Dictionaries:
  - Infer values type as `AnyStruct`
  - Infer keys type from the first key
    - Ensure all keys belongs to the same type

______


turbolent:
@SupunS Sorry this has been open for so long and I didn't review it yet. Do we need this to ship the feature?

SupunS:
Yeah, we would need it for the next release.

This would be needed for the new storage, since now the types are stored. Also needed for adding the container mutation check: https://github.com/onflow/cadence/pull/1103

runtime/convertValues.go
```diff
@@ -551,6 +554,52 @@ func importDictionaryValue(
...
+		} else {
+			// Infer the keys type from the first entry
+
+			key := keysAndValues[0]
+			keyStaticType = key.StaticType()
+			keySemaType := inter.ConvertStaticToSemaType(keyStaticType)
+
+			if sema.IsSubType(keySemaType, sema.NumberType) {
+				keyStaticType = interpreter.PrimitiveStaticTypeNumber
+			} else if sema.IsSubType(keySemaType, sema.PathType) {
```
turbolent, line 572: What is the reason for specifically handling `Number` and `Path` subtypes here? Is this because `IsValidDictionaryKeyType` allows all subtypes of these types and we want to infer the supertype?

SupunS: Yes, exactly. Dictionary Keys could be a mix of numeric sub-types/ path sub-types, and can be still valid. So here we always go for the safest option, by inferring the most generic type.
i.e: sending a value similar to: `var map: {Number:String} = {1: "int", 2.3: "float"}`

turbolent: Makes sense 👍 

Could you please add this as a comment?

turbolent (approved):

runtime/convertValues.go
SupunS: Added in 6d7e59d

PR #1072, opened 2021-07-14
 Report a hint for casting an expression of type `T` to same type `T`

### LINKED ISSUE #1047: Report warnings for redundant casts
### Issue To Be Solved

SImple cast operator (`as`) is another way of annotating an expression to mark its type. Now that we support inferring type from context, sometimes simple casts can be redundant if the expression already has type info. 

There are a few ways an expression can have type info:

- Case I: Available through the parent context. e.g: type annotation.
  ```kotlin
  var x = 5 as Int8    // this is ok

  var x: Int8 = 5 as Int8   // Cast is redundant because '5' would be Int8 even without the cast.
  ```
- Case II: Expression is self-typed and is same as the casted type.
  ```kotlin
   var x = (5 as Int8) as Int8    // Inner expression is already 'Int8'.
                                  // So casting again to same type 'Int8' is redundant.

   var y = "hello" as String    // Inner expression is already 'String'.
                                // So casting again to same type 'String' is redundant.
  ```

### Suggest A Solution

It is possible to report a warning if there are redundant casts like above.

- [x] Case I: Can be easily validated by checking if the casted type `T` is the same as the contextually expected type `R`. i.e: is `T == R`
- [x] Case II: 
  - It is a bit tricky to handle Case II, because sometimes, the expression gets it type from the cast. e.g: `5 as Int8`. 
  - So need a way to differentiate the cases where an expression is 'self-typed to be `T`' vs 'inferred from the cast to be `T`'.
- [x] Make sure no warnings are reported for valid cases.

SupunS:
Closes #1047

## Description

This handles the Case-II in #1047: Expression is self-typed and is the same as casted type.
e.g:
```kotlin
// Invalid cases

 var x = "hello" as String    // Inner expression is already 'String'.
                              // So casting again to same type 'String' is redundant.

var y = [3, 5, 2] as [Int]    // Inner expression is already '[Int]' even without the cast.
                              // So casting again to same type '[Int]' is redundant.


// Valid cases

var z = [3, 5, 2] as [UInt8]  // Inner expression is '[UInt8]' with the cast, but would be '[Int]' if the cast is not there.
                              // So the casting is NOT redundant.
```
______


runtime/sema/check_casting_expression.go
```diff
@@ -412,3 +421,204 @@ func FailableCastCanSucceed(subType, superType Type) bool {
 
 	return true
 }
+
+// IsCastRedundant checks whether a simple cast is redundant.
+// Checks for two cases:
+//    - Case I: Contextually expected type is same as casted type (target type).
+//    - Case II: Expression is self typed, and is same as the casted type (target type).
+func IsCastRedundant(expr ast.Expression, exprInferredType, targetType, expectedType Type) bool {
+	if expectedType != nil && expectedType.Equal(targetType) {
```
dsainati1, line 430: While Cadence doesn't currently have type aliases, we can probably future proof this against such a feature by changing this `equal` check to instead check whether both `expectedType` and `targetType` are subtypes of each other. At the moment this seems like it will only give the redundant cast hint when the two types are syntactically equal, but not when they are semantically equal.

runtime/sema/check_casting_expression.go
```diff
@@ -412,3 +421,204 @@ func FailableCastCanSucceed(subType, superType Type) bool {
...
+	// If there is no expected type (e.g: var-decl with no type annotation),
+	// then the simple-cast might be used as a way of marking the type of the variable.
+	// Therefore it is ok for the target type to be a super-type.
+	// But being the exact type as expression's type is redundant.
+	// e.g:
+	//   var x: Int8 = 5
+	//   var y = x as Int8     // <-- not ok: `y` will be of type `Int8` with/without cast
+	//   var y = x as Integer  // <-- ok	: `y` will be of type `Integer`
+	return exprType != nil &&
+		exprType.Equal(targetType)
```
dsainati1, line 623: same as above

runtime/sema/check_casting_expression.go
SupunS: Hmm, that's a good point! But that might depend on how we define the `equal` method for type aliases. Like for e.g: In Go, type aliases are 'equal' at the user-level, and it would come down to how we are going to implement it.
We use this equal method across the checker, and all such places might affect when we introduce such a feature.

runtime/sema/check_casting_expression.go
```diff
@@ -412,3 +421,204 @@ func FailableCastCanSucceed(subType, superType Type) bool {
...
+	if expectedType != nil && expectedType.Equal(targetType) {
+		return true
+	}
+
+	checkCastVisitor := &CheckCastVisitor{}
+
+	return checkCastVisitor.isCastRedundant(expr, exprInferredType, targetType)
+}
+
+type CheckCastVisitor struct {
```
turbolent, line 439: Maybe move this type to its own file

runtime/sema/check_casting_expression.go
```diff
@@ -412,3 +421,204 @@ func FailableCastCanSucceed(subType, superType Type) bool {
...
+//    - Case I: Contextually expected type is same as casted type (target type).
+//    - Case II: Expression is self typed, and is same as the casted type (target type).
+func IsCastRedundant(expr ast.Expression, exprInferredType, targetType, expectedType Type) bool {
+	if expectedType != nil && expectedType.Equal(targetType) {
+		return true
+	}
+
+	checkCastVisitor := &CheckCastVisitor{}
+
+	return checkCastVisitor.isCastRedundant(expr, exprInferredType, targetType)
```
turbolent, line 436: Great idea to factor the casting check into its own type, though I think there is no need for a pointer type here (only the mutating functions need a pointer receiver):

```suggestion
	return CheckCastVisitor{}.isCastRedundant(expr, exprInferredType, targetType)
```

runtime/sema/check_casting_expression.go
```diff
@@ -154,7 +158,12 @@ func (checker *Checker) VisitCastingExpression(expression *ast.CastingExpression
 		return rightHandType
 
 	case ast.OperationCast:
-		if checker.expectedType != nil && checker.expectedType.Equal(rightHandType) {
+		// If there are errors in the lhs-expr, then the target type is considered as
+		// the inferred-type of the expression. i.e: exprActualType == rightHandType
+		// Then, it is not possible to determine whether the target type is redundant.
+		// Therefore don't check for redundant casts, if there are errors.
+		if !hasErrors &&
+			IsCastRedundant(leftHandExpression, exprActualType, rightHandType, checker.expectedType) {
```
turbolent, line 166: Thinking about hints and how the recent community effort to build a Cadence linter a bit more: I think it's great to have the hints, but maybe we should only have them during development, instead of always, i.e. also during production execution on the networks.

As a first step we could put code that generates hints, like this one, behind an interpreter option which is disabled by default. We would have it enabled it in the language server. 

This hint calculation looks a bit expensive, so maybe it would be good to gate it already in this PR.

Further, (and out of scope for this PR) we could maybe also factor the hints out of the type checker and have them be an optional second pass, or have them as linting/analyzer passes, like in Go: https://arslan.io/2019/06/13/using-go-analysis-to-write-a-custom-linter/.

runtime/sema/check_casting_expression.go
```diff
@@ -412,3 +421,204 @@ func FailableCastCanSucceed(subType, superType Type) bool {
...
+	result := expr.AcceptExp(d)
+	return result.(bool)
+}
+
+func (d *CheckCastVisitor) VisitBoolExpression(_ *ast.BoolExpression) ast.Repr {
+	return d.isTypeRedundant(BoolType, d.targetType)
+}
+
+func (d *CheckCastVisitor) VisitNilExpression(_ *ast.NilExpression) ast.Repr {
+	return d.isTypeRedundant(TypeOfNil, d.targetType)
```
turbolent, line 467: Maybe `TypeOfNil` should be renamed to `NilType` to be consistent in the naming (cf. `BoolType`). It wasn't introduced in this PR, but would be nice to clean it up 🙏

runtime/sema/check_casting_expression.go
```diff
@@ -412,3 +421,204 @@ func FailableCastCanSucceed(subType, superType Type) bool {
...
+
+	return checkCastVisitor.isCastRedundant(expr, exprInferredType, targetType)
+}
+
+type CheckCastVisitor struct {
+	exprInferredType Type
+	targetType       Type
+}
+
+var _ ast.ExpressionVisitor = &CheckCastVisitor{}
```
turbolent, line 444: Not quite sure, but could a simpler solution like be sufficient here: Would it be enough to just infer the type of the expression without the expected type, and compare it to the casted type (target type)?

Basically, in `IsCastRedundant`:

```go
if expectedType == nil {
	expectedType = // ... infer type for casted expression
} 
return expectedType.Equal(targetType)
```

turbolent (commented):
Nice work!

bluesign:
I know it is not totally related but below will give hint?

```
  var a: String = "deniz"
  var b = &a as &String
```

Currently I guess the one below is not possible also.
```
  var a: String = "deniz"
  var b = &a 
```

turbolent:
@bluesign `&t as &T` is actually not a casting expression, but an expression that creates a reference, it just happens to have the "same" syntax, so I think this PR won't address this. In fact, we require the explicit type when taking a reference, so that the code clearly states what the type of the reference is (e.g. it might be a different type, or an auth reference)

runtime/sema/check_casting_expression.go
SupunS: `CheckCastVisitor` maintains an updates the `exprInferredType` and `targetType` as it goes into nested expressions (e.g: array/dictionary literals). So going to need a pointer receiver.

runtime/sema/check_casting_expression.go
SupunS: yeah, it would be really clean to separate out the linter to a separate phase, then we can also enable/disable the entire phase, rather than gating everywhere. It might need some additional effort though, so like you've suggested, would be good to handle it separately. I'll create an issue to track.

Will add a flag to enable/disable this only check for now.

runtime/sema/check_casting_expression.go
SupunS: The problem with that is, sometimes the casted/target type is used as the annotation to infer the type for the expression. e.g: `[1, 3] as [UInt8]` Then it would en

So we could have to run one pass with the expected type, and another pass without the expected type and compare the results of the two. But two passes would be even more expensive (`CheckCastVIsitor` doesn't recursively visit all the nested expressions, so it's fairly less expensive), also we would need a way to disable the error reporting in the second pass.

turbolent (commented):
Looks good, nice work! 

Should be ready to get merged once gated behind a disabled option, so we don't always run this (especially not on production).

runtime/sema/check_casting_expression.go
turbolent: Makes sense 👍

runtime/sema/check_casting_expression.go
turbolent: Right, the mutating functions should have a pointer receiver, but the value here itself doesn't have to be, see e.g. https://go.dev/play/p/iF22kMUH_MV

SupunS:
@turbolent It was added in https://github.com/onflow/cadence/pull/1072/commits/11e9ecc2b61f04ce301dc4f2761c88f396b04fd7. It's turned off by default.
Need to bump the version in LS to enable it there. I guess we should do that separately.

runtime/sema/check_casting_expression.go
SupunS: `(d *CheckCastVisitor) IsRedundantCast()` is actually the mutating function, and is recursively invoked by the others.

turbolent:
@SupunS Ah, my bad, I had missed it 👍

runtime/sema/checker.go
```diff
@@ -235,6 +236,16 @@ func WithPositionInfoEnabled(enabled bool) Option {
 	}
 }
 
+// WithLintingEnabled returns a checker option which enables/disables
+// advanced linting.
+//
+func WithLintingEnabled(enabled bool) Option {
```
turbolent, line 242: Must also be propagated to sub-interpreters in `NewSubInterpreter`

turbolent (approved):
Looks good, great work! 

Just one line missing in `NewSubInterpreter` (see comment), but then should be ready to be merged 👍

runtime/sema/checker.go
SupunS: This flag is for the checker. I think we don't need to pass it down?

PR #1076, opened 2021-07-16
Interface default implementation support

### LINKED ISSUE #989: Adding default implementations to interface methods
### Issue To Be Solved

In order to evolve contracts and standards like the NFT contract Interface having some way to specify the default behavior of a interface method is needed. 

Say you want to add a `getName(): String` method to all INFT types. In order to do that safely the sanes way is IMHO to add support for default implementations. 

### Suggested Solution

Interface methods can today contain pre and post blocks. One suggestion from bluesign  is that they can also contain a default block. And when implementing an interface methods that have a default block can be left unimplemented in the implementing code

### Context

See here for context https://github.com/onflow/flow-nft/issues/9

bluesign:
Closes #989

## Description

Add default implementation support to interface methods. 

______


bjartek:
@turbolent any chance you could take a peak at this.

turbolent:
Thank you again for your work here @bluesign! The review and merge will have to wait a while, as we currently have a backlog of work to integrate. Thank you for you patience

runtime/tests/checker/interface_test.go
```diff
@@ -1991,3 +1998,117 @@ func TestCheckInvalidInterfaceUseAsTypeSuggestion(t *testing.T) {
...
+	
+	struct Test:  InterfaceA, InterfaceB  {
+
+	}
+
+    `)
+
+	errs := ExpectCheckerErrors(t, err, 1)
+
+	require.IsType(t, &sema.ConformanceError{}, errs[0])
```
bjartek, line 2085: I do not understand why this should be an error. Why it it not allowed to use the default value from InterfaceA here?

bluesign: Normally if the interface does not have a default implementation, later on, adds one, should be a non-breaking change for the contract that implements that interface. But if we allow this case, then we lose this guarantee.

bjartek: This make sense as long as you only allow a single interface to have a default implementation.

bluesign: @bjartek sorry I don't follow, can you explain a bit more?

bjartek: I see your point on why it needs to be like this :)

bluesign: Actually at first I was thinking order of conformances could be used, but it can be little confusing to new developers.

bjartek:
If I have an interface that requires a method and that has a default implementation that appears to not be allowed now. In my GeneircNFT repo https://github.com/bjartek/flow-nfmt/blob/nftv2/contracts/GenericNFT.cdc i tried to make borrowNFT be default implemented but it does not work. If possible this should be fixed.

runtime/tests/checker/interface_test.go
```diff
@@ -145,17 +145,26 @@ func TestCheckInvalidInterfaceWithFunctionImplementation(t *testing.T) {
 					`
                       %s interface Test {
                           fun test(): Int {
+							  pre{
+
+							  }
+							  post{
+
+							  }
```
turbolent, line 153: Are these strictly needed? Could they be removed?

runtime/tests/checker/interface_test.go
```diff
@@ -145,17 +145,26 @@ func TestCheckInvalidInterfaceWithFunctionImplementation(t *testing.T) {
...
                              return 1
                           }
                       }
+					  
+					  %s TestUser: Test{
+
+					  }
                     `,
-					kind.Keyword(),
+					kind.Keyword(), kind.Keyword(),
```
turbolent, line 162:
```suggestion
					kind.Keyword(),
```

runtime/tests/checker/interface_test.go
```diff
@@ -145,17 +145,26 @@ func TestCheckInvalidInterfaceWithFunctionImplementation(t *testing.T) {
 					`
                       %s interface Test {
```
turbolent, line 146:
```suggestion
                      %[1]s interface Test {
```

runtime/tests/checker/interface_test.go
```diff
@@ -145,17 +145,26 @@ func TestCheckInvalidInterfaceWithFunctionImplementation(t *testing.T) {
...
+
+							  }
+							  post{
+
+							  }
                              return 1
                           }
                       }
+					  
+					  %s TestUser: Test{
```
turbolent, line 158:
```suggestion
					  %[1]s TestUser: Test{
```

runtime/tests/checker/interface_test.go
```diff
@@ -1991,3 +1998,117 @@ func TestCheckInvalidInterfaceUseAsTypeSuggestion(t *testing.T) {
...
+
+	_, err := ParseAndCheck(t, `
+	struct interface InterfaceA {
+		fun test(): Int{
+			return 41
+		}
+	}
+
+	struct interface InterfaceB {
+		fun test(): Int{
```
turbolent, line 2015:
```suggestion
		fun test(): Int {
			return 41
		}
	}

	struct interface InterfaceB {
		fun test(): Int {
```

runtime/tests/checker/interface_test.go
```diff
@@ -1991,3 +1998,117 @@ func TestCheckInvalidInterfaceUseAsTypeSuggestion(t *testing.T) {
...
+			return 41
+		}
+	}
+	
+	struct Test:  InterfaceA, InterfaceB   {
+		fun test(): Int{
```
turbolent, line 2050:
```suggestion
		fun test(): Int{
			return 41
		}
	}

	struct interface InterfaceB {
		fun test(): Int {
			return 41
		}
	}
	
	struct Test:  InterfaceA, InterfaceB {
		fun test(): Int {
```

runtime/tests/checker/interface_test.go
```diff
@@ -1991,3 +1998,117 @@ func TestCheckInvalidInterfaceUseAsTypeSuggestion(t *testing.T) {
...
 	)
 }
+
+//MultipleInterfaceDefaultImplementation
+func TestCheckInvalidMultipleInterfaceDefaultImplementation(t *testing.T) {
+
+	t.Parallel()
+
+	_, err := ParseAndCheck(t, `
+	struct interface InterfaceA {
```
turbolent, line 2008: Use spaces instead of tabs in code literals. Maybe just do a search-and-replace for tabs and replace them with 4 spaces, then reformat using gofmt.

runtime/tests/checker/interface_test.go
```diff
@@ -1991,3 +1998,117 @@ func TestCheckInvalidInterfaceUseAsTypeSuggestion(t *testing.T) {
...
+		fun test(): Int{
+			return 41
+		}
+	}
+
+	struct interface InterfaceB {
+		fun test(): Int
+	}
+	
+	struct Test:  InterfaceA, InterfaceB  {
```
turbolent, line 2077:
```suggestion
	struct interface InterfaceA {
		fun test(): Int {
			return 41
		}
	}

	struct interface InterfaceB {
		fun test(): Int
	}
	
	struct Test:  InterfaceA, InterfaceB {
```

runtime/tests/checker/interface_test.go
```diff
@@ -1991,3 +1998,117 @@ func TestCheckInvalidInterfaceUseAsTypeSuggestion(t *testing.T) {
...
+			return 41
+		}
+	}
+
+	struct interface InterfaceB {
+		fun test(): Int
+	}
+	
+	struct Test:  InterfaceA, InterfaceB   {
+		fun test(): Int{
```
turbolent, line 2105:
```suggestion
		fun test(): Int {
			return 41
		}
	}

	struct interface InterfaceB {
		fun test(): Int
	}
	
	struct Test:  InterfaceA, InterfaceB {
		fun test(): Int {
```

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -8294,3 +8294,66 @@ func BenchmarkNewInterpreter(b *testing.B) {
...
+
+	inter := parseCheckAndInterpret(t, `
+
+	struct interface InterfaceA {
+		fun test(): Int{
+			return 42
+		}
+	}
+	
+	struct Test: InterfaceA   {
```
turbolent, line 8311:
```suggestion
		fun test(): Int {
			return 42
		}
	}
	
	struct Test: InterfaceA {
```

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -8294,3 +8294,66 @@ func BenchmarkNewInterpreter(b *testing.B) {
...
+	inter := parseCheckAndInterpret(t, `
+
+	struct interface InterfaceA {
+		fun test(): Int{
+			return 41
+		}
+	}
+	
+	struct Test: InterfaceA   {
+		fun test(): Int{
```
turbolent, line 8342:
```suggestion
	struct interface InterfaceA {
		fun test(): Int {
			return 41
		}
	}
	
	struct Test: InterfaceA {
		fun test(): Int {
```

turbolent (commented):
Great work!

bluesign:
@turbolent thanks for the review, I am planning to jump in tomorrow.

runtime/interpreter/interpreter.go
```diff
@@ -1668,6 +1674,29 @@ func (interpreter *Interpreter) compositeDestructorFunction(
 	}
 }
 
+func (interpreter *Interpreter) interfaceFunctions(
```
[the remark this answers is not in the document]
bluesign: @turbolent I think this would require a flag to check if it is interface / composite function then because interface on checking if interface function has an implementation. 

Though I am not 100% sure if that check is required.

runtime/sema/check_composite_declaration.go
```diff
@@ -1017,23 +1022,39 @@ func (checker *Checker) checkCompositeConformance(
...
+						(*overridden)[name] = true
+					}
+
+				}
 			}
-			return
 		}
 
-		if !checker.memberSatisfied(compositeMember, interfaceMember) {
+		if compositeMember != nil && !checker.memberSatisfied(compositeMember, interfaceMember) {
```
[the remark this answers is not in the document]
bluesign: here covers the case where compositeMember is inherited from interface. so now it can be missing (nil) but can cause no error in !ok section (if interface implements)

runtime/sema/check_composite_declaration.go
```diff
@@ -964,7 +967,9 @@ func (checker *Checker) checkCompositeConformance(
 	interfaceType *InterfaceType,
 	compositeKindMismatchIdentifier ast.Identifier,
 	options compositeConformanceCheckOptions,
+	overridden *map[string]bool,
```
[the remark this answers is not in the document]
bluesign: I changed the type but somehow return value felt wrong here a bit, as it is a bit like aggregate (or seen) implementations. But I can switch to that if you prefer that way

turbolent:
@bluesign Would it be OK to commit some improvements to this PR, or would you rather like a separate PR against this one?

runtime/interpreter/interpreter.go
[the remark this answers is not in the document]
turbolent: I see, compared to `compositeFunction`, `interfaceFunctions` checks if the declaration has statements, makes sense 👍 
Good idea to only allocate the map if it is needed 👌

runtime/sema/check_composite_declaration.go
```diff
@@ -1730,15 +1762,6 @@ func (checker *Checker) checkSpecialFunction(
 	)
 
 	switch containerKind {
-	case ContainerKindInterface:
```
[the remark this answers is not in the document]
turbolent: I see that `checkInterfaceSpecialFunctionBlock` was inlined below (https://github.com/onflow/cadence/pull/1076/files#diff-c0b32cc7cd7dca791c9db58c1c67445a5d4f64f069f008ad62a2bbab14a0f48aR183), but is it OK to remove the call/check here?

runtime/sema/errors.go
```diff
@@ -1012,6 +1012,34 @@ func (e *DuplicateConformanceError) Error() string {
 
 func (*DuplicateConformanceError) isSemanticError() {}
 
+// MultipleInterfaceDefaultImplementationsError
+
+// TODO: just make this a warning?
+
```
turbolent, line 1018: I think it makes sense to have this be an error for now, i.e. be more conservative

```suggestion
//
```

turbolent:
Re-reviewed the implementation, it looks good and makes sense. There's one open question about some code removal still that would be good to address. I have  some improvements that I could push up to the PR or in a separate PR as mentioned above.

I'll look at the tests on Monday, I think we're close with getting this in, thanks for the patience everyone!

runtime/tests/checker/interface_test.go
```diff
@@ -230,9 +233,8 @@ func TestCheckInvalidInterfaceWithInitializerImplementation(t *testing.T) {
 				),
 			)
 
-			errs := ExpectCheckerErrors(t, err, 1)
+			require.NoError(t, err)
```
turbolent, line 236: That's an interesting "edge case": Should we allow initializes and destructors as default implementations? I think we should rather exclude them, have this test keep failing.

runtime/tests/checker/interface_test.go
```diff
@@ -1991,3 +1992,117 @@ func TestCheckInvalidInterfaceUseAsTypeSuggestion(t *testing.T) {
 		errs[0].(*sema.InvalidInterfaceTypeError).ExpectedType,
 	)
 }
+
+//MultipleInterfaceDefaultImplementation
+func TestCheckInvalidMultipleInterfaceDefaultImplementation(t *testing.T) {
```
turbolent, line 1997: I guess this is a typo?

```suggestion
func TestCheckInvalidMultipleInterfaceDefaultImplementation(t *testing.T) {
```

runtime/tests/checker/interface_test.go
```diff
@@ -1991,3 +1992,117 @@ func TestCheckInvalidInterfaceUseAsTypeSuggestion(t *testing.T) {
...
+    struct Test:  InterfaceA, InterfaceB {
+        fun test(): Int {
+            return 42
+        }
+    }
+
+
+    `)
+
+	require.NoError(t, err)
```
turbolent, line 2052: 👍

runtime/tests/checker/interface_test.go
turbolent: Agreed that this should be an error, and using order of conformances is not a good idea, as it might not be obvious 👍

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -8294,3 +8294,66 @@ func BenchmarkNewInterpreter(b *testing.B) {
 		}
 	})
 }
+
+//InterfaceDefaultImplementation
+func TestInterpretInterfaceDefaultImplementation(t *testing.T) {
```
turbolent, line 8299:
```suggestion
func TestInterpretInterfaceDefaultImplementation(t *testing.T) {
```

turbolent (commented):
Tests look good, too!

Given that the implementation of the language feature also supports default implementations for type requirements, could you please also add tests for them?
They could literally just be "copies" of the existing tests

Also, ideally we would have documentation for this new language feature.
I'd also be happy to merge this PR without documentation, and then we could add the documentation in a separate follow up PR.

bluesign:
> @bluesign Would it be OK to commit some improvements to this PR, or would you rather like a separate PR against this one?

Ofc It's ok. Just I am on a pentest in London right now, I will be able to check more on Monday/Tuesday when I am back.

turbolent:
I'm currently adding tests for the feature using when type requirements, and it looks like the interface's type's default methods are declared in the concrete type's nested type. Looking into it

turbolent:
Alternatively, maybe just for now, and given we might remove type requirements soon, we forbid default methods in type requirements?

turbolent:
Finished implementing default functions for type requirements in the checker, remaining is implementing support for it in the interpreter.

Finally, we need to still discuss / decide on special functions (initializers and destructors) and potentially implement them and test the behaviour.

bluesign:
Thanks @turbolent, I think initializers and destructors are also can be needed but I am not sure about what can be the side effects  there. ( So I think I am a bit more leaning to against ) 

As every resource has most likely have an initializer and destructor, this can have problem when you are trying to compose something with multiple interfaces.  Maybe instead of override, we need to call all interface initializers and destructors somehow. ( but then we are getting into dangerous waters again )

runtime/tests/checker/interface_test.go
```diff
@@ -1991,3 +1992,538 @@ func TestCheckInvalidInterfaceUseAsTypeSuggestion(t *testing.T) {
...
+		t.Parallel()
+
+		_, err := ParseAndCheck(t, `
+          struct interface IA {
+              let x: Int
+
+              fun test(): Int {
+                  return self.x
+              }
+          }
```
SupunS, line 2479: I'm more interested in the runtime behavior of this example. How would the inherrited function pick the correct variable scope at runtime?

Probably good to add as a test case?

runtime/interpreter/interpreter.go
```diff
@@ -1881,6 +1897,33 @@ func (interpreter *Interpreter) compositeDestructorFunction(
...
+	lexicalScope *VariableActivation,
+) map[string]FunctionValue {
+
+	functionDeclarations := members.Functions()
+	functionCount := len(functionDeclarations)
+
+	var functions map[string]FunctionValue
+	if functionCount > 0 {
+		functions = make(map[string]FunctionValue, functionCount)
+	}
```
SupunS, line 1911: Could early-exit from here
```suggestion
	if functionCount == 0 {
		return nil
	}
	
	functions := make(map[string]FunctionValue, functionCount)
```
```

File: runtime/tests/checker/interface_test.go (hunk 18 above)
Comment (SupunS) [reply runtime/tests/checker/interface_test.go:236]: I agree that it makes sense to not allow default implementations for init/destructor. I don't have any concrete examples as to why it's bad, but just 'feels' it could introduce some trouble. Given that it's an addition, we can always add it in the future, if needed.

**Review by SupunS — COMMENTED** [review]
Looks good to me! Just have one question on the interface variables.

**Comment by bjartek** [comment]
Totally agree with @SupunS here. If you can say default implement a destructor and destroy things and you forget to override that in your impl it can be bad.

**Comment by bluesign** [comment]
> Totally agree with @SupunS here. If you can say default implement a destructor and destroy things and you forget to override that in your impl it can be bad. 


Not that I disagree ( I have the same feeling ) but how this can go bad?

**Comment by bjartek** [comment]
Well lets say the default impl of a destructor is to burn something because it does by default contain anything useful. Then you a dev add some escrow to it and forget to override destructor. Now you suddebly burned ft.

**Comment by bluesign** [comment]
> Well lets say the default impl of a destructor is to burn something because it does by default contain anything useful. Then you a dev add some escrow to it and forget to override destructor. Now you suddebly burned ft. 

Very good point.

**Review by robert-e-davidson3 — APPROVED** [review]

**Comment by bjartek** [comment]
Any progress here? I still think it is valid to not allow default impl of init/destroy. This feature is sorely needed.

File: runtime/tests/checker/interface_test.go (hunk 22 above)
Comment (bluesign) [reply runtime/tests/checker/interface_test.go:2479]: Hey @SupunS  sorry for the late answer, I could just get back into this, I will add a test case. 

But I am not too versed in variable scopes in Cadence tbh, so any help or clarification is much appreciated.

**Comment by bluesign** [comment]
@turbolent 

> The type checker should report an error for all special function declarations (e.g. initializers, destructors) with statements in interfaces, Add a test case for the new error above

Are there anything else than initializers, destructors I should disallow ? 

Also any meaningful error message suggestion for this ?

**Comment by turbolent** [comment]
@bluesign 
> Are there anything else than initializers, destructors I should disallow ?

Initializers and destructors are currently the only kinds of special functions (`ast.SpecialFunctionDeclaration`), but it is maybe best to just forbid default methods for all special functions in general. 

> Also any meaningful error message suggestion for this ?

There are no errors defined for this yet, but maybe an error message like this should suffice for now (we can improve it later): `fmt.Sprintf("%s may not be defined as a default function", kind.Name())`

**Comment by turbolent** [comment]
Merging `master` into this PR's branch should fix CI

**Comment by joshuahannan** [comment]
Looks like there are a bunch of changes to the docs, best practices, and such that aren't related to default implementation support. Did you also make a bunch of edits to those or have you just not pulled the latest from master or something?

**Comment by bluesign** [comment]
@joshuahannan  yeah I messed up when fetching upstream I guess :( trying to fix

**Comment by bluesign** [comment]
I think I managed to do all, just @SupunS's variable scope comment left, but I have no idea tbh how to approach that.

File: runtime/tests/checker/interface_test.go (hunk 22 above)
Comment (SupunS) [reply runtime/tests/checker/interface_test.go:2478]: Hey @bluesign , Sorry for being so late to get back on this. I went through the code once more, and I think now I understand how the scoping work with the default functions. The [defaultFunctions wrapper method](https://github.com/onflow/cadence/blob/6c92d06d9a6261797a0cf2edbd142bdb7f3cce78/runtime/interpreter/interpreter.go#L2000) always takes the current scope/variableActivation.

I added a test-case too: https://github.com/onflow/cadence/commit/eb899b0fca520c7aa93823ed54631a5ec63cd823 (this is in a forked branch) to verify the behaviour. I can push it to this PR, if that's OK?

Comment (bluesign) [reply runtime/tests/checker/interface_test.go:2478]: @SupunS sure thanks a lot

**Comment by bluesign** [comment]
@turbolent I don't see the problem here, can you expand a bit?

**Comment by turbolent** [comment]
> @turbolent I don't see the problem here, can you expand a bit?

Yes, an example would be:

```cadence
struct interface SI {
    fun default() {
        self.s = 1
    }
}

struct S: SI {
    var s: Int

    init() {
        self.s = 0
    }
}
```

Here, `SI.default` should not have access to `S.s`

turbolent:
Resolved the merge conflicts and added new conformances to the errors (they are user errors)

turbolent:
Implemented the missing case for checking if type requirements have default special functions in [b1ad20e](https://github.com/onflow/cadence/pull/1076/commits/b1ad20e2cb50c0f500b52e5fe110298a2b51bcec)

turbolent:
@bluesign The first example case from my post above (field only exists in concrete type) already had a test case. 

The second example case (field exist, but is not `let` or `var`) didn't have a test case, so I added a it in [2d29b42](https://github.com/onflow/cadence/pull/1076/commits/2d29b4288037ed597b25023c1590663d166ad558). It didn't pass, so fixed the implementation. 

Looks like everything is there now! 🎉  

@SupunS could you please have a final look? I'll have a final look as well

runtime/tests/checker/interface_test.go
```diff
@@ -1996,6 +1996,682 @@ func TestCheckInvalidInterfaceUseAsTypeSuggestion(t *testing.T) {
...
+
+	t.Parallel()
+
+	t.Run("interface", func(t *testing.T) {
+
+		t.Parallel()
+
+		_, err := ParseAndCheck(t, `
+          struct interface IA {
+              x: Int
```
SupunS, line 2545: oh, never knew defining a field without var/let is valid in an interface 🤔

SupunS (approved):
LGTM! 👌

turbolent:
Added some minimal documentation

turbolent (approved):
LGTM too. Thanks everyone!

bluesign:
Thanks a lot @turbolent, for explanation and all the work. I am really glad this feature finally will be out.

psiemens:
This is so exciting! Great work everybody! 👏

PR #1090, opened 2021-07-27
Add public account contracts

### LINKED ISSUE #1075: PublicAccount contracts and contract names
### Issue To Be Solved

Currently it is only possible to get a known deployed contract from an authorized account (`AuthAccount.Contracts.get`). 

`PublicAccount` does not provide a way to get a deployed contract.

In addition, it would be nice to also query the names of all deployed contracts of any account.

### Suggest A Solution

- [ ] Add a function `fun getNames(): [String]` to `AuthAccount.Contracts`, which returns the names of all deployed contracts of the account
  - [ ] This requires adding a function (e.g. `GetAccountContractNames`) to the `runtime.Interface` and implementing it in flow-go's FVM. The data is already available there
- [ ] Add the `PublicAccount.Contracts` type (similar to `AuthAccount.Contracts`), which only exposes a `get` function and a `getNames` function like in `AuthAccount.Contracts`

### Context

We need this functionality to evaluate the impact of changes made to Cadence on contracts deployed to the existing networks. Similar to the Source Compatiblity Suite, we would query all contracts deployed to a network, and parse and check them first with the current version, and then again with the modified version of Cadence. This way we can detect potential regressions in the parser and checker

SupunS:
Closes #1075

## Description

This PR:
- Adds a `contracts` field to public account to retrieve all contracts deployed under the account.
- Adds a `get()` method to get a deployed contract by name
- Add a `names` field to `PublicAccount.Contracts` to retrieve the names of all contracts deployed under the account.

______

### DIFF
--- a/runtime/tests/checker/account_test.go
+++ b/runtime/tests/checker/account_test.go
@@ -1417,4 +1417,50 @@ func TestPublicAccountContracts(t *testing.T) {
 		assert.IsType(t, &sema.InvalidAssignmentAccessError{}, errors[0])
 		assert.IsType(t, &sema.AssignmentToConstantMemberError{}, errors[1])
 	})
+
+	t.Run("add contract", func(t *testing.T) {
+		_, err := ParseAndCheckAccount(t, `
+            fun test() {
+                publicAccount.contracts.add(name: "foo", code: "012".decodeHex())
+            }
+	    `)
+
+		require.Error(t, err)
+		errors := ExpectCheckerErrors(t, err, 1)
+
+		require.IsType(t, &sema.NotDeclaredMemberError{}, errors[0])
+		notDeclaredError := errors[0].(*sema.NotDeclaredMemberError)
+		assert.Equal(t, "add", notDeclaredError.Name)
+	})
+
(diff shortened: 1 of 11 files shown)


runtime/tests/checker/account_test.go
```diff
@@ -1383,3 +1383,38 @@ func TestAuthAccountContracts(t *testing.T) {
 		assert.IsType(t, &sema.AssignmentToConstantMemberError{}, errors[1])
 	})
 }
+
+func TestPublicAccountContracts(t *testing.T) {
```
turbolent, line 1387: Could you please add tests that attempt to call mutating functions, i.e `add`, `update`, and `remove`, and fail? Just to make sure that we don't accidentally allow that (now or in the future)

turbolent (approved):
Looks great! Just one comment to add some more negative tests to ensure we don't allow unauthorized contract updates

runtime/tests/checker/account_test.go
SupunS: Thats a good idea! Added in 3cd8f8f

PR #1103, opened 2021-08-04
Add container mutation check for arrays and dictionaries

### LINKED ISSUE #1098: Panic when value is inserted into container with incorrect type
### Issue To Be Solved

It should not be possible to store a value in a container that is not a subtype of the container's static type.

### Suggested Solution

When a value is inserted into a container (array or dictionary), check that the dynamic type of the inserted value is a subtype of the containers static type, i.e array static type's element type, or dictionary static type's key/value type.

SupunS:
Closes #1098

______


runtime/interpreter/value.go
```diff
@@ -9508,3 +9524,22 @@ func NewPublicAccountKeysValue(getFunction FunctionValue) *CompositeValue {
 		fields:              fields,
 	}
 }
+
+func checkContainerMutation(
```
turbolent, line 9558: Minor: Maybe make the interpreter the receiver, i.e. `func (inter *Interpreter) checkContainerMutation(...`

runtime/interpreter/value.go
```diff
@@ -9508,3 +9524,22 @@ func NewPublicAccountKeysValue(getFunction FunctionValue) *CompositeValue {
...
+	memberStaticType StaticType,
+	value Value,
+	getLocationRange func() LocationRange,
+) {
+
+	value = inter.unbox(value)
+
+	memberType := inter.ConvertStaticToSemaType(memberStaticType)
+
+	if !inter.IsSubType(value.DynamicType(inter, DynamicTypeResults{}), memberType) {
```
turbolent, line 9567: 👍

turbolent (approved):
Nice!

runtime/runtime_test.go
```diff
@@ -917,33 +917,6 @@ func TestRuntimeTransactionWithArguments(t *testing.T) {
 			},
 			expectedLogs: []string{"42"},
 		},
-		{
```
turbolent, line 920: Why was this not valid before? 🤔

turbolent (approved):
Great work! 👍

runtime/runtime_test.go
SupunS: Oh this is still invalid. However, this now returns an error at the value conversion step (even before args are validated for types), because we now cannot insert the invalid element into the dictionary anymore. 

But the catch is, when I implemented it, the value-conversion used to panic (didn't return errors), and therefore cannot assert it here like the rest of the tests. So I removed it from here, since it is already covered in the below tests:  https://github.com/onflow/cadence/pull/1103/files#diff-863660c79fae3d9b381c2dff3025cd88dc46eb62571b754e7695acbfeab4fbfeR2186

turbolent: Oh, I totally misread the code. The script is fine, it is the argument / value that is invalid!

Also, thank you for the explanation why it got removed, that makes sense 👍

runtime/interpreter/value.go
SupunS: Added with 4af18e7

turbolent (approved):
Nice!

PR #1113, opened 2021-08-13
introduce TokenStream interface

### LINKED ISSUE #1094: Extract token stream interface from parser
### Issue To Be Solved

The parser-guided fuzzer needs to have a couple alternate implementations of token buffering and token access, and it needs to be switchable at time of Parse().  Previously I tried to use a buildtag to isolate the different implementations, but the parser-guided fuzzer needs to switch at runtime, and needs more than 2 implementations which makes the buildtag approach even more difficult.

### Tradeoff

* There could be a performance hit due to the indirection, but existing prototype work shows no noticeable change in the running time for the full cadence test suite.

### Context

The parser-guided-fuzzer is blocked on this.

jsatdapr:
Closes #1094

## Description

We want to write fuzzers with custom TokenStream implementations.  This PR introduces a TokenStream interface and a default implementation.
______


SupunS (commented):
Looks good! just a couple of questions/comments.

runtime/parser2/lexer/lexer.go
```diff
@@ -74,9 +76,44 @@ type lexer struct {
...
+			Range: ast.Range{
+				StartPos: pos,
+				EndPos:   pos,
+			},
+		}
+	}
+	return token
+}
+
+func (l *lexer) Close(_ *[]error) {
```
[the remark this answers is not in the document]
jsatdapr: We don't need errors now, true.  I will propose it again on another PR when it becomes necessary.

runtime/parser2/lexer/lexer.go
```diff
@@ -74,9 +76,44 @@ type lexer struct {
 	startPos position
 }
 
-func Lex(ctx context.Context, input string) chan Token {
+type TokenStream interface {
+	Next() Token
```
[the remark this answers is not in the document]
jsatdapr: Added documentation for Next and others.

runtime/parser2/lexer/lexer.go
```diff
@@ -74,9 +76,47 @@ type lexer struct {
...
+	Code() string
+}
+
+func (l *lexer) Next() Token {
+	token, ok := <-l.tokens
+	if !ok {
+		pos := ast.Position{
+			Offset: l.endOffset - 1,
+			Line:   l.endPos().line,
+			Column: l.endPos().column - 1,
```
turbolent, line 94: Maybe refactor `l.endPos()` into a variable, so it is only computed once

runtime/parser2/lexer/lexer.go
```diff
@@ -74,9 +76,47 @@ type lexer struct {
...
+	}
+	return token
+}
+
+func (l *lexer) Close() {
+	l.cancelLexer()
+}
+
+func (l *lexer) Code() string {
+	return l.input
```
turbolent, line 112: From the documentation of `Code` it seemed to me at first that the function would only return the lexed code up-to when `Close` was called

turbolent (approved):
Nice!

runtime/parser2/lexer/lexer.go
turbolent: Refactored in fd6b5b3

runtime/parser2/lexer/lexer.go
turbolent: 313b77b

PR #1114, opened 2021-08-13
Remove obsolete script/transaction parameter check

### LINKED ISSUE #1099: Allow AnyStruct as script parameter type
## Issue To Be Solved

Currently it is not possible to declare a script which has a `AnyStruct` or `AnyResource` parameter type.

## Suggested Solution

The check following check should probably be removed / replace:

https://github.com/onflow/cadence/blob/fa0c3f625056cbaf001acc8567658c8aaabdbffb/runtime/interpreter/interpreter.go#L796-L802

- `AnyStruct` values should be allowed, as the argument validation should now already check if the value is importable
- `AnyResource` is not enough? Reject all resource types (i.e. check `IsResourceType()` instead?

SupunS:
Closes #1099

## Description

This check is now obsolete as we do check for importability of types at: 
https://github.com/onflow/cadence/blob/fa0c3f625056cbaf001acc8567658c8aaabdbffb/runtime/runtime.go#L214-L223

There are already tests available to cover the different cases:
https://github.com/onflow/cadence/blob/fa0c3f625056cbaf001acc8567658c8aaabdbffb/runtime/program_params_validation_test.go#L180-L194
______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -793,13 +793,6 @@ func (interpreter *Interpreter) prepareInvoke(
 	preparedArguments := make([]Value, len(arguments))
 	for i, argument := range arguments {
 		parameterType := parameters[i].TypeAnnotation.Type
-		// TODO: value type is not known, reject for now
-		switch parameterType {
-		case sema.AnyStructType, sema.AnyResourceType:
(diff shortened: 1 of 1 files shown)


turbolent (approved):
Looks good! Would be nice to get this into 0.19

PR #1125, opened 2021-08-24
Add container static types

### LINKED ISSUE #870: Implement `StaticType()` for arrays and dictionaries

SupunS:
Closes #870

## Description

Add static type information to containers (arrays and dictionaries)

______


runtime/interpreter/value.go
```diff
@@ -680,6 +683,8 @@ func NewDeferredArrayValue(
 	version uint16,
 ) *ArrayValue {
 	return &ArrayValue{
+		// TODO: type
```
turbolent, line 686: Looks like we missed a TODO here. Could you please resolve it?

runtime/interpreter/value.go
```diff
@@ -7354,6 +7432,8 @@ func NewDeferredDictionaryValue(
 	version uint16,
 ) *DictionaryValue {
 	return &DictionaryValue{
+		// TODO: type
```
turbolent, line 7435: Looks like we missed a TODO here for dictionaries too. Could you please resolve it?

turbolent (commented):
Looks like there are two, similar, TODOs that I think we should resolve before merging.

Other than that, I think we already reviewed all the PRs individually, so don't need a full re-review of it all, and a quick look over it also looks good.

Can't wait to ⛵ this!

runtime/interpreter/value.go
SupunS: Ah right! We can simply remove the TODO (+ field initializing). Static type is populated when the content is loaded to memory, so we dont need to provide it at the constructor. Same for dictionary as well.

turbolent: Ah, right! 👍

SupunS: Removed it with a comment: 5b2c6ce

turbolent: Thank you for resolving them and also adding comments! 👌

turbolent (approved):
🎉

PR #1138, opened 2021-09-10
Add limit on call stack depth

SupunS:
## Description

Porting https://github.com/dapperlabs/cadence-internal/pull/32

______

### DIFF
--- a/runtime/errors.go
+++ b/runtime/errors.go
@@ -74,6 +74,19 @@ func (e ComputationLimitExceededError) Error() string {
 	)
 }
 
+// CallStackLimitExceededError
+
+type CallStackLimitExceededError struct {
+	Limit uint64
+}
+
+func (e CallStackLimitExceededError) Error() string {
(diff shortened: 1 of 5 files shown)


turbolent:
Even though it turned out that the call stack depth wasn't the real problem we saw on Testnet, it might be still useful to have this as a feature. Maybe we can merge it and disable it for now?

SupunS:
Yeah, that was my feeling as well. It is good to enforce it from the language in a first-class manner. 

What if we still keep it, but with a slightly higher value? (e.g: 2000)

turbolent:
Yes, I think it's good to merge, and agreed 2000 should be OK. I tagged it as a breaking change, because we should flag this change of behaviour to users/developers

turbolent (approved):

turbolent:
It would be good to make this limit configurable, like the computation limit (but in a separate PR)

PR #1144, opened 2021-09-23
 Add static type to host functions

### LINKED ISSUE #1136: Add static types to HostFunctionValue
### Issue To Be Solved
Currently, host functions don't have static types

### Suggested Solution

Pass the static type during construction of host function value and store it in value itself.

SupunS:
Closes #1136

______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -2640,6 +2640,7 @@ var stringFunction = func() Value {
 				bytes, _ := ByteArrayValueToByteSlice(argument)
 				return NewStringValue(hex.EncodeToString(bytes))
 			},
+			sema.StringTypeEncodeHexFunctionType,
 		),
 	)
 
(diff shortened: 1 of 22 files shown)


runtime/interpreter/interpreter.go
```diff
@@ -2175,50 +2188,56 @@ func (interpreter *Interpreter) functionConditionsWrapper(
...
+					preConditions,
+					body,
+					rewrittenPostConditions,
+					returnType,
+				)
+			},
+
+			// This is an internally created and used function, and can
+			// never pass around as a value. Hence, type is not required.
+			nil,
```
turbolent, line 2239: 👍

runtime/interpreter/interpreter.go
```diff
@@ -2175,50 +2188,56 @@ func (interpreter *Interpreter) functionConditionsWrapper(
...
+					beforeStatements,
+					preConditions,
+					body,
+					rewrittenPostConditions,
+					returnType,
+				)
+			},
+
+			// This is an internally created and used function, and can
+			// never pass around as a value. Hence, type is not required.
```
turbolent, line 2238:
```suggestion
			// never be passed around as a value. Hence, the type is not required.
```

runtime/interpreter/interpreter.go
```diff
@@ -2580,6 +2599,10 @@ var converterFunctionValues = func() []converterFunction {
 			func(invocation Invocation) Value {
 				return convert(invocation.Arguments[0])
 			},
+
+			// Converter functions are not pass around as values.
+			// Hence, type is not required.
```
turbolent, line 2604:
```suggestion
			// Converter functions are not passed around as values.
			// Hence, the type is not required.
```

runtime/tests/interpreter/container_mutation_test.go
```diff
@@ -258,6 +259,104 @@ func TestArrayMutation(t *testing.T) {
...
+				},
+				Options: []interpreter.Option{
+					interpreter.WithPredeclaredValues(values),
+				},
+			},
+		)
+
+		require.NoError(t, err)
+
+		// TODO: Shouldn't throw an error once dynamic subtyping for functions is implemented.
```
turbolent, line 294: Could you please create a follow-up issue for this TODO?

turbolent (approved):
Great work! 👏 

It might be worth it to cache the result of the `ConvertStaticToSemaType` in the values to avoid convering them each time. Maybe best to do this in a follow-up PR

runtime/tests/interpreter/container_mutation_test.go
SupunS: Created: https://github.com/onflow/cadence/issues/1145

SupunS:
> It might be worth it to cache the result of the ConvertStaticToSemaType in the values to avoid converting the static types to sema types each time. Maybe best to do this in a follow-up PR

For function types, it shouldn't be too much of an overhead, since it only creates a thin wrapper. But definitely, we could cache these, and also cache parameter types, return types, etc in the `FunctionStaticType`.

PR #1146, opened 2021-09-23
Fix array function member type creation

SupunS:
## Description

For deferred values, the static type may not have been built by the time the members are accessed. So, use `v.StaticType()` function which ensures the meta info is built.

The same is already handled properly for dictionaries and composites.

______

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -1012,7 +1012,11 @@ func (v *ArrayValue) GetMember(inter *Interpreter, getLocationRange func() Locat
 				v.Append(inter, getLocationRange, invocation.Arguments[0])
 				return VoidValue{}
 			},
-			sema.ArrayAppendFunctionType(inter.ConvertStaticToSemaType(v.Type.ElementType())),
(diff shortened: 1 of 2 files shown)


runtime/interpreter/value.go
```diff
@@ -1012,7 +1012,12 @@ func (v *ArrayValue) GetMember(inter *Interpreter, getLocationRange func() Locat
 				v.Append(inter, getLocationRange, invocation.Arguments[0])
 				return VoidValue{}
 			},
-			sema.ArrayAppendFunctionType(inter.ConvertStaticToSemaType(v.Type.ElementType())),
+			func() *sema.FunctionType {
+				arrayStaticType := v.StaticType().(ArrayStaticType)
```
turbolent, line 1016: Maybe just inline this variable, as it is only used once, and the function wrapper might not be optimized away (just guessing here)

turbolent (approved):
👍

runtime/interpreter/value.go
SupunS: Yea, I separate them out for better readability, but makes sense to inline

PR #1018, opened 2021-06-18
Reuse base activation across all interpreters

SupunS:
## Description
While checking for interpreter reusability, I noticed that we could avoid re-registering the built-in values/functions, and reuse them across all interpreters.

**Aproach:**

Initialize a base activation only once. Then reuse the base activation as the root of all the activations created for the new interpreters.

**Perf Impact:**
```
name                                  old time/op    new time/op    delta
NewInterpreter/new_interpreter-4        7.86µs ± 0%    0.96µs ± 0%  -87.83%  (p=0.008 n=5+5)
NewInterpreter/new_sub-interpreter-4    8.56µs ± 1%    1.61µs ± 2%  -81.21%  (p=0.008 n=5+5)

name                                  old alloc/op   new alloc/op   delta
NewInterpreter/new_interpreter-4        4.94kB ± 0%    0.68kB ± 0%  -86.24%  (p=0.008 n=5+5)
NewInterpreter/new_sub-interpreter-4    5.25kB ± 0%    0.98kB ± 0%  -81.33%  (p=0.079 n=4+5)

name                                  old allocs/op  new allocs/op  delta
NewInterpreter/new_interpreter-4          65.0 ± 0%      12.0 ± 0%  -81.54%  (p=0.008 n=5+5)
NewInterpreter/new_sub-interpreter-4      82.0 ± 0%      29.0 ± 0%  -64.63%  (p=0.008 n=5+5)
```
______

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #1149, opened 2021-09-27
Cache host function static type

### LINKED ISSUE #1147: Cache static types for function
### Issue To Be Solved

Static types for function were added with: https://github.com/onflow/cadence/pull/1144

Cache the created types in the value as an optimization
  - Cache static type in the value
  - Cache parameter/return/etc types in the static function type.
  - Do the same for other places, if any

SupunS:
Closes #1147

______

### DIFF
--- a/runtime/interpreter/function.go
+++ b/runtime/interpreter/function.go
@@ -20,6 +20,7 @@ package interpreter
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/onflow/cadence/runtime/ast"
 	"github.com/onflow/cadence/runtime/common"
@@ -61,6 +62,8 @@ type InterpretedFunctionValue struct {
 	PreConditions    ast.Conditions
 	Statements       []ast.Statement
 	PostConditions   ast.Conditions
+	staticType       StaticType
+	staticTypeOnce   sync.Once
 }
 
 func (f *InterpretedFunctionValue) String() string {
@@ -88,7 +91,11 @@ func (*InterpretedFunctionValue) DynamicType(_ *Interpreter, _ SeenReferences) D
 }
 
 func (f *InterpretedFunctionValue) StaticType() StaticType {
-	return ConvertSemaToStaticType(f.Type)
+	f.staticTypeOnce.Do(func() {
+		f.staticType = ConvertSemaToStaticType(f.Type)
+	})
+
+	return f.staticType
 }
 
 func (f *InterpretedFunctionValue) Copy() Value {
@@ -140,6 +147,8 @@ type HostFunctionValue struct {
 	Function        HostFunction
 	NestedVariables *StringVariableOrderedMap
 	Type            *sema.FunctionType
+	staticType      StaticType
+	staticTypeOnce  sync.Once
 }
 
 func (f *HostFunctionValue) String() string {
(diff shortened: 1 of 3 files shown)


runtime/interpreter/function.go
```diff
@@ -61,6 +62,8 @@ type InterpretedFunctionValue struct {
 	PreConditions    ast.Conditions
 	Statements       []ast.Statement
 	PostConditions   ast.Conditions
+	staticType       StaticType
+	staticTypeOnce   sync.Once
```
turbolent, line 66: There is no concurrency in the interpreter and the interpreter is not used in a concurrent way, so using sync.Once is not necessary (here and for the other cases below)

SupunS: There are some functions that we use across interpreters (e.g: [typeFunction](https://github.com/onflow/cadence/blob/master/runtime/interpreter/interpreter.go#L2638-L2640), etc ) could that cause issues?

turbolent: Ah, that's a good point, I didn't think of that! I guess you are right, which is unfortunate. I guess we should maybe refactor things like that to be independent for each interpreter, instead globally shared, potentially even used concurrently

turbolent:
@SupunS Would you be able to post a benchmark comparison, just to make sure this does not actually regress performance? I'm worried the locking might have a big overhead, potentially more than the gain from caching

SupunS:
Yeah, checking that now. Also removing locking means we have to remove the base-activation which is a singleton now, which might affect runtime performance (we made it a singleton as an optimization). Running a benchmark for that as well

SupunS:
@turbolent These are the benchmark results with locks. (_old_ - with locks, _new_ - without locks)
```
name                                  old time/op    new time/op    delta
FunctionStaticType/Get_type-8           2.12ns ± 1%    2.10ns ± 1%     ~     (p=0.063 n=4+5)
FunctionStaticType/Get_inner_types-8    5.75ns ± 0%    3.71ns ± 1%  -35.58%  (p=0.016 n=4+5)

name                                  old alloc/op   new alloc/op   delta
FunctionStaticType/Get_type-8            0.00B          0.00B          ~     (all equal)
FunctionStaticType/Get_inner_types-8     0.00B          0.00B          ~     (all equal)

name                                  old allocs/op  new allocs/op  delta
FunctionStaticType/Get_type-8             0.00           0.00          ~     (all equal)
FunctionStaticType/Get_inner_types-8      0.00           0.00          ~     (all equal)
```

But that would mean we have to make the base activation non-singleton. This means newing the interpreter takes a lot of time:
```
name                                  old time/op    new time/op    delta
NewInterpreter/new_interpreter-8        1.18µs ± 3%   15.37µs ± 1%  +1197.32%  (p=0.008 n=5+5)
NewInterpreter/new_sub-interpreter-8    1.86µs ± 1%   16.32µs ± 3%   +779.10%  (p=0.008 n=5+5)

name                                  old alloc/op   new alloc/op   delta
NewInterpreter/new_interpreter-8          912B ± 0%    16711B ± 0%  +1732.39%  (p=0.008 n=5+5)
NewInterpreter/new_sub-interpreter-8    1.25kB ± 0%   17.05kB ± 0%  +1265.95%  (p=0.008 n=5+5)

name                                  old allocs/op  new allocs/op  delta
NewInterpreter/new_interpreter-8          14.0 ± 0%     286.0 ± 0%  +1942.86%  (p=0.008 n=5+5)
NewInterpreter/new_sub-interpreter-8      33.0 ± 0%     305.0 ± 0%   +824.24%  (p=0.008 n=5+5)
```
Basically undoing the changes in: https://github.com/onflow/cadence/pull/1018

turbolent:
@SupunS and I synced offline:

The first benchmark compares:
- current master: always re-computing, no caching, and no locks
- this PR: caching, with locks
The later is slower. 

The second benchmark is for an alternative to caching, where we don't use locks, but then have to undo the base activation optimization. This would be a dramatic regression

We decided to close this PR and leave the current non-caching code on master as-is.

PR #1158, opened 2021-10-01
Consider constant-sized type's size when checking dynamic array type

turbolent:
______

### DIFF
--- a/runtime/convertValues_test.go
+++ b/runtime/convertValues_test.go
@@ -1711,7 +1711,7 @@ func TestArgumentPassing(t *testing.T) {
 		// TODO: Enable below once https://github.com/onflow/cadence/issues/712 is fixed.
 		// TODO: Add a malformed argument test for capabilities
 		//{
-		//    label:         "Capability",
+		//    name:         "Capability",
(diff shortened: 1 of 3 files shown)


runtime/convertValues_test.go
```diff
@@ -2170,19 +2170,37 @@ func TestMalformedArgumentPassing(t *testing.T) {
...
 			}),
 			expectedErrType: &MalformedValueError{},
 		},
 		{
-			label:         "Array with wrong size",
-			typeSignature: "[String; 2]",
+			label:         "Constant-size array with malformed element",
+			typeSignature: "[Foo; 1]",
 			exportedValue: cadence.NewArray([]cadence.Value{
 				malformedStruct1,
```
SupunS, line 2184: oh so this was failing before reaching the size check (which wasn't there), because the inner element is malformed?

SupunS (approved):
Good catch!

runtime/convertValues_test.go
turbolent: Correct, the test was failing, but for the "wrong reason"

PR #1175, opened 2021-10-14
Remove CompositeDeclarations check from FunctionEntryPointDeclaration

janezpodhostnik:
References https://github.com/onflow/flow-cli/issues/395

## Description

It is very useful for a script to define a Resource and use it as a return type.

While this is already possible, the checker did not correctly return the `FunctionEntryPointDeclaration` if this was the case.

This was reported as a bug in the CLI https://github.com/onflow/flow-cli/issues/395, but  can also be seen in the playground.

______

### DIFF
--- a/runtime/sema/entrypoint.go
+++ b/runtime/sema/entrypoint.go
@@ -26,7 +26,7 @@ const FunctionEntryPointName = "main"
 
 // FunctionEntryPointDeclaration returns the entry point function declaration, if any.
 //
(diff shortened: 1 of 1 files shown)


turbolent (approved):
Good point! Thank you for fixing this

@MaxStalker had asked me about this the other day and I had looked into it: I couldn't remember why I had been so defensive here, probably mostly thinking about transactions, where it is not allowed, though it is allowed in scripts and actually often used.

PR #1202, opened 2021-10-28
Allow importing type and capability values

### LINKED ISSUE #491: Allow import of type values (`cadence.TypeValue`)
### Issue To Be Solved

It should be possible to import type values (`cadence.TypeValue`), e.g. when sending them as an argument in a transaction or script.

### Suggested Solution

- [ ] Add support for importing a `cadence.TypeValue`
  - [ ] Convert `cadence.TypeValue` in `importValue` to `interpreter.TypeValue`
  - [ ] Check that the type exists, fail if not
- [ ] Tests

dsainati1:
Closes #491, #712

## Description

This does two primary things:

1) Change the representation of types in `cadence.TypeValue` and `cadence.Capability` to be a `cadence.Type` instead of a `string`. This also requires implementation of encoding and decoding types to and from JSON. 

2) Implement `ImportType` to convert a `cadence.Type` into a `interpreter.StaticType`

3) Update importing to allow for the importing of capability and type values
______


runtime/convertValues.go
```diff
@@ -457,6 +469,43 @@ func importPathValue(v cadence.Path) interpreter.PathValue {
...
+	address cadence.Address,
+	borrowType cadence.Type,
+) (
+	interpreter.CapabilityValue,
+	error,
+) {
+
+	switch borrowType.(type) {
+	case cadence.ReferenceType:
+		return interpreter.CapabilityValue{
```
turbolent, line 496: Careful here: It is only safe to import **public** capabilities, i.e. ones with a public path. See the description of #712. 

The tests below only check that the import of a capability with a storage path succeeds, which should not. Could you please also add tests for the other path domains and assert that only the import of public ones succeeds?

I'm not quote sure if this should be checked here. 
@SupunS Where would the proper place be? I remember you had worked on checking imported values before, but I don't quite remember the mechanics.

runtime/convertValues.go
```diff
@@ -457,6 +469,43 @@ func importPathValue(v cadence.Path) interpreter.PathValue {
...
+			Path:       importPathValue(path),
+			Address:    interpreter.NewAddressValueFromBytes(address.Bytes()),
+			BorrowType: ImportType(borrowType),
+		}, nil
+	}
+	return interpreter.CapabilityValue{}, fmt.Errorf(
+		"cannot import capability of type '%s', '%s' must be a reference type",
+		borrowType.ID(),
+		borrowType.ID(),
+	)
```
turbolent, line 506: Instead of a single-case switch statement, just use a conditional cast.

Also, bail out early instead of nesting the happy path, i.e. invert the logic, e.g.

```suggestion
	_, ok := borrowType.(cadence.ReferenceType)
	if !ok {
		return interpreter.CapabilityValue{}, fmt.Errorf(
			"cannot import capability: expected reference, got '%s'",
			borrowType.ID(),
		)
	}
	return interpreter.CapabilityValue{
		Path:       importPathValue(path),
		Address:    interpreter.NewAddressValueFromBytes(address.Bytes()),
		BorrowType: ImportType(borrowType),
	}, nil
```

turbolent (commented):
Nice work also adding support for importing capabilities

runtime/convertValues.go
SupunS: I think that validation should probably go to `CapabilityDynamicType.IsImportable()`. Checking `isImportable` is already done when importing values, so it should get automatically validated there. We would have to include pathType info to the `CapabilityDynamicType` to make it work.

runtime/convertValues_test.go
```diff
@@ -2961,6 +3374,339 @@ func TestRuntimeStringValueImport(t *testing.T) {
...
+				Interface: runtimeInterface,
+				Location:  TestLocation,
+			},
+		)
+
+		require.Error(t, err)
+	})
+}
+
+func TestCapabilityValueImport(t *testing.T) {
```
SupunS, line 3466: Is it possible to merge these test cases into `TestRuntimeArgumentPassing` and `TestRuntimeMalformedArgumentPassing`? or does this perform any additional checks?

runtime/convertValues_test.go
```diff
@@ -2961,6 +3374,339 @@ func TestRuntimeStringValueImport(t *testing.T) {
...
+		require.NoError(t, err)
+
+		rt := NewInterpreterRuntime()
+
+		runtimeInterface := &testRuntimeInterface{
+			decodeArgument: func(b []byte, t cadence.Type) (value cadence.Value, err error) {
+				return json.Decode(b)
+			},
+			log: func(s string) {
+				assert.Equal(t, s, "Capability<&Int>(address: 0x100000000000000, path: /public/foo)")
```
SupunS, line 3499: Similar to above, might want to add another assertion to make sure this gets executed

runtime/convertValues_test.go
```diff
@@ -2961,6 +3374,339 @@ func TestRuntimeStringValueImport(t *testing.T) {
...
+		encodedArg, err := json.Encode(typeValue)
+		require.NoError(t, err)
+
+		rt := NewInterpreterRuntime()
+
+		runtimeInterface := &testRuntimeInterface{
+			decodeArgument: func(b []byte, t cadence.Type) (value cadence.Value, err error) {
+				return json.Decode(b)
+			},
+			log: func(s string) {
```
SupunS, line 3447: I guess we wouldn't need the log functions (also in the script), since the test would fail even before it reaches there. Also in other places.

runtime/convertValues_test.go
```diff
@@ -2961,6 +3374,339 @@ func TestRuntimeStringValueImport(t *testing.T) {
...
+				Source:    []byte(script),
+				Arguments: [][]byte{encodedArg},
+			},
+			Context{
+				Interface: runtimeInterface,
+				Location:  TestLocation,
+			},
+		)
+
+		require.Error(t, err)
```
SupunS, line 3462: Maybe assert the error type is `MalformedValueError`

runtime/convertValues_test.go
```diff
@@ -2961,6 +3374,339 @@ func TestRuntimeStringValueImport(t *testing.T) {
...
+	})
+
+	t.Run("missing", func(t *testing.T) {
```
SupunS, line 3655: Shall we rename `missing` it to be more intuitive of what the test case checks?

SupunS (commented):
Looks good!

Shall we also address this TODO and the one below that: https://github.com/onflow/cadence/blob/master/runtime/convertValues_test.go#L1840-L1841
This was added earlier when we didn't have the type/capability importing support.

runtime/convertValues_test.go
dsainati1: The those tests generally seem to test simpler types, so for these more complex ones I want to keep them separate. I got rid of the TODOs above though since they are now redundant.

runtime/convertValues.go
```diff
@@ -550,6 +561,45 @@ func importPathValue(v cadence.Path) interpreter.PathValue {
...
+	borrowType cadence.Type,
+) (
+	*interpreter.CapabilityValue,
+	error,
+) {
+
+	_, ok := borrowType.(cadence.ReferenceType)
+
+	if !ok {
+		return &interpreter.CapabilityValue{}, fmt.Errorf(
```
turbolent, line 589: It's more idiomatic to return nil, which also avoids an unnecessary allocation: 

```suggestion
		return nil, fmt.Errorf(
```

runtime/convertValues_test.go
```diff
@@ -1243,7 +1651,7 @@ func TestExportTypeValue(t *testing.T) {
 
 		actual := exportValueFromScript(t, script)
 		expected := cadence.TypeValue{
-			StaticType: "S.test.S",
+			StaticType: &cadence.StructType{QualifiedIdentifier: "S", Location: TestLocation, Fields: []cadence.Field{}},
```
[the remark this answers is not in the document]
turbolent: Maybe break lines here like Supun suggested to make this more readable

runtime/convertValues_test.go
```diff
@@ -1304,7 +1712,12 @@ func TestExportTypeValue(t *testing.T) {
 
 		assert.Equal(t,
 			cadence.TypeValue{
-				StaticType: "S.test.S{S.test.SI}",
+				StaticType: cadence.RestrictedType{
+					Type: &cadence.StructType{QualifiedIdentifier: "S", Location: TestLocation, Fields: []cadence.Field{}},
+					Restrictions: []cadence.Type{
+						&cadence.StructInterfaceType{QualifiedIdentifier: "SI", Location: TestLocation, Fields: []cadence.Field{}},
+					},
```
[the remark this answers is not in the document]
turbolent: Same here, maybe break lines as suggested

turbolent (commented):
Looks good! The PR is only missing some type conversions, which could also be done in a follow up PR, and some of Supun suggestions are still open

runtime/convertTypes.go
```diff
@@ -386,5 +390,157 @@ func exportCapabilityType(t *sema.CapabilityType, results map[sema.TypeID]cadenc
...
+	case cadence.PublicPathType:
+		return interpreter.PrimitiveStaticTypePublicPath
+	case cadence.PrivatePathType:
+		return interpreter.PrimitiveStaticTypePrivatePath
+	case cadence.CapabilityType:
+		return interpreter.CapabilityStaticType{
+			BorrowType: ImportType(t.BorrowType),
+		}
+	}
+	panic(fmt.Sprintf("cannot export type of type %T", t))
```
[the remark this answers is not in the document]
dsainati1: I will do this in a follow up PR, I filed #1219 to document this, and will put up a PR fixing this later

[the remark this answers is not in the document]
turbolent: Perfect, thank you!

runtime/convertValues_test.go
```diff
@@ -2961,6 +3365,335 @@ func TestRuntimeStringValueImport(t *testing.T) {
...
+				Source:    []byte(script),
+				Arguments: [][]byte{encodedArg},
+			},
+			Context{
+				Interface: runtimeInterface,
+				Location:  TestLocation,
+			},
+		)
+
+		require.Error(t, err)
```
turbolent, line 3454: Looks like the last commit reverted the assertion suggestion again: [`bec485a` (#1202)](https://github.com/onflow/cadence/pull/1202/commits/bec485ae1d813bf91c260c231067ad19fd6b5c8f#diff-863660c79fae3d9b381c2dff3025cd88dc46eb62571b754e7695acbfeab4fbfeL3454-R3454)

runtime/interpreter/interpreter.go
```diff
@@ -3746,26 +3746,26 @@ func (interpreter *Interpreter) GetCapabilityFinalTargetStorageKey(
 	}
 }
 
-func (interpreter *Interpreter) ConvertStaticToSemaType(staticType StaticType) sema.Type {
+func (interpreter *Interpreter) OptionallyConvertStaticToSemaType(staticType StaticType) (sema.Type, error) {
```
turbolent, line 3749: Usually the "Go idiomatic" naming convention is to name the function that returns an error without a prefix, and the one that panics with a `Must` prefix, but this also works

turbolent (approved):
@dsainati1 This looks like it is working now, but I still suggest applying Supun's code formatting suggestions. We try to keep lines short for readability reasons

PR #1208, opened 2021-11-01
Fix integer conversion

### LINKED ISSUE #1126: Number Types Convert Incorrectly
This code:

```
  log(Int(UInt64(1)))
  log(Int(UInt64(18446744073709551615)))
```

logs:

```
18:20:43 Script 1
18:20:43 Script -1
```

https://play.onflow.org/f7457773-a35b-41f9-9296-cb0d73add49d?type=script&id=247df2fe-9acd-48bd-bba0-6f232801e478

That behaviour is unexpected. I have just encountered some code relying on it, and I have warned the developer that they should not do so.

@turbolent (thank you for looking at this!) says:

> I think this is a bug, caused by an internal conversion from uint64 to int, which silently overflows in Go

and that we should clean up code in the golang codebase marked:

> `// TODO: handle overflow`

I would like to attempt to fix this but I am slammed this week. If this can wait I can take it on, but if not someone else may need to.

turbolent:
Closes #1126 

## Description

- Fix conversion of integers > `math.MaxInt64`: 
  Implement `BigNumberValue` for `UInt64` and `Word64`
- Handle overflow/underflow when converting number values to `int` (in `ToInt()`)
- Add an extensive test case which performs number conversions between all integer types, for various values ("typical" value, overflowing, underflowing, and max value)

______

### DIFF
--- a/runtime/tests/interpreter/integers_test.go
+++ b/runtime/tests/interpreter/integers_test.go
@@ -471,7 +471,7 @@ func TestInterpretIntegerConversion(t *testing.T) {
 		inter := parseCheckAndInterpret(t,
 			fmt.Sprintf(
 				`
-				  fun test(value: %[1]s): %[2]s {
+                  fun test(value: %[1]s): %[2]s {
                       return %[2]s(value)
                   }
 				`,
--- a/runtime/tests/interpreter/integers_test.go
+++ b/runtime/tests/interpreter/integers_test.go
@@ -21,8 +21,10 @@ package interpreter_test
 import (
 	"fmt"
 	"math"
+	"math/big"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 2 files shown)


runtime/tests/interpreter/integers_test.go
```diff
@@ -453,6 +455,224 @@ func TestInterpretIntegerLiteralTypeConversionInReturnOptional(t *testing.T) {
...
+		expectedValue interpreter.Value,
+		expectedError error,
+	) {
+
+		inter := parseCheckAndInterpret(t,
+			fmt.Sprintf(
+				`
+				  fun test(value: %[1]s): %[2]s {
+                      return %[2]s(value)
+                  }
```
SupunS, line 476:
```suggestion
                  fun test(value: %[1]s): %[2]s {
                      return %[2]s(value)
                  }
```

SupunS (approved):

PR #1211, opened 2021-11-01
Add converter functions to produce path values from strings

### LINKED ISSUE #1105: Add support for casting String to *Path (StoragePath or CapabilityPath)
### Issue To Be Solved
A convenience method to cast a String to *Path would be convenient and ergonomic from a developer perspective. It would allow for easy dynamic path specification for resources/capabilities, such that one wouldn't need to 'hardcode' entire *Paths.

### Suggested Solution
Add the following methods to String:

* `"/storage/somePath".to_storagePath()`
* `"/public/somePath".to_capabilityPath()`

These methods will simply cast the path contained in the String to their corresponding *Path type.

dsainati1:
Closes #1105

## Description

This adds three converter functions to produce paths from strings:

```
fun PublicPath(_ identifier: string): PublicPath?
fun PrivatePath(_ identifier: string): PrivatePath?
fun StoragePath(_ identifier: string): StoragePath?
```

______


runtime/interpreter/value.go
```diff
@@ -9337,6 +9338,78 @@ func (PathValue) IsStorable() bool {
...
+	}
+
+	return NewSomeValueOwningNonCopying(PathValue{
+		Domain:     common.PathDomainPublic,
+		Identifier: stringValue.Str,
+	})
+}
+
+func ConvertPrivatePath(value Value) Value {
+	stringValue, ok := value.(*StringValue)
```
turbolent, line 9366: Maybe avoid the code duplication in `ConvertPublicPath`, `ConvertPrivatePath`, and `ConvertStoragePath`: They all share the same code, the only difference is the domain. Maybe introduce a helper function that accepts the domain as a parameter

runtime/sema/check_path_expression.go
```diff
@@ -19,40 +19,59 @@
...
+			return &InvalidPathDomainError{
+				ActualDomain: domainString,
+				Range:        ast.NewRangeFromPositioned(expression.Domain),
+			}
+		})
+		return PathType
+	}
+
+	// Check that the identifier is valid
+	isValidIdentifier := regexp.MustCompile(`^[A-Za-z_][A-Za-z0-9_]*$`).MatchString
```
turbolent, line 56: This re-compiles the regular expression on each call of `CheckPathLiteral`. 
Refactor the regular expression out into a global

runtime/sema/check_path_expression.go
```diff
@@ -19,40 +19,59 @@
...
+		func(errFunc func(err *ast.PathExpression) error) {
+			checker.report(errFunc(expression))
+		},
+	)
 
 	return ty
 }
 
-func CheckPathLiteral(expression *ast.PathExpression) (Type, error) {
+func CheckPathLiteral(domainString, identifier string, report func(func(err *ast.PathExpression) error)) Type {
```
turbolent, line 41: Maybe instead of requiring a path expression to produce the error, generalize it to just what is needed to produce the error, the range. That way a caller can e.g. pass an empty location range

runtime/sema/errors.go
```diff
@@ -2703,6 +2703,15 @@ func (e *InvalidPathDomainError) Error() string {
 	return "invalid path domain"
```
turbolent, line 2703: Good idea to add the actual identifier in `InvalidPathIdentifierError.Error`, maybe do the same here

runtime/sema/type.go
```diff
@@ -3219,6 +3219,51 @@ func init() {
...
 		),
 	)
+
+	BaseValueActivation.Set(
+		PublicPathType.String(),
+		baseFunctionVariable(
+			PublicPathType.String(),
+			&FunctionType{
+				Parameters: []*Parameter{{
+					Identifier:     "path",
```
turbolent, line 3229: Naming the parameter "path" looks like it is expected to be a full path, but only an identifier is expected.

```suggestion
					Identifier:     "identifier",
```

runtime/sema/type.go
```diff
@@ -3219,6 +3219,51 @@ func init() {
...
+		),
+	)
+
+	BaseValueActivation.Set(
+		PrivatePathType.String(),
+		baseFunctionVariable(
+			PrivatePathType.String(),
+			&FunctionType{
+				Parameters: []*Parameter{{
+					Identifier:     "path",
```
turbolent, line 3244:
```suggestion
					Identifier:     "identifier",
```

runtime/sema/type.go
```diff
@@ -3219,6 +3219,51 @@ func init() {
...
+		),
+	)
+
+	BaseValueActivation.Set(
+		StoragePathType.String(),
+		baseFunctionVariable(
+			StoragePathType.String(),
+			&FunctionType{
+				Parameters: []*Parameter{{
+					Identifier:     "path",
```
turbolent, line 3259:
```suggestion
					Identifier:     "identifier",
```

runtime/tests/checker/path_test.go
```diff
@@ -105,3 +105,44 @@ func TestCheckPath(t *testing.T) {
 		assert.IsType(t, &sema.InvalidPathDomainError{}, errs[0])
 	})
 }
+
+func TestConvertStringToPath(t *testing.T) {
```
turbolent, line 109:
```suggestion
func TestCheckConvertStringToPath(t *testing.T) {
```

runtime/tests/interpreter/path_test.go
```diff
@@ -55,3 +56,86 @@ func TestInterpretPath(t *testing.T) {
 		})
 	}
 }
+
+func TestConvertStringToPath(t *testing.T) {
```
turbolent, line 60:
```suggestion
func TestInterpretConvertStringToPath(t *testing.T) {
```

runtime/tests/interpreter/path_test.go
```diff
@@ -55,3 +56,86 @@ func TestInterpretPath(t *testing.T) {
...
+		t.Run(fmt.Sprintf("valid: %s", domain.Identifier()), func(t *testing.T) {
+
+			t.Parallel()
+
+			domainType := domainTypes[domain]
+
+			inter := parseCheckAndInterpret(t,
+				fmt.Sprintf(
+					`
+                      let x = %[1]s("foo")!
```
turbolent, line 80:
```suggestion
                      let x = %[1]s(identifier: "foo")!
```

runtime/tests/interpreter/path_test.go
```diff
@@ -55,3 +56,86 @@ func TestInterpretPath(t *testing.T) {
...
+		t.Run(fmt.Sprintf("invalid identifier 2: %s", domain.Identifier()), func(t *testing.T) {
+
+			t.Parallel()
+
+			domainType := domainTypes[domain]
+
+			inter := parseCheckAndInterpret(t,
+				fmt.Sprintf(
+					`
+                      let x = %[1]s("2")
```
turbolent, line 104:
```suggestion
                      let x = %[1]s(identifier: "2")
```

runtime/tests/interpreter/path_test.go
```diff
@@ -55,3 +56,86 @@ func TestInterpretPath(t *testing.T) {
...
+		t.Run(fmt.Sprintf("invalid identifier -: %s", domain.Identifier()), func(t *testing.T) {
+
+			t.Parallel()
+
+			domainType := domainTypes[domain]
+
+			inter := parseCheckAndInterpret(t,
+				fmt.Sprintf(
+					`
+                      let x = %[1]s("fo-o")
```
turbolent, line 125:
```suggestion
                      let x = %[1]s(identifier: "fo-o")
```

runtime/tests/checker/path_test.go
```diff
@@ -105,3 +105,44 @@ func TestCheckPath(t *testing.T) {
...
+
+			domainType := domainTypes[domain]
+
+			checker, err := ParseAndCheck(t,
+				fmt.Sprintf(
+					`
+                      let x: %[1]s? = %[1]s("/%[2]s/foo")
+                    `,
+					domainType.String(),
+					domain.Identifier(),
```
turbolent, line 132: Shouldn't the call require an argument label `identifier:`? Maybe add a test that checks the argument label is required.

Also, shouldn't the call only accept an identifier and not a full path? (Is not checked statically, but still)

Maybe also remove the type annotation for the variable declaration: As it is explicitly given, the assertion below always succeeds. Instead, you want to assert that the inferred type for the variable matches the return type of the converter function:

```suggestion
                      let x = %s(identifier: "foo")
                    `,
					domainType.String(),
```

turbolent (commented):
👍

runtime/sema/errors.go
```diff
@@ -2700,7 +2700,16 @@ type InvalidPathDomainError struct {
 }
 
 func (e *InvalidPathDomainError) Error() string {
-	return "invalid path domain"
+	return fmt.Sprintf("invalid path domain %s", e.ActualDomain)
```
turbolent, line 2703: 👍

runtime/literal.go
```diff
@@ -101,13 +101,18 @@ func arrayLiteralValue(elements []ast.Expression, elementType sema.Type) (cadenc
...
 	if !ok {
 		return nil, LiteralExpressionTypeError
 	}
 
-	pathType, err := sema.CheckPathLiteral(pathExpression)
+	pathType, err := sema.CheckPathLiteral(
+		pathExpression.Domain.Identifier,
+		pathExpression.Identifier.Identifier,
+		ast.NewRangeFromPositioned(pathExpression.Domain),
+		ast.NewRangeFromPositioned(pathExpression.Identifier),
```
turbolent, line 114: The ranges are only needed for errors, which is the unlikely case. 

Maybe refactor the range parameters to functions of type `func () ast.Range`

runtime/sema/check_path_expression.go
```diff
@@ -19,29 +19,44 @@
...
 
-	domain, ok := common.AllPathDomainsByIdentifier[domainIdentifier.Identifier]
+	// Check that the domain is valid
+	domain, ok := common.AllPathDomainsByIdentifier[domainString]
 	if !ok {
 		return PathType, &InvalidPathDomainError{
-			ActualDomain: domainIdentifier.Identifier,
-			Range:        ast.NewRangeFromPositioned(domainIdentifier),
+			ActualDomain: domainString,
+			Range:        ast.NewRangeFromPositioned(domainRange),
```
turbolent, line 51: The domain range is already a range, no need to construct a new range from the existing range:

```suggestion
			Range:        domainRange,
```

runtime/sema/check_path_expression.go
```diff
@@ -19,29 +19,44 @@
...
+		}
+	}
+
+	// Check that the identifier is valid
+	if !isValidIdentifier(identifier) {
+		return PathType, &InvalidPathIdentifierError{
+			ActualIdentifier: identifier,
+			Range:            ast.NewRangeFromPositioned(idRange),
```
turbolent, line 59: The identifier range is already a range, no need to construct a new range from the existing range:

```suggestion
			Range:            idRange,
```

runtime/sema/type.go
```diff
@@ -3259,6 +3259,51 @@ func init() {
...
 		),
 	)
+
+	BaseValueActivation.Set(
+		PublicPathType.String(),
+		baseFunctionVariable(
+			PublicPathType.String(),
+			&FunctionType{
+				Parameters: []*Parameter{{
+					Identifier:     "identifier",
```
turbolent, line 3269: Not specifying an argument label makes the checker assume the identifier to be the argument label. 
Something seems to be wrong here, because the checker test below does not pass the argument label:
https://github.com/onflow/cadence/pull/1211/files#diff-1ddde7bafb81fcdd1cd19e775f812508485a4291b71f533f4f975fe4f9504f95R129

Also, the documentation states there is no argument label required. I'd suggest requiring the argument label `identifier`, to avoid confusion by users which might assume calls like `PublicPath("/public/foo")` work

docs/language/accounts.mdx
```diff
@@ -326,6 +326,21 @@ Both `StoragePath` and `CapabilityPath` are subtypes of `Path`.
   storagePath.toString()  // is "/storage/path"
   ```
 
+There are also utilities to produce paths from strings:
+
+```cadence
+fun PublicPath(_ identifier: string): PublicPath?
+fun PrivatePath(_ identifier: string): PrivatePath?
+fun StoragePath(_ identifier: string): StoragePath?
```
turbolent, line 334:
```suggestion
fun PublicPath(identifier: string): PublicPath?
fun PrivatePath(identifier: string): PrivatePath?
fun StoragePath(identifier: string): StoragePath?
```

runtime/interpreter/value.go
```diff
@@ -10765,6 +10766,44 @@ func (PathValue) IsStorable() bool {
...
 }
 
+func convertPath(domain common.PathDomain, value Value) Value {
+	stringValue, ok := value.(*StringValue)
+	if !ok {
+		return NilValue{}
+	}
+
+	_, err := sema.CheckPathLiteral(
+		"public",
```
turbolent, line 10776:
```suggestion
		domain,
```

runtime/interpreter/value.go
```diff
@@ -10765,6 +10766,44 @@ func (PathValue) IsStorable() bool {
...
+		stringValue.Str,
+		func() ast.Range {
+			return ReturnEmptyLocationRange().Range
+		},
+		func() ast.Range {
+			return ReturnEmptyLocationRange().Range
+		},
```
turbolent, line 10783: No need to construct a full location range (location + range), just to get an empty range:

```suggestion
		func() ast.Range {
			return ast.Range{}
		},
		func() ast.Range {
			return ast.Range{}
		},
```

And then further, maybe create a function like `ReturnEmptyLocationRange`:

```go
func ReturnEmptyRange() ast.Range {
	return ast.Range{}
}
```

```suggestion
		ReturnEmptyRange,
		ReturnEmptyRange,
```

runtime/sema/type.go
```diff
@@ -2990,6 +2990,7 @@ func baseFunctionVariable(name string, ty *FunctionType, docString string) *Vari
 	return &Variable{
 		Identifier:      name,
 		DeclarationKind: common.DeclarationKindFunction,
+		ArgumentLabels:  ty.ArgumentLabels(),
```
turbolent, line 2993: 👍

turbolent (approved):
Great work! 👏

turbolent:
@dsainati1 There has been many requests from the community for this feature. Maybe we can backport it to v0.23 to get it out earlier?

PR #1212, opened 2021-11-01
doc(06-marketplace-compose.mdx): update ownerVault access control keyword

GuillaumeBAECHLER:
## Description

Hi there :wave:

This is the nitpicky one again.
I changed the access control keyword for the `ownerVault` from `access(account)` to `pub` in the `Marketplace.cdc` template.

Later in the documentation (L292), there is a reference to this same line but with the access keyword `pub`.

I chose to keep the `pub` keyword because I feel that the capability-based access control here is sufficient to safely expose the vault in the code (right?). On the other hand, I don't see the point of making the vault accessible at this level of access, we can just check the vault directly if needed (so `access(account)` is enought for the `purchase` function or further possible contracts to work with the vault ref).

What is the best practice? (sorry if it's not that important, but I think access control in smart contracts is something very important and I want to make sure I get everything correctly :sweat_smile: )

If we should to keep `access(account)`, I'll make the change on line 292.
______

### DIFF
--- a/docs/tutorial/06-marketplace-compose.mdx
+++ b/docs/tutorial/06-marketplace-compose.mdx
@@ -200,7 +200,7 @@ pub contract Marketplace {
     // The fungible token vault of the owner of this sale.
     // When someone buys a token, this resource can deposit
     // tokens into their account.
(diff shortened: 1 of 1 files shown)


joshuahannan (approved):
Thank you!

PR #1214, opened 2021-11-03
Zero-pad string representation of addresses

### LINKED ISSUE #1131: Zero-padding of Address#toString()
### Instructions

When you execute Address#toString() in Cadence, it does not return a zero-padded result.
If the length of Address is fixed, I think it should be zero-padded result.

### Problem

When I used Cadence Script to get metadata containing address information as `{String:String}` type value and compared it with a string, I could not check that it was the same address.

### Steps to Reproduce

CADENCE: V0.18.0

```
  let addr: Address = 0x004b19d1e7c70d25
  log(addr.toString())
  // -> "0x4b19d1e7c70d25"
```

turbolent:
Closes #1131 

______

### DIFF
--- a/runtime/common/address.go
+++ b/runtime/common/address.go
@@ -87,6 +87,10 @@ func (a Address) ShortHexWithPrefix() string {
 	return fmt.Sprintf("0x%s", strings.TrimLeft(hexString, "0"))
 }
 
+func (a Address) HexWithPrefix() string {
+	return fmt.Sprintf("0x%x", [AddressLength]byte(a))
+}
+
 // HexToAddress converts a hex string to an Address.
 func HexToAddress(h string) (Address, error) {
 	trimmed := strings.TrimPrefix(h, "0x")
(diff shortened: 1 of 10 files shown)


jordanschalm (approved):
🎸

runtime/sema/check_composite_declaration.go
```diff
@@ -1229,8 +1229,21 @@ func (checker *Checker) checkTypeRequirement(
 	for _, nestedCompositeDeclaration := range containerDeclaration.Members.Composites() {
 		nestedCompositeIdentifier := nestedCompositeDeclaration.Identifier.Identifier
 		if nestedCompositeIdentifier == declaredCompositeType.Identifier {
+			// If we detected a second nested composite declaration with the same identifier,
```
SupunS, line 1232: I think this is not related to `address` fix. Is it intentionally included?

turbolent: Argh, good flag! I accidentally committed this in the wrong PR, removing

SupunS (approved):
Looks good! Thank you!

PR #1216, opened 2021-11-03
Add new syntax to bind for-in loop indices 

### LINKED ISSUE #402: for-in loop should bind index
My issue was posted in the cadence discord right here: 
https://discordapp.com/channels/613813861610684416/621847359894061056/738918199940350036

### Issue To Be Solved

Is there an easy way to get the index of an item in an array? 
In Go, for example, you can do something like

```go
for i, j := range array {
    // ... 
}
```

in cadence I found 

```swift
let array = [45, 65, 32]
for element in array {
    if element == 32 {
        //return index (2)? 
    }
}
```

**Now the for-in loop only binds the value and not the index**

### Suggest A Solution

Would it be possible to handle the index in a for-loop in the same way this happens, by example, in Go? 

```go
for **index**, element := range array {
... 
}
```

### 

I feel like this is a more intuitive way of getting the index so we don't have to make an index variable.

dsainati1:
Closes #402

## Description

This adds a new piece of syntax to for loops to the language to allow users to bind the index into the array being iterated over.

```
for x, y in array {
}
```

This introduces two variables `x` and `y` into the scope of the loop body; `y` is the element variable as in a normal for-in loop, while `x` is of type `int`, is initialized to 0 and is incremented by one at the end of each iteration of the loop, effectively tracking the index of the array being looped over. 

______

### DIFF
--- a/runtime/interpreter/interpreter_statement.go
+++ b/runtime/interpreter/interpreter_statement.go
@@ -288,11 +288,13 @@ func (interpreter *Interpreter) VisitForStatement(statement *ast.ForStatement) a
 	}
 
 	var indexVariable *Variable
-	var one = NewIntValueFromInt64(1)
+	var index int64
+	var indexValue IntValue
 	if statement.Index != nil {
+		indexValue = NewIntValueFromInt64(0)
 		indexVariable = interpreter.declareVariable(
 			statement.Index.Identifier,
-			NewIntValueFromInt64(0),
+			indexValue,
 		)
 	}
 
@@ -329,7 +331,8 @@ func (interpreter *Interpreter) VisitForStatement(statement *ast.ForStatement) a
 		}
 
 		if indexVariable != nil {
-			indexVariable.SetValue(indexVariable.GetValue().(IntValue).Plus(one))
+			index++
+			indexValue.BigInt.SetInt64(index)
 		}
 	}
 }
--- a/runtime/interpreter/interpreter_statement.go
+++ b/runtime/interpreter/interpreter_statement.go
@@ -287,6 +287,15 @@ func (interpreter *Interpreter) VisitForStatement(statement *ast.ForStatement) a
 		panic(ExternalError{err})
 	}
 
+	var indexVariable *Variable
+	var one = NewIntValueFromInt64(1)
+	if statement.Index != nil {
+		indexVariable = interpreter.declareVariable(
+			statement.Index.Identifier,
+			NewIntValueFromInt64(0),
+		)
+	}
+
 	for {
 		var atreeValue atree.Value
 		atreeValue, err = iterator.Next()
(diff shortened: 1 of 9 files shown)


runtime/interpreter/interpreter_statement.go
```diff
@@ -287,6 +287,15 @@ func (interpreter *Interpreter) VisitForStatement(statement *ast.ForStatement) a
...
 	}
 
+	var indexVariable *Variable
+	var one = NewIntValueFromInt64(1)
+	if statement.Index != nil {
+		indexVariable = interpreter.declareVariable(
+			statement.Index.Identifier,
+			NewIntValueFromInt64(0),
+		)
+	}
```
turbolent, line 297: Maybe optimize and avoid the allocation when increasing by keeping a plain indexing Go variable and updating the Cadence variable's value in-place:

```suggestion
	var indexVariable *Variable
	var index int64
	var indexValue IntValue
	if statement.Index != nil {
		indexValue = NewIntValueFromInt64(index)
		indexVariable = interpreter.declareVariable(
			statement.Index.Identifier,
			indexValue,
		)
	}
```

runtime/interpreter/interpreter_statement.go
```diff
@@ -318,6 +327,10 @@ func (interpreter *Interpreter) VisitForStatement(statement *ast.ForStatement) a
 		case functionReturn:
 			return result
 		}
+
+		if indexVariable != nil {
+			indexVariable.SetValue(indexVariable.GetValue().(IntValue).Plus(one))
+		}
```
turbolent, line 333: Second part of suggestion above:
```suggestion
		if indexVariable != nil {
			index++
			indexValue.BigInt.SetInt64(index)
		}
```

turbolent (commented):
Nice work! 👍 

BTW: We usually use a branch naming scheme like `<name>/<issue-nunber>-<short-name>` to avoid conflicts, e.g. for this PR/branch `daniel/402-for-loop-index`

runtime/interpreter/interpreter_statement.go
```diff
@@ -318,6 +329,11 @@ func (interpreter *Interpreter) VisitForStatement(statement *ast.ForStatement) a
 		case functionReturn:
 			return result
 		}
+
+		if indexVariable != nil {
+			index++
+			indexValue.BigInt.SetInt64(index)
```
SupunS, line 335: Is it safe to change the existing value, or should we assign a new value? I feel it might also mutate the usages of previous indexes as well.

like for eg: what if a user stores the indexes in an array, and prints them later?

turbolent: Ah, that's a good point. I had suggested this above to avoid allocation, but I think you are right

dsainati1: I added a test for this, and there are is indeed an error in this case. Nice catch

turbolent: @dsainati1 Nice, good idea to add a test! Feel free to revert to your original code, that worked and I think was safe 👍

turbolent (approved):
Nice work! 👍

PR #1229, opened 2021-11-09
Add a clone function to all values

turbolent:
## Description

Add a function that allows cloning a value. This enables implementing a storage migration which reads in existing values and re-stores them as new values.
______


runtime/interpreter/value.go
```diff
@@ -815,6 +828,10 @@ func (v *StringValue) Transfer(
 	return v
 }
 
+func (v *StringValue) Clone(_ *Interpreter) Value {
+	return NewStringValue(v.Str)
```
SupunS, line 832: Not sure, do we need to copy the bytes of `v.Str` here?

SupunS (approved):

runtime/interpreter/value.go
turbolent: No need I think, because strings are immutable

PR #1232, opened 2021-11-10
Improve state decoding tool

turbolent:
## Description

- Handle slab not found errors gracefully (report them)
- Ignore and report empty lines
______

### DIFF
--- a/runtime/cmd/decode-state-values/main.go
+++ b/runtime/cmd/decode-state-values/main.go
@@ -26,6 +26,7 @@ import (
 	"compress/gzip"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"flag"
 	"io"
 	"log"
@@ -210,7 +211,7 @@ type interpreterStorage struct {
 
 var _ interpreter.Storage = &interpreterStorage{}
 
-func (i interpreterStorage) ValueExists(_ *interpreter.Interpreter, _ common.Address, key string) bool {
+func (i interpreterStorage) ValueExists(_ *interpreter.Interpreter, _ common.Address, _ string) bool {
 	panic("unexpected ValueExists call")
 }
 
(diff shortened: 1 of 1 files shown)


runtime/cmd/decode-state-values/main.go
```diff
@@ -258,111 +259,137 @@ func load() {
...
 		// Check the key is a non-root slab or a storage path
 		key := storageKey[2]
 
 		var address atree.Address
 		copy(address[:], storageKey[0])
 
-		// If the key is for a slab (format '$' + storage index),
-		// then attempt to decode the slab
+		err := loadStorageKey(key, address, data, inter, slabStorage)
+		var slabNotFoundErr *atree.SlabNotFoundError
```
fxamacker, line 274: I'm not sure I understand when `loadStorageKey` would return `SlabNotFoundError`.

runtime/cmd/decode-state-values/main.go
```diff
@@ -258,111 +259,137 @@ func load() {
...
+		if errors.As(err, &slabNotFoundErr) {
+			slabNotFoundErrCount++
+		}
```
fxamacker, line 277: I think `atree.SlabNotFoundError` might not be returned from `loadStorageKey`.

turbolent: `loadStorageKey` calls `StoredValue` on the decoded slab, which in turn might more slabs. The decoded state might be corrupt, which might cause atree to load a slab which does not exist, resulting in a `SlabNotFoundError`

runtime/cmd/decode-state-values/main.go
```diff
@@ -258,111 +259,137 @@ func load() {
...
-									address, key,
-								)
-							}
-						}
+			reader := bytes.NewReader(data)
+			decoder := interpreter.CBORDecMode.NewStreamDecoder(reader)
+			storable, err := interpreter.DecodeStorable(decoder, atree.StorageIDUndefined)
+			if err != nil {
+				log.Printf(
+					"Failed to decode storable @ 0x%x %s: %s (data: %x)",
```
SupunS, line 342: not sure if `Printf` function already does this, but do we need to add a newline at the end of the log, here and other places?
```suggestion
					"Failed to decode storable @ 0x%x %s: %s (data: %x)\n",
```

SupunS (approved):

PR #1240, opened 2021-11-16
Allow scripts to access authorized accounts

### LINKED ISSUE #539: Scripts should have access to authorized accounts
### Issue To Be Solved

It is currently not possible to query account storage of accounts, as scripts do not have access to authorized accounts.

### Suggested Solution

- Add a global function `fun getAuthAccount(_ address: Address): AuthAccount`,
  which returns an authorized account for a given address, 
  to scripts, but *not* transactions!

dsainati1:
Closes #539 

## Description

Because scripts discard their changes upon completion, there is no harm in allowing them to access `AuthAccount`s. This adds a function:

```
getAuthAccount(_ address: Address): AuthAccount
```

That loads an `AuthAccount` in the same way that `getAccount` loads a `PublicAccount`. This is only available in scripts; trying to use this function outside of a script will cause an error in the checker. 

______


docs/language/accounts.mdx
```diff
@@ -156,6 +156,17 @@ to the `prepare` phase of the transaction.
...
+  using the built-in `getAuthAccount` function:
+
+  ```cadence
+  fun getAuthAccount(_ address: Address): AuthAccount
+  ```
+
+  This function is only available in scripts; as they discard 
+  their changes upon completion. Attempting to use this function
+  outside of a script will cause a type error. 
+
```
joshuahannan, line 169: Can you add some clarification about what you can do with this `AuthAccount` object in the script? You can still perform all of the regular operations with it, right?

joshuahannan (commented):
Exciting! 🎉  🥳

runtime/cmd/parse/main.go
```diff
@@ -1,3 +1,4 @@
+//go:build !wasm
```
turbolent, line 1: Was this accidentally added?

runtime/runtime.go
```diff
@@ -1475,15 +1475,37 @@ func (r *interpreterRuntime) standardLibraryFunctions(
...
+			stdlib.NewStandardLibraryFunction(
+				"getAuthAccount",
+				&sema.FunctionType{
+					Parameters: []*sema.Parameter{{
+						Label:          sema.ArgumentLabelNotRequired,
+						Identifier:     "address",
+						TypeAnnotation: sema.NewTypeAnnotation(&sema.AddressType{}),
+					}},
+					ReturnTypeAnnotation: sema.NewTypeAnnotation(sema.AuthAccountType),
+				},
```
turbolent, line 1500: Maybe factor this our into a global, so we don't have to allocate it again on each call

turbolent (approved):
Great work!

bluesign:
Why not extend `PublicAccount` with functions from `AuthAccount` in script context ?

runtime/cmd/parse/main.go
dsainati1: Ah yea, for some reason `make fix-lint` adds this every time I run it, and I must have forgotten to drop the changes before committing this time.

dsainati1:
> Why not extend `PublicAccount` with functions from `AuthAccount` in script context ?

Having the same type mean different things in different contexts is confusing; we don't want to change what a `PublicAccount` means depending on where a script is being used. Instead, it's a much simpler change to allow people to access `AuthAccount`s depending on whether they are in a script or a transaction, and this way each of the types mean the same thing everywhere. The only thing changing is whether they are accessible.

bluesign:
I know it is already merged, but here is my 2 cents. 

This is the diff of AuthAccount from PublicAccount:

```
struct AuthAccount {
    fun addPublicKey(_ publicKey: [UInt8])
    fun removePublicKey(_ index: Int)
    fun save<T>(_ value: T, to: StoragePath)
    fun load<T>(from: StoragePath): T?
    fun copy<T: AnyStruct>(from: StoragePath): T?
    fun borrow<T: &Any>(from: StoragePath): T?
    fun getCapability<T>(_ path: CapabilityPath): Capability<T>
    fun link<T: &Any>(_ newCapabilityPath: CapabilityPath, target: Path): Capability<T>?
    fun unlink(_ path: CapabilityPath)
    struct Contracts {
        fun add(
            name: String,
            code: [UInt8],
            ... contractInitializerArguments
        ): DeployedContract
        fun update__experimental(name: String, code: [UInt8]): DeployedContract
        fun remove(name: String): DeployedContract?
    }
    struct Keys {
        fun add(
            publicKey: PublicKey,
            hashAlgorithm: HashAlgorithm,
            weight: UFix64
        ): AccountKey
        fun revoke(keyIndex: Int): AccountKey?
    }
}
```

Below is the function that makes sense in script ( read only ): (considering getCapability is useless)

```
struct AuthAccount {
    fun load<T>(from: StoragePath): T?
    fun copy<T: AnyStruct>(from: StoragePath): T?
    fun borrow<T: &Any>(from: StoragePath): T?
}
```

technically load and copy are same (in script context) borrow is useless when you have load. So we end up with:

```
fun load<T>(from: StoragePath): T?
```

In my opinion exposing 14 functions where in reality 1 is needed outweights the benefit. Also `PublicAccount` has feeling of read-only vs `AuthAccount` has feeling of read/write. 


instead of `getAuthAccount` (which is exposed to scripts only), maybe it would be better to have something like:

`fun loadFromAccount<T>(Address, path: StoragePath): T?` 

But I think it would also make sense to just put `load` to `PublicAccount`. And just panic with a nice error message when in Transaction context.

turbolent:
@bluesign The difference between a transaction context and a script context is that in a transaction context state changes are written, and in the script context the changes are ignored. 

It is much simpler for the implementation to behave exactly the same in both contexts and I would assume that it is also the same for developers/users. I don't really think that changing the definitions and APIs (e.g. adding new / removing existing functions) is worth it. 

Having the existing definitions behave the same in a scripting context as in a transaction context is also useful for future features like testing, where you might want to perform state-changing actions, but are not interested in the state changes being applied. If the API were reduced to just a "read-only" API it would be limiting for some actions you might want to perform.

bluesign:
> Having the existing definitions behave the same in a scripting context as in a transaction context is also useful for future features like testing, where you might want to perform state-changing actions, but are not interested in the state changes being applied. If the API were reduced to just a "read-only" API it would be limiting for some actions you might want to perform.

Oh thats exactly the point that I missed, I now see how this can be a powerful feature. Thanks for the clarification.

bluesign:
Any eta on this to be included in Cadence ?

dsainati1:
This will be out in Secure Cadence I believe

Issue #1217, opened 2021-11-04
Check possibility of using static-type for runtime subtype checking


SupunS:
### Issue To Be Solved

Check the possibility of using static-subtype check in place of dynamic-subtype check at runtime for better performance. The speed gains are likely to come from type-checking of collections.

SupunS:
So I checked the possibility of doing this, and the perf gains. Below is a summary:

**Assumptions:**
- The static type for any value is known at the time of the value creation, and stays unchanged throughout the runtime (type safety). 
- Since DynamicType is always a subset of StaticType, if the StaticType satisfies some condition, then the dynamic type would also satisfy the same condition.
- Thus, checking the static type to make decisions on the value should be sufficient.

**Gains:** 
The benchmark results are:

```
name                            old time/op    new time/op    delta
CheckValueTransferTargetType-8    26.9µs ± 1%     0.1µs ± 3%  -99.72%  (p=0.008 n=5+5)
TransferAndConvert1K-8             116µs ± 3%      77µs ± 5%  -33.46%  (p=0.008 n=5+5)
TransferAndConvert10K-8           1.19ms ± 2%    0.90ms ± 6%  -24.29%  (p=0.008 n=5+5)

name                            old alloc/op   new alloc/op   delta
CheckValueTransferTargetType-8    16.6kB ± 0%     0.0kB ± 0%  -99.71%  (p=0.008 n=5+5)
TransferAndConvert1K-8            66.0kB ± 0%    48.8kB ± 0%  -26.00%  (p=0.008 n=5+5)
TransferAndConvert10K-8            685kB ± 1%     517kB ± 0%  -24.45%  (p=0.008 n=5+5)

name                            old allocs/op  new allocs/op  delta
CheckValueTransferTargetType-8      5.00 ± 0%      1.00 ± 0%  -80.00%  (p=0.008 n=5+5)
TransferAndConvert1K-8               123 ± 0%       119 ± 0%   -3.25%  (p=0.008 n=5+5)
TransferAndConvert10K-8            1.11k ± 0%     1.11k ± 0%   -0.45%  (p=0.029 n=4+4)
```
Here the three benchmarks are:
- [CheckValueTransferTargetType](https://github.com/onflow/cadence/blob/ad6895c077a591788fe299d8308ec3f3b2685329/runtime/interpreter/interpreter.go#L1926) method for an array with 1,000 elements 
[...]

PR #1243, opened 2021-11-16
Add test for casting storage ref to an invalid type

SupunS:
## Description

There were no tests for the following scenario:
 - Saving struct value of type `T` in an account
 - Borrowing it as a super-type `auth AnyStruct`
 - Downcast it to a different type `auth R`

This was the only scenario that doesn't work if we do https://github.com/onflow/cadence/issues/1217 (_there's a solution for it, I'll update the original git-issue with more context/info_), but realized there were no test cases for this particular case (all tests were passing!)

______


turbolent (approved):
Thank you for adding a test for this 👍

PR #1248, opened 2021-11-16
Refactor account storage to use atree ordered maps

### LINKED ISSUE #1222: Refactor account storage to atree OMT
### Issue To Be Solved

Currently account storage is mapped 1:1 to registers. Iteration over an account's stored values is only possible by adding an index. 

### Suggested Solution

Refactor account storage to use an atree OMT. This enables storage iteration.

turbolent:
Closes #1222 

## Description

Refactor account storage to use atree ordered maps: Instead of storing each path-value pair under a separate storage key, store the values in atree ordered maps, separated by domain.

Deploying this will require a storage migration.

______


runtime/interface.go
```diff
@@ -131,6 +131,4 @@ type Metrics interface {
 	ProgramParsed(location common.Location, duration time.Duration)
 	ProgramChecked(location common.Location, duration time.Duration)
 	ProgramInterpreted(location common.Location, duration time.Duration)
-	ValueEncoded(duration time.Duration)
-	ValueDecoded(duration time.Duration)
```
turbolent, line 135: These functions are not used anymore: encoding and decoding is done inside of the atree library. We might want to bring these metric callbacks back at some point

runtime/interpreter/errors.go
```diff
@@ -395,21 +395,6 @@ func (e TypeLoadingError) Error() string {
...
-	Path  []string
-}
-
-func (e EncodingUnsupportedValueError) Error() string {
-	return fmt.Sprintf(
-		"encoding unsupported value to path [%s]: %[2]T, %[2]v",
-		strings.Join(e.Path, ","),
-		e.Value,
-	)
-}
```
turbolent, line 411: This is just a cleanup: Code producing this error was removed earlier in c3ac75224f3d585c95980bed71cb29e8fb3a0bea

runtime/interpreter/interpreter.go
```diff
@@ -3369,15 +3382,17 @@ func (interpreter *Interpreter) authAccountCopyFunction(addressValue AddressValu
 
 func (interpreter *Interpreter) authAccountReadFunction(addressValue AddressValue, clear bool) *HostFunctionValue {
 
+	address := addressValue.ToAddress()
```
turbolent, line 3385: Converted addresses can be cached and don't have to be recomputed on each function invocation

runtime/interpreter/interpreter.go
```diff
@@ -4048,6 +4077,9 @@ func (interpreter *Interpreter) ValidateAtreeValue(v atree.Value) {
 			return info.Equal(other.(StaticType))
 		case compositeTypeInfo:
 			return info.Equal(other)
+		case EmptyTypeInfo:
```
turbolent, line 4080: Storage maps are atree ordered maps without any type info. They map storage paths to arbitrarily typed values, so static type information is useless

runtime/interpreter/storage.go
```diff
@@ -85,27 +83,55 @@ func ConvertStoredValue(value atree.Value) (Value, error) {
 	}
 }
 
-type StorageKey struct {
-	Address common.Address
-	Key     string
-}
+func DecodeTypeInfo(dec *cbor.StreamDecoder) (atree.TypeInfo, error) {
```
turbolent, line 86: Moved up and added support for decoding CBOR nil == empty type info

runtime/interpreter/storagemap.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+}
+
+func (s StorageMap) StorageID() atree.StorageID {
+	return s.orderedMap.StorageID()
+}
+
+// StorageMapIterator is an iterator over StorageMap
+//
+type StorageMapIterator struct {
```
turbolent, line 189: This iterator will be used in the future for exposing an storage iteration API to Cadence programs

runtime/runtime.go
```diff
@@ -2369,8 +2362,7 @@ func (r *interpreterRuntime) newAuthAccountContractsChangeFunction(
 				context.SetCode(context.Location, string(code))
 
 				panic(fmt.Errorf(
-					"invalid %s: the name argument must match the name of the declaration"+
-						"name argument: %q, declaration name: %q",
+					"invalid %s: the name argument must match the name of the declaration: got %q, expected %q",
```
turbolent, line 2365: The error message was missing a dot between the two sentences (see `+` operator). I fixed this and also made it consistent with the other error messages (got/expected)

runtime/runtime_test.go
```diff
@@ -5475,7 +5446,7 @@ func TestRuntimeContractWriteback(t *testing.T) {
 		},
 		storage: newTestLedger(nil, onWrite),
 		getSigningAccounts: func() ([]Address, error) {
-			return []Address{common.BytesToAddress(addressValue.Bytes())}, nil
+			return []Address{Address(addressValue)}, nil
```
turbolent, line 5449: Cleanup: No need to convert from array to slice and back to array, can convert between arrays directly

runtime/storage.go
```diff
@@ -283,117 +232,117 @@ func (s *Storage) recordContractUpdate(
...
 
 		var err error
 		wrapPanic(func() {
 			err = s.Ledger.SetValue(
-				address[:],
-				[]byte(storageKey.Key),
-				encoded,
+				write.storageKey.Address[:],
+				[]byte(write.storageKey.Key),
+				write.data,
```
fxamacker, line 345: Not sure, but if `write.data` is a storage index referring to an empty map, should we remove the entry instead of setting a value?

fxamacker (commented):
Nice work!  Just a few comments/questions.

runtime/storage.go
```diff
@@ -283,117 +232,117 @@ func (s *Storage) recordContractUpdate(
...
+		a := updates[i].Key
+		b := updates[j].Key
+		return a.IsLess(b)
+	})
+}
+
+func (s *Storage) commitContractUpdates(inter *interpreter.Interpreter) {
 
-	var accountStorageEntries []AccountStorageEntry
+	var contractUpdates []ContractUpdate
```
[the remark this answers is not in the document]
turbolent: Great idea, refactored and optimized in 9e0890b5ab5afb003c3a7b2ecce42f6f4be260ee 👍

runtime/storage.go
```diff
@@ -33,23 +33,39 @@ import (
...
+}
+
 type Storage struct {
 	*atree.PersistentSlabStorage
-	// NOTE: temporary, will be refactored to dictionary
-	writes          map[interpreter.StorageKey]atree.Storable
-	readCache       map[interpreter.StorageKey]atree.Storable
-	contractUpdates map[interpreter.StorageKey]atree.Storable
+	writes          map[StorageKey][]byte
+	readCache       map[StorageKey][]byte
```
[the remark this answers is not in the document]
turbolent: Great idea! Refactored in 668434b05ccb5d72087872e818b6c69f84eff8bc 👍

runtime/storage.go
```diff
@@ -165,97 +122,89 @@ func (s *Storage) readStorable(storageKey interpreter.StorageKey) atree.Storable
...
+		Domain:  domain,
 	}
 
-	s.readCache[storageKey] = readStorable
+	storageMap = s.storageMaps[key]
+	if storageMap == nil {
 
-	return readStorable
-}
+		data := s.read(address, domain)
```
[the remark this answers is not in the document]
turbolent: Yep, same as https://github.com/onflow/cadence/pull/1248#discussion_r751879621, great idea! Refactored in 668434b

runtime/storage.go
```diff
@@ -33,23 +33,39 @@ import (
...
+	}
+}
+
 type Storage struct {
 	*atree.PersistentSlabStorage
-	// NOTE: temporary, will be refactored to dictionary
-	writes          map[interpreter.StorageKey]atree.Storable
-	readCache       map[interpreter.StorageKey]atree.Storable
-	contractUpdates map[interpreter.StorageKey]atree.Storable
+	writes          map[StorageKey][]byte
```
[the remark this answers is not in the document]
turbolent: Good idea to use the same key type, I didn't notice they had the same fields (through different "purposes"), refactored in 0d054738ec1d0a5c60bc10f7dd7f59f74e7c7dd5 👍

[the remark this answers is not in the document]
turbolent: > Also, what do you think about changing writes map value from `[]byte` to `atree.StorageIndex` to be more explicit?

Good idea! I tried this, and it compiles, but for whatever reason I couldn't quite figure out yet, some tests now fail and cannot load the contract value. It looks like the storage map for the `contract` domain is a different one (?), so maybe there is now some odd storage index mix-up / re-use?

I pushed the code to branch `bastian/1222-atree-account-storage-2`, the refactor commit is c830533f79e5c65abf77f40a9127e4cb4a936ae9.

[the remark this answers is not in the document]
fxamacker: The problem is `write.StorageIndex` inside range loop is being overwritten.  

In storage.go, `Storage.Commit()` function calls `Ledger.SetValue` in a range loop.

The loop needs to be changed from `for _, write := range writes` to `for i:=0; i<len(writes); i++`.

Here's an example in a standalone program to show the difference.
https://play.golang.com/p/ZxwtGXQnj0H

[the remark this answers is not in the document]
turbolent: Thank you for finding the bug! Fixed in 6e2fd29da66cf0a939648cbb6518edb5c911c672

runtime/storage.go
```diff
@@ -283,117 +232,117 @@ func (s *Storage) recordContractUpdate(
...
-					err = storable.Encode(encoder)
-				},
-				func(metrics Metrics, duration time.Duration) {
-					metrics.ValueEncoded(duration)
-				},
-			)
-			if err != nil {
-				return err
-			}
+	var writes []write
```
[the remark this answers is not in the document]
turbolent: Good idea! Refactored in 5df17787f6b2139dcd1a192524a6d4dc248ad18c

turbolent:
@fxamacker Removed the unnecessary check for `writes` when loading an existing/creating a new storage map in b8efa1091a03bc0ea7550be89533632f6e122911 like you suggested 👍

fxamacker:
@turbolent Do you want to merge `bastian/1222-atree-account-storage-2` back to `bastian/1222-atree-account-storage`?

turbolent:
@fxamacker yup, forgot that

runtime/storage.go
```diff
@@ -283,117 +177,143 @@ func (s *Storage) recordContractUpdate(
 			panic(err)
 		}
 
-		s.contractUpdates[storageKey] = storable
+		s.contractUpdates[key] = storable
```
fxamacker, line 180: Not sure, but is there a leak here?

Contract storable is created and stored in `contractUpdates` map in `recordContractUpdate()`.  In `writeContractUpdate()` contract storable is converted to Value object and is inserted to underlying ordered map.  Ordered map creates another storable for this contract during insertion.

runtime/storage.go
```diff
@@ -283,117 +177,143 @@ func (s *Storage) recordContractUpdate(
...
 
 		var err error
 		wrapPanic(func() {
 			err = s.Ledger.SetValue(
-				address[:],
-				[]byte(storageKey.Key),
-				encoded,
+				write.storageKey.Address[:],
+				[]byte(write.storageKey.Key),
+				write.storageIndex[:],
```
fxamacker, line 316: What do you think about removing `writes` cache entry after `Ledger.SetValue` is returned?  I think get storage used function can call `Commit` multiple times and cause the same value to be written to ledger more than once.

fxamacker (commented):
Great work!  :+1:  Just a few comments/questions.

runtime/storage.go
```diff
@@ -61,201 +55,101 @@ func NewStorage(
...
 		Address: address,
-		Key:     key,
+		Key:     name,
 	}
 
 	// Remove existing, if any
 
-	existingStorable, ok := s.contractUpdates[storageKey]
+	existingStorable, ok := s.contractUpdates[key]
 	if ok {
```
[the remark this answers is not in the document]
turbolent: Yes, great catch! 

`contractUpdates` explicitly keeps `nil` entries to record contract removals, so in the case of a transaction removing a contract and adding a contract back in the same transaction, this will fail.

I've fixed this and added a test case in 7cce3ae769b5cf2b4eb1b62ec9f8d5506400be31

runtime/storage.go
turbolent: Yes, good catch! I guess there is no need to keep contract updates as storables, they can just be kept as values in `contractUpdates`, and then `writeContractUpdate` implicitly gets the storable during the insertion into the storage map? I've refactored this in d3a4ef2efc4e82c2825a9bb39960a48ce1277e48, what do you think?

runtime/storage.go
turbolent: I guess that makes sense, I added this in 61dee3f69f21b2fd695aef133a688a965393cc6c

runtime/storage.go
fxamacker: LGTM!  :+1:

fxamacker (approved):
Great work!

turbolent:
@fxamacker As discussed, refactored the APIs to pass `Value`/`nil` vs `SomeValue`/`NilValue` in 97c952d + a409e30

turbolent:
@SupunS Could you please also have a look? 🙏

fxamacker (approved):
LGTM! 👍

runtime/interpreter/interpreter.go
```diff
@@ -3627,33 +3635,29 @@ func (interpreter *Interpreter) authAccountLinkFunction(addressValue AddressValu
...
-
 			capabilityPath := invocation.Arguments[0].(PathValue)
 
-			capabilityKey := PathToStorageKey(capabilityPath)
-
-			value := interpreter.ReadStored(address, capabilityKey)
-
-			switch value := value.(type) {
-			case NilValue:
-				return value
```
SupunS, line 3643: Is it no longer possible to get a `NilValue`  from `interpreter.ReadStored()`?

runtime/interpreter/interpreter.go
```diff
@@ -3801,62 +3803,58 @@ func (interpreter *Interpreter) capabilityCheckFunction(
...
+	seenPaths := map[PathValue]struct{}{}
 	paths := []PathValue{path}
 
 	for {
 		// Detect cyclic links
 
-		if _, ok := seenKeys[key]; ok {
-			return "", false, CyclicLinkError{
+		if _, ok := seenPaths[path]; ok {
+			return PathValue{}, false, CyclicLinkError{
```
SupunS, line 3825: Is it possible/safe to use a predefined constant for `PathValue{}` (i.e: something like `EmptyPath`) here and in other places?

runtime/storage.go
```diff
@@ -61,344 +54,241 @@ func NewStorage(
...
-	storageKey := interpreter.StorageKey{
+func (s *Storage) GetStorageMap(address common.Address, domain string) (storageMap *interpreter.StorageMap) {
+	key := interpreter.StorageKey{
 		Address: address,
-		Key:     key,
+		Key:     domain,
 	}
 
-	// Check locally
+	storageMap = s.storageMaps[key]
```
SupunS, line 71: Should we gracefully handle the non-existing keys?
```go
storageMap, ok := s.storageMaps[key]
```

SupunS (approved):
I don't have much knowledge on the atree-specific changes. But the cadence changes and the general Go codes look fine!

runtime/interpreter/interpreter.go
turbolent: I refactored the internal storage API in 97c952d81aa097191bdf0e0335d2dd74f344b99a: 
Previously, reads returned an optional, `NilValue` if the read returns nothing, and `SomeValue` if there is a returned value. No the function returns a Go `nil` if the read returns nothing, and the stored value itself if there is one, without wrapping it in a `SomeValue`. (Similarly I refactored `writeStored` from `OptionalValue` to Go `nil`/`Value`). This removes the rather unnecessary boxing / allocations, as suggested by @fxamacker 👍

runtime/interpreter/interpreter.go
turbolent: Good idea! Refactored the remaining cases and added comments everywhere in fa1b621875708ed5c95cbd028cc4a56d2438eb97

runtime/interpreter/interpreter.go
turbolent: Good idea, refactored in 436457968ec35e198d8a6e9c455ac4155ce2d7ae

runtime/storage.go
turbolent: The value type of `storageMaps` is `*interpreter.StorageMap`, so Go will use the default value for the type here when the lookup fails, which is `nil`, so this should already be safe (and is tested)

PR #1250, opened 2021-11-17
Improve API for BLS crypto functions

dsainati1:
Based on discussion in https://github.com/onflow/flow-go/pull/1605, it seems as though users would like these functions to return a default value in case of a crypto library error. 

______


runtime/cmd/parse/main.go
```diff
@@ -1,3 +1,4 @@
+//go:build !wasm
```
turbolent, line 1: Why was this added? Maybe remove it?
```suggestion
```

runtime/runtime.go
```diff
@@ -3277,18 +3287,21 @@ func aggregateBLSPublicKeys(
...
 	var key *PublicKey
 	wrapPanic(func() {
-		key, err = runtimeInterface.AggregateBLSPublicKeys(publicKeys)
+		key, err = runtimeInterface.BLSAggregatePublicKeys(publicKeys)
 	})
 
+	// if the crypto layer produces an error, we have invalid input, return nil
 	if err != nil {
-		return nil, err
+		return interpreter.NilValue{}, nil //nolint
```
turbolent, line 3295: Maybe only nolint the specific lint that is hit (error is ignored?), instead of all linting

runtime/runtime.go
```diff
@@ -3246,13 +3246,19 @@ func verifyBLSPOP(
...
+
 	var valid bool
 	wrapPanic(func() {
 		valid, err = runtimeInterface.BLSVerifyPOP(publicKey, signature)
 	})
 
+	// if the crypto layer produces an error, we have invalid input, return false
 	if err != nil {
-		return false, err
+		return interpreter.BoolValue(false), nil //nolint
```
turbolent, line 3261: Same here, maybe don't ignore all linting, just the specific lint

runtime/runtime.go
```diff
@@ -3246,13 +3246,19 @@ func verifyBLSPOP(
 		return false, err
 	}
 
+	// if public key is invalid, verify returns false
+	if !publicKey.IsValid {
+		return interpreter.BoolValue(false), nil
+	}
```
turbolent, line 3252: The implementation of `BLSVerifyPOP` should instead check `Validated`, and validate the public key if necessary. Then, if the key is invalid, return false. 

```suggestion
```

See a detailed explanation in https://axiomzen.slack.com/archives/CG0B7CJAJ/p1637366567211300?thread_ts=1637289659.195200&cid=CG0B7CJAJ

tarakby: I think the key is guaranteed to be validated already (instantiating a PublicKey in Cadence includes the validation), so it should be safe to just check `publicKey.IsValid`.

turbolent:
@dsainati1 After discussions with @tarakby and @dete in https://axiomzen.slack.com/archives/CG0B7CJAJ/p1637289659195200, I issued up #1268 and @robert-e-davidson3 is going to work on it. Once done, we can rebase this PR and it'll make thinks much simpler (no `IsValid` / `Validated` checking necessary in Cadence or FVM)

turbolent (approved):
Looks good! Nice

PR #1252, opened 2021-11-17
add `forceSave` function to save and overwrite stored data

### LINKED ISSUE #1244: Add a `forceSave` function to overwrite stored data
See #376 for context

Proposed solution

```
fun forceSave<T>(_ value: T, to path: Path): Bool
```

This function would save `value` to the specified `path`, as well as returning `true` if a value was previously at that path, and `false` otherwise.

dsainati1:
Closes #1244 

## Description

Adds the following method to `AuthAccount` objects:
```
fun forceSave<T>(_ value: T, to path: Path): Bool
```

This will save a value to the specified path, overwriting any data present (and destroying it if it is a resource). Returns whether or not any data was overwritten. 

______

### DIFF
--- a/docs/language/accounts.mdx
+++ b/docs/language/accounts.mdx
@@ -104,6 +104,7 @@ to the `prepare` phase of the transaction.
       // Account storage API (see the section below for documentation)
 
       fun save<T>(_ value: T, to: StoragePath)
+      fun forceSave<T>(_ value: T, to: StoragePath): Bool
       fun load<T>(from: StoragePath): T?
       fun copy<T: AnyStruct>(from: StoragePath): T?
 
@@ -358,6 +359,20 @@ to all its stored objects.
   If there is already an object stored under the given path, the program aborts.
 
   The path must be a storage path, i.e., only the domain `storage` is allowed.
+
+- `cadence•fun forceSave<T>(_ value: T, to: StoragePath): Bool`
+  
+  Saves the given object into the account's storage at the given path.
+  Resources are moved into storage, and structures are copied.
+
+  If there is already an object stored under the given path, the stored
+  object will be overwritten. If the stored object was a resource, it will
+  be destroyed.  
+
+  The path must be a storage path, i.e., only the domain ` + "`storage`" + ` is allowed.
+
+  This function returns true if a previously stored object was overwritten, and
+  false if the given path had no stored object. 
 
 - `cadence•fun load<T>(from: StoragePath): T?`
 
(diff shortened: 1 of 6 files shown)


joshuahannan:
Is it safe to add stuff like this and the `clear` method? It might encourage people to use bad practices for handling resources

dsainati1:
`forceSave` and `clear` are both safe with respect to resources because they will destroy any resources that are potentially saved at the destination. Neither method really adds any new functionality to the language; they only make certain patterns easier. 

```
account.clear(/storage/x)
``` 
is equivalent to 
```
let r <- account.load<@R>(/storage/x)
destroy r
```

while 
```
account.forceSave(<-r, /storage/x)
```
is equivalent to 
```
account.clear(/storage/x)
account.save(<-r, /storage/x)
```

joshuahannan:
I understand that they are just simpler patters for things that already exist, but I feel like this kind of blurs the line on the rule "resource cannot be lost, they have to be explicitly destroyed". To me, this is not explicit destruction and kind of breaks that rule, and I could see some people reverting to using these by default instead of the more explicit versions that we currently use, which could cause unintended loss of resources in some circumstances.

I would support adding these for non-resource types, but it makes me nervous to add them for resource types

bluesign:
I agree with @joshuahannan, when you give developer hammer, everything seems like nail to them, for example: when setting up FUSD Vault imagine developer accidentally override. 

I think storage system ( cadence part ) needs a rethinking, instead of adding features.  Those changes are not solving any serious problem, like storage sandboxing etc... 

I was expecting after storage improvements, there would be some work on this area. IMO at least we need:

- sandboxing ( dapp isolation ) 
- capability revoke 
- loading types which does not exist anymore 

 If it was up to me, I would even get rid of Paths totally.

turbolent:
@dete Any thoughts here?

turbolent:
@bluesign Fair enough on the feedback on this PR in particular, but this PR isn't the right place to vent your expectations about the roadmap. Please keep the conversation productive and focused on this PR in particular. Thanks

bluesign:
> @bluesign Fair enough on the feedback on this PR in particular, but this PR isn't the right place to vent your expectations about the roadmap. Please keep the conversation productive and focused on this PR in particular. Thanks

sorry about that, I went off topic a bit. :(

turbolent:
@dsainati1 Should we maybe consider this change for Stable Cadence?

dsainati1:
> @dsainati1 Should we maybe consider this change for Stable Cadence?

See comment on https://github.com/onflow/cadence/pull/1253

dsainati1:
Seems like this feature is not wanted; closing.

PR #1253, opened 2021-11-17
add a `clear` method to delete saved data in storage

### LINKED ISSUE #1245: Add a `clear` function to erase stored data
See #376 for context

Proposed Solution:

```
fun clear(_ path: Path): Bool
```

This would delete the data stored at `path`, returning `true` if there was data to be cleared and `false` otherwise.

dsainati1:
Closes #1245

Adds a new method to the `AuthAccount` object:

```
fun clear(_ path: Path): Bool
```

This function deletes whatever is present at the specified path, destroying any resources, and returns whether anything was present originally. 

______

### DIFF
--- a/docs/language/accounts.mdx
+++ b/docs/language/accounts.mdx
@@ -377,6 +377,20 @@ to all its stored objects.
 
   The path must be a storage path, i.e., only the domain `storage` is allowed.
 
+- `cadence•fun clear(_ path: StoragePath): Bool`
+
+  Clears an object from the account's storage which is stored under the given path
+
+  If there is an object stored, the stored resource or structure is removed from storage, and 
+  destroyed if it is a resource.
+
+  When the function returns, the storage no longer contains an object under the given path.
+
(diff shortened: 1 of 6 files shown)


turbolent:
@dete Any thoughts?

turbolent:
@dsainati1 Should we maybe consider this change for Stable Cadence?

dsainati1:
> @dsainati1 Should we maybe consider this change for Stable Cadence?

Seems to me like based on the discussion on https://github.com/onflow/cadence/pull/1252, the sentiment is against these two methods, the argument being that any operation that destroys a resource should be explicit; having methods that can implicitly destroy resources is not a desirable feature for the language. 

However I would be willing to put this to a FLIP if it's something we want to push for anyways.

turbolent:
@dsainati1 Sounds good! Added it to the agenda of the next LDM, maybe we can decide there together if we want to close this and the other related PR, or open a FLIP for them

dsainati1:
Seems like this feature is not wanted; closing.

PR #1254, opened 2021-11-18
add `typeAt` function for inspecting the type of objects in storage

### LINKED ISSUE #1246: add a `typeAt` function to get the type of stored data
See #376 for context

Proposed Solution:

```
fun typeAt(_ path: Path): Type?
```

This would look up the value stored at the given path, and return its type if present, and nil otherwise.

dsainati1:
Closes #1246

## Description

Adds a method to `AuthAccount`s to inspect the type of an object in storage:

```
cadence•fun typeAt(_ path: StoragePath): Type?
```

if the storage path is empty, the function returns nil. Otherwise, it returns the type of the value stored there.
______


runtime/sema/authaccount_type.go
```diff
@@ -277,6 +284,32 @@ var AuthAccountTypeLoadFunctionType = func() *FunctionType {
...
+				TypeAnnotation: NewTypeAnnotation(StoragePathType),
+			},
+		},
+		ReturnTypeAnnotation: NewTypeAnnotation(
+			&OptionalType{
+				Type: MetaType,
+			},
+		),
+	}
+}()
```
turbolent, line 311: No need for an immediately invoked closure here:

```suggestion
var AuthAccountTypeTypeAtFunctionType = &FunctionType{
	Parameters: []*Parameter{
		{
			Label:          ArgumentLabelNotRequired,
			Identifier:     "path",
			TypeAnnotation: NewTypeAnnotation(StoragePathType),
		},
	},
	ReturnTypeAnnotation: NewTypeAnnotation(
		&OptionalType{
			Type: MetaType,
		},
	),
}
```

runtime/interpreter/interpreter.go
```diff
@@ -3415,6 +3415,35 @@ func (interpreter *Interpreter) authAccountSaveFunction(addressValue AddressValu
...
+			value := interpreter.ReadStored(address, key)
+
+			switch value := value.(type) {
+			case NilValue:
+				return value
+
+			case *SomeValue:
+				return TypeValue{
+					Type: value.Value.StaticType(),
+				}
```
turbolent, line 3436: Result needs to be wrapped in a `SomeValue`. Maybe add a test/assertion

```suggestion
				return NewSomeValueNonCopying(
					TypeValue{
						Type: value.Value.StaticType(),
					},
				)
```

runtime/tests/interpreter/account_test.go
```diff
@@ -205,6 +206,91 @@ func TestInterpretAuthAccount_save(t *testing.T) {
...
+
+			  fun saveS() {
+				let s = S()
+				destroy account.load<@R>(from: /storage/x)
+			 	account.save(s, to: /storage/x)
+			  }
+
+              fun typeAt(): Type? {
+				return account.typeAt(/storage/x)
+              }
```
turbolent, line 241: The test case is a bit brittle here: The bug above where the result of the function wasn't wrapped in a `SomeValue` properly wasn't caught, because the non-optional value was automatically boxed due to the `Type?` return type. Maybe return `AnyStruct` instead and add a comment why

docs/language/accounts.mdx
```diff
@@ -359,6 +359,14 @@ to all its stored objects.
 
   The path must be a storage path, i.e., only the domain `storage` is allowed.
 
+- `cadence•fun typeAt(_ path: StoragePath): Type?`
```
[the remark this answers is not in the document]
dsainati1: I named it `typeAt` to avoid confusion with the `Type` function, which is used to create a runtime type. They are slightly different, as one has a capital letter and the other doesn't, but I still think it might be confusing to have both of these builtins have the same name.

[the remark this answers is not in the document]
turbolent: Hmm, I hadn't thought of that at all. Given that this is a function on `AuthAccount`, I don't really think it'll be confused with functionality for run-time types

docs/language/accounts.mdx
```diff
@@ -370,6 +370,14 @@ to all its stored objects.
 
   The path must be a storage path, i.e., only the domain `storage` is allowed.
 
+- `cadence•fun type(at path: StoragePath): Type?`
+
+  Reads the type of an object from the account's storage which is stored under the given path, or nil if no object is stored under the given path.
+
+  If there is an object stored, the type of the object is returned without modifying the stored object. 
+
+  The path must be a storage path, i.e., only the domain ` + "`storage`" + ` is allowed
```
turbolent, line 379:
```suggestion
  The path must be a storage path, i.e., only the domain `storage` is allowed
```

turbolent (approved):
Nice! Just one small issue in the documentation, but should be good to merge after

PR #1270, opened 2021-11-22
Include dependencies in contributing doc, plus gitignore.

robert-e-davidson3:
## Description
I added a section describing the requirements to build and test cadence. Aside from `go`, `wabt` is also needed.

I also added some paths to .gitignore since they were generated when I ran `make build` or `make test`.

Notes:
- I didn't create an issue. Do I need to?
- Not sure which label(s) to use.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,9 @@ suppressions
 
 languageserver/test/languageserver
 coverage.txt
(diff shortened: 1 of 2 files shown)


turbolent (approved):
Nice! First PR 🎉👏

PR #1273, opened 2021-11-24
Removed PublicKey.isValid and .Validated

### LINKED ISSUE #1268: Clean up PublicKey
### Issue To Be Solved

Currently it is possible to construct invalid public key. Representing invalid public keys is not useful and complicates the implementation of much of the functionality using public keys (e.g. signature verification).

### Suggested Solution

- In the `PublicKey` constructor: Keep validating the inputs, but instead of setting the `isValid` field, abort (panic) if the public key data is invalid
- Remove the `IsValid` and `Validated` fields from the `runtime.PublicKey` type
- Remove the `isValid` field from the Cadence `PublicKey` type, it is effectively constantly true

robert-e-davidson3:
Closes #1268 

## Description

Panics when a public key isn't valid. This makes `PublicKey.isValid` and `PublicKey.Validated` unnecessary, so they're removed.

Changes Cadence:
1. `PublicKey` no longer has an `isValid` field. 
2. `PublicKey` can no longer be invalid.

### DIFF
--- a/docs/language/crypto.mdx
+++ b/docs/language/crypto.mdx
@@ -96,7 +96,6 @@ pub enum SignatureAlgorithm: UInt8 {
 struct PublicKey {
     let publicKey: [UInt8]
     let signatureAlgorithm: SignatureAlgorithm
-    let isValid: Bool
 
     /// Verifies a signature under the given tag, data and public key.
     /// It uses the given hash algorithm to hash the tag and data.
@@ -146,8 +145,7 @@ publicKey.publicKey = []                                            // Not allow
 publicKey.publicKey[2] = 4      // No effect
 ```
 
-The validity of a public key can be checked using the `isValid` field.
-Verifications performed with an invalid public key (using `verify()` method) will always fail.
+Invalid public keys cannot be constructed so public keys are always valid.
 
 ### Signature verification
 
--- a/runtime/account_test.go
+++ b/runtime/account_test.go
@@ -22,6 +22,8 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/onflow/cadence/runtime/interpreter"
+
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
@@ -237,8 +239,6 @@ var accountKeyA = &AccountKey{
 	PublicKey: &PublicKey{
 		PublicKey: []byte{1, 2, 3},
 		SignAlgo:  sema.SignatureAlgorithmECDSA_P256,
-		IsValid:   false,
-		Validated: true,
 	},
 	HashAlgo:  sema.HashAlgorithmSHA3_256,
 	Weight:    100,
@@ -250,8 +250,6 @@ var accountKeyB = &AccountKey{
 	PublicKey: &PublicKey{
 		PublicKey: []byte{4, 5, 6},
 		SignAlgo:  sema.SignatureAlgorithmECDSA_secp256k1,
-		IsValid:   false,
-		Validated: false,
 	},
 	HashAlgo:  sema.HashAlgorithmSHA3_256,
 	Weight:    100,
@@ -275,7 +273,7 @@ func TestRuntimeAuthAccountKeys(t *testing.T) {
 		storage := newTestAccountKeyStorage()
 		rt := newTestInterpreterRuntime()
 		runtimeInterface := getAccountKeyTestRuntimeInterface(storage)
-
+		addPublicKeyValidation(runtimeInterface, nil)
 		addAuthAccountKey(t, rt, runtimeInterface)
 
 		assert.Equal(t, []*AccountKey{accountKeyA}, storage.keys)
@@ -289,6 +287,7 @@ func TestRuntimeAuthAccountKeys(t *testing.T) {
 		storage := newTestAccountKeyStorage()
 		rt := newTestInterpreterRuntime()
 		runtimeInterface := getAccountKeyTestRuntimeInterface(storage)
+		addPublicKeyValidation(runtimeInterface, nil)
 
 		addAuthAccountKey(t, rt, runtimeInterface)
 
@@ -312,7 +311,7 @@ func TestRuntimeAuthAccountKeys(t *testing.T) {
 		assert.Equal(
 			t,
 			[]string{
-				"AccountKey(keyIndex: 0, publicKey: PublicKey(publicKey: [1, 2, 3], signatureAlgorithm: SignatureAlgorithm(rawValue: 1), isValid: false), hashAlgorithm: HashAlgorithm(rawValue: 3), weight: 100.00000000, isRevoked: false)",
+				"AccountKey(keyIndex: 0, publicKey: PublicKey(publicKey: [1, 2, 3], signatureAlgorithm: SignatureAlgorithm(rawValue: 1)), hashAlgorithm: HashAlgorithm(rawValue: 3), weight: 100.00000000, isRevoked: false)",
 			},
 			storage.logs,
 		)
@@ -325,6 +324,7 @@ func TestRuntimeAuthAccountKeys(t *testing.T) {
 		storage := newTestAccountKeyStorage()
 		rt := newTestInterpreterRuntime()
 		runtimeInterface := getAccountKeyTestRuntimeInterface(storage)
+		addPublicKeyValidation(runtimeInterface, nil)
 
 		addAuthAccountKey(t, rt, runtimeInterface)
 
@@ -351,6 +351,7 @@ func TestRuntimeAuthAccountKeys(t *testing.T) {
 		storage := newTestAccountKeyStorage()
 		rt := newTestInterpreterRuntime()
 		runtimeInterface := getAccountKeyTestRuntimeInterface(storage)
+		addPublicKeyValidation(runtimeInterface, nil)
 
 		addAuthAccountKey(t, rt, runtimeInterface)
 
@@ -379,6 +380,7 @@ func TestRuntimeAuthAccountKeys(t *testing.T) {
 		storage := newTestAccountKeyStorage()
 		rt := newTestInterpreterRuntime()
 		runtimeInterface := getAccountKeyTestRuntimeInterface(storage)
+		addPublicKeyValidation(runtimeInterface, nil)
 
 		addAuthAccountKey(t, rt, runtimeInterface)
 
@@ -425,6 +427,7 @@ func TestRuntimeAuthAccountKeysAdd(t *testing.T) {
 
 	storage := newTestAccountKeyStorage()
 	runtimeInterface := getAccountKeyTestRuntimeInterface(storage)
+	addPublicKeyValidation(runtimeInterface, nil)
 
 	nextTransactionLocation := newTransactionLocationGenerator()
 
@@ -814,9 +817,6 @@ func accountKeyExportedValue(
 
 					// Signature Algo
 					newSignAlgoValue(signAlgo),
-
-					// valid
-					cadence.Bool(false),
 				},
 			},
 
@@ -919,6 +919,12 @@ func addAuthAccountKey(t *testing.T, runtime Runtime, runtimeInterface *testRunt
 	require.NoError(t, err)
 }
 
+func addPublicKeyValidation(runtimeInterface *testRuntimeInterface, returnError error) {
+	runtimeInterface.validatePublicKey = func(_ *PublicKey) error {
+		return returnError
+	}
+}
+
 func encodeArgs(argValues []cadence.Value) [][]byte {
 	args := make([][]byte, len(argValues))
 	for i, arg := range argValues {
@@ -1011,7 +1017,7 @@ func TestRuntimePublicKey(t *testing.T) {
 	t.Run("Constructor", func(t *testing.T) {
 		script := `
             pub fun main(): PublicKey {
-                let publicKey =  PublicKey(
+                let publicKey = PublicKey(
                     publicKey: "0102".decodeHex(),
                     signatureAlgorithm: SignatureAlgorithm.ECDSA_P256
                 )
@@ -1025,6 +1031,7 @@ func TestRuntimePublicKey(t *testing.T) {
 		runtimeInterface := &testRuntimeInterface{
 			storage: storage,
 		}
+		addPublicKeyValidation(runtimeInterface, nil)
 
 		value, err := executeScript(script, runtimeInterface)
 		require.NoError(t, err)
@@ -1037,9 +1044,6 @@ func TestRuntimePublicKey(t *testing.T) {
 
 				// Signature Algo
 				newSignAlgoValue(sema.SignatureAlgorithmECDSA_P256),
-
-				// valid
-				cadence.Bool(false),
 			},
 		}
 
@@ -1058,92 +1062,84 @@ func TestRuntimePublicKey(t *testing.T) {
             }
         `
 
-		runtimeInterface := &testRuntimeInterface{
-			validatePublicKey: func(publicKey *PublicKey) (bool, error) {
-				return true, nil
-			},
-		}
+		runtimeInterface := &testRuntimeInterface{}
 
 		_, err := executeScript(script, runtimeInterface)
 		require.Error(t, err)
 		assert.Contains(t, err.Error(), "value of type `PublicKey` has no member `validate`")
 	})
 
-	t.Run("IsValid", func(t *testing.T) {
-		for _, validity := range []bool{true, false} {
-			script := `
-              pub fun main(): Bool {
+	t.Run("Construct PublicKey in Cadence code", func(t *testing.T) {
+		script := `
+              pub fun main(): PublicKey {
                   let publicKey = PublicKey(
                       publicKey: "0102".decodeHex(),
                       signatureAlgorithm: SignatureAlgorithm.ECDSA_P256
                   )
 
-                  return publicKey.isValid
+                  return publicKey
               }
             `
+
+		fakeError := &fakeError{}
+		for _, errorToReturn := range []error{fakeError, nil} {
 			invoked := false
-			validateMethodReturnValue := validity
 
 			storage := newTestLedger(nil, nil)
 
 			runtimeInterface := &testRuntimeInterface{
 				storage: storage,
-				validatePublicKey: func(publicKey *PublicKey) (bool, error) {
+				validatePublicKey: func(publicKey *PublicKey) error {
 					invoked = true
-					return validateMethodReturnValue, nil
+					return errorToReturn
 				},
 			}
 
 			value, err := executeScript(script, runtimeInterface)
-			require.NoError(t, err)
 
-			assert.True(t, invoked)
-			assert.Equal(t, cadence.Bool(validateMethodReturnValue), value)
+			assert.True(t, invoked, "validatePublicKey was not invoked")
 
+			if errorToReturn == nil {
+				assert.NotNil(t, value)
+				assert.NoError(t, err)
+			} else {
+				assert.Nil(t, value)
+				assert.ErrorAs(t, err, &errorToReturn)
+				assert.ErrorAs(t, err, &interpreter.InvalidPublicKeyError{})
+			}
 		}
 	})
 
-	t.Run("IsValid - publicKey from host env", func(t *testing.T) {
-
+	t.Run("PublicKey from host env", func(t *testing.T) {
 		storage := newTestAccountKeyStorage()
 		storage.keys = append(storage.keys, accountKeyA, accountKeyB)
 
-		for index, key := range storage.keys {
-			script := fmt.Sprintf(
-				`
-                  pub fun main(): Bool {
-                      // Get a public key from host env
-                      let acc = getAccount(0x02)
-                      let publicKey = acc.keys.get(keyIndex: %d)!.publicKey
-                      return publicKey.isValid
-                  }
-                `,
+		for index := range storage.keys {
+			script := fmt.Sprintf(`
+                          pub fun main(): PublicKey {
+                              // Get a public key from host env
+                              let acc = getAccount(0x02)
+                              let publicKey = acc.keys.get(keyIndex: %d)!.publicKey
+                              return publicKey
+                          }`,
 				index,
 			)
 
 			invoked := false
-			validateMethodReturnValue := true
 
 			runtimeInterface := getAccountKeyTestRuntimeInterface(storage)
-			runtimeInterface.validatePublicKey = func(publicKey *PublicKey) (bool, error) {
+			runtimeInterface.validatePublicKey = func(publicKey *PublicKey) error {
 				invoked = true
-				return validateMethodReturnValue, nil
+				return nil
 			}
 
 			value, err := executeScript(script, runtimeInterface)
-			require.NoError(t, err)
 
-			// If already validated, then the validation func shouldn't get re-invoked
-			assert.NotEqual(t, key.PublicKey.Validated, invoked)
+			// skip validation when key comes from host env aka FVM
+			assert.False(t, invoked, "validatePublicKey was not invoked")
 
-			// If validated, `isValid` should have the same value as `publicKey.IsValid`.
-			// Otherwise, it should give the value returned by the `validate()` func.
-			isValid := validateMethodReturnValue
-			if key.PublicKey.Validated {
-				isValid = key.PublicKey.IsValid
-			}
-
-			assert.Equal(t, cadence.Bool(isValid), value)
+			assert.IsType(t, cadence.Struct{}, value)
+			assert.Nil(t, err)
 		}
 	})
 
@@ -1181,6 +1177,7 @@ func TestRuntimePublicKey(t *testing.T) {
 				return true, nil
 			},
 		}
+		addPublicKeyValidation(runtimeInterface, nil)
 
 		value, err := executeScript(script, runtimeInterface)
 		require.NoError(t, err)
@@ -1240,7 +1237,6 @@ func TestRuntimePublicKey(t *testing.T) {
 
                 publicKey.publicKey = []
                 publicKey.signatureAlgorithm = SignatureAlgorithm.ECDSA_secp256k1
-                publicKey.isValid = true
 
                 return publicKey
             }
@@ -1259,14 +1255,12 @@ func TestRuntimePublicKey(t *testing.T) {
 		require.ErrorAs(t, err, &checkerErr)
 
 		errs := checkerErr.Errors
-		require.Len(t, errs, 6)
+		require.Len(t, errs, 4)
 
 		assert.IsType(t, &sema.InvalidAssignmentAccessError{}, errs[0])
 		assert.IsType(t, &sema.AssignmentToConstantMemberError{}, errs[1])
 		assert.IsType(t, &sema.InvalidAssignmentAccessError{}, errs[2])
 		assert.IsType(t, &sema.AssignmentToConstantMemberError{}, errs[3])
-		assert.IsType(t, &sema.InvalidAssignmentAccessError{}, errs[4])
-		assert.IsType(t, &sema.AssignmentToConstantMemberError{}, errs[5])
 	})
 
 	t.Run("raw-key mutability", func(t *testing.T) {
@@ -1288,6 +1282,7 @@ func TestRuntimePublicKey(t *testing.T) {
 		runtimeInterface := &testRuntimeInterface{
 			storage: storage,
 		}
+		addPublicKeyValidation(runtimeInterface, nil)
 
 		_, err := executeScript(script, runtimeInterface)
 
@@ -1321,6 +1316,7 @@ func TestRuntimePublicKey(t *testing.T) {
 		runtimeInterface := &testRuntimeInterface{
 			storage: storage,
 		}
+		addPublicKeyValidation(runtimeInterface, nil)
 
 		value, err := executeScript(script, runtimeInterface)
 		require.NoError(t, err)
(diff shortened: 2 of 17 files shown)


runtime/runtime.go
```diff
@@ -3163,11 +3153,6 @@ func NewPublicKeyValue(
 			getLocationRange func() interpreter.LocationRange,
 			publicKeyValue *interpreter.CompositeValue,
 		) interpreter.BoolValue {
-			// If the public key is already validated, avoid re-validating, and return the cached result.
-			if publicKey.Validated {
-				return interpreter.BoolValue(publicKey.IsValid)
-			}
-
```
robert-e-davidson3, line 3170: Is there a good way to avoid needing to re-validate now that there's no `isValid` field to act as a cache? I don't know the codebase well enough.

It might not be enough of a performance cost to worry about though.

runtime/interpreter/value.go
```diff
@@ -11541,14 +11541,11 @@ func NewPublicKeyValue(
...
-		interpreter,
-		getLocationRange,
-		sema.PublicKeyIsValidField,
-		validatePublicKey(interpreter, getLocationRange, publicKeyValue),
-	)
+	// Validate the public key.
+	isValid := validatePublicKey(interpreter, getLocationRange, publicKeyValue)
+	if !isValid {
+		panic("invalid public key")
+	}
```
robert-e-davidson3, line 11548: This is the big change so bugs are most likely here, or as a result of this change.

runtime/convertValues_test.go
```diff
@@ -4139,8 +4013,8 @@ func TestRuntimePublicKeyImport(t *testing.T) {
 
 	t.Run("Missing raw public key", func(t *testing.T) {
 		script := `
-            pub fun main(key: PublicKey): Bool {
-                return key.isValid
+            pub fun main(key: PublicKey): PublicKey {
+                return key
```
robert-e-davidson3, line 4017: I'm not sure if this change is sensible. Since there isn't a `PublicKey.isValid` anymore, I just did the minimum change to make the syntax work.

runtime/convertValues_test.go
```diff
@@ -3912,48 +3839,6 @@ func TestRuntimePublicKeyImport(t *testing.T) {
...
-		}
-
-		_, err := executeScript(t, script, publicKey, runtimeInterface)
-		require.Error(t, err)
-
-		var argErr *InvalidEntryPointArgumentError
-		require.ErrorAs(t, err, &argErr)
-	})
-
-	t.Run("Invalid content in public key", func(t *testing.T) {
```
robert-e-davidson3, line 3936: I _think_ deleting this makes sense because it's no longer possible for `PublicKey` arguments to be invalid.

The test was failing because it was getting the new panic message I created `"invalid public key"`. For some reason this happened even when I used the `validatePublicKey` test code that I use elsewhere.

docs/language/crypto.mdx
```diff
@@ -142,8 +141,7 @@ publicKey.publicKey = []                                            // Not allow
 publicKey.publicKey[2] = 4      // No effect
 ```
 
-The validity of a public key can be checked using the `isValid` field.
-Verifications performed with an invalid public key (using `verify()` method) will always fail.
+Invalid public keys cannot be constructed so public keys are always valid.
```
robert-e-davidson3, line 144: Docs change.

Are any other docs changes needed? I see CHANGELOG has some relevant info but I'm not sure when that gets updated.

runtime/account_test.go
```diff
@@ -891,6 +884,9 @@ func getAccountKeyTestRuntimeInterface(storage *testAccountKeyStorage) *testRunt
 		decodeArgument: func(b []byte, t cadence.Type) (value cadence.Value, err error) {
 			return json.Decode(b)
 		},
+		validatePublicKey: func(publicKey *PublicKey) (bool, error) {
+			return true, nil
+		},
```
robert-e-davidson3, line 889: I believe this is necessary because before, tests would set `isValid` to `true` to bypass validation checking or let it be `false` if it didn't matter for the test. Now that invalid public keys panic, those tests fail. But I'm not confident that this isn't masking some deeper problem.

runtime/interpreter/value_test.go
```diff
@@ -2965,10 +2965,66 @@ func TestPublicKeyValue(t *testing.T) {
 		)
 
 		require.Equal(t,
-			"PublicKey(publicKey: [1, 7, 3], signatureAlgorithm: SignatureAlgorithm(rawValue: 2), isValid: true)",
+			"PublicKey(publicKey: [1, 7, 3], signatureAlgorithm: SignatureAlgorithm(rawValue: 2))",
 			key.String(),
 		)
 	})
+
+	t.Run("Panics when PublicKey is invalid", func(t *testing.T) {
```
robert-e-davidson3, line 2973: Test I added.

docs/language/crypto.mdx
turbolent: No, this should be all. It gets pulled by github.com/onflow/flow and ends up on docs.onflow.org

runtime/account_test.go
turbolent: Maybe double check where `getAccountKeyTestRuntimeInterface` is/was called and see if it expected the key to be valid/invalid. See below

runtime/account_test.go
```diff
@@ -1069,84 +1065,6 @@ func TestRuntimePublicKey(t *testing.T) {
 		assert.Contains(t, err.Error(), "value of type `PublicKey` has no member `validate`")
 	})
 
-	t.Run("IsValid", func(t *testing.T) {
```
turbolent, line 1072: Maybe instead of removing the tests, keep them and inject `validatePublicKey` once with it returning true and assert the program executes, and add another time with `validatePublicKey` returning false and assert the program aborts

runtime/convertValues_test.go
```diff
@@ -3770,72 +3763,6 @@ func TestRuntimePublicKeyImport(t *testing.T) {
 		cadence.NewUInt8(2),
 	})
 
-	t.Run("Test IsValid", func(t *testing.T) {
```
turbolent, line 3773: Don't remove the test but keep it and just remove the read of `isValid` in the body. This tests also asserts if import of public keys properly validates them, not just that `isValid` returns the appropriate value

runtime/convertValues_test.go
turbolent: Same here: Don't remove the test and keep it. The test asserts that the checking of the arguments that are sent along with scripts/transactions are properly validated. The invalid input here (strings in the key) should still result in `InvalidEntryPointArgumentError`s

runtime/interpreter/value.go
```diff
@@ -11541,14 +11541,11 @@ func NewPublicKeyValue(
...
-	publicKeyValue.SetMember(
-		interpreter,
-		getLocationRange,
-		sema.PublicKeyIsValidField,
-		validatePublicKey(interpreter, getLocationRange, publicKeyValue),
-	)
+	// Validate the public key.
+	isValid := validatePublicKey(interpreter, getLocationRange, publicKeyValue)
+	if !isValid {
+		panic("invalid public key")
```
turbolent, line 11547: Instead of using a string here, declare a type in `runtime/errors.go` and make it conform to `error`. In addition, provide position information. The easiest way to do that is by embedding `interpreter.LocationRange` and passing it from here by calling `getLocationRange()`

runtime/interpreter/value_test.go
turbolent: Good idea to also add a unit test for the internal constructor function 👍

runtime/runtime.go
turbolent: I think this was the main reason for having the `Validated` field of the internal public keys (Go type `PublicKey`, which is passed between Cadence and FVM). I'm not sure if we can assume that the keys from FVM are always valid. Maybe the PR should not get rid of the `Validated` field. 

Thoughts @tarakby ?

turbolent (commented):
Nice work!

runtime/account_test.go
SupunS, line 1072: It would be good to keep the test, and instead of asserting the `IsValid` field, we can test constructing an invalid `PublicKey` inside a cadence code, and checking if the program terminates with a panic.

Some more context to avoid any confusion: There are 3 ways a public-key can be constructed:
- Through a cadence code (this test)
- PublicKey sent by the fvm, and re-constructed at the fvm-cadence interface (below test)
- Passed in as an argument to the script/transaction ([argument passing test](https://github.com/onflow/cadence/pull/1273/files#r755736755))

runtime/account_test.go
```diff
@@ -1069,84 +1065,6 @@ func TestRuntimePublicKey(t *testing.T) {
...
-			value, err := executeScript(script, runtimeInterface)
-			require.NoError(t, err)
-
-			assert.True(t, invoked)
-			assert.Equal(t, cadence.Bool(validateMethodReturnValue), value)
-
-		}
-	})
-
-	t.Run("IsValid - publicKey from host env", func(t *testing.T) {
```
SupunS, line 1106: Similarly here, can test constructing an invalid `PublicKey` through the fvm-cadence interface.

runtime/runtime.go
SupunS: parallel reviews! 😅

turbolent: 😆

runtime/interpreter/value.go
robert-e-davidson3: @SupunS What should it contain? I'm giving it the `ArrayValue` but I'm not sure that's the best information to put there. (change pushed)

runtime/interpreter/value.go
```diff
@@ -11541,14 +11541,11 @@ func NewPublicKeyValue(
...
-	// Validate the public key, and initialize 'isValid' field.
-
-	publicKeyValue.SetMember(
-		interpreter,
-		getLocationRange,
-		sema.PublicKeyIsValidField,
-		validatePublicKey(interpreter, getLocationRange, publicKeyValue),
-	)
+	// Validate the public key.
+	isValid := validatePublicKey(interpreter, getLocationRange, publicKeyValue)
```
tarakby, line 11545: In the call to `validatePublicKey`, there will be a call to `runtimeInterface.ValidatePublicKey` which errors if the key is invalid. In that case there will be an internal panic with an error different than `InvalidPublicKeyError`. 
It would be good to return that internal panic as `InvalidPublicKeyError` too for consistency.

Moreover, that internal error contains the reason for the key invalidity, and it would be good to log it to the user. Maybe `func (e *InvalidPublicKeyError) Error()` should log it ?

runtime/runtime.go
tarakby: If I understand correctly, this function is handling keys coming from FVM? 
If yes, keys coming from the FVM are always valid, there is no need to re-validate them. 

The `validated` field existed before to avoid re-validating the key when the `isValid` field is read on Cadence. That way validation is run only once per Cadence key and the result is cached in `isValid`.

tarakby (commented):
Thanks for the updates 👌🏼 

`PublicKey` has been an immutable structure in Cadence. I just wanted to make sure this PR didn't change that (otherwise the key validation is obsolete). The runtime code is too complicated for me to understand 😅

runtime/interpreter/value.go
robert-e-davidson3: Can you point me to an example of logging we do at runtime? Or do you mean including the original error in the InvalidPublicKeyError struct, and including it in `fmt.Sprintf`?

runtime/account_test.go
robert-e-davidson3: Aha! The test runtime defaults PublicKey validation to `false` if the `validatePublicKey` field isn't specified. Most of the tests don't care if it's valid or not.

I'll make it default to `true`.

runtime/interpreter/value.go
robert-e-davidson3: What should I store in that struct?

I tried to settle on using a string because the places in which I need to instantiate `InvalidPublicKey` do not have the same data type for PublicKey. It's either `ArrayValue` or `CompositeValue`, though `[]byte` seems ideal. There's code to convert `ArrayValue` to `CompositeValue` for `PublicKey` but it does validation so I would need to refactor it.

But using a string causes the linter to fail because I use `fmt.Sprintf("%s", foo)`. Unfortunately there is no `String(foo)` like it suggests and `string(foo)` can't actually serialize `ArrayValue`.

I made a hack around this by adding a space like `%s ` but that's not great.

I suppose I could instead make the error struct take several types and have it print the first one that isn't `nil`... thoughts?

---

linter error:
https://github.com/onflow/cadence/runs/4388776276?check_suite_focus=true#step:5:24

runtime/interpreter/value.go
tarakby: I meant the latter, including the original error text in the returned `InvalidPublicKeyError`

runtime/runtime.go
```diff
@@ -786,7 +802,18 @@ func validateArgumentParams(
 			}
 		}
 
-		arg, err := importValue(inter, value, parameterType)
+		var arg interpreter.Value
+		maybeErr := panicToInvalidPublicKeyError(func() {
```
robert-e-davidson3, line 806: Needed to wrap the panic caused by validation of invalid public key, which happens in `importValue`.

SupunS (commented):
Nice! Overall looks good!
Just [one question](https://github.com/onflow/cadence/pull/1273#discussion_r764449746) over the re-validation of public keys.

runtime/runtime.go
```diff
@@ -786,7 +802,25 @@ func validateArgumentParams(
...
+		var arg interpreter.Value
+		panicError := panicToError(func() {
+			// if importing an invalid public key, this call panics
+			arg, err = importValue(inter, value, parameterType)
+		})
+
+		if panicError != nil {
+			// The only potential user error here is InvalidPublicKeyError,
+			// so panic if any other error occurs.
+			if _, ok := panicError.(interpreter.InvalidPublicKeyError); ok {
```
turbolent, line 814: Why do we have to special case `interpreter.InvalidPublicKeyError` here? Maybe just do it for any recovered error

turbolent (commented):
The error handling is a bit messy at the moment. There's a parallel effort to clean up error handling, specifically propagating user and implementation errors properly and more elegantly, so eventually we will be able to clean up the error handling for this too, but we don't have to wait for it

runtime/runtime.go
```diff
@@ -786,7 +802,25 @@ func validateArgumentParams(
...
+				return nil, &InvalidEntryPointArgumentError{
+					Index: i,
+					Err:   panicError,
+				}
+			} else {
+				panic(panicError)
+			}
+		}
```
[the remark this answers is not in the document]
SupunS: ah yes, I misunderstood what that method does 👍

runtime/runtime.go
robert-e-davidson3: My understanding is that we only want to turn user-caused panics into errors. Here that should only be `InvalidPublicKeyError` so everything else should actually panic.

Would it make more sense to wrap everything that isn't a go panic? Like how `wrapPanic` does it but returning an error instead of panicking with an `ExternalError`.

runtime/account_test.go
```diff
@@ -1058,92 +1050,96 @@ func TestRuntimePublicKey(t *testing.T) {
...
 
+				value, err := executeScript(script, runtimeInterface)
+
+				assert.True(t, invoked, "validatePublicKey was invoked")
+
+				if panics {
+					assert.Nil(t, value)
+					assert.Error(t, err)
+					// verifies the exact error
+					assert.Equal(t, fakeError, errors.Unwrap(errors.Unwrap(err)))
```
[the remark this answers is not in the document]
turbolent: `assert.ErrorIs` uses [`errors.Is`](https://pkg.go.dev/errors#Is), which compares by equality. 
Given that you are likely only interested in asserting that the Go error type matches, you probably want to use `assert.ErrorAs` (which uses [`errors.As`](https://pkg.go.dev/errors#As))

runtime/runtime.go
robert-e-davidson3: I went with keeping `panicToError` since it seems like a good function for composability. Then I call it within `userPanicToError`, which panics at go errors while returning every other error. So no more special case for `InvalidPublicKeyError`.

runtime/runtime.go
```diff
@@ -3215,23 +3234,23 @@ func validatePublicKey(
 	getLocationRange func() interpreter.LocationRange,
 	publicKeyValue *interpreter.CompositeValue,
 	runtimeInterface Interface,
-) interpreter.BoolValue {
-
+) (interpreter.BoolValue, error) {
```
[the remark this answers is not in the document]
robert-e-davidson3: @turbolent Is it fine to leave this returning two values?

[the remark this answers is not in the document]
turbolent: Sorry, I missed this! I'm unfortunately not quite following, I'll dedicate some time tomorrow to try this and investigate.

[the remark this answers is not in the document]
turbolent: @tarakby When would decoding the public key fail, and it is not an `invalidInputsError`? Are there really three different outcomes for public key validation, and if so, what are those?

runtime/runtime_test.go
```diff
@@ -438,7 +438,7 @@ func (i *testRuntimeInterface) ImplementationDebugLog(message string) error {
 
 func (i *testRuntimeInterface) ValidatePublicKey(key *PublicKey) (bool, error) {
 	if i.validatePublicKey == nil {
-		return false, nil
+		return true, nil
```
turbolent, line 441: The default implementation for all mock functions is to fail. If a successful result is expected, I'd recommend passing a `validatePublicKey` function in, i.e. `alwaysValidates` which got introduced above

runtime/runtime.go
```diff
@@ -2984,7 +3029,7 @@ func (r *interpreterRuntime) newAccountKeysGetFunction(
 					inter,
 					invocation.GetLocationRange,
 					accountKey,
-					inter.PublicKeyValidationHandler,
+					alwaysValidates, // key from FVM has already been validated
```
turbolent, line 3032: @tarakby Can we assume that the FVM always returns keys that are both validated *and* valid?

runtime/runtime.go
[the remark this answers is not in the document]
tarakby: The key decoding fails with a non-`invalidInputsError` when there was an unexpected execution error (C layer issue, memory..).

A key validation outputs: 
 - `true, nil` when the bytes represent a valid key and there was no execution error
 - `false, nil` when the bytes do not represent a valid key and there was no execution error ( `invalidInputsError` from `DecodePublicKey` shows that the execution went well but the bytes do not represent any valid key)
 - `false, error` is returned on **master** when an unexpected error occurred. **However**, this [PR](https://github.com/onflow/flow-go/pull/1605) updates this case to panic instead. The PR is not merged yet, it looks like it's blocked by the current PR: https://github.com/onflow/flow-go/pull/1605#issuecomment-983860729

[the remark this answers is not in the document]
turbolent: Thanks for the explanation @tarakby! In that case it's probably best to only return a single return value, an error, and `nil` means valid, and non-`nil` a user error. If the FVM side then runs into an `invalidInputsError` , which is an implementation error, it should panic (like in the linked open PR), it should not be returned (confusion of an implementation error as a user error)

runtime/runtime.go
tarakby: Just to make sure I'm answering the right question, this PR is about the raw `PublicKey` structure, while the function above is about an account public key (wrapper around a raw public key). 

But in general, yes data coming from the FVM is validated and valid.

runtime/runtime.go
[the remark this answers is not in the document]
robert-e-davidson3: I changed it so `ValidatePublicKey` only returns an error.

I assume this will break the FVM-Cadence integration because the `Interface.ValidatePublicKey` method in `cadence/runtime/interface.Go` changed. Do I need to do anything to resolve that?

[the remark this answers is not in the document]
turbolent: @robert-e-davidson3 FVM depends on Cadence, so once we release Cadence, we will have to update FVM and adjust the implementation there. There's nothing else to do here on the Cadence side

runtime/account_test.go
```diff
@@ -1058,92 +1060,95 @@ func TestRuntimePublicKey(t *testing.T) {
...
             `
+
+		fakeError := &fakeError{}
+		for _, errorToReturn := range []error{fakeError, nil} {
 			invoked := false
-			validateMethodReturnValue := validity
 
 			storage := newTestLedger(nil, nil)
 
+
```
turbolent, line 1088:
```suggestion
```

runtime/account_test.go
```diff
@@ -1058,92 +1060,95 @@ func TestRuntimePublicKey(t *testing.T) {
...
+			//	assert.Nil(t, value)
+			//	assert.Error(t, err)
+			//	assert.ErrorAs(t, err, &fakeError)
+			//} else if validity {
+			//	assert.NotNil(t, value)
+			//	assert.NoError(t, err)
+			//} else {
+			//	assert.Error(t, err)
+			//	assert.ErrorAs(t, err, &interpreter.InvalidPublicKeyError{})
+			//}
```
turbolent, line 1118:
```suggestion
```

runtime/account_test.go
```diff
@@ -1058,92 +1060,95 @@ func TestRuntimePublicKey(t *testing.T) {
...
 
-			assert.True(t, invoked)
-			assert.Equal(t, cadence.Bool(validateMethodReturnValue), value)
+			assert.True(t, invoked, "validatePublicKey was not invoked")
+
+			if errorToReturn == nil {
+				assert.NotNil(t, value)
+				assert.NoError(t, err)
+			} else {
+				assert.Nil(t, value)
```
turbolent, line 1105: Maybe add an assertion for the error here, e.g. `err` is the same as `errorToReturn`

runtime/convertValues_test.go
```diff
@@ -3895,6 +3889,9 @@ func TestRuntimePublicKeyImport(t *testing.T) {
 				verifyInvoked = true
 				return true, nil
 			},
+			validatePublicKey: func(publicKey *PublicKey) error {
```
turbolent, line 3892: Maybe just use `addPublicKeyValidation(runtimeInterface, nil)` here

runtime/crypto_test.go
```diff
@@ -91,6 +91,7 @@ func TestRuntimeCrypto_verify(t *testing.T) {
 			assert.Equal(t, HashAlgorithmSHA3_256, hashAlgorithm)
 			return true, nil
 		},
+		validatePublicKey: func(_ *PublicKey) error { return nil },
```
turbolent, line 94: Maybe use `addPublicKeyValidation(runtimeInterface, nil)` here and below

runtime/interface.go
```diff
@@ -120,7 +120,7 @@ type Interface interface {
 	// ImplementationDebugLog logs implementation log statements on a debug-level
 	ImplementationDebugLog(message string) error
 	// ValidatePublicKey verifies the validity of a public key.
-	ValidatePublicKey(key *PublicKey) (bool, error)
+	ValidatePublicKey(key *PublicKey) error
```
turbolent, line 123: 👍

runtime/interpreter/errors.go
```diff
@@ -591,3 +591,19 @@ func (e InvalidOperandsError) Error() string {
...
+	PublicKey *ArrayValue
+	Err       error
+	LocationRange
+}
+
+func (e *InvalidPublicKeyError) Error() string {
+	return fmt.Sprintf("invalid public key %s", e.PublicKey)
+}
+
+
```
turbolent, line 606:
```suggestion
```

runtime/interpreter/value.go
```diff
@@ -14857,14 +14857,15 @@ func NewPublicKeyValue(
...
-	publicKeyValue.SetMember(
-		interpreter,
-		getLocationRange,
-		sema.PublicKeyIsValidField,
-		validatePublicKey(interpreter, getLocationRange, publicKeyValue),
-	)
+	var err error
+	err = validatePublicKey(interpreter, getLocationRange, publicKeyValue)
+	if err != nil {
+		panic(&InvalidPublicKeyError{
```
turbolent, line 14863:
```suggestion
		panic(InvalidPublicKeyError{
```

runtime/interpreter/value_test.go
```diff
@@ -3177,10 +3165,61 @@ func TestPublicKeyValue(t *testing.T) {
...
+			NewIntValueFromInt64(int64(publicKeyBytes[2])),
+		)
+
+		sigAlgo := stdlib.NewSignatureAlgorithmCase(
+			inter,
+			sema.SignatureAlgorithmECDSA_secp256k1.RawValue(),
+		)
+
+		assert.PanicsWithError(t,
+			(&InvalidPublicKeyError{PublicKey: publicKey, Err: fakeError}).Error(),
```
turbolent, line 3212:
```suggestion
			(InvalidPublicKeyError{PublicKey: publicKey, Err: fakeError}).Error(),
```

runtime/convertValues_test.go
```diff
@@ -3821,27 +3810,35 @@ func TestRuntimePublicKeyImport(t *testing.T) {
...
 					assert.True(t, publicKeyValidated)
-					assert.Equal(t, actual, cadence.NewBool(publicKeyActualValidity))
+
+					// Invalid PublicKey errors but valid does not.
+					if publicKeyActualError == nil {
+						require.NoError(t, err)
+					} else {
+						assert.Error(t, err)
+						var invalidEntryPointArgumentError *InvalidEntryPointArgumentError
+						var invalidPublicKeyError *interpreter.InvalidPublicKeyError
```
turbolent, line 3830:
```suggestion
						var invalidPublicKeyError interpreter.InvalidPublicKeyError
```

runtime/interpreter/errors.go
```diff
@@ -591,3 +591,19 @@ func (e InvalidOperandsError) Error() string {
...
 }
+
+// InvalidPublicKeyError is reported during PublicKey creation, if the PublicKey is invalid.
+type InvalidPublicKeyError struct {
+	PublicKey *ArrayValue
+	Err       error
+	LocationRange
+}
+
+func (e *InvalidPublicKeyError) Error() string {
```
turbolent, line 602:
```suggestion
func (e InvalidPublicKeyError) Error() string {
```

runtime/interpreter/errors.go
```diff
@@ -591,3 +591,19 @@ func (e InvalidOperandsError) Error() string {
...
+	return fmt.Sprintf("invalid public key %s", e.PublicKey)
+}
+
+
+func (e *InvalidPublicKeyError) Unwrap() error {
```
turbolent, line 607:
```suggestion
func (e InvalidPublicKeyError) Unwrap() error {
```

runtime/interpreter/value.go
```diff
@@ -14857,14 +14857,15 @@ func NewPublicKeyValue(
...
-	// Validate the public key, and initialize 'isValid' field.
-
-	publicKeyValue.SetMember(
-		interpreter,
-		getLocationRange,
-		sema.PublicKeyIsValidField,
-		validatePublicKey(interpreter, getLocationRange, publicKeyValue),
-	)
+	var err error
+	err = validatePublicKey(interpreter, getLocationRange, publicKeyValue)
```
turbolent, line 14861:
```suggestion
	err := validatePublicKey(interpreter, getLocationRange, publicKeyValue)
```

runtime/interpreter/value_test.go
```diff
@@ -3154,18 +3151,9 @@ func TestPublicKeyValue(t *testing.T) {
 			NewIntValueFromInt64(3),
 		)
 
-		sigAlgo := NewCompositeValue(
+		sigAlgo := stdlib.NewSignatureAlgorithmCase(
```
turbolent, line 3154: Good idea! 👍

turbolent (commented):
Great work! Just a few minor things, but looks great overall 👏

runtime/interpreter/value.go
robert-e-davidson3: Oops! I did this because the Goland debugger was complaining that it didn't know what `err` was.

turbolent (approved):
Nice work! 👍 

Fixed the last linting errors. @SupunS Could you please have a final look given this is a critical component?

runtime/runtime.go
```diff
@@ -3521,3 +3543,11 @@ func hash(
 
 	return interpreter.ByteSliceToByteArrayValue(inter, result)
 }
+
+func alwaysValidates(
```
SupunS, line 3547: This function actually 'neverValidates'? 🤔  Maybe rename it to something like `noOp`?

SupunS (approved):
Look good! 🎉

runtime/runtime.go
turbolent: Haha, naming is hard. I guess the intention was here "the result is always valid", but you are right, the function does not perform any validation, it could be misunderstood. I do not have a good idea, but would recommend to still include something in the name that expresses the function is for public key validation – the parameters do not indicate it, and the function could get accidentally used for other purposes and its behaviour changed in the future.

Also, the function seems to be duplicated in the test code. Maybe export here and remove the duplicate in the test code.

tarakby:
Thank you everyone for the push to merge this PR 🙏🏼 

Just to make sure the PR didn't update an existing behaviour: is `PublicKey` still an immutable structure in Cadence?

SupunS:
> Just to make sure the PR didn't update an existing behavior: is PublicKey still an immutable structure in Cadence?

Yup, this is still true. There are some tests in place to verify this behavior: https://github.com/onflow/cadence/blob/3534776138b44a9a4dce18ecdc38d75335e5d178/runtime/account_test.go#L1230-L1333

runtime/runtime.go
robert-e-davidson3: Clarified!

Issue #1269, opened 2021-11-22
Move linter to a separate phase


SupunS:
### Issue To Be Solved

Currently, linter errors (i.e: hints such as redundant casts, etc.) are also checked and reported at the `checker`. Moving these non-erroneous checks /code-smells to a separate phase could make the checker cleaner. It also allows enabling/disabling such advanced but non-mandatory checks depending on the use-case (e.g: on-chain vs dev-tools)

turbolent:
I really like the simplicity of [Go's `analysis` package](https://pkg.go.dev/golang.org/x/tools/go/analysis).
It should be possible to implement something similar for Cadence with little effort: We already have the AST walking implemented, and have to make the elaboration available. It does not provide complete type information for all elements, but should be sufficient for most tasks, and could be extended.

turbolent:
The analysis framework is now in place, so we can easily refactor the existing linters to be based on the new framework. 
For example, they could be moved to the new contract analyzer tool and eventually to the CLI.

turbolent:
Once #1732 is merged, we basically have a linter which can be extended with more linting rules / analyzers.

Two follow-up tasks:
- [ ] Maybe move this tool from `tools/contract-analyzer` into a repo `cadence-lint`
- [ ] Refactor the existing linting rules out of the checker and into separate analyzers (instances can be found by searching for `lintingEnabled` in `sema`)

turbolent:
With Secure Cadence released, now would be a good time to clean up this tech debt and turn it into a more useful tool.

@dsainati1 would you be maybe interested in picking this up?

dsainati1:
Sure! This seems cool

PR #1275, opened 2021-11-24
Disable existing hints if linting is disabled

turbolent:
Depends on #1072 

### Description

Just like disabling the new hint in #1072, also disable the existing hints when linting is disabled (the default).

______

### DIFF
--- a/runtime/sema/check_casting_expression.go
+++ b/runtime/sema/check_casting_expression.go
@@ -124,7 +124,7 @@ func (checker *Checker) VisitCastingExpression(expression *ast.CastingExpression
 						Range:        ast.NewRangeFromPositioned(leftHandExpression),
 					},
 				)
-			} else if IsSubType(leftHandType, rightHandType) {
+			} else if checker.lintEnabled && IsSubType(leftHandType, rightHandType) {
 
 				switch expression.Operation {
 				case ast.OperationFailableCast:
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -3126,14 +3126,16 @@ func numberFunctionArgumentExpressionsChecker(targetType Type) ArgumentExpressio
 		switch argument := argument.(type) {
 		case *ast.IntegerExpression:
 			if CheckIntegerLiteral(argument, targetType, checker.report) {
-
-				suggestIntegerLiteralConversionReplacement(checker, argument, targetType, invocationRange)
+				if checker.lintEnabled {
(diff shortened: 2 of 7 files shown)


runtime/sema/check_casting_expression.go
```diff
@@ -124,7 +124,7 @@ func (checker *Checker) VisitCastingExpression(expression *ast.CastingExpression
 						Range:        ast.NewRangeFromPositioned(leftHandExpression),
 					},
 				)
-			} else if IsSubType(leftHandType, rightHandType) {
+			} else if checker.lintEnabled && IsSubType(leftHandType, rightHandType) {
```
dsainati1, line 127: Instead of having to add multiple checks, why not add a check in `checker.Hint`? We can add a new hint only if `lintEnabled` is true, and the function can be a no-op otherwise.

runtime/sema/type.go
```diff
@@ -3126,14 +3126,16 @@ func numberFunctionArgumentExpressionsChecker(targetType Type) ArgumentExpressio
 		switch argument := argument.(type) {
 		case *ast.IntegerExpression:
 			if CheckIntegerLiteral(argument, targetType, checker.report) {
-
-				suggestIntegerLiteralConversionReplacement(checker, argument, targetType, invocationRange)
+				if checker.lintEnabled {
```
dsainati1, line 3129: same as above, maybe move this to `checker.Hint`?

runtime/sema/type.go
```diff
@@ -3126,14 +3126,16 @@ func numberFunctionArgumentExpressionsChecker(targetType Type) ArgumentExpressio
...
+					suggestIntegerLiteralConversionReplacement(checker, argument, targetType, invocationRange)
+				}
 			}
 
 		case *ast.FixedPointExpression:
 			if CheckFixedPointLiteral(argument, targetType, checker.report) {
-
-				suggestFixedPointLiteralConversionReplacement(checker, targetType, argument, invocationRange)
+				if checker.lintEnabled {
```
dsainati1, line 3136: same as above

dsainati1 (commented):
This makes sense; I also don't think it makes sense to report hints when linting is not enabled.

runtime/sema/check_casting_expression.go
turbolent: The expensive part is not the reporting, but the checks for the linting. There is a follow-up issue to move the linting code out of the checker in #1269, the conditionals are only temporary to optimize the current implementation.

SupunS (approved):

Issue #209, opened 2020-07-04
Implement pretty print for Cadence


MaxStalker:
### Issue To Be Solved

At the moment there is no pretty print solution for Cadence, which leads to simple mistakes - double quotes for String literals, for example - and waste of time, when you would reformat your code to make it more readable. 

### (Optional): Suggest A Solution
[Prettier](https://prettier.io/) is a de-facto goto solution for pretty print in web development. And it have support for [plugins](https://prettier.io/docs/en/plugins.html). Main repository also have multiple implementations - [Swift, for example](https://github.com/prettier/plugin-swift/) - which can serve as inspiration and guidance to build one for Cadence.

turbolent:
A formatter / pretty printer is something we've been wanting for quite a while, but have not gotten around too. 

There are two things we need before we can pretty print:

1. Generate JSON from the AST where all position info (start and end) is fully calculated.
   Currently it is sometimes calculated from child nodes (for performance reasons: less work if not needed)

2. Add support for keeping "trivia" in the AST (that’s e.g. what it is called in Swift’s AST), i.e. leading and trailing whitespace and comments

Both aren’t hard, just a bit of work.

Once we add support for trivia, we can then also use it for other purposes, e.g. generate nice API documentation, use it as doc strings in the completion items in the VS Code extension.

I think if Prettier provides us some infrastructure which would save us some work, we should use it.
As far as I understand it provides a pretty print document AST, and the task of a plugin is to translate its AST to Prettier's AST.

As an aside: The other day I played with Go’s WebAssembly target, and successfully compiled the new parser to WebAssembly in the browser. The code can be found in an unmerged branch: https://github.com/onflow/cadence/compare/bastian/astexplorer. I think we could use it to maybe avoid having to shell out to the parser (and have users build the Go parser)

turbolent:
I opened #210 for the JSON serialization improvements for the AST

turbolent:
I opened a first PR for #210 to demonstrate what needs to be done: #211

turbolent:
Created #308 for retaining comments in the AST

turbolent:
After all 5 PRs are merged, #210 should be done, and the AST is ready to be consumed!

For development, the `cmd/parse` tool can be used with the `json` flag to get the output.

Given a file `program.cdc`:

```
struct X {}
```

The JSON representation can be obtained with:

```sh
$ go run ./cmd/parse -json program.cdc
[
  {
    "path": "program.cdc",
    "Program": {
[...]

PR #1280, opened 2021-11-29
Pretty print statements

turbolent:
Work towards #209

## Description

Implement pretty-printing functions for all statement AST nodes:

- `ReturnStatement`
- `BreakStatement`
- `ContinueStatement`
- `IfStatement`
- `WhileStatement`
- `ForStatement`
- `EmitStatement`
- `AssignmentStatement`
- `SwapStatement`
- `ExpressionStatement`

Also:
- Add missing JSON marshaling tests
- Allow inserting tabs (4 spaces) in the pretty printer tool's text area

______

### DIFF
--- a/runtime/ast/statement.go
+++ b/runtime/ast/statement.go
@@ -403,6 +403,10 @@ type ExpressionStatement struct {
 	Expression Expression
 }
 
+var _ Statement = &ExpressionStatement{}
+
+func (*ExpressionStatement) isStatement() {}
+
 func (s *ExpressionStatement) StartPosition() Position {
 	return s.Expression.StartPosition()
 }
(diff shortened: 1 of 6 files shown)


runtime/ast/statement.go
```diff
@@ -251,6 +331,36 @@ func (s *ForStatement) EndPosition() Position {
 	return s.Block.EndPosition()
 }
 
+const forStatementForKeywordSpaceDoc = prettier.Text("for ")
+const forStatementSpaceInKeywordSpaceDoc = prettier.Text(" in ")
+
+func (s *ForStatement) Doc() prettier.Doc {
+	doc := prettier.Concat{
```
SupunS, line 338: need to concat?

runtime/ast/statement.go
```diff
@@ -251,6 +331,36 @@ func (s *ForStatement) EndPosition() Position {
...
+		forStatementForKeywordSpaceDoc,
+	}
+
+	if s.Index != nil {
+		doc = append(
```
SupunS, line 343: Is there a difference between using an array (append) vs `prettier.Concat{}`

SupunS (approved):
Nice!

runtime/ast/statement.go
turbolent: Yes, we always need the concatenation, as https://github.com/onflow/cadence/pull/1280/files#diff-cb6a372e592ff206bc8f5ad13e85ba6527d1e6914f281f0dd263479c7f3a278eR350 always appends more after the conditionally appended index (https://github.com/onflow/cadence/pull/1280/files#diff-cb6a372e592ff206bc8f5ad13e85ba6527d1e6914f281f0dd263479c7f3a278eR342)

runtime/ast/statement.go
turbolent: `prettier.Concat` is a prettier document, but is also just a slice of documents, so we can use the Go `append` to add to it

PR #1281, opened 2021-11-29
Pretty print types

turbolent:
Work towards #209

## Description

Implement pretty-printing functions for all type AST nodes:

- `TypeAnnotation`
- `NominalType`
- `OptionalType`
- `VariableSizedType`
- `ConstantSizedType`
- `DictionaryType`
- `FunctionType`
- `ReferenceType`
- `RestrictedType`
- `InstantiationType`

Also, include type annotations in expressions.

______

### DIFF
--- a/runtime/ast/type.go
+++ b/runtime/ast/type.go
@@ -140,6 +140,8 @@ type OptionalType struct {
 	EndPos Position `json:"-"`
 }
 
+var _ Type = &OptionalType{}
+
 func (*OptionalType) isType() {}
 
 func (t *OptionalType) String() string {
@@ -152,6 +154,15 @@ func (t *OptionalType) StartPosition() Position {
 
 func (t *OptionalType) EndPosition() Position {
 	return t.EndPos
+}
+
+const optionalTypeSymbolDoc = prettier.Text("?")
+
+func (t *OptionalType) Doc() prettier.Doc {
+	return prettier.Concat{
+		t.Type.Doc(),
+		optionalTypeSymbolDoc,
+	}
 }
 
 func (t *OptionalType) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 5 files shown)


robert-e-davidson3 (approved):

runtime/ast/type.go
```diff
@@ -314,6 +454,22 @@ func (t *ReferenceType) EndPosition() Position {
...
+const referenceTypeAuthKeywordSpaceDoc = prettier.Text("auth ")
+const referenceTypeSymbolDoc = prettier.Text("&")
+
+func (t *ReferenceType) Doc() prettier.Doc {
+	var doc prettier.Concat
+	if t.Authorized {
+		doc = append(doc, referenceTypeAuthKeywordSpaceDoc)
+	}
+
+	return append(
```
SupunS, line 466: same question as in the prev PR, should `prettier.Concat{}` be used here?

SupunS (approved):
🎉

runtime/ast/type.go
```diff
@@ -271,6 +367,48 @@ func (t *FunctionType) String() string {
...
+				functionTypeStartDoc,
+				prettier.Indent{
+					Doc: parametersDoc,
+				},
+				prettier.SoftLine{},
+				functionTypeEndDoc,
+			},
+		},
+		functionTypeTypeSeparatorSpaceDoc,
+		t.ReturnTypeAnnotation.Doc(),
```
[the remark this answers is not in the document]
turbolent: Unlike function declarations, function types must always specify the return type explicitly, so `fun () {}` has type `((): Void)`

runtime/ast/type.go
turbolent: `prettier.Concat`, like other documents, can be nested, so this could be `return prettier.Concat{doc, ...` here too, but there is no semantic difference between nesting of flattening them, expect for the latter being more efficient.

PR #1286, opened 2021-11-30
`load`, `copy` and `borrow` cause a force cast 

### LINKED ISSUE #1247: Improve Storage API of `load`, `copy` and `borrow`
Originally proposed in #376:
> 
> The load, copy, and borrow functions, as well as the capability borrow function, currently return nil both when the storage location is empty, and when the storage location contains a value, but is does not have the requested type. The inability to differentiate the two cases makes debugging hard.
> 
> The function has basically three return value cases:
> 
>     Nothing is stored
>     A value is stored, but it does not have the requested type
>     A value is stored, and it has the requested type
> 
> For case 2:
> 
>     Maybe abort the program, i.e. treat the requested type as a force-cast.
>     Maybe the emulator should log a warning that
>

dsainati1:
Closes #1247

## Description

This changes the storage API to cause `load`, `copy` and `borrow` to produce a runtime error if the supplied type argument does not match the stored value. 

______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -3567,7 +3567,11 @@ func (interpreter *Interpreter) authAccountBorrowFunction(addressValue AddressVa
 			// which reads the stored value
 			// and performs a dynamic type check
 
-			if reference.ReferencedValue(interpreter) == nil {
+			value, errFunc := reference.dereference(interpreter)
+			if errFunc != nil {
+				panic(errFunc(invocation.GetLocationRange()))
+			}
+			if value == nil {
 				return NilValue{}
 			}
 
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -3500,7 +3500,10 @@ func (interpreter *Interpreter) authAccountReadFunction(addressValue AddressValu
 
 				dynamicType := value.Value.DynamicType(interpreter, SeenReferences{})
 				if !interpreter.IsSubType(dynamicType, ty) {
-					return NilValue{}
+					panic(ForceCastTypeMismatchError{
+						ExpectedType:  ty,
+						LocationRange: invocation.GetLocationRange(),
+					})
 				}
 
 				inter := invocation.Interpreter
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -10176,7 +10176,7 @@ func (v *StorageReferenceValue) StaticType() StaticType {
 	}
 }
 
-func (v *StorageReferenceValue) ReferencedValue(interpreter *Interpreter) *Value {
+func (v *StorageReferenceValue) dereference(interpreter *Interpreter) (*Value, func(LocationRange) error) {
 	switch referenced := interpreter.ReadStored(v.TargetStorageAddress, v.TargetKey).(type) {
 	case *SomeValue:
 		value := referenced.Value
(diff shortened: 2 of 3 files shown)


runtime/interpreter/interpreter.go
```diff
@@ -3564,7 +3567,11 @@ func (interpreter *Interpreter) authAccountBorrowFunction(addressValue AddressVa
 			// which reads the stored value
 			// and performs a dynamic type check
 
-			if reference.ReferencedValue(interpreter) == nil {
```
SupunS, line 3567: Do we need to do the same for the other places like `capabilityBorrowFunction()`, `exportValueWithInterpreter`, etc ?

dsainati1: What do you mean?

dsainati1:
Does this change merit a FLIP? On the one hand, it is a pretty small change design/implementation wise, and there is not much to discuss on those fronts. On the other hand, it has a pretty widespread affect on how people write code. I am curious whether or not this would benefit from going through the full FLIP process.

runtime/interpreter/interpreter.go
SupunS: There are a few other places as well (like: `capabilityBorrowFunction`, `exportValueWithInterpreter`) where the value is dereferenced (`reference.ReferencedValue()` is used). I was wondering whether do we need to panic in those places as well or continue to silently fail.

dsainati1: Ah, yea that is a question worth considering. For the scope of this PR I think just changing `load`, `copy` and `borrow` makes sense, but if we want to extend this behavior to anywhere we query storage then we would need to change the behavior there was well.

SupunS: The original issue has mentioned about handling the capability burrow as well.
> The load, copy, and borrow functions, as well as the capability borrow function, currently return nil both when the storage location is empty...

That's why I was wondering.
Is the plan to handle that separately? or am I missing something.

dsainati1: ah I missed that the capability borrow function was also mentioned here. I will add that to this PR.

SupunS (commented):
Looks good! 🎉

turbolent:
> Does this change merit a FLIP? 

Yes, we should definitely pitch this to the community, then if there is interest, propose the breaking change through a FLIP and get agreement on it.

Thank you for already creating a PR with the changes 👍 

I could imagine putting this change behind a feature flag / option, then release it, disabled by default, in the emulator.
This would allow users to evaluate the impact of the changes on their code.

dsainati1:
The FLIP outlining this change has been approved and merged: https://github.com/onflow/flow/pull/718

runtime/interpreter/value.go
```diff
@@ -10314,25 +10314,37 @@ func (v *StorageReferenceValue) StaticType() StaticType {
 	}
 }
 
-func (v *StorageReferenceValue) ReferencedValue(interpreter *Interpreter) *Value {
-
+func (v *StorageReferenceValue) dereference(interpreter *Interpreter) (*Value, func(LocationRange) error) {
```
turbolent, line 10317: Maybe make this more Go idiomatic, by returning just an error value. As the location range is needed, add a `getLocationRange func() LocationRange` parameter to the `dereference` function itself. This pattern is used by many other functions in the codebase.

turbolent (commented):
Nice, looks good! Just one comment regarding generating the error value, which requires the location range

turbolent (approved):
Nice! Ready to get merged once the FLIP has been approved

PR #1303, opened 2021-12-07
Taking a reference to an optional value now produces an optionally-typed reference

### LINKED ISSUE #747: BUG: Should not be able to statically access the fields of a optionally casted reference
### Problem

When creating a reference to an object stored in a dictionary using `as?`, the checker does not report a static error when trying to access a field of the optional reference.

### Steps to Reproduce

1. Copy this contract
```
pub contract TestCollection {
    pub resource IDRes {
        pub let id: UInt32
        init() {
            self.id = 0
        }
    }

    pub resource Collection {
        access(self) var idResources: @{String: IDRes}
        init() {
            self.idResources <- {}
        }

        destroy() {
            destroy self.idResources
        }

        pub fun borrowID(_ key: String, _ id: UInt32): UInt32 {
            let idref = &self.idResources[key] as? &IDRes

            let id = idref.id

            return id
        }
    }

    init() {
        let col <- create Collection()

        let del = col.borrowID("hello", 1)

        destroy col
    }
}
```
2. Deploy the contract and see this error:
```
WARN[0004] ERR [08e509] Execution failed:
error: dereference failed
  --> 01cf0e2f2f715450.TestCollection:23:21
   |
23 |             let id = idref.id
```

### Acceptance Criteria

It should cause a static error because it is trying to access the field of an optional, right?

dsainati1:
Closes #747

## Description

It was possible to obtain a reference to an optional by using a dictionary access index expression, which could then have its fields accessed as if it were not optional, leading to the bug in the linked issue. 

This removes the ability to create a reference to an optional this way. Instead, to create a reference to the contents of a dictionary, one can create an optional reference like so:

```cadence
let dict: {String : S} = {}
let s = &dict[""] as &S? // s has type &S?
```

When the expression on the left of the `as` is optional, the target type on the right side must also be optional to match.

This is described in more detail in this FLIP: https://github.com/onflow/flow/pull/722 
______


dsainati1:
Note that this breaks the Topshot contract test, which implies that the existing Topshot contract would rely on the current, buggy behavior. I am unsure what the best way to resolve this is, or whether it simply indicates that this approach is flawed.

turbolent:
This was not a bug, but a feature. I still think it makes sense to remove it, but we should pitch it to the community and then propose a FLIP before merging this, as it is a breaking change and some contracts rely on this feature.

Could you please remove the `Bugfix` label and add the  `Breaking Change` label?

runtime/interpreter/interpreter_expression.go
```diff
@@ -819,10 +819,23 @@ func (interpreter *Interpreter) VisitReferenceExpression(referenceExpression *as
 
 	result := interpreter.evalExpression(referenceExpression.Expression)
 
-	return &EphemeralReferenceValue{
-		Authorized:   borrowType.Authorized,
-		Value:        result,
-		BorrowedType: borrowType.Type,
+	switch result := result.(type) {
```
turbolent, line 822: I'm not quite sure, but I think the type check shouldn't be on the dynamic type of the value, but the static type of the referenced expression

runtime/sema/check_reference_expression.go
```diff
@@ -34,9 +34,16 @@ func (checker *Checker) VisitReferenceExpression(referenceExpression *ast.Refere
 
 	var referenceType *ReferenceType
 	var targetType, referencedType Type
+	var returnType Type
 
 	if !resultType.IsInvalidType() {
 		var ok bool
+		// indexed access to dictionaries require an optional reference as their
+		// target type, so allow one level of optionals
```
turbolent, line 42: Maybe just mention this as an example and document the general case:

 ```suggestion
		// Reference expressions may reference a value which has an optional type.
		// For example, the result of indexing into a dictionary is an optional:
		//
		// let ints: {Int: String} = {0: "zero"}
		// let ref: &T? = &ints[0] as &T?   // read as (&T)?
		//
		// In this case the reference expression's type is an optional type.
		// Unwrap it one level to get the actual reference type
```

File: runtime/sema/check_reference_expression.go [hunk 3]
```diff
@@ -47,35 +54,24 @@ func (checker *Checker) VisitReferenceExpression(referenceExpression *ast.Refere
...
-			referencedType = optionalReferencedType.Type
-		}
+	_, referencedType = checker.visitExpression(referencedExpression, targetType)
 
-	} else {
-		// If the referenced expression is not an index expression, check it normally
-		_, referencedType = checker.visitExpression(referencedExpression, targetType)
+	// Unwrap the optional one level, but not infinitely
+	if optionalReferencedType, ok := referencedType.(*OptionalType); ok {
+		// if referenced type is optional, require that target type is also optional
```
turbolent, line 72:
```suggestion
		// If the referenced type is optional, then the return type is optional
```

runtime/sema/check_reference_expression.go
```diff
@@ -47,35 +54,24 @@ func (checker *Checker) VisitReferenceExpression(referenceExpression *ast.Refere
...
+		referencedType = optionalReferencedType.Type
+		returnType = &OptionalType{Type: referenceType}
 	}
 
 	if _, ok := referencedType.(*OptionalType); ok {
```
turbolent, line 77: This if-statement looks like dead code now: It is the same code as above (`referencedType.(*OptionalType)`), and cannot occur anymore, as `referencedType` is unwrapped and re-assigned above.

I think this statement can be removed, and the error type `OptionalTypeReferenceError` too.

runtime/tests/checker/reference_test.go
```diff
@@ -1012,10 +1009,45 @@ func TestCheckReferenceExpressionOfOptional(t *testing.T) {
...
+	t.Run("upcast to optional", func(t *testing.T) {
+
+		t.Parallel()
+
+		_, err := ParseAndCheck(t, `
+          let i: Int = 1
+          let ref = &i as &Int?
+        `)
+
+		require.NoError(t, err)
```
turbolent, line 1050: Maybe assert the type of `ref` to be an optional of a reference to `Int`

runtime/tests/checker/reference_test.go
```diff
@@ -1012,10 +1009,45 @@ func TestCheckReferenceExpressionOfOptional(t *testing.T) {
           let ref = &i as &Int?
         `)
 
-		errs := ExpectCheckerErrors(t, err, 2)
+		require.NoError(t, err)
```
turbolent, line 1012: Maybe also assert the type of `ref` to be an optional of a reference to `Int`

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -8020,25 +8020,23 @@ func TestInterpretNonStorageReferenceToOptional(t *testing.T) {
...
-              destroy xs
-              return name
-          }
+		  fun testNil(): String {
+			let xs: @{String: Foo} <- {}
+			let ref = (&xs["no"] as &Foo?)!
+			let name = ref.name
+			destroy xs
+			return name
+		  }
```
turbolent, line 8037: Please keep the spaces in string literals, i.e. revert the changes

turbolent (commented):
Nice work! 👏

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -9497,3 +9494,39 @@ func TestInterpretArrayTypeInference(t *testing.T) {
...
+              let x: Int? = 1
+			  let y = &x as &Int?
+			  return y!
+          }
+
+		  fun absent(): &Int {
+			let x: Int? = nil
+			let y = &x as &Int?
+			return y!
+		}
```
SupunS, line 9514:
```suggestion
          fun present(): &Int {
              let x: Int? = 1
              let y = &x as &Int?
              return y!
          }

          fun absent(): &Int {
              let x: Int? = nil
              let y = &x as &Int?
              return y!
          }
```

runtime/tests/checker/reference_test.go
```diff
@@ -1079,6 +1111,84 @@ func TestCheckInvalidDictionaryAccessReference(t *testing.T) {
...
+
+	_, err := ParseAndCheck(t, `
+		pub struct S {
+			pub let foo : Number
+			init() {
+				self.foo = 0
+			}
+		}
+		let dict: {String : S} = {}
+		let s = &dict[""] as &S
```
SupunS, line 1165: 👌

runtime/sema/check_reference_expression.go
```diff
@@ -34,9 +34,16 @@ func (checker *Checker) VisitReferenceExpression(referenceExpression *ast.Refere
...
+	var returnType Type
 
 	if !resultType.IsInvalidType() {
 		var ok bool
+		// indexed access to dictionaries require an optional reference as their
+		// target type, so allow one level of optionals
+		optType, optOk := resultType.(*OptionalType)
+		if optOk {
+			resultType = optType.Type
+		}
```
SupunS, line 46: maybe leave a blank line after the block
```suggestion
		}

```

SupunS (commented):
Once the suggestions are addressed, PR looks good to me!

runtime/interpreter/interpreter_expression.go
dsainati1: We do need to do this switch on the dynamic type so we know whether or not to move the `*SomeValue` constructor. However you are right in that this check should only occur if the static type of the expression is itself optional.

dsainati1:
The FLIP proposing this change has been approved and merged.

runtime/missingmember_test.go
```diff
@@ -836,15 +836,15 @@ pub contract GarmentNFT: NonFungibleToken {
...
        }
 
        // Parameters: id: The ID of the NFT to get the reference for
        //
        // Returns: A reference to the NFT
        pub fun borrowGarment(id: UInt64): &GarmentNFT.NFT? {
            if self.ownedNFTs[id] != nil {
-               let ref = &self.ownedNFTs[id] as auth &NonFungibleToken.NFT
+               let ref = (&self.ownedNFTs[id] as auth &NonFungibleToken.NFT?)!
                return ref as! &GarmentNFT.NFT
```
turbolent, line 848: Even though this makes the code valid, and the code itself is just a test case for something else and was provided by the community, we should maybe still update this code so it gracefully handles the case and to avoid developers seeing this and copying it as-is.

Concretely, in the cases where the functions return an optional (e.g. here, but not in the function above), maybe instead use the nil-coalescing operator `??` to return `nil` if the dictionary lookup fails, instead of aborting the execution:

```suggestion
               let ref = &self.ownedNFTs[id] as auth &NonFungibleToken.NFT?
               return ref ?? ref as! &GarmentNFT.NFT
```

runtime/interpreter/interpreter_expression.go
```diff
@@ -823,11 +823,39 @@ func (interpreter *Interpreter) VisitReferenceExpression(referenceExpression *as
 		interpreter.trackReferencedResourceKindedValue(result.StorageID(), result)
 	}
 
-	return &EphemeralReferenceValue{
-		Authorized:   borrowType.Authorized,
-		Value:        result,
-		BorrowedType: borrowType.Type,
+	switch typ := borrowType.(type) {
+	case *sema.OptionalType:
```
turbolent, line 827: 👍

runtime/tests/checker/reference_test.go
```diff
@@ -1079,6 +1131,84 @@ func TestCheckInvalidDictionaryAccessReference(t *testing.T) {
...
+	t.Parallel()
+
+	_, err := ParseAndCheck(t, `
+		pub struct S {
+			pub let foo : Number
+			init() {
+				self.foo = 0
+			}
+		}
+		let dict: {String : S} = {}
```
turbolent, line 1163:
```suggestion
			pub let foo: Number
			init() {
				self.foo = 0
			}
		}
		let dict: {String: S} = {}
```

runtime/tests/checker/reference_test.go
```diff
@@ -1079,6 +1131,84 @@ func TestCheckInvalidDictionaryAccessReference(t *testing.T) {
...
```
turbolent, line 1184:
```suggestion
			pub let foo: Number
			init() {
				self.foo = 0
			}
		}
		let dict: {String: S} = {}
```

runtime/tests/checker/reference_test.go
```diff
@@ -1079,6 +1131,84 @@ func TestCheckInvalidDictionaryAccessReference(t *testing.T) {
...
+	require.IsType(t, &sema.TypeMismatchError{}, errs[0])
+}
+
+func TestCheckArrayAccessReference(t *testing.T) {
+
+	t.Parallel()
+
+	_, err := ParseAndCheck(t, `
+		pub struct S {
+			pub let foo : Number
```
turbolent, line 1199:
```suggestion
			pub let foo: Number
```

turbolent (approved):
Looks good, great work! 👏 Just some minor style suggestions, good to merge after

PR #1315, opened 2021-12-16
Add array slice function

### LINKED ISSUE #1314: Array slice function
### Issue To Be Solved

Cadence currently does not provide a function to create a slice of an array, so creating a slice must be implemented in Cadence and is therefore inefficient.

### Suggested Solution

- Add a function `fun slice(from: Int, upTo: Int): [T]` to arrays, similar to how it exists for strings
- To implement such a function inside of Cadence efficiently:
  - Add an additional iteration function `func Iterate(startIndex, endIndex uint64, fn ArrayIterationFunc) error` to `atree.Array` which starts at a given index
  - Use `atree.NewArrayFromBatchData` to construct the new sub-array using the new iteration function

turbolent:
Closes #1314 

## Description

- Add a function `fun slice(from: Int, upTo: Int): [T]` to non-resource-kinded arrays (variable and constant-sized), similar to how it exists for strings
- Depends on latest atree feature, [array range iterators](https://github.com/onflow/atree/pull/238)
	
______


docs/language/values-and-types.mdx
```diff
@@ -1076,6 +1076,27 @@ are available for both variable-sized and fixed-sized or variable-sized arrays.
...
 
+- `cadence•fun slice(from: Int, upTo: Int): [T]`
+
+  Returns an array slice of the elements
+  in the given array from start index `from` up to,
+  but not including, the end index `upTo`.
+  This function creates a new array whose length is `upTo - from`.
+  It does not modify the original array.
+  If either of the parameters are out of
+  the bounds of the array, the function will fail.
```
dsainati1, line 1087: What if `upTo < from`? This should probably fail in some way, but we should specify the behavior. This case also appears unspecified in the `slice` function for strings as well.

runtime/interpreter/errors.go
```diff
@@ -349,6 +349,34 @@ func (e ArrayIndexOutOfBoundsError) Error() string {
...
+}
+
+func (e ArraySliceIndicesError) Error() string {
+	return fmt.Sprintf(
+		"slice indices [%d:%d] are out of bounds (size %d)",
+		e.FromIndex, e.UpToIndex, e.Size,
+	)
+}
+
+// InvalidSliceIndexError is returned when a slice index is invalid, such as fromIndex > upToIndex
```
dsainati1, line 368: This answers my question above about invalid slice ranges, but this should be included in the documentation.

runtime/interpreter/value.go
```diff
@@ -1136,6 +1143,14 @@ func (v *ArrayValue) handleIndexOutOfBoundsError(err error, index int, getLocati
...
+		panic(ArrayIndexOutOfBoundsError{
+			Index:         index,
+			Size:          v.Count(),
+			LocationRange: getLocationRange(),
+		})
+	}
+
 	storable, err := v.array.Get(uint64(index))
 	if err != nil {
 		v.handleIndexOutOfBoundsError(err, index, getLocationRange)
```
dsainati1, line 1156: isn't this `panic` redundant given the call to `handleIndexOutOfBoundsError` later?

runtime/interpreter/value.go
```diff
@@ -1152,6 +1167,13 @@ func (v *ArrayValue) SetKey(interpreter *Interpreter, getLocationRange func() Lo
 }
 
 func (v *ArrayValue) Set(interpreter *Interpreter, getLocationRange func() LocationRange, index int, element Value) {
+	if index < 0 {
+		panic(ArrayIndexOutOfBoundsError{
```
dsainati1, line 1171: same question as above

runtime/interpreter/value.go
```diff
@@ -1225,6 +1247,13 @@ func (v *ArrayValue) InsertKey(interpreter *Interpreter, getLocationRange func()
 }
 
 func (v *ArrayValue) Insert(interpreter *Interpreter, getLocationRange func() LocationRange, index int, element Value) {
+	if index < 0 {
+		panic(ArrayIndexOutOfBoundsError{
```
dsainati1, line 1251: same question as above

runtime/interpreter/value.go
```diff
@@ -1251,6 +1280,14 @@ func (v *ArrayValue) RemoveKey(interpreter *Interpreter, getLocationRange func()
 }
 
 func (v *ArrayValue) Remove(interpreter *Interpreter, getLocationRange func() LocationRange, index int) Value {
+	if index < 0 {
+		panic(ArrayIndexOutOfBoundsError{
```
dsainati1, line 1284: same question as above

docs/language/values-and-types.mdx
turbolent: Good question! Clarified in the docstrings and documentation in ce4487eb2e16b444667ea9d654407ff3de22b376

runtime/interpreter/errors.go
turbolent: Good idea to add it to the documentation! Added it there and to the docstrings in ce4487eb2e16b444667ea9d654407ff3de22b376

runtime/interpreter/value.go
turbolent: I thought so too, but turns out it is not.

Initially the Cadence functions here performed bounds checks. Given that atree was also performing bounds checks, we decided to remove the bounds checks from Cadence. 

However, the bounds checks in atree were not equivalent: The index type in Cadence is `int` (signed), whereas in atree, it is `uint64` (unsigned). Atree is only performing a bounds check if the index against the size of the array, but not if the index is negative. Negative indices were still handled, but indirectly: The conversion of a negative `int` to `uint64` results in a very large number, which is then out-of-bounds.

I cleaned this up by properly checking for negative integers before converting to `uint64`

SupunS (approved):

docs/language/values-and-types.mdx
```diff
@@ -1076,6 +1076,27 @@ are available for both variable-sized and fixed-sized or variable-sized arrays.
   let containsKitty = numbers.contains("Kitty")
   ```
 
+- `cadence•fun slice(from: Int, upTo: Int): [T]`
```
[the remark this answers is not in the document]
turbolent: The `String.slice` function already uses these same parameter names / argument labels, and yes, I intentionally chose the longer `upTo` to indicate that it is exclusive. This was inspired by Swift's naming scheme, e.g. https://developer.apple.com/documentation/swift/array/1688336-prefix

runtime/interpreter/value.go
```diff
@@ -1670,6 +1724,81 @@ func (v *ArrayValue) IsResourceKinded(interpreter *Interpreter) bool {
...
+func (v *ArrayValue) Slice(
+	interpreter *Interpreter,
+	from IntValue,
+	to IntValue,
+	getLocationRange func() LocationRange,
+) Value {
+	fromIndex := from.ToInt()
+	toIndex := to.ToInt()
+
+	if fromIndex < 0 || toIndex < 0 {
```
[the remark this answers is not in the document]
turbolent: We only need to check the lower bound before converting from `int` (signed) to `uint64` (unsigned), and `RangeIterator` will check the upper bound, i.e. this avoids checking twice.

I added a comment here and to the other cases in the code where the same applies in commit e91375870e9c4e12d306e2c304b3556dd9cff535

PR #1318, opened 2021-12-17
Add more graceful handling for go type assertions in `runtime/sema`

dsainati1:
## Description

This adds more graceful go handling for some type assertions in the `runtime/sema` directory, which previously were unhandled. This should not result in any behavioral differences. 

______

### DIFF
--- a/languageserver/server/server.go
+++ b/languageserver/server/server.go
@@ -753,7 +753,6 @@ func (s *Server) DocumentHighlight(
 
 	position := conversion.ProtocolToSemaPosition(params.Position)
 	occurrences := checker.Occurrences.FindAll(position)
-
 	// If there are no occurrences,
 	// then try the preceding position
 	if len(occurrences) == 0 && position.Column > 0 {
@@ -801,7 +800,6 @@ func (s *Server) Rename(
 
 	position := conversion.ProtocolToSemaPosition(params.Position)
 	occurrences := checker.Occurrences.FindAll(position)
-
 	// If there are no occurrences,
 	// then try the preceding position
 	if len(occurrences) == 0 && position.Column > 0 {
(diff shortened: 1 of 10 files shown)


languageserver/server/server.go
```diff
@@ -753,12 +753,18 @@ func (s *Server) DocumentHighlight(
 
 	position := conversion.ProtocolToSemaPosition(params.Position)
 	occurrences := checker.Occurrences.FindAll(position)
+	if occurrences == nil {
+		return nil, nil
+	}
```
SupunS, line 758: If the result is nil, then the below condition should get executed. i.e:`len()` of a nil array is also zero.

It is safe to skip the nil check here and below, as `len()` and `range` expressions are nil-safe.

runtime/sema/member_accesses.go
```diff
@@ -57,15 +57,22 @@ func (m *MemberAccesses) Find(pos Position) *MemberAccess {
...
 }
 
 func (m *MemberAccesses) All() []MemberAccess {
 	values := m.tree.Values()
 	accesses := make([]MemberAccess, len(values))
 	for i, value := range values {
-		accesses[i] = value.(MemberAccess)
+		access, ok := value.(MemberAccess)
+		if !ok {
+			return nil
```
SupunS, line 73: would `continue` be a better option? I'm ok with both.

SupunS (approved):
Nice! 👌

runtime/sema/member_accesses.go
dsainati1: I think this depends on what we'd like this function to do. If one of the elements of `values` is not a `MemberAccess` even though this function is meant to be called on a `MemberAccesses` collection, that feels like a significant enough issue that we'd like to fail the entire operation and return `nil`. However, if you feel like we should just recover as much as possible and return those values that are valid then we can `continue`.

SupunS: I felt returning `nil` would be somewhat closer to returning an empty array (the result of this is more likely to be looped over the `range`). Maybe if we want to completely fail the operation then panicking with a dedicated error would be even better I guess.
My initial thought was, if we are silently failing, we could make it to be resilient as much as possible.

PR #1319, opened 2021-12-18
Support recursive supertype inferring for containers

### LINKED ISSUE #1225: Support Recursive supertype inferring for containers
### Issue To Be Solved

Currently, common supertype inferring does not support covariance for containers. e.g: common supertype `[Int]` and `[String]` is inferred to be `AnyStruct`. We can improve this to recursively infer common supertype for elements as well, which should result `[AnyStruct]` instead of `AnyStruct`

### Suggested Solution

- [ ] Recursively infer common supertype for containers
- [ ] For dictionaries, this going to need a cadence type to represent all 'keyable types'.

SupunS:
Closes #1225

## Description

This PR add $subject for arrays and dictionaries.

______

### DIFF
--- a/runtime/sema/type_tags.go
+++ b/runtime/sema/type_tags.go
@@ -590,11 +590,11 @@ func findSuperTypeFromLowerMask(joinedTypeTag TypeTag, types []Type) Type {
 
 	// All derived types goes here.
 	case constantSizedTypeMask:
-		return getConstantSizedArraySuperType(types)
+		return commonSuperTypeOfConstantSizedArrays(types)
 	case variableSizedTypeMask:
-		return getVariableSizedSuperType(types)
+		return commonSuperTypeOfVariableSizedArrays(types)
 	case dictionaryTypeMask:
-		return getDictionarySuperType(types)
+		return commonSuperTypeOfDictionaries(types)
 	case referenceTypeMask,
 		genericTypeMask,
 		functionTypeMask,
@@ -653,10 +653,10 @@ func getSuperTypeOfDerivedTypes(types []Type) Type {
 	return prevType
 }
 
-func getVariableSizedSuperType(types []Type) Type {
+func commonSuperTypeOfVariableSizedArrays(types []Type) Type {
 	// We reach here if all types are variable-sized arrays.
-	// Therefore, check for member types, and decide the
-	// common supertype based on the element types.
+	// Therefore, decide the common supertype based on the element types.
+
 	elementTypes := make([]Type, 0)
 
 	for _, typ := range types {
@@ -668,7 +668,7 @@ func getVariableSizedSuperType(types []Type) Type {
 
 		arrayType, ok := typ.(*VariableSizedType)
 		if !ok {
-			panic(fmt.Errorf("expected variabled-sized array type, found %t", typ))
+			panic(fmt.Errorf("expected variabled-sized array type, found %s", typ))
 		}
 
 		elementTypes = append(elementTypes, arrayType.ElementType(false))
@@ -685,10 +685,10 @@ func getVariableSizedSuperType(types []Type) Type {
 	}
 }
 
-func getConstantSizedArraySuperType(types []Type) Type {
+func commonSuperTypeOfConstantSizedArrays(types []Type) Type {
 	// We reach here if all types are constant-sized arrays.
-	// Therefore, check for member types, and decide the
-	// common supertype based on the element types.
+	// Therefore, decide the common supertype based on the element types.
+
 	elementTypes := make([]Type, 0)
 	var prevType *ConstantSizedType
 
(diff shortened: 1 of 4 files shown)


runtime/sema/type_tags.go
```diff
@@ -650,6 +653,120 @@ func getSuperTypeOfDerivedTypes(types []Type) Type {
...
+	for _, typ := range types {
+		// 'Never' type doesn't affect the supertype.
+		// Hence, ignore them
+		if typ == NeverType {
+			continue
+		}
+
+		arrayType, ok := typ.(*ConstantSizedType)
+		if !ok {
+			panic(fmt.Errorf("expected constnant-sized array type, found %s", typ))
```
dsainati1, line 704:
```suggestion
			panic(fmt.Errorf("expected constant-sized array type, found %s", typ))
```

runtime/sema/type_tags.go
```diff
@@ -650,6 +653,120 @@ func getSuperTypeOfDerivedTypes(types []Type) Type {
...
+	for _, typ := range types {
+		// 'Never' type doesn't affect the supertype.
+		// Hence, ignore them
+		if typ == NeverType {
+			continue
+		}
+
+		arrayType, ok := typ.(*VariableSizedType)
+		if !ok {
+			panic(fmt.Errorf("expected variabled-sized array type, found %s", typ))
```
dsainati1, line 671:
```suggestion
			panic(fmt.Errorf("expected variable-sized array type, found %s", typ))
```

runtime/sema/type_tags.go
```diff
@@ -650,6 +653,120 @@ func getSuperTypeOfDerivedTypes(types []Type) Type {
...
+		valueTypes = append(valueTypes, dictionaryType.ValueType)
+		keyTypes = append(keyTypes, dictionaryType.KeyType)
+	}
+
+	keySuperType := LeastCommonSuperType(keyTypes...)
+	valueSuperType := LeastCommonSuperType(valueTypes...)
+
+	if keySuperType == InvalidType ||
+		valueSuperType == InvalidType ||
+		!IsValidDictionaryKeyType(keySuperType) {
```
dsainati1, line 760: I'm not sure we want to produce an invalid type when the dictionary key is invalid. Consider a case where we have two dictionaries: `{String: Int}` and `{Int: String}`. In this case, both `keySuperType` and `valueSuperType` are going to be `AnyStruct`, but I don't think this means the inferred type should itself be invalid; we probably want to use the appropriate `AnyStruct` or `AnyResource` in this case.

dsainati1 (commented):
This is exciting! Thanks for following up on this!

runtime/sema/type_tags.go
SupunS: Currently, the checker doesn't allow having heterogeneous key types, when type inferring is not required (i.e: type annotations are present). So the type inferring also follows the same checker rules. e.g: Handling this case in particular: https://github.com/onflow/cadence/blob/051bf1599729d43a1e4af2d054654371739ab5a2/runtime/tests/checker/type_inference_test.go#L1136-L1146

Here, by returning an invalid type, the type-inferring says "I can't figure out a type", and then the checker says "let's ask from user".

The reason why inferring a generic type like `AnyStruct` doesn't work is because the runtime-value also requires a specific static-type/dynamic-type for runtime type checking. That runtime type is created based on the type that the checker computes.

dsainati1: I see! That makes sense.

SupunS: Or are you suggesting keeping the old behavior for such cases?

dsainati1: I was assuming that the common supertype of `{String: Int}` and `{Int: Int}` would be `AnyStruct`

SupunS: Right, so that pretty much what we have now. I think that's a good idea to fall back to the broader type. That makes this change also backward compatible 👍

SupunS: added in https://github.com/onflow/cadence/pull/1319/commits/09128222d8724a2db70412724912cb18d845ffd7

runtime/sema/type_tags.go
```diff
@@ -650,6 +653,123 @@ func getSuperTypeOfDerivedTypes(types []Type) Type {
...
+
+	keySuperType := LeastCommonSuperType(keyTypes...)
+	valueSuperType := LeastCommonSuperType(valueTypes...)
+
+	if keySuperType == InvalidType ||
+		valueSuperType == InvalidType {
+		return InvalidType
+	}
+
+	if !IsValidDictionaryKeyType(keySuperType) {
```
turbolent, line 763: 👍

turbolent (approved):
Very nice! 👏

PR #1332, opened 2022-01-05
Fix resource loss analysis and definite initialization analysis

### LINKED ISSUE #1325: Resource object may lost when using switch-case to destroy
### Problem
Hi guys,
We are from the Increment Labs which focusing on the flow development and ecosystem. We know cadence takes resouce security testing seriously. However, we found that:
> If a switch-case without default is used to destroy the resource object, the compiler fails to make a judgement on whether the resource is passed correctly. When passing a param which is not captured by the switch-case, the resource disappears.

### Steps to Reproduce

Testcase on playground:
https://play.onflow.org/813992f4-5e98-4449-8421-9a753b607d5f?type=account&id=0
```js
    // deploy successfully
    pub fun test(n:Int, res:@R) {
        // switch with no default
        switch n {
            case 1:
                log("1")
                destroy res
            case 2:
                log("2")
                destroy res
        }
    }
```
```js
    let r <- HelloWorld.mintR()
    // if pass n = 3, switch failed in catching and resource disappears
    HelloWorld.test(n: 3, res: <-r)
```

turbolent:
Closes #1325 

Based heavily on #1329, thank you @SupunS 🙏 

Investigating the issue I also discovered that the definite initialization analysis suffered from the same problem.

## Description

- Fix and simplify resource tracking / resource loss detection (82c260aab267981de61551782775294a2af569d2):
  -  Consider jumps (break and continue statements) when merging then/else branches' invalidations into the parent
  - Don't record resource invalidations if the code adding the invalidation is unreachable
- Consider jumps in definite initialization analysis and improve dead code reporting (88b9fbe6f850a5eb0a483961cdf53fc472776088)
  - In addition to when maybe a return statement occured, initializations should also be ignored when maybe a jump (break or continue statement) occured
  - In addition to when a definite return statement occurs, code is also unreachable if a definite jump (beak or continue statement) occurred before
- Add tests for resource loss analysis and definite initialization analysis fixes

______

### DIFF
--- a/runtime/sema/return_info.go
+++ b/runtime/sema/return_info.go
@@ -21,6 +21,7 @@ package sema
 type ReturnInfo struct {
 	MaybeReturned      bool
 	DefinitelyReturned bool
+	MaybeJumped        bool
 	DefinitelyHalted   bool
 	DefinitelyJumped   bool
 }
@@ -33,6 +34,10 @@ func (ri *ReturnInfo) MergeBranches(thenReturnInfo *ReturnInfo, elseReturnInfo *
 	ri.DefinitelyReturned = ri.DefinitelyReturned ||
 		(thenReturnInfo.DefinitelyReturned &&
 			elseReturnInfo.DefinitelyReturned)
+
+	ri.MaybeJumped = ri.MaybeJumped ||
+		thenReturnInfo.MaybeJumped ||
+		elseReturnInfo.MaybeJumped
 
 	ri.DefinitelyJumped = ri.DefinitelyJumped ||
 		(thenReturnInfo.DefinitelyJumped &&
--- a/runtime/sema/return_info.go
+++ b/runtime/sema/return_info.go
@@ -22,6 +22,7 @@ type ReturnInfo struct {
 	MaybeReturned      bool
 	DefinitelyReturned bool
 	DefinitelyHalted   bool
+	DefinitelyJumped   bool
 }
 
 func (ri *ReturnInfo) MergeBranches(thenReturnInfo *ReturnInfo, elseReturnInfo *ReturnInfo) {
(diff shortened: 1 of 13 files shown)


runtime/sema/return_info.go
```diff
@@ -21,7 +21,9 @@ package sema
 type ReturnInfo struct {
 	MaybeReturned      bool
 	DefinitelyReturned bool
+	MaybeJumped        bool
```
SupunS, line 24: Is it now possible to merge `xxxReturned` and `xxxJumped` to a single flag?
Are there any use-cases of each being used individually? 🤔

runtime/tests/checker/resources_test.go
```diff
@@ -2698,10 +2733,11 @@ func TestCheckInvalidResourceLossThroughReturn(t *testing.T) {
       }
     `)
 
-	errs := ExpectCheckerErrors(t, err, 2)
+	errs := ExpectCheckerErrors(t, err, 3)
 
 	assert.IsType(t, &sema.ResourceLossError{}, errs[0])
 	assert.IsType(t, &sema.UnreachableStatementError{}, errs[1])
+	assert.IsType(t, &sema.ResourceLossError{}, errs[2])
```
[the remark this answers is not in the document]
turbolent: Yes, there is now an additional error reported, but "over reporting" isn't a problem: The first resource loss error is reported by the resource loss check by the return statement, and the second is returned by the resource loss check at the end of the function. Both are correct and I don't think it is worth the effort to reduce over reporting

runtime/sema/return_info.go
turbolent: Maybe, I'll have a look 👍

runtime/tests/checker/resources_test.go
```diff
@@ -5049,3 +5086,384 @@ func TestCheckEmptyResourceCollectionMove(t *testing.T) {
...
+                    default:
+                        destroy r
+                }
+            }
+        `)
+
+		require.NoError(t, err)
+	})
+
+	t.Run("switch-case: break in one case", func(t *testing.T) {
```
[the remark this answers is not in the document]
turbolent: Good catch!

```suggestion
	t.Run("switch-case: destroy missing in default case", func(t *testing.T) {
```

runtime/sema/return_info.go
turbolent: Merged `MaybeReturned` and `MaybeJumped` in 3021117. `DefinitelyReturned` is used in `Checker.checkFunctionExits` independently to determine if a function with a return type also has a return statement, so `DefinitelyReturned` and `DefinitelyJumped` cannot be merged

SupunS: I see. Thank you for checking!

SupunS (approved):
Nice! LGTM! 👌

PR #1344, opened 2022-01-06
Fix references to resource-kinded values

### LINKED ISSUE #1320: Ownership of nested resources is inconsistent, breaks backwards compatibility
### Problem

Flow seems to handle the `.owner` of nested resources inconsistently, and appears to have made a breaking change in `.owner` behaviour in the most recent (Dec 8, 2021) spork.

[One would expect](https://docs.onflow.org/cadence/language/composite-types/#resource-owner) that the `.owner` of a nested resource should always be pointing to the latest account which it had been stored in, and always be the same as its nesting resource. However, Flow inconsistently exhibits unintuitive behaviour.

Before the Dec 8 spork (emulator version 0.28.1), the `.owner` of a nested resource is consistently the same as its nesting resource. However, if a resource is stored in an array, which is in an account's storage, the resource's `.owner` will not be recorded and remain `nil`.

After the Dec 8 spork (emulator version 0.30.1),  the `.owner` of a nested resource is not recorded and remains `nil` even when its nesting resource's `.owner` changes.

### Steps to Reproduce

This can be verified with a test contract and transaction, as follows:
```
// TestContract.cdc
pub contract TestContract {
        pub resource TestResource {}
        pub resource TestNestingResource {
                pub let testResources : @[TestResource]

                init () {
                        self .testResources <- [ <- create TestResource () ] }
                destroy () {
                        destroy self .testResources } }
        pub fun makeTestResource () : @TestResource {
                return <- create TestResource () }
        pub fun makeTestNestingResource () : @TestNestingResource {
                return <- create TestNestingResource () }

        pub event Log (info : String)
        pub fun log (_ info : String) : Void {
                emit Log (info: info) } }
```

```
// test-transaction.cdc
import TestContract from 0xTESTCONTRACT

transaction () {

        prepare (account : AuthAccount) {

                var nestingResource <- TestContract .makeTestNestingResource ()
                var nestingResourceRef = & nestingResource as &TestContract.TestNestingResource
                var nestedElementResourceRef = & (nestingResource .testResources [0]) as &TestContract.TestResource

                // At this point the nesting and nested resources are not in storage
                TestContract .log ("nesting resource owner: " .concat (nestingResourceRef .owner ?.address ?.toString () ?? "nil"))
                TestContract .log ("nested resource owner: " .concat (nestedElementResourceRef .owner ?.address ?.toString () ?? "nil"))

                account .save (<- nestingResource, to: /storage/test)

                // At this point the nesting and nested resources are both in storage
                TestContract .log ("nesting resource owner: " .concat (nestingResourceRef .owner ?.address ?.toString () ?? "nil"))
                TestContract .log ("nested resource owner: " .concat (nestedElementResourceRef .owner ?.address ?.toString () ?? "nil"))

                var nestingResource2 <- account .load <@TestContract.TestNestingResource> (from: /storage/test) !

                // At this point the nesting and nested resources are both taken out of storage
                TestContract .log ("nesting resource owner: " .concat (nestingResourceRef .owner ?.address ?.toString () ?? "nil"))
                TestContract .log ("nested resource owner: " .concat (nestedElementResourceRef .owner ?.address ?.toString () ?? "nil"))

                var removedResource <- nestingResource2 .testResources .remove (at: 0)

                // At this point the nesting and nested resources are both still out of storage, and the nested resource is removed from the nesting resource
                TestContract .log ("nesting resource owner: " .concat (nestingResourceRef .owner ?.address ?.toString () ?? "nil"))
                TestContract .log ("nested resource owner: " .concat (nestedElementResourceRef .owner ?.address ?.toString () ?? "nil"))

                destroy removedResource
                destroy nestingResource2 !

                var resourceArray <- [ <- TestContract .makeTestResource () ]
                var elementResourceRef = & (resourceArray [0]) as &TestContract.TestResource

                // At this point the resource array and the nested resource are not in storage
                TestContract .log ("element resource owner: " .concat (elementResourceRef .owner ?.address ?.toString () ?? "nil"))

                account .save (<- resourceArray, to: /storage/test)

                // At this point the resource array and the nested resource are both in storage
                TestContract .log ("element resource owner: " .concat (elementResourceRef .owner ?.address ?.toString () ?? "nil"))

                var resourceArray2 <- account .load <@[TestContract.TestResource]> (from: /storage/test) !

                // At this point the resource array and the nested resource are both taken out of storage
                TestContract .log ("element resource owner: " .concat (elementResourceRef .owner ?.address ?.toString () ?? "nil"))

                var removedResource2 <- resourceArray2 .remove (at: 0)

                // At this point the resource array and the nested resource are both still out of storage, and the nested resource is removed from the resource array
                TestContract .log ("element resource owner: " .concat (elementResourceRef .owner ?.address ?.toString () ?? "nil"))

                destroy removedResource2
                destroy resourceArray2 } }
```

Before the Dec 8 spork (emulator version 0.28.1), this script gives the result by inspecting events:
```
nesting resource owner: nil
nested resource owner: nil
nesting resource owner: 0xf8d6e0586b0a20c7
nested resource owner: 0xf8d6e0586b0a20c7
nesting resource owner: 0xf8d6e0586b0a20c7
nested resource owner: 0xf8d6e0586b0a20c7
nesting resource owner: 0xf8d6e0586b0a20c7
nested resource owner: 0xf8d6e0586b0a20c7
element resource owner: nil
element resource owner: 0xf8d6e0586b0a20c7
element resource owner: nil
element resource owner: nil
```

After the Dec 8 spork (emulator version 0.30.1), this script gives the result by inspecting events:
```
nesting resource owner: nil
nested resource owner: nil
nesting resource owner: 0xf8d6e0586b0a20c7
nested resource owner: nil
nesting resource owner: 0xf8d6e0586b0a20c7
nested resource owner: nil
nesting resource owner: 0xf8d6e0586b0a20c7
nested resource owner: nil
element resource owner: nil
element resource owner: nil
element resource owner: nil
element resource owner: nil
```

turbolent:
Closes #1320 

## Description

Track referenced resource-kinded values and update them when another value representing the same value is transferred.

_____

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -279,6 +279,8 @@ type Storage interface {
 	CheckHealth() error
 }
 
+type ReferencedValues map[atree.StorageID]map[ReferenceTrackedValue]struct{}
+
 type Interpreter struct {
 	Program                        *Program
 	Location                       common.Location
@@ -315,6 +317,8 @@ type Interpreter struct {
 	atreeValueValidationEnabled    bool
 	atreeStorageValidationEnabled  bool
 	tracingEnabled                 bool
+	// TODO: ideally this would be a weak map, but Go has no weak references
+	referencedValues ReferencedValues
 }
 
 type Option func(*Interpreter) error
@@ -577,6 +581,15 @@ func withTypeCodes(typeCodes TypeCodes) Option {
 	}
 }
 
+// withReferencedValues returns an interpreter option which sets the referenced values.
+//
+func withReferencedValues(referencedValues ReferencedValues) Option {
+	return func(interpreter *Interpreter) error {
+		interpreter.referencedValues = referencedValues
+		return nil
+	}
+}
+
 // WithDebugger returns an interpreter option which sets the given debugger
 //
 func WithDebugger(debugger *Debugger) Option {
@@ -615,6 +628,7 @@ func NewInterpreter(program *Program, location common.Location, options ...Optio
 			InterfaceCodes:       map[sema.TypeID]WrapperCode{},
 			TypeRequirementCodes: map[sema.TypeID]WrapperCode{},
 		}),
+		withReferencedValues(map[atree.StorageID]map[ReferenceTrackedValue]struct{}{}),
 	}
 
 	for _, option := range defaultOptions {
@@ -1230,7 +1244,7 @@ func (interpreter *Interpreter) visitFunctionBody(
 	}
 
 	// If there is a return type, declare the constant `result`.
-	// If it is a resource type, the constant has the same type as a referecne to the return type.
+	// If it is a resource type, the constant has the same type as a reference to the return type.
 	// If it is not a resource type, the constant has the same type as the return type.
 
 	if returnType != sema.VoidType {
@@ -2506,6 +2520,7 @@ func (interpreter *Interpreter) NewSubInterpreter(
 		WithAtreeValueValidationEnabled(interpreter.atreeValueValidationEnabled),
 		WithAtreeStorageValidationEnabled(interpreter.atreeStorageValidationEnabled),
 		withTypeCodes(interpreter.typeCodes),
+		withReferencedValues(interpreter.referencedValues),
 		WithPublicAccountHandler(interpreter.publicAccountHandler),
 		WithPublicKeyValidationHandler(interpreter.PublicKeyValidationHandler),
 		WithSignatureVerificationHandler(interpreter.SignatureVerificationHandler),
@@ -4354,5 +4369,29 @@ func (interpreter *Interpreter) ValidateAtreeValue(v atree.Value) {
 			}
 		}
 	}
+}
 
+func (interpreter *Interpreter) trackReferencedValue(id atree.StorageID, value ReferenceTrackedValue) {
+	values := interpreter.referencedValues[id]
+	if values == nil {
+		values = map[ReferenceTrackedValue]struct{}{}
+		interpreter.referencedValues[id] = values
+	}
+	values[value] = struct{}{}
+}
+
+func (interpreter *Interpreter) updateReferencedValues(
+	currentStorageID atree.StorageID,
+	newStorageID atree.StorageID,
+	updateFunc func(value Value),
+) {
+	values := interpreter.referencedValues[currentStorageID]
+	if values == nil {
+		return
+	}
+	for value := range values {
+		updateFunc(value)
+	}
+	interpreter.referencedValues[newStorageID] = values
+	interpreter.referencedValues[currentStorageID] = nil
 }
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -279,7 +279,7 @@ type Storage interface {
 	CheckHealth() error
 }
 
-type ReferencedValues map[atree.StorageID]map[ReferenceTrackedValue]struct{}
+type ReferencedResourceKindedValues map[atree.StorageID]map[ReferenceTrackedResourceKindedValue]struct{}
 
 type Interpreter struct {
 	Program                        *Program
@@ -318,7 +318,7 @@ type Interpreter struct {
 	atreeStorageValidationEnabled  bool
 	tracingEnabled                 bool
 	// TODO: ideally this would be a weak map, but Go has no weak references
-	referencedValues ReferencedValues
+	referencedResourceKindedValues ReferencedResourceKindedValues
 }
 
 type Option func(*Interpreter) error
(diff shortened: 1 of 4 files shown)


runtime/interpreter/value.go
```diff
@@ -1592,7 +1607,21 @@ func (v *ArrayValue) Transfer(
...
+		// (but currently point to an outdated Go instance of the value)
+
 		v.array = array
+
+		newStorageID := array.StorageID()
+		interpreter.updateReferencedResource(
+			currentStorageID,
+			newStorageID,
+			func(value ReferenceTrackedResourceKindedValue) {
+				value.(*ArrayValue).array = array
```
dsainati1, line 1621: Does this count as an unchecked cast? This wouldn't be picked up by the linter, but we are implicitly ignoring the boolean result here. Unsure whether or not it's worth doing the cast separately.

runtime/interpreter/value.go
```diff
@@ -11740,7 +11773,22 @@ func (v *CompositeValue) Transfer(
...
+
 		v.dictionary = dictionary
+
+		newStorageID := dictionary.StorageID()
+
+		interpreter.updateReferencedResource(
+			currentStorageID,
+			newStorageID,
+			func(value ReferenceTrackedResourceKindedValue) {
+				value.(*CompositeValue).dictionary = dictionary
```
dsainati1, line 11788: same question as above

runtime/interpreter/value.go
```diff
@@ -12651,7 +12705,22 @@ func (v *DictionaryValue) Transfer(
...
+
 		v.dictionary = dictionary
+
+		newStorageID := dictionary.StorageID()
+
+		interpreter.updateReferencedResource(
+			currentStorageID,
+			newStorageID,
+			func(value ReferenceTrackedResourceKindedValue) {
+				value.(*DictionaryValue).dictionary = dictionary
```
dsainati1, line 12720: same question as above

runtime/storage_test.go
```diff
@@ -2268,3 +2268,517 @@ transaction {
 	)
 	require.NoError(t, err)
 }
+
+func TestRuntimeReferenceOwnerAccess(t *testing.T) {
```
SupunS, line 2272: Just thinking out loud here: What would be the (expected) behavior if:
- A resource value `x` is created
- A reference `xRef` is created for `x`
- `x` is stored in account `A`
- load the value in account `A`; say `y`
- Take a reference `yRef` to `y`
- Store `y` in a different account `B`
-  What would be the owners of `xRef` and `yRef` respectively 🤔 

Not sure, maybe worth adding a test-case?

runtime/interpreter/value.go
turbolent: Yes, it definitely does! Good point that we shouldn't introduce new type unchecked assertions like this one and the ones below 👍

turbolent: Fixed in 7295252fe948a5fa81f01fdd8c574234493a67a4

runtime/interpreter/value.go
turbolent: Fixed in 7295252fe948a5fa81f01fdd8c574234493a67a4

runtime/interpreter/value.go
turbolent: Fixed in 7295252fe948a5fa81f01fdd8c574234493a67a4

runtime/interpreter/interpreter.go
```diff
@@ -315,6 +317,8 @@ type Interpreter struct {
 	atreeValueValidationEnabled    bool
 	atreeStorageValidationEnabled  bool
 	tracingEnabled                 bool
+	// TODO: ideally this would be a weak map, but Go has no weak references
+	referencedResourceKindedValues ReferencedResourceKindedValues
```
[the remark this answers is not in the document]
turbolent: Yes, this mapping would fill up if the interpreter is used for a prolonged period of time, e.g. for multiple transactions, scripts, etc. AFAIK this is not the case at the moment. @janezpodhostnik Can you confirm this?

The problem this PR addresses is that there exist multiple values that represent the same resource. For example, when a resource is stored in a container, e.g. an array, then the Go value representing the `CompositeValue` is converted to a storable representation (`Storable`) and the value is discarded. When the array element is accessed, the storable is converted back to a value. If the value that was inserted had a reference, it now points to a difference `CompositeValue` instance then the one that was created from the indexing access. That means it is not possible to keep the references in the value itself.

runtime/storage_test.go
turbolent: After this operation, both references point to the same resource, which is stored in account B, so they should both report the owner to be B.

I wrote a test for this, but it is not passing, I'm looking into it.

Thank you for coming up with another test case!

turbolent: The tracking had a bug where it would loose the tracked resources when the source and target ID where the same, which was only exercised through your test case (!). Thank you again for encouraging the "basic" case (resource is not in a container or nested), but also testing the behaviour over multiple operations 👌 

Fixed in 1dc3d579a7b14886fe370d13c9d50d29530ac0f2

SupunS: Nice! Thank you for adding the test and for the fix! 👍

SupunS (approved):
LGTM! 🎉

PR #1348, opened 2022-01-10
Fix parsing of 0_ integer literals

turbolent:
Closes dapperlabs/cadence-private-issues#14
______

### DIFF
--- a/runtime/parser2/lexer/state.go
+++ b/runtime/parser2/lexer/state.go
@@ -210,7 +210,7 @@ func numberState(l *lexer) stateFn {
 			}
 			l.emitValue(TokenHexadecimalIntegerLiteral)
 
-		case '0', '1', '2', '3', '4', '5', '6', '7', '8', '9':
+		case '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '_':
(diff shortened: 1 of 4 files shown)


runtime/parser2/lexer/state.go
```diff
@@ -210,7 +210,7 @@ func numberState(l *lexer) stateFn {
 			}
 			l.emitValue(TokenHexadecimalIntegerLiteral)
 
-		case '0', '1', '2', '3', '4', '5', '6', '7', '8', '9':
+		case '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '_':
```
SupunS, line 213: would this allow numbers starting with `_`? e.g: `_0`?

turbolent: No, this case only may occur after an initial `0`, see https://github.com/onflow/cadence/pull/1348/files#diff-e258ba2fee01ca54837c0062a323e801197f430aeff7ce6185b0130f51b511beR189 above

turbolent: Added in 71d7af0d0707dd41c745b2ba8c91c410b231626d

SupunS: Thank you!

PR #1352, opened 2022-01-11
trying to add indexOf to array

### LINKED ISSUE #1276: Array indexOf implementation
### Issue To Be Solved
Arrays in Cadence have a limited API, but they do feature the function **contains**.
https://docs.onflow.org/cadence/language/values-and-types/#array-fields-and-functions

While this function is useful, it leaves something to be desired from the caller's perspective. Namely: **where is it?**
To answer this valuable question, the caller will then have to traverse the array themselves and find the thing they **know** is in there (_and just got found!_ <- this is repeated work, right?)

### Suggested Solution
The act of determining **does this array contain that thing**, is nothing more than a return-type change away from **where is this thing in that array**?

Arguably this function is a superset of both implementations:
```
indexOf(thing: T): UInt64? {
  if (!contains(thing)) {
    // find its index and return it
  }
  // return nil // "contains = false"
}
```

bjartek:
Closes #1276

## Description

If you want to remove an element from a cadence array right now you have to use loops in cadence. This is inneficient. 
This PR implements the `indexOf` method that in combination with remove can do that. 

______


docs/language/values-and-types.mdx
```diff
@@ -1079,6 +1079,22 @@ are available for both variable-sized and fixed-sized or variable-sized arrays.
   let containsKitty = numbers.contains("Kitty")
   ```
 
+- `cadence•fun indexOf(_ element: T): Int?`
```
turbolent, line 1082: I'd recommend renaming this function to `firstIndex(of: T): Int?`, [like in Swift](https://developer.apple.com/documentation/swift/array/2994720-firstindex)

docs/language/values-and-types.mdx
```diff
@@ -1079,6 +1079,22 @@ are available for both variable-sized and fixed-sized or variable-sized arrays.
...
+  // Declare an array of integers.
+  let numbers = [42, 23, 31, 12]
+
+  //Check if the array contains 31
+  let index= numbers.indexOf(31)
+  // `index` is 2
+
+  //Check if the array contains 22
+  let index= numbers.indexOf(31)
+  // `index` is nil
```
turbolent, line 1095:
```suggestion
  let index = numbers.indexOf(31)
  // `index` is 2

  //Check if the array contains 22
  let index = numbers.indexOf(31)
  // `index` is nil
```

runtime/interpreter/value.go
```diff
@@ -1348,6 +1348,33 @@ func (v *ArrayValue) RemoveLast(interpreter *Interpreter, getLocationRange func(
 	return v.Remove(interpreter, getLocationRange, v.Count()-1)
 }
 
+func (v *ArrayValue) IndexOf(interpreter *Interpreter, getLocationRange func() LocationRange, needleValue Value) OptionalValue {
+
+	needleEquatable, ok := needleValue.(EquatableValue)
+	if !ok {
+		panic(errors.NewUnreachableError())
+	}
```
turbolent, line 1356: At run-time the value is required to be equatable, so this must also be checked statically. See the comment for the implementation in `sema`

runtime/tests/checker/arrays_dictionaries_test.go
```diff
@@ -753,6 +753,19 @@ func TestCheckInvalidArrayRemoveLastFromConstantSized(t *testing.T) {
...
+	_, err := ParseAndCheck(t, `
+      fun test(): Int? {
+          let x = [1, 2, 3]
+          return x.indexOf(2)
+      }
+    `)
+
+	require.NoError(t, err)
+}
 func TestCheckArrayContains(t *testing.T) {
```
turbolent, line 769:
```suggestion
}

func TestCheckArrayContains(t *testing.T) {
```

runtime/tests/checker/arrays_dictionaries_test.go
```diff
@@ -753,6 +753,19 @@ func TestCheckInvalidArrayRemoveLastFromConstantSized(t *testing.T) {
 	assert.IsType(t, &sema.NotDeclaredMemberError{}, errs[0])
 }
 
+func TestCheckArrayIndexOf(t *testing.T) {
```
turbolent, line 756: Please also add some tests for invalid cases, like unequatable values, and resource-kinded values. See the test cases above

turbolent (commented):
Nice work, looks good! Maybe rename, and please also add interpreter tests

docs/language/values-and-types.mdx
bjartek: good point, will do.

runtime/tests/checker/arrays_dictionaries_test.go
bjartek: ok, will do.

runtime/interpreter/value.go
bjartek: I copied this code block from Contains, since it was very similiar and it is not implemented there. Is that wrong then?

runtime/tests/checker/arrays_dictionaries_test.go
bjartek: I have tried adding more tests here, are there any in particular you feel are missing now.

runtime/interpreter/value.go
turbolent: Right, I missed that 👍

runtime/interpreter/value.go
```diff
@@ -1348,6 +1348,33 @@ func (v *ArrayValue) RemoveLast(interpreter *Interpreter, getLocationRange func(
...
+		counter++
+		// continue iteration
+		return true
+	})
+
+	if result {
+		value := NewIntValueFromInt64(int64(counter))
+		return NewSomeValueNonCopying(value)
+	}
+	return NilValue{}
```
turbolent, line 1375:
```suggestion
	var counter int64
	var result bool
	v.Iterate(func(element Value) (resume bool) {
		if needleEquatable.Equal(interpreter, getLocationRange, element) {
			result = true
			// stop iteration
			return false
		}
		counter++
		// continue iteration
		return true
	})

	if result {
		value := NewIntValueFromInt64(counter)
		return NewSomeValueNonCopying(value)
	}
	return NilValue{}
```

docs/language/values-and-types.mdx
```diff
@@ -1079,6 +1079,23 @@ are available for both variable-sized and fixed-sized or variable-sized arrays.
...
+  let numbers = [42, 23, 31, 12]
+
+  //Check if the array contains 31
+
+  let index= numbers.firstIndex(of:31)
+  // `index` is 2
+
+  //Check if the array contains 22
+  let index= numbers.firstIndex(of:22)
+  // `index` is nil
```
turbolent, line 1096:
```suggestion
  // Check if the array contains 31
  let index = numbers.firstIndex(of: 31)
  // `index` is 2

  // Check if the array contains 22
  let index = numbers.firstIndex(of: 22)
  // `index` is nil
```

runtime/tests/checker/arrays_dictionaries_test.go
```diff
@@ -753,6 +753,72 @@ func TestCheckInvalidArrayRemoveLastFromConstantSized(t *testing.T) {
...
+
+	t.Parallel()
+
+	_, err := ParseAndCheck(t, `
+      fun test(): Int? {
+          let x = [1, 2, 3]
+					return x.firstIndex(of:2)
```
turbolent, line 763:
```suggestion
          return x.firstIndex(of: 2)
```

runtime/tests/checker/arrays_dictionaries_test.go
```diff
@@ -753,6 +753,72 @@ func TestCheckInvalidArrayRemoveLastFromConstantSized(t *testing.T) {
...
+	_, err := ParseAndCheck(t, `
+      fun test(): Int? {
+          let x = [1, 2, 3]
+					return x.firstIndex(of:5)
+      }
+    `)
+
+	require.NoError(t, err)
+}
+
```
turbolent, line 783: This test is a duplicate of the one above, the argument passed to the function does not matter for the static analysis.

```suggestion
```

runtime/tests/checker/arrays_dictionaries_test.go
```diff
@@ -753,6 +753,72 @@ func TestCheckInvalidArrayRemoveLastFromConstantSized(t *testing.T) {
...
+func TestCheckArrayFirstIndexWrongType(t *testing.T) {
+
+	t.Parallel()
+
+	_, err := ParseAndCheck(t, `
+      fun test(): Int? {
+          let x = [1, 2, 3]
+					return x.firstIndex(of:"foo")
```
turbolent, line 791:
```suggestion
          return x.firstIndex(of: "foo")
```

runtime/tests/checker/arrays_dictionaries_test.go
```diff
@@ -753,6 +753,72 @@ func TestCheckInvalidArrayRemoveLastFromConstantSized(t *testing.T) {
...
+func TestCheckInvalidResourceFirstIndex(t *testing.T) {
+
+	t.Parallel()
+
+	_, err := ParseAndCheck(t, `
+      resource X {}
+
+      fun test(): Int? {
+          let xs <- [<-create X()]
+					return <-xs.firstIndex(of: "foo")
```
turbolent, line 807: The return value is always `Int?`, what matter is the array element type, so maybe fix some of the errors like the invalid move operation for the returned value, and also make the argument a resource.

```suggestion
          return xs.firstIndex(of: <-create X())
```

This change suggestion also requires updating the error assertions below.

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -9615,3 +9615,51 @@ func TestInterpretArrayTypeInference(t *testing.T) {
...
+func TestInterpretArrayFirstIndex(t *testing.T) {
+
+	t.Parallel()
+
+	inter := parseCheckAndInterpret(t, `
+      let xs = [1, 2, 3]
+
+			fun test() :Int? {
+				return xs.firstIndex(of:2)
+      }
```
turbolent, line 9628:
```suggestion
      fun test(): Int? {
          return xs.firstIndex(of: 2)
      }
```

runtime/tests/checker/arrays_dictionaries_test.go
```diff
@@ -753,6 +753,72 @@ func TestCheckInvalidArrayRemoveLastFromConstantSized(t *testing.T) {
...
+
+	assert.IsType(t, &sema.InvalidResourceArrayMemberError{}, errs[0])
+	assert.IsType(t, &sema.NotEquatableTypeError{}, errs[1])
+	assert.IsType(t, &sema.TypeMismatchError{}, errs[2])
+	assert.IsType(t, &sema.MissingMoveOperationError{}, errs[3])
+	assert.IsType(t, &sema.InvalidMoveOperationError{}, errs[4])
+	assert.IsType(t, &sema.ResourceLossError{}, errs[5])
+	assert.IsType(t, &sema.ResourceLossError{}, errs[6])
+}
+
```
turbolent, line 821: Could you please also add a checker test for a non-equatable type here? For example, arrays are not equatable, so testing an array of arrays would be suitable

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -9615,3 +9615,51 @@ func TestInterpretArrayTypeInference(t *testing.T) {
...
+
+func TestInterpretArrayFirstIndexDoesNotExist(t *testing.T) {
+
+	t.Parallel()
+
+	inter := parseCheckAndInterpret(t, `
+      let xs = [1, 2, 3]
+
+			fun test() :Int? {
+				return xs.firstIndex(of:5)
```
turbolent, line 9652:
```suggestion

      fun test(): Int? {
      return xs.firstIndex(of: 5)
```

turbolent (commented):
@bjartek Thank you for updating and adding more tests. There are some (minor) whitespace issues and it would be great if you could add one more checker test case for non-equatable values, then it should be good to go!

docs/language/values-and-types.mdx
```diff
@@ -1079,6 +1079,23 @@ are available for both variable-sized and fixed-sized or variable-sized arrays.
   let containsKitty = numbers.contains("Kitty")
   ```
 
+- `cadence•fun firstIndex(of:: T): Int?`
+ Returns the index of the first matching elmement of type `T`, or nil if no match
```
turbolent, line 1083:
```suggestion
- `cadence•fun firstIndex(of: T): Int?`
 Returns the index of the first matching element of type `T`, or nil if no match.
 Available if `T` is not resource-kinded and equatable.
```

runtime/sema/type.go
```diff
@@ -1511,6 +1511,10 @@ type ArrayType interface {
 	isArrayType()
 }
 
+const arrayTypeFirstIndexFunctionDocString = `
+Returns the index of the first element matching the given object in the array, nil if no match
```
turbolent, line 1515:
```suggestion
Returns the index of the first element matching the given object in the array, nil if no match.
 Available if the array element type is not resource-kinded and equatable.
```

runtime/tests/checker/arrays_dictionaries_test.go
bjartek: I added test with array of arrays.

turbolent (approved):
Looks great, nice work! 👏 

@SupunS Could you please also have a look?

turbolent:
@SupunS @dsainati1 when you get a change, could you please have a second look? Thank you!

runtime/sema/type.go
```diff
@@ -1621,6 +1626,44 @@ func getArrayMembers(arrayType ArrayType) map[string]MemberResolver {
...
+					report(
+						&InvalidResourceArrayMemberError{
+							Name:            identifier,
+							DeclarationKind: common.DeclarationKindFunction,
+							Range:           targetRange,
+						},
+					)
+				}
+
+				// TODO: implement Equatable interface: https://github.com/dapperlabs/bamboo-node/issues/78
```
SupunS, line 1648: Is this comment needed? Probably outdated now?

SupunS (approved):
Looks good! 🎉 

Thank you for the patient on the long delay for the review 🙏

runtime/sema/type.go
turbolent: Yes, the comment/TODO is still relevant, but the URL is outdated and should be removed:
```suggestion
				// TODO: implement Equatable interface
```

PR #1353, opened 2022-01-11
added replace to string

bjartek:
## Description

Add replace(from:String, to:String) to String.

______

### DIFF
--- a/runtime/sema/string_type.go
+++ b/runtime/sema/string_type.go
@@ -120,6 +120,17 @@ func init() {
 					)
 				},
 			},
+			"replace": {
+				Kind: common.DeclarationKindField,
+				Resolve: func(identifier string, _ ast.Range, _ func(error)) *Member {
+					return NewPublicConstantFieldMember(
+						t,
+						identifier,
+						StringTypeReplaceFunctionType,
+						stringTypeReplaceFunctionDocString,
+					)
+				},
+			},
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


turbolent:
Thank you for contributing this @bjartek!

One open question is what the behaviour in regards to equivalent Unicode sequences should be. Cadence strings are encoded as UTF-8 and logically consist of Unicode grapheme clusters, like in Swift, but unlike Go, where strings are just byte arrays.

This also means that strings replacements should operate on logical characters, i.e. Unicode grapheme clusters, and not purely on bytes.

For example, these two strings are equivalent: `"Caf\u{65}\u{301}" == "Caf\u{e9}"`.

However, observe the result in Go:

```go
strings.ReplaceAll("Caf\u0065\u0301", "\u00E9", "X") == "Caf\u0065\u0301"
```

Go's `strings.ReplaceAll` operates on bytes, and the searched byte sequence does not occur in the source.

Now compare this to Swift:

```swift
"Caf\u{65}\u{301}".replacingOccurrences(of: "\u{E9}", with: "X") == "CafX"
```

I don't know of the top of my head what algorithm is used for the replacement, it could be that Unicode normalization is involved, like when testing equality.

bjartek:
Good questions. I have not though that far tbh. Maybe we can look at a different package to do the replacement in go. Or call the method replaceAscii or something?

bjartek:
this is over a. year old @turbolent should we just close it?

runtime/sema/string_type.go
```diff
@@ -195,3 +206,23 @@ var StringTypeToLowerFunctionType = &FunctionType{
...
 `
+
+var StringTypeReplaceFunctionType = &FunctionType{
+	Parameters: []*Parameter{
+		{
+			Identifier:     "from",
+			TypeAnnotation: NewTypeAnnotation(StringType),
+		},
+		{
+			Identifier:     "upTo",
```
bluesign, line 217: s/upTo/to

bluesign:
oh I thought this is new PR, this would be very nice to have.

turbolent:
Yes, if you want to. We can always re-open it

PR #1363, opened 2022-01-15
add identity field to AuthAccount

### LINKED ISSUE #423: Proving access to an authorized account
### Issue To Be Solved

A common use-case is proving access to an authorized account `AuthAccount`.

### Suggest A Solution

- Declare a new type `Identity` that cannot be constructed
- Add a new field `let identity: &Identity` to `AuthAccount`. The reference ensures the value cannot be stored

bjartek:
Closes #423 

## Description
Add an Identity type and a field to AuthAccount that retuns a reference to it. 

______


bjartek:
@turbolent do you have any hints on how to go forward here, There is quite a lot boilerplate and I am not sure I have finished it all.

joshuahannan:
I don't know if this makes sense. It encourages `msg.sender` access control list patterns that we have explicitly tried to avoid in Cadence.

bjartek:
My usecase for this is to add context to events, since in some scenarios it is really hard to add "the other side" to an event.

turbolent:
@joshuahannan Thank you for your feedback. Maybe we should propose such an addition through a FLIP?

joshuahannan:
Maybe a FLIP would be good. Though I guess it wouldn't hurt to have a feature like this on the chance that it could be useful, but I worry that people would use it the wrong way. But that might not be the responsibility of the language to enforce and could be something that is just viewed as a best practice

bluesign:
I agree with @joshuahannan in this case, but also I feel like we lost the war on msg.sender pattern with resource owner already. 

This will be more clear vs people abusing resource owner for sure. 

But eventually FLIP can be good to see other alternatives, community use cases etc.

docs/language/accounts.mdx
```diff
@@ -167,6 +171,25 @@ to the `prepare` phase of the transaction.
   which discard their changes upon completion. 
   Attempting to use this function outside of a script will cause a type error. 
 
+## Identity
+Identity can only be obtained as a reference from an AuthAccount and has a single `address` field. 
+
+The main use case for Identity is to send it from a transaction into a contract method to know the address of who called it.
```
robert-e-davidson3, line 177:
```suggestion
The main use case for Identity is to send it from a transaction into a contract method to know the caller's address.
```

docs/language/accounts.mdx
```diff
@@ -167,6 +171,25 @@ to the `prepare` phase of the transaction.
...
+```cadence
+
+// example fun 
+pub fun someMethod(identity: &Identity) {
+   log("I was called by ".concat(identity.address))
+}
+
+transaction() {
+    prepare(signer: AuthAccount) {
+				someMethod(signer.identity)
```
robert-e-davidson3, line 188: Whitespace shenanigans in a few places in this diff. Are they just an artifact of the medium or is the source code going to look weird?

runtime/sema/identity_type.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"github.com/onflow/cadence/runtime/common"
+)
+
+const IdentityTypeName = "Identity"
+const IdentityAddressField = "address"
+
+// IdentityType represents the Identity of an Auth Account
+// Identity can only be fetch as a reference from AuthAccount so it guarantees that the address in it comes from one of the signers.
```
robert-e-davidson3, line 29:
```suggestion
// Identity can only be fetched as a reference from AuthAccount so it guarantees that the address in it comes from one of the signers.
```

robert-e-davidson3 (approved):

bjartek:
Sure

turbolent:
The [FLIP proposing the addition of this feature ](https://github.com/onflow/flow/pull/945)has been rejected, so closing this implementation PR.

Thank you very much for your PR @bjartek, it sparked a very fruitful discussion, and hopefully an even better solution to the problem it tries to solve. 🙏

PR #1374, opened 2022-01-22
Meter string creation

turbolent:
Work towards dapperlabs/cadence-private-issues#2

Start metering the allocation of string values. 
They are allocated in various places, e.g. 
- In the interpreter, by string operations like concatenating strings
- In the lexer when creating AST nodes for string literals
- When importing values (transaction/script arguments)
- When decoding stored values

There are still some unclear / unmetered cases, e.g.
- The result of a FVM operation, like querying the names of contracts deployed to an account
- The result of computing type IDs and qualified strings

______


runtime/convertValues.go
```diff
@@ -446,7 +446,8 @@ func importValue(inter *interpreter.Interpreter, value cadence.Value, expectedTy
 	case cadence.Bool:
 		return interpreter.BoolValue(v), nil
 	case cadence.String:
-		return interpreter.NewStringValue(string(v)), nil
+		// TODO: meter?
```
dsainati1, line 449: I feel like this probably does need metering; if we only meter strings that users create from within their smart contracts rather than strings that they import from storage or elsewhere, this creates a weird incentive for people to never create strings in their code and instead load them from outside.

runtime/interpreter/big_test.go
```diff
@@ -0,0 +1,81 @@
...
+	assert.LessOrEqual(t, 1, OverEstimateIntStringLength(1))
+	assert.LessOrEqual(t, 1, OverEstimateIntStringLength(9))
+	assert.LessOrEqual(t, 2, OverEstimateIntStringLength(10))
+	assert.LessOrEqual(t, 2, OverEstimateIntStringLength(99))
+	assert.LessOrEqual(t, 3, OverEstimateIntStringLength(100))
+	assert.LessOrEqual(t, 3, OverEstimateIntStringLength(999))
+	assert.LessOrEqual(t, 4, OverEstimateIntStringLength(1000))
+}
+
+func TestOverEstimateBigIntStringLength(t *testing.T) {
```
dsainati1, line 46: Given the above points about the bounding concerns for very very large numbers, we should probably have a test for some extremely large `BigInt`s here.

runtime/interpreter/decode.go
```diff
@@ -239,7 +239,8 @@ func (d Decoder) decodeStorable() (atree.Storable, error) {
 }
 
 func (d Decoder) decodeString(v string) *StringValue {
-	return NewStringValue(v)
+	// TODO: meter? but string is already decoded
```
dsainati1, line 242: similarly to above, we don't want to incentive people to try to use storage to get around metering.

runtime/interpreter/value.go
```diff
@@ -594,14 +595,20 @@ type StringValue struct {
...
-func NewStringValue(str string) *StringValue {
+func NewUnmeteredStringValue(str string) *StringValue {
 	return &StringValue{
 		Str: str,
 		// a negative value indicates the length has not been initialized, see Length()
 		length: -1,
 	}
 }
 
+func NewStringValue(interpreter *Interpreter, memoryUsage MemoryUsage, stringConstructor func() string) *StringValue {
```
dsainati1, line 606: If I understand correctly, the API requires the memory usage to be computed before the string is constructed so that we can elect to abort execution and not construct the string if we would go over our allotted memory. However, this API also complicates the creation of strings, and will ultimately complicate the construction of all kinds of values once we extend metering to everything else. 

Would it be possible to optimistically proceed with the creation of the string here (or int or other primitives) and check the used memory at the end, and then abort if that puts us over the limit? We might need to define a max string size or int size that is also checked to ensure we don't run out of memory as a result of the creation of a single value, but this would result in a much simpler implementation.

runtime/interpreter/interpreter_expression.go
```diff
@@ -513,7 +513,8 @@ func (interpreter *Interpreter) VisitFixedPointExpression(expression *ast.FixedP
 }
 
 func (interpreter *Interpreter) VisitStringExpression(expression *ast.StringExpression) ast.Repr {
-	return NewStringValue(expression.Value)
+	// TODO: meter?
```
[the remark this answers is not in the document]
turbolent: They would not be unmetered, it's just that we do not need to (re)meter here: The string value was already allocated during parsing, so the metering should be there

runtime/interpreter/value.go
```diff
@@ -594,14 +595,20 @@ type StringValue struct {
...
+	interpreter.useMemory(memoryUsage)
+	str := stringConstructor()
+	return NewUnmeteredStringValue(str)
```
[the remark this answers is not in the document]
turbolent: The length of a string is computed lazily / only if needed. -1 is used as a sentinel value to indicate that the length was not calculated yet. See the comment above

runtime/interpreter/value.go
turbolent: Yes, I tried to design the API in such a way that the upfront memory usage calculation is "enforced" as much as possible: The constructor requires the upfront calculated measuring, and the allocation is performed lazily, after metering. I agree that this makes the construction of values slightly more work, but every allocation has the potential for an unmetered code path, so IMHO the extra effort required is worth the benefit.

I agree that it would be even better if we could devise an API that is simpler and still has the same effect.

runtime/interpreter/decode.go
turbolent: Agreed, we should meter decoded values. The problem here is that it by the time this function the value constructor is called here, the allocation already happened, in the CBOR library.

runtime/convertValues.go
turbolent: 👍 

Added in c4c2d5e

runtime/interpreter/big_test.go
turbolent: Good point! Improving the tests, and testing `math.MinInt` uncovered a bug. Fixed the estimation and improved the tests in 435d171

runtime/interpreter/decode.go
turbolent: Pinged Faye to ask how we can get the size of the string from the CBOR stream decoder before calling the string decoding function (`StreamDecoder.DecodeString`)

turbolent: All strings which are decoded are also metered now (not just string values, but also e.g in locations, `StringAtreeValue`, etc)

runtime/interpreter/value.go
```diff
@@ -1026,7 +1077,7 @@ func (v *StringValue) Transfer(
 }
 
 func (v *StringValue) Clone(_ *Interpreter) Value {
-	return NewStringValue(v.Str)
+	return NewUnmeteredStringValue(v.Str)
```
robert-e-davidson3, line 1080: Why is this unmetered?

turbolent: `Clone` is an internal function that is not used by the interpreter. It can e.g. be used for a state migration

runtime/parser2/parser.go
```diff
@@ -45,6 +46,8 @@ type parser struct {
 	backtrackingCursorStack []int
 	// bufferedErrorsStack is the stack of parsing errors encountered during buffering
 	bufferedErrorsStack [][]error
+	// memoryGauge is used to meter memory usage
+	memoryGauge common.MemoryGauge
```
SupunS, line 50: This field seems unused

runtime/runtime.go
```diff
@@ -2961,7 +2981,8 @@ func (r *interpreterRuntime) newAccountContractsGetNamesFunction(
 
 		values := make([]interpreter.Value, len(names))
 		for i, name := range names {
-			values[i] = interpreter.NewStringValue(name)
+			// TODO: meter?
+			values[i] = interpreter.NewUnmeteredStringValue(name)
```
SupunS, line 2985: I think we should meter all cadence-values that get created at the runtime interface (e.g: values passed over from FVM), since its one of the three ways a value can be constructed in cadence (i.e: in-line, program-arguments, sent from FVM)

SupunS (approved):
Approach looks good to me! The comments I added are mostly for refactoring, so +1 to do them in a separate PR.

runtime/convertValues.go
```diff
@@ -554,6 +554,20 @@ func importValue(inter *interpreter.Interpreter, value cadence.Value, expectedTy
 	return nil, fmt.Errorf("cannot import value of type %T", value)
 }
 
+func importString(inter *interpreter.Interpreter, v cadence.String) *interpreter.StringValue {
+	memoryUsage := common.MemoryUsage{
+		Kind:   common.MemoryKindString,
+		Amount: uint64(len(v)),
+	}
```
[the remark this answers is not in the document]
turbolent: Good idea, added in [e39bdef](https://github.com/onflow/cadence/pull/1374/commits/e39bdefd76653bb7c70811185ed90a5d08433d34)

runtime/interpreter/decode.go
```diff
@@ -1540,3 +1399,237 @@ func decodeCompositeTypeInfo(dec *cbor.StreamDecoder) (atree.TypeInfo, error) {
 		kind:                common.CompositeKind(kind),
 	}, nil
 }
+
+func DecodeStaticType(decoder *cbor.StreamDecoder, memoryGauge common.MemoryGauge) (StaticType, error) {
+	return TypeDecoder{
+		decoder:     decoder,
+		memoryGauge: memoryGauge,
+	}.decodeStaticType()
```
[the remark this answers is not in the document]
turbolent: Good point, I'll refactor 👍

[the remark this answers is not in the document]
turbolent: Refactored in 83650f276a4c633bcf2f8bab4fa21b2af1ecf170

runtime/interpreter/decode.go
```diff
@@ -1540,3 +1399,237 @@ func decodeCompositeTypeInfo(dec *cbor.StreamDecoder) (atree.TypeInfo, error) {
...
+			return nil, err
+		}
+		return emptyTypeInfo, nil
+
+	default:
+		return nil, fmt.Errorf("invalid type info CBOR type: %d", ty)
+	}
+}
+
+type LocationDecoder struct {
```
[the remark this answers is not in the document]
turbolent: Refactored in 83650f276a4c633bcf2f8bab4fa21b2af1ecf170

runtime/interpreter/value.go
```diff
@@ -2266,7 +2307,18 @@ func getNumberValueMember(v NumberValue, name string, typ sema.Type) Value {
 	case sema.ToStringFunctionName:
 		return NewHostFunctionValue(
 			func(invocation Invocation) Value {
-				return NewStringValue(v.String())
+				interpreter := invocation.Interpreter
+				memoryUsage := common.MemoryUsage{
+					Kind:   common.MemoryKindString,
+					Amount: uint64(OverEstimateNumberStringLength(v)),
```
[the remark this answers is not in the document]
turbolent: Good idea, refactored in 0f4f8afb0f32e7da763bf5a40c0fba840c5fc8b7

runtime/parser2/lexer/state.go
```diff
@@ -290,6 +292,13 @@ func identifierState(l *lexer) stateFn {
 
 func stringState(l *lexer) stateFn {
 	l.scanString('"')
+	memoryGauge := l.memoryGauge
+	if memoryGauge != nil {
+		memoryGauge.UseMemory(common.MemoryUsage{
+			Kind:   common.MemoryKindString,
+			Amount: uint64(l.wordLength()),
+		})
+	}
```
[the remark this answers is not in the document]
turbolent: I was hesitant to do that here, actually even in the interpreter, because it would change the semantics, i.e. it would always allocate the memory usage value, even if metering is disabled (there is no gauge)

runtime/parser2/parser.go
turbolent: Right, and the parser/lexer integration was not working properly at all. Fixed this in cbe726b01ea028157cc902e342ea2cd59fce4837

runtime/runtime.go
turbolent: Agreed, I will address this in a follow-up PR 👍

runtime/interpreter/storage_test.go
```diff
@@ -36,7 +36,7 @@ func TestCompositeStorage(t *testing.T) {
 
 	t.Parallel()
 
-	storage := NewInMemoryStorage()
+	storage := NewInMemoryStorage(nil)
```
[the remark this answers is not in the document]
turbolent: I thought about it, but given it's longer, I didn't bother. I'll replace it for tests in a follow-up PR

turbolent:
Once green I'll merge this into the feature branch and will address the follow-up work in (a) separate PR(s):
- `TypeValue.GetMember` `identifier`: Type ID generation must be metered (have a metered string builder)
- `CharacterValue.GetMember` `toString`: Can probably meter with amount 1
- Meter values FVM values (newAccountContractsGetNamesFunction (Account.Contracts.names))
- Replace `NewInMemoryStorage(nil)` with `NewUnmeteredInMemoryStorage()` in tests
- Add tests	
- Re-evaluate unmetered cases. Might still want to meter `StringValue` instance, maybe with amount 1

SupunS (approved):
Checked the most recent changes/commits, and LGTM! 👌

PR #1392, opened 2022-01-31
Add CharacterValue type

### LINKED ISSUE #1373: Missing dedicated CharacterValue leads to incorrect run-time type value
### Problem

Indexing into a `String` should result in a `Character`. However, string indexing is implemented in the interpreter as returning a `StringValue`. This leads to `getType` returning a run-time type value of incorrect type `String`, but it should be `Character`.

### Steps to Reproduce

```swift
let c: Character = "x"[0]
c.getType()  // returns `Type<String>()`, but should return `Type<Character>()`
```

### Suggested Solution

Ideas:
- Introduce `CharacterValue` and return it in `StringValue.GetKey`
- Add a flag `IsSingleCharacter bool` to `StringValue`, and return the appropriate static and dynamic type based on the flag. Return a `StringValue` with `IsSingleCharacter: true` in `StringValue.GetKey`

dsainati1:
Closes #1373

## Description

Cadence previously lacked a runtime character value, causing some inconsistency between the static `Character` type and values with that type at runtime. This adds runtime `Character` values, which are produced by indexing into a string and by creating a character variable like so: `let x: Character = "a"`.

______

### DIFF
--- a/encoding/json/encoding_test.go
+++ b/encoding/json/encoding_test.go
@@ -95,12 +95,12 @@ func TestEncodeCharacter(t *testing.T) {
 		{
 			"a",
 			cadence.NewCharacter('a'),
-			`{"type":"Bool","value":'a'}`,
+			`{"type":"Character","value":97}`,
 		},
 		{
 			"b",
-			cadence.NewCharacter('a'),
-			`{"type":"Bool","value":'b'}`,
+			cadence.NewCharacter('b'),
+			`{"type":"Character","value":98}`,
 		},
 	}...)
 }
--- a/encoding/json/encoding_test.go
+++ b/encoding/json/encoding_test.go
@@ -94,12 +94,12 @@ func TestEncodeCharacter(t *testing.T) {
 	testAllEncodeAndDecode(t, []encodeTest{
 		{
 			"a",
-			cadence.NewCharacter(byte('a')),
+			cadence.NewCharacter('a'),
 			`{"type":"Bool","value":'a'}`,
 		},
 		{
 			"b",
-			cadence.NewCharacter(byte('a')),
+			cadence.NewCharacter('a'),
 			`{"type":"Bool","value":'b'}`,
 		},
 	}...)
--- a/encoding/json/encoding_test.go
+++ b/encoding/json/encoding_test.go
@@ -87,6 +87,24 @@ func TestEncodeBool(t *testing.T) {
 	}...)
 }
 
+func TestEncodeCharacter(t *testing.T) {
+
+	t.Parallel()
+
+	testAllEncodeAndDecode(t, []encodeTest{
+		{
+			"a",
+			cadence.NewCharacter(byte('a')),
+			`{"type":"Bool","value":'a'}`,
+		},
+		{
+			"b",
+			cadence.NewCharacter(byte('a')),
+			`{"type":"Bool","value":'b'}`,
+		},
+	}...)
+}
+
 func TestEncodeString(t *testing.T) {
 
 	t.Parallel()
--- a/runtime/tests/interpreter/string_test.go
+++ b/runtime/tests/interpreter/string_test.go
@@ -205,3 +205,23 @@ func TestInterpretStringToLower(t *testing.T) {
 		result,
 	)
 }
+
+func TestInterpretStringAccess(t *testing.T) {
+
+	t.Parallel()
+
+	inter := parseCheckAndInterpret(t, `
+	fun test(): Type {
+		let c: Character = "x"[0]
+		return c.getType() 
+	}
+	`)
+
+	result, err := inter.Invoke("test")
+	require.NoError(t, err)
+
+	require.Equal(t,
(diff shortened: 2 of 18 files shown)


turbolent (commented):
Looks good, great work!

encoding/json/encoding_test.go
```diff
@@ -87,19 +87,55 @@ func TestEncodeBool(t *testing.T) {
...
+
+		t.Parallel()
+		_, err := cadence.NewCharacter(`\a`)
+		require.Error(t, err)
+	})
+
+	t.Run("ok complex", func(t *testing.T) {
+
+		t.Parallel()
+		_, err := cadence.NewCharacter(`\u{75}\u{308}`)
```
turbolent, line 118: 👌

runtime/format/character.go
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package format
+
+import (
+	"github.com/rivo/uniseg"
+)
+
+func IsValidCharacter(s string) bool {
```
turbolent, line 25: Good idea to add this function and re-use it👌 

Package `format` is so far meant for rendering Cadence values to strings. Maybe move the function to e.g. `sema` or `common`

runtime/format/fix.go
```diff
@@ -23,12 +23,12 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/onflow/cadence/runtime/sema"
+	"github.com/onflow/cadence/fixedpoint"
```
turbolent, line 26: Great idea! This removes the dependency of `format` on `sema` 👍

values.go
```diff
@@ -189,8 +189,11 @@ func (v Bytes) String() string {
 //
 type Character string
 
-func NewCharacter(b string) Character {
-	return Character(b)
+func NewCharacter(b string) (Character, error) {
+	if !format.IsValidCharacter(b) {
+		return "_", fmt.Errorf("invalid character: %s", b)
```
turbolent, line 194: Best to follow the Go idiom of returning the empty value for the type, i.e., for `string`, and empty `""`. Callers should only use one of the return values (first if second/error is nil, or only the second/error if non-nil), similar to https://github.com/onflow/cadence/pull/1392/files#diff-e9b60dd44adee3235a03e234d28b9711ab20c09f79ea62bc6be1d924dc59a4ecR258

```suggestion
		return "", fmt.Errorf("invalid character: %s", b)
```

If really needed, maybe consider returning [REPLACEMENT CHARACTER / U+FFFD](https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character) instead of the arbitrary underscore `_`

runtime/tests/interpreter/string_test.go
```diff
@@ -205,3 +205,139 @@ func TestInterpretStringToLower(t *testing.T) {
...
+	t.Parallel()
+
+	inter := parseCheckAndInterpret(t, `
+        let a: Character = "ü"
+        let b: Character = "\u{FC}"
+        let c: Character = "\u{75}\u{308}"
+        let d: Character = "y"
+        let x = a == b
+        let y = a == c
+		let z = a == d
```
turbolent, line 320:
```suggestion
        let z = a == d
```

runtime/tests/interpreter/string_test.go
```diff
@@ -205,3 +205,139 @@ func TestInterpretStringToLower(t *testing.T) {
...
+	result, err := inter.Invoke("test")
+	require.NoError(t, err)
+
+	require.Equal(t,
+		interpreter.NewStringValue("x"),
+		result,
+	)
+}
+
+func TestCompareCharacters(t *testing.T) {
```
turbolent, line 309:
```suggestion
func TestInterpretCompareCharacters(t *testing.T) {
```

runtime/tests/interpreter/string_test.go
```diff
@@ -205,3 +205,139 @@ func TestInterpretStringToLower(t *testing.T) {
...
+	result, err := inter.Invoke("test")
+	require.NoError(t, err)
+
+	require.Equal(t,
+		interpreter.TypeValue{Type: interpreter.PrimitiveStaticTypeString},
+		result,
+	)
+}
+
+func TestConvertCharacterToString(t *testing.T) {
```
turbolent, line 289:
```suggestion
func TestInterpretConvertCharacterToString(t *testing.T) {
```

runtime/sema/checker.go
```diff
@@ -1928,15 +1929,13 @@ func (checker *Checker) checkFieldsAccessModifier(fields []*ast.FieldDeclaration
 // i.e. it has exactly one grapheme cluster.
 //
 func (checker *Checker) checkCharacterLiteral(expression *ast.StringExpression) {
-	length := uniseg.GraphemeClusterCount(expression.Value)
-
-	if length == 1 {
+	if format.IsValidCharacter(expression.Value) {
```
turbolent, line 1932: 🙏

turbolent (approved):
Nice work! Just a couple minor things and then ready to get merged 👍

Issue #1394, opened 2022-02-02
Interface conformance requirements for contract updates is too restrictive


SupunS:
### Problem

The documents suggest that adding/removing/changing interface conformance for a struct/resource is possible during updating a contract: https://docs.onflow.org/cadence/language/contract-updatability/#structs-resources-and-interfaces

However, it is restricted in the implementation: https://github.com/onflow/cadence/blob/0ce830012ae0f3d9358b18a16ecbdb17bf20a95a/runtime/contract_update_validation.go#L492

### Suggested solution

Conformance check is only required for enums, as changing enum's type can cause data inconsistencies. However, this restriction can be lifted for other types of composite declarations.

bluesign:
@SupunS We had a recent discussion on Discord. Do we only consider storage as breaking for contract updates?

Or are we also considering breaking other contracts ? ( With composability using this contract )

SupunS:
We only consider maintaining the data consistency during the contract update validations, and by doing so, trying to maintain the 'type safety' of cadence. There is a note on the docs that also specifies the goal: https://docs.onflow.org/cadence/language/contract-updatability/#validation-goals

On contrary, trying to protect the other smart contract's composability would bring too many restrictions. I believe maintaining composability is a bigger and separate problem to be solved, that may need a separate solution, like versioning smart contracts, etc.

bluesign:
> We only consider maintaining the data consistency during the contract update validations

Do you mean by breaking every time changes ? or does it possibility break changes included ?

SupunS:
Yeah, if there is any 'potential to break', then we do not allow such updates. TBH I haven't fully think through all the cases, and what changes to conformances are OK and what are not. But there's definitely a mismatch between the docs and the implementation so we should rectify that.

My initial thought was it's ok to lift some/most of these restrictions on conformances for structs/resources, but I could be wrong.

turbolent:
Currently, the validation requires the conformances to stay the same, ignoring order, i.e., same count and elements; and most developers are currently trying to add a conformance. So maybe we can just focus on what allowing that would entail? I guess adding a conformance is fine?

As @SupunS noted, if adding a conformance also requires changing existing data (i.e: adding a new field that is enforced by the new conformance), then the separate check for adding new fields would reject such an update. Adding a conformance which only requires functions should be safe, or at least I can’t think of a counter-example

turbolent:
This has been released in 0.21.1

PR #1395, opened 2022-02-02
Allows adding new interface conformances during a contract update

SupunS:
Work towards #1394

## Description

This PR allows adding interface conformances during a contract update.

Removing existing conformances are continue to be not allowed.

______

### DIFF
--- a/runtime/contract_update_validation.go
+++ b/runtime/contract_update_validation.go
@@ -494,30 +494,33 @@ func (validator *ContractUpdateValidator) checkConformances(
 	newDecl *ast.CompositeDeclaration,
 ) {
 
+	// Here it is assumed enums will always have one and only one conformance.
+	// This is enforced by the checker.
+	// Therefore, below check for multiple conformances is only applicable
+	// for non-enum type composite declarations. i.e: structs, resources, etc.
+
 	oldConformances := oldDecl.Conformances
 	newConformances := newDecl.Conformances
 
-	if len(oldConformances) != len(newConformances) {
-		validator.report(&ConformanceCountMismatchError{
-			Expected: len(oldConformances),
-			Found:    len(newConformances),
-			Range:    ast.NewRangeFromPositioned(newDecl.Identifier),
-		})
+	// All the existing conformances must have a match. Order is not important.
+	// Having extra new conformance is OK. See: https://github.com/onflow/cadence/issues/1394
+	for _, oldConformance := range oldConformances {
+		found := false
+		for _, newConformance := range newConformances {
+			err := oldConformance.CheckEqual(newConformance, validator)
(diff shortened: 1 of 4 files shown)


runtime/contract_update_validation.go
```diff
@@ -494,30 +494,33 @@ func (validator *ContractUpdateValidator) checkConformances(
...
-
-		// If the lengths are not the same, trying to match the conformance
-		// may result in too many regression errors. hence return.
-		return
-	}
+	// All the existing conformances must have a match. Order is not important.
+	// Having extra new conformance is OK. See: https://github.com/onflow/cadence/issues/1394
+	for _, oldConformance := range oldConformances {
+		found := false
+		for _, newConformance := range newConformances {
```
turbolent, line 509: Maybe we can optimize the quadratic search here to avoid blow-up (in a follow-up PR). I guess because the outer loop break on first missing, it is actually OK?

runtime/contract_update_validation_test.go
```diff
@@ -1961,22 +1953,351 @@ func TestContractUpdateValidationDisabled(t *testing.T) {
...
+                    self.a = 0
+                }
+            }`
 
 		err := deployAndUpdate(t, "Test1", oldCode, newCode)
 		require.NoError(t, err)
 	})
 }
+
+func TestConformanceChanges(t *testing.T) {
```
turbolent, line 1976: Good idea to add full runtime tests 👍

turbolent (approved):
Great work!

runtime/contract_update_validation.go
SupunS: Optimized it here: 3c22c34

runtime/contract_update_validation.go
```diff
@@ -506,10 +506,14 @@ func (validator *ContractUpdateValidator) checkConformances(
...
 		found := false
-		for _, newConformance := range newConformances {
+		for index, newConformance := range newConformances {
 			err := oldConformance.CheckEqual(newConformance, validator)
 			if err == nil {
 				found = true
+
+				// Remove the matched conformance, so we don't have to check it again.
+				// i.e: optimization
+				newConformances = append(newConformances[:index], newConformances[index+1:]...)
```
turbolent, line 516: Great idea! Double checked with https://go.dev/play/p/N233f-oGIDh, as I was not quite sure how modifying `newConformances` here affects the `range` over it / the inner loop

turbolent (approved):
Nice work!

runtime/contract_update_validation.go
SupunS: It always breaks from the inner-loop after modification, so should be safe. Thanks for double-checking!

PR #1399, opened 2022-02-04
Panic when testRuntimeInterface methods need hooks but lack them

robert-e-davidson3:
## Description

The runtime test interface has hooks for the FVM methods of the "Interface" API. Most have defaults for when hooks aren't set. A few instead throw `invalid memory address or nil pointer dereference`. This PR gives them defaults where appropriate else a more descriptive panic.

Update: Switching to make all of the methods panic when they don't have a hook.

______

### DIFF
--- a/runtime/runtime_test.go
+++ b/runtime/runtime_test.go
@@ -409,22 +409,37 @@ func (i *testRuntimeInterface) Hash(data []byte, tag string, hashAlgorithm HashA
 }
 
 func (i *testRuntimeInterface) SetCadenceValue(owner common.Address, key string, value cadence.Value) (err error) {
+	if i.setCadenceValue == nil {
+		panic("must specify testRuntimeInterface.setCadenceValue")
+	}
 	return i.setCadenceValue(owner, key, value)
 }
 
 func (i *testRuntimeInterface) GetAccountBalance(address Address) (uint64, error) {
+	if i.getAccountBalance == nil {
+		panic("must specify testRuntimeInterface.getAccountBalance")
+	}
 	return i.getAccountBalance(address)
 }
 
 func (i *testRuntimeInterface) GetAccountAvailableBalance(address Address) (uint64, error) {
+	if i.getAccountAvailableBalance == nil {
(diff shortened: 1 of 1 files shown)


runtime/runtime_test.go
```diff
@@ -409,22 +409,37 @@ func (i *testRuntimeInterface) Hash(data []byte, tag string, hashAlgorithm HashA
 }
 
 func (i *testRuntimeInterface) SetCadenceValue(owner common.Address, key string, value cadence.Value) (err error) {
+	if i.setCadenceValue == nil {
+		panic("must specify testRuntimeInterface.setCadenceValue")
```
turbolent, line 413: Good idea! Maybe even do this consistently for all functions, currently some just return a default of no hook is provided. Better to have to explicitly provide functions that are called in the tests

turbolent (approved):

turbolent:
@robert-e-davidson3 Please follow the checklist in the PR template as well as you can. It helps with avoiding mistakes, labeling helps with generating the release notes, etc. Thank you!

runtime/runtime_test.go
robert-e-davidson3: Ah I just switched some to having defaults! I can switch them all over though. I agree that being explicit is better since defaults tend to hide bugs.

turbolent: Agreed, better to switch to `panic`s, it's a great idea!

robert-e-davidson3: A LOT of tests fail when I eliminate the defaults. I created https://github.com/onflow/cadence/issues/1400 so we can account for the work and maybe assign it later depending on priority. It's a constrained enough problem that I should be able to it without needing much if any help but I don't want to push more urgent stuff back.

However, there are a few more methods that currently panic with a generic message instead of a nice one. I'm going to add them to this PR before asking for another review (hopefully super quick) then merging into master.

turbolent: Sounds good! Thank you for trying and opening the separate PR 👍

turbolent (approved):

PR #1424, opened 2022-02-15
Fix definite field initialization analysis 

### LINKED ISSUE #1405: Definite initialization is too conservative
### Problem

PR #1332 fixed definite initialization analysis, but also made it more conservative. 

### Steps to Reproduce

For example, the following test cases fail, but should not:

```go

func TestCheckFieldInitializationAfterJump(t *testing.T) {

	t.Parallel()

	t.Run("while, continue", func(t *testing.T) {

		t.Parallel()

		_, err := ParseAndCheck(t, `
            struct Test {
                var foo: Int

                init(foo: Int) {
                    while true {
                        continue
                    }
                    self.foo = foo
                }
            }
        `)

		require.NoError(t, err)
	})

	t.Run("while, break", func(t *testing.T) {

		t.Parallel()

		_, err := ParseAndCheck(t, `
            struct Test {
                var foo: Int

                init(foo: Int) {
                    while true {
                        break
                    }
                    self.foo = foo
                }
            }
        `)

		require.NoError(t, err)
	})

	t.Run("for, continue", func(t *testing.T) {

		t.Parallel()

		_, err := ParseAndCheck(t, `
            struct Test {
                var foo: Int

                init(foo: Int) {
                    for i in [] {
                        continue
                    }
                    self.foo = foo
                }
            }
        `)

		require.NoError(t, err)
	})

	t.Run("for, break", func(t *testing.T) {

		t.Parallel()

		_, err := ParseAndCheck(t, `
            struct Test {
                var foo: Int

                init(foo: Int) {
                    for i in [] {
                        break
                    }
                    self.foo = foo
                }
            }
        `)

		require.NoError(t, err)
	})

    t.Run("switch, break", func(t *testing.T) {

		t.Parallel()

		_, err := ParseAndCheckWithPanic(t, `
          struct Test {
              let n: Int

              init(n: Int) {
                  switch n {
                  case 1:
                      break
                  }
                  self.n = n
              }
          }
        `)

		require.NoError(t, err)
	})

}
```

### Suggested Solution

Return info tracking should probably differentiate between switch-case jumps (break) and loop breaks/continues

SupunS:
Closes #1405

## Description

`MaybeJumpedOrReturned` (now `MaybeReturned`) was used only to determine whether a branch of the function has a 'return'. Marking it as `true` for jumps (`break` and `continue` statements) was going against the above assumption/usage.

Reverted the change of using that flag for both jumps and returns, and continue to use it for returns only.
______

### DIFF
--- a/runtime/sema/return_info.go
+++ b/runtime/sema/return_info.go
@@ -19,18 +19,18 @@
 package sema
 
 type ReturnInfo struct {
-	// MaybeJumpedOrReturned indicates that the (branch of) the function
-	// contains a potentially taken return, break, or continue statement
-	MaybeJumpedOrReturned bool
-	DefinitelyReturned    bool
-	DefinitelyHalted      bool
-	DefinitelyJumped      bool
+	// MaybeReturned indicates that the (branch of) the function
+	// contains a potentially taken return statement
+	MaybeReturned      bool
+	DefinitelyReturned bool
+	DefinitelyHalted   bool
+	DefinitelyJumped   bool
 }
 
 func (ri *ReturnInfo) MergeBranches(thenReturnInfo *ReturnInfo, elseReturnInfo *ReturnInfo) {
-	ri.MaybeJumpedOrReturned = ri.MaybeJumpedOrReturned ||
-		thenReturnInfo.MaybeJumpedOrReturned ||
-		elseReturnInfo.MaybeJumpedOrReturned
+	ri.MaybeReturned = ri.MaybeReturned ||
+		thenReturnInfo.MaybeReturned ||
+		elseReturnInfo.MaybeReturned
 
 	ri.DefinitelyReturned = ri.DefinitelyReturned ||
 		(thenReturnInfo.DefinitelyReturned &&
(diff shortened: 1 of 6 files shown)


runtime/sema/return_info.go
```diff
@@ -19,18 +19,18 @@
 package sema
 
 type ReturnInfo struct {
-	// MaybeJumpedOrReturned indicates that the (branch of) the function
-	// contains a potentially taken return, break, or continue statement
-	MaybeJumpedOrReturned bool
-	DefinitelyReturned    bool
-	DefinitelyHalted      bool
-	DefinitelyJumped      bool
+	// MaybeReturned indicates that (a branch of) the function
```
turbolent, line 22: Thank you for fixing the typo here (double "the"). Sorry to be so pernickety, but maybe use "the" branch in the parentheses, as this return info is not for some branch of the function, but "this particular one" / "the current one".
Just worries "a branch" could be misinterpreted to mean "some branch" / "another branch" in the function.

```suggestion
	// MaybeReturned indicates that (the branch of) the function
```

turbolent (approved):
Nice work! 👏 Glad it was an easy fix, I couldn't quite see it 🙂

PR #1425, opened 2022-02-16
Various regression tests

robert-e-davidson3:
Closes https://github.com/dapperlabs/cadence-private-issues/issues/29

## Description

Adds regression tests for Cadence. These bugs were found by Joe Soroka's fuzz testing.

______


runtime/various_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"testing"
+
+	"github.com/onflow/cadence/runtime/parser2"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/onflow/cadence/runtime/sema"
+)
+
+func TestVariousAgainstRegression(t *testing.T) {
```
turbolent, line 32: I think it might be better to move each test to the related component (parser/lexer, sema/checker, interpreter). This allows running the component's tests individually and detect regressions in one component earlier in the development. For example, if we break the parser, all parser tests could pass and then only when running the runtime tests we would realize there is a regression. 

For example, the first test case is a parser test case, so it should go into `runtime/parser2/parser_test.go`.
The next five test cases are checker tests, so they should go into `runtime/tests/checker`.
For the test cases where no error is expected, the tests should be "duplicated" in both the checker tests and the interpreter tests, so we can ensure they are successfully checked, but then also successfully interpreted (other successful tests are "duplicated" in a similar way)

robert-e-davidson3: @turbolent I moved the tests around and restructured them to fit with the other tests in those files.

runtime/tests/checker/contract_test.go
```diff
@@ -708,3 +708,43 @@ func TestCheckInvalidContractNestedTypeShadowing(t *testing.T) {
 		})
 	}
 }
+
+func TestBadContractNesting(t *testing.T) {
```
turbolent, line 712:
```suggestion
func TestCheckBadContractNesting(t *testing.T) {
```

runtime/tests/checker/interface_test.go
```diff
@@ -1991,3 +1991,24 @@ func TestCheckInvalidInterfaceUseAsTypeSuggestion(t *testing.T) {
 		errs[0].(*sema.InvalidInterfaceTypeError).ExpectedType,
 	)
 }
+
+func TestBadStructInterface(t *testing.T) {
```
turbolent, line 1995:
```suggestion
func TestCheckBadStructInterface(t *testing.T) {
```

runtime/tests/checker/resources_test.go
```diff
@@ -8616,3 +8616,64 @@ func TestCheckResourceInvalidationWithConditionalExprInDestroy(t *testing.T) {
 	assert.IsType(t, &sema.InvalidConditionalResourceOperandError{}, errs[0])
 	assert.IsType(t, &sema.InvalidConditionalResourceOperandError{}, errs[1])
 }
+
+func TestBadResourceInterface(t *testing.T) {
```
turbolent, line 8620:
```suggestion
func TestCheckBadResourceInterface(t *testing.T) {
```

turbolent (approved):
Looks good, thank you for adding these and putting them into the related "test suites" 👍 

Only got a couple minor naming suggestions for keeping consistency, nothing major

PR #1432, opened 2022-02-17
Add test for CharacterValue.HashInput

robert-e-davidson3:
Closes nothing

## Description

For some reason we don't have a test for `CharacterValue` hashing so I added one.

I used the example of a large grapheme cluster from the spec according to [this blog post](https://manishearth.github.io/blog/2017/01/14/stop-ascribing-meaning-to-unicode-code-points/).

I didn't verify that the output value is correct because my intent here is just to add test coverage in case of regressions. But I can if desired.

______


turbolent (approved):
Great catch and great test case! 👍

PR #1434, opened 2022-02-18
Add a package to run a script against all accounts

turbolent:
## Description

Add a package that allows running a script against all accounts on a network.

Includes a tool to fetch all contracts and write them into a CSV file.

Code is taken from internal tool authored by @tehranifar, @farbodg, and @janezpodhostnik 🙏 

______


tools/batch-script/batch.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// Config defines that application's config
+type Config struct {
+	// BachSize is the number of addresses for which to run each script
+	BatchSize int
+	// ContinualPoll enables the script to continue polling Flow at the most recent block height after
+	// it is completed the current block height. If AtBlockHeight is set to a value greater than 0,
+	// ContinualPoll is ignored (it will be treated as false).
+	ContinualPoll bool
```
janezpodhostnik, line 51: This is not used. This is a leftover from the fraud-worker code.

janezpodhostnik (approved):
Nice!

tools/batch-script/address_provider.go
```diff
@@ -0,0 +1,224 @@
...
+type AddressProvider struct {
+	log              zerolog.Logger
+	lastAddress      flow.Address
+	generator        *flow.AddressGenerator
+	lastAddressIndex uint
+	referenceBlockID flow.Identifier
+	currentIndex     uint
+}
+
+const endOfAccountsError = "storage used is not initialized or not initialized correctly"
```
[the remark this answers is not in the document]
turbolent: Yeah, when defining `getAccount` we didn't define the behaviour for accounts that do not exist. We should maybe define this better, e.g. abort execution, or allow it and add a field indicating if the account exists

tools/batch-script/get_contracts.cdc
```diff
@@ -0,0 +1,25 @@
+pub fun main(addresses: [Address]): {Address: {String: String}} {
+    let accountContracts: {Address: {String: String}} = {}
+
+    for address in addresses {
+        let account = getAccount(address)
+        let contracts: {String: String} = {}
+
+        if account.storageUsed > 0 as UInt64 {
+            // this produces a deterministic error if account does not exist yet
+        }
```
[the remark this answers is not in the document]
turbolent: Good catch! Removed in 4c4a1239b97aee32b4e0fe333f1b971e43c641d3

tools/batch-script/batch.go
turbolent: Good catch! Removed in 4c4a1239b97aee32b4e0fe333f1b971e43c641d3

tools/batch-script/get_contracts.cdc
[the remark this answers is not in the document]
turbolent: Right 👍 Removed in fc21c88e8b79bba010bd487c78b12b0c667a357e

tools/batch-script/address_provider.go
```diff
@@ -0,0 +1,224 @@
...
+			return true, nil
+		}
+		if strings.Contains(err.Error(), endOfAccountsError) {
+			return false, nil
+		}
+		return false, err
+	}
+
+	// We assume address #2 exists
+	lastAddressIndex, err := ap.getLastAddress(1, 2, true, addressExistsAtIndex)
```
robert-e-davidson3, line 96: Maybe short-circuit the check if address `#2` doesn't exist?

tools/batch-script/address_provider.go
```diff
@@ -0,0 +1,224 @@
...
+// If the last address is at index 7 the algorithm goes like this:
+// (x,y) <- lower and upper index
+// 0. start at (1,2); address exists at 2
+// 1. (2,4): address exists at 4
+// 2. (4,8): address doesnt exist at 8
+// 3. (4,8): check address (8 - 4) / 2 = 6  address exists so next pair is (6,8)
+// 4. (6,8): check address 7 address exists so next pair is (7,8)
+// 5. (7,8): check address (8 - 7) / 2 = 7 ... ok already checked so this is the last existing address
+//
+func (p *AddressProvider) getLastAddress(
```
robert-e-davidson3, line 122: Go really needs to get fast generics so this kind of thing can be generalized :/

turbolent: Hmm, how would generics help here, the types are always constant, no?

tools/batch-script/address_provider.go
turbolent: I just copied the internal code as is and didn't notice this. @janezpodhostnik Do you know why address 2 is assumed to exist? (maybe you @tehranifar?)

tehranifar: Seems like the code was refactored by @janezpodhostnik and @farbodg
The original version was meant to drive a script that operates on a range of accounts, e.g. `30, 31, 32`. If one of those accounts doesn't exist the script panics so the code had to try with `30, 31` and then `30` to make sure it's capturing the end range correctly. The new version is probably achieving the same.

janezpodhostnik: I added that assumption (that account 2 exists) because all the networks the program would be run on had account 2.

During bootstrapping we create at least 4 accounts.

If this is a problematic assumption, the code just needs to test that address no. 2 exists, and if it doesn't address no. 1 is the last one.

tools/batch-script/address_provider.go
```diff
@@ -0,0 +1,224 @@
...
+var brokenAddresses = map[flow.Address]struct{}{
+	flow.HexToAddress("bf48a20670f179b8"): {},
+	flow.HexToAddress("5eba0297874a2bfd"): {},
+	flow.HexToAddress("474ec037bcd8accf"): {},
+	flow.HexToAddress("b0e80595d267f4eb"): {},
+}
+
+func (p *AddressProvider) GenerateAddressBatches(addressChan chan<- []flow.Address, batchSize int) {
+	var done bool
+	for {
```
SupunS, line 191: could use `!done` as the condition (while-loop version)
```suggestion
	for !done {
```

tools/batch-script/address_provider.go
```diff
@@ -0,0 +1,224 @@
...
+			addresses = append(addresses, addr)
+		}
+
+		if len(addresses) > 0 {
+			addressChan <- addresses
+		}
+
+		if done {
+			break
+		}
```
SupunS, line 216: and then remove this snippet

SupunS (approved):
I have a very little understanding of the logic; mostly reviewed the general Go code. Looks fine to me!

PR #1438, opened 2022-02-22
Container metering

dsainati1:
## Description

Based on discussion in https://github.com/dapperlabs/cadence-private-issues/issues/2, this adds metering for container values. The metering here is constant, and does not need to depend on the types or quantities of the values contained, as those are already tracked separately. 

______


runtime/metering_test.go
```diff
@@ -0,0 +1,146 @@
...
+		},
+		Context{
+			Interface: runtimeInterface,
+			Location:  utils.TestLocation,
+		},
+	)
+
+	require.NoError(t, err)
+
+	assert.Equal(t, meter[common.MemoryKindString], uint64(39))
```
dsainati1, line 144: cc @turbolent Does this value seem correct to you? It was larger than I was expecting for the two strings that seem to be getting allocated here, but there is probably some allocation happening behind the scenes that I am missing.

runtime/metering_test.go
```diff
@@ -0,0 +1,146 @@
...
+		},
+		Context{
+			Interface: runtimeInterface,
+			Location:  utils.TestLocation,
+		},
+	)
+
+	require.NoError(t, err)
+
+	assert.Equal(t, meter[common.MemoryKindString], uint64(4))
```
dsainati1, line 92: cc @turbolent Does this value seem correct to you? It was larger than I was expecting for the string that seems to be getting allocated here, but there is probably some allocation happening behind the scenes that I am missing.

runtime/interpreter/simplecompositevalue.go
```diff
@@ -55,6 +57,11 @@ func NewSimpleCompositeValue(
 	fieldFormatters map[string]func(Value, SeenReferences) string,
 	stringer func(SeenReferences) string,
 ) *SimpleCompositeValue {
+	inter.UseMemory(common.MemoryUsage{
+		Kind:   common.MemoryKindComposite,
+		Amount: uint64(1),
```
SupunS, line 62: cast to `uint64` is redundant for literals
```suggestion
		Amount: 1,
```

runtime/context.go
```diff
@@ -54,3 +54,7 @@ func (c *Context) InitializeCodesAndPrograms() {
 		c.programs = map[common.LocationID]*ast.Program{}
 	}
 }
+
+func (c Context) UseMemory(usage common.MemoryUsage) {
```
SupunS, line 58: Whats the usage of this function?

dsainati1: It's a leftover from before; I can remove it now I think.

SupunS (commented):
LGTM!

runtime/interpreter/interpreter.go
```diff
@@ -4483,6 +4483,15 @@ func (interpreter *Interpreter) UseMemory(usage common.MemoryUsage) {
...
 }
 
+// UseKnownMemory uses a pre-determined amount of memory
+//
+func (interpreter *Interpreter) UseKnownMemory(kind common.MemoryKind, amount uint64) {
+	interpreter.UseMemory(common.MemoryUsage{
+		Kind:   kind,
+		Amount: amount,
+	})
+}
```
robert-e-davidson3, line 4493: All of the simple types use `1` for `Amount` too. Since this is such a common use case, maybe a more specific version of this function would be better.

```suggestion
// UseConstantMemory uses a pre-determined amount of memory
//
func (interpreter *Interpreter) UseConstantMemory(kind common.MemoryKind) {
	interpreter.UseMemory(common.MemoryUsage{
		Kind:   kind,
		Amount: uint64(1),
	})
}
```

robert-e-davidson3 (approved):

SupunS:
@dsainati1 Wanted to get this PR in before others, because subsequent PRs were depending on some of the changes introduced in this PR. So I took the liberty of opening a new PR by cherry-picking your commits to resolve the conflicts. 
There were too many conflicts in this PR to resolve, not from your changes, but from the merging of master to the base branch. Cherry-picking had only a trivial conflicts so went with that option. 

Closing this one as the changes are added with https://github.com/onflow/cadence/pull/1472

Thank you for the changes! 🎉

PR #1442, opened 2022-02-23
Memory metering for small types

robert-e-davidson3:
Related to https://github.com/onflow/cadence/pull/1374

## Description

Every bool uses the same amount of memory so this metering is simple: anywhere you're creating a cadence boolean, add the boolean "kind" meter with Amount set to "1".

The same is true of nil.

______

- [X] Targeted PR against `master` branch (bastian/memory-metering actually)


runtime/interpreter/value.go
```diff
@@ -15260,6 +15421,20 @@ type PathValue struct {
...
 
+func NewUnmeteredPathValue(domain common.PathDomain, identifier string) PathValue {
+	return PathValue{domain, identifier}
+}
+
+func NewPathValue(memoryGauge common.MemoryGauge, domain common.PathDomain, identifier string) PathValue {
+	if memoryGauge != nil {
+		memoryGauge.UseMemory(common.MemoryUsage{
+			Kind:   common.MemoryKindPathValue,
+			Amount: uint64(len(identifier)),
```
robert-e-davidson3, line 15432: I use two approaches to length:
1. Make sure it's at least zero.
2. Assume it's non-negative.

Is the second approach safe or is the first necessary?

runtime/interpreter/value.go
```diff
@@ -15506,6 +15681,21 @@ type CapabilityValue struct {
 	BorrowType StaticType
 }
 
+func NewUnmeteredCapabilityValue(address AddressValue, path PathValue, borrowType StaticType) *CapabilityValue {
+	return &CapabilityValue{address, path, borrowType}
+}
+
+func NewCapabilityValue(memoryGauge common.MemoryGauge, address AddressValue, path PathValue, borrowType StaticType) *CapabilityValue {
```
SupunS, line 15688: Same here

runtime/interpreter/value.go
```diff
@@ -15696,6 +15886,23 @@ type LinkValue struct {
 	Type       StaticType
 }
 
+func NewUnmeteredLinkValue(targetPath PathValue, staticType StaticType) LinkValue {
+	return LinkValue{targetPath, staticType}
+}
+
+func NewLinkValue(memoryGauge common.MemoryGauge, targetPath PathValue, staticType StaticType) LinkValue {
```
SupunS, line 15893: Same here

runtime/interpreter/interpreter.go
```diff
@@ -4758,7 +4767,7 @@ func (interpreter *Interpreter) invalidateResource(value Value) {
 // UseMemory delegates the memory usage to the interpreter's memory gauge, if any.
 //
 func (interpreter *Interpreter) UseMemory(usage common.MemoryUsage) {
-	if interpreter.memoryGauge == nil {
+	if interpreter == nil || interpreter.memoryGauge == nil {
```
SupunS, line 4770: I don't think interpreter can be `nil` for any valid cases. If it is, then probably a bug / wrong usage of the method, and might be OK to crash. Was there any particular scenario this was `nil`?

runtime/interpreter/decode.go
```diff
@@ -294,14 +294,14 @@ func (d StorableDecoder) decodeStorable() (atree.Storable, error) {
...
-func (d StorableDecoder) decodeCharacter(v string) (CharacterValue, error) {
+func (d StorableDecoder) decodeCharacter(memoryGauge common.MemoryGauge, v string) (CharacterValue, error) {
 	if !sema.IsValidCharacter(v) {
 		return "", fmt.Errorf(
 			"invalid character encoding: %s",
 			v,
 		)
 	}
-	return NewCharacterValue(v), nil
+	return NewCharacterValue(memoryGauge, v), nil
```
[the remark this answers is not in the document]
robert-e-davidson3: Good point!

runtime/interpreter/interpreter.go
```diff
@@ -3187,7 +3191,12 @@ var converterFunctionValues = func() []converterFunction {
 		convert := declaration.convert
 		converterFunctionValue := NewUnmeteredHostFunctionValue(
 			func(invocation Invocation) Value {
-				return convert(invocation.Arguments[0])
+				val := convert(invocation.Arguments[0])
+				if val == nil {
+					return NewNilValue(invocation.Interpreter)
+				} else {
+					return val
+				}
```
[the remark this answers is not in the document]
robert-e-davidson3: Gotta rebase/merge eventually anyway.

runtime/interpreter/interpreter.go
```diff
@@ -3933,17 +3946,14 @@ func (interpreter *Interpreter) authAccountLinkFunction(addressValue AddressValu
...
 
 			// Write new value
 
 			borrowStaticType := ConvertSemaToStaticType(borrowType)
 
-			linkValue := LinkValue{
-				TargetPath: targetPath,
-				Type:       borrowStaticType,
-			}
+			linkValue := NewLinkValue(interpreter, targetPath, borrowStaticType)
```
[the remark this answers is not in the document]
robert-e-davidson3: Ah yeah this is inconsistent in a bunch of places. I'll change them.

runtime/interpreter/decode.go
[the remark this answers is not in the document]
robert-e-davidson3: I like this idea but doing so would require changing `DecodeStorable` in `decode.go` and possibly its callers (they all pass `nil` so far).

An alternative is to change `UseConstantMemory` to be a function acting instead of a method:

```

//
func UseConstantMemory(memoryGauge common.MemoryGauge, kind common.MemoryKind) {
	memoryGauge.UseMemory(common.MemoryUsage{
		Kind:   kind,
		Amount: 1,
	})
}
```

That makes sense anyway IMO since every kind of MemoryGauge will handle constants the same way.

If there're no objections, I'll convert `UseConstantMemory` to a function.

runtime/interpreter/value.go
```diff
@@ -15009,16 +15163,27 @@ func (*EphemeralReferenceValue) DeepRemove(_ *Interpreter) {
...
 	result := AddressValue{}
 	copy(result[common.AddressLength-len(b):], b)
 	return result
 }
 
+func NewAddressValue(memoryGauge common.MemoryGauge, a common.Address) AddressValue {
+	return NewAddressValueFromBytes(memoryGauge, a[:])
+}
+
+func NewAddressValueFromBytes(memoryGauge common.MemoryGauge, b []byte) AddressValue {
```
turbolent, line 15176: The last parameter should be lazily evaluated, e.g. by passing it as a function, so that the metering happens before the actual allocation

robert-e-davidson3: I should do this for all of the variable types, right?
- TypeValue
- CharacterValue
- StorageReferenceValue
- PathValue
- AddressValue

runtime/interpreter/interpreter.go
robert-e-davidson3: This turned out to be due to a test using `nil` instead of an interpreter. AFAICT there's no reason to omit the interpreter there so I added it back.

runtime/interpreter/value.go
robert-e-davidson3: Done.

turbolent (approved):
Looks good, but metering of type value is tricky. Maybe split up the PR and submit the type value metering separately, so we can already get the metering for the other values in

dsainati1 (commented):
Can you add some tests to `memory_metering_test.go`? Testing the imported values is good but I think we want to also test the metering for values created in programs.

runtime/convertValues.go
```diff
@@ -591,11 +591,38 @@ func importString(inter *interpreter.Interpreter, v cadence.String) *interpreter
...
+		inter,
+		memoryUsage,
+		func() string {
+			return s
+		},
+	)
+}
+
+func importAddress(inter *interpreter.Interpreter, v cadence.Address) interpreter.AddressValue {
+	memoryUsage := common.NewAddressMemoryUsage(len(v))
```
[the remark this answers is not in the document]
robert-e-davidson3: Yup!

runtime/interpreter/interpreter_expression.go
```diff
@@ -951,8 +956,11 @@ func (interpreter *Interpreter) VisitForceExpression(expression *ast.ForceExpres
...
 
-	return PathValue{
-		Domain:     domain,
-		Identifier: expression.Identifier.Identifier,
-	}
+	return NewPathValue(
+		interpreter,
+		common.NewPathMemoryUsage(expression.Identifier.Identifier),
+		func() (common.PathDomain, string) {
+			return domain, expression.Identifier.Identifier
```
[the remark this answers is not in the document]
robert-e-davidson3: Good point!

Looking into this, I notice another problem: the decoder is metering the Identifier string but importValue isn't.

I'm going to meter the Identifier separately and meter Path as a constant because then it'll just be a container.

runtime/interpreter/decode.go
```diff
@@ -134,14 +134,14 @@ func (d StorableDecoder) decodeStorable() (atree.Storable, error) {
 		if err != nil {
 			return nil, err
 		}
-		storable = BoolValue(v)
+		storable = NewBoolValue(d.memoryGauge, v)
```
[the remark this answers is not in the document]
robert-e-davidson3: This turned out to be kind of complicated for AddressValue but maybe I fixed a potential bug? I switched it from grabbing all bytes at once to using the stream's NextSize() call first.

runtime/interpreter/value.go
```diff
@@ -423,6 +450,21 @@ var _ Value = VoidValue{}
...
+	return VoidValue{}
+}
+
+func NewVoidValue(memoryGauge common.MemoryGauge) VoidValue {
+	if memoryGauge != nil {
+		memoryGauge.UseMemory(common.MemoryUsage{
+			Kind:   common.MemoryKindVoid,
+			Amount: 1,
+		})
+	}
```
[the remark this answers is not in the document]
robert-e-davidson3: Passing an `Interpreter` here caused issues upstream so I turned `UseConstantMemory` into a function. It works fine against any kind of `MemoryGauge` so it doesn't benefit from being a method.

```

//
func UseConstantMemory(memoryGauge MemoryGauge, kind MemoryKind) {
	if memoryGauge == nil {
		return
	}
		Kind:   kind,
		Amount: 1,
	})
}
```

runtime/interpreter/interpreter.go
```diff
@@ -3378,9 +3379,12 @@ var typeFunction = NewUnmeteredHostFunctionValue(
...
 		ty := typeParameterPair.Value
 
-		return TypeValue{
-			Type: ConvertSemaToStaticType(ty),
-		}
+		staticType := ConvertSemaToStaticType(ty)
+		return NewTypeValue(
+			invocation.Interpreter,
+			common.NewTypeMemoryUsage(staticType.String()),
+			func() StaticType { return staticType },
```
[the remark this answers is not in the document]
robert-e-davidson3: At Bastian's suggestion, I'm going to leave the `TypeValue` metering in a half-done state because it's complex. Ticket for that: https://github.com/onflow/cadence/issues/1503.

runtime/convertValues.go
```diff
@@ -612,16 +639,19 @@ func importTypeValue(
 	   import is invalid */
```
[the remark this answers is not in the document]
robert-e-davidson3: Issue for finishing `TypeValue`: https://github.com/onflow/cadence/issues/1503

runtime/convertValues.go
```diff
@@ -612,16 +639,19 @@ func importTypeValue(
...
+		// unmetered because when err != nil, value should be ignored
+		return interpreter.EmptyTypeValue, err
 	}
 
-	return interpreter.TypeValue{
-		Type: typ,
-	}, nil
+	return interpreter.NewTypeValue(
+		inter,
+		common.NewTypeMemoryUsage(typ.String()),
```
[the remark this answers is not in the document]
robert-e-davidson3: Issue for finishing `TypeValue`: https://github.com/onflow/cadence/issues/1503

dsainati1 (approved):
Added tests look good to me!

runtime/interpreter/decode.go
```diff
@@ -973,12 +1004,18 @@ func (d StorableDecoder) decodeType() (TypeValue, error) {
...
-	}, nil
+	if staticType == nil {
+		return EmptyTypeValue, nil
+	}
+
+	return NewTypeValue(
+		d.memoryGauge,
+		common.NewTypeMemoryUsage(staticType.String()),
+		func() StaticType { return staticType },
+	), nil
```
SupunS, line 1018: Similarly, let's use the unmetered constructor, so that we can introduce TypeValue related changes all at once.

runtime/interpreter/interpreter.go
```diff
@@ -1906,7 +1903,7 @@ func EnumConstructorFunction(
 
 			caseValue, ok := lookupTable[string(rawValueArgumentBigEndianBytes)]
 			if !ok {
-				return NilValue{}
+				return NewNilValue(inter.memoryGauge)
```
SupunS, line 1906:
```suggestion
				return NewNilValue(inter)
```

runtime/interpreter/interpreter.go
```diff
@@ -3981,17 +3990,15 @@ func (interpreter *Interpreter) authAccountLinkFunction(addressValue AddressValu
...
 			// Write new value
 
 			borrowStaticType := ConvertSemaToStaticType(borrowType)
 
-			linkValue := LinkValue{
-				TargetPath: targetPath,
-				Type:       borrowStaticType,
-			}
+			// Unmetered here because interpreter.writeStored(...) meters it
+			linkValue := NewUnmeteredLinkValue(targetPath, borrowStaticType)
```
SupunS, line 4001: IINM `writeStored` doesn't meter upon storing. It only meter when a value is read-back from storage (decoding).

runtime/interpreter/interpreter_expression.go
```diff
@@ -443,7 +443,11 @@ func (interpreter *Interpreter) VisitIntegerExpression(expression *ast.IntegerEx
 	value := expression.Value
 
 	if _, ok := typ.(*sema.AddressType); ok {
-		return NewAddressValueFromBytes(value.Bytes())
+		bytes := value.Bytes()
+		return NewAddressValueFromBytes(
+			interpreter,
+			bytes,
+		)
```
SupunS, line 450: `value.Bytes()` allocates new memory. Must delay call to it, and make the constructor `NewAddressValueFromBytes` accept an anon-func.
```suggestion
		return NewAddressValueFromBytes(
			interpreter,
			value.Bytes,
		)
```

runtime/interpreter/value.go
```diff
@@ -16533,27 +16652,40 @@ func (*EphemeralReferenceValue) DeepRemove(_ *Interpreter) {
...
 	return result
 }
 
+func NewAddressValue(
+	memoryGauge common.MemoryGauge,
+	address common.Address,
+) AddressValue {
+	return NewAddressValueFromBytes(
+		memoryGauge,
+		address[:],
```
SupunS, line 16667: Same here, go internally allocates a new slice for `address[:]`. Need to pass this as an anon-func

runtime/interpreter/interpreter_expression.go
```diff
@@ -915,27 +919,24 @@ func (interpreter *Interpreter) VisitReferenceExpression(referenceExpression *as
...
-				&EphemeralReferenceValue{
-					Authorized:   innerBorrowType.Authorized,
-					Value:        result.InnerValue(interpreter, getLocationRange),
-					BorrowedType: innerBorrowType.Type,
-				},
+				NewEphemeralReferenceValue(
+					interpreter,
+					innerBorrowType.Authorized,
+					result.InnerValue(interpreter, getLocationRange),
+					innerBorrowType.Type),
```
SupunS, line 926:
```suggestion
					innerBorrowType.Type,
				),
```

runtime/interpreter/value.go
```diff
@@ -16709,7 +16841,7 @@ func (v AddressValue) Transfer(
 }
 
 func (v AddressValue) Clone(_ *Interpreter) Value {
-	return v
+	return NewUnmeteredAddressValue(v[:])
```
SupunS, line 16844: leaving this as is should be ok. so we won't allocate extra memory (new slices) during internal operation such as migrations, etc.

runtime/tests/interpreter/container_mutation_test.go
```diff
@@ -895,7 +895,10 @@ func TestDictionaryMutation(t *testing.T) {
 
 		owner := newTestPublicAccountValue(
 			inter,
-			interpreter.NewAddressValue(common.Address{0x1}),
+			interpreter.NewAddressValue(
+				inter,
+				common.Address{0x1},
+			),
```
SupunS, line 901:
```suggestion
			interpreter.NewUnmeteredAddressValue(common.Address{0x1}),
```

SupunS (commented):
Most of it Looks good! Just few suggestions/questions before merging

runtime/imported_values_memory_metering_test.go
```diff
@@ -427,3 +429,192 @@ func TestMemoryMeteringErrors(t *testing.T) {
 		assert.ErrorIs(t, err, testMemoryError{})
 	})
 }
+
+func TestImportedValueMemoryMeteringIterate(t *testing.T) {
```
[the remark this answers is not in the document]
robert-e-davidson3: Oops that was a temporary name since the approach to test-writing was very different. It's iterating over an array of tests...

I'll give it a better name.

runtime/interpreter/decode.go
```diff
@@ -712,8 +740,12 @@ func (d StorableDecoder) decodeSome() (SomeStorable, error) {
 	}, nil
 }
 
-func checkEncodedAddressLength(addressBytes []byte) error {
-	actualLength := len(addressBytes)
+func checkEncodedAddressLength(decoder *cbor.StreamDecoder) error {
+	actualLength, err := decoder.NextSize()
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
robert-e-davidson3: If it doesn't then it may need to be extended. We test for Address length so I assume it's possible for the length to be arbitrarily large. We can't meter effectively without knowing the length ahead of time.

runtime/interpreter/value.go
robert-e-davidson3: It's weird to me that that our `Clone` methods return the original most of the time. For e.g. `BoolValue` that doesn't matter because changing the returned value won't change the original. But for slices like `AddressValue`, changes to the copy affect the original.

I suppose this is a naming argument. This isn't the right PR for that argument though so I'll revert it.

runtime/interpreter/value.go
robert-e-davidson3: I think it's fine.

This call should only happen when the address is at most 8 bytes long so we can treat it as a constant. Constants use a known, smallish amount of memory, so it's safe to allocate them ahead of time.

robert-e-davidson3: Updated to allocate first.

SupunS: Actually, the suggestion here was to do the metering first and then the allocation. That is what we agreed earlier: https://github.com/dapperlabs/cadence-private-issues/issues/2#issuecomment-1065622097

runtime/interpreter/decode.go
```diff
@@ -973,12 +998,14 @@ func (d StorableDecoder) decodeType() (TypeValue, error) {
...
-		return TypeValue{}, fmt.Errorf("invalid type encoding: %w", err)
+		return EmptyTypeValue, fmt.Errorf("invalid type encoding: %w", err)
 	}
 
-	return TypeValue{
-		Type: staticType,
-	}, nil
+	if staticType == nil {
+		return EmptyTypeValue, nil
+	}
```
SupunS, line 1006: This condition is sort of redundant, It's same as calling the `NewUnmeteredTypeValue` constructor I suppose.

SupunS (commented):
There are still some unaddressed/unresolved comments. We should try to get all suggestions addressed before merging a PR, and probably get another round of quick review after the changes.

Can you please open a follow-up PR addressing them?

runtime/interpreter/interpreter.go
robert-e-davidson3: That's true. `writeStored` here eventually calls `interpreter.ValidateAtreeValue`, which eventually calls `decodeLink`.

How's this?

 	// This is unmetered because this value won't be exposed to the user. It's internal.
	// The value returned to the user is metered because it goes through `decodeLink(...)`,
	// which is called by `writeStored(...)` below, after going through `ValidateAtreeValue(...)`.

SupunS: The `ValidateAtreeValue()` is hidden behind the`atreeValueValidationEnabled` flag, and is not always enabled. e.g: it is not enabled on mainnet due to performance reasons. 
On the other hand, at the time of decoding, the decoder creates another `LinkedValue` instance, so there are in fact two values in memory. So it's ok/correct to meter it here, regardless of whether it was metered again at atree validation (`decodeLink()`)

runtime/interpreter/decode.go
```diff
@@ -737,12 +774,10 @@ func (d StorableDecoder) decodeAddress() (AddressValue, error) {
 		return AddressValue{}, err
 	}
 
-	err = checkEncodedAddressLength(addressBytes)
-	if err != nil {
-		return AddressValue{}, err
-	}
-
-	address := NewAddressValueFromBytes(addressBytes)
+	address := NewAddressValueFromBytes(
```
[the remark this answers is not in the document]
robert-e-davidson3: https://github.com/onflow/cadence/pull/1512

runtime/interpreter/interpreter.go
robert-e-davidson3: https://github.com/onflow/cadence/pull/1512

runtime/interpreter/interpreter.go
robert-e-davidson3: https://github.com/onflow/cadence/pull/1512

runtime/interpreter/decode.go
robert-e-davidson3: https://github.com/onflow/cadence/pull/1512

PR #1450, opened 2022-02-25
Expand interpreter tracing for composite, array and dictionary types

ramtinms:
## Description
This PR expands tracing functionality inside cadence for composite, array and dictionary types. these open traces are disabled by default and upon enabling it, they would be passed to the interface's recordTrace method for further aggregation. These traces are great tools for analyzing the performance of the interpreter when handling these types.

______

### DIFF
--- a/runtime/interpreter/interpreter_tracing_test.go
+++ b/runtime/interpreter/interpreter_tracing_test.go
@@ -1,0 +1,91 @@
+/*
+ * Cadence - The resource-oriented smart contract programming language
+ *
+ * Copyright 2022 Dapper Labs, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package interpreter_test
+
+import (
+	"testing"
+	"time"
+
+	"github.com/onflow/cadence/runtime/common"
+	"github.com/onflow/cadence/runtime/interpreter"
+	"github.com/onflow/cadence/runtime/sema"
+	"github.com/onflow/cadence/runtime/tests/utils"
+	"github.com/opentracing/opentracing-go"
+	"github.com/stretchr/testify/require"
+)
+
+func TestInterpreterTracing(t *testing.T) {
+
+	t.Parallel()
+
+	t.Run("composite tracing", func(t *testing.T) {
+		storage := interpreter.NewInMemoryStorage()
+
+		elaboration := sema.NewElaboration()
+		elaboration.CompositeTypes[testCompositeValueType.ID()] = testCompositeValueType
+
+		traceOps := make([]string, 0)
+		inter, err := interpreter.NewInterpreter(
+			&interpreter.Program{
+				Elaboration: elaboration,
+			},
+			utils.TestLocation,
+			interpreter.WithOnRecordTraceHandler(
+				func(inter *interpreter.Interpreter,
+					operationName string,
+					duration time.Duration,
+					logs []opentracing.LogRecord) {
+					traceOps = append(traceOps, operationName)
+				},
+			),
+			interpreter.WithStorage(storage),
+			interpreter.WithTracingEnabled(true),
+		)
+		require.NoError(t, err)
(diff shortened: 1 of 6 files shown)


turbolent:
Just realized: Tracing is disabled by default, so tests will not exercise any of this code. 

Could you please add this after https://github.com/onflow/cadence/blob/master/runtime/tests/interpreter/interpreter_test.go#L97-L97:

```go
			interpreter.WithOnRecordTraceHandler(
				func(
					_ *interpreter.Interpreter,
					_ string,
					_ time.Duration,
					_ []opentracing.LogRecord,
				) {
					// NO-OP
				},
			),
			interpreter.WithTracingEnabled(true),
```

turbolent:
Please follow and fill out the checklist, thank you

runtime/interpreter/interpreter_tracing_test.go
```diff
@@ -0,0 +1,203 @@
...
+					operationName string,
+					duration time.Duration,
+					logs []opentracing.LogRecord) {
+					traceOps = append(traceOps, operationName)
+				},
+			),
+			interpreter.WithStorage(storage),
+			interpreter.WithTracingEnabled(true),
+		)
+		require.NoError(t, err)
```
turbolent, line 57: This setup code is repeated in each test. Consider defining a helper function for it

runtime/interpreter/value.go
```diff
@@ -11515,6 +11553,13 @@ func NewCompositeValue(
 	address common.Address,
 ) *CompositeValue {
 
+	if interpreter.tracingEnabled {
+		startTime := time.Now()
+		defer func() {
+			interpreter.reportCompositeValueConstructTrace(address.String(), qualifiedIdentifier, kind.String(), time.Since(startTime))
```
turbolent, line 11559: Please chop down the argument list for 3+ arguments (here and in the other cases below):

```suggestion
			interpreter.reportCompositeValueConstructTrace(
				address.String(), 
				qualifiedIdentifier, 
				kind.String(), 
				time.Since(startTime),
			)
```

turbolent (commented):
Looks good!

runtime/interpreter/value.go
ramtinms: updated all calls

runtime/interpreter/interpreter_tracing_test.go
ramtinms: added a helper function

runtime/interpreter/interpreter_tracing_test.go
```diff
@@ -28,34 +28,41 @@ import (
 
 	"github.com/onflow/cadence/runtime/common"
 	"github.com/onflow/cadence/runtime/interpreter"
-	"github.com/onflow/cadence/runtime/sema"
 	"github.com/onflow/cadence/runtime/tests/utils"
 )
 
+func setupInterpreterWithTracingCallBack(
```
turbolent, line 34: 👌

turbolent (approved):
Looks good, nice work! 

Seems like some tests fail.

Should we also port this to 0.23? (the next Cadence release)

turbolent:
@ramtinms I looked at the failing tests. `Transfer` functions are bit tricky, OK if I fix the tracing calls?
Basically, the tracing is using a defer, and by the time the trace is reported, the value has already changed

ramtinms:
its not a must-have for 0.23, but would make our life easier if its backported.

turbolent:
@ramtinms In f9d8c816bbf0287fc1ee788df58fba26f9061b7e I:
- Removed the tracing from `Clone` calls. This is an internal function which is never called by the interpreter, it is only used by e.g. a state migration
- Unified how the tracing information is prepared
- Moved the preparation of the tracing information before the deferal (when possible), so it does not access potentially invalidated information/state (e.g. `Transfer` sets `ArrayValue`/`DictionaryValue`/`CompositeValue`'s `array`/`dictionary` `nil` if it is a resource

ramtinms:
changes looks great, thanks for taking care of it

turbolent (approved):

turbolent:
@SupunS Could you please have a second look?

runtime/interpreter/value.go
```diff
@@ -12186,6 +12385,23 @@ func (v *CompositeValue) Transfer(
 		v.checkInvalidatedResourceUse(getLocationRange)
 	}
 
+	if interpreter.tracingEnabled && v.dictionary != nil {
```
SupunS, line 12388: I think its safe to assume that `v.dictionary != nil` is already checked in the `checkInvalidatedResourceUse` check.

SupunS (approved):
Looks good to me!

runtime/interpreter/value.go
turbolent: Ah, yes, that was temporary debugging code 👍 
```suggestion
	if interpreter.tracingEnabled {
```

PR #1461, opened 2022-03-01
Check transfer and use of transfered or destroyed resources at runtime

turbolent:
Work towards dapperlabs/cadence-private-issues#19

## Description

Add a defensive check for resources: When a resource is transferred (moved), create a new instance (Go struct), instead of reusing the existing one. Also, mark the transferred resource as invalidated, and detect uses after invalidation.

We already had an existing check for uses of referenced resources after their destruction.

Combined, this allows checking the use (transfer or access) of invalidated (transferred or destroyed) resources.

______


SupunS (approved):
🎉

PR #1469, opened 2022-03-02
Moving computation metering and limit enforcement responsibility to the environment

ramtinms:
## Description

This PR
- moves the computation metering call to the interface, so FVM would enforce the limits going forward, this would enable tracking some heavy operations on the FVM side such as crypto calls as well. 
- updates the interpreter to report computations directly instead of through the onStatement, onLoop, etc calls, so we could cover more kinds of computation reporting 
- adds new computation kinds (array/dictionary/composite value create/transfer/destroy) 
______


janezpodhostnik (approved):
This looks good to me.

runtime/runtime.go
```diff
@@ -1578,32 +1543,45 @@ func (r *interpreterRuntime) meteringInterpreterOptions(runtimeInterface Interfa
 	return []interpreter.Option{
 		interpreter.WithOnStatementHandler(
 			func(_ *interpreter.Interpreter, _ ast.Statement) {
-				checkComputationLimit(1)
+				var err error
+				wrapPanic(func() {
+					err = runtimeInterface.MeterComputation(OpTypeStatement, 1)
```
dsainati1, line 1548: Is there every going to be a situation in which we meter more than one of a computation type at a time? If not IMO we should get rid of the numeric argument here.

ramtinms: for some other types like CopyOfArray would be capturing the size of input instead of 1.

runtime/metering.go
```diff
@@ -0,0 +1,59 @@
+package runtime
```
turbolent, line 1: Need the copyright header

runtime/metering.go
```diff
@@ -0,0 +1,59 @@
...
+
+type MetringOperationType uint
```
turbolent, line 3: Both "Operation" and "Type" seem redundant, maybe just `ComputationOperation`?

runtime/metering.go
```diff
@@ -0,0 +1,59 @@
...
+
+const (
+	// base [0-10)
```
turbolent, line 6: Is there any reason for leaving gaps? What if in the future the gaps are filled but more elements need to be added to a "category"?

runtime/metering.go
```diff
@@ -0,0 +1,59 @@
...
+	OpTypeStatement MetringOperationType = iota
```
turbolent, line 7: Prefix with the type name, e.g. 

```suggestion
	ComputationOperationStatement ComputationOperation = iota
```

runtime/metering.go
```diff
@@ -0,0 +1,59 @@
...
+	_
+	_
+	_
+	_
+	_
+	_
+
+	// value operations [10-40)
+	OpTypeCompositeValueCreate
+	OpTypeCompositeValueCopy
```
turbolent, line 20: Transfer is copy/move. Probably not needed

runtime/runtime_test.go
```diff
@@ -5282,12 +5285,24 @@ func TestRuntimeComputationLimit(t *testing.T) {
...
+
+			checkComputationLimit := func(_ MetringOperationType, intensity uint) error {
+				computationUsed += intensity
+				if computationUsed <= computationLimit {
+					return nil
+				}
+				panic(ComputationLimitExceededError{
+					Limit: computationLimit,
+				})
+			}
```
turbolent, line 5298: This should be moved to the FVM now, correct?

runtime/metering.go
ramtinms: yeah you're right, changed it

runtime/metering.go
ramtinms: removed

runtime/runtime_test.go
ramtinms: yeah, just kept it here for testing that there is no side effect due to the change until we add proper tests on the FVM side, then we could remove these

runtime/metering.go
ramtinms: renamed all to match how memoryKinds are defined

runtime/metering.go
ramtinms: added

runtime/metering.go
ramtinms: yeah, renamed to the way memoryKind is defined

runtime/interface.go
```diff
@@ -84,10 +84,9 @@ type Interface interface {
 	EmitEvent(cadence.Event) error
 	// GenerateUUID is called to generate a UUID.
 	GenerateUUID() (uint64, error)
-	// GetComputationLimit returns the computation limit. A value <= 0 means there is no limit
-	GetComputationLimit() uint64
-	// SetComputationUsed reports the amount of computation used.
-	SetComputationUsed(used uint64) error
+	// MeterComputation is a callback method for metering computation, it returns error
+	// when computation passes the limit (set by the environment)
+	MeterComputation(operationType common.ComputationKind, intensity uint) error
```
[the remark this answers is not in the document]
ramtinms: hmm, was thinking most of the time this number might be really small, so allocating uint64 might be unnecessary, but we could change it in the FVM side whenever needed, thoughts?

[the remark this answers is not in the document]
janezpodhostnik: My reasoning was that the final value for execution effort (computation) would actually be a UFix64 so that it can be sent to the FlowFees smart contract. 

But like you have pointed out the conversion can happen in the FVM

runtime/interpreter/interpreter.go
```diff
@@ -425,6 +435,16 @@ func WithOnResourceOwnerChangeHandler(handler OnResourceOwnerChangeFunc) Option
 	}
 }
 
+// WithOnMeterComputationFuncHandler returns an interpreter option which sets
+// the given function as the meter computation handler.
+//
+func WithOnMeterComputationFuncHandler(handler OnMeterComputationFunc) Option {
```
turbolent, line 441: Needs to be propagated in `NewSubInterpreter`

runtime/interpreter/value.go
```diff
@@ -1944,6 +1948,7 @@ func (v *ArrayValue) Transfer(
 	remove bool,
 	storable atree.Storable,
 ) Value {
+	interpreter.ReportComputation(common.ComputationKindTransferArrayValue, uint(v.Count()))
```
turbolent, line 1951: `Count` uses `v.array`, which may be nil if the array is a resource and got invalidated. Move this line after the if statement below

turbolent: Maybe also double-check the other functions where `ReportComputation` is called and ensure it is after `checkInvalidatedResourceUse`

runtime/runtime_test.go
```diff
@@ -5215,109 +5214,6 @@ func TestRuntimeResourceOwnerFieldUseDictionary(t *testing.T) {
 	)
 }
 
-func TestRuntimeComputationLimit(t *testing.T) {
```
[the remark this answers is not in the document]
ramtinms: I planned to move it to the FVM side, but you're right I'll update this and keep it.

runtime/interpreter/interpreter.go
ramtinms: nice catch

runtime/interpreter/value.go
ramtinms: updated at 7f84226

runtime/interpreter/interpreter.go
ramtinms: updated at 7f84226

runtime/runtime_test.go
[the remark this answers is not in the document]
ramtinms: updated test at 4c958b1

turbolent (approved):
Nice work!

PR #1480, opened 2022-03-04
Fix value transfer type check

### LINKED ISSUE #1479: Cannot cast &FungibleToken.Vault to &FungibleToken.Vault{FungibleToken.Balance}
### Problem

When trying to cast a `&FungibleToken.Vault` type to a `&FungibleToken.Vault{FungibleToken.Balance}` type, you get this weird error: "Error invalid transfer of value: expected &FungibleToken.Vault{FungibleToken.Balance}". 

The same is true for other restricted types, like `&FungibleToken.Vault{FungibleToken.Receiver}` and `&FungibleToken.Vault{FungibleToken.Provider}`. I believe this is also a bug for &NonFungibleToken.Collection as well.

I have reproduced this error in both the playground and on the current version of TestNet. 

image

### Steps to Reproduce

https://play.onflow.org/0d50d978-20eb-4328-b75c-4aff5d42240f?type=account&id=f7be17c7-953e-4728-ac7b-e23337d5e215&storage=none

If you run the transaction in the playground above, you should see the error at hand.

turbolent:
Closes #1479 

## Description

When checking `T <: U{Vs}`, we currently require `T == U`, assuming both are composite values. 
However, the latter may actually be a type requirement.

Extend the subtyping check and require `T <: U`.

______

### DIFF
--- a/runtime/tests/interpreter/transfer_test.go
+++ b/runtime/tests/interpreter/transfer_test.go
@@ -34,52 +34,98 @@ func TestInterpretTransferCheck(t *testing.T) {
 
 	t.Parallel()
 
-	ty := &sema.CompositeType{
-		Location:   utils.TestLocation,
-		Identifier: "Fruit",
-		Kind:       common.CompositeKindStructure,
-	}
+	t.Run("String value as composite", func(t *testing.T) {
 
-	valueDeclarations := stdlib.StandardLibraryValues{
-		{
-			Name: "fruit",
-			Type: ty,
-			// NOTE: not an instance of the type
-			ValueFactory: func(_ *interpreter.Interpreter) interpreter.Value {
-				return interpreter.NewStringValue("fruit")
+		t.Parallel()
+
+		ty := &sema.CompositeType{
+			Location:   utils.TestLocation,
+			Identifier: "Fruit",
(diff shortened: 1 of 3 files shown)


turbolent:
Added another test case which shows that this is not just a problem for references, but an issue with composite dynamic type and restricted static type.

turbolent:
Added yet another check which shows that this is actually a limitation of the type checker. The dynamic subtype check in the interpreter relies on this check to succeed

runtime/tests/interpreter/transfer_test.go
```diff
@@ -34,52 +34,139 @@ func TestInterpretTransferCheck(t *testing.T) {
...
+		)
+		require.NoError(t, err)
+
+		_, err = inter.Invoke("test")
+		require.Error(t, err)
+
+		require.ErrorAs(t, err, &interpreter.ValueTransferTypeError{})
+	})
+
+	t.Run("contract and restricted type", func(t *testing.T) {
```
turbolent, line 91: These tests passed the type checking (correctly) and only failed at run-time due to the defensive value transfer check, which in addition to the static check, also checks subtyping for the underlying value.

dsainati1 (approved):

SupunS (approved):

PR #1481, opened 2022-03-04
Add a custom linter which detects Value composite literals

turbolent:

## Description

Add a new linter which detects composite literals for `Value`s.

It is not enabled by default and can be run using: `make lint linters=constructorcheck`

______


robert-e-davidson3 (approved):

dsainati1 (approved):
Looks good to me.

I assume that like the other linters this will not run on test files, but for this one I think we should also disable it in `value.go` so that we don't need to add lint ignore comments all over the place.

turbolent:
@dsainati1 yes, the linter is disabled for test files: https://github.com/onflow/cadence/pull/1481/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9R28

> I think we should also disable it in `value.go` 
The linter is not enabled (yet) by default, but once we are done with the metering, there should only be very few cases where we, because we should always use the metering/non-metering constructor functions instead of constructing values using a composite literal.

PR #1482, opened 2022-03-04
Add test-cases for container value iteration metering

SupunS:
Work towards https://github.com/dapperlabs/cadence-private-issues/issues/2

## Description

Container values created when converting atree-value to cadence value (`ConvertStoredValue`) is not metered at the moment.

This is due to the technical complexity, as `ConvertStoredValue` method is used in various places directly/indirectly, where the interpreter instance is not available. e.g: Iterator(), walk(), RecursiveString(), String() methods of values, etc.

We should probably meter these as well, but asserting the current behavior for the moment.

______

- [ ] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #1484, opened 2022-03-07
Meter values created during atree value conversion 

dsainati1:

## Description

As described in https://github.com/onflow/cadence/pull/1482, we were previously not metering these values due to the effort involved. This adds the metering for the creation of these values.

This is mostly threading the interpreter through in more places than before, with one exception. During the `RecursiveString` method on complex values, we iterate through these values, which can call the `ConvertStoredValue` function. We should not need to meter values created here, as they will be discarded immediately once the string is computed. 
______


runtime/interpreter/storage.go
```diff
@@ -33,31 +33,42 @@ import (
...
-func ConvertStoredValue(value atree.Value) (Value, error) {
+func MustConvertUnmeteredStoredValue(value atree.Value) Value {
+	converted, err := ConvertStoredValue(nil, value)
+	if err != nil {
+		panic(err)
+	}
+	return converted
+}
+
+func ConvertStoredValue(gauge common.MemoryGauge, value atree.Value) (Value, error) {
```
dsainati1, line 61: This PR adds metering for the array, dictionary, and composite cases in this function, but we don't do anything for any other values types. If values are being loaded from storage (hence the atree), do we need to "retroactively" meter them once they move from storage into the memory of the program?

runtime/interpreter/storage.go
```diff
@@ -33,31 +33,42 @@ import (
...
 	// TODO: Meter container values created below.
 	//   This is currently technically challenging because, this method is used in various
 	//   places directly/indirectly, where the interpreter instance is not available.
 	//   e.g: Iterator(), value walk(), RecursiveString(), String() method of values, etc.
```
SupunS, line 65: I think we can remove this TODO now?

runtime/interpreter/storage.go
SupunS: Its already handled by the decoder IINM
e.g:
https://github.com/onflow/cadence/blob/b5ebaced3696485167163b6dcd80335e2df2700d/runtime/interpreter/decode.go#L68-L81

runtime/interpreter/value.go
```diff
@@ -1549,9 +1549,13 @@ func (v *ArrayValue) RecursiveString(seenReferences SeenReferences) string {
...
 	i := 0
-	v.Walk(func(value Value) {
-		values[i] = value.RecursiveString(seenReferences)
+
+	_ = v.array.Iterate(func(element atree.Value) (resume bool, err error) {
+		// ok to not meter anything created as part of this iteration, since we will discard the result
+		// upon creating the string
+		values[i] = MustConvertUnmeteredStoredValue(element).RecursiveString(seenReferences)
 		i++
+		return true, nil
```
SupunS, line 1558: makes sense 👍

runtime/tests/interpreter/memory_metering_test.go
```diff
@@ -93,10 +93,7 @@ func TestRuntimeArrayMetering(t *testing.T) {
 		_, err := inter.Invoke("main")
 		require.NoError(t, err)
 
-		// TODO:
-		// Iteration create new values for array typed elements.
-		// Currently, these are not metered. Only the initial 4-values are captured.
-		assert.Equal(t, uint64(4), meter.getMemory(common.MemoryKindArray))
+		assert.Equal(t, uint64(16), meter.getMemory(common.MemoryKindArray))
```
SupunS, line 96: Oh wow, that increased the values drastically. I was expecting it to be 10 (4-initial values, 3 for iteration, and 3 for assignment). Looking at the above test, it seems the allocations on array-literal also has increased.

Maybe good to add comment with a break-down on where the allocations happened?

runtime/interpreter/storage.go
```diff
@@ -33,31 +33,42 @@ import (
...
 
 	switch value := value.(type) {
 	case *atree.Array:
+		if gauge != nil {
+			gauge.UseMemory(common.NewConstantMemoryUsage(common.MemoryKindArray))
+		}
 		return &ArrayValue{
 			Type:  value.Type().(ArrayStaticType),
 			array: value,
```
turbolent, line 74: Maybe we should have constructors (private?) for the container values that are constructed from their underlying atree values.

For example here, it might make sense to add a constructor function `func newArrayValueFromAtreeValue(array *atree.Array, memoryGauge common.MemoryGauge)` and call it here.

`NewArrayValueWithIterator` would also call this function at the end of its body. The metering in `NewArrayValueWithIterator` would also move to the new `newArrayValueFromAtreeValue`.

That way we only meter in one place, the constructor

runtime/common/metering.go
```diff
@@ -27,6 +27,13 @@ type MemoryGauge interface {
 	UseMemory(usage MemoryUsage)
 }
 
+func NewConstantMemoryUsage(kind MemoryKind) MemoryUsage {
```
robert-e-davidson3, line 30: Here's my approach:

```

//
func UseConstantMemory(memoryGauge MemoryGauge, kind MemoryKind) {
	if memoryGauge == nil {
		return
	}
		Kind:   kind,
		Amount: 1,
	})
}
```

runtime/interpreter/storage.go
```diff
@@ -33,31 +33,42 @@ import (
...
 	// TODO: Meter container values created below.
 	//   This is currently technically challenging because, this method is used in various
 	//   places directly/indirectly, where the interpreter instance is not available.
 	//   e.g: Iterator(), value walk(), RecursiveString(), String() method of values, etc.
 
 	switch value := value.(type) {
 	case *atree.Array:
+		if gauge != nil {
+			gauge.UseMemory(common.NewConstantMemoryUsage(common.MemoryKindArray))
+		}
```
robert-e-davidson3, line 71: From my above comment: my approach would look like this here:

```
common.UseConstantMemory(gauge, common.MemoryKindArray)
```

runtime/interpreter/value.go
```diff
@@ -1549,9 +1549,13 @@ func (v *ArrayValue) RecursiveString(seenReferences SeenReferences) string {
 	values := make([]string, v.Count())
 
 	i := 0
-	v.Walk(func(value Value) {
-		values[i] = value.RecursiveString(seenReferences)
+
+	_ = v.array.Iterate(func(element atree.Value) (resume bool, err error) {
+		// ok to not meter anything created as part of this iteration, since we will discard the result
+		// upon creating the string
```
robert-e-davidson3, line 1555: Is there a risk that Go's garbage collection won't catch this quickly enough?

My concern is that the memory is still being allocated even if it will (eventually) be deallocated so a well-crafted transaction could trigger this code path repeatedly. But that's just a guess since I don't know how this code is used now how Go actually does garbage collection.

runtime/tests/interpreter/memory_metering_test.go
SupunS: Thanks @dsainati1 for the complex Analysis. This is very useful!

Since it's the interpreter-implementation that allocates more memory than what's probably needed, and the memory metering correctly captures them, I would think it's close to actual memory usage, and should be fine. (Given our goal is to be as close as possible to actual usage.

But definitely we should optimize these, maybe as a follow-up task after secure cadence. Couple items I could think of inspired from these findings are:
- DynamicType checking - This one have been under radar for a while. I think https://github.com/onflow/cadence/issues/1217 (this guy again!) would actually solve this. See the `CheckValueTransferTargetType` alloc improvements on the benchmark results.
- Perhaps we don't need to transfer/copy for literals. Can skip, as an optimization.

runtime/interpreter/value.go
turbolent: We still have to meter the amount of storage used before allocating though. The stored value could e.g. be a very large string value. 

`MustConvertUnmeteredStoredValue` should be `MustConvertStoredValue` with a memory gauge and pass it to `ConvertStoredValue`

dsainati1: The issue with this is that this would require the `String` method on all values to take an `Interpreter` argument. Is this something we want?

SupunS: What do you all think of creating an issue for this and merging this PR in the current state? Since it will be an addition to the existing metering, we can get back to it once we have a conclusion.

SupunS (approved):
Changes look good to me!

For the string conversion metering, we can create an issue (like I suggested here: https://github.com/onflow/cadence/pull/1484#discussion_r826083750) and add it separately, so we can get these changes in.

PR #1485, opened 2022-03-07
String metering is length + 1 to account for empty strings

robert-e-davidson3:
Related to https://github.com/dapperlabs/cadence-private-issues/issues/2.

## Description

An array of a million empty strings would be metered as if it were an empty array. This patch fixes that by metering every string as `length + 1` instead of simply `length`.

______

- [X] Targeted PR against `feature/memory-metering` branch


dsainati1 (approved):
Makes sense to me

SupunS (approved):

PR #1490, opened 2022-03-09
Meter memory for IntValue and UIntValue creation, addition, multiplication, and negation

turbolent:

______

### DIFF
--- a/runtime/common/metering.go
+++ b/runtime/common/metering.go
@@ -33,3 +33,10 @@ func NewStringMemoryUsage(length int) MemoryUsage {
 		Amount: uint64(length),
 	}
 }
+
+func NewBigIntMemoryUsage(bytes int) MemoryUsage {
+	return MemoryUsage{
+		Kind:   MemoryKindBigInt,
+		Amount: uint64(bytes),
+	}
+}
--- a/runtime/common/metering.go
+++ b/runtime/common/metering.go
@@ -18,6 +18,11 @@
 
 package common
 
+import (
+	"math/big"
+	"unsafe"
+)
+
 type MemoryUsage struct {
 	Kind   MemoryKind
 	Amount uint64
@@ -39,4 +44,29 @@ func NewBigIntMemoryUsage(bytes int) MemoryUsage {
 		Kind:   MemoryKindBigInt,
 		Amount: uint64(bytes),
 	}
+}
+
+func max(a, b int) int {
+	if a > b {
+		return a
+	}
+	return b
+}
+
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+const bigIntWordSize = int(unsafe.Sizeof(big.Word(0)))
+
+func NewPlusBigIntMemoryUsage(a, b *big.Int) MemoryUsage {
+	return NewBigIntMemoryUsage(
+		max(
+			len(a.Bits()),
+			len(b.Bits()),
+		) + bigIntWordSize,
+	)
 }
--- a/runtime/common/metering.go
+++ b/runtime/common/metering.go
@@ -62,17 +62,22 @@ func min(a, b int) int {
 
 const bigIntWordSize = int(unsafe.Sizeof(big.Word(0)))
 
+func BigIntByteLength(v *big.Int) int {
+	return len(v.Bits()) * bigIntWordSize
+}
+
 func NewPlusBigIntMemoryUsage(a, b *big.Int) MemoryUsage {
 	return NewBigIntMemoryUsage(
 		max(
-			len(a.Bits()),
-			len(b.Bits()),
+			BigIntByteLength(a),
+			BigIntByteLength(b),
 		) + bigIntWordSize,
 	)
 }
 
 func NewMulBigIntMemoryUsage(a, b *big.Int) MemoryUsage {
 	return NewBigIntMemoryUsage(
-		len(a.Bits()) + len(b.Bits()),
+		BigIntByteLength(a) +
+			BigIntByteLength(b),
 	)
 }
--- a/runtime/common/metering.go
+++ b/runtime/common/metering.go
@@ -53,16 +53,11 @@ func max(a, b int) int {
 	return b
 }
 
-func min(a, b int) int {
-	if a < b {
-		return a
-	}
-	return b
-}
-
 const bigIntWordSize = int(unsafe.Sizeof(big.Word(0)))
 
 func BigIntByteLength(v *big.Int) int {
+	// NOTE: big.Int.Bits() actually returns bytes:
+	// []big.Word, where big.Word = uint
 	return len(v.Bits()) * bigIntWordSize
 }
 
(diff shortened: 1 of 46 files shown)


runtime/tests/interpreter/memory_metering_test.go
```diff
@@ -688,3 +685,180 @@ func TestRuntimeOptionalValueMetering(t *testing.T) {
...
+        `
+
+		meter := newTestMemoryGauge()
+		inter := parseCheckAndInterpretWithMemoryMetering(t, script, meter)
+
+		_, err := inter.Invoke("main")
+		require.NoError(t, err)
+
+		// creation: 8 + 8
+		// result: 16 = max(8, 8) + 8
```
dsainati1, line 818: It seems like these values are always multiples of `8`, since this the base size of a BigInt. For simplicity can we divide everything by `8` so these numbers more accurately reflect how many numbers have been allocated? Or is there a scenario where an amount of memory could be used that isn't a multiple of 8.

turbolent: `big.Int` is (currently) based on a `[]uint`. That is an implementation detail though, as it may e.g. use a `uint128` if/once it becomes available in Go.

runtime/common/metering.go
```diff
@@ -33,3 +38,41 @@ func NewStringMemoryUsage(length int) MemoryUsage {
...
+	// []big.Word, where big.Word = uint
+	return len(v.Bits()) * bigIntWordSize
+}
+
+func NewPlusBigIntMemoryUsage(a, b *big.Int) MemoryUsage {
+	return NewBigIntMemoryUsage(
+		max(
+			BigIntByteLength(a),
+			BigIntByteLength(b),
+		) + bigIntWordSize,
```
SupunS, line 69: Will be good to have a comment with the calculation logic

runtime/interpreter/decode.go
```diff
@@ -81,6 +82,17 @@ func decodeString(dec *cbor.StreamDecoder, memoryGauge common.MemoryGauge) (stri
 	return dec.DecodeString()
 }
 
+func decodeBigInt(dec *cbor.StreamDecoder, memoryGauge common.MemoryGauge) (*big.Int, error) {
```
SupunS, line 85: Maybe make this a receiver-method of `StorableDecoder`, so the `*cbor.StreamDecoder` and `memoryGauge` can be obtained from the receiver

runtime/interpreter/value.go
```diff
@@ -2492,21 +2500,55 @@ type IntValue struct {
 	BigInt *big.Int
 }
 
-func NewIntValueFromInt64(value int64) IntValue {
-	return NewIntValueFromBigInt(big.NewInt(value))
+func NewIntValueFromInt64(memoryGauge common.MemoryGauge, value int64) IntValue {
+	const int64Size = int(unsafe.Sizeof(int64(0)))
+	return NewIntValueFromBigInt(
+		memoryGauge,
+		common.NewBigIntMemoryUsage(int64Size),
```
SupunS, line 2507: `common.NewBigIntMemoryUsage(int64Size)` is a fixed value for all Int64 values it seems. Can move to a module-level variable?

runtime/interpreter/value.go
```diff
@@ -6121,29 +6193,63 @@ type UIntValue struct {
 	BigInt *big.Int
 }
 
-func NewUIntValueFromUint64(value uint64) UIntValue {
-	return NewUIntValueFromBigInt(new(big.Int).SetUint64(value))
+func NewUIntValueFromUint64(memoryGauge common.MemoryGauge, value uint64) UIntValue {
+	const uint64Size = int(unsafe.Sizeof(uint64(0)))
+	return NewUIntValueFromBigInt(
+		memoryGauge,
+		common.NewBigIntMemoryUsage(uint64Size),
```
SupunS, line 6200: Similarly, `common.NewBigIntMemoryUsage(uint64Size)` could be calculated only once

SupunS (approved):
LGTM!

runtime/interpreter/value.go
```diff
@@ -2492,21 +2500,55 @@ type IntValue struct {
 	BigInt *big.Int
 }
 
-func NewIntValueFromInt64(value int64) IntValue {
-	return NewIntValueFromBigInt(big.NewInt(value))
+func NewIntValueFromInt64(memoryGauge common.MemoryGauge, value int64) IntValue {
```
robert-e-davidson3, line 2503: Thought: instead of passing in MemoryGauge in these sorts of functions, we pass in the value that will become the Amount in the MemoryGauge. That should simplify the calls. Maybe call it something like memoryWeight?

robert-e-davidson3 (approved):

runtime/common/metering.go
turbolent: Yes, I'll update the formulas and add comments once https://github.com/dapperlabs/cadence-private-issues/issues/32 is complete

runtime/interpreter/decode.go
turbolent: Good idea! 

I replicated `decodeString`, which is called from different decoders (storable, type, location), but you are right `decodeBigInt` is only ever called from `StorableDecoder` 👍 

Refactored in d47975221eb129b00395f9f03a5b477b23567472

runtime/interpreter/value.go
turbolent: Good point, refactored in 05a3f20542f1f94c0a04a33b7107229f7829688c

runtime/interpreter/value.go
turbolent: Also refactored in 05a3f20542f1f94c0a04a33b7107229f7829688c

runtime/interpreter/value.go
turbolent: Here in particular it is constant. Do you have an example for the other constructor?

turbolent:
Merging this for now, so we can continue, but we can continue the conversation for improvements on Slack and address the suggestions in follow-up PRs

runtime/interpreter/value.go
robert-e-davidson3: Oh sorry I misread that.

Anyway, my example would be StringValue:
```
func NewStringValue(memoryGauge common.MemoryGauge, memoryUsage common.MemoryUsage, stringConstructor func() string) *StringValue {
	if memoryGauge != nil {
		memoryGauge.UseMemory(memoryUsage)
	}
	str := stringConstructor()
	return NewUnmeteredStringValue(str)
}
```

Becomes:
```
func NewStringValue(memoryGauge common.MemoryGauge, memoryWeight uint64, stringConstructor func() string) *StringValue {
	if memoryGauge != nil {
			Kind:   MemoryKindString,
			Amount: memoryWeigh,
		})
	}
	str := stringConstructor()
	return NewUnmeteredStringValue(str)
}
```

Probably either stop using `NewStringMemoryUsage` or make it something like `func NewStringMemoryWeight(length int) uint64`.

If we do this then we might be able to get rid of the `MemoryUsage`  struct. Instead, `UseMemory` would take `Kind` and `Amount` as arguments.

PR #1494, opened 2022-03-09
Add memory metering for `Int8`, `Int16`, `Int32` and `Int64` Values

SupunS:

## Description

______

### DIFF
--- a/runtime/interpreter/decode.go
+++ b/runtime/interpreter/decode.go
@@ -82,6 +82,11 @@ func decodeString(dec *cbor.StreamDecoder, memoryGauge common.MemoryGauge) (stri
 	return dec.DecodeString()
 }
 
+func decodeInt64(d StorableDecoder) (int64, error) {
+	common.UseMemory(d.memoryGauge, int64MemoryUsage)
+	return d.decoder.DecodeInt64()
+}
+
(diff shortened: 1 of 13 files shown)


runtime/interpreter/decode.go
```diff
@@ -350,6 +350,8 @@ func (d StorableDecoder) decodeInt() (IntValue, error) {
 }
 
 func (d StorableDecoder) decodeInt8() (Int8Value, error) {
+	common.UseMemory(d.memoryGauge, int8MemoryUsage)
+
 	v, err := d.decoder.DecodeInt64()
```
turbolent, line 355: As a `int64` is decoded, it should be metered as such. Maybe define a function `decodeInt64` which meters 8 bytes and calls the `StreamDecoder.DecodeInt64`, e.g. like the `decodeString` helper function

turbolent (commented):
Nice, thank you!

robert-e-davidson3 (approved):

runtime/interpreter/decode.go
SupunS: added with e76c1f6

turbolent (approved):
Great work! 👍

PR #1497, opened 2022-03-11
[WIP] Memory measurement calibration

dsainati1:
https://github.com/dapperlabs/cadence-private-issues/issues/33

This PR (not to be merged) should help us compute the weights for each of the abstract memory kinds. The methodology is as follows:

We can consider the memory measurement to be some equation `M = c1*x1 + c2*x2 + ... + C`, where `M` denotes the actual memory used by Go when running a transaction, `xn` and `cn` are the number of abstract memory units of each kind and a constant factor for each of them, and then `C` is some constant amount of overhead memory used by Go. Each program generates its own equation. To solve this, we need a program that creates values of each memory kind to generate `n` equations, which we can then solve. 

The programs here generate 10000 of each value in order to reduce noise, so that `C` does not dominate the equation.


dsainati1:
We still need programs that create:
* BlockValue
* LinkValue
* CapabilityValue
* VoidValue
* StorageReferenceValue

turbolent:
Nice work! 

I think it's fine if we merge this as a tool, maybe move it to `tools` where the other tools are.

dsainati1:
One thing I have noticed is that the `string_test` program is only reporting 4 units of string memory. Is there some kind of optimization we have here that is only allocating the literal once and then making every assignment refer to that same literal? This is probably good for us to have, but it does mean I'm going to need a different way to create 10000 simple values for the purposes of this test.

turbolent:
Yes, string literals are only metered once, as they're immutable. You could maybe try e.g. concatenating?

turbolent:
@dsainati1 Is this still needed?

dsainati1:
Nope!

PR #1505, opened 2022-03-15
Refactor memory-metering interface API

SupunS:

## Description

Refactor the memory-metering interface API to be in-lined with the computation metering API.

old-signature:
```go
UseMemory(usage MemoryUsage)
```
new-signature:
```go
MeterMemory(usage common.MemoryUsage) error
```
Returned `error` is handled by Cadence.
______


janezpodhostnik (approved):
The changes look good, but my question is if the `MeterMemory` function should accept two parameters like `MeterComputation(operationType common.ComputationKind, intensity uint) error` does.

SupunS:
That's a good question! I had the same thought when adding changes. The reason why I went with the single `common.MemoryUsage` parameter is, in future if we needed to pass more properties through this API, by having this wrapper object (`common.MemoryUsage`) we can seamlessly add them, without changing/breaking the API.

janezpodhostnik:
That is true, but if the `common.MemoryUsage` changes, wouldn't that still be a breaking change for anyone implementing the runtime interface.

SupunS:
Additions of fields would be backward compatible. Yes, they would have to update the code to use the new field, but the code will continue to compile without any changes.

turbolent (approved):

ramtinms (approved):

PR #1513, opened 2022-03-17
Refactor to meter address value before creation, whenever possible

SupunS:

## Description

Just a refactor, to delay the creation of address whenever possible.

______

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -16652,38 +16652,49 @@ func (*EphemeralReferenceValue) DeepRemove(_ *Interpreter) {
 //
 type AddressValue common.Address
 
-func NewUnmeteredAddressValue(b []byte) AddressValue {
+func NewUnmeteredAddressValueFromBytes(b []byte) AddressValue {
 	result := AddressValue{}
 	copy(result[common.AddressLength-len(b):], b)
 	return result
 }
 
+// NewAddressValue constructs an address-value from a `common.Address`.
+//
+// NOTE:
+// This method must only be used if the `address` value is already constructed,
+// and/or already loaded onto memory. This is a convenient method for better performance.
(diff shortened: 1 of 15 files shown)


runtime/interpreter/value.go
```diff
@@ -16652,38 +16652,49 @@ func (*EphemeralReferenceValue) DeepRemove(_ *Interpreter) {
...
+			result[:common.AddressLength],
+			uint64(uint64Value),
+		)
 
-	binary.BigEndian.PutUint64(
-		result[:common.AddressLength],
-		uint64(uint64Value),
-	)
-	return NewAddressValue(memoryGauge, result)
+		return
```
robert-e-davidson3, line 16694: unnecessary return

```suggestion
```

runtime/runtime.go
```diff
@@ -1775,18 +1773,21 @@ func (r *interpreterRuntime) newCreateAccountFunction(
...
+		addressGetter := func() (address common.Address) {
+			var err error
+			wrapPanic(func() {
+				address, err = context.Interface.CreateAccount(payerAddress)
+			})
+			if err != nil {
+				panic(err)
+			}
+
+			return
```
robert-e-davidson3, line 1785: unnecessary return

```suggestion
```

robert-e-davidson3 (approved):
rad! job a couple of trivial suggestions

runtime/interpreter/value.go
SupunS: This is actually needed - it returns the `result`.
its a shortened form of `return result`

runtime/runtime.go
SupunS: same as above

dsainati1 (approved):

runtime/interpreter/value.go
robert-e-davidson3: TIL. Solidity does that too.

PR #1514, opened 2022-03-18
atree string value metering

dsainati1:

## Description

Composite type field keys are `atreeStringValue`s, which consume memory and need to be metered. Here, we create a new `MemoryKindRawString` which is used to meter raw strings, rather than `StringValue`s. This comes into play when composite or other complex types like `Path`, `Location` or `Composite` fields are being metered. 

______


turbolent (approved):
Nice!

PR #1521, opened 2022-03-21
Add memory metering for fixed-size numbers

SupunS:

______


runtime/interpreter/value.go
```diff
@@ -12772,7 +12835,12 @@ func ConvertFix64(value Value) Fix64Value {
...
-		return Fix64Value(value)
+		return NewFix64Value(
+			memoryGauge,
+			func() int64 {
+				return int64(value)
+			},
+		)
 
 	case BigNumberValue:
 		v := value.ToBigInt()
```
turbolent, line 12846: Needs to be metered before this allocation occurs. See e.g. `ConvertInt64`

turbolent (commented):
Nice!

runtime/interpreter/value.go
SupunS: ah yes, I had the entire function as the converter, but then refactored to make it a bit cleaner, and have missed these two in the process. Fixed them in: b669fa35e80b4803f1ad35551c79062d4b4a8d0b

robert-e-davidson3 (approved):

dsainati1:
All the fixed-size numbers use the same amount of memory (hence the fixed-size), so I don't see the need to use constructor functions for them like we did with strings. We can just have the creation functions for these values meter this constant amount and then do their normal operation.

dsainati1:
Yea I guess my concern is that it complicates the API unnecessarily and results in extra computation because of the deferred function calls, when I think there is no possible harm done by metering the allocation after it happens in this case, since it is a fixed small size.

SupunS:
This is the pattern we have followed for all other values (e.g all number types, all simple types, etc)

turbolent (approved):

turbolent:
@SupunS Looks good! Didn't realize there were follow-up commits, sorry about the delay

PR #1522, opened 2022-03-21
Check address length before decoding

SupunS:

## Description

@robert-e-davidson3 and I had a discussion regarding this in PR https://github.com/onflow/cadence/pull/1442#discussion_r828514562 and decided not to change this, until the behaviour is confirmed.

Came across this thread yesterday and realized it has been already discussed in https://axiomzen.slack.com/archives/CG0B7CJAJ/p1646701886736249 and concluded that it's safe to use `NextSize` to get the size of `bytes` before decoding.

______

### DIFF
--- a/runtime/interpreter/decode.go
+++ b/runtime/interpreter/decode.go
@@ -738,8 +738,7 @@ func (d StorableDecoder) decodeSome() (SomeStorable, error) {
 	}, nil
 }
 
-func checkEncodedAddressLength(addressBytes []byte) error {
-	actualLength := len(addressBytes)
+func checkEncodedAddressLength(actualLength int) error {
 	const expectedLength = common.AddressLength
 	if actualLength > expectedLength {
 		return fmt.Errorf(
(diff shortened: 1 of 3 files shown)


runtime/interpreter/decode.go
```diff
@@ -1673,6 +1682,9 @@ func (d LocationDecoder) decodeAddressLocation() (common.Location, error) {
 	// Address
 
 	// Decode address at array index encodedAddressLocationAddressFieldKey
+	//
+	// TODO: Use `decodeAddressBytes` and remove the `checkEncodedAddressLength` below
```
turbolent, line 1686: Can this already be done in this PR?

turbolent (approved):
Nice!

runtime/interpreter/decode.go
SupunS: I thought about it, but then it will introduce some inconsistency - i.e: locations will be metered during decoding, but not when created in-line.

robert-e-davidson3 (approved):
rad!

PR #1527, opened 2022-03-22
Forbid reinitialization of variable, resource-kinded fields

turbolent:
## Description

In addition to constant fields, also variable resource-kinded fields may not be re-initialized.

______

### DIFF
--- a/runtime/sema/check_assignment.go
+++ b/runtime/sema/check_assignment.go
@@ -274,6 +274,8 @@ func (checker *Checker) visitMemberExpressionAssignment(
 		)
 	}
 
+	targetIsConstant := member.VariableKind == ast.VariableKindConstant
+
 	// If this is an assignment to a `self` field, it needs special handling
 	// depending on if the assignment is in an initializer or not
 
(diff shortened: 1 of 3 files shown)


runtime/sema/check_assignment.go
```diff
@@ -294,17 +296,26 @@ func (checker *Checker) visitMemberExpressionAssignment(
...
 
-				// If the field is constant and it has already previously been
-				// initialized, report an error for the repeated assignment
+				// If the field is constant,
+				// or it is variable and resource-kinded,
+				// and it has already previously been initialized,
+				// report an error for the repeated assignment / initialization
+				//
+				// Assigning to a variable, resource-kinded field is invalid,
+				// because the initial value would get lost.
```
robert-e-davidson3, line 305: It might be clearer to use more well-named variables than to have a large comment. You started this with `targetIsConstant`.

robert-e-davidson3 (approved):

SupunS (approved):
Thanks for fixing!

runtime/sema/check_assignment.go
turbolent: What is your concrete suggestion?

robert-e-davidson3: On second thought, it wouldn't be good to run `initializedFieldMembers.Contains(accessedSelfMember)` if unnecessary. It could be put into a function but that's probably less clear overall.

Anyway, here's what I was thinking:

```
...

targetIsConstant := member.VariableKind == ast.VariableKindConstant
targetIsResource := member.TypeAnnotation.Type.IsResourceType()
targetIsInitialized := functionActivation.InitializationInfo.InitializedFieldMembers.Contains(accessedSelfMember)

...

// Report error for repeated assignment/initialization of constants or resource variables.
// Cannot assign to resource-kinded fields because the initial value would be lost.

if (targetIsConstant || targetIsResource) && targetIsInitialized {

...
```

PR #1531, opened 2022-03-24
Memory metering for variables

dsainati1:

## Description

As discussed in the issue above, we need to meter variables as well as values, as only metering values does not accurately capture all cases, such as one where variables are created many times from the same values, as in a loop. This adds a new memory kind for variables and meters a constant each time a new variable is created. 

______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -3239,7 +3239,7 @@ var converterFunctionValues = func() []converterFunction {
 			if converterFunctionValue.NestedVariables == nil {
 				converterFunctionValue.NestedVariables = map[string]*Variable{}
 			}
-			// these variables are do need to be metered as they are only ever declared once,
+			// these variables are not needed to be metered as they are only ever declared once,
 			// and can be considered base interpreter overhead
 			converterFunctionValue.NestedVariables[name] = NewVariableWithValue(nil, value)
 		}
@@ -3427,7 +3427,7 @@ func defineBaseValue(activation *VariableActivation, name string, value Value) {
 	if activation.Find(name) != nil {
 		panic(errors.NewUnreachableError())
 	}
-	// these variables are do need to be metered as they are only ever declared once,
+	// these variables are not needed to be metered as they are only ever declared once,
 	// and can be considered base interpreter overhead
 	activation.Set(name, NewVariableWithValue(nil, value))
 }
@@ -3450,7 +3450,7 @@ var stringFunction = func() Value {
 		if functionValue.NestedVariables == nil {
 			functionValue.NestedVariables = map[string]*Variable{}
 		}
-		// these variables are do need to be metered as they are only ever declared once,
+		// these variables are not needed to be metered as they are only ever declared once,
 		// and can be considered base interpreter overhead
 		functionValue.NestedVariables[name] = NewVariableWithValue(nil, value)
 	}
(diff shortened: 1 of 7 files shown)


runtime/interpreter/interpreter.go
```diff
@@ -3239,7 +3239,9 @@ var converterFunctionValues = func() []converterFunction {
 			if converterFunctionValue.NestedVariables == nil {
 				converterFunctionValue.NestedVariables = map[string]*Variable{}
 			}
-			converterFunctionValue.NestedVariables[name] = NewVariableWithValue(value)
+			// these variables are do need to be metered as they are only ever declared once,
+			// and can be considered base interpreter overhead
+			converterFunctionValue.NestedVariables[name] = NewVariableWithValue(nil, value)
```
SupunS, line 3244: I guess the comment should be 'not metered'
```suggestion
			// these variables are not needed to be metered as they are only ever declared once,
			// and can be considered base interpreter overhead
			converterFunctionValue.NestedVariables[name] = NewVariableWithValue(nil, value)
```

Maybe add a new 'NewUnmeteredVariableWithValue' constructor, just so that its more explicit

runtime/interpreter/interpreter.go
```diff
@@ -3425,7 +3427,9 @@ func defineBaseValue(activation *VariableActivation, name string, value Value) {
 	if activation.Find(name) != nil {
 		panic(errors.NewUnreachableError())
 	}
-	activation.Set(name, NewVariableWithValue(value))
+	// these variables are do need to be metered as they are only ever declared once,
```
SupunS, line 3430:
```suggestion
	// these variables are not needed to be metered as they are only ever declared once,
```

runtime/interpreter/interpreter.go
```diff
@@ -3446,7 +3450,9 @@ var stringFunction = func() Value {
 		if functionValue.NestedVariables == nil {
 			functionValue.NestedVariables = map[string]*Variable{}
 		}
-		functionValue.NestedVariables[name] = NewVariableWithValue(value)
+		// these variables are do need to be metered as they are only ever declared once,
```
SupunS, line 3453:
```suggestion
		// these variables are not needed to be metered as they are only ever declared once,
```

runtime/interpreter/variable.go
```diff
@@ -36,13 +38,19 @@ func (v *Variable) SetValue(value Value) {
 	v.value = value
 }
 
-func NewVariableWithValue(value Value) *Variable {
+func NewVariableWithValue(gauge common.MemoryGauge, value Value) *Variable {
+	if gauge != nil {
```
SupunS, line 42: `common.UseMemory()` method already have this `nil` check. should be safe to pass `gauge` as is

SupunS (commented):
Thanks for adding this!

Can you also add some tests for global-variables and parameters as well?

runtime/interpreter/variable.go
```diff
@@ -36,13 +38,15 @@ func (v *Variable) SetValue(value Value) {
 	v.value = value
 }
 
-func NewVariableWithValue(value Value) *Variable {
+func NewVariableWithValue(gauge common.MemoryGauge, value Value) *Variable {
+	common.UseMemory(gauge, common.NewConstantMemoryUsage(common.MemoryKindVariable))
```
turbolent, line 42: Maybe declare a global for it and reuse it here and in `NewVariableWithGetter`:

```go
var variableMemoryUsage = common.NewConstantMemoryUsage(common.MemoryKindVariable)
```

PR #1533, opened 2022-03-24
Add memory metering for tokens

SupunS:

Also added a test to benchmark parsing with and without metering. Didn't show me a much difference between the two.

```
goos: darwin
goarch: amd64
pkg: github.com/onflow/cadence/runtime/parser2
cpu: Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz

BenchmarkParseFungibleToken/Without_memory_metering-8         	    6214	    163507 ns/op	  198849 B/op	    1074 allocs/op
BenchmarkParseFungibleToken/With_memory_metering-8            	    6052	    167140 ns/op	  198849 B/op	    1074 allocs/op
```
______

### DIFF
--- a/runtime/parser2/lexer/lexer.go
+++ b/runtime/parser2/lexer/lexer.go
@@ -270,10 +270,30 @@ func (l *lexer) endPos() position {
 }
 
 func (l *lexer) emitType(ty TokenType) {
+	if l.memoryGauge != nil {
+		usage := l.typeMemoryUsage(ty)
+
+		// Don't use `common.MemoryUsae()` to avoid redundant `nil` check.
+		err := l.memoryGauge.MeterMemory(usage)
+		if err != nil {
+			panic(err)
+		}
+	}
+
 	l.emit(ty, nil, l.startPosition(), true)
 }
 
 func (l *lexer) emitValue(ty TokenType) {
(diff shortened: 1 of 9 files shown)


runtime/parser2/lexer/lexer.go
```diff
@@ -270,10 +270,30 @@ func (l *lexer) endPos() position {
...
 func (l *lexer) emitType(ty TokenType) {
+	if l.memoryGauge != nil {
+		usage := l.typeMemoryUsage(ty)
+
+		// Don't use `common.MemoryUsae()` to avoid redundant `nil` check.
+		err := l.memoryGauge.MeterMemory(usage)
+		if err != nil {
+			panic(err)
+		}
+	}
```
SupunS, line 281: I'm not sure whether we need to meter these types of tokens as they are not part of the AST (discarded). But at the same time, they are available in the token stream - so it's sort of in-memory

turbolent (approved):
Great work! 👍

runtime/parser2/lexer/lexer.go
turbolent: Yes, I think we need to meter them, as they are allocations resulting from user interaction (code submission) and kept in memory, even if just temporarily

PR #1536, opened 2022-03-25
Meter identifiers

SupunS:

## Description

______


dsainati1 (commented):
One question about this: it seems like we are only metering the number of identifiers, and not including the length, so an extremely long identifier would count the same as a one-character identifier. Are we planning to meter the program length to prevent people from just allocating a small number of long identifiers?

SupunS:
This PR adds metering to the identifier wrapper. The content of the identifier is metered as part of tokens-metering: https://github.com/onflow/cadence/pull/1533

dsainati1 (approved):

PR #1538, opened 2022-03-26
Memory metering for primitive StaticType

### LINKED ISSUE #1503: StaticType memory metering
Note that when the static type metering is done, this ticket also includes metering for `TypeValue`. This will be fairly easy because all of its components will be metered.

Relates to https://github.com/dapperlabs/cadence-private-issues/issues/2, https://github.com/dapperlabs/cadence-private-issues/issues/36 and https://github.com/onflow/cadence/pull/1374.

Blocked by #1442

- [ ] ArrayStaticType
- [ ] CapabilityStaticType
- [ ] CompositeStaticType
- [ ] ConstantSizedStaticType
- [ ] DictionaryStaticType
- [ ] FunctionStaticType
- [ ] InterfaceStaticType
- [ ] OptionalStaticType
- [ ] PrimitiveStaticType - @robert-e-davidson3 
- [ ] ReferenceStaticType
- [ ] RestrictedStaticType
- [ ] VariableSizedStaticType

robert-e-davidson3:
Works towards #1503 
Works towards https://github.com/dapperlabs/cadence-private-issues/issues/2, https://github.com/dapperlabs/cadence-private-issues/issues/36

## Description

Whenever a StaticType is allocated in a way that a user might influence, note the increase in memory usage.

______


SupunS:
@robert-e-davidson3 It would be good to split the changes to several PRs, to make it easy to review and can to get the changes in quickly. Metering just the primitive-statictypes sounds like a good sub-set; maybe make this PR ready to merge as-is, and send the other changes in separate PRs?

robert-e-davidson3:
> @robert-e-davidson3 It would be good to split the changes to several PRs, to make it easy to review and can to get the changes in quickly. Metering just the primitive-statictypes sounds like a good sub-set; maybe make this PR ready to merge as-is, and send the other changes in separate PRs?

Done! I have the rest of the changes stashed so I can bring them back for the subsequent PR.

robert-e-davidson3:
It looks like primitive static types are only used for arrays and dictionaries. Is that correct?
@turbolent @SupunS

runtime/convertTypes.go
```diff
@@ -423,97 +423,97 @@ func importCompositeType(t cadence.CompositeType) interpreter.CompositeStaticTyp
 	}
 }
 
-func ImportType(t cadence.Type) interpreter.StaticType {
+func ImportType(memoryGauge common.MemoryGauge, t cadence.Type) interpreter.StaticType {
 	switch t := t.(type) {
 	case cadence.AnyType:
-		return interpreter.PrimitiveStaticTypeAny
+		return interpreter.NewPrimitiveStaticType(memoryGauge, interpreter.PrimitiveStaticTypeAny)
```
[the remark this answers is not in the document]
robert-e-davidson3: They're passed by value so they can still take up space... can there be an array of static types? I've only found primitive static types to be instantiated with arrays and dictionaries and then they'd be implicitly covered by array/dictionary size.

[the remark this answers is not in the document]
dsainati1: There are [functions](https://docs.onflow.org/cadence/language/run-time-types/#constructing-a-run-time-type) (or rather there will be functions) that create run-time `TypeValue`s that thus implicitly create `StaticType`s. So yes, you could in theory construct an arbitrary array of runtime types.

runtime/interpreter/value.go
```diff
@@ -94,7 +94,7 @@ type Value interface {
 	Accept(interpreter *Interpreter, visitor Visitor)
 	Walk(interpreter *Interpreter, walkChild func(Value))
 	DynamicType(interpreter *Interpreter, seenReferences SeenReferences) DynamicType
-	StaticType() StaticType
+	StaticType(interpreter *Interpreter) StaticType
```
dsainati1, line 97: Is there a reason that `StaticType` needs to take an `*Interpreter` instead of a `MemoryGauge`. I didn't realize this until I needed to write https://github.com/onflow/cadence/pull/1558, but I think we should try to make these argument types as general as possible.

robert-e-davidson3: I went with `*Interpreter` to match the other methods.

dsainati1: Yeah, we should probably change the rest of those over as well.

SupunS: An interpreter is needed for the `StaticType()` method for the runtime subtype check change https://github.com/onflow/cadence/pull/1487 (if we are also including it that is). I would suggest having the interpreter only for this method because then it would cater to both requirements + fewer conflicts.
 
I agree it makes sense to have a `MemoryGauge` parameter for other methods though.

turbolent (approved):

PR #1546, opened 2022-03-31
Improve metering of containers

dsainati1:
https://github.com/dapperlabs/cadence-private-issues/issues/36
Closes https://github.com/dapperlabs/cadence-private-issues/issues/34

## Description

We were previously only metering a constant amount of memory per container created, which did not accurately reflect the reality of the underlying atree representation. This meters containers on creation based on length (for arrays and dictionaries) and number of fields (for composites), and also adds additional constant metering for calls to append and insert for these types. 

______


runtime/common/metering.go
```diff
@@ -50,6 +50,27 @@ func NewConstantMemoryUsage(kind MemoryKind) MemoryUsage {
 	}
 }
 
+func NewArrayMemoryUsage(length int) MemoryUsage {
+	return MemoryUsage{
+		Kind:   MemoryKindArray,
+		Amount: uint64(length) + 1, // +1 to account for empty arrays
```
robert-e-davidson3, line 56: I thought us doing this in the strings was messing up your calculations? Or is it fine?

My approach was to add another MemoryKind so every container would have two: a constant one for its existence and a variable one for its length.

turbolent: Agreed, it might be better to split this up. The code reads fine here, but is confusing below:
https://github.com/onflow/cadence/pull/1546/files#diff-424575c293dc2bc3a28aab845829d7656fa83a7be4e026de8906935e23bbd52eR1652

> common.NewArrayMemoryUsage(0)

This reads like the cost of allocating an "empty array", but really is the allocation cost of the addition of an element.

runtime/interpreter/value.go
```diff
@@ -1694,6 +1699,9 @@ func (v *ArrayValue) Insert(interpreter *Interpreter, getLocationRange func() Lo
 		})
 	}
 
+	// length increases by 1
+	common.UseMemory(interpreter, common.NewArrayLengthUsage(0))
```
robert-e-davidson3, line 1703: Should be `1`?
```suggestion
	common.UseMemory(interpreter, common.NewArrayLengthUsage(1))
```

runtime/tests/interpreter/memory_metering_test.go
```diff
@@ -60,60 +60,73 @@ func TestInterpretArrayMetering(t *testing.T) {
...
-                let y: [[String]] = [[]]
-                let z: [[[Bool]]] = [[[]]]
-            }
-        `
+	            pub fun main() {
+	                let x: [Int8] = []
+	                let y: [[String]] = [[]]
+	                let z: [[[Bool]]] = [[[]]]
+	            }
+	        `
```
robert-e-davidson3, line 68: Doesn't really matter but cadence's error reporting doesn't count tabs correctly so spaces are preferred.

robert-e-davidson3 (changes requested):

robert-e-davidson3 (approved):
lgtm!

SupunS (approved):
LGTM!

PR #1550, opened 2022-04-01
[0.23] Fix overflow error in UInt to int conversion

turbolent:
## Description

The function `UIntValue.ToInt` was incorrectly rejecting valid values, which caused a simple program like the following to fail with an overflow error, even though it should not:

```kotlin
Welcome to Cadence v0.23.2!
Type '.help' for assistance.

1> [""][0 as UInt]
panic: overflow
```

This was a regression introduced in #1208.

Fix the bug and also add tests covering all number types/values' `ToInt` and also `ToBigInt`.

______

### DIFF
--- a/runtime/runtime_test.go
+++ b/runtime/runtime_test.go
@@ -1136,11 +1136,7 @@ func TestRuntimeTransactionWithArguments(t *testing.T) {
 			if tc.check != nil {
 				tc.check(t, err)
 			} else {
-				if !assert.NoError(t, err) {
-					for err := err; err != nil; err = errors.Unwrap(err) {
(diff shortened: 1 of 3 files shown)


runtime/runtime_test.go
```diff
@@ -1136,11 +1136,7 @@ func TestRuntimeTransactionWithArguments(t *testing.T) {
 			if tc.check != nil {
 				tc.check(t, err)
 			} else {
-				if !assert.NoError(t, err) {
```
turbolent, line 1139: Remove the unnecessary logging resulting in noise here and below

SupunS (approved):

PR #1562, opened 2022-04-06
Add a checker option to short-circuit error reporting

turbolent:
## Description

We can short-circuit error reporting in some environments. For example, we want to report all errors in a developer tool, but we do not need to report all errors on-chain.

Add a checker option which stops error reporting after the first reported error. 
The new option is disabled by default, but should be enabled in the flow-go EN.

______

### DIFF
--- a/runtime/sema/checker.go
+++ b/runtime/sema/checker.go
@@ -250,7 +250,9 @@ func WithLintingEnabled(enabled bool) Option {
 }
 
 // WithErrorShortCircuitingEnabled returns a checker option which enables/disables
-// error short-circuiting.
+// error short-circuiting in the checker.
+// When enabled, the checker will stop running once it encounters an error.
+// When disabled (the default), the checker reports the error then continues checking.
 //
 func WithErrorShortCircuitingEnabled(enabled bool) Option {
 	return func(checker *Checker) error {
--- a/runtime/sema/checker.go
+++ b/runtime/sema/checker.go
@@ -115,7 +115,8 @@ type Checker struct {
 	checkHandler                       CheckHandlerFunc
 	expectedType                       Type
 	memberAccountAccessHandler         MemberAccountAccessHandlerFunc
-	lintEnabled                        bool
+	lintingEnabled                     bool
+	errorShortCircuitingEnabled        bool
 	// memoryGauge is used for metering memory usage
 	memoryGauge common.MemoryGauge
 }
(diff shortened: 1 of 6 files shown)


turbolent:
cc @janezpodhostnik @ramtinms

robert-e-davidson3 (approved):

runtime/sema/checker.go
```diff
@@ -241,7 +242,17 @@ func WithPositionInfoEnabled(enabled bool) Option {
...
 func WithLintingEnabled(enabled bool) Option {
 	return func(checker *Checker) error {
-		checker.lintEnabled = enabled
+		checker.lintingEnabled = enabled
+		return nil
+	}
+}
+
+// WithErrorShortCircuitingEnabled returns a checker option which enables/disables
+// error short-circuiting.
```
robert-e-davidson3, line 251: The function name does a pretty good job of communicating this. I recommend describing error short-circuiting:
```suggestion
// WithErrorShortCircuitingEnabled returns a checker option which enables/disables
// error short-circuiting in the checker.
// When enabled, the checker will stop running once it encounters an error.
// When disabled (the default), the checker reports the error then continues checking.
```

robert-e-davidson3 (approved):

runtime/sema/checker.go
turbolent: Nice! Adding

runtime/sema/checker.go
```diff
@@ -365,11 +381,29 @@ func (checker *Checker) IsChecked() bool {
...
 func (checker *Checker) Check() error {
 	if !checker.IsChecked() {
 		checker.Elaboration.setIsChecking(true)
 		checker.errors = nil
 		check := func() {
+			if checker.errorShortCircuitingEnabled {
+				defer func() {
+					switch recovered := recover().(type) {
+					case stopChecking:
+						// checking should stop
```
SupunS, line 395: If the error is in an import, would it (need to?) terminate checking the original program?

turbolent: That's a good question. I'd say the stop should propagate: If a program is imported, and check it reports an error, then checking the importing program should also stop. 

I'll add a test case for this

turbolent: Added a test case in 57906f4aee6331bfc056d64e379ef3ba553a7982

SupunS (approved):
Nice! Thanks for adding the test

PR #1566, opened 2022-04-08
Check reference expression operator

### LINKED ISSUE #1406: Parser/Checker do not reject invalid reference operators
### Problem

Reference expressions are always of the form `&v as &T`.

However, the parser parses the expression as a casting expression `v op T`, where op is one of `as`, `as!`, `as?`.

The parser or checker should reject the invalid operators `as!`  and `as?` in reference expressions.

### Issues

Many contracts, e.g. even the core contracts, use invalid operators

turbolent:
Closes #1406 

## Description

Reference expressions and casting expressions have the same syntax, so reference expressions are parsed by parsing casting expressions. 

Ensure the casting expression has the correct operator (`as`).

At least make the current syntax / parsing strict, but we still might want to change the syntax has discussed.

______

### DIFF
--- a/runtime/ast/expression.go
+++ b/runtime/ast/expression.go
@@ -1234,6 +1234,7 @@ func (e *FunctionExpression) MarshalJSON() ([]byte, error) {
 type CastingExpression struct {
 	Expression                Expression
 	Operation                 Operation
+	OperationPos              Position
 	TypeAnnotation            *TypeAnnotation
 	ParentVariableDeclaration *VariableDeclaration `json:"-"`
 }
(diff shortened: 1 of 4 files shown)


robert-e-davidson3 (approved):

dsainati1:
I agree that this is an improvement over not doing the check, but I think this error is going to be very confusing to people. If people were using `as!` and `as?` instead of `as` because they didn't understand the difference between a reference and a cast, complaining that they are using the wrong one is not going to make sense. IMO at least until we have time to address the syntax similarity we should include something in the text of the error that either explains the difference or at least links to the docs.

SupunS (approved):
This sounds like a good temporary solution to me

turbolent:
@dsainati1 Agreed. We won't merge this for Secure Cadence, I just opened it so we would at least start to do something about it, even if just reporting *some* error until we have a proper decision about what to do here (e.g change the syntax, like discussed before).

> [...] we should include something in the text of the error that either explains the difference or at least links to the docs.
Could you maybe add this? You might be better at phrasing this than I would

turbolent:
Replaced by #1661

Issue #1570, opened 2022-04-12
[LS] Better cache invalidation for imported program checkers


sideninja:
### Issue To Be Solved
The language server creates a new checker every time when checking the imported programs as seen here, this might be quite expensive and was done as a quick patch to a problem in production.

### Suggested Solution
The caching should be reverted and stay in place but there should be a cache invalidation system that is smarter, potentially by analyzing an inversed dependency tree and just invalidating the dependencies that changed.

sideninja:
Base test suite indicates changes are processed correctly. Trying to set up the playground version at the time the issue was reported. Coming across some difficulties setting up that version.

sideninja:
I see everything 👁️  😄  yes I did and thank you for that. It will serve well with this for sure. I'm working today to reproduce the bug locally with local playground setup. I want to make sure this is indeed causing the bug since when I now wrote tests it seemed like the cached checkers work ok.

sideninja:
This recording (sorry for potato resolution) is showing that the bug with added members is not possible to recreate. I’ve chosen the version before any updates and even though there are bunch of other issues with UI (a lot of them are now fixed) the one with member not being visible is not reproducable by me. I was monitoring the LS communication and the responses are correct. This makes me conclude it probably isn’t a bug related to LS. I will make sure to extend tests in LS potentially to find any problems, but the base test I made for this problem is working which is another indication it’s not a problem from the LS. 
cc @turbolent @srinjoyc 

https://user-images.githubusercontent.com/75445744/167633602-4ff07673-2dfb-4689-9d15-4d949617e3c6.mov

sideninja:
Waiting to sync with Bastian.

KhaledAylii:
@sideninja @turbolent any news on this issue? Making changes in a contract and redeploying does not reflect in the types imported by other accounts or transaction type checkers - pretty much a development dead end

KhaledAylii:
update: refreshing page solves the problem for now

sideninja:
@KhaledAylii I plan to work on this in the following days.

PR #1572, opened 2022-04-12
[LS] Remove caching of checkers for imported programs

### LINKED ISSUE #1571: [LS] New members not visible from imported contracts
The language server does not produce valid member declaration lists for imported programs. The consequence is described in the issue here: https://github.com/onflow/flow-playground/issues/234

sideninja:
Closes #1571

## Description
When LS is resolving the imports it uses a cached checker which doesn't include any changes to the imported program resulting in the current program not seeing those changes in diagnostics / code completions. This PR makes sure the checker is initialized every time, which arguably can be expensive but it's meant to be a quick patch on the blocking problem in production, a better way would be to have a "reversed" dependency tree and figure out which dependencies actually change and not just do a new re-init every time, this is been documented in an issue for future https://github.com/onflow/cadence/issues/1570.

______


languageserver/Makefile
```diff
@@ -30,3 +30,7 @@ build:
 .PHONY: test
 test:
 	cd test && npm install && npm t
+
+.PHONY: wasm
+wasm:
```
turbolent, line 35: 👌

languageserver/server/server.go
```diff
@@ -1788,27 +1788,24 @@ func (s *Server) getDiagnostics(
 
 					importedLocationID := importedLocation.ID()
 
-					importedChecker, ok := s.checkers[importedLocationID]
```
turbolent, line 1791: 👍 

Maybe add a comment explaining that we re-check all imports for the moment and do not cache imported checkers, because we do not implement proper dependency tracking yet which would be necessary for proper cache invalidation. In addition, maybe also link to the new issue

languageserver/test/index.test.ts
```diff
@@ -221,4 +221,98 @@ describe("diagnostics", () => {
...
+          })
+        }
+
+        resolve(docsNotifications)
+      })
+
+    })
+  }
+
+  test("script with import", async() => {
```
turbolent, line 265: 👌

turbolent (approved):

PR #1579, opened 2022-04-13
Meter additional container length as a function of original size

dsainati1:
Arrays and dictionaries, being backed by atrees underneath, do not consume a linear amount of memory as they grow. Rather, the size of these structures increases faster and faster the larger they become, as more nodes need to be added at higher levels of the tree as the number of leaves increases. 

Based on discussion with @ramtinms, we can use `log(Count())` of the underlying tree to (over)estimate the amount of overhead necessary to grow the tree. 

______

### DIFF
--- a/runtime/common/metering.go
+++ b/runtime/common/metering.go
@@ -163,7 +163,6 @@ func NewDictionaryAdditionalSizeUsage(originalSize, additionalSize int) MemoryUs
 	}
 	return MemoryUsage{
 		Kind: MemoryKindDictionarySize,
-		// size of b+ tree grows logarithmically with the size of the tree
 		Amount: newAmount,
 	}
 }
--- a/runtime/common/metering.go
+++ b/runtime/common/metering.go
@@ -128,10 +128,10 @@ func NewArrayMemoryUsages(length int) (MemoryUsage, MemoryUsage) {
 		}
 }
 
-func NewArrayLengthUsage(length int) MemoryUsage {
+func NewArrayAdditionalLengthUsage(originalLength, additionalLength int) MemoryUsage {
 	return MemoryUsage{
 		Kind:   MemoryKindArrayLength,
-		Amount: uint64(length),
+		Amount: uint64(originalLength + additionalLength),
 	}
 }
 
@@ -145,10 +145,10 @@ func NewDictionaryMemoryUsages(length int) (MemoryUsage, MemoryUsage) {
 		}
 }
 
-func NewDictionarySizeUsage(length int) MemoryUsage {
+func NewDictionaryAdditionalSizeUsage(originalSize, additionalSize int) MemoryUsage {
 	return MemoryUsage{
 		Kind:   MemoryKindDictionarySize,
-		Amount: uint64(length),
+		Amount: uint64(originalSize + additionalSize),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


runtime/common/metering.go
```diff
@@ -145,10 +153,18 @@ func NewDictionaryMemoryUsages(length int) (MemoryUsage, MemoryUsage) {
...
+		newAmount = uint64(originalSize + additionalSize)
+	} else {
+		// size of b+ tree grows logarithmically with the size of the tree
+		newAmount = uint64(math.Log2(float64(originalSize)) + float64(additionalSize))
+	}
 	return MemoryUsage{
-		Kind:   MemoryKindDictionarySize,
-		Amount: uint64(length),
+		Kind: MemoryKindDictionarySize,
+		// size of b+ tree grows logarithmically with the size of the tree
```
robert-e-davidson3, line 166: Duplicate
```suggestion
```

robert-e-davidson3 (approved):

turbolent:
@dsainati1 let's please get at least one review from a code owner before merging in PRs

dsainati1:
Why did it even allow me to merge it at all? I thought a Code owner review was required to merge in the first place

runtime/common/metering.go
```diff
@@ -145,10 +153,17 @@ func NewDictionaryMemoryUsages(length int) (MemoryUsage, MemoryUsage) {
 		}
 }
 
-func NewDictionarySizeUsage(length int) MemoryUsage {
+func NewDictionaryAdditionalSizeUsage(originalSize, additionalSize int) MemoryUsage {
+	var newAmount uint64
```
turbolent, line 157: The code here is a duplicate of the above, maybe factor the logic out. We might forget to update one of the functions

runtime/common/metering.go
```diff
@@ -145,10 +153,17 @@ func NewDictionaryMemoryUsages(length int) (MemoryUsage, MemoryUsage) {
...
+	if originalSize <= 1 {
+		newAmount = uint64(originalSize + additionalSize)
```
turbolent, line 159: Does this assume `additionalSize` is always 1? If so, maybe remove the parameter and simplify to `newAmount = 1`. 

If not: What happens if `originalSize <= 1` and `additionalSize > 1`?  Shouldn't this still calculate the new number of intermediate nodes allocated (which won't be the same as the number of new elements)?

runtime/common/metering.go
```diff
@@ -145,10 +153,17 @@ func NewDictionaryMemoryUsages(length int) (MemoryUsage, MemoryUsage) {
...
+	} else {
+		// size of b+ tree grows logarithmically with the size of the tree
+		newAmount = uint64(math.Log2(float64(originalSize)) + float64(additionalSize))
```
turbolent, line 162: Followed up on the formula here, not quite sure log2 is a correct over-estimation for the atree data structures: 
https://axiomzen.slack.com/archives/C0355LG8V6J/p1649889273942269?thread_ts=1649706053.685719&cid=C0355LG8V6J

turbolent:
@dsainati1 Because you're merging into a non-`master` branch, only `master` has a code owner requirement

turbolent:
Added a branch protection rule for `feature/*`, same as `master`

PR #1585, opened 2022-04-14
[0.23] Add start of a Cadence analysis package

turbolent:
Work towards #1269 

⚠️ Targets v0.23 branch, so we can use this package already e.g. to detect breaking changes of the upcoming Cadence release. We'll want to still merge this to `master` too

## Description

- Add a `tools/analysis` package, inspired by the Go tools [`golang.org/x/tools/go/packages`](https://pkg.go.dev/golang.org/x/tools/go/packages) and [`golang.org/x/tools/go/analysis`](https://pkg.go.dev/golang.org/x/tools/go/analysis) packages.

  The package allows parsing and optional type checking of locations, with custom import resolution.

  For example, it can be used in the future to implement separate linters (analysis passes), or detect outdated code due to breaking changes.

- Add a command line tool which reads in contracts from a CSV file, parses and checks them, and runs analyzers on the programs
  - Currently there are no analyzers yet, but this is a skeleton for e.g. adding analyzers for upcoming breaking changes
  - This tool is already useful for parsing and checking Mainnet contracts and determining overall regressions / breakages
  - We can integrate this tool into the [source compatibility suite](https://github.com/onflow/cadence/tree/master/compat), which already analyzes programs for regressions (however, by pulling code from GitHub instead of a network)

- Also, refactor and fix the default predeclared values and types for Flow in the checker and interpreter for the REPL and CLI, and use this default behaviour in the analysis package

cc @sideninja @j1010001 

______


runtime/sema/resolve.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+type AddressContractNamesResolver func(address common.Address) ([]string, error)
+
+// AddressLocationHandlerFunc returns a location handler
+// which returns a single location for non-address locations,
+// and uses the given address contract names resolve function
+// to get all contract names for an address
+//
+func AddressLocationHandlerFunc(resolveAddressContractNames AddressContractNamesResolver) LocationHandlerFunc {
```
turbolent, line 33: This was extracted from the language server. Even though not used by the `runtime` package, it is still the "default" behaviour. A future PR refactor the LS to use this handler, too

sideninja:
Awesome!
I have some questions, is this analysis package within the analysis.go meant to be a single place to do analysis or could it be that we define a simple analyzer interface, something like: 
```go
type Analyzer interface {
	Analyze(*ast.Program, *sema.Elaboration) error
}
```
Then you could have multiple files in the package (e.g. `unnecessary-casting.go`) implementing a specific and straightforward analyzer for that purpose. All those could then be passed to the "main" analyzer and further configured. I hope this makes sense.

turbolent:
@sideninja Yes, the idea is to use this package to implement individual analyzers, not one large analyzer that detects multiple different things. 

Also, In #1587 I added functions to perform optimized AST traversal. If multiple analyzers are run on the same AST it may be worth it to pay the overhead/cost of using the inspector over normal repeated traversal of the AST, as subsequent traversals are much faster.

robert-e-davidson3 (approved):

PR #1586, opened 2022-04-15
feat: add alchemy API callout

alxflw:
This adds a link to the Alchemy API repo to make the NFT discoverable


albeethekid (approved):

joshuahannan (commented):
I don't really think this is the place for a callout like this. This is focused on teaching the basics of Cadence development, not as an explicit resource for app developers. If we want to put a link to a separate dapp development page [like this one](https://docs.onflow.org/dapp-development/) and put the alchemy link there, that is fine, but not directly here.

turbolent:
Agreed with @joshuahannan on this.

Linking to developer resources is great, but the Cadence tutorials are not the right place, they are meant to teach the language.

PR #1590, opened 2022-04-18
Adjust formulas for calculating memory usage of arithmetic operations on big integers

turbolent:
Based on https://www.notion.so/dapperlabs/Metering-arithmetic-operations-7acc852f89834e78911f8f42c6bb842d. 

Not complete, I left some questions in the document.

______

### DIFF
--- a/runtime/common/metering.go
+++ b/runtime/common/metering.go
@@ -250,18 +250,13 @@ func BigIntByteLength(v *big.Int) int {
 	return len(v.Bits()) * bigIntWordSize
 }
 
-func NewPlusBigIntMemoryUsage(a, b *big.Int) MemoryUsage {
-	maxWordLength := max(
-		len(a.Bits()),
-		len(b.Bits()),
-	)
-	return NewBigIntMemoryUsage(
-		(maxWordLength + 4) *
-			bigIntWordSize,
-	)
-}
+// big.Int memory metering:
+// - |x| is len(x.Bits()), which is the length in words
+//
 
-func NewMinusBigIntMemoryUsage(a, b *big.Int) MemoryUsage {
+func NewPlusBigIntMemoryUsage(a, b *big.Int) MemoryUsage {
+	// max(|a|, |b|) + 5
+
 	maxWordLength := max(
 		len(a.Bits()),
 		len(b.Bits()),
@@ -272,7 +267,26 @@ func NewMinusBigIntMemoryUsage(a, b *big.Int) MemoryUsage {
 	)
 }
 
+func NewMinusBigIntMemoryUsage(a, b *big.Int) MemoryUsage {
+	// max(|a|, |b|) + 4
+
+	maxWordLength := max(
+		len(a.Bits()),
+		len(b.Bits()),
+	)
+	return NewBigIntMemoryUsage(
+		(maxWordLength + 4) *
+			bigIntWordSize,
+	)
+}
+
 func NewMulBigIntMemoryUsage(a, b *big.Int) MemoryUsage {
+	// if min(|a|, |b|) <= 40:
+	//     |a| + |b| + 4
+	// else:
+	//     n = min(|a|, |b|)
+	//     3 * n + max(6 * n, |a| + |b|) + 8
+
 	aWordLength := len(a.Bits())
 	bWordLength := len(b.Bits())
 	minWordLength := min(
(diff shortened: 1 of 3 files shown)


runtime/common/metering.go
```diff
@@ -217,121 +232,170 @@ func max(a, b int) int {
...
+		resultWordLength = aWordLength
+	} else if b.Cmp(bigOneHundred) < 0 {
+		resultWordLength = aWordLength - bWordLength + 5
+	} else {
+		recursionCost := int(unsafe.Sizeof(uintptr(0))) +
+			9*bWordLength +
+			int(math.Floor(float64(aWordLength)/float64(bWordLength))) + 12
+		recursionDepth := 2 * b.BitLen()
+		resultWordLength = 3*bWordLength + 4 + recursionCost*recursionDepth
+	}
```
SupunS, line 313: Might be useful to have a comment with the used equations for reference purposes

SupunS (approved):

runtime/common/metering.go
turbolent: Good idea, added in [c094707](https://github.com/onflow/cadence/pull/1590/commits/c09470796f1fe6e1eec16cf8ab37c9724ac76e52)

tarakby:
FYI, I am back today and answered the questions [here](https://www.notion.so/dapperlabs/Metering-arithmetic-operations-7acc852f89834e78911f8f42c6bb842d).

runtime/common/metering.go
```diff
@@ -217,121 +232,255 @@ func max(a, b int) int {
...
+
+	var resultWordLength int
+	if a.Cmp(b) < 0 || b.Cmp(bigOne) == 0 {
+		resultWordLength = aWordLength + 4
+	} else if b.Cmp(bigOneHundred) < 0 {
+		resultWordLength = aWordLength - bWordLength + 5
+	} else {
+		recursionCost := int(unsafe.Sizeof(uintptr(0))) +
+			9*bWordLength +
+			int(math.Floor(float64(aWordLength)/float64(bWordLength))) + 12
```
tarakby, line 333: On the document, I wrote `Floor` because that's the math specific operation. In software languages, integer division implicitly returns that floor. 
We could replace `int(math.Floor(float64(aWordLength)/float64(bWordLength)))` by `aWordLength/bWordLength`.

runtime/common/metering.go
```diff
@@ -217,121 +232,255 @@ func max(a, b int) int {
...
-			BigIntByteLength(a),
-			BigIntByteLength(b),
-		),
+		resultWordLength * bigIntWordSize,
 	)
 }
 
+var invalidLeftShift = errors.New("invalid left shift of non-Int64")
+
 func NewBitwiseLeftShiftBigIntMemoryUsage(a, b *big.Int) MemoryUsage {
```
tarakby, line 420: I don't think the `shift` value should be a big integer, a `uint` is more than enough.
Then in the metering `b/word_size` would avoid using a big integer division ( I guess the metering itself should be very light and should not allocate memory)

runtime/common/metering.go
```diff
@@ -217,121 +232,255 @@ func max(a, b int) int {
...
+	//     |a| + b/word_size + 5
+
+	aWordLength := len(a.Bits())
+
+	var resultWordLength int
+	if b.Sign() == 0 {
+		resultWordLength = aWordLength + 4
+	} else {
+		// TODO: meter the allocation of the metering itself
+		shiftByteLengthBig := new(big.Int).Div(b, bigIntWordSizeAsBig)
```
tarakby, line 433: This the operation I mentioned in the last comment.

runtime/common/metering.go
```diff
@@ -217,121 +232,255 @@ func max(a, b int) int {
...
+	}
 	return NewBigIntMemoryUsage(
-		// TODO: https://github.com/dapperlabs/cadence-private-issues/issues/32
-		BigIntByteLength(a) +
-			BigIntByteLength(b),
+		resultWordLength * bigIntWordSize,
 	)
 }
 
 func NewBitwiseRightShiftBigIntMemoryUsage(a, b *big.Int) MemoryUsage {
```
tarakby, line 446: Same here for the `shift` value that should be `uint`

tarakby (commented):
I left a late review. 
- The code is overall correct except some small errors. 
- I also noticed some edge cases to tighten the upper bound were omitted (like zero big integers), maybe to simplify the code?
- If Cadence is accepting big integer shift values, it might make sense to update the API in the future to only accept 32 or 64-bits shift values.

PR #1591, opened 2022-04-19
Meter Atree DataSlab and MetaDataSlab directly

dsainati1:
Based on additional discussion after https://github.com/onflow/cadence/pull/1579, it became evident that metering length of arrays and dictionaries naively was not sufficient, and that it is also necessary to meter updates to array and dictionary elements. 

This PR redoes the array and dictionary metering completely to now meter `MemoryKindDataSlab` and `MemoryKindMetaDataSlab` instead of estimating the size of the atrees. We also now use a more accurate overestimate (detailed in the comments of this PR) to result in less overestimation. 

This also new tests. 

______


runtime/common/metering.go
```diff
@@ -134,17 +134,20 @@ func NewArrayMemoryUsages(length int) (MemoryUsage, MemoryUsage) {
 		}
 }
 
-func NewArrayAdditionalLengthUsage(originalLength, additionalLength int) MemoryUsage {
-	var newAmount uint64
-	if originalLength <= 1 {
-		newAmount = uint64(originalLength + additionalLength)
+func atreeAdditionalMemoryUsage(originalSize int) uint64 {
```
turbolent, line 137: Isn't the calculation based on the new size?

runtime/common/metering.go
```diff
@@ -134,17 +134,20 @@ func NewArrayMemoryUsages(length int) (MemoryUsage, MemoryUsage) {
...
+	if originalSize <= 1 {
+		return uint64(originalSize + 1)
```
turbolent, line 139: @fxamacker How many nodes does an atree data structure have at minimum? Is it the case that there is always at least one `MetaDataSlab` + one` DataSlab`? Or is it possible for a root to be just a single `DataSlab` (i.e. no `MetaDataSlab`, e.g. if the container is empty or the total size of the elements is smaller than the threshold)?

runtime/common/metering.go
dsainati1: Unsure. @fxamacker when splitting, is the worst case based on the depth of the original tree or the depth of the resulting tree?

runtime/common/metering.go
fxamacker: > How many nodes does an atree data structure have at minimum? 

At minimum, atree container has one DataSlab (leaf node) as root node.

> Is it the case that there is always at least one MetaDataSlab + one DataSlab? 

No, atree container can have 
- one DataSlab (leaf node) as root node, or 
- at least 3 nodes (2 DataSlabs + 1 MetadataSlabs) with MetadataSlab as root node
 
> Or is it possible for a root to be just a single DataSlab (i.e. no MetaDataSlab, e.g. if the container is empty or the total size of the elements is smaller than the threshold)?

Yes.  When total size of elements in the container is less than threshold,  DataSlab is the root node.

runtime/common/metering.go
fxamacker: > when splitting, is the worst case based on the depth of the original tree or the depth of the resulting tree?

The worst case is based on the depth of the original tree.

dsainati1: In that case this should be correct.

turbolent: Thanks for clarifying @fxamacker !

fxamacker (commented):
Nice work!  👍  I mainly focused on atree array metering.  After our meeting with Bastian yesterday, I looked into atree again and have a few thoughts.

As discussed, we know the formula is overestimation since we want to account for the worst case situation.  And it wasn't clear how to reduce the amount overestimated without introducing risk of it being too low for some edge case.

I think we might have a more realistic estimation for number of DataSlab (leaf nodes) and increase metering only when new DataSlab can be allocated in the worst case (which causes new MetaDataSlab to be allocated).

- If array is homogenous with fix-sized types, such as `Int`, number of leaf nodes can be computed as `num_of_element * element_size / default_slab_size`
- Otherwise (heterogenous array or array with variable-sized types), number of leaf nodes can be `num_of_element / 2` because non-root data slab will contain at least 2 elements.

In `atreeAddtionalMemoryUsage()`, we can compute number of leaf nodes with `originalSize` and number of leaf nodes with `newSize`.  If they differ, we can use the formula to estimate number of new nodes.  Otherwise, no new allocation.

Also, we may want to add atree metering in `NewArrayMemoryUsages()`.

Not sure, but since overhead for DataSlab and MetaDataSlab is different, do we want to have separate kind for them?

Since we are likely to overestimate, we might want to test some edge-cases to:
- repeatedly append and remove elements
- repeatedly update existing elements

fxamacker (commented):
Thanks for the update!  Nice work!  Suggestions focused on:
1.  Using Metered Max Count
2.  Formula to Compute Number of Branch Nodes
3.  Incrementing Slab Counts

@turbolent @dsainati1 - Especially in Using Metered Max Count, please let me know if the suggested change missed any edge cases.
   
### Using Metered Max Count

I think when metering atree array/map, we might need to save max metered array/dict count, and pass it to `AdditionalAtreeMemoryUsage` as `meteredCount` in addition to `newCount`.
```
func AdditionalAtreeMemoryUsage(
    meteredCount uint64, 
    newCount uint64, 
    elementSize uint,
) (MemoryUsage, MemoryUsage)
```

Example sequence of activity:

1.  100 elements are added to an array
2.  100 elements are removed
3.  100 new element are added

If we just meter new elements based on current count and new count, we will meter 200 elements, effectively double the worst case.

If we use max metered element count as base, we will meter 100 elements for the example sequence of activity.

### Formula to Compute Number of Branch Nodes

I think we should use a new formula to compute number of branch nodes (MetadataSlab) in `atreeNodes()`.  

`atreeNodes()` should return the number of leaf nodes and number of branch nodes based on element count and element size.  

Old Formula: 
```Go
branchNodes = uint64(math.Ceil(math.Log2(float64(leafNodes))))
```
Old formula computes depth of a binary tree - 1.  It actually is number of new branch node by adding one more leaf node in the worst case.  

:warning: ~Suggested New Formula~ (sorry for suggesting this, I found this can under count): 
```Go
const n = 20 // n-ary tree
branchNodes = uint64(math.Ceil(float64(leafNodes - 1) / float64(n - 1)))
```

:point_right: Updated New Formula:

I revised formula for number of branch nodes to avoid under counting:
```Go
const n = 20 // n-ary tree
// Compute atree height from number of leaf nodes and n-ary
height := 1 + int(math.Ceil(math.Log(float64(leafNodeCount))/math.Log(float64(n))))

// Compute number of branch nodes from leaf node count and height
branchNodes := 0
for i := 1; i < height; i++ {
	branchNodes += int(math.Ceil(float64(leafNodes) / math.Pow(float64(n), float64(i))))
}
```

New formula computes total number of branch nodes from total number of leaf nodes with n-ary.  I think n is closer to 20 (conservatively).  

UPDATE:  ⚠️ Replaced suggested new formula with one that doesn't under count number of branch nodes.

### Incrementing Slab Counts

Maybe we don't need to update metering when element is updated, because

- If array/dict element static type is fix-sized, setting new value has no effect on number of atree slabs.
- Otherwise, we estimate for the worst-case already (2 elements per slab).

Also, I think `DictionaryValue.Insert()` handles element with existing key.  If so, we should only meter new atree nodes for non-existing key.

dsainati1:
> If we just meter new elements based on current count and new count, we will meter 200 elements, effectively double the worst case.

Does the atree not shrink when elements are removed? If we have 100 elements in it and we remove all of them, does the tree still have the same number of nodes?

runtime/interpreter/value.go
```diff
@@ -15487,7 +15520,15 @@ func (v *DictionaryValue) Insert(
...
-	common.UseMemory(interpreter, common.NewDictionaryAdditionalSizeUsage(int(v.dictionary.Count()), 1))
+	keySize := v.Type.KeyType.elementSize()
+	valueSize := v.Type.ValueType.elementSize()
+	var elementSize uint
+	if keySize != 0 && valueSize != 0 {
+		elementSize = keySize + valueSize
+	}
+	dataSlabs, metaDataSlabs := common.AdditionalAtreeMemoryUsage(v.dictionary.Count(), elementSize)
+	common.UseMemory(interpreter, dataSlabs)
+	common.UseMemory(interpreter, metaDataSlabs)
```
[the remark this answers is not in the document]
fxamacker: Great point! :+1:   Maybe we should add a comment to indicate we can overestimate if existing key is updated.

fxamacker:
> > If we just meter new elements based on current count and new count, we will meter 200 elements, effectively double the worst case.
> 
> Does the atree not shrink when elements are removed? If we have 100 elements in it and we remove all of them, does the tree still have the same number of nodes?

Yes, if we remove all elements, atree array shrinks to one DataSlab.

dsainati1:
> New formula computes total number of branch nodes from total number of leaf nodes with n-ary. I think n is closer to 20 (conservatively).

Is this always true even in the worst case? If we don't have any information about the element size of the values in the atree, we have to assume the worst case, otherwise there will be a possible exploit from filling an array or dictionary with large values and causing us to underestimate how many branch nodes are allocated.

fxamacker:
> > New formula computes total number of branch nodes from total number of leaf nodes with n-ary. I think n is closer to 20 (conservatively).
> 
> Is this always true even in the worst case? If we don't have any information about the element size of the values in the atree, we have to assume the worst case, otherwise there will be a possible exploit from filling an array or dictionary with large values and causing us to underestimate how many branch nodes are allocated.

@dsainati1 Yes, I think we are assuming the worst case: 
- only two elements in a non-root data slab when element size is unknown 
- 20-ary is the number of referenced child nodes in half filled MetadataSlab.
- the number of branch nodes is computed with two previous assumptions (in updated new formula).

:warning:  BTW, I discovered my suggested formula for number of branch nodes can under count, so I came up with an updated new formula.  See [updated summary with new formula using for-loop](https://github.com/onflow/cadence/pull/1591#pullrequestreview-952402335).

fxamacker (commented):
Nice work!

If we always want to meter new elements (even in the case of removing and adding elements), I think you can remove `if v.maxCount <= v.array.Count() {` and just call `AdditionalAtreeMemoryUsage()`.

BTW, I found my previously suggested formula can underestimate number of branch nodes (sorry!), so I replaced it with a new formula that uses a for-loop.

runtime/interpreter/value.go
```diff
@@ -14996,13 +15027,25 @@ func newDictionaryValueFromOrderedMap(
 	staticType DictionaryStaticType,
 ) *DictionaryValue {
 
-	baseUse, lengthUse := common.NewDictionaryMemoryUsages(int(dict.Count()))
+	keySize := staticType.KeyType.elementSize()
+	valueSize := staticType.ValueType.elementSize()
```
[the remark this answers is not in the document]
fxamacker: I see.  I think we check if `staticType != nil` in `NewArrayValue()`.  Do we want to remove it there?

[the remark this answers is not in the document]
dsainati1: This is because some tests manually create array values without specifying a static type, so to preserve this behavior of the tests without crashing I added it there, but in actual programs it should never be `nil`.

fxamacker (approved):
LGTM!  I mainly focused on non-test code.

runtime/interpreter/value.go
```diff
@@ -1638,7 +1645,13 @@ func (v *ArrayValue) RecursiveString(seenReferences SeenReferences) string {
 func (v *ArrayValue) Append(interpreter *Interpreter, getLocationRange func() LocationRange, element Value) {
 
 	// length increases by 1
-	common.UseMemory(interpreter, common.NewArrayAdditionalLengthUsage(int(v.array.Count()), 1))
+	dataSlabs, metaDataSlabs := common.AdditionalAtreeMemoryUsage(
```
SupunS, line 1648: Don't we need to incorporate elementOverhead for arrays, similar to dictionaries/composites?

dsainati1: No, according to @fxamacker's comment here: https://github.com/onflow/flow-go/pull/2357#discussion_r862923066 the overhead is only for the key/value pair of atree OrderedMaps, and thus arrays do not have the same overhead cost per element.

SupunS (approved):
LGTM!

runtime/common/metering.go
```diff
@@ -127,62 +130,114 @@ func NewConstantMemoryUsage(kind MemoryKind) MemoryUsage {
...
+		// will always contain at least two elements.
+		leafNodeCount = uint64(math.Ceil(float64(count) / 2))
+	}
+	if leafNodeCount < 1 {
+		leafNodeCount = 1 // there will always be at least one data slab
+	}
+
+	branchNodeCount = 0
+	if leafNodeCount >= 2 {
+		const n = 20 // n-ary tree
```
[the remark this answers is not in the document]
dsainati1: Yes, based on the discussion above this is a conservative overestimate of the branching factor.

runtime/interpreter/primitivestatictype.go
```diff
@@ -186,6 +186,82 @@ const (
...
+	case PrimitiveStaticTypeInt32,
+		PrimitiveStaticTypeUInt32,
+		PrimitiveStaticTypeWord32:
+		return 4
+	case PrimitiveStaticTypeInt64,
+		PrimitiveStaticTypeUInt64,
+		PrimitiveStaticTypeWord64,
+		PrimitiveStaticTypeFix64,
+		PrimitiveStaticTypeUFix64:
+		return 8
```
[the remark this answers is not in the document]
dsainati1: This is an estimation of the number of values of each type that can be packed into one (1024 byte) atree data slab, the idea being that we can use this information and the atree's `Count` in order to get a more accurate estimation of the number of data slabs in an atree. cc @fxamacker are the values stored in an atree `cbor` encoded; do we need to add the tag size here?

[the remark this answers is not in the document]
turbolent: Thank you for the clarification, adding a comment, and adjusting the values.

Last concern I have is that the `StaticType.elementSize` and `Value.ByteLength` functions become "out-of-sync". They return different values (the first returns the max, the second is based on the concrete value), but changing the latter could still cause the former to become outdated / under-estimating. 

It is hard to prevent this, but maybe a warning/remding comment in the `Value.ByteLength` implementations could help, e.g. `// NOTE: ensure to keep StaticType.elementSize in sync!`. Please feel free to add this in a follow-up PR, it's not blocking this one

turbolent (approved):

PR #1594, opened 2022-04-20
Fix common supertype calculation for optionals

### LINKED ISSUE #1592: `commonSuperTypeOfComposites` does not handle optionals
### Problem

`commonSuperTypeOfComposites` does not handle optional types.

For example, given a composite type `T`, finding the least-common super type of `T` and `T?` calls `commonSuperTypeOfComposites`, which assumes that all types are (non-optional composite types).

### Expected Behaviour

- The case of a type being optional should at least be handled gracefully, e.g. by resulting in `AnyStruct`/`AnyResource`, depending on the composite kind
- An improved solution would be to properly infer the "most-optional" type of the composite types. For example, for a type list `T`, `T??`, `T?`, the most-optional type would be `T??`, a common supertype of all types. 

### Steps to Reproduce

Subtest case reproducing the problem in `TestCommonSuperType`:

```go

	t.Run("optional composite types", func(t *testing.T) {
		t.Parallel()

		testLocation := common.StringLocation("test")

		ty := &CompositeType{
			Location:   testLocation,
			Identifier: "T",
			Kind:       common.CompositeKindStructure,
			Members:    NewStringMemberOrderedMap(),
		}

		optionalType := &OptionalType{
			Type: ty,
		}

		tests := []testCase{
			{
				name: "mixed",
				types: []Type{
					optionalType,
					ty,
				},
				expectedSuperType: optionalType,
			},
		}

		testLeastCommonSuperType(t, tests)
	})

```

SupunS:
Closes #1592

## Description

The optionals were not properly unwrapped before going into the struct-common-supertype finding path.

Solutions:
- Unwrap the optionals before going further
- Re-wrap the same amount of level of optionals, before returning (if not already contain optionals).

With this, supertype of `T`, `T?`, `T??` will be `T??`
______


runtime/sema/type_test.go
```diff
@@ -871,25 +871,6 @@ func TestCommonSuperType(t *testing.T) {
...
-			},
-			{
-				name: "optional with heterogeneous types",
-				types: []Type{
-					nilType,
-					Int8Type,
-					StringType,
-				},
-				expectedSuperType: AnyStructType,
-			},
```
SupunS, line 892: Moved these two tests to a dedicated test for optionals

runtime/sema/type_tags.go
```diff
@@ -870,3 +877,47 @@ func commonSuperTypeOfComposites(types []Type) Type {
...
+	for i := 0; i < levels; i++ {
+		typ = &OptionalType{
+			Type: typ,
+		}
+	}
+
+	return typ
+}
+
+func max(a, b int) int {
```
dsainati1, line 917: Does this not already exist in the standard library?

SupunS: stdlibs doesn't have a one for `int`. That one takes `float64` and its implementation is a bit heavy

turbolent (approved):
Great fix!

PR #1598, opened 2022-04-21
Meter string conversion of interpreter values

SupunS:

## Description

Introduced a `ToMeteredString()` method to be used only when string conversion is performed during the interpreter/runtime, where metering is needed.

Reduces the complexity of having to update all places where `RecursiveString()`/`String()` method is called, but metering is not needed. e.g: tooling, etc.
______


runtime/interpreter/function.go
```diff
@@ -101,6 +102,12 @@ func (f *InterpretedFunctionValue) RecursiveString(_ SeenReferences) string {
 	return f.String()
 }
 
+func (f *InterpretedFunctionValue) ToMeteredString(memoryGauge common.MemoryGauge, _ SeenReferences) string {
+	typeString := f.Type.String()
```
SupunS, line 106: We need to separately meter string conversions of runtime types. I'll open a subsequent PR for that

dsainati1 (approved):
Nice! This is a much better solution than requiring that `String` and `RecursiveString` have access to an interpreter.

robert-e-davidson3 (approved):

runtime/format/string.go
```diff
@@ -25,3 +25,28 @@ import (
...
+			// ASCII non-printable characters (i.e: out of range from space through DEL-1)
+			if 0x20 > r || r > 0x7E {
+				asciiNonPrintableChars++
+			}
+		}
+	}
+
+	// len = printableChars + (escapedChars x 2) + (nonPrintableChars x 8) + (quote x 2)
+	l := (len(s) - asciiNonPrintableChars - escapedChars) + escapedChars*2 + asciiNonPrintableChars*8 + 2
+	return l
```
turbolent, line 51:
```suggestion
	return (len(s) - asciiNonPrintableChars - escapedChars) + escapedChars*2 + asciiNonPrintableChars*8 + 2
```

runtime/interpreter/account.go
```diff
@@ -111,9 +112,11 @@ func NewAuthAccountValue(
 	}
 
 	var str string
-	stringer := func(_ SeenReferences) string {
+	stringer := func(memoryGauge common.MemoryGauge, _ SeenReferences) string {
 		if str == "" {
-			str = fmt.Sprintf("AuthAccount(%s)", address)
+			common.UseMemory(memoryGauge, common.NewRawStringMemoryUsage(13))
```
turbolent, line 117: Maybe add this to the list of known memory usages up top. Same for `PublicAccount`, `AuthAccount.Contracts`, etc below

turbolent (commented):
Sorry for the late review, missed this

runtime/interpreter/account.go
SupunS: Good idea! Added with 0b763c8a88be1f222c7e286b24f15a6a00db0f19

turbolent (approved):
Nice!

PR #1605, opened 2022-04-25
Separate metering for simple composites

dsainati1:
With the changes in https://github.com/onflow/cadence/pull/1591, it no longer makes sense to meter `SimpleCompositeValue`s (which do not have an underlying atree) like regular `CompositeValue`s, which do. 
______


runtime/common/metering.go
```diff
@@ -197,6 +197,13 @@ func NewCompositeMemoryUsages(size uint64, element_size uint) (MemoryUsage, Memo
 	}, leaves, branches
 }
 
+func NewSimpleCompositeMemoryUsage(length int) MemoryUsage {
+	return MemoryUsage{
+		Kind:   MemoryKindSimpleComposite,
+		Amount: uint64(length) + 1, // +1 to account for empty strings
```
SupunS, line 203: Why is an additional +1 needed?

runtime/interpreter/simplecompositevalue.go
```diff
@@ -56,10 +56,8 @@ func NewSimpleCompositeValue(
 	stringer func(SeenReferences) string,
 ) *SimpleCompositeValue {
 
-	baseUsage, dataUsage, metaDataUsage := common.NewCompositeMemoryUsages(uint64(len(fields)), 0)
-	common.UseMemory(inter, baseUsage)
-	common.UseMemory(inter, dataUsage)
-	common.UseMemory(inter, metaDataUsage)
+	common.UseMemory(inter, common.NewConstantMemoryUsage(common.MemoryKindCompositeBase))
```
SupunS, line 59: Is the base same for both composites and simple-composites? The fields of the two go-structs are different, so the weights also need to be different.

Also, the base memory-usage (i.e: `common.NewConstantMemoryUsage(common.MemoryKindCompositeBase)`) can be pre computed / 
 computed only once

docs/language/accounts.mdx
```diff
@@ -88,6 +88,19 @@ to the `prepare` phase of the transaction.
...
+      // Adds a public key to the account.
+      // The public key must be encoded together with their signature algorithm, hashing algorithm and weight.
+      // This method is currently deprecated and is available only for the backward compatibility.
+      // `keys.add` method can be use instead.
+      fun addPublicKey(_ publicKey: [UInt8])
+
+      // Revokes the key at the given index.
+      // This method is currently deprecated and is available only for the backward compatibility.
+      // `keys.revoke` method can be use instead.
+      fun removePublicKey(_ index: Int)
```
SupunS, line 102: I guess these should go away if re-sync with the feature-branch?

dsainati1: Ah yea, i will resync everything to fix this.

SupunS (approved):

PR #1610, opened 2022-04-28
Pool lexers to reduce allocations and improve performance

turbolent:
## Description

@simonhf pointed out that execution is allocating a lot of memory, specifically emitting tokens in the token stream (`append` on a slice). Pre-allocating an array reduces the overhead of append.

I then also realized that the token stream itself is allocated and not used after parsing, so we can reduce allocations by introducing an object pool for lexers.

This significantly reduces amount of allocated memory and also increases performance as a side-effect.

Also, improve the benchmarking workflow:
- Enable memory statistics for all benchmarks
- Sort benchmarks by name. This makes it easier to find benchmarks in the output (e.g. see performance improvement -> check memory benchmark results)

______

### DIFF
--- a/runtime/parser2/lexer/lexer.go
+++ b/runtime/parser2/lexer/lexer.go
@@ -103,6 +103,7 @@ func Lex(input string) TokenStream {
 		prevEndOffset: 0,
 		current:       EOF,
 		prev:          EOF,
+		tokens:        make([]Token, 0, 2048),
 	}
 	l.run(rootState)
 	return l
--- a/runtime/parser2/lexer/lexer.go
+++ b/runtime/parser2/lexer/lexer.go
@@ -20,6 +20,7 @@ package lexer
 
 import (
 	"fmt"
+	"sync"
 	"unicode/utf8"
 
 	"github.com/onflow/cadence/runtime/ast"
@@ -95,16 +96,35 @@ func (l *lexer) Revert(cursor int) {
 	l.cursor = cursor
 }
 
+func (l *lexer) clear() {
+	l.startOffset = 0
+	l.endOffset = 0
+	l.prevEndOffset = 0
+	l.current = EOF
+	l.prev = EOF
+	l.canBackup = false
+	l.startPos = position{line: 1}
+	l.cursor = 0
+	l.tokens = l.tokens[:0]
+	l.tokenCount = 0
+}
+
+func (l *lexer) Reclaim() {
+	pool.Put(l)
+}
+
+var pool = sync.Pool{
+	New: func() interface{} {
+		return &lexer{
+			tokens: make([]Token, 0, 2048),
+		}
+	},
+}
+
 func Lex(input string) TokenStream {
-	l := &lexer{
-		input:         input,
-		startPos:      position{line: 1},
-		endOffset:     0,
-		prevEndOffset: 0,
-		current:       EOF,
-		prev:          EOF,
-		tokens:        make([]Token, 0, 2048),
-	}
+	l := pool.Get().(*lexer)
+	l.clear()
+	l.input = input
 	l.run(rootState)
 	return l
 }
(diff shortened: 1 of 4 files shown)


runtime/parser2/lexer/lexer.go
```diff
@@ -95,15 +96,35 @@ func (l *lexer) Revert(cursor int) {
...
-	l := &lexer{
-		input:         input,
-		startPos:      position{line: 1},
-		endOffset:     0,
-		prevEndOffset: 0,
-		current:       EOF,
-		prev:          EOF,
-	}
+	l := pool.Get().(*lexer)
+	l.clear()
```
SupunS, line 126: Wouldn't it be better for the `Reclaim()` to do the clearing (i.e: `l.clear()`) before putting it back in the pool? Then the memory (e.g: tokens array) is freed up while the lexer is in the pool.

SupunS (commented):
Nice optimization! 👌    Just a one question

janezpodhostnik (approved):
LGTM.

runtime/parser2/lexer/lexer.go
turbolent: But wouldn't that defeat the optimization, re-using the tokens array? Allocating the lexer itself isn't the expensive part, it is the tokens array in it.

SupunS: ah right, the array isn't cleared. I misread the 'slicing' as 'array re-initialization'.

turbolent: But yeah, I think the clearing / initialization is better to do after getting the item from the pool, for two reasons:
- It mimics initialization like the "construction" would do
- The returned item may not be a recycled one, but a newly constructed one (the "construction function" passed to the the pool), so we would have to call clear/init in two places: when the lexer is constructed, and when it is to be reclaimed

SupunS (approved):

runtime/parser2/lexer/lexer.go
```diff
@@ -95,15 +96,35 @@ func (l *lexer) Revert(cursor int) {
...
+}
+
+func (l *lexer) Reclaim() {
+	pool.Put(l)
+}
+
+var pool = sync.Pool{
+	New: func() interface{} {
+		return &lexer{
+			tokens: make([]Token, 0, 2048),
```
simonhf, line 119: @turbolent sorry, got here after the PR has been merged. But left a comment here [1] suggesting that the `2048` be increased 10 fold to e.g. `20480`. Why? Already found a contract which increases the token capacity to almost 16,000...

[1] https://axiomzen.slack.com/archives/CG0B7CJAJ/p1652314319091899?thread_ts=1651116257.517069&cid=CG0B7CJAJ

turbolent: We should probably not optimize for large cases, but for the average. Otherwise we have to pay for this large memory usage for all parses, which often encounter way fewer tokens

simonhf: @turbolent Thinking out loud: What do you think about the idea (for a future PR) of limiting the total number of lexers to e.g. some multiple of the number of CPU cores on the box? So on mainnet ENs it might be e.g. 2 * 128 = 256 possible lexer instances? In theory this would limit the memory footprint for the token slices and stop the ENs from being overwhelmed by a surge of concurrent Cadence script lexing via GRPC requests? There would have to be some queuing mechanism so that if all n lexer instances are currently in use, then the n+1 wannabe user would be queued and wait for one of the existing lexer instances to finish?

turbolent: Limiting the number of concurrent executions, e.g. for scripts, sounds like a good idea. However, I would recommend implementing this at a higher level than the lexer, even higher than Cadence itself, i.e. in the FVM/EN.

simonhf: It would be good to revisit this idea after this PR hits mainnet and we can see the effect on the lexer garbage production for a while. We'll then know if further optimization is necessary to curb the lexer garbage production.

Who would be a good person to approach to implement this at the higher level? Maybe @Kay-Zee has suggestions? :-)

PR #1611, opened 2022-04-28
Meter import/export of cadence values and types

robert-e-davidson3:
Closes https://github.com/dapperlabs/cadence-private-issues/issues/38

## Description

Metering import and export of the cadence.Value and cadence.Type data structures.

Added expected weights to notion doc: https://www.notion.so/dapperlabs/3686f94810664b51833ba922ecde5552?v=bd3091ef17a04684ab96d4bc97d9d492
______


values.go
```diff
@@ -1286,10 +1636,23 @@ type Capability struct {
...
+	common.UseConstantMemory(gauge, common.MemoryKindCadenceCapability)
+	return NewUnmeteredCapability(path, address, borrowType)
+}
+
 func (Capability) isValue() {}
 
-func (Capability) Type() Type {
-	return CapabilityType{}
+func (v Capability) Type(gauge common.MemoryGauge) Type {
+	return NewCapabilityType(gauge, v.BorrowType) // TODO is this correct? original was nil for BorrowType
```
robert-e-davidson3, line 1655: I need advice here @turbolent.

encoding/json/decode.go
```diff
@@ -333,12 +355,19 @@ func (d *Decoder) decodeInt64(valueJSON interface{}) cadence.Int64 {
 		panic(ErrInvalidJSONCadence)
 	}
 
-	return cadence.NewInt64(i)
+	return cadence.NewInt64(d.gauge, i)
 }
 
 func (d *Decoder) decodeInt128(valueJSON interface{}) cadence.Int128 {
 	bigInt := d.decodeBigInt(valueJSON)
```
turbolent, line 362: The current code is a good start, in the future we actually also need to meter allocations here: The JSON string is converted to a bigint value, which already allocates, which should go into the function below:
https://github.com/onflow/cadence/pull/1611/files#diff-486c5071ade706b28d51beacaf8b7cbe4267a892e708877e4ef37e3cfbc464d4R367.

We can address this in a follow-up PR though, not a blocker for this one

turbolent (commented):
It looks like this depends on the static type metering?

encoding/json/decode.go
robert-e-davidson3: I went ahead and did this change since I needed to fix up the PR anyway.

robert-e-davidson3:
I redid the branch so it comes off of `feature/memory-metering`.

turbolent:
This changes the API for the "public API", the `cadence` package, so labeling as a breaking change

turbolent (approved):
Nice work!

@SupunS could you please have a last look?

SupunS (commented):
Nice! changes looks good!

I have one question/suggestion wrt API change: Since the value and type creation APIs are used across different toolings (sdk, etc.) and can impact many downstream codes, what if we keep the current constructors as is, and introduce a new 'metered' constructor.

e.g: for `UInt8`, keep `NewUInt8()` as is, and introduce a new constructor `NewMeteredUInt8(memoryGauge)` to be used internally within cadence. So the external users don't have any impact. This might be just a simple renaming, on top of this PR

robert-e-davidson3:
> Nice! changes looks good!
> 
> I have one question/suggestion wrt API change: Since the value and type creation APIs are used across different toolings (sdk, etc.) and can impact many downstream codes, what if we keep the current constructors as is, and introduce a new 'metered' constructor.
> 
> e.g: for `UInt8`, keep `NewUInt8()` as is, and introduce a new constructor `NewMeteredUInt8(memorGauge)` to be used internally within cadence. So the external users don't have any impact. This might be just a simple renaming, on top of this PR

I like that idea.

values.go
turbolent: Yep, this makes sense 👍 It was previously not included

turbolent:
@SupunS Great idea! That might reduce / avoid breakage 👍

values.go
```diff
@@ -55,10 +56,15 @@ func NewVoid() Void {
...
+	common.UseConstantMemory(memoryGauge, common.MemoryKindCadenceVoid)
+	return NewVoid()
+}
+
 func (Void) isValue() {}
 
-func (Void) Type() Type {
-	return VoidType{}
+func (Void) Type(gauge common.MemoryGauge) Type {
+	return NewVoidType(gauge)
```
SupunS, line 67: Maybe we can introduce a new method here as well, keeping the old one as it is, similar to constructors.
Can do it in a subsequent PR.

SupunS (approved):

values.go
robert-e-davidson3: Oh good point - that also alters the interface.

A second PR sounds good to me. This wouldn't be as easy a change as the constructor renamings.

robert-e-davidson3: https://github.com/dapperlabs/cadence-private-issues/issues/46

PR #1613, opened 2022-04-29
Improve constant-sized type parsing

turbolent:
## Description

Optimize parsing of the size in constant-sized types by directly checking for an integer literal.

______


dsainati1 (approved):

robert-e-davidson3 (approved):

SupunS (approved):

SupunS (approved):
Thanks for adding the negative literal test. I initially did wonder what would happen for the negative literals, but realized it would still be handled, because the next token is the not the 'literal'. Always nice to have a test case 👌

PR #1614, opened 2022-04-29
Suggest optional chaining when member access is on optional type

turbolent:
## Description

Took 10 minutes to improve the developer experience: 

Developers often attempt to access a member of an optional type. 
The existing error is correct (the optional type does not have the requested member), but is not helpful.

When then the unwrapped type has the requested member, suggest optional chaining instead of a direct access.

Also, increase coverage for all errors, by producing their secondary error, if available.

______

### DIFF
--- a/runtime/sema/errors.go
+++ b/runtime/sema/errors.go
@@ -703,6 +703,13 @@ func (e *NotDeclaredMemberError) Error() string {
 }
 
 func (e *NotDeclaredMemberError) SecondaryError() string {
+	if optionalType, ok := e.Type.(*OptionalType); ok {
+		members := optionalType.Type.GetMembers()
+		name := e.Name
+		if _, ok := members[name]; ok {
+			return fmt.Sprintf("type is optional, consider optional-chaining: ?.%s", name)
+		}
+	}
 	return "unknown member"
 }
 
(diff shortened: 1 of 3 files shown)


bluesign:
Just commenting to learn more, why we don't suggest unwrapping here ?

Consider below struct: 

```
pub struct S {
    pub  fun a():String {
        return ""
    }
}

Currently we suggest this usage:
```
let s: S? = S()
var r:String = s?.a()!
```

vs 

```
let s: S? = S()
var r:String = s!.a()
```

turbolent:
There are two parts to your question:
1. Suggestions for accessing the member (field or function) of the optional value (e.g. access `a` of `s`)
2. Suggestions for getting the non-optional result of the access (e.g. get `String` from `String?` result)

This PR only tries to address 1, adding a suggestion for accessing the member. Adding a suggestion for getting the non-optional result of accessing the member could be done in a follow up PR / is out of scope.

In general, force-casting is a code-smell, and should only be performed if the developer knows the value will be non-nil. 

As the checker cannot know for either case 1 or 2 if the value is non-nil, it is better to recommend optional-chaining, or e.g. optional-binding with `if-let`.

robert-e-davidson3 (approved):

SupunS (approved):

PR #1615, opened 2022-05-02
Adding a github actions workflow to trigger the doc site refresh endpoint

codingone21:
Closes #???

## Description

* Used tj-actions/changed-files@v18.7 to get all the doc files that were edited
* Logs more detailed information (added, deleted, modified) for the diff files
* Calls the doc site's refresh endpoint with the content paths, SHA, and repository name

TODO: 
* The Github Action job will trigger the locally exposed WEB_URL (env) which is a temporary solution until the web server is live and public.
* This workflow will likely be copied to other repos related to the doc site
______

### DIFF
--- a/.github/workflows/refresh-doc.yml
+++ b/.github/workflows/refresh-doc.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - hannah-github-actions
 
 env:
   WEB_URL: 'https://nasty-penguin-86.loca.lt' # This needs to be updated to the live web server!
--- a/.github/workflows/refresh-doc.yml
+++ b/.github/workflows/refresh-doc.yml
@@ -2,11 +2,10 @@ name: refresh-doc
 on:
   push:
     branches:
-      - main
       - hannah-github-actions
 
 env:
-  WEB_URL: 'https://tender-bird-86.loca.lt' # This needs to be updated to the live web server!
+  WEB_URL: 'https://cold-pigs-fail-174-94-59-90.loca.lt' # This needs to be updated to the live web server!
 
 jobs:
   build:
--- a/.github/workflows/refresh-doc.yml
+++ b/.github/workflows/refresh-doc.yml
@@ -1,0 +1,75 @@
+name: refresh-doc
+on:
+  push:
+    branches:
+      - main
+
+env:
+  WEB_URL: 'https://nasty-penguin-86.loca.lt' # This needs to be updated to the live web server!
+
+jobs:
+  build:
+    runs-on: macos-latest
+    steps:
+    - uses: actions/checkout@v3
+      
+    - name: Changed Files
+      id: changed-files
+      uses: tj-actions/changed-files@v18.7
+      with: # Any file in the docs
+        files: |
+          docs/**
+    
+    # --- Log all diff info ---
+    - name: Log the files commit SHA if any md files have diffs
+      if: steps.changed-files.outputs.any_modified == 'true'
+      run: |
+        for file in ${{ steps.changed-files.outputs.all_modified_files }}; do
+          echo "$file was modified" with the commit ${{ github.sha }}
+        done
+
+    - name: Log any added (A) files
+      if: steps.changed-files.outputs.added_files != ''
+      run: |
+        for file in ${{ steps.changed-files.outputs.added_files }}; do
+          echo "$file was added"
+        done
+
+    - name: Log any deleted (D) files
+      if: steps.changed-files.outputs.any_deleted == 'true'
+      run: |
+        for file in ${{ steps.changed-files.outputs.deleted_files }}; do
+          echo "$file was deleted"
+        done
+
+    - name: Log any modified (M) files
+      if: steps.changed-files.outputs.modified_files != ''
+      run: |
+        for file in ${{ steps.changed-files.outputs.all_modified_files }}; do
+          echo "$file was modified"
+        done
+
+    # --- CALL REFRESH ENDPOINT USING PYTHON ---
+    - name: Setup Python
+      run: pip install requests
+    - name: Compile all files with diff into a request body using Python
+      if: steps.changed-files.outputs.any_modified == 'true'
+      run: |
+        import requests
+
+        list_diff_paths = "${{ steps.changed-files.outputs.all_modified_files }}".split(' ')
+        list_added_paths = "${{ steps.changed-files.outputs.added_files }}".split(' ')
+        list_deleted_paths = "${{ steps.changed-files.outputs.deleted_files }}".split(' ')
+        list_modified_paths = "${{ steps.changed-files.outputs.modified_files }}".split(' ')
+
+        print("All Diff", list_diff_paths)
+        print("A", list_added_paths, "D", list_deleted_paths, "M", list_modified_paths)
+
+        response = requests.post('${{ env.WEB_URL }}/action/refresh', data = '{ \
+          "repo": "${{ github.event.repository.name }}", \
+          "contentPaths": list_diff_paths, \
+          "commitSha": ${{ github.sha }} \
+        }')
+        
+        print("Request Response", response)
+      shell: python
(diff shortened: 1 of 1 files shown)


.github/workflows/refresh-doc.yml
```diff
@@ -0,0 +1,75 @@
...
+        list_added_paths = "${{ steps.changed-files.outputs.added_files }}".split(' ')
+        list_deleted_paths = "${{ steps.changed-files.outputs.deleted_files }}".split(' ')
+        list_modified_paths = "${{ steps.changed-files.outputs.modified_files }}".split(' ')
+
+        print("All Diff", list_diff_paths)
+        print("A", list_added_paths, "D", list_deleted_paths, "M", list_modified_paths)
+
+        response = requests.post('${{ env.WEB_URL }}/action/refresh', data = '{ \
+          "repo": "${{ github.event.repository.name }}", \
+          "contentPaths": list_diff_paths, \
```
justinbarry, line 70: I'm not familiar with how python works. Will it interpolate list_diff_paths in json format here?

.github/workflows/refresh-doc.yml
```diff
@@ -0,0 +1,75 @@
...
+      
+    - name: Changed Files
+      id: changed-files
+      uses: tj-actions/changed-files@v18.7
+      with: # Any file in the docs
+        files: |
+          docs/**
+    
+    # --- Log all diff info ---
+    - name: Log the files commit SHA if any md files have diffs
```
justinbarry, line 24: Is this debug code?

.github/workflows/refresh-doc.yml
```diff
@@ -0,0 +1,75 @@
+name: refresh-doc
+on:
+  push:
+    branches:
+      - hannah-github-actions
+
+env:
+  WEB_URL: 'https://cold-pigs-fail-174-94-59-90.loca.lt' # This needs to be updated to the live web server!
```
justinbarry, line 8: We can use something like https://requestbin.net/r/b9atr5pk?inspect temporarily until the endpoint is up.

.github/workflows/refresh-doc.yml
```diff
@@ -0,0 +1,75 @@
...
+        print("A", list_added_paths, "D", list_deleted_paths, "M", list_modified_paths)
+
+        response = requests.post('${{ env.WEB_URL }}/action/refresh', data = '{ \
+          "repo": "${{ github.event.repository.name }}", \
+          "contentPaths": list_diff_paths, \
+          "commitSha": ${{ github.sha }} \
+        }')
+        
+        print("Request Response", response)
+      shell: python
```
justinbarry, line 75: If the webrequest fails for some reason with it fail the build in this repo? If so, is that the behavior we want?

justinbarry (dismissed):
Made a few comments. Feel free to reach out with any questions.

.github/workflows/refresh-doc.yml
```diff
@@ -0,0 +1,75 @@
+name: refresh-doc
+on:
+  push:
+    branches:
+      - hannah-github-actions
```
turbolent, line 5: I guess this should be changed to `master` once ready for merging?

```suggestion
      - master
```

.github/workflows/refresh-doc.yml
```diff
@@ -0,0 +1,75 @@
...
+
+env:
+  WEB_URL: 'https://cold-pigs-fail-174-94-59-90.loca.lt' # This needs to be updated to the live web server!
+
+jobs:
+  build:
+    runs-on: macos-latest
```
turbolent, line 12: Is macOS required? Can this also run on a Linux machine?

turbolent:
Thank you for the PR @codingone21! 

Could you please follow and fill out the checklist?

.github/workflows/refresh-doc.yml
codingone21: Good catch! I had to reformat this to be string-concatenated instead in my new fix.

.github/workflows/refresh-doc.yml
codingone21: I added all the logging for viewing purposes! I think when Mackenzie comes back we might change the payload to be more descriptive (modified vs added vs deleted), so I just logged them for now!

.github/workflows/refresh-doc.yml
codingone21: Totally! Could you help me set this up?

.github/workflows/refresh-doc.yml
codingone21: I'll add the condition "continue-on-error: true" to prevent this! I thought I did but I might have removed it while debugging.

.github/workflows/refresh-doc.yml
codingone21: Eventually yes, but atm I don't see a huge reason to change this to master, since we don't have the live web server working anyways. I think we can still do all the testing on this branch and make little changes in the markdown files for testing before relying on the master branch.

.github/workflows/refresh-doc.yml
codingone21: Will update this to 'ubuntu-latest'

codingone21:
> Thank you for the PR @codingone21!
> 
> Could you please follow and fill out the checklist?

Yep! I learned that my commit message and branch name doesn't follow the guidelines, and I'll keep this in mind for future PRs.

.github/workflows/refresh-doc.yml
justinbarry: Will do!

.github/workflows/refresh-doc.yml
```diff
@@ -0,0 +1,90 @@
...
+        list_added_paths = "${{ steps.changed-files.outputs.added_files }}".split(' ')
+        list_deleted_paths = "${{ steps.changed-files.outputs.deleted_files }}".split(' ')
+        list_modified_paths = "${{ steps.changed-files.outputs.modified_files }}".split(' ')
+
+        print("All Diff", list_diff_paths)
+        print("A", list_added_paths, "D", list_deleted_paths, "M", list_modified_paths)
+
+        # Format path to string
+        list_diff_paths_formatted = ', '.join('"{0}"'.format(p) for p in hardcoded_diff_paths)
+        list_diff_paths_formatted = '[' + hardcoded_diff_paths_formatted + ']'
```
turbolent, line 80: Maybe better to just use a JSON encoder here to generate the JSON data, e.g. [`json.dumps`](https://docs.python.org/3/library/json.html#json.dumps)

.github/workflows/refresh-doc.yml
```diff
@@ -0,0 +1,90 @@
...
+
+        # Format payload
+        payload = '{"repo": "${{ github.event.repository.name }}", "contentPaths": ' + list_diff_paths_formatted + ', "commitSha": "${{ github.sha }}"}'
```
turbolent, line 83: Maybe better to just use a JSON encoder here to generate the JSON data, e.g. [`json.dumps`](https://docs.python.org/3/library/json.html#json.dumps), e.g. 

```python
payloads = json.dumps({"repo": ...
```

.github/workflows/call-refresh-doc.yml
```diff
@@ -0,0 +1,13 @@
...
+on:
+  push:
+    branches:
+      - hannah-github-actions
+
+jobs:
+  call-workflow-passing-data:
+    uses: onflow/cadence/.github/workflows/refresh-doc.yml@hannah-github-actions
+    with:
+      username: codingone21
```
turbolent, line 11: Is this for testing purposes? Should this go into `master`?

codingone21: This should eventually be replaced to master in a later PR, but since we are still in testing stages, I wanted to keep the changes in my branch! I don't think we would need this on master yet.

codingone21: I ended using @master for the yaml file, since I do most of the testing locally anyways. Didn't want to complicate it too much by making different versions of the same branch 😅

turbolent (approved):
Looks good! Cannot judge if this is works, but let's try it out :-)

PR #1616, opened 2022-05-02
 Meter variable activation

SupunS:
Work towards https://github.com/dapperlabs/cadence-private-issues/issues/35

## Description

Adds metering for variable activation creation cost.

______


dsainati1 (approved):

robert-e-davidson3 (approved):

runtime/activations/activations.go
```diff
@@ -115,6 +123,14 @@ func (a *ValueTypeActivation) Set(name string, value ValueType) {
 //
 type ValueTypeActivations struct {
 	activations []*ValueTypeActivation
+	memoryGauge common.MemoryGauge
+}
+
+func NewValueTypeActivations(memoryGauge common.MemoryGauge) *ValueTypeActivations {
+	// No need to meter.
+	return &ValueTypeActivations{
```
turbolent, line 131: This does not seem to be used? 

In case it is: Why do we not need to meter it?

runtime/compiler/local_activations.go
```diff
@@ -113,6 +121,14 @@ func (a *LocalActivation) Set(name string, value *Local) {
 //
 type LocalActivations struct {
 	activations []*LocalActivation
+	memoryGauge common.MemoryGauge
+}
+
+func NewLocalActivations(memoryGauge common.MemoryGauge) *LocalActivations {
+	// No need to meter.
+	return &LocalActivations{
```
turbolent, line 129: Same question as above

runtime/activations/activations.go
SupunS: This is the template code used for code-generation.

In the variable activation case, 'activation list' is created only once per interpreter. So we don't need to meter it.

I didn't want to add the word 'interpreter' to the template, and wanted to keep the comment also generic. Maybe can say  "created only once per execution" in the comment. Or do you have any suggestions?

turbolent: Oh, I didn't realize this was the templated code, thank you for clarifying 👍 

Maybe this function should be just called `NewUnmetered...` and not take a memory gauge, given that it does not meter?

runtime/compiler/local_activations.go
```diff
@@ -198,7 +214,7 @@ func (a *LocalActivations) Pop() {
 func (a *LocalActivations) CurrentOrNew() *LocalActivation {
 	current := a.Current()
 	if current == nil {
-		return NewLocalActivation(nil)
+		return NewLocalActivation(a.memoryGauge, nil)
```
turbolent, line 217: Unrelated to this particular PR:
It's interesting that the tests do not exercise this code path and similar cases below. 

I guess because we always have an initial activation in the activations before using/accessing the activations? 
Maybe we can remove the nil cases.

runtime/activations/activations.go
SupunS: Even though this constructor doesn't meter `Activations` , it keeps a ref to the memory gauge, and uses it to meter each new `Activation` and activation-entries initialization. This was done to avoid having to pass down the memory-gauge for each method.
i,e: it doesn't meter the outer list, but meter each entry in the list

turbolent: Ah, OK. Could you please add this as a comment to the function?

turbolent (approved):

runtime/compiler/local_activations.go
SupunS: Yup, might be possible. I'll add this in a separate PR with some tests, to keep this one to a minimal changes

PR #1620, opened 2022-05-04
Limit number of tokens which can be replayed (reparsed)

turbolent:
## Description

The parser may reparse parts of the program in cases of ambiguity.

Limit this to a sensible number of tokens. We may want to adjust this limit.

I validated this limit by parsing all Mainnet contracts and all of them remain parsable.

______

### DIFF
--- a/runtime/parser2/parser.go
+++ b/runtime/parser2/parser.go
@@ -250,6 +250,7 @@ func (p *parser) replayBuffered() {
 	backtrackCursor := p.backtrackingCursorStack[lastIndex]
 
 	replayedCount := p.replayedTokensCount + uint(cursor-backtrackCursor)
+	// Check for overflow (uint) and for exceeding the limit
 	if replayedCount < p.replayedTokensCount || replayedCount > tokenReplayLimit {
 		panic(fmt.Errorf("program too ambiguous, replay limit of %d tokens exceeded", tokenReplayLimit))
 	}
--- a/runtime/parser2/parser.go
+++ b/runtime/parser2/parser.go
@@ -45,6 +45,8 @@ type parser struct {
 	backtrackingCursorStack []int
 	// bufferedErrorsStack is the stack of parsing errors encountered during buffering
 	bufferedErrorsStack [][]error
+	// replayedCount is the number of replayed tokens
+	replayedCount uint
 }
 
 // Parse creates a lexer to scan the given input string,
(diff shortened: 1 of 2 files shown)


runtime/parser2/parser.go
```diff
@@ -229,13 +231,27 @@ func (p *parser) acceptBuffered() {
 	}
 }
 
+// replayLimit is a sensible limit for how many tokens may be replayed
+// while parsing a program
+const replayLimit = 2 << 12
```
dsainati1, line 236: How did we come up with this value? 

It seems like this is setting a total limit on the amount of backtracking for the whole program. Would it not be better to instead impose a (smaller) limit on the amount of backtracking that can happen at once? I.e. impose a limit for each backtracking cursor on the stack instead of a total limit for the stack as a whole?

This has the benefit of being able to localize errors for users should they encounter them; instead of complaining that the entire program is too ambiguous at some arbitrary location where we pass the total limit, we can error at the actual location that has too much backtracking.

turbolent: I determined it mostly through experimentation, there is no perfect value. We need to find a limit that rejects invalid programs as quickly as possible, while not rejecting valid programs.

There are problems with both local and global limits. With the global limit, the user may encounter the limit in a locally small ambiguity, because the parser already re-parsed too many tokens before. With the local limit, finding a sensible limit is hard: A small value will reject too many programs, a slightly larger value will not reject invalid programs quickly.

Maybe we can have both local and global limits to balance user experience and security.

turbolent: Another idea: Maybe we can enforce the limit locally to each top-level ambiguity, instead of the whole program? We could reset `replayedCount` in `acceptBuffered`

turbolent: Made the token replay limit "local" to each top-level reparse in f2c81d9. Nested replays count towards the top-most buffering, so errors stay local.

dsainati1 (approved):
Nice! Thanks for keeping the errors local.

runtime/parser2/parser.go
```diff
@@ -227,15 +230,32 @@ func (p *parser) acceptBuffered() {
...
 	// Pop the last backtracking cursor from the stack
 	// and revert the lexer back to it
 
 	lastIndex := len(p.backtrackingCursorStack) - 1
-	cursor := p.backtrackingCursorStack[lastIndex]
-	p.tokens.Revert(cursor)
+	backtrackCursor := p.backtrackingCursorStack[lastIndex]
+
+	replayedCount := p.replayedTokensCount + uint(cursor-backtrackCursor)
+	if replayedCount < p.replayedTokensCount || replayedCount > tokenReplayLimit {
```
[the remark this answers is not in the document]
turbolent: Oh, that's a good question! It is to check for overflow (unlikely, but still). I'll add a comment 👍

turbolent, line 253:
```suggestion
	// Check for overflow (uint) and for exceeding the limit
	if replayedCount < p.replayedTokensCount || replayedCount > tokenReplayLimit {
```

SupunS (approved):

PR #1624, opened 2022-05-05
Add element overhead to atree array metering

SupunS:

______


dsainati1 (approved):

runtime/interpreter/value.go
```diff
@@ -2305,8 +2311,9 @@ func (v *ArrayValue) Clone(interpreter *Interpreter) Value {
 		panic(ExternalError{err})
 	}
 
-	baseUsage, dataSlabs, metaDataSlabs := common.NewArrayMemoryUsages(v.array.Count(), v.elementSize)
+	baseUsage, elementUsage, dataSlabs, metaDataSlabs := common.NewArrayMemoryUsages(v.array.Count(), v.elementSize)
 	common.UseMemory(interpreter, baseUsage)
+	common.UseMemory(interpreter, elementUsage)
 	common.UseMemory(interpreter, dataSlabs)
 	common.UseMemory(interpreter, metaDataSlabs)
```
fxamacker, line 2318: Do we need to meter memory in `ArrayValue.Clone()`?

```
	// Clone returns a new value that is equal to this value.
	// NOTE: not used by interpreter, but used externally (e.g. state migration)
	// NOTE: memory metering is unnecessary for Clone methods
	Clone(interpreter *Interpreter) Value
```

fxamacker (commented):
Looks good!  

I mainly focused on non-test code and I left one minor comment about `ArrayValue.Clone()`.

runtime/interpreter/value.go
SupunS: ah yes, we don't need it there. I only updated the existing places. 

Maybe we can remove them in a separate PR. Should be harmless to have it for now. Thanks for catching that!

fxamacker (approved):

PR #1464, opened 2022-03-01
Check use of invalidated resources at runtime

SupunS:
Work towards https://github.com/dapperlabs/cadence-private-issues/issues/19

## Description

This adds second layer of runtime validation to check use of invalidated (moved/destroyed) resources. This check relies on:
- An external mapping, which keeps the resource-to-variable mapping.
- Assumes there could be at-most one mapping for each resource value at any given time.

______

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #1625, opened 2022-05-05
Fix erroneous report of resource invalidation when executing a pre-condition

dsainati1:
Fixes issue reported in Discord

## Description

The safety analysis for resource validation, which was introduced in #1464, was not properly handling pre and post conditions for functions in composites that implemented interfaces. In order to implement this behavior, we "wrap" the composite function in another "function" that contains the code for the pre and post conditions, and redeclare all the parameters. If one of the parameters is given a resource-kinded argument, this tricks the analysis into thinking this resource is being duplicated, when it is not. 

To handle this, we invalidate any resources that are parameters to the precondition before executing the body, then re-validate them after the body, before executing the post condition. 

______


runtime/interpreter/interpreter.go
```diff
@@ -2485,10 +2485,40 @@ func (interpreter *Interpreter) functionConditionsWrapper(
...
 						//   from the inner function, otherwise it is lost
 
 						returnValue := inner.invoke(invocation)
+
+						/* Restore resources invalidated before execution of the body */
+						for val, variable := range vars { //nolint:maprangecheck
+							interpreter.invalidateResource(val)
+							interpreter.resourceVariables[val] = variable
+						}
+
```
turbolent, line 2521: Great solution to the problem: Temporarily un-tracking the resources before the inner function starts tracking them, then restoring the tracking after the inner function finished makes sense 👍 

A few things:

- We usually use // comments everywhere 
- Maybe mention the code which starts tracking of the resource-kinded arguments, `bindParameterArguments` above
- Only allocate if necessary. Many functions have no resource-kinded arguments
- In the loop body, continue early (just like in functions, where we return early)
- Iteration using `range` over Go maps is non-deterministic, like CI warned about
  - Ignoring it is safe at the moment, but the body of loop may perform side-effects in the future (e.g. report errors), so this could become unsafe in the future
  - There is actually no need to use a map, as the map is only iterated over, it is not used for lookups
  - You can just e.g. use a slice of tuples, which is already ordered

```suggestion
						// Pre- and post-condition wrappers "re-declare" the same
						// parameters as are used in the actual body of the function,
						// see the use of bindParameterArguments at the start of this function wrapper.
						//
						// When these parameters are given resource-kinded arguments,
						// this can trick the resource analysis into believing that these
						// resources exist in multiple variables at once
						// (one for each condition wrapper + the function itself).
						//
						// This is not the case, however, as execution of the pre- and post-conditions
						// occurs strictly before and after execution of the body respectively.
						//
						// To prevent the analysis from reporting a false positive here,
						// when we enter the body of the wrapped function,
						// we invalidate any resources that were assigned to parameters by the precondition block,
						// and then restore them after execution of the wrapped function,
						// for use by the post-condition block.

						type argumentVariable struct {
							variable *Variable
							value    ResourceKindedValue
						}

						var argumentVariables []argumentVariable
						for _, argument := range invocation.Arguments {
							resourceKindedValue := interpreter.resourceForValidation(argument)
							if resourceKindedValue == nil {
								continue
							}

							argumentVariables = append(
								argumentVariables,
								argumentVariable{
									variable: interpreter.resourceVariables[resourceKindedValue],
									value:    resourceKindedValue,
								},
							)

							interpreter.invalidateResource(resourceKindedValue)
						}

						// NOTE: It is important to actually return the value returned
						//   from the inner function, otherwise it is lost

						returnValue := inner.invoke(invocation)

						// Restore the resources which were temporarily invalidated
						// before execution of the inner function

						for _, argumentVariable := range argumentVariables {
							value := argumentVariable.value
							interpreter.invalidateResource(value)
							interpreter.resourceVariables[value] = argumentVariable.variable
						}
```

runtime/missingmember_test.go
```diff
@@ -4163,3 +4163,837 @@ transaction(recipient: Address, amount: UFix64) {
 	)
 	require.NoError(t, err)
 }
+
+func TestRuntimeMissingMemberExampleMarketplace(t *testing.T) {
+
```
turbolent, line 4168: Parallelize all new tests:

```suggestion

	t.Parallel()

```

runtime/missingmember_test.go
```diff
@@ -4163,3 +4163,837 @@ transaction(recipient: Address, amount: UFix64) {
...
+	exampleTokenAddress, err := common.HexToAddress("0x1")
+	require.NoError(t, err)
+
+	exampleNFTAddress, err := common.HexToAddress("0x2")
+	require.NoError(t, err)
+
+	exampleMarketplaceAddress, err := common.HexToAddress("0x3")
+	require.NoError(t, err)
+
+	const exampleTokenContract = `
```
turbolent, line 4180: We already have some example and "real" contracts in the codebase, maybe see if it is possible to use them instead. See `examples.go`, `ft_test.go`, and `nft_test.go`

runtime/runtime_test.go
```diff
@@ -7236,3 +7237,496 @@ func TestRuntimeComputationMetring(t *testing.T) {
 		})
 	}
 }
+
+func TestRuntimeResourcePrecondition(t *testing.T) {
```
turbolent, line 7241: Could these be just interpreter tests, or do they also rely on behaviour in the runtime package?
If not, please move them to the interpreter tests, which reduces the amount of boilerplate.

Could you please also test two additional scenarios:

- Test with non-resource kinded arguments (e.g. by adding an additional non-resource-kinded parameter, so we have coverage for the `continue` branch in the for-loop which temporarily untracks resources in the condition function wrapper, above)
- Test with multiple interfaces, so that multiple wrappings with temporary untracking and restoring of resource-kinded arguments is tested

turbolent (commented):
Great work here @dsainati1, both the analysis of the root problem and the solution are 👌 

Like you described, in addition to the inner function body, also the condition wrapper functions bind the arguments, which leads to multiple variables being associated with the same resource.

Temporarily un-tracking the resource-kinded arguments and restoring the tracking makes sense.

runtime/missingmember_test.go
dsainati1: Unfortunately it looks like the marketplace playground example that this test is based on has more complexity than the tokens in our examples, but less complexity than a real contract, so I don't think any of the pre-existing token contracts in our tests are going to work.

runtime/runtime_test.go
dsainati1: I think https://github.com/onflow/cadence/pull/1625/files#diff-74ffa81b6938856f695ba112a721503dfc5ff2e852bc65d79d15ee4868f4bb97R7598 should test multiple wrappings of functions, but let me know if you had something else in mind.

turbolent: Ah,  `TestInterpreterResourceDoubleWrappedCondition` looks good 👍

turbolent (approved):
Great work! 

@SupunS can you please also have a look, as you authored the run-time resource tracking facility 🙏

SupunS (approved):
Nice! 👌

PR #1626, opened 2022-05-06
Meter location ID generation

SupunS:

## Description

Introduced a new method `MeteredID()` to use only in places where metering is needed. This was to avoid unnecessary changes to places where metering is not needed.

______


runtime/common/location.go
```diff
@@ -88,6 +90,21 @@ func NewLocationID(parts ...string) LocationID {
...
+func NewMeteredLocationID(memoryGauge MemoryGauge, parts ...string) LocationID {
+	jointString := joinStrings(memoryGauge, parts)
+	return LocationID(jointString)
+}
+
+func joinStrings(memoryGauge MemoryGauge, parts []string) string {
+	l := 0
+	for _, part := range parts {
+		l += len(part) + 1
+	}
```
robert-e-davidson3, line 102: This will overcount by one. Maybe this approach?

```suggestion
	if len(parts) == 0 {
		return ""
	}
	l := len(parts) - 1 // count the join character "."
	for _, part := range parts {
		l += len(part)
	}
```

SupunS: Yeah, I opted for slight overestimation (1 is very trivial), over complexity

turbolent (approved):

PR #1628, opened 2022-05-06
Revert removal of checker caching / #1572

turbolent:
Until we have #1570, let us at least remove the spurious errors which are reported at the moment, but reverting the removal of the checker caching: It is important to return the same checker for a location, as the checker performs type equality checks by Go pointer – returning different instances for the same logical types results in confusing errors. 

______


sideninja (approved):
Thank you, I was working on this now, but it makes sense to just revert first.

PR #1630, opened 2022-05-06
Track inner values of resource-kinded optionals

### LINKED ISSUE #1539: Accessing resource reference causes "missing value for member" error if resource has been moved
### Problem

I recently upgraded my project from cadence v0.18.1 to v0.23.0 (flow-cli v0.28.3 -> v0.31.2) and started getting `missing value for member` errors in contracts that didn't have any problems before. I traced the issue to the use of references to resources where the reference was created before the resource was transferred to a different account — after transfer, the reference is non-nil, but any attempt to access properties of the reference result in the error "missing value for member <property-name>". Previously, the references remained valid after transfer and properties could be accessed without error.

### Steps to Reproduce

I have reproduced the issue in the Flow Playground by slightly modifying the "Cadence Marketplace Tutorial" playground:

#### Changes:
* Add `getReference` method to Collection resource in ExampleNFT contract:
    ```
        pub fun getReference(id: UInt64): &NFT {
            return &self.ownedNFTs[id] as &NFT
        }
    ```

* Use the new `getReference` method in ExampleMarketplace purchase method to obtain a reference to the NFT that is being purchased; access the `id` property of the reference after the NFT has been transfered to the buyer:
    ```
      // pub fun purchase(...) {
            // ...

            let nftRef = self.ownerCollection.borrow()!.getReference(id: tokenID)
            log("NFT Reference before transfer:")
            log(nftRef)

            // deposit the NFT into the buyers collection
            receiverReference.deposit(token: <- self.ownerCollection.borrow()!.withdraw(withdrawID: tokenID))

            log("NFT Reference after transfer:")
            log(nftRef)
            log(nftRef.id)
    ```

#### Steps to reproduce:
https://play.onflow.org/ebde5a12-6292-40d5-86a1-42c9eca8b611
1. Deploy contracts for 0x01, 0x02, 0x03 (ExampleToken, ExampleNFT, ExampleMarketplace)
2. Run transactions 1 through 5. Transaction 5 ("Purchase NFT") will fail with the `missing value for member` error

dsainati1:
Closes #1539

## Description

With the change in https://github.com/onflow/cadence/pull/1303, it is no longer possible to obtain a reference to an optional value, as all such references are automatically transformed into optional references. However, these references were not properly tracked by the logic added in https://github.com/onflow/cadence/pull/1344, meaning that the linked issue was still possible despite the cause seemingly being removed. This adds proper tracking for the inner values of resource-kinded optionals. 

______

### DIFF
--- a/runtime/interpreter/interpreter_expression.go
+++ b/runtime/interpreter/interpreter_expression.go
@@ -901,10 +901,15 @@ func (interpreter *Interpreter) VisitReferenceExpression(referenceExpression *as
 
 			getLocationRange := locationRangeGetter(interpreter.Location, referenceExpression.Expression)
 
+			innerValue := result.InnerValue(interpreter, getLocationRange)
+			if result, ok := innerValue.(ReferenceTrackedResourceKindedValue); ok {
+				interpreter.trackReferencedResourceKindedValue(result.StorageID(), result)
+			}
+
 			return NewSomeValueNonCopying(
 				&EphemeralReferenceValue{
 					Authorized:   innerBorrowType.Authorized,
-					Value:        result.InnerValue(interpreter, getLocationRange),
+					Value:        innerValue,
 					BorrowedType: innerBorrowType.Type,
 				},
 			)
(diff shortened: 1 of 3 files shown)


turbolent (approved):
Great work! 

For the other resource-kinded objects (composites, dictionaries, and arrays) we could track via their dedicated storage ID (as they are mutable, they are stored in separate slabs).
However, optionals (`SomeValue`) are potentially inlined in storage, so do not have a dedicated storage ID.

I thought at first this is an issue, but as you showed, tracking the potentially resource-kinded inner value of the optional works fine, nice 👍

PR #1633, opened 2022-05-06
cadence.Value.Type method back to original param list

robert-e-davidson3:
Closes https://github.com/dapperlabs/cadence-private-issues/issues/46

## Description

The SDK etc use the cadence.Value and cadence.Type types. This PR undoes a change to the cadence.Value.Type method so the API will remain unchanged in the next Cadence release.

The cadence.Value.MeteredType method was added for when memory metering is needed.

Another change: I added compile-time interface type checking for the cadence.Value types. ex: `var _ Value = Void{}`.

______


SupunS (commented):
Looks good!

SupunS (approved):
Thanks for adding these!

PR #1635, opened 2022-05-07
adjust big int metering minor errors

tarakby:
## Description

This PR addresses late comments on the PR https://github.com/onflow/cadence/pull/1590. 
It adjusts some minor errors in metering big integers operations. 

- Some metering functions are still a bit "loose" as they don't treat edge cases. This results in over metering but does not result in errors. 
- the shift functions are also not adjusted: the `shift` parameter in right and left shift should be a 32 or 64 bit integer, instead of a big integer.

______

- [x] Targeted PR against feature branch

### DIFF
--- a/runtime/common/metering.go
+++ b/runtime/common/metering.go
@@ -425,11 +425,18 @@ func BigIntByteLength(v *big.Int) int {
 //
 
 func NewPlusBigIntMemoryUsage(a, b *big.Int) MemoryUsage {
-	// max(|a|, |b|) + 5
+	// if |a|==|b|==0, then 0
+	// else max(|a|, |b|) + 5
+	aWordLength := len(a.Bits())
+	bWordLength := len(b.Bits())
+
+	if aWordLength == 0 && bWordLength == 0 {
+		return NewBigIntMemoryUsage(0)
+	}
 
 	maxWordLength := max(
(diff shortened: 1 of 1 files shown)


runtime/common/metering.go
```diff
@@ -425,11 +425,18 @@ func BigIntByteLength(v *big.Int) int {
...
 func NewPlusBigIntMemoryUsage(a, b *big.Int) MemoryUsage {
-	// max(|a|, |b|) + 5
+	// if |a|==|b|==0, then 0
+	// else max(|a|, |b|) + 5
+	aWordLength := len(a.Bits())
+	bWordLength := len(b.Bits())
+
+	if aWordLength == 0 && bWordLength == 0 {
+		return NewBigIntMemoryUsage(0)
+	}
```
turbolent, line 435: I ignored the 0 cases, because I understood you as saying that the library always pre-allocates some words, but this does not hurt 👍

turbolent (approved):
Thank you!

runtime/common/metering.go
tarakby: You're right, the lib over allocates but only in the general cases. 
We can add these edge cases anytime later if you think we should "tighten" the metering.

Issue #1659, opened 2022-05-18
Clean up parse error handling


turbolent:
### Context

The parser is currently implemented in a way where anything that’s not specifically a syntax error becomes one.
This is a convenience mechanism to allow the parser to easily report syntax errors.

When a syntax problem is encountered there are currently four (maybe more?) “kinds” how that is reported: 
- `panic` with a string, e.g. `panic("expected comma, got start of parameter")`
- `panic` with a “plain” / non-`SyntaxError`, e.g. `panic(fmt.Errorf("expected parameter or end of parameter list, got %s", p.current.Type))`
- `parser.report` with a “plain” / non-`SyntaxError`, e.g. `report(fmt.Errorf("missing start of string literal: expected '\"'"))
- `parser.report` with a `SyntaxError`

`panic`s abort the parse immediately, and if the value that’s recovered (e.g. string or plain error) is not a `SyntaxError`, it is automatically wrapped into one at the current parse location: The panics are recovered at the top-level entry point of parsing, and turned into a `parser.report`: 
https://github.com/onflow/cadence/blob/66eac2a2dfac1eb474b7c11dbe73be50d3adbb18/runtime/parser2/parser.go#L82-L93
`parser.report` in turn performs the wrapping: if whatever error is reported is not yet a syntax error, it becomes one:
https://github.com/onflow/cadence/blob/66eac2a2dfac1eb474b7c11dbe73be50d3adbb18/runtime/parser2/parser.go#L134-L142

Even though convenient, this has two, related problems:
- It makes it impossible to report explicit non-syntax errors, i.e. fatal errors
- Non explicit non-syntax errors, e.g. Go runtime errors are turned into syntax errors

### Suggested Solution

- Remove the convenience syntax error reporting scheme and require syntax errors to be always reported as such. The verbosity can be reduced by introducing a convenience function to produce a syntax error
- Treat non-syntax errors as fatal errors

PR #1660, opened 2022-05-18
Wrap memory error in FatalError and check for it when parsing

robert-e-davidson3:
Closes https://github.com/dapperlabs/cadence-private-issues/issues/52

## Description

Two related changes:
1. The errors returned from the FVM can have types that Cadence is unaware of. This is causing a problem when there's an error when metering memory usage (typically when out of memory). So here I wrap those errors inside a new error type, `FatalError`.
2. Parsing's error handling is a mishmash of strategies. Compounding this is that panicking is used to implement lookahead in the parser. Eventually we will explicitly wrap every parsing error as a parsing error and treat all other errors as fatal. But for this PR we do the inverse: wrap known-to-be-fatal errors in `FatalError` and check for them when handling errors while parsing. This relates to https://github.com/onflow/cadence/issues/1659.

______

### DIFF
--- a/runtime/parser2/expression_test.go
+++ b/runtime/parser2/expression_test.go
@@ -19,7 +19,9 @@
 package parser2
 
 import (
+	"errors"
 	"fmt"
+	"github.com/onflow/cadence/runtime/common"
 	"math"
 	"math/big"
 	"math/rand"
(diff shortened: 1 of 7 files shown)


turbolent (commented):
Looks good! Could you please add a test case?

robert-e-davidson3:
> Looks good! Could you please add a test case?

Good call! I missed a path.

runtime/parser2/expression_test.go
```diff
@@ -313,6 +315,25 @@ func TestParseAdvancedExpression(t *testing.T) {
...
+		var panicMsg interface{}
+		(func() {
+			defer func() {
+				panicMsg = recover()
+			}()
+
+			ParseExpression("1 < 2 > 3", gauge)
+		})()
+
+		require.IsType(t, common.FatalError{}, panicMsg)
```
turbolent, line 334: Maybe also assert the inner error is a memory limit error (and not potentially some other error in the future)

runtime/parser2/parser_test.go
```diff
@@ -493,6 +494,22 @@ func TestParseArgumentList(t *testing.T) {
...
+		var panicMsg interface{}
+		(func() {
+			defer func() {
+				panicMsg = recover()
+			}()
+
+			ParseArgumentList(`(1, b: true)`, gauge)
+		})()
+
+		require.IsType(t, common.FatalError{}, panicMsg)
```
turbolent, line 511: Same here

turbolent (commented):
Thank you for adding tests!

turbolent (approved):
Nice!

PR #1663, opened 2022-05-19
Fix local token replay limit, add global replay limit

turbolent:
## Description

The local replay was not correctly implemented: 
An ambiguity does not just end when the buffered tokens are accepted, it also ends after replaying the buffered tokens and re-parsing using a different rule. 
Explicitly state when an ambiguity starts and ends, and reset the local token replay limit.

In addition, add a global replay limit, as the local replay limit might not be sufficient in rejecting programs that are too ambiguous.

I parsed all Mainnet contracts using these changes and they all still parse.

TODO: Wrap errors in fatal errors, depends on #1660 

______

### DIFF
--- a/runtime/parser2/parser.go
+++ b/runtime/parser2/parser.go
@@ -278,11 +278,11 @@ func (p *parser) acceptBuffered() {
 
 // localTokenReplayCountLimit is a sensible limit for how many tokens may be replayed
 // until the top-most ambiguity ends.
-const localTokenReplayCountLimit = 2 << 6
+const localTokenReplayCountLimit = 1 << 7
 
 // globalTokenReplayCountLimit is a sensible limit for how many tokens may be replayed
 // during a parse
-const globalTokenReplayCountLimit = 2 << 8
+const globalTokenReplayCountLimit = 1 << 9
 
 func checkReplayCount(total, additional, limit uint, kind string) uint {
 	newTotal := total + additional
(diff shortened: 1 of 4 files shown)


runtime/parser2/parser.go
```diff
@@ -252,14 +257,24 @@ func (p *parser) acceptBuffered() {
...
-	// Reset the replayed tokens count
-	p.replayedTokensCount = 0
 }
 
-// tokenReplayLimit is a sensible limit for how many tokens may be replayed
-// until the replay buffer is accepted.
-const tokenReplayLimit = 2 << 12
+// localTokenReplayCountLimit is a sensible limit for how many tokens may be replayed
+// until the top-most ambiguity ends.
+const localTokenReplayCountLimit = 2 << 6
```
turbolent, line 264:
```suggestion
const localTokenReplayCountLimit = 1 << 7
```

runtime/parser2/parser.go
```diff
@@ -252,14 +257,24 @@ func (p *parser) acceptBuffered() {
...
+
+// globalTokenReplayCountLimit is a sensible limit for how many tokens may be replayed
+// during a parse
+const globalTokenReplayCountLimit = 2 << 8
```
turbolent, line 268:
```suggestion
const globalTokenReplayCountLimit = 1 << 9
```

robert-e-davidson3 (approved):

dsainati1 (approved):

SupunS (approved):

PR #1668, opened 2022-05-20
Add token limit

turbolent:
## Description

Limit the amount of tokens that may be emitted during a lex.

______

### DIFF
--- a/runtime/parser2/lexer/lexer.go
+++ b/runtime/parser2/lexer/lexer.go
@@ -28,7 +28,7 @@ import (
 )
 
 // tokenLimit is a sensible limit for how many tokens may be emitted
-const tokenLimit = 2 << 18
+const tokenLimit = 1 << 19
 
 type TokenLimitReachedError struct {
 	ast.Position
--- a/runtime/parser2/lexer/lexer.go
+++ b/runtime/parser2/lexer/lexer.go
@@ -27,6 +27,19 @@ import (
 	"github.com/onflow/cadence/runtime/common"
 )
 
+// tokenLimit is a sensible limit for how many tokens may be emitted
+const tokenLimit = 2 << 18
+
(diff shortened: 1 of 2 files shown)


runtime/parser2/lexer/lexer.go
```diff
@@ -27,6 +27,19 @@ import (
 	"github.com/onflow/cadence/runtime/common"
 )
 
+// tokenLimit is a sensible limit for how many tokens may be emitted
+const tokenLimit = 2 << 18
```
robert-e-davidson3, line 31: Why `2 << 18` instead of `1 << 19`?

runtime/parser2/lexer/lexer.go
```diff
@@ -229,6 +242,11 @@ func (l *lexer) acceptOne(r rune) bool {
 
 // emit writes a token to the channel.
 func (l *lexer) emit(ty TokenType, val interface{}, rangeStart ast.Position, consume bool) {
+
+	if len(l.tokens) >= tokenLimit {
+		panic(TokenLimitReachedError{})
```
robert-e-davidson3, line 247: Should this be wrapped in `FatalError`? I'm not sure of the flow of control so I really don't know.

robert-e-davidson3 (approved):

runtime/parser2/lexer/lexer.go
turbolent: Tokenization happens before parsing, so there is no real need to wrap it. We might want to clean up / unify the error handling in general

runtime/parser2/lexer/lexer.go
turbolent, line 31:
```suggestion
const tokenLimit = 1 << 19
```

SupunS (approved):

Issue #1683, opened 2022-06-01
Investigate diverging testnet execution results


j1010001:
While replaying blocks on testnet we noticed a transaction that used to pass and now fails, this is preventing us from replaying blocks on testnet ATM. We need to investigate and decide which result is correct and fix the underlying issue that caused the results to diverge.

More details in this thread: https://axiomzen.slack.com/archives/CG0B7CJAJ/p1653500963838309

PR #1689, opened 2022-06-02
Avoid type-checking of the old program during contract updates

SupunS:
Closes https://github.com/onflow/cadence/issues/1683

## Description

Rechecking the existing (old) code of the contract during contract updates can lead to type checking errors, if the dependencies of the contract have changed e.g: API change.

Therefore, avoid checking the old code, and instead only parse.

______

### DIFF
--- a/runtime/contract_update_test.go
+++ b/runtime/contract_update_test.go
@@ -37,6 +37,12 @@ func TestContractUpdateWithDependencies(t *testing.T) {
 	runtime := newTestInterpreterRuntime()
 	accountCodes := map[common.LocationID][]byte{}
 	signerAccount := common.MustBytesToAddress([]byte{0x1})
+	fooLocation := common.AddressLocation{
+		Address: signerAccount,
+		Name:    "Foo",
+	}
+
+	var checkGetSetProgram, getProgramCalled, setProgramCalled bool
 
 	runtimeInterface := &testRuntimeInterface{
 		getCode: func(location Location) (bytes []byte, err error) {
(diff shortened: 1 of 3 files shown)


runtime/contract_update_test.go
```diff
@@ -0,0 +1,200 @@
...
+		},
+		emitEvent: func(event cadence.Event) error {
+			return nil
+		},
+		decodeArgument: func(b []byte, t cadence.Type) (value cadence.Value, err error) {
+			return json.Decode(b)
+		},
+
+		// Always force to get the old program from the source during the update.
+		getProgram: func(location Location) (*interpreter.Program, error) {
```
turbolent, line 73: 👍 

Maybe we can also add a `setProgram` function to the interface, and then assert here and in `setProgram`, that during the update (after the initial deployment), the cache isn't read from or written to

turbolent (approved):
Nice!

runtime/contract_update_test.go
SupunS: `setProgram` and `getProgram` do get called when parsing the 'current' program. So it may not be possible to isolate the cache read/writes during parsing the 'old-program' vs 'current-program' for assertion.
We could test 'what' is read from the cache, but I guess that's already tested by `TestRuntimeContractUpdateProgramCaching`?

turbolent: Right, they get called for the update. In the old version of the code, there were multiple calls to get/set program.
I tried to add a best-effort assertion for this in [02293b0](https://github.com/onflow/cadence/pull/1689/commits/02293b0fae79385fb0820318bd44581c8d8d93c0): assert get/set is only called once.

SupunS: looks good! Thanks for adding that

PR #1694, opened 2022-06-03
Properly support super types in ConformsToStaticType implementations

### LINKED ISSUE #1693: `Value.ConformsToStaticType` implementations do not always properly handle supertypes
### Problem

Some implementations of `Value.ConformsToStaticType` do not always properly handle supertypes, e.g. `PathValue.ConformsToStaticType` with `PrimitiveStaticTypeAnyStruct` should return true, but does not.

### Steps to Reproduce

For example, sending a transaction where the parameter type is `[AnyStruct]`, and an element is a path, incorrectly fails, but should succeed.

turbolent:
Closes #1693

## Description

Change `ConformsToStaticType` so it only determines if the value conforms to its own static type, recursively.

______

### DIFF
--- a/runtime/convertValues.go
+++ b/runtime/convertValues.go
@@ -1036,7 +1036,6 @@ func importCapability(
 ) {
 
 	_, ok := borrowType.(cadence.ReferenceType)
-
 	if !ok {
 		return nil, fmt.Errorf(
 			"cannot import capability: expected reference, got '%s'",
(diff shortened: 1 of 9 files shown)


turbolent:
@SupunS great summary, correct! I've added a comment in 9d3d900cd19e8d4b79cb264f13f3527db611f3f0

Issue #1690, opened 2022-06-02
Documentation for breaking changes of Secure Cadence


turbolent:
### Issue To Be Solved

Given that the release of Secure Cadence is imminent, we should update the documentation to reflect the future behaviour.

### Suggested Solution

In particular, we should update the documentation to reflect the following breaking changes:

- [x] #1214 @turbolent 
- [x] #1027 @SupunS
- [x] #1303 @dsainati1 
- [x] #1360 @SupunS 

We can mostly just copy-and-paste the new behaviour from the [breaking changes announcement in the forum](https://forum.onflow.org/t/breaking-changes-coming-with-secure-cadence-release/3052).

We should maybe wait with merging related PRs until the sporks are closer.

turbolent:
@SupunS Ah, great! I only checked the PRs themselves, sorry for the ping

PR #1700, opened 2022-06-06
update documentation for optional references

dsainati1:
For https://github.com/onflow/cadence/issues/1690

Update the documentation on optional references to reflect the changes in https://github.com/onflow/cadence/pull/1303.

______

### DIFF
--- a/docs/language/references.md
+++ b/docs/language/references.md
@@ -33,13 +33,13 @@ If the referenced value is nil, the reference itself will be nil. If the referen
 exists, then forcing the optional reference will yield a reference to that value:
 
 ```cadence
-let nil: String? = nil
-let nilRef = &n as &String? // r has type &String?
-let n = r! // error, forced nil value
+let nilValue: String? = nil
+let nilRef = &nilValue as &String? // r has type &String?
+let n = nilRef! // error, forced nil value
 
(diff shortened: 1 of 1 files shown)


docs/language/references.md
```diff
@@ -28,6 +28,20 @@ helloRef.length // is `5`
...
 ```
 
+If you attempt to reference an optional value, you will receive an optional reference.
+If the referenced value is nil, the reference itself will be nil. If the referenced value
+exists, then forcing the optional reference will yield a reference to that value:
+
+```cadence
+let nil: String? = nil
+let nilRef = &n as &String? // r has type &String?
+let n = r! // error, forced nil value
```
joshuahannan, line 38:
```suggestion
let nilRef = &nil as &String? // r has type &String?
let n = nilRef! // error, forced nil value
```

Same suggestion for the other example

joshuahannan (approved):

docs/language/references.md
```diff
@@ -28,6 +28,20 @@ helloRef.length // is `5`
...
+
+let str: String? = ""
+let strRef = &n as &String? // r has type &String?
```
robert-e-davidson3, line 41: Where does `r` come from?

PR #1706, opened 2022-06-07
Differentate invalidated resource use errors

turbolent:
Closes dapperlabs/cadence-private-issues#55

## Description

Differentiate between use after destruction (user error) and use after invalidation (internal error)

______

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -1486,13 +1486,15 @@ func (v *ArrayValue) IsImportable(inter *Interpreter) bool {
 	return importable
 }
 
-func (v *ArrayValue) checkInvalidatedResourceUse(interpreter *Interpreter, getLocationRange func() LocationRange) {
+func (v *ArrayValue) checkDestroyedResourceUse(getLocationRange func() LocationRange) {
 	if v.isDestroyed {
 		panic(DestroyedResourceError{
 			LocationRange: getLocationRange(),
 		})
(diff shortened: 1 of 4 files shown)


runtime/interpreter/value.go
```diff
@@ -1487,7 +1487,13 @@ func (v *ArrayValue) IsImportable(inter *Interpreter) bool {
 }
 
 func (v *ArrayValue) checkInvalidatedResourceUse(interpreter *Interpreter, getLocationRange func() LocationRange) {
-	if v.isDestroyed || (v.array == nil && v.IsResourceKinded(interpreter)) {
+	if v.isDestroyed {
+		panic(DestroyedResourceError{
+			LocationRange: getLocationRange(),
+		})
+	}
```
SupunS, line 1494: Wouldn't this hide the internal error?
 Maybe it's ok to treat that as an internal error since it is not supposed to reach here

turbolent: Treating use of a destroyed resource as an internal error will likely result in many reports and make the alert to noisy.

SupunS: It would not come to the runtime check under normal circumstances, wouldn't it? Use via references is already handled separately?

turbolent: Clarified offline

SupunS (commented):
LGTM!

PR #1718, opened 2022-06-13
Separate user errors from internal errors

### LINKED ISSUE #1708: Separate user errors from internal errors
### Issue To Be Solved

All the run-time panics are gracefully handled at the `interpterRuntime` and returned to the caller (FVM): 
https://github.com/onflow/cadence/blob/56e99979e110584457c0ea0c0b03d9582a6d16c8/runtime/runtime.go#L237-L238

Some of these runtime panics are user errors. e.g: `ContainerMutationError`, while some of the errors are implementation errors. e.g: `UnreachableError`.

There needs to be a way to distinguish between user errors and internal errors. This will:
- Allow the caller (FVM) to act differently, depending on the kind of error
- Enable setting up alerts for internal errors.( e.g: https://github.com/onflow/cadence/issues/1686)

### Suggested Solution

- [x] Introduce two interfaces:
  - InternalError
  - UserError
- [x] Make sure each concrete error type implements the correct interface
- [ ] [TBD] Maybe also refactor the `ExternalError` to follow a similar pattern? Currently ExternalError is a concrete type.
- [x] Refactor generic use of go-errors to throw more specific errors. e.g: `fmt.Errorf()` and `errors.New()` must be replaced to use a known implementation of either `UserError` or `InternalError` to avoid them being misclassified. 
  - e.g: `fmt.Errorf()` has been used for both user errors as well as internal errors
  - Need to figure out which category of errors each of them belongs to.
  - Maybe don't need to refactor parser/lexer errors, as they are caught and wrapped at the parser.
  - Can skip modules that are not related to runtime. e.g: wasm compiler, toolings (LS, etc.), etc.

SupunS:
Closes #1708

## Description

This PR does the following refactor:
- Introduced two interfaces:
  - InternalError
    - Errors happen due to implementation bugs. e.g: Unreachable paths,`InvalidatedResourceError` 
    - Internal errors are always encouraged to be thrown (as opposed to gracefully handling/ returning from methods)
    - They should not be recovered (except at an entry point. i.e: interpreter runtime)
  - UserError
    - Errors happen due to problems in user code. e.g: Syntax errors, Semantic errors, etc.
    - OK to recover
- Conform each concrete error type to the correct interface
- Refactor generic use of go-errors to throw more specific errors. e.g: `fmt.Errorf()` and `errors.New()` are replaced to use a `DefaultUserError` (implements `UserError`) or `UnexpectedError` (implements `InternalError`) to avoid them being misclassified.

This PR does **not** refactor: 
- Parser/Lexer errors, as they are caught and wrapped at the parser with a user error. Can refactor this in a follow-up PR, to separate go-panics (internal errors) from user errors. 
- Modules that are not related to runtime. e.g: wasm compiler, toolings (LS, etc.), etc.

### FVM Changes:
PR: https://github.com/onflow/flow-go/pull/2608

______


runtime/errors/errors.go
```diff
@@ -66,3 +106,107 @@ type ParentError interface {
...
+func (e MemoryError) Error() string {
+	return fmt.Sprintf("memory error: %s", e.Err.Error())
+}
+
+func (e MemoryError) IsUserError() {}
+
+// UnexpectedError is the default implementation of InternalError interface.
+// It's a generic error that wraps an implementation error.
+//
+type UnexpectedError struct {
```
dsainati1, line 129: What is the difference between an unexpected error and an unreachable error? Is it possible for an unexpected error to occur with, e.g. a malformed FVM, while unreachable errors can never occur at all?

SupunS: `UnexpectedError` is a generic error to be thrown in any unhandled cases. It's essentially a wrapper for go-errors, to make sure all errors that are thrown are known 'cadence errors'.
i.e: replaces all existing `fmt.Errorf()` / `errors.new()`.

`UnreachableError` doesn't wrap anything. it only prints the stack trace.

dsainati1 (approved):

runtime/errors/errors.go
SupunS: oh sorry, I didn't see the last part of the question.
Both of these errors should never occur. Errors caused by malformed FVM are wrapped as `ExternalError` (this was already there).

runtime/errors/errors.go
```diff
@@ -21,8 +21,44 @@ package errors
...
+
+	"golang.org/x/xerrors"
 )
 
+// InternalError is thrown on an implementation error.
+// A program should never throw an InternalError in an ideal world.
+// e.g: UnreachableError
+//
+// InternalError s must always be thrown, and let it propagate up the call stack.
+//
```
turbolent, line 33:
```suggestion
// InternalError is an implementation error, e.g an unreachable code path (UnreachableError).
// A program should never throw an InternalError in an ideal world.
// 
// InternalError s must always be thrown and not be caught (recovered), i.e. be propagated up the call stack.
//
```

runtime/errors.go
```diff
@@ -74,6 +74,8 @@ func (e CallStackLimitExceededError) Error() string {
 	)
 }
 
+func (CallStackLimitExceededError) IsUserError() {}
```
turbolent, line 77: Maybe also add "assertions" the type implements the interface, here and for the other errors below:

```suggestion
var _ UserError = CallStackLimitExceededError{}

func (CallStackLimitExceededError) IsUserError() {}
```

runtime/common/metering.go
```diff
@@ -811,17 +811,3 @@ func NewAtreeEncodedSlabMemoryUsage(slabsCount uint) MemoryUsage {
 		Amount: uint64(slabsCount),
 	}
 }
-
-// FatalError indicates an error that should end
-// the Cadence parsing, checking, or interpretation.
-type FatalError struct {
```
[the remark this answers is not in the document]
turbolent: 👌

runtime/convertValues.go
```diff
@@ -194,10 +193,9 @@ func exportValueWithInterpreter(
 			return nil, nil
 		}
 		return exportValueWithInterpreter(*referencedValue, inter, seenReferences)
+	default:
+		return nil, errors.NewUnexpectedError("cannot export value of type %T", value)
```
turbolent, line 197: 👌

runtime/convertValues.go
```diff
@@ -771,9 +769,13 @@ func importValue(
 			v.Address,
 			v.BorrowType,
 		)
+	case cadence.Link:
+		return nil, errors.NewDefaultUserError("cannot import value of type %T", value)
+	default:
+		// This means the implementation has unhandled types.
+		// Hence, return an internal error
+		return nil, errors.NewUnexpectedError("cannot import value of type %T", value)
```
turbolent, line 777: 👌

runtime/errors/errors.go
```diff
@@ -21,8 +21,44 @@ package errors
...
+type InternalError interface {
+	error
+	IsInternalError()
+}
+
+// UserError is an error thrown for an error in the user-code.
```
turbolent, line 39:
```suggestion
// UserError is an error thrown for an error in the user-code, e.g. exceeding a metering limit.
```

runtime/errors/errors.go
turbolent: I was also wondering what the difference between the two types is. Maybe we can merge them, or alternatively, I guess an `UnreachableError` could be wrapped by/nested in an `UnexptectedError`? ("it was unexpected this code path was reached" – compared with e.g. another `UnexpectedError` which wraps a concrete "unreachable error", like in `importValue`)

runtime/interpreter/errors.go
```diff
@@ -43,6 +46,8 @@ func (e *unsupportedOperation) Error() string {
 	)
 }
 
+func (e *unsupportedOperation) IsUserError() {}
```
turbolent, line 49: This is actually an internal/implementation error, it is e.g. thrown when a program that type-checked could not be interpreted, e.g. currently missing support for certain binary and unary operators.

runtime/errors/errors.go
```diff
@@ -35,10 +71,14 @@ type UnreachableError struct {
 	Stack []byte
 }
 
+var _ InternalError = UnreachableError{}
+
 func (e UnreachableError) Error() string {
 	return fmt.Sprintf("unreachable\n%s", e.Stack)
 }
 
+func (e UnreachableError) IsInternalError() {}
```
turbolent, line 80:
```suggestion
func (UnreachableError) IsInternalError() {}
```

runtime/errors/errors.go
```diff
@@ -66,3 +106,107 @@ type ParentError interface {
...
+
+func (e MemoryError) Unwrap() error {
+	return e.Err
+}
+
+func (e MemoryError) Error() string {
+	return fmt.Sprintf("memory error: %s", e.Err.Error())
+}
+
+func (e MemoryError) IsUserError() {}
```
turbolent, line 124:
```suggestion
func (MemoryError) IsUserError() {}
```

runtime/errors/errors.go
```diff
@@ -66,3 +106,107 @@ type ParentError interface {
...
+
+func (e UnexpectedError) Unwrap() error {
+	return e.Err
+}
+
+func (e UnexpectedError) Error() string {
+	return e.Err.Error()
+}
+
+func (e UnexpectedError) IsInternalError() {}
```
turbolent, line 149:
```suggestion
func (UnexpectedError) IsInternalError() {}
```

runtime/interpreter/errors.go
```diff
@@ -80,16 +85,7 @@ func (e PositionedError) Error() string {
...
-// It contains the recovered value.
-//
-type ExternalError struct {
-	Recovered any
-}
-
-func (e ExternalError) Error() string {
-	return fmt.Sprint(e.Recovered)
-}
+func (e PositionedError) IsUserError() {}
```
turbolent, line 88: Here and for the other errors below:

```suggestion
func (PositionedError) IsUserError() {}
```

runtime/interpreter/interpreter.go
```diff
@@ -1125,14 +1124,20 @@ func (interpreter *Interpreter) InvokeTransaction(index int, arguments ...Value)
 func (interpreter *Interpreter) RecoverErrors(onError func(error)) {
 	if r := recover(); r != nil {
 		var err error
+
+		// Recover all errors, because interpreter can be directly invoked by FVM.
```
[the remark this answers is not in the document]
turbolent: I guess it's better to recover here then in the runtime, as it is now. Maybe we can remove the recovery logic there now?

runtime/parser2/errors.go
```diff
@@ -211,6 +226,8 @@ var _ ParseError = &MissingCommaInParameterListError{}
 
 func (*MissingCommaInParameterListError) isParseError() {}
 
+func (e *MissingCommaInParameterListError) IsUserError() {}
```
turbolent, line 229: Also, maybe add a type assertion like for the other parse errors above.

```suggestion
func (*MissingCommaInParameterListError) IsUserError() {}
```

runtime/parser2/expression.go
```diff
@@ -598,7 +598,7 @@ func defineLessThanOrTypeArgumentsExpression() {
 				defer func() {
 					err := recover()
 					// Fatal errors should abort parsing
```
turbolent, line 600: Maybe update the comment above.

Maybe this shouldn't check memory errors specifically, but e.g. non-`ParseError`s?

runtime/tests/errors_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+	"github.com/onflow/cadence/runtime"
+	"github.com/onflow/cadence/runtime/interpreter"
+	"github.com/onflow/cadence/runtime/parser2"
+)
+
+// TestErrorInterfaceConformance checks whether all the error structs implement
+// one of the interfaces.
+//
+func TestErrorInterfaceConformance(t *testing.T) {
```
turbolent, line 40: Nice! 👌

turbolent (commented):
Great work! 👏

runtime/errors.go
SupunS: There are 215+ user errors, and adding the assertions manually is a lot of work. So I took the lazy approach of asserting it via a test-case 😄

turbolent: Of course! OK if I quickly add them?

SupunS: for sure, thanks!

runtime/interpreter/errors.go
SupunS: Fixed it in https://github.com/onflow/cadence/pull/1718/commits/ff5e6de74cf4a640f71b6e91f882b25952966e46

runtime/parser2/expression.go
SupunS: Ideally Yes. But many parser errors are thrown with `fmt.Errorf()`, so they won't get caught as `ParseError`. We need to throw specific errors (i.e: replace `fmt.Errorf()` with a known `ParseError` similar to what we did in this PR for interpreter errors).

I was hoping to do that in a separate PR, to avoid this becoming too large.

runtime/errors.go
turbolent: "quickly" 😅 added in [15943b7](https://github.com/onflow/cadence/pull/1718/commits/15943b792bd25ef03c993ad1d21488e3d77127fa)

runtime/parser2/expression.go
turbolent: Ah, that makes sense! Yes, we can/should do that in a separate PR.
BTW, I had discussed the errors in the parser with @robert-e-davidson3 before, who had asked about it:

> when a syntax problem is encountered there are currently four (maybe more?) “kinds” how that is reported: 
> - `panic` with a string, e.g. `panic("expected comma, got start of parameter")`
> - `panic` with a “plain” error, e.g. `panic(fmt.Errorf("expected parameter or end of parameter list, got %s", p.current.Type))`
> - `parser.report` with a “plain” error, e.g. `report(fmt.Errorf("missing start of string literal: expected '\"'"))`
> - `parser.report` with a `SyntaxError`
> `panic`s abort the parse immediately, and if the value that’s recovered (e.g. string or plain error) is not a `SyntaxError`, it is automatically wrapped into one at the current parse location

Maybe we can put this into a separate issue

runtime/interpreter/interpreter.go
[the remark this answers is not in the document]
SupunS: Not sure whether all errors are covered by the `(Interpreter) RecoverErrors()`. i.e: Can there be errors thrown at runtime, outside of the interpreter 🤔 I'll check.

runtime/errors/errors.go
SupunS: The `UnexpectedError` was introduced to make go-errors conform to the `InternalError` interface. Given that `UnreachableError` already implements the `InternalError`, wrapping it with `UnexpectedError` doesn't add much value – Only adds complexity to the error chain.

So I merged the two (https://github.com/onflow/cadence/pull/1718/commits/098d47b5ebf124b25d120248c509fa5fa40624b6) as I suggested earlier. So the `UnexpectedError` has both a cause (error) and the stack trace – and the trace info is available for all scenarios.

Now then, this makes the `UnreachableError` redundant, so removed it, and simply used a built-in error string.

turbolent: Nice!

turbolent (approved):
Great work!

turbolent:
@SupunS Sorry for not following up on this PR earlier, I didn't notice you addresses the comments. Please feel free to re-request a review (🔄 button), then the reviewer gets a notification. Thanks!

PR #960, opened 2021-05-28
Improve the crypto doc

tarakby:
## Description

Add more details to the crypto contract and native crypto features :
 - move the public key section to `crypto.md` as `PublicKey` is a cryptographic structure rather than an account structure. 
 - add details about signature verification.
 - add details about hashing with and without tags.

PR #1719, opened 2022-06-13
Cadence Style Guide

alilloig:
Closes [#960](https://github.com/onflow/flow/issues/960)
## Description
Developers need better guidelines for how to write clean cadence code. Define those guidelines at `docs/cadence-style-guide.md`. Early PR containing only table of contents to seek consensus about which points should the guide contain and check everyone's thoughts on conflictive aspects.

______

### DIFF
--- a/docs/cadence-style-guide.md
+++ b/docs/cadence-style-guide.md
@@ -1,21 +1,23 @@
-# Cadence Style Guide
-
+# Cadence <- Style Guide
+## Table of contents
+1. [Line length](#line-length)
+1. [Comments](#comments)
+1. [Identation](#identation)
+1. [Whitespace between code lines](#whitespace-between-code-lines)
+1. [Declaring variables and constants](#whitespace-between-code-lines)
+1. [Bracket spaces](#bracket-spaces)
+1. [Panic messages](#panic-messages)
+1. [Naming and capitalization](#naming-and-capitalization)
 ## Line length
-
+A line lenght of 80 characters is recomended for Cadence. This should be strictly enforced when commenting but could be observed in a flexible way, prioritazing code readability, when dealing with complex code sentences as the capability related ones.
 ## Comments
 ### High level documentation at the begining of files (contracts, transactions and scripts)
 ### Commenting functions
-
 ## Identation
-
 ## Whitespace between code lines
-
 ## Declaring variables and constants
-
 ## Bracket spaces
-
 ## Panic messages
-
 ## Naming and capitalization
 ### Contracts
 ### Fields
--- a/docs/cadence-style-guide.md
+++ b/docs/cadence-style-guide.md
@@ -1,0 +1,23 @@
+# Cadence Style Guide
+
+## Line length
+
+## Comments
+### High level documentation at the begining of files (contracts, transactions and scripts)
+### Commenting functions
+
+## Identation
+
+## Whitespace between code lines
+
+## Declaring variables and constants
+
+## Bracket spaces
+
+## Panic messages
+
+## Naming and capitalization
+### Contracts
+### Fields
+### Functions
+### Paths
--- a/docs/cadence-style-guide.md
+++ b/docs/cadence-style-guide.md
@@ -1,25 +1,65 @@
 # Cadence <- Style Guide
 ## Table of contents
+1. [Identation](#identation)
 1. [Line length](#line-length)
 1. [Comments](#comments)
-1. [Identation](#identation)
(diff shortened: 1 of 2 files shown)


joshuahannan (commented):
Looks good so far! Maybe we can also add a section for access control modifier naming? `access(self)` and `priv` currently mean the same thing, as does `access(all)` and `pub`, so we could give recommendations about that

alilloig:
> Looks good so far! Maybe we can also add a section for access control modifier naming? `access(self)` and `priv` currently mean the same thing, as does `access(all)` and `pub`, so we could give recommendations about that

Yeah that makes a lot of sense! I´m prone to access(self) and pub.

turbolent:
Great idea to start on this! I have a lot of thoughts :-) 

I would recommend proposing this as a FLIP, so everyone can chime in

turbolent:
Some pointers:
- [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/)
- [Google's Swift Style Guide](https://google.github.io/swift/)
- [Kotlin Coding Conventions](https://kotlinlang.org/docs/coding-conventions.html)
- [Rust Style Guidelines](https://doc.rust-lang.org/1.0.0/style/)

alilloig:
Great thank you!! I was randomly looking at the Go style guide or the airbnb js style guide for instance

joshuahannan:
@turbolent Why do you think this needs to be a FLIP? It isn't a new big feature or anything. Couldn't the discussion just happen in this issue or in a PR?

turbolent:
As far as I understand, this guideline is going to recommend a certain style to be the "preferred" ("correct?") style, which will then also have implications for tooling, e.g. the code formatter should probably generate code in that style.

It is a good idea to involve the community and ask for feedback.

For example, [Rust even has a "meta RFC" for the *process* of proposing style guide changes](https://rust-lang.github.io/rfcs/1607-style-rfcs.html).

bluesign:
This is a great initiative. I also think this can be FLIP, though I am in more favour of having it's on repository. 

Problem with this being FLIP is, this feels like ruleset, and discussing different rules in a single place will not be enough. If we make this a repo: 

- We can create issues for questions, where they can be answered. 
- We can create PRs for proposals, where they can be discussed separately and merged or closed. 

Also I think this can be good place for ideas to grow for extending Cadence standard library or fix common problems with the wrong usage of the language. For example: when I last reviewed some contracts for community, I figured out there are very common misconceptions, but couldn't find a place to put those as feedback.

pgebheim:
I agree with @bluesign  -- a FLIP should be seen as a repository of technical standards or governance actions, which provide clear reasonsing, argumentation, transparency, and history for decision making and consensus.

A style guide is inherently a guide and doesnt actually define interoperability standards.

As such, I think this makes sense as its own repo and can be the schelling point for where people start, but I would expect this guide to be forked depending upon the audience over time.

turbolent:
I wasn't suggesting that the style guide should become part of the specification, but once it is in the Cadence repo, it might be seen as such. In any case, we should make sure to make this clear.

My main point was that we already have a process for proposals to Flow, FLIPs, and we should use it instead of having an ad-hoc discussion in this PR.  

> FLIP should be seen as a repository of technical standards or governance actions, which provide clear reasoning, argumentation, transparency, and history for decision making and consensus.
> A style guide is inherently a guide and doesn't actually define interoperability standards.

I don't think we ever said anywhere that FLIPs are just for technical/interoperability standards and governance actions, I always just saw them as a means of having a process for any kind of proposal, with the main benefits being the ones you stated @pgebheim (reasoning, consensus, etc.). 

If a FLIP isn't the right place, then by all means we can also have the same discussion here.

bluesign:
@turbolent My comment was more about discussing many unrelated items in a single FLIP ( or single issue ) is not being productive. 

Neither FLIP nor single issue is good for discussing totally unrelated rules in a ruleset. But also they are too related to separate into different FLIPs. 

My proposal was to make style guide a separate repository. So each item can be discussed around issues or PRs. 

Also I think this can be little more expanded into best practices ( not stuck in style only )

alilloig:
I also really like the idea of having a repo for gathering all cadence best practices. Structuring the discussion around each point in one PR/issue also makes a lot of sense!! Even if we are not strictly just talking about the code style guide on the repo the `cadence-style-guide` name seems right to me.

turbolent:
I haven't yet, please do, that would be much appreciated! 👍

alilloig:
do we all feel good with `cadence-style-guide` as the repo name?

PR #1733, opened 2022-06-17
Allow decoding of types in pre-0.3.0 format (just type ID)

turbolent:
## Description

Make it possible to use the current decoder for values that were encoded in the pre-0.3.0 format, where static types were encoded as a type ID.

For example, this allows off-chain consumers to decode older events.

______

### DIFF
--- a/encoding/json/decode.go
+++ b/encoding/json/decode.go
@@ -35,17 +35,30 @@ import (
 
 // A Decoder decodes JSON-encoded representations of Cadence values.
 type Decoder struct {
-	dec   *json.Decoder
-	gauge common.MemoryGauge
+	dec                          *json.Decoder
+	gauge                        common.MemoryGauge
+	allowUnstructuredStaticTypes bool
+}
+
+type Option func(*Decoder)
+
+func WithAllowUnstructuredStaticTypes(allow bool) Option {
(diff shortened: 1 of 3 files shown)


encoding/json/decode.go
```diff
@@ -1045,11 +1045,13 @@ func (d *Decoder) decodeType(valueJSON interface{}, results typeDecodingResults)
...
 
-	typeID, ok := valueJSON.(string)
-	if ok {
+	if typeID, ok := valueJSON.(string); ok {
 		if result, ok := results[typeID]; ok {
 			return result
 		}
+
+		// Backwards-compatibility for format <0.3.0
+		return cadence.TypeID(typeID)
```
SupunS, line 1054: Probably need to add this behind a flag, because this is also used for decoding program arguments, which could cause problems

turbolent: Good point, added in [85b3ed3](https://github.com/onflow/cadence/pull/1733/commits/85b3ed329f0fe7769bcb334d50ca64aaed5bd9de)

turbolent: Also added a test case in 942041d98b4055a35a0c0ae7015eeeaf991053e8

SupunS (approved):
👍

PR #1735, opened 2022-06-20
Change script and transaction locations to arrays

turbolent:
Closes #???

## Description

Scripts and transactions are always identified by a 256-bit identifier. 

Make the location types 32-byte arrays. With this change all locations can be used directly as `map` keys, locations can be compared, and a conversion to a location ID is not necessary.

This or a future PR should refactor other cases where location IDs are used as keys for indices (`map`) and use locations directly.
______


turbolent:
Benchmarked this locally, shows a good performance improvement:

```
name                             old time/op    new time/op    delta
RuntimeFungibleTokenTransfer-12     758µs ±17%     651µs ± 9%  -14.16%  (p=0.000 n=20+19)
```

PR #1736, opened 2022-06-20
Add support for adding breakpoints to the debugger

turbolent:
____


bluesign:
Do we want debugger to `step into` each statement ? 

Can't we add just some statement to Cadence like 'debugger()' which will trigger callback to debugger? 

So to put breakpoint, debugger will just replace statement at location with `debugger()` statement, it will not care about other statements executed. When breakpoint hit it will put the original statement at location. 

Same callback also can be used to transfer context after pause etc.

turbolent:
@bluesign 

> Do we want debugger to step into each statement ?

Yes, the debugger should be able to step all statements of the program. It already does at the moment, see the `Debugger.Next` function.

> Can't we add just some statement to Cadence like 'debugger()' which will trigger callback to debugger?

Yes, a function like `debugger()` which triggers a pause in the debugger could be added.

> So to put breakpoint, debugger will just replace statement at location with debugger() statement, it will not care about other statements executed. When breakpoint hit it will put the original statement at location.

The debugger UI would need to perform this source modification. The debugging API we are currently trying to implement, the [Debug Adapter Protocol](https://microsoft.github.io/debug-adapter-protocol/), has high-level functions, like [`SetBreakpoints`](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_SetBreakpoints) to set breakpoints.

In addition, this approach will also allow us to e.g. add support for function breakpoints, conditional breakpoints, "logpoints", etc.

runtime/debugger_test.go
```diff
@@ -115,3 +117,89 @@ func TestRuntimeDebugger(t *testing.T) {
...
+
+		runtime.SetDebugger(debugger)
+
+		err := runtime.ExecuteTransaction(
+			Script{
+				Source: []byte(`
+                  transaction {
+                      prepare(signer: AuthAccount) {
+			    	      let answer = 42
+                          log("Hello, World!")
```
SupunS, line 170:
```suggestion
                          let answer = 42
                          log("Hello, World!")
```

SupunS (approved):
Nice! 👏

PR #1739, opened 2022-06-21
[LS] Use a hosted emulator from flowkit

### LINKED ISSUE #1691: [LS] Use hosted emulator
The language server expects the client to handle the blockchain state and execution which in turn makes every client duplicate that logic and defies the purpose of avoiding the matrix approach.
The language server should be changed in a way to leverage the flowkit hosted emulator feature, with that it could manage the blockchain state by itself.

sideninja:
Closes #1691 

## Description
This PR changes the usage of the emulator. It now uses an emulator as a package programmatically leveraging the flowkit implementation instead of relaying on a separate emulator service being spawned. This makes the language server less complex and avoids emulator state synchronization. 

______

### DIFF
--- a/languageserver/test/index.test.ts
+++ b/languageserver/test/index.test.ts
@@ -53,7 +53,6 @@ async function withConnection(f: (connection: ProtocolConnection) => Promise<voi
   if (enableFlowClient) {
     initOpts = {
       configPath: "./flow.json",
-      emulatorState: 1,
       activeAccountName: "service-account",
       activeAccountAddress: "0xf8d6e0586b0a20c7"
     }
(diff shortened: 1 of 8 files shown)


languageserver/test/index.test.ts
```diff
@@ -48,12 +49,22 @@ async function withConnection(f: (connection: ProtocolConnection) => Promise<voi
 
   connection.listen()
 
+  let initOpts = null
+  if (enableFlowClient) {
+    initOpts = {
+      configPath: "./flow.json",
+      activeAccountName: "service-account",
+      activeAccountAddress: "0xf8d6e0586b0a20c7"
```
turbolent, line 57: What's this address? Maybe add a comment explaining it

Issue #1740, opened 2022-06-21
Refactor parser error handling


SupunS:
### Issue To Be Solved

This is a follow-up improvement from https://github.com/onflow/cadence/issues/1708. PR https://github.com/onflow/cadence/pull/1718 refactors interpreter/runtime errors, and we need to do the same for parser errors.
i.e: replace fmt.Errorf() with a known ParseError

Once this is done, it is also possible to change parser's error recovery (https://github.com/onflow/cadence/pull/1718#discussion_r902016383) to be refactored to only recover `ParserError`s

### Related Issue:
https://github.com/onflow/cadence/issues/1659

> when a syntax problem is encountered there are currently four (maybe more?) “kinds” how that is reported: 
> - `panic` with a string, e.g. `panic("expected comma, got start of parameter")`
> - `panic` with a “plain” error, e.g. `panic(fmt.Errorf("expected parameter or end of parameter list, got %s", p.current.Type))`
> - `parser.report` with a “plain” error, e.g. `report(fmt.Errorf("missing start of string literal: expected '\"'"))`
> - `parser.report` with a `SyntaxError`
> `panic`s abort the parse immediately, and if the value that’s recovered (e.g. string or plain error) is not a `SyntaxError`, it is automatically wrapped into one at the current parse location

PR #1746, opened 2022-06-22
Refactor parser errors

### LINKED ISSUE #1659: Clean up parse error handling

SupunS:
Closes #1659, #1740

## Description

Depends on https://github.com/onflow/cadence/pull/1718

- Convert all reported errors to known syntax errors (`SyntaxError`)
- Report/Recover only `ParserErrors`s. Non-parser-errors are propagated up the call stack
______


runtime/literal.go
```diff
@@ -82,7 +81,7 @@ func ParseLiteralArgumentList(
 	parameterCount := len(parameterTypes)
 
 	if argumentCount != parameterCount {
-		return nil, fmt.Errorf(
+		return nil, parser.NewUnpositionedSyntaxError(
```
dsainati1, line 84: Are we planning on. going back to add positions here (and to the other unpositioned errors generally)? Errors that don't tell you their source code location are pretty unhelpful.

runtime/parser/declaration.go
```diff
@@ -596,7 +594,7 @@ func isNextTokenCommaOrFrom(p *parser) bool {
 	}
 }
 
-func parseHexadecimalLocation(memoryGauge common.MemoryGauge, literal string) common.AddressLocation {
+func parseHexadecimalLocation(p *parser, memoryGauge common.MemoryGauge, literal string) common.AddressLocation {
```
dsainati1, line 597: Doesn't seem like we need to pass the latter two arguments to this function anymore since they are both accessible from the parser.

runtime/parser/expression.go
```diff
@@ -1348,15 +1344,15 @@ func defaultExprMetaLeftDenotation(
 	return result, false
 }
 
-func exprLeftBindingPower(token lexer.Token) int {
+func exprLeftBindingPower(p *parser, token lexer.Token) int {
```
dsainati1, line 1347: similar to above, this function probably only needs the parser argument.

runtime/literal.go
SupunS: For some reason, this `ParseLiteralArgumentList` method is only used within tests. So the position info is not available. Not sure why it is not in test-utils or something like that (maybe it was used in the parser before, but not anymore?), but I just let it be.

runtime/parser/declaration.go
SupunS: Ah yes. Refactored in c42f4ae

dsainati1 (approved):

runtime/parser/expression.go
```diff
@@ -1357,36 +1354,32 @@ func applyExprNullDenotation(p *parser, token lexer.Token) ast.Expression {
...
-		panic(fmt.Errorf("unexpected token in expression: %s", token.Type))
+		p.panicSyntaxError("unexpected token in expression: %s", token.Type)
 	}
 	return leftDenotation(p, token, left)
 }
 
 // parseStringLiteral parses a whole string literal, including start and end quotes
 //
-func parseStringLiteral(literal string) (result string, errs []error) {
-	report := func(err error) {
```
turbolent, line 1376: Nice cleanup!

turbolent (approved):
Nice work!

PR #1764, opened 2022-06-24
Refactor parser to always return syntax errors

### LINKED ISSUE #1650: Refactor parser from using panics to error return values
### Context

The parser is currently using panics to abort parsing when an unrecoverable parse error is encountered.

### Definition of Done

Replace panics with an explicit error return value

SupunS:
Closes #1650

## Description

Depends on #1718 and #1746

______


runtime/parser/expression.go
```diff
@@ -161,29 +172,36 @@ func defineExpr(def any) {
...
 
 	default:
 		panic(errors.NewUnreachableError())
 	}
+
+	// It is OK to panic here, since this method is only called inside `init()`.
+	// Returning the error from here explodes the error handling in `init` function.
+	// So panic it here, which has the same effect.
+	if err != nil {
+		panic(err)
```
turbolent, line 203: It's already refactored now, but we can keep the panics in the parser "setup" functions, like in `setExprLeftDenotation` – they're only called when the parser package is initialized, not during usage/parsing itself

runtime/parser/expression.go
```diff
@@ -590,7 +632,8 @@ func defineLessThanOrTypeArgumentsExpression() {
 
 			var argumentsStartPos ast.Position
 
-			(func() {
+			// Ignore the error
+			_ = func() error {
```
turbolent, line 636: We can now use the error here to determine if parsing as an invocation succeeded and can remove `isInvocation` now (basically `isInvocation` iff `err != nil`)

runtime/parser/transaction.go
```diff
@@ -52,15 +52,21 @@ func parseTransactionDeclaration(p *parser, docString string) *ast.TransactionDe
 
 	var parameterList *ast.ParameterList
 	if p.current.Is(lexer.TokenParenOpen) {
-		parameterList = parseParameterList(p)
+		parameterList, _ = parseParameterList(p)
```
turbolent, line 55: Does the error need to be handled here?

runtime/parser/expression.go
```diff
@@ -509,28 +532,47 @@ func init() {
...
 				left,
 				tokenRange.EndPos,
-			)
+			), nil
 		},
 	})
 
 	defineNestedExpression()
-	defineInvocationExpression()
+	err = defineInvocationExpression()
```
turbolent, line 545: As noted above, these setup functions can just keep panicing, they'll never be called in a way where the error needs to be handled specifically

turbolent (commented):
Great work!

runtime/parser/expression.go
SupunS: reverted in 30f9a812d5725ccbbda8a4ca0a4c823738d709af

runtime/parser/transaction.go
SupunS: Ah yes, this was automatically added (and hence ignored) by the IDE. Fixed it in 30f9a812d5725ccbbda8a4ca0a4c823738d709af

PR #1783, opened 2022-07-05
Fail invalidated-resource use with a proper error

### LINKED ISSUE #1781: Cadence Error in testnet
### Error
 [Error Code: 1101] cadence runtime error Execution failed:
error: runtime error: invalid memory address or nil pointer dereference

### Transaction
https://testnet.flowscan.org/transaction/7d98e36a67133618a706049b5e5e64de10ec8af5b9f71e1a3dd5a3f7ddf8b7d1

### Description
These functions work fine around a month ago. Emulator won't throw any error.  Addtional tests show that it works fine before A.203ea2804f661311.LockProxy: 212 (tx is supposed to end in line 216), so I think maybe it is not a contract error. Actually, it looks from golang.

SupunS:
Closes #1781

## Description

Trying to get the `v.StorageID()` can cause nil-pointer dereference error if the resource is already invalidated. Hence move the `v.StorageID()` below the invalidated resource check, to give a proper error, instead of a nil-pointer.

______

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -14327,11 +14327,11 @@ func (v *CompositeValue) Destroy(interpreter *Interpreter, getLocationRange func
 
 	interpreter.ReportComputation(common.ComputationKindDestroyCompositeValue, 1)
 
-	storageID := v.StorageID()
-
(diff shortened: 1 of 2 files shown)


turbolent (approved):
👍

PR #1307, opened 2021-12-13
Reject duplicate cases in switch-statement

SupunS:
Closes #530

## Description

This is not an urgent PR. Just a nice to have feature I worked on sometimes back.

______

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #1787, opened 2022-07-06
Separate type equality checker from contract update validator

SupunS:
## Description

This PR separates the AST type equality checker from the contract update validator, so that type equality checking can be re-used. No change is done to the checker's logic.

This is needed for moving the [duplicate switch-case checking](https://github.com/onflow/cadence/pull/1307) to the linter repo. Because the switch-case checking currently depends on the type-checker to do the type equality check. But once it is moved out from the type-checker, it no longer can rely on the checker. Instead, the same 'type equality check' can be performed using the AST only, similar to how it is done in the contract update validator. 

______

### DIFF
--- a/runtime/contract_update_validation.go
+++ b/runtime/contract_update_validation.go
@@ -23,15 +23,15 @@ import (
 
 	"github.com/onflow/cadence/runtime/ast"
 	"github.com/onflow/cadence/runtime/common"
-	"github.com/onflow/cadence/runtime/sema"
 )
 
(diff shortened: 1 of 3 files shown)


turbolent (commented):
I guess the code was literally just moved to the new file? Or where there any additional changes needed?

SupunS:
Yes, just moved. No changes in the logic.

turbolent (approved):
Nice!

PR #1789, opened 2022-07-06
Update go.js in NPM packages to Go 1.18.1

turbolent:
## Description

#1786 improved the `cadene-parser` NPM package. @SaveTheRbtz noted the `go.js` file is outdated. 

Update the `go.js` files in the NPM packages to Go 1.18.1. 

This also reverts the refactor to TypeScript, which makes updating the file unnecessarily difficult / hard to maintain and adds no value, as we are not owning this code, but rather just ship it.

______

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #1786, opened 2022-07-06
Make `@onflow/cadence-parser` support node.js (Hybrid ESM/CJS)

Pure-Peace:
After merging this, `@onflow/cadence-parser` can be used directly in node.js,
developers can easily use this library to develop some libraries (such as a library for generating contract schemas, similar to typechain).

And I tested its use in `tools/astexplorer` (browser), everything works fine.

```javascript
const {CadenceParser} = require('@onflow/cadence-parser');
const fs = require('fs');
const path = require('path');

(async () => {
  const parser = await CadenceParser.create(
    await fs.promises.readFile(path.join(__dirname, './node_modules/@onflow/cadence-parser/dist/cadence-parser.wasm'))
  );

  const ast = parser.parse(`
  pub contract HelloWorld {
    pub fun hello() {
      log("Hello, world!")
    }
  }
`);

  console.log(ast);
})();
```

## Description

______

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [ ] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #1790, opened 2022-07-06
`@onflow/cadence-parser` bump to go 1.18 `wasm_exec.js`

Pure-Peace:
Improve for #1786
https://github.com/onflow/cadence/pull/1786#discussion_r915109554
https://github.com/onflow/cadence/pull/1786#discussion_r915143777

#1789

## Description

______

### DIFF
--- a/npm-packages/cadence-parser/README.md
+++ b/npm-packages/cadence-parser/README.md
@@ -5,8 +5,10 @@ so it can be used in tools written in JavaScript.
 
 ## Usage
 
+Broswer
+
 ```js
-import {CadenceParser} from "@onflow/cadence-parser"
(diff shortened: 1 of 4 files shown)


Pure-Peace:
> Thank you for the improvements! I had not realized you opened this at almost the same time I made the same changes in #1789.
> 
> Keeping the wasm_exec.js file as close to the original as possible means we can update it more easily, and type annotations add little, as the code is not owned by the Cadence project, so I reverted the refactor to TS.
> 
> However, it would still be nice to get your README and build script improvements in 👍

Yes, we should reduce some unnecessary maintenance costs, so I am thinking of some other modifications.
Revert the ts refactoring first, then add something like polyfill node.js... It should be better.

PR #1785, opened 2022-07-06
Fix documentation, add linter

turbolent:
## Description

- Fix all the broken links
- Add a linter that checks the links and integrate it into CI
- Improve operator docs
______

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #1795, opened 2022-07-06
Repair markdown for new site.

10thfloor:
## Description

Fixes markdown rendering errors in docs as a result of upgrading the docs site from MDX v1 -> v2
Description of migration steps here: https://mdxjs.com/migrating/v2/

Original issue here: https://github.com/onflow/next-docs-v1/issues/292

______

### DIFF
--- a/docs/tutorial/10-resources-compose.mdx
+++ b/docs/tutorial/10-resources-compose.mdx
@@ -132,6 +132,7 @@ These definitions show how a Kitty resource could own hats.
 
 The hats are stored in a variable in the Kitty resource.
 
+```cadence
     // place where the Kitty hats are stored
     `pub let items: <-{String: KittyHat}`
 
--- a/docs/tutorial/10-resources-compose.mdx
+++ b/docs/tutorial/10-resources-compose.mdx
@@ -134,8 +134,7 @@ The hats are stored in a variable in the Kitty resource.
 
 ```cadence
     // place where the Kitty hats are stored
-    `pub let items: <-{String: KittyHat}`
-
+    pub let items: <-{String: KittyHat}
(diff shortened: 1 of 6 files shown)


docs/tutorial/10-resources-compose.mdx
```diff
@@ -133,7 +133,7 @@ These definitions show how a Kitty resource could own hats.
 The hats are stored in a variable in the Kitty resource.
 
     // place where the Kitty hats are stored
```
turbolent, line 135: Wrap in a code fence instead:

```suggestion
```cadence
    // place where the Kitty hats are stored
```

docs/tutorial/10-resources-compose.mdx
```diff
@@ -133,7 +133,7 @@ These definitions show how a Kitty resource could own hats.
...
-    pub let items: <-{String: KittyHat}
+    `pub let items: <-{String: KittyHat}`
```
turbolent, line 137:
```suggestion
    pub let items: <-{String: KittyHat}
```
```

turbolent (approved):
Nice! I also fixed most of these yesterday in #1785

turbolent:
@10thfloor Please follow/fill out the check list, thanks

docs/tutorial/10-resources-compose.mdx
10thfloor: This needs to be in a code fence, since MDX2 will interpret the { } as an JS expression template, and fail to render the page.

MDX 2 expects anything inside { } to be valid JS and There is no way to turn this off afaik.

turbolent: Yep, see above. My suggestion was to wrap it in a block code fence (```), instead of an inline one (`)

turbolent (approved):

PR #1797, opened 2022-07-07
refactor ExecuteTransaction body into Executor object

pattyshack:
## Description

The original ExecuteTransaction body was split into multiple methods (prep work for parallelizing transaction preprocessing)
______

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -43,6 +43,59 @@ type Script struct {
 
 type importResolutionResults map[common.Location]bool
 
+// Executor is a continuation which represents a full unit of transaction/script
+// execution.
+//
+// The full unit of execution is divided into stages:
+//  1. Preprocess() initializes the executor in preparation for the actual
+//     transaction execution (e.g., parse / type check the input).  Note that
+//     the work done by Preprocess() should be embrassingly parallel.
+//  2. Execute() performs the actual transaction execution (e.g., run the
+//     interpreter to produce the transaction result).
+//  3. Result() returns the result of the full unit of execution.
+//
+// (TODO(patrick): maybe add Cleanup/Postprocess in the future)
+type Executor interface {
+	// Preprocess prepares the transaction/script for execution.
+	//
+	// This function returns an error if the program has errors (e.g., syntax
+	// errors, type errors).
+	//
+	// This method may be called multiple times.  Only the first call will
+	// trigger meaningful work; subsequent calls will return the cached return
+	// value from the original call (i.e., an Executor implementation must
+	// guard this method with sync.Once).
+	Preprocess() error
+
+	// Execute executes the transaction/script.
+	//
+	// This function returns an error if Preprocess failed or if the execution
+	// fails.
+	//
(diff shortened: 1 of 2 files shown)


pattyshack:
I'm not familiar with the code base.  If parts of Preprocess should remain in Execute, let me know and I'll adjust accordingly.

turbolent:
@pattyshack Can you please follow and fill out the template that was provided (https://github.com/onflow/cadence/blob/master/.github/pull_request_template.md)?

turbolent (commented):
Just double checking: On the flow-go/FVM side, have we made sure that we can do this optimization, i.e. that contract updates are only effective at the end of the block, and not in the following transaction?

runtime/runtime.go
```diff
@@ -43,6 +43,59 @@ type Script struct {
...
+//
+// The full unit of execution is divided into stages:
+//  1. Preprocess() initializes the executor in preparation for the actual
+//     transaction execution (e.g., parse / type check the input).  Note that
+//     the work done by Preprocess() should be embrassingly parallel.
+//  2. Execute() performs the actual transaction execution (e.g., run the
+//     interpreter to produce the transaction result).
+//  3. Result() returns the result of the full unit of execution.
+//
+// (TODO(patrick): maybe add Cleanup/Postprocess in the future)
```
turbolent, line 57:
```suggestion
// TODO: maybe add Cleanup/Postprocess in the future
```

turbolent (commented):
@pattyshack There are still some build errors, can you please fix them?

turbolent:
@janezpodhostnik @ramtinms Can you maybe help here in answering the question above, i.e. do contract updates become effective in the following transaction, or in the following block?

pattyshack:
done.  my bad.

janezpodhostnik:
Contract updates are done at the end of the transaction currently https://github.com/onflow/flow-go/blob/de5c66130a83a6fdc015e91ed1b6ec8ac2f1e93d/fvm/transactionEnv.go#L1255. 

> Each flow transaction is compose of at least 3 cadence transactions. These 3 cadence transactions should be easily parallelizable (with a crap ton of refactoring)

The 2 "inner transactions" (fee deduction, and storage checks, and potentially more depending on the transaction) are not actually using a script, they are calling InvokeContractFunction, not sure that this would fit in this model. Maybe if the `InvokeContractFunction` would also be moved to the Executor (and perhaps `GetValue` as well).

Transaction Invocation, storage check and fee deduction are currently not trivially paralelizable as you need to:
- invoke the transaction to change storage to run the storage check
- deduct the fees to run the storage check (storage depends on the balance of the same vault that the fees are deducted from)

you could potentially run transaction invocation and fee deduction at the same time, but this is not true for all transactions.

The two things that could already be run in parallel are the two branches we have for successful transactions and for failed transactions:

1. In case of success: invoke transaction, deduct fees, check storage
2. In case of failure: deduct fees

If branch 1 fails we need to do branch 2, which could have already been running while 1 was running.

turbolent (approved):

PR #1798, opened 2022-07-07
Update anti-patterns.mdx

masterEye-07:
Update comment in "Array or dictionary fields should be private"


joshuahannan (approved):
Thanks!

PR #1775, opened 2022-06-30
[LS] Update flowkit event API change

sideninja:
## Description
This PR implements changes made to the flowkit API in the PR: https://github.com/onflow/flow-cli/pull/538

______

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #1801, opened 2022-07-08
Update language server to use new linter for hints

dsainati1:
Closes https://github.com/onflow/cadence/issues/1269

This updates the language servers hint architecture to use the new linter instead of Cadence's hints, which have been removed. 

______


languageserver/server/server.go
```diff
@@ -3116,7 +3137,22 @@ func convertHint(
 				},
 			}
 		}
+		protocolDiagnostic = protocol.Diagnostic{
+			Message: linterDiagnostic.Message,
+			// protocol.SeverityHint doesn't look prominent enough in VS Code,
+			// only the first character of the range is highlighted.
+			Severity: protocol.SeverityInformation,
+			Range:    protocolRange,
+		}
```
turbolent, line 3146: Duplicated here and below. Maybe refactor out the assignment to `protocolDiagnostic` after the switch. The only difference in the cases is the message

turbolent (commented):
Nice!

languageserver/server/server.go
```diff
@@ -1963,6 +1984,8 @@ func (s *Server) getDiagnostics(
 		diagnostics = append(diagnostics, diagnostic)
 	}
 
+	analysisProgram.Run(maps.Values(linter.Analyzers), report)
```
turbolent, line 1987: Might be good to "cache" the result of `maps.Values(linter.Analyzers)`, maybe even on the linter side, e.g. through a getter function that memoizes

languageserver/server/server.go
```diff
@@ -1953,8 +1959,23 @@ func (s *Server) getDiagnostics(
...
+		Elaboration: checker.Elaboration,
+		Location:    checker.Location,
+		Code:        text,
+	}
+
+	var reportLock sync.Mutex
+
+	report := func(linterDiagnostic analysis.Diagnostic) {
+		reportLock.Lock()
+		defer reportLock.Unlock()
```
turbolent, line 1977: Is this lock required because the analysis below runs concurrently?

dsainati1: Yeah, it looks like it based on the implementation in `tools/analysis/program.go` as well as this being how the linter itself implements its report function.

turbolent: Probably not a big deal, but there might be some contention here. We might want to change `analysis.Program.Run` to not accept a single reporting function that is used for all diagnostics of all given analyzers, but report diagnostics for each analyzer separately, like Go does: https://github.com/golang/tools/blob/d6c099e3c1a39214a49a5b7c5c5faa604650e581/go/analysis/internal/checker/checker.go#L687.
We can do that in separate PR(s) though.

languageserver/server/server.go
```diff
@@ -3116,7 +3133,18 @@ func convertHint(
 				},
 			}
 		}
+		message = linterDiagnostic.Message
+	default:
+		message = linterDiagnostic.Message
 	}
```
turbolent, line 3139:
```suggestion
	}
	
	if message == "" {
		message = linterDiagnostic.Message
	}
```

turbolent (approved):

psiemens:
Is this good to merge? It's blocking https://github.com/onflow/cadence/pull/1775, which we need for some CLI work. (cc @sukantoraymond).

Thanks for making this change BTW! The new linter package looks awesome.

dsainati1:
This is blocked on https://github.com/onflow/cadence-lint/pull/3, but as soon as that merges I will merge this as well

PR #1804, opened 2022-07-12
Use generics for ordered map, lazily allocate inner map and list

turbolent:
## Description

- Use proper Go generics for the ordered map implementation
- Remove all the generated implementations 
- Optimize the allocations of the inner map and list by making them lazy

______


runtime/common/orderedmap/orderedmap.go
```diff
@@ -108,23 +121,39 @@ func (om *KeyTypeValueTypeOrderedMap) Delete(key KeyType) (oldValue ValueType, p
 }
 
 // Len returns the length of the ordered map.
-func (om *KeyTypeValueTypeOrderedMap) Len() int {
+func (om OrderedMap[K, V]) Len() int {
+	if om.pairs == nil {
+		return 0
+	}
+
```
SupunS, line 128: I think builtin `len()` function already handles `nil`. But no harm having it here I guess

runtime/sema/import.go
```diff
@@ -21,14 +21,15 @@ package sema
...
 	"github.com/onflow/cadence/runtime/ast"
 	"github.com/onflow/cadence/runtime/common"
+	"github.com/onflow/cadence/runtime/common/orderedmap"
 )
 
 // Import
 
 type Import interface {
-	AllValueElements() *StringImportElementOrderedMap
+	AllValueElements() *orderedmap.OrderedMap[string, ImportElement]
```
SupunS, line 30: maybe define a type/alias for `*orderedmap.OrderedMap[string, ImportElement]`

runtime/sema/type.go
```diff
@@ -155,7 +156,7 @@ type Type interface {
 	//
 	Unify(
 		other Type,
-		typeParameters *TypeParameterTypeOrderedMap,
+		typeParameters *orderedmap.OrderedMap[*TypeParameter, Type],
```
SupunS, line 159: similarly, maybe it's worth defining type aliases for these commonly used map types.

SupunS (approved):
`+400 −3,235 ` 😍

runtime/sema/import.go
turbolent: Good point! Added in [9045c19](https://github.com/onflow/cadence/pull/1804/commits/9045c1981258031fe0f9946dea9c4963b44d7e9d). That shaved off another ~100 LOC :-)

runtime/sema/type.go
turbolent: Added in [9045c19](https://github.com/onflow/cadence/pull/1804/commits/9045c1981258031fe0f9946dea9c4963b44d7e9d)

SupunS (approved):

SaveTheRbtz:
Yey!  Once available we could also switch to a generic-based version of containers (e.g. `container/list` is still interface-based)

PR #1807, opened 2022-07-12
Don't copy code string if there's no error

pattyshack:
Closes #???

## Description

The code string is only needed for pretty printing on error.  We don't need to make a copy of the string during normal parsing

______

### DIFF
--- a/runtime/errors.go
+++ b/runtime/errors.go
@@ -39,10 +39,17 @@ type Error struct {
 }
 
 func newError(err error, context Context) Error {
+	codes := make(map[common.Location]string, len(context.codes))
+
+	// Regardless of iteration order, the final result will be the same.
+	for location, code := range context.codes { //nolint:maprangecheck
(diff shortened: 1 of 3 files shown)


runtime/errors.go
```diff
@@ -39,10 +39,15 @@ type Error struct {
 }
 
 func newError(err error, context Context) Error {
+	codes := make(map[common.Location]string, len(context.codes))
+	for location, code := range context.codes {
```
pattyshack, line 43: I'm not really sure why ranging over a map is a lint error in cadence.  Could you give me a quick explanation?

turbolent (approved):
Good idea! Maybe we should look into how the code of transactions and scripts are passed from requests in EN/FVM all the way to Cadence, and how Cadence fetches contract code from FVM – there might be many accidental/unintentional `[]byte`/`string` conversions along the way

PR #1808, opened 2022-07-12
Switch to generic list

turbolent:
## Description

As suggested in https://github.com/onflow/cadence/pull/1804#issuecomment-1182579272, use a generic version of the linked list in the ordered map implementation. 

We can switch to the one in the standard library once/if it becomes available in a future release of Go.

______

### DIFF
--- a/check-headers.sh
+++ b/check-headers.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-files=$(find . -name \*.go -type f -print0 | xargs -0 grep -L -E '(Licensed under the Apache License)|(Code generated (from|by))')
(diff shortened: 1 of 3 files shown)


SaveTheRbtz (approved):
ARM64 shows a nice 3% perf bump:
```
$ benchstat old new                                                                                                               
name                              old time/op    new time/op    delta
ContractInterfaceFungibleToken-8    22.1µs ± 2%    21.5µs ± 2%  -2.42%  (p=0.000 n=10+9)

name                              old alloc/op   new alloc/op   delta
ContractInterfaceFungibleToken-8    25.8kB ± 0%    25.1kB ± 0%  -2.79%  (p=0.000 n=10+10)

name                              old allocs/op  new allocs/op  delta
ContractInterfaceFungibleToken-8       441 ± 0%       441 ± 0%    ~     (all equal)
```

Issue #1806, opened 2022-07-12
Syntax and checking for immutable values and functions


dsainati1:
### Issue To Be Solved

A number of issues and PRs currently would benefit from the ability to know whether or not a function is "pure" or has the potential to mutate state. Examples of issues that would be helped by this would be https://github.com/onflow/cadence/issues/1805, https://github.com/onflow/cadence/issues/1304, https://github.com/onflow/flow/pull/1043 and the follow-up work on https://github.com/onflow/flow/pull/703. 

### Suggested Solution

This would add a new piece of syntax to the language allowing users to annotate their functions as "pure" or "impure". The exact nature of this syntax is up for consideration (What specific names would we use? What would be the default purity for an unannotated function? Which functions require annotations?), but this syntax would then be used to power a semantic analysis as part of type checking to enforce that "pure" functions do not have any side effects: i.e. they do not modify or assign to any external state, and they only call "pure" functions within their bodies. "Impure" functions would have no such restrictions on what operations they can perform and which functions they can call.

dsainati1:
A couple thoughts for consideration here:

1) We would probably need/want to require that all public/exported functions be explicitly annotated as pure or impure, because the potential bubble effects of making a change to one function could impact the purity of functions arbitrarily far away. Requiring that public functions be explicitly annotated helps keep the scope of any potential analysis we do small, and also prevents an internal implementation change in one contract from affecting the type checking of another contract. 

2) What would we do in the case of recursive or mutually recursive functions? The easy thing would be to just require an explicit annotation on functions that are recursive, but in theory we could also do some kind of fixed-point analysis. 

3) Do we want to extend this to variables, fields and references? One could imagine a `const`-like declaration for composites or containers that passing them as arguments to or using them as a receiver for any impure functions.

turbolent:
We should clearly define what the attributes of the functions are. For example, GCC has support for [`pure`](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-pure-function-attribute) and [`const`](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-const-function-attribute) functions.

dsainati1:
> If I define a function pure I cannot change it later to impure right?

You can change the purity annotation, but this would cause type errors in any pure functions that use this function, or in any other places where impure functions are not allowed. 

> Isn't pure will be dead on arrival with this? I mean we can say post/pre conditions can call only pure etc but, are there any other use case?
[...]

PR #1818, opened 2022-07-14
Add machinery for parsing and checking `view` functions

dsainati1:
Works towards https://github.com/onflow/cadence/issues/1806

## Description

This adds the beginnings of a purity analysis to determine whether functions are pure or impure (have side effects or not). In the current implementation, annotating a function `view fun foo():T` indicates that the function is pure. If no purity annotation is provided, we default to a function being impure. 

Purity is a subtype of impurity for functions. 

______


runtime/sema/check_function.go
```diff
@@ -173,11 +182,17 @@ func (checker *Checker) checkFunction(
...
-				)
+				isPure := checker.InNewPurityScope(functionType.Purity == PureFunction, func() {
+					checker.visitFunctionBlock(
+						functionBlock,
+						functionType.ReturnTypeAnnotation,
+						checkResourceLoss,
+					)
+				})
+
+				if functionType.Purity == UnknownPurity {
```
dsainati1, line 193: This is the core of the analysis; we check the function in a new "purity scope", which starts pure and then will be monotonically updated to "impure" as soon as an impurity is observed. If we are enforcing purity, as we would do if the function has an explicit `pure` annotation, then this will cause a type error. Otherwise if the function has no annotation we update its purity to be the purity of the body block.

runtime/sema/checker.go
```diff
@@ -381,6 +389,40 @@ func (checker *Checker) IsChecked() bool {
...
+func (checker *Checker) PopPurityScope() PurityCheckScope {
+	scope := checker.CurrentPurityScope()
+	checker.purityCheckScopes = checker.purityCheckScopes[:len(checker.purityCheckScopes)-1]
+	return scope
+}
+
+func (checker *Checker) ObserveImpureOperation(operation ast.Element) {
+	scope := checker.CurrentPurityScope()
+	// purity is monotonic, if we already know this scope is impure, there's no need to continue
+	if !scope.CurrentPurity {
```
dsainati1, line 409: Ideally this should keep the performance of the analysis within the realm of reason since we won't be recording every single impure event, only the first one in each function body.

dsainati1 (commented):
~~This is still a work in progress; things that still need doing:~~

~~1) Actually determining when events are impure. Currently there are no actual impure events, but we can add them by calling `ObserveImpureOperation`, which cause the checker to record the current scope as impure, and if we are inside a function that has been declared to be `pure`, throw an error. Currently impure operations are planned to be:~~
* ~~Calling an impure function~~
* ~~Mutating a value declared outside of the current scope (this includes the function parameters). Note that it won't always be possible to know if a value is external or local (e.g. a value in an array), so we must be conservative here. ~~
* ~~Writing to storage~~
* ~~Mutating a reference (since even if the reference was created in this scope, it's not always possible to know whether it refers to an external value)~~


~~ Some other operations where it is not clear whether they should be impure: ~~
* ~~ Logging a value ~~
* ~~ Emitting an event ~~

~~ 2) Topological sort of top level declarations. Currently code like this does not work with this analysis: ~~
```
fun foo(): (pure (Int): Void) {
    return bar
}

fun bar(_ n: Int) {}
```
~~ Since we currently check `foo` before `bar`, `bar` has not yet had its purity determined when the time comes to check whether it is compatible with the return type of `foo`. If we want to allow purity to be inferred from function bodies, we will need to sort top level function declarations by use so that `bar` is checked first in this scenario. ~~

~~ This also lets us easily deal with the recursive case: any functions that still have indeterminate purity after sorting will necessarily be (mutually) recursive, so they will need explicit purity annotations. ~~


~~ 3) Special treatment of public-level functions. Instead of allowing inference like functions local to a contract or composite, public functions should default to `impure` if no annotation is provided, and require a `pure` annotation in order to be considered `pure`. ~~

EDIT: This comment is no longer true, but I am preserving the text for posterity

dsainati1:
This has been updated to reflect the changes in https://github.com/onflow/flow/pull/1056: functions with no annotation are now `impure` by default, the ability to explicitly annotate something as `impure` has been removed, and there is no plan to add a future inference pass.

robert-e-davidson3 (approved):

turbolent (commented):
Great work!

Is this already implementing the rules as lined out in the FLIP?

dsainati1:
> Is this already implementing the rules as lined out in the FLIP?

Yes, this is just the machinery behind parsing and typechecking purity; so this part remains mostly the same whether or not we have an inference pass.

runtime/interpreter/interpreter.go
```diff
@@ -3134,6 +3136,7 @@ func init() {
 		),
 	)
 
+	// TODO: Add an option to this for a pure function
```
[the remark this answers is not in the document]
turbolent: Ah, that makes sense now. Feel free leave it in 👍

runtime/sema/check_function.go
```diff
@@ -423,7 +433,7 @@ func (checker *Checker) declareBefore() {
 func (checker *Checker) VisitFunctionExpression(expression *ast.FunctionExpression) ast.Repr {
 
 	// TODO: infer
-	functionType := checker.functionType(expression.ParameterList, expression.ReturnTypeAnnotation)
+	functionType := checker.functionType(expression.Purity, expression.ParameterList, expression.ReturnTypeAnnotation)
```
SupunS, line 436: easy on the eyes:
```suggestion
	functionType := checker.functionType(
		expression.Purity,
		expression.ParameterList,
		expression.ReturnTypeAnnotation,
	)
```

runtime/sema/checker.go
```diff
@@ -381,6 +389,40 @@ func (checker *Checker) IsChecked() bool {
 	return checker.isChecked
 }
 
+func (checker *Checker) CurrentPurityScope() PurityCheckScope {
+	return checker.purityCheckScopes[len(checker.purityCheckScopes)-1]
+}
+
+func (checker *Checker) PushNewPurityScope(enforce bool) {
+	checker.purityCheckScopes = append(checker.purityCheckScopes, PurityCheckScope{EnforcePurity: enforce, CurrentPurity: true})
```
SupunS, line 397:
```suggestion
	checker.purityCheckScopes = append(
		checker.purityCheckScopes,
		PurityCheckScope{
			EnforcePurity: enforce,
			CurrentPurity: true,
		},
	)
```

runtime/sema/checker.go
```diff
@@ -595,7 +637,7 @@ func (checker *Checker) checkTopLevelDeclarationValidity(declarations []ast.Decl
 }
 
 func (checker *Checker) declareGlobalFunctionDeclaration(declaration *ast.FunctionDeclaration) {
-	functionType := checker.functionType(declaration.ParameterList, declaration.ReturnTypeAnnotation)
+	functionType := checker.functionType(declaration.Purity, declaration.ParameterList, declaration.ReturnTypeAnnotation)
```
SupunS, line 640:
```suggestion
	functionType := checker.functionType(
		declaration.Purity,
		declaration.ParameterList,
		declaration.ReturnTypeAnnotation,
	)
```

encoding/json/decode.go
```diff
@@ -908,13 +909,26 @@ func (d *Decoder) decodeFieldType(valueJSON any, results typeDecodingResults) ca
...
-func (d *Decoder) decodeFunctionType(returnValue, parametersValue, id any, results typeDecodingResults) cadence.Type {
+func (d *Decoder) decodePurity(purity any) cadence.FunctionPurity {
+	functionPurity := toString(purity)
+	if functionPurity == "pure" {
+		return cadence.PureFunction
+	} else if functionPurity == "impure" {
+		return cadence.ImpureFunction
+	} else {
+		panic(ErrInvalidJSONCadence)
+	}
```
SupunS, line 920: Can we make "purity" optional in the JSON, so that users only have to add it if it is pure (less breaking)

encoding/json/decode.go
```diff
@@ -908,13 +909,26 @@ func (d *Decoder) decodeFieldType(valueJSON any, results typeDecodingResults) ca
 	)
 }
 
-func (d *Decoder) decodeFunctionType(returnValue, parametersValue, id any, results typeDecodingResults) cadence.Type {
+func (d *Decoder) decodePurity(purity any) cadence.FunctionPurity {
+	functionPurity := toString(purity)
+	if functionPurity == "pure" {
+		return cadence.PureFunction
+	} else if functionPurity == "impure" {
```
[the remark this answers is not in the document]
dsainati1: I like the flexibility of keeping it an enum; this way if we choose to expand it later we have the option in place.

encoding/json/decode.go
```diff
@@ -908,13 +909,23 @@ func (d *Decoder) decodeFieldType(valueJSON any, results typeDecodingResults) ca
 	)
 }
 
-func (d *Decoder) decodeFunctionType(returnValue, parametersValue, id any, results typeDecodingResults) cadence.Type {
+func (d *Decoder) decodePurity(purity any) cadence.FunctionPurity {
+	functionPurity := toString(purity)
```
SupunS, line 913: Are there any tests for the field non-existing case? `toString()` seems to be panicking if the value is not a string..

encoding/json/decode.go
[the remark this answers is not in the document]
SupunS: makes sense!

runtime/ast/function_declaration.go
```diff
@@ -26,8 +26,23 @@ import (
 	"github.com/onflow/cadence/runtime/common"
 )
 
+type FunctionPurity int
+
+const (
+	UnspecifiedPurity FunctionPurity = iota
+	ViewFunction      FunctionPurity = 1
```
turbolent, line 33: Enums in the codebase use "prefix-naming". Also, given that iota is used, no need to provide anything for the remaining cases.

```suggestion
	FunctionPurityUnspecified FunctionPurity = iota
	FunctionPurityView
```

runtime/parser/declaration.go
```diff
@@ -78,26 +81,51 @@ func parseDeclaration(p *parser, docString string) (ast.Declaration, error) {
...
 				return parseVariableDeclaration(p, access, accessPos, docString)
 
 			case keywordFun:
-				return parseFunctionDeclaration(p, false, access, accessPos, docString)
+				return parseFunctionDeclaration(p, false, access, accessPos, purity, purityPos, docString)
 
 			case keywordImport:
+				if purity != ast.UnspecifiedPurity {
+					return nil, p.syntaxError("invalid purity modifier for import")
+				}
```
turbolent, line 95: Add test cases for uncovered new code

runtime/parser/declaration.go
```diff
@@ -1157,10 +1212,16 @@ func parseSpecialFunctionDeclaration(
...
+	purityPos *ast.Position,
 	identifier ast.Identifier,
 ) (*ast.SpecialFunctionDeclaration, error) {
 
 	startPos := identifier.Pos
+	// access modifier will precede purity if both exist
+	if purityPos != nil {
+		startPos = *purityPos
+	}
 	if accessPos != nil {
```
turbolent, line 1225: If both purity and access modifier are specified, the latter will override the former:

```suggestion
	} else if accessPos != nil {
```

runtime/parser/expression.go
```diff
@@ -846,8 +846,11 @@ func defineIdentifierExpression() {
 					token.Range.StartPos,
 				), nil
 
+			case keywordView:
+				p.next()
+				return parseFunctionExpression(p, token, ast.ViewFunction)
```
turbolent, line 851: Next token isn't verified to be `fun` keyword and is just skipped

runtime/parser/function.go
```diff
@@ -194,10 +205,16 @@ func parseFunctionDeclaration(
...
+	purityPos *ast.Position,
 	docString string,
 ) (*ast.FunctionDeclaration, error) {
 
 	startPos := p.current.StartPos
+	// access modifier will precede purity if both exist
+	if purityPos != nil {
+		startPos = *purityPos
+	}
 	if accessPos != nil {
```
turbolent, line 218: Same as above:

```suggestion
	} else if accessPos != nil {
```

runtime/parser/statement.go
```diff
@@ -94,7 +94,7 @@ func parseStatement(p *parser) (ast.Statement, error) {
 			return parseForStatement(p)
 		case keywordEmit:
 			return parseEmitStatement(p)
-		case keywordFun:
+		case keywordFun, keywordView:
```
turbolent, line 97: Same as above. Add a separate case for the `view` keyword and ensure the `fun` keyword follows.
Like above, add a purity parameter to `parseFunctionDeclarationOrFunctionExpressionStatement`

runtime/parser/statement.go
```diff
@@ -155,6 +155,12 @@ func parseFunctionDeclarationOrFunctionExpressionStatement(p *parser) (ast.State
 
 	startPos := p.current.StartPos
 
+	purity := parsePurityAnnotation(p)
+
+	if secondPurity := parsePurityAnnotation(p); secondPurity != ast.UnspecifiedPurity {
+		return nil, p.syntaxError("invalid second purity modifier")
+	}
```
turbolent, line 162: This shouldn't be needed, no?

runtime/sema/check_composite_declaration.go
```diff
@@ -1247,6 +1266,11 @@ func (checker *Checker) memberSatisfied(compositeMember, interfaceMember *Member
 				return false
 			}
 
+			// Functions are covariant in their purity
+			if compositeMemberFunctionType.Purity != interfaceMemberFunctionType.Purity && compositeMemberFunctionType.Purity != ViewFunction {
```
turbolent, line 1270: 👍

runtime/sema/check_composite_declaration.go
```diff
@@ -1902,6 +1931,19 @@ func (checker *Checker) checkSpecialFunction(
...
+	fnType := checker.Elaboration.FunctionDeclarationFunctionTypes[function]
+	member, present := selfType.MemberMap().Get(function.Identifier.Identifier)
+	if present {
+		// members resolvers are created before the purity analysis is performed, so we update
+		// unresolved purities with the correct values
+		fnMember, ok := member.TypeAnnotation.Type.(*FunctionType)
+		if ok {
+			fnMember.Purity = fnType.Purity
+		}
+	}
```
turbolent, line 1944: Fail early, e.g. `if !present { return }` (or panic?), `if !ok { return }`, ...

runtime/sema/checker.go
```diff
@@ -78,6 +79,11 @@ type ImportHandlerFunc func(checker *Checker, importedLocation common.Location,
 
 type MemberAccountAccessHandlerFunc func(checker *Checker, memberLocation common.Location) bool
 
+type PurityCheckScope struct {
+	EnforcePurity bool
```
turbolent, line 83: What does "enforce" mean? Maybe add a comment

types.go
```diff
@@ -1397,19 +1397,29 @@ func (t *ContractInterfaceType) InterfaceInitializers() [][]Parameter {
 
 // Function
 
+type FunctionPurity int
+
+const (
+	ImpureFunction = iota
+	ViewFunction   = 1
```
turbolent, line 1404: Same as for the other enum:

```suggestion
	FunctionPurityUnspecified FunctionPurity = iota
	FunctionPurityView
```

runtime/sema/checker.go
```diff
@@ -78,6 +79,11 @@ type ImportHandlerFunc func(checker *Checker, importedLocation common.Location,
...
+	CurrentPurity bool
```
turbolent, line 84: Maybe use the enum?

```suggestion
	CurrentPurity FunctionPurity
```

runtime/parser/declaration.go
dsainati1: This is intentional: if both purity and access are specified on a function declaration, we want to use the earlier of the two to compute the start of the function declaration. That said I have a more organized way of doing this that I will update to.

encoding/json/encode.go
```diff
@@ -644,6 +645,13 @@ func prepareFields(fieldTypes []cadence.Field, results typePreparationResults) [
 	return fields
 }
 
+func preparePurity(purity cadence.FunctionPurity) string {
+	if purity == cadence.FunctionPurityView {
+		return "view"
+	}
+	return "impure"
+}
```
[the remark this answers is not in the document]
SupunS: Another option is, given that the word `impure` only occurs in json encoding, we can also choose to not to write the field if it's not `view`. I think we already treat it as optional when decoding.

runtime/parser/statement.go
```diff
@@ -151,9 +159,9 @@ func parseStatement(p *parser) (ast.Statement, error) {
 	}
 }
 
-func parseFunctionDeclarationOrFunctionExpressionStatement(p *parser) (ast.Statement, error) {
+func parseFunctionDeclarationOrFunctionExpressionStatement(p *parser, purity ast.FunctionPurity, purityPos *ast.Position) (ast.Statement, error) {
```
SupunS, line 162:
```suggestion
func parseFunctionDeclarationOrFunctionExpressionStatement(
    p *parser,
    purity ast.FunctionPurity,
    purityPos *ast.Position,
) (ast.Statement, error) {
```

runtime/sema/errors.go
```diff
@@ -3819,3 +3821,11 @@ func (e *ExternalMutationError) SecondaryError() string {
 		e.ContainerType.QualifiedString(),
 	)
 }
+
+type PurityError struct {
```
SupunS, line 3825: Make sure the error implements the correct interfaces:
```go
var _ SemanticError = &PurityError{}
var _ errors.UserError = &PurityError{}
```

It is a bit strange though, `TestErrorInterfaceConformance` is supposed to catch this and fail 🤔

SupunS (approved):
Overall looks good to me 👍

runtime/sema/type.go
```diff
@@ -192,6 +192,12 @@ type MemberResolver struct {
 	) *Member
 }
 
+// supertype of interfaces and composites
+type NominalType interface {
+	Type
+	MemberMap() *StringMemberOrderedMap
+}
```
dsainati1, line 199: Delete this
```suggestion
```

runtime/sema/type.go
```diff
@@ -192,6 +192,12 @@ type MemberResolver struct {
 	) *Member
 }
 
+// supertype of interfaces and composites
```
dsainati1, line 195:
```suggestion
```

dsainati1:
@SupunS @turbolent i just checked, heterogeneous function types always just produce `AnyStruct` for their least common supertype. If we change this in the future we'll need to account for purity, but for now it's fine.

runtime/parser/statement.go
```diff
@@ -151,9 +159,13 @@ func parseStatement(p *parser) (ast.Statement, error) {
...
 
-func parseFunctionDeclarationOrFunctionExpressionStatement(p *parser) (ast.Statement, error) {
+func parseFunctionDeclarationOrFunctionExpressionStatement(
+	p *parser,
+	purity ast.FunctionPurity,
+	purityPos *ast.Position,
+) (ast.Statement, error) {
 
-	startPos := p.current.StartPos
+	startPos := *ast.EarlierPosition(&p.current.StartPos, purityPos)
```
turbolent, line 168: 👌

runtime/parser/transaction.go
```diff
@@ -86,7 +86,7 @@ func parseTransactionDeclaration(p *parser, docString string) (*ast.TransactionD
 			identifier := p.tokenToIdentifier(p.current)
 			// Skip the `prepare` keyword
 			p.next()
-			prepare, err = parseSpecialFunctionDeclaration(p, false, ast.AccessNotSpecified, nil, identifier)
+			prepare, err = parseSpecialFunctionDeclaration(p, false, ast.AccessNotSpecified, nil, ast.FunctionPurityUnspecified, nil, identifier)
```
turbolent, line 89: Maybe chop down argument list here

runtime/sema/type.go
turbolent: I guess this was used after all?

runtime/sema/type.go
```diff
@@ -2470,6 +2494,13 @@ func formatFunctionType(
 	var builder strings.Builder
 	builder.WriteRune('(')
 
+	if len(purity) > 0 {
+		builder.WriteString(purity)
+		if spaces {
+			builder.WriteRune(' ')
```
turbolent, line 2500:
```suggestion
			builder.WriteByte(' ')
```

runtime/sema/type.go
```diff
@@ -2504,8 +2535,30 @@ func formatFunctionType(
 
 // FunctionType
 //
+type FunctionPurity int
+
+const (
+	ImpureFunction = 0
+	ViewFunction   = 1
```
turbolent, line 2542:
```suggestion
	FunctionPurityImpure FunctionPurity = iota
	FunctionPurityView
```

runtime/sema/type.go
```diff
@@ -2504,8 +2535,30 @@ func formatFunctionType(
...
+)
+
+func Purity(b bool) FunctionPurity {
```
turbolent, line 2545: Maybe improve the function name and parameter name. What is it used for?

turbolent (commented):
Looks great! Just a couple minor things

PR #1824, opened 2022-07-18
Switch to OpenTelemetry tracing library

SaveTheRbtz:
Issue: https://github.com/onflow/flow-go/issues/2796

## Description
This diff switches [deprecated](https://github.com/opentracing/specification/issues/163) OpenTracing-based library with standardized OpenTelemetry API.  This is also needed since we are migrating Flow production tracing to OTEL Collector that prefers native OTEL formats and libraries.

OpenTelemetry says the following in their docs:
> There is no hard requirement to migrate existing application code to OpenTelemetry. However, we do recommend migrating from any OpenTracing instrumentation libraries – libraries used to instrument web frameworks, HTTP clients, database clients, etc. – to their OpenTelemetry equivalents. This will improve support, as many OpenTracing libraries will be retired and may no longer be updated.

PS. While here, regenerated `go.cap` though I have no idea what i'm doing there.
PPS. This commit will break flow-go interfaces, so it needs to be coupled with a similar diff there.  (Not sure if the "Breaking change" label should be used for this or this is only for user-visible changes.)
______

### DIFF
--- a/go.cap
+++ b/go.cap
@@ -1,10 +1,8 @@
 github.com/onflow/cadence ()
 
 github.com/davecgh/go-spew/spew (file)
-github.com/onflow/cadence/runtime/interpreter ()
 github.com/onflow/cadence/runtime/parser (file)
 github.com/onflow/cadence/runtime/pretty (runtime)
-github.com/opentracing/opentracing-go (network)
 github.com/stretchr/testify/assert (runtime, file, network)
 github.com/zeebo/blake3/internal/consts (file)
 github.com/klauspost/cpuid/v2 (runtime, file)
--- a/go.mod
+++ b/go.mod
@@ -12,11 +12,11 @@ require (
 	github.com/leanovate/gopter v0.2.9
 	github.com/logrusorgru/aurora v0.0.0-20200102142835-e9ef32dff381
 	github.com/onflow/atree v0.4.0
-	github.com/opentracing/opentracing-go v1.2.0
 	github.com/rivo/uniseg v0.2.1-0.20211004051800-57c86be7915a
 	github.com/schollz/progressbar/v3 v3.8.3
 	github.com/stretchr/testify v1.7.3
 	github.com/turbolent/prettier v0.0.0-20220320183459-661cc755135d
+	go.opentelemetry.io/otel v1.8.0
 	go.uber.org/goleak v1.1.10
 	golang.org/x/crypto v0.0.0-20210921155107-089bfa567519
 	golang.org/x/text v0.3.7
--- a/go.sum
+++ b/go.sum
@@ -15,6 +15,7 @@ github.com/fxamacker/circlehash v0.3.0 h1:XKdvTtIJV9t7DDUtsf0RIpC1OcxZtPbmgIH7ek
 github.com/fxamacker/circlehash v0.3.0/go.mod h1:3aq3OfVvsWtkWMb6A1owjOQFA+TLsD5FgJflnaQwtMM=
 github.com/go-test/deep v1.0.5 h1:AKODKU3pDH1RzZzm6YZu77YWtEAq6uh1rLIAQlay2qc=
 github.com/go-test/deep v1.0.5/go.mod h1:QV8Hv/iy04NyLBxAdO9njL0iVPN1S4d/A3NVv1V36o8=
+github.com/google/go-cmp v0.5.8 h1:e6P7q2lk1O+qJJb4BtCQXlK8vWEO8V1ZeuEdJNOqZyg=
 github.com/k0kubun/go-ansi v0.0.0-20180517002512-3bf9e2903213/go.mod h1:vNUNkEQ1e29fT/6vq2aBdFsgNPmy8qMdSay1npru+Sw=
 github.com/klauspost/cpuid/v2 v2.0.12/go.mod h1:g2LTdtYhdyuGPqyWyv7qRAmj1WBqxuObKfj5c0PQa7c=
 github.com/klauspost/cpuid/v2 v2.0.14 h1:QRqdp6bb9M9S5yyKeYteXKuoKE4p0tGlra81fKOpWH8=
@@ -46,8 +47,6 @@ github.com/mitchellh/colorstring v0.0.0-20190213212951-d06e56a500db h1:62I3jR2Em
 github.com/mitchellh/colorstring v0.0.0-20190213212951-d06e56a500db/go.mod h1:l0dey0ia/Uv7NcFFVbCLtqEBQbrT4OCwCSKTEv6enCw=
 github.com/onflow/atree v0.4.0 h1:+TbNisavAkukAKhgQ4plWnvR9o5+SkwPIsi3jaeAqKs=
 github.com/onflow/atree v0.4.0/go.mod h1:7Qe1xaW0YewvouLXrugzMFUYXNoRQ8MT/UsVAWx1Ndo=
-github.com/opentracing/opentracing-go v1.2.0 h1:uEJPy/1a5RIPAJ0Ov+OIO8OxWu77jEv+1B0VhjKrZUs=
-github.com/opentracing/opentracing-go v1.2.0/go.mod h1:GxEUsuufX4nBwe+T+Wl9TAgYrxe9dPLANfrWvHYVTgc=
 github.com/pkg/term v1.1.0 h1:xIAAdCMh3QIAy+5FrE8Ad8XoDhEU4ufwbaSozViP9kk=
 github.com/pkg/term v1.1.0/go.mod h1:E25nymQcrSllhX42Ok8MRm1+hyBdHY0dCeiKZ9jpNGw=
 github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZbAQM=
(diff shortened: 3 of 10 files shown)


turbolent:
@SaveTheRbtz Could you please benchmark locally and see if this change has a significant impact?

simonhf:
> @SaveTheRbtz Could you please benchmark locally and see if this change has a significant impact?

@turbolent FYI I think there are two schools of thought about this:
1. It doesn't matter how fast or slow open tracing is because we can just tweak the sampling rate appropriately, e.g. let's say the new API is twice as slow, so we'd just halve the sampling rate from every 16th TX to every 32nd TX... no harm done?
2. Use a faster way to collect the actual timing info (the part which is speed critical), and then do all the other processing asynchronously (the part where it doesn't matter how long it takes because we have lots of spare CPU).

I did some timing of the old Jaeger instrumentation here [1] and discovered that surprisingly it's even slower than using zerolog just for the stats collection part which presumably doesn't actually send off any data to the Jaeger server.

FYI I also suggest using a different form of instrumentation which only records the minimum info -- e.g. timestamp and context -- during the time critical code, into a simple memory buffer. Then at the end of the time critical code, the recorded info can be processed asynchronously and even sent to multiple places concurrently, e.g. zlog, prometheus, and Jaeger. I show a working PoC on dummy code here [1] too. This method is the absolute fastest way to record the minimum info and in tests only spends about 50 nanoseconds per record event.

This means if we wanted to, we can record everything without sampling and it's still much faster than sampling with Jaeger. Why? In tests then just recording in the time critical code is about 50 times fasters than logging, which is already faster than Jaeger. Therefore if Jaeger is sampling every 16th TX, because this other technique is 50 times faster, it's possible to sample all TXs and still be faster. This also means that we can add almost arbitrary extra timing instrumentation and still sample all TXs and still be faster...

[1] https://www.notion.so/dapperlabs/Examine-instrumentation-timing-noodle-on-best-design-9abf3b0ce0c34b7fac6e4dc0c4f67603#e190789d2e33497aaa395f34da11712b

SaveTheRbtz:
@turbolent on the cadence side perf impact is negligible since it only operates on (already escaping) data structures.  Maybe even slightly faster on the newer version of the code given that it does not have a separate `Time.Now()` call.   That said, let me run cadence specific tests just in case.

<details>
It is quite hard to get a clean run on my laptop and even on a linux server (likely due to Turbo boost), but at least it shows that allocations did not change.
  ```
name                                              old time/op    new time/op    delta
pkg:github.com/onflow/cadence/runtime goos:darwin goarch:arm64
RuntimeFungibleTokenTransfer-8                       348µs ± 2%     348µs ± 4%    ~     (p=0.541 n=8+9)
RuntimeResourceDictionaryValues-8                   2.35ms ± 3%    2.37ms ± 5%    ~     (p=0.631 n=10+10)
pkg:github.com/onflow/cadence/runtime/interpreter goos:darwin goarch:arm64
ValueIsSubtypeOfSemaType-8                          41.6ns ± 1%    41.5ns ± 2%    ~     (p=0.305 n=9+10)
pkg:github.com/onflow/cadence/runtime/parser goos:darwin goarch:arm64
ParseDeploy/byte_array-8                            6.84ms ± 1%    6.90ms ± 1%  +0.95%  (p=0.000 n=10+9)
ParseDeploy/decode_hex-8                             597µs ± 2%     600µs ± 1%    ~     (p=0.113 n=10+9)
ParseInfix-8                                        4.00µs ± 1%    4.01µs ± 1%    ~     (p=0.085 n=10+10)
ParseFungibleToken/Without_memory_metering-8        89.7µs ± 1%    90.3µs ± 1%    ~     (p=0.065 n=10+9)
ParseFungibleToken/With_memory_metering-8            109µs ± 2%     114µs ± 3%  +3.95%  (p=0.000 n=10+10)
ParseArray-8                                        4.62ms ± 1%    4.73ms ± 2%  +2.43%  (p=0.000 n=9+10)
pkg:github.com/onflow/cadence/runtime/sema goos:darwin goarch:arm64
QualifiedIdentifierCreation/One_level-8             2.06ns ± 1%    2.04ns ± 1%  -0.81%  (p=0.001 n=9+9)
QualifiedIdentifierCreation/Three_levels-8          63.2ns ± 2%    62.3ns ± 0%  -1.49%  (p=0.000 n=9+9)
pkg:github.com/onflow/cadence/runtime/tests/checker goos:darwin goarch:arm64
ContractInterfaceFungibleToken-8                    21.0µs ± 1%    20.9µs ± 0%  -0.78%  (p=0.001 n=9+10)
CheckContractInterfaceFungibleTokenConformance-8    65.7µs ± 1%    64.8µs ± 3%  -1.49%  (p=0.019 n=9+9)
pkg:github.com/onflow/cadence/runtime/tests/interpreter goos:darwin goarch:arm64
NewInterpreter/new_interpreter-8                     524ns ± 2%     521ns ± 1%    ~     (p=0.055 n=10+10)
NewInterpreter/new_sub-interpreter-8                1.15µs ± 3%    1.13µs ± 1%  -1.44%  (p=0.027 n=10+10)
InterpretRecursionFib-8                             1.64ms ± 3%    1.61ms ± 2%  -2.35%  (p=0.006 n=10+9)

name                                              old alloc/op   new alloc/op   delta
pkg:github.com/onflow/cadence/runtime goos:darwin goarch:arm64
RuntimeFungibleTokenTransfer-8                       233kB ± 1%     233kB ± 1%    ~     (p=0.549 n=10+9)
RuntimeResourceDictionaryValues-8                   2.23MB ± 0%    2.23MB ± 0%    ~     (p=0.497 n=10+9)
pkg:github.com/onflow/cadence/runtime/interpreter goos:darwin goarch:arm64
ValueIsSubtypeOfSemaType-8                           48.0B ± 0%     48.0B ± 0%    ~     (all equal)
pkg:github.com/onflow/cadence/runtime/parser goos:darwin goarch:arm64
ParseDeploy/byte_array-8                            4.38MB ± 3%    4.38MB ± 1%    ~     (p=0.853 n=10+10)
ParseDeploy/decode_hex-8                             214kB ± 0%     214kB ± 0%    ~     (p=0.438 n=10+10)
ParseInfix-8                                        2.17kB ± 0%    2.17kB ± 0%    ~     (p=1.000 n=10+10)
ParseFungibleToken/Without_memory_metering-8        36.3kB ± 0%    36.3kB ± 0%    ~     (p=0.641 n=10+10)
ParseFungibleToken/With_memory_metering-8           36.3kB ± 0%    36.3kB ± 0%    ~     (p=0.126 n=8+10)
ParseArray-8                                        2.78MB ± 1%    2.78MB ± 1%    ~     (p=0.853 n=10+10)
pkg:github.com/onflow/cadence/runtime/sema goos:darwin goarch:arm64
QualifiedIdentifierCreation/One_level-8              0.00B          0.00B         ~     (all equal)
QualifiedIdentifierCreation/Three_levels-8           64.0B ± 0%     64.0B ± 0%    ~     (all equal)
pkg:github.com/onflow/cadence/runtime/tests/checker goos:darwin goarch:arm64
ContractInterfaceFungibleToken-8                    25.1kB ± 0%    25.1kB ± 0%  -0.00%  (p=0.003 n=10+8)
CheckContractInterfaceFungibleTokenConformance-8    63.6kB ± 0%    63.6kB ± 0%  -0.00%  (p=0.029 n=10+7)
pkg:github.com/onflow/cadence/runtime/tests/interpreter goos:darwin goarch:arm64
NewInterpreter/new_interpreter-8                      928B ± 0%      928B ± 0%    ~     (all equal)
NewInterpreter/new_sub-interpreter-8                1.46kB ± 0%    1.46kB ± 0%    ~     (all equal)
InterpretRecursionFib-8                             1.51MB ± 0%    1.51MB ± 0%    ~     (p=0.598 n=10+10)

name                                              old allocs/op  new allocs/op  delta
pkg:github.com/onflow/cadence/runtime goos:darwin goarch:arm64
RuntimeFungibleTokenTransfer-8                       4.16k ± 0%     4.16k ± 0%    ~     (all equal)
RuntimeResourceDictionaryValues-8                    37.4k ± 0%     37.4k ± 0%    ~     (all equal)
pkg:github.com/onflow/cadence/runtime/interpreter goos:darwin goarch:arm64
ValueIsSubtypeOfSemaType-8                            1.00 ± 0%      1.00 ± 0%    ~     (all equal)
pkg:github.com/onflow/cadence/runtime/parser goos:darwin goarch:arm64
ParseDeploy/byte_array-8                              105k ± 0%      105k ± 0%    ~     (p=0.671 n=10+10)
ParseDeploy/decode_hex-8                              77.0 ± 0%      77.0 ± 0%    ~     (all equal)
ParseInfix-8                                          66.0 ± 0%      66.0 ± 0%    ~     (all equal)
ParseFungibleToken/Without_memory_metering-8         1.06k ± 0%     1.06k ± 0%    ~     (all equal)
ParseFungibleToken/With_memory_metering-8            1.06k ± 0%     1.06k ± 0%    ~     (all equal)
ParseArray-8                                         70.0k ± 0%     70.0k ± 0%    ~     (p=0.838 n=10+10)
pkg:github.com/onflow/cadence/runtime/sema goos:darwin goarch:arm64
QualifiedIdentifierCreation/One_level-8               0.00           0.00         ~     (all equal)
QualifiedIdentifierCreation/Three_levels-8            2.00 ± 0%      2.00 ± 0%    ~     (all equal)
pkg:github.com/onflow/cadence/runtime/tests/checker goos:darwin goarch:arm64
ContractInterfaceFungibleToken-8                       441 ± 0%       441 ± 0%    ~     (all equal)
CheckContractInterfaceFungibleTokenConformance-8     1.05k ± 0%     1.05k ± 0%    ~     (all equal)
pkg:github.com/onflow/cadence/runtime/tests/interpreter goos:darwin goarch:arm64
NewInterpreter/new_interpreter-8                      13.0 ± 0%      13.0 ± 0%    ~     (all equal)
NewInterpreter/new_sub-interpreter-8                  42.0 ± 0%      42.0 ± 0%    ~     (all equal)
InterpretRecursionFib-8                              33.5k ± 0%     33.5k ± 0%    ~     (all equal)

  ```
  
</details>

PS. Flow tracing microbenchmarks are here: https://github.com/onflow/flow-go/pull/2806 (i'm missing a second reviewer there =))
PPS. I'm of the first school of thought: IMHO if we get into the nano-second territory we should just use pprof (incl. trace).  For everything else we can always tweak sampling probability.  (Tests are there to check that we do not block and maintain a sane amount of allocations.)

turbolent:
@simonhf @SaveTheRbtz Thank you for the info. Looks good!

go.cap
```diff
@@ -1,10 +1,8 @@
 github.com/onflow/cadence ()
 
 github.com/davecgh/go-spew/spew (file)
-github.com/onflow/cadence/runtime/interpreter ()
```
ramtinms, line 4: is this expected ?

ramtinms (approved):
looks great to me, thanks for using attributes instead of events (aka logs) here, makes more sense

turbolent (approved):

go.cap
turbolent: I guess no entry is equivalent to `()`? @fxamacker do you maybe know?

Issue #1823, opened 2022-07-18
Avoid re-creating interpreter and checker configuration


turbolent:
### Issue To Be Solved

The checker and interpreter are currently configured through options – functions which accept the checker/interpreter and modify it. This is a common pattern in Go code, but can be a performance issue: Checkers and interpreters are created often, even during the lifetime of a transaction/script execution, as there is one checker and interpreter for each location.

In addition, most of the configuration is always the same for each environment (script or transaction), so the result of the configuration is mostly the same each time it is performed.

### Suggested Solution

Refactor the configuration part out of the checker and interpreter into a reusable value. This should speed up checker/interpreter initialization and reduce allocations.

PR #1825, opened 2022-07-19
Optimize predeclared values, refactor runtime into reusable part

### LINKED ISSUE #1822: Avoid re-declaring predeclared values
### Issue To Be Solved

The checker and interpreter accept an option to inject type and value declarations. For example, this allows defining standard library functions like `assert` and `panic`.

This is currently implemented through checker and interpreter options which declare "predeclared values" in the activation.

This works, but is inefficient: For each new checker and interpreter instance, the list of predeclared values is iterated, and a variable is allocated and defined in the activation. As there is one checker / interpreter per location, this has a significant cost. It is one of the major costs of creating a new checker / interpreter.

In addition, for a certain environment, script or transaction execution, the set of declarations is always the same. For example, all transactions and scripts have the `assert`, `panic`, `getAccount`, etc functions.

### Suggested Solution

Refactor the way predeclared values are declared for the checker and interpreter: Create a new activation, declare the predeclared values, then allow it to be passed in as an option, which links it into the variable activations stack.

This allows the activation and associated variables to be created once, then reused for multiple checker / interpreters.

This optimization should improve both performance and reduce allocations.

turbolent:
Closes #1822

## Description

- Optimize checker/interpreter predeclared values by refactoring redeclaration of variables in activation for each checker/interpreter instantiation to a single activation creation which can be injected and thus reused
-  Moved the majority of the standard library implementation from `runtime` to the `stdlib` package. Instead of requiring the full `runtime.Interface`, each value declaration depends on a separate interface. This allows standard library declarations to be declared once/reused, by making the target (`runtime.Interface`) switchable. It also ensures the declarations are implemented consistently and allows them to be tested individually.
- Refactored the majority of the reusable part of the runtime into an "environment". The environment can be reused across multiple transactions/scripts. Only the interface and storage is re-targeted.
- This change unlocks #1823: In a follow-up PR, the checker and interpreter options will get refactored into a reused configuration.

This is a large refactor. It is best reviewed commit by commit. We should have a review session to go over it together.

______


runtime/cmd/cmd.go
```diff
@@ -74,18 +74,14 @@ var checkers = map[common.Location]*sema.Checker{}
...
 	codes map[common.Location]string,
-	impls stdlib.FlowBuiltinImpls,
 ) (
 	[]sema.Option,
 	[]interpreter.Option,
 ) {
 
-	semaPredeclaredValues, interpreterPredeclaredValues :=
-		stdlib.FlowDefaultPredeclaredValues(impls)
+	// TODO: checker and interpreter base activations
```
turbolent, line 82: In a follow up PR it would be nice to refactor the command line runner and REPL to use the environment and avoid duplicating checker and interpreter creation/configuration.

runtime/sema/checker.go
```diff
@@ -122,34 +122,16 @@ type Checker struct {
...
-			variable := checker.declareValue(declaration)
-			if variable == nil {
-				continue
-			}
-			checker.Elaboration.GlobalValues.Set(variable.Identifier, variable)
-			checker.Elaboration.EffectivePredeclaredValues[variable.Identifier] = declaration
-		}
-
+		checker.baseValueActivation = baseValueActivation
 		return nil
```
turbolent, line 128: This is one of the main optimizations: Instead of re-declaring the predeclared values (and types below; same for the predeclared values in the interpreter), a fully prepared activation is linked into the variable activations stack

tools/analysis/programs.go
```diff
@@ -87,16 +87,11 @@ func (programs Programs) check(
 	error,
 ) {
 
-	semaPredeclaredValues, _ :=
-		stdlib.FlowDefaultPredeclaredValues(stdlib.FlowBuiltinImpls{})
```
turbolent, line 91: Same here: In a follow-up PR, we can maybe use the new environment code instead

runtime/account_test.go
```diff
@@ -1866,7 +1868,7 @@ func TestGetAuthAccount(t *testing.T) {
 			},
 		}
 
-		_, err := rt.ExecuteScript(
+		err := rt.ExecuteTransaction(
```
turbolent, line 1871: Actually run a transaction, not a script with a transaction location

runtime/config.go
```diff
@@ -0,0 +1,39 @@
...
+
+package runtime
+
+import (
+	"github.com/onflow/cadence/runtime/interpreter"
+)
+
+// Config is a constant/read-only configuration of an environment.
+//
+type Config struct {
```
turbolent, line 27: Refactor the "option function" based configuration to a simple struct

runtime/contract_update_test.go
```diff
@@ -95,8 +95,6 @@ func TestContractUpdateWithDependencies(t *testing.T) {
 		},
 	}
 
-	runtime.SetContractUpdateValidationEnabled(true)
```
turbolent, line 98: Contract update validation should not be an option, it should always be enabled

runtime/environment.go
```diff
@@ -0,0 +1,1147 @@
...
+}
+
+func (e *interpreterEnvironment) newInterpreter(
+	location common.Location,
+	program *interpreter.Program,
+) (*interpreter.Interpreter, error) {
+
+	publicKeyValidationHandler := e.newPublicKeyValidationHandler()
+
+	options := []interpreter.Option{
```
turbolent, line 508: In a follow-up PR we can now make this a configuration struct that can be reused for all interpreters: All options only depend on the environment.

runtime/predeclaredvalues_test.go
```diff
@@ -19,130 +19,111 @@
...
+	"github.com/onflow/cadence"
 	"github.com/onflow/cadence/runtime/common"
+	"github.com/onflow/cadence/runtime/interpreter"
 	"github.com/onflow/cadence/runtime/sema"
-	"github.com/onflow/cadence/runtime/tests/checker"
+	"github.com/onflow/cadence/runtime/stdlib"
+	"github.com/onflow/cadence/runtime/tests/utils"
 )
 
 func TestRuntimePredeclaredValues(t *testing.T) {
```
turbolent, line 35: The option to selectively declare values based on location is gone. It was not used and it is a reasonable trade-off for the performance gain

runtime/stdlib/account.go
```diff
@@ -0,0 +1,1718 @@
...
+}
+
+type AccountCreator interface {
+	EventEmitter
+	AuthAccountHandler
+	// CreateAccount creates a new account.
+	CreateAccount(payer common.Address) (address common.Address, err error)
+}
+
+func NewAuthAccountConstructor(creator AccountCreator) StandardLibraryValue {
```
turbolent, line 78: Each standard library value declaration now only depends on a dedicated interface. This allows the declarations to be used independent of the runtime package, and makes them individually testable.

The environment implements these interfaces and forwards the calls to the current `runtime.Interface` of the environment (the environment is reused across multiple script and transaction invocations, but its interface is swapped)

runtime/sema/checker.go
SupunS: nice! That makes total sense

runtime/stdlib/account.go
SupunS: I assume the code was just 'moved' and updated to rely on the new interface, with no change to the logic?

runtime/stdlib/contract_update_validation.go
```diff
@@ -16,19 +16,21 @@
  * limitations under the License.
  */
 
-package runtime
+package stdlib
```
SupunS, line 19: Not sure if stdlib is the correct place for this

runtime/stdlib/crypto.go
```diff
@@ -43,8 +43,7 @@ var CryptoChecker = func() *sema.Checker {
 		location,
 		nil,
 		false,
-		sema.WithPredeclaredValues(BuiltinFunctions.ToSemaValueDeclarations()),
-		sema.WithPredeclaredTypes(BuiltinTypes.ToTypeDeclarations()),
+		// TODO: base values and types
```
SupunS, line 46: How come the type checking for Crypto contract works without the built-in types? (e.g: `PublicKey`, etc.)?

runtime/stdlib/hashalgorithm.go
```diff
@@ -25,100 +25,97 @@ import (
...
-			inter,
-			getLocationRange,
-			dataValue,
-			nil,
-			hashAlgoValue,
-		)
-	},
-	sema.HashAlgorithmTypeHashFunctionType,
-)
+func hashAlgorithmHashFunction(hashAlgoValue interpreter.MemberAccessibleValue) *interpreter.HostFunctionValue {
```
SupunS, line 54: previously this function was only defined once, but now it creates a one for every enum-case of the hash algorithm. Is it no longer possible to keep the old behaviour (optimized)?

SupunS (commented):
This is a great optimization! Numbers on ft-transfer benchmark looks very promising!

runtime/stdlib/account.go
turbolent: Yes, I only updated the "external calls", the code was mostly just moved over. We can maybe go over it side-by-side in a review session

runtime/convertValues.go
```diff
@@ -1570,14 +1569,14 @@ func importHashAlgorithm(
 		)
 	}
 
-	return stdlib.NewHashAlgorithmCase(inter, uint8(rawValue)), nil
+	// TODO: return existing
```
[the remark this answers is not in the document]
turbolent: Implemented in #1888

runtime/environment.go
```diff
@@ -0,0 +1,1147 @@
...
+	)
+
+	return interpreter.NewInterpreter(
+		program,
+		location,
+		options...,
+	)
+}
+
+func (e *interpreterEnvironment) onStatementHandler() interpreter.OnStatementFunc {
```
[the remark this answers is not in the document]
turbolent: Improved in 16f51686198cce2cab82c33bd8a9973d6d5706b8

runtime/interpreter/interpreter.go
```diff
@@ -363,6 +360,7 @@ type Interpreter struct {
 	resourceVariables                    map[ResourceKindedValue]*Variable
 	memoryGauge                          common.MemoryGauge
 	CallStack                            *CallStack
+	baseActivation                       *VariableActivation
```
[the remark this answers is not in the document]
turbolent: The idea is that instead of hard-coding to a default (`BaseActivation`), it is now possible to provide a custom one

runtime/interpreter/interpreter.go
```diff
@@ -1850,7 +1833,10 @@ func (interpreter *Interpreter) declareEnumConstructor(
 	intType := sema.IntType
 
 	enumCases := declaration.Members.EnumCases()
-	caseValues := make([]*CompositeValue, len(enumCases))
+	caseValues := make([]struct {
+		Value    MemberAccessibleValue
+		RawValue IntegerValue
+	}, len(enumCases))
```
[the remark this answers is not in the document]
turbolent: Good idea, added in [9f50172](https://github.com/onflow/cadence/pull/1825/commits/9f50172b9c2742ef91b9317287f15a043b8af5d4)

runtime/environment.go
```diff
@@ -0,0 +1,1147 @@
...
+	"github.com/onflow/cadence/runtime/ast"
+	"github.com/onflow/cadence/runtime/common"
+	"github.com/onflow/cadence/runtime/errors"
+	"github.com/onflow/cadence/runtime/interpreter"
+	"github.com/onflow/cadence/runtime/parser"
+	"github.com/onflow/cadence/runtime/sema"
+	"github.com/onflow/cadence/runtime/stdlib"
+)
+
+type Environment interface {
```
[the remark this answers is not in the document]
turbolent: Correct. Basically, it is the reusable parts, extracted from the runtime. The idea is that the environment can be prepared once for e.g all transactions, then re-used by just re-configuring (e.g. runtime interface and storage for a particular transaction).

I tried to find a name for this, but struggled. In hindsight, I don't even really like the way we now have runtime and environment – maybe environment should be merged back into runtime, and made re-configurable?

runtime/runtime.go
```diff
@@ -441,145 +304,47 @@ func scriptExecutionFunction(
...
-
-	storage := NewStorage(context.Interface, memoryGauge)
-
-	var interpreterOptions []interpreter.Option
-	var checkerOptions []sema.Option
+	storage := NewStorage(context.Interface, context.Interface)
 
-	functions := r.standardLibraryFunctions(
-		context,
+	environment := context.Environment
```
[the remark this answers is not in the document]
turbolent: Correct. The idea is that a user of the runtime *may* provide an environment through the context, which allows re-using it across multiple transactions. That's the core idea of this whole optimization: prepare what is needed for checking and interpretation once, e.g. activations, storage, etc., then just reconfigure it. 

The user does not *have to* provide an environment in the context, in which case a new environment is created (basically the current behaviour)

runtime/stdlib/contract_update_validation.go
turbolent: Yeah, it isn't really "standard library" code, but it is needed by the standard library's account contract deployment/update code, and I could not think of another place. Do you have any ideas where it could live instead?

runtime/stdlib/crypto.go
turbolent: Turns out the Crypto contract does not actually use any of the stdlib values. It does indeed use the `PublicKey` type – but not the value (constructor)! I'll remove this TODO.

runtime/stdlib/hashalgorithm.go
turbolent: That was actually done in #1813. I'll open a follow-up PR

turbolent: Looked into it. The problem is that #1813 refactored the enum cases from `CompositeValue`s to `SimpleCompositeValue`. For `CompositeValue`, `GetMember` returns a `BoundFunctionValue`, which sets `Invocation.Self`. But `SimpleCompositeValue` does not do this, as it internally makes no distinction between fields and functions.

We can either:
- Keep the current code as-is. The overhead of having a separate function per case should be negligible; or
- Distinguish between fields and functions in `SimpleCompositeValue` and return a `BoundFunctionValue` for functions

Thoughts?

turbolent:
CI is failing for a potentially flaky LS test. Pinged @sideninja

sideninja:
@turbolent I can't seem to reproduce that flakiness locally. I've created a PR I think might help fix the test issue, but I also haven't detected that during the CI, so making these changes based on the stack from this CI: https://github.com/onflow/cadence/pull/1890

turbolent:
@sideninja Thank you for investigating. 🙏 I'll re-run the job and see if it shows up again

sideninja:
> @sideninja Thank you for investigating. 🙏 I'll re-run the job and see if it shows up again

Now it's something else?

runtime/sema/check_import_declaration.go
```diff
@@ -149,7 +149,6 @@ func (checker *Checker) importResolvedLocation(resolvedLocation ResolvedLocation
 		checker.valueActivations,
 		resolvedLocation.Identifiers,
 		allValueElements,
-		imp.IsImportableValue,
```
dsainati1, line 152: As discussed in the meeting; would be good to double check with a test that it is indeed safe to remove the check that these values are importable.

dsainati1 (approved):
Reviewed this as a group in a meeting; this looks good to me! Very exciting!

turbolent:
@sideninja it's happening again and I can reproduce it locally, even on master. It only happens 1 out of ~5000 calls ... looked into it and can't see how this could ever produce anything else, though testify's mock library uses quite some low-level hacks, hmm

runtime/interface.go
```diff
@@ -79,11 +80,11 @@ type Interface interface {
 	// RevokeEncodedAccountKey removes a key from an account by index, add returns the encoded key.
 	RevokeEncodedAccountKey(address Address, index int) (publicKey []byte, err error)
 	// AddAccountKey appends a key to an account.
-	AddAccountKey(address Address, publicKey *PublicKey, hashAlgo HashAlgorithm, weight int) (*AccountKey, error)
+	AddAccountKey(address Address, publicKey *stdlib.PublicKey, hashAlgo HashAlgorithm, weight int) (*stdlib.AccountKey, error)
```
SupunS, line 83: Do we want to expose `stdlib.*` stuff via the interface? or maybe define an alias in the runtime package (similar to `Address`, `HashAlgorithm`, etc.)

SupunS (approved):
Looks good! 🎉

turbolent:
@dsainati1 I had another look to re-familiarize myself with that part: The check is not necessary anymore, because `GlobalValues` and `GlobalTypes` is not set anymore.

https://github.com/onflow/cadence/pull/1825/files#diff-74d0954852bf407e9259a7c9bb55e8af093f080f00f5e35689c250886c29860aL134-L135

runtime/interface.go
turbolent: Great idea! Added type aliases in d264d5fbf588fa843bf9f9588acc4c93a6523a42

runtime/stdlib/hashalgorithm.go
SupunS: I still need to go through the code again to fully understand what this means / and the impact.

Meantime this is certainly not a blocker for merging the PR 👍

sideninja:
> 

Can we try merging the PR I made in? I think it could be related to the fact I'm returning nils. https://github.com/onflow/cadence/pull/1890

Issue #331, opened 2020-08-20
Cadence testing framework


turbolent:
### Issue To Be Solved

It would be nice to write tests for Cadence in Cadence.

### Suggest A Solution

Once the [Go testing library](https://github.com/onflow/cadence-testing-go) is ready, it could be exposed to Cadence

turbolent:
Here's a first proposal for the framework:


# Cadence Testing Framework

## Assertions/Expectations/Matchers

- `fun assert(_ condition: Bool, message: String)`:
  - Accepts boolean expression
  - Simple implementation: Abort test instead of aborting whole program
  - Improved implementation: Could analyze expression, like [Scala Test's assert macro](https://www.scalatest.org/user_guide/using_assertions)
    - Translate expression to a matcher

- `fun fail(message: string)`:
  - Aborts the current test
  - Message argument is optional

- `fun expect(_ value: AnyStruct): Expectation`:
  - Returns an expectation for the given value
  - Open question: how could we support resources, e.g. `expect(optionalResource).to(not(beNil()))`

- Expectation
  - An expectation is an object that allows a value to be tested against a matcher

    ```cadence
    pub struct Expectation {

        let value: AnyStruct

        pub init(value: AnyStruct) {
            self.value = value
        }

        pub fun to(_ matcher: Matcher) {
            if !matcher.test(self.value) {
                fail()
            }
        }
    }
    ```

- Matcher
  - A matcher is an object that consists of a test function and associated utility functionality

    ```cadence
    pub struct Matcher {

        pub let test: ((value: AnyStruct): Bool)

        pub init(test: ((value: AnyStruct): Bool)) {
            self.test = test
        }

        /// Combine this matcher with the given matcher.
        /// Returns a new matcher that succeeds if this and the given matcher succeed

        pub fun and(_ other: Matcher): Matcher {
            return Matcher(test: fun (value: AnyStruct): Bool {
                return self.test(value) && other.test(value)
            })
        }

        /// Combine this matcher with the given matcher.
        /// Returns a new matcher that succeeds if this and the given matcher succeed

        pub fun or(_ other: Matcher): Matcher {
            return Matcher(test: fun (value: AnyStruct): Bool {
                return self.test(value) || other.test(value)
            })
        }
    }
    ```

  - Built-in matcher functions:
    - `fun equal(_ value): Matcher`:
      returns a matcher that succeeds if the tested value is equal to the given value
    - `fun haveElementCount(_ count: Int): Matcher`:
[...]

PR #1827, opened 2022-07-19
[Test Framework] Start development of Cadence test framework

SupunS:

## Description

This PR lays the foundation for the test framework.

For now, this currently provides the functionality of:
 - Running a single test method
 - Running all test methods in a script.

    e.g: _tests.cdc_
    ```cadence
    pub fun testFunc1() {
        assert(false)
    }
    pub fun testFunc2() {
        assert(true)
    }
    ```
    Above cadence test-script (_tests.cdc_) can be run using the provided Go-utility method (can be used in dev-tools):
    ```go
    results := RunTests(code)
    ```

 - Starting a blockchain and running a cadence script against it.
   ``` cadence
    import Test

    pub fun testExecutingScript() {

        var blockchain = Test.newEmulatorBlockchain()
        var result = blockchain.executeScript("pub fun main(): Int {  return 2 + 3 }")

        assert(result.status == Test.ResultStatus.succeeded)
        assert((result.returnValue! as! Int) == 5)

        log(result.returnValue)
    } 
   ```

### NOTE:
The execution of the test script itself (i.e: test runner) currently uses a local environment with limited functionality. Need to improve this by using fvm/emulator as the environment for the runner.

______


runtime/stdlib/test.go
```diff
@@ -0,0 +1,471 @@
...
+		if len(invocation.Arguments) > 1 {
+			messageValue, ok := invocation.Arguments[1].(*interpreter.StringValue)
+			if !ok {
+				panic(errors.NewUnreachableError())
+			}
+			message = messageValue.Str
+		}
+
+		if !condition {
+			panic(AssertionError{
```
robert-e-davidson3, line 219: This is out of scope for developing the test framework but do you plan for the framework to eventually support tests that continue after failure, collecting the errors along the way? I am thinking of Go's `assert` vs `require` for tests.

runtime/stdlib/test.go
```diff
@@ -0,0 +1,471 @@
...
+
+	succeededVar := resultStatusConstructor.NestedVariables[succeededCaseName]
+	succeeded := succeededVar.GetValue()
+
+	// Create a 'ScriptResult' by calling its constructor.
+
+	scriptResultConstructorVar := inter.Activations.Find(scriptResultTypeName)
+	scriptResultConstructor, ok := scriptResultConstructorVar.GetValue().(*interpreter.HostFunctionValue)
+	if !ok {
+		panic(errors.NewUnexpectedError("invalid type for constructor"))
```
robert-e-davidson3, line 436: Maybe include the invalid type in the error message?
(here and elsewhere)

test-framework/emulator_backend.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func NewEmulatorBackend() *EmulatorBackend {
+	return &EmulatorBackend{
+		blockchain: newBlockchain(),
+	}
+}
+
+func (e *EmulatorBackend) RunScript(code string, args []interpreter.Value) interpreter.ScriptResult {
+	// TODO: maybe re-use interpreter? Only needed for value conversion
```
robert-e-davidson3, line 51: Value conversion uses the interpreter for its Elaboration. Is there a potential conflict from reusing it here? Like if a contract is updated between calls.

I know we do some kind of cache invalidation but I'm not sure if it applies here.

test-framework/emulator_backend.go
```diff
@@ -0,0 +1,159 @@
...
+				subInterpreter, err := inter.NewSubInterpreter(program, location)
+				if err != nil {
+					panic(err)
+				}
+				return interpreter.InterpreterImport{
+					Interpreter: subInterpreter,
+				}
+
+			default:
+				panic(errors.NewUnexpectedError("importing programs not implemented"))
```
robert-e-davidson3, line 155: This is a little less ambiguous:
```suggestion
				panic(errors.NewUnexpectedError("importing of programs not implemented"))
```

test-framework/test_framework.go
```diff
@@ -0,0 +1,226 @@
...
+		panic(err)
+	}
+
+	err = checker.Check()
+	if err != nil {
+		panic(err)
+	}
+
+	// TODO: validate test function signature
+	//   e.g: no return values, no arguments, etc.
```
robert-e-davidson3, line 88: Maybe it would later be useful to accept a string argument as the test's name/identifier.

test-framework/test_framework.go
```diff
@@ -0,0 +1,226 @@
...
+		panic(err)
+	}
+
+	return program, inter
+}
+
+func newInterpreterFromChecker(checker *sema.Checker) (*interpreter.Interpreter, error) {
+	predeclaredInterpreterValues := stdlib.BuiltinFunctions.ToInterpreterValueDeclarations()
+	predeclaredInterpreterValues = append(predeclaredInterpreterValues, stdlib.BuiltinValues.ToInterpreterValueDeclarations()...)
+	predeclaredInterpreterValues = append(predeclaredInterpreterValues, stdlib.HelperFunctions.ToInterpreterValueDeclarations()...)
```
robert-e-davidson3, line 106: Here's a utility that's helped for my encoding tests. It's overkill for a one-off like this but if this pattern is common outside of the encoding tests then it may be worth using:

```
func Concat[T any](deep ...[]T) []T {
	length := 0
	for _, b := range deep {
		length += len(b)
	}

	flat := make([]T, 0, length)
	for _, b := range deep {
		flat = append(flat, b...)
	}

	return flat
}
```

I went with iterating twice but allocating memory once. This probably isn't optimal because I didn't measure the tradeoff.

test-framework/test_framework.go
```diff
@@ -0,0 +1,226 @@
...
+		interpreter.WithStorage(interpreter.NewInMemoryStorage(nil)),
+		interpreter.WithTestFramework(NewEmulatorBackend()),
+		interpreter.WithPredeclaredValues(predeclaredInterpreterValues),
+		interpreter.WithImportLocationHandler(func(inter *interpreter.Interpreter, location common.Location) interpreter.Import {
+			switch location {
+			case stdlib.CryptoChecker.Location:
+				program := interpreter.ProgramFromChecker(stdlib.CryptoChecker)
+				subInterpreter, err := inter.NewSubInterpreter(program, location)
+				if err != nil {
+					panic(err)
```
robert-e-davidson3, line 120: Should this be wrapped in `errors.NewUnexpectedError`?

test-framework/test_framework.go
```diff
@@ -0,0 +1,226 @@
...
+				}
+				return interpreter.InterpreterImport{
+					Interpreter: subInterpreter,
+				}
+
+			default:
+				panic(errors.NewUnexpectedError("importing programs not implemented"))
```
robert-e-davidson3, line 137:
```suggestion
				panic(errors.NewUnexpectedError("importing of programs not implemented"))
```

test-framework/test_framework.go
```diff
@@ -0,0 +1,226 @@
...
+				var elaboration *sema.Elaboration
+				switch importedLocation {
+				case stdlib.CryptoChecker.Location:
+					elaboration = stdlib.CryptoChecker.Elaboration
+
+				case stdlib.TestContractLocation:
+					elaboration = stdlib.TestContractChecker.Elaboration
+
+				default:
+					return nil, errors.NewUnexpectedError("importing programs not implemented")
```
robert-e-davidson3, line 200:
```suggestion
					return nil, errors.NewUnexpectedError("importing of programs not implemented")
```

test-framework/test_framework_test.go
```diff
@@ -0,0 +1,110 @@
...
+    `
+
+	err := RunTest(code, "testFunc1")
+	assert.Error(t, err)
+
+	err = RunTest(code, "testFunc2")
+	assert.NoError(t, err)
+}
+
+func TestExecuteScript(t *testing.T) {
```
robert-e-davidson3, line 67: Needs a test for when `assert` has a second argument.

robert-e-davidson3 (approved):

runtime/stdlib/test.go
SupunS: Yea, we can iteratively add them. For now, this function is a sort of duplicate of the built-in `assert()` function and has the same behaviour. One remaining question is whether to keep both or remove one of them.

runtime/stdlib/test.go
SupunS: That's a good idea. I'm planning to open a separate PR to refactor and improve all errors/panics (for e.g: some panics have no error messages atm, etc.). I'll include this in the same 👍

test-framework/test_framework.go
SupunS: This kind of appends only happens once during initialization usually (here it's only for tests, which are local), so allocations aren't really a problem. So using append twice would be fine/simple?

test-framework/test_framework.go
SupunS: We could, but doesn't really need to. The wrapping was introduced specially for the cadence <-> fvm interface, so the fvm can 'fail transaction' vs 'stop the network' depending on the error type. In tests, we don't really have that requirement. All errors are recovered at the entry point and the test is failed.

test-framework/go.mod
```diff
@@ -0,0 +1,96 @@
+module github.com/onflow/cadence/test
+
+go 1.18
+
+require (
+	github.com/onflow/cadence v0.24.6
+	github.com/onflow/flow-emulator v0.33.4-0.20220719174752-88332a9af217
```
dsainati1, line 7: This makes me wary of putting another distinct module in the Cadence repo along with Cadence itself and the Language Server. Have we given any thought to putting this in a different repo? We could move this to the same repo as the linter, for example, and maybe rename it from `cadence-lint` to `cadence-tooling`?

SupunS: Yea, this can go to a separate repo eventually. Kept it inside cadence for now, because it makes it easier during the initial phase (i.e: needs a lot of coordinated changes between the test-stdlib and this module).

Agree that all tooling should go to a separate repo/repos 👍

runtime/stdlib/test.go
```diff
@@ -0,0 +1,472 @@
...
+
+	return result, nil
+}
+
+// createScriptResult Creates a "ScriptResult" using the return value of the executed script.
+//
+func createScriptResult(inter *interpreter.Interpreter, returnValue interpreter.Value) interpreter.Value {
+	// Lookup and get 'ResultStatus.succeeded' value.
+
+	resultStatusConstructorVar := inter.Activations.Find(resultStatusTypeName)
```
turbolent, line 423: Maybe just export `Interpreter.findVariable` instead of exporting `Interpreter.activations`

runtime/stdlib/test.go
```diff
@@ -0,0 +1,472 @@
...
+
+	succeededVar := resultStatusConstructor.NestedVariables[succeededCaseName]
+	succeeded := succeededVar.GetValue()
+
+	// Create a 'ScriptResult' by calling its constructor.
+
+	scriptResultConstructorVar := inter.Activations.Find(scriptResultTypeName)
+	scriptResultConstructor, ok := scriptResultConstructorVar.GetValue().(*interpreter.HostFunctionValue)
+	if !ok {
+		panic(errors.NewUnexpectedError("invalid type for constructor"))
```
turbolent, line 437: Same code as above for `ResultStatus`. Maybe create a helper function

runtime/stdlib/test.go
```diff
@@ -0,0 +1,472 @@
...
+		panic(errors.NewUnexpectedError(
+			"invalid type for %s. expected function",
+			emulatorBackendExecuteScriptFunctionName,
+		))
+	}
+
+	return functionType
+}()
+
+var emulatorBackendExecuteScriptFunction = interpreter.NewUnmeteredHostFunctionValue(
```
turbolent, line 364: Instead of making `Interpreter` know about a testing framework by adding a `TestFramework` field to it, just to be able to get access in this host function, maybe (if possible) make `emulatorBackendExecuteScriptFunction` a function that gets a test framework and returns a host function (`func newEmulatorBackendExecuteScriptFunction(testFramework TestFramework) *interpreter.HostFunctionValue`)

runtime/stdlib/contracts/test.cdc
```diff
@@ -0,0 +1,92 @@
+// Test contract is the standard library that provides testing functionality in Cadence.
+//
+// TODO: Name is yet to be finalized
+//
+pub contract Test {
```
turbolent, line 5: 👌

test-framework/test_framework_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+            pub fun test() {
+                var blockchain = Test.newEmulatorBlockchain()
+                var result = blockchain.executeScript(
+                    "pub fun main(a: Int, b: Int): Int {  return a + b }",
+                    [2, 3]
+                )
+
+                assert(result.status == Test.ResultStatus.succeeded)
+                assert((result.returnValue! as! Int) == 5)
```
turbolent, line 141: 😍

turbolent (commented):
Great work! 👏

runtime/stdlib/test.go
SupunS: Good idea, added with f2ad6e78c3deb87555f9d542e3776c4104072846

runtime/stdlib/test.go
SupunS: added with f2ad6e78c3deb87555f9d542e3776c4104072846

runtime/stdlib/test.go
SupunS: Yes, should be possible.

Unfortunately the follow-up PRs also depend on this pattern and there is some refactoring around the test runner which could severely conflict with this change. I will refactor once the first four PRs (up to adding transaction support) are merged. That way I can also change it once for all test-stdlib functions. WDYT?

turbolent: Of course, please feel free to do work in a follow-up PR if that makes life easier 👍

turbolent (approved):
👏

PR #1830, opened 2022-07-20
Robert/custom codec

### LINKED ISSUE #1636: Make composite type information of elaboration serializable
### Issue To Be Solved

We would like to store composite type information, which is computed by the type checker and saved in the elaboration, in the execution state. This would allow us to load type information as-is, without having to potentially re-parse and re-check contracts.

### Suggested Solution

- [ ] Determine which parts of the elaboration should be saved and if it is available
  - Composite members?
- [ ] (Add missing information to elaboration, if missing ^)
- [ ] Determine serialization format
  - CBOR, like Cadence value serialization?
  - Or format closer to external encoding format? Not determined yet
- [ ] Implement serialization
- [ ] Implement deserialization
- [ ] Tests

robert-e-davidson3:
Closes #1636 

## Description

This is a custom codec for [de]serializing the sema types and the parts of the Elaboration needed for succinctly describing encoded Cadence values.

The goal for this PR is to have a codec in place to support the rest of the External Cadence Value Encoding epic rather than to have code that is ready for production in itself. The TODOs sprinkled throughout this PR reflect that goal: some of the TODOs point out potential bugs.

______

### DIFF
--- a/encoding/custom/sema_codec/codec_test.go
+++ b/encoding/custom/sema_codec/codec_test.go
@@ -1029,9 +1029,15 @@ func TestSemaCodecInterfaceType(t *testing.T) {
 			InitializerParameters: parameters,
 		}
 
-		// TODO container type
+		empty := &sema.InterfaceType{
+			Members: &sema.StringMemberOrderedMap{},
+		}
 
-		// TODO nested types
+		interfaceType.SetContainerType(empty)
+
+		nestedTypes := &sema.StringTypeOrderedMap{}
+		nestedTypes.Set("none", empty)
+		interfaceType.SetNestedTypes(nestedTypes)
 
 		encoder, decoder, buffer := NewTestCodec()
 
@@ -1082,9 +1088,22 @@ func TestSemaCodecInterfaceType(t *testing.T) {
 			[]byte(parameters[0].Identifier),
 			[]byte{byte(sema_codec.EncodedBoolTrue)},
 
-			[]byte{byte(sema_codec.EncodedSemaNilType)}, // no container type
+			[]byte{byte(sema_codec.EncodedSemaInterfaceType)}, // container type is empty interface
+			[]byte{sema_codec.NilLocationPrefix[0]},
+			[]byte{0, 0, 0, 0},
+			[]byte{byte(common.CompositeKindUnknown)},
+			[]byte{byte(sema_codec.EncodedBoolFalse)},
+			[]byte{0, 0, 0, 0},
+			[]byte{byte(sema_codec.EncodedBoolTrue)},
+			[]byte{byte(sema_codec.EncodedBoolTrue)},
+			[]byte{byte(sema_codec.EncodedSemaNilType)},
+			[]byte{byte(sema_codec.EncodedBoolTrue)},
 
-			[]byte{byte(sema_codec.EncodedBoolTrue)}, // no nested types
+			[]byte{byte(sema_codec.EncodedBoolFalse)}, // nested type
+			[]byte{0, 0, 0, 1},
+			[]byte{0, 0, 0, 4},
+			[]byte("none"),
+			[]byte{byte(sema_codec.EncodedSemaPointerType), 0, 0, 0, 0xb4}, // nested type is also container type
 		)
 
 		assert.Equal(t, expected, buffer.Bytes(), "encoded bytes differ")
@@ -1116,7 +1135,8 @@ func TestSemaCodecInterfaceType(t *testing.T) {
 
 			assert.Equal(t, parameters, i.InitializerParameters, "parameters")
 
-			assert.Nil(t, i.GetContainerType(), "container type")
+			assert.Equal(t, i.GetContainerType(), empty, "container type")
+			assert.Equal(t, i.GetNestedTypes(), nestedTypes, "nested types")
 		default:
 			assert.Fail(t, "Decoded type is not *sema.InterfaceType")
 		}
--- a/encoding/custom/sema_codec/codec_test.go
+++ b/encoding/custom/sema_codec/codec_test.go
@@ -553,7 +553,7 @@ func TestSemaCodecBadTypes(t *testing.T) {
 	t.Run("unknown type", func(t *testing.T) {
 		_, decoder, buffer := NewTestCodec()
 
-		fakeType := byte(255)
+		fakeType := byte(0xff)
 		buffer.Write([]byte{fakeType})
 
 		_, err := decoder.Decode()
--- a/encoding/custom/sema_codec/codec_test.go
+++ b/encoding/custom/sema_codec/codec_test.go
@@ -5,12 +5,13 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
 	"github.com/onflow/cadence/encoding/custom/sema_codec"
 	"github.com/onflow/cadence/runtime/ast"
 	"github.com/onflow/cadence/runtime/common"
 	"github.com/onflow/cadence/runtime/sema"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 )
 
 func TestSemaCodecSimpleTypes(t *testing.T) {
--- a/encoding/custom/sema_codec/codec_test.go
+++ b/encoding/custom/sema_codec/codec_test.go
@@ -1,0 +1,1643 @@
+package sema_codec_test
+
+import (
+	"bytes"
+	"fmt"
+	"github.com/onflow/cadence/encoding/custom/sema_codec"
+	"github.com/onflow/cadence/runtime/ast"
+	"github.com/onflow/cadence/runtime/common"
+	"github.com/onflow/cadence/runtime/sema"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestSemaCodecSimpleTypes(t *testing.T) {
+	t.Parallel()
+
+	type TestInfo struct {
+		SimpleType *sema.SimpleType
+		Type       sema_codec.EncodedSema
+	}
+
+	tests := []TestInfo{
+		{sema.AnyType, sema_codec.EncodedSemaSimpleTypeAnyType},
+		{sema.AnyResourceType, sema_codec.EncodedSemaSimpleTypeAnyResourceType},
+		{sema.AnyStructType, sema_codec.EncodedSemaSimpleTypeAnyStructType},
+		{sema.BlockType, sema_codec.EncodedSemaSimpleTypeBlockType},
+		{sema.BoolType, sema_codec.EncodedSemaSimpleTypeBoolType},
+		{sema.CharacterType, sema_codec.EncodedSemaSimpleTypeCharacterType},
+		{sema.DeployedContractType, sema_codec.EncodedSemaSimpleTypeDeployedContractType},
+		{sema.InvalidType, sema_codec.EncodedSemaSimpleTypeInvalidType},
+		{sema.MetaType, sema_codec.EncodedSemaSimpleTypeMetaType},
+		{sema.NeverType, sema_codec.EncodedSemaSimpleTypeNeverType},
+		{sema.PathType, sema_codec.EncodedSemaSimpleTypePathType},
+		{sema.StoragePathType, sema_codec.EncodedSemaSimpleTypeStoragePathType},
+		{sema.CapabilityPathType, sema_codec.EncodedSemaSimpleTypeCapabilityPathType},
+		{sema.PublicPathType, sema_codec.EncodedSemaSimpleTypePublicPathType},
+		{sema.PrivatePathType, sema_codec.EncodedSemaSimpleTypePrivatePathType},
+		{sema.StorableType, sema_codec.EncodedSemaSimpleTypeStorableType},
+		{sema.StringType, sema_codec.EncodedSemaSimpleTypeStringType},
+		{sema.VoidType, sema_codec.EncodedSemaSimpleTypeVoidType},
+	}
+
+	for _, typ := range tests {
+		t.Run(typ.SimpleType.Name, func(t *testing.T) {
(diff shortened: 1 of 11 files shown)


runtime/sema/type.go
```diff
@@ -3761,6 +3792,10 @@ func (t *CompositeType) GetNestedTypes() *StringTypeOrderedMap {
 	return t.nestedTypes
 }
 
+func (t *CompositeType) SetNestedTypes(nestedTypes *StringTypeOrderedMap) {
+	t.nestedTypes = nestedTypes
+}
```
robert-e-davidson3, line 3796: I needed a way to set the nested types when decoding. There are a couple of other options though:
1. From what I can tell, the only CompositeTypes with nested types are built-ins. These could be encoded as enums instead of encoded themselves.
2. To account for recursive types, it is necessary to save the CompositeType in the decoder's typeDefs map before decoding further. A deferral mechanism can be added so the CompositeType can be referred to during decoding before it is instantiated.

runtime/sema/type.go
```diff
@@ -3695,6 +3717,15 @@ func (t *CompositeType) IsExternallyReturnable(results map[*Member]bool) bool {
...
 }
 
+func (t *CompositeType) HasComputedMembers() bool {
+	return t.hasComputedMembers
+}
+
+// SetHasComputedMembers is intended to only be called for deserialization.
+func (t *CompositeType) SetHasComputedMembers(val bool) {
+	t.hasComputedMembers = val
+}
```
robert-e-davidson3, line 3726: This might be unnecessary for two possible reasons:
1. If computable members are only present in built-in CompositeTypes and we choose to encode built-in types as an enum then there's no need to ever encode hasComputedMembers because it would always be `false` for any encoded CompositeType.
2. Any CompositeType where `hasComputedMembers == true` is considered not-storable. If a type is not storable then does it ever need to be transmitted? If not then there's no need to encode such types.

runtime/sema/publicaccount_type.go
```diff
@@ -116,9 +116,9 @@ var PublicAccountType = func() *CompositeType {
 var PublicAccountKeysType = func() *CompositeType {
 
 	accountKeys := &CompositeType{
-		Identifier: AccountKeysTypeName,
-		Kind:       common.CompositeKindStructure,
-		importable: false,
```
robert-e-davidson3, line 121: I renamed `importable` to `ImportableWithoutLocation` to make it accessible and better describe the effect this value has on behavior.

This may be unnecessary: if CompositeTypes with `false` for `importable` are never transmitted then there's no need to encode them.

runtime/sema/type.go
```diff
@@ -454,6 +455,8 @@ type OptionalType struct {
 	Type Type
 }
 
+var _ Type = &OptionalType{}
```
robert-e-davidson3, line 458: Lines like this one are code cleanup: we intend these types to match these interfaces so it's desirable (IMO) to make that relationship explicit. My understanding is that the compiler optimizes away these statements so the only effect is a compile-time error if the type or the interface change.

dsainati1 (approved):
Lots of TODOs but this seems like a good base for the epic

runtime/sema/type.go
```diff
@@ -3557,7 +3572,13 @@ func (t *CompositeType) QualifiedIdentifier() string {
 }
 
 func (t *CompositeType) ID() TypeID {
+	if t == nil {
+		panic("*CompositeType is nil")
+	}
 	t.initializeIdentifiers()
+	if t.cachedIdentifiers == nil {
+		panic("*CompositeType.cachedIdentifiers is nil")
```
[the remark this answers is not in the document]
robert-e-davidson3: Ah yeah it is. Thanks for catching it!

SupunS:
@robert-e-davidson3 Given that this is still WIP, I think we should have this in a feature branch. Feature branches have CI/CD enabled, and the same reviewer requirements as the master.
Since the changes are also big, might be better to have a code walk-through session before getting it onto the master branch.

robert-e-davidson3:
> @robert-e-davidson3 Given that this is still WIP, I think we should have this in a feature branch (branch name starts with `feature/`). i.e: Fork a branch from master, name it as `feature/something`, and open this PR against that branch. Feature branches have CI/CD enabled, and the same reviewer requirements as the master. Since the changes are also big, might be better to have a code walk-through session before getting it onto the master branch.

That's a good idea. I'll open a revert PR and put this in a feature branch.

robert-e-davidson3:
> > @robert-e-davidson3 Given that this is still WIP, I think we should have this in a feature branch (branch name starts with `feature/`). i.e: Fork a branch from master, name it as `feature/something`, and open this PR against that branch. Feature branches have CI/CD enabled, and the same reviewer requirements as the master. Since the changes are also big, might be better to have a code walk-through session before getting it onto the master branch.
> 
> That's a good idea. I'll open a revert PR and put this in a feature branch.

@SupunS https://github.com/onflow/cadence/pull/1847

SupunS:
Thanks! Sorry, I should have suggested that earlier. It went completely out of my radar

encoding/custom/sema_codec/encode.go
```diff
@@ -0,0 +1,1101 @@
...
+}
+
+func (e *SemaEncoder) write(b []byte) (err error) {
+	_, err = e.w.Write(b)
+	return
+}
+
+func EncodeArray[T any](e *SemaEncoder, arr []T, encodeFn func(T) error) (err error) {
+	// TODO save a bit in the array length for nil check?
+	err = e.EncodeBool(arr == nil)
```
turbolent, line 1037: Maybe flip this? When reading the test cases above, it was surprising that `false` meant there is an array to follow. 

Maybe also differentiate between optional and non-optional arrays? This function not only encodes an array, but also handles the case where the array is nil. That means that in cases where there always is an array, an additional byte is wasted. The simplest solution might be to just always write just the size and write 0 for the nil case?

encoding/custom/sema_codec/encode.go
```diff
@@ -0,0 +1,1101 @@
...
+	EncodedBoolTrue
+)
+
+func (e *SemaEncoder) EncodeBool(boolean bool) (err error) {
+	b := EncodedBoolFalse
+	if boolean {
+		b = EncodedBoolTrue
+	}
+
+	return e.write([]byte{byte(b)})
```
turbolent, line 893: This allocates a slice, maybe add a dedicated single-byte write function

encoding/custom/sema_codec/codec_test.go
```diff
@@ -0,0 +1,1705 @@
...
+		{sema.Word64Type, sema_codec.EncodedSemaNumericTypeWord64Type},
+		{sema.FixedPointType, sema_codec.EncodedSemaNumericTypeFixedPointType},
+		{sema.SignedFixedPointType, sema_codec.EncodedSemaNumericTypeSignedFixedPointType},
+		{sema.Fix64Type, sema_codec.EncodedSemaFix64Type},
+		{sema.UFix64Type, sema_codec.EncodedSemaUFix64Type},
+	}
+
+	for _, typ := range tests {
+		t.Run(typ.SimpleType.String(), func(t *testing.T) {
+			t.Parallel()
```
turbolent, line 114: using `t.Parallel` in a sub-test nested in a for-loop doesn't work, see https://github.com/onflow/cadence/pull/1851#discussion_r946218284

encoding/custom/sema_codec/codec_test.go
```diff
@@ -0,0 +1,1705 @@
...
+		common.AddressLocationPrefix,
+		common.IdentifierLocationPrefix,
+		common.ScriptLocationPrefix,
+		common.StringLocationPrefix,
+		common.TransactionLocationPrefix,
+		common.REPLLocationPrefix,
+		sema_codec.NilLocationPrefix,
+	} {
+		t.Run(fmt.Sprintf("prefix: %s", prefix), func(t *testing.T) {
+			t.Parallel()
```
turbolent, line 845: Same as above

encoding/custom/sema_codec/decode.go
```diff
@@ -0,0 +1,1069 @@
...
+
+		default:
+			err = fmt.Errorf("unknown type identifier: %d", typeIdentifier)
+		}
+	}
+
+	return
+}
+
+func (d *SemaDecoder) EncodePointer() (t sema.Type, err error) {
```
turbolent, line 144: Decode?

runtime/sema/type.go
turbolent: Correct! Thank you for adding these 👍

encoding/custom/sema_codec/decode.go
```diff
@@ -0,0 +1,1069 @@
...
+// Decode reads custom-encoded bytes from the io.Reader and decodes them to a
+// Sema type. There is no assumption about the top-level Sema type so the first
+// byte must specify the top-level type. Usually this will be a CompositeType.
+//
+// This function returns an error if the bytes represent a custom encoding that
+// is malformed, does not conform to the custom specification, or contains
+// an unknown composite type.
+func (d *SemaDecoder) Decode() (t sema.Type, err error) {
+	return d.DecodeType()
+}
```
turbolent, line 75: Maybe just remove this function if all it does is calling another function?

encoding/custom/sema_codec/decode.go
```diff
@@ -0,0 +1,1069 @@
...
+func (d *SemaDecoder) EncodePointer() (t sema.Type, err error) {
+	bufferOffset, err := d.DecodeLength()
+	if err != nil {
+		return
+	}
+
+	if knownType, defined := d.typeDefs[bufferOffset]; defined {
+		t = knownType
+	} else {
+		err = fmt.Errorf(`pointer to unknown type: %d`, bufferOffset)
```
turbolent, line 153: Could you please add tests for this and other uncovered code reported by codecov, especially the error/unhappy cases?

encoding/custom/sema_codec/encode.go
```diff
@@ -0,0 +1,1101 @@
...
+		return e.EncodeTypeIdentifier(EncodedSemaAddressType)
+	case nil:
+		return e.EncodeTypeIdentifier(EncodedSemaNilType)
+	}
+
+	// Recursable types
+	if bufferOffset, usePointer := e.typeDefs[t]; usePointer {
+		return e.EncodePointer(bufferOffset)
+	}
+	e.typeDefs[t] = e.w.length + 1 // point to the encoded type, not its type identifier
```
turbolent, line 114: Cool!

encoding/custom/sema_codec/encode.go
```diff
@@ -0,0 +1,1101 @@
...
+		err = encodeFn(v)
+		if err != nil {
+			return
+		}
+	}
+
+	return
+}
+
+type LengthyWriter struct {
```
turbolent, line 1092: Maybe don't export this

encoding/custom/sema_codec/decode.go
```diff
@@ -0,0 +1,1069 @@
...
+			return
+		}
+
+		mapToPopulate[common.TypeID(k)] = v
+	}
+
+	return
+}
+
+type LocatedReader struct {
```
turbolent, line 1060: Maybe don't export this

turbolent (commented):
Great work!

turbolent:
What was the reasoning for a fully custom binary format and not leveraging an existing format like CBOR?

encoding/custom/sema_codec/encode.go
robert-e-davidson3: I originally translated nil array into array of size zero but then the encoded and decoded types differed when an array was nil. I don't know if that matters outside of making tests a little more complex.

Also, I don't know if most nil-able values will ever actually be nil.

encoding/custom/sema_codec/codec_test.go
```diff
@@ -0,0 +1,1705 @@
...
+			&sema.GenericType{TypeParameter: &sema.TypeParameter{
+				Name:      name,
+				TypeBound: sema.Int32Type,
+				Optional:  true,
+			}},
+			Concat(
+				[]byte{byte(sema_codec.EncodedSemaGenericType)},
+				[]byte{0, 0, 0, byte(len(name))},
+				[]byte(name),
+				[]byte{byte(sema_codec.EncodedSemaNumericTypeInt32Type)},
```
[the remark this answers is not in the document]
robert-e-davidson3: It would be encoded as a "nil type":
`				[]byte{byte(sema_codec.EncodedSemaNilType)},`

(adding test to demonstrate this)

encoding/custom/sema_codec/encode.go
robert-e-davidson3: In my next PR, this will be in a library package, `common_codec`, since it's useful for the cadence.Value codec as well.

robert-e-davidson3:
> What was the reasoning for a fully custom binary format and not leveraging an existing format like CBOR?

My chief motivation is that our scaling needs are immense and a custom format will always have a higher potential efficiency. I'm not certain the benefit outweighs the cost but I think it's likely.

I tested CBOR but it didn't show much improvement over JSON in terms of encoded size. There are solutions but they require a lot of custom code. At a guess, writing the custom code for CBOR would be 50% to 80% of the effort of the custom codec.

bluesign:
> custom format will always have a higher potential efficiency

This is true but usually unrelated 

> I'm not certain the benefit outweighs the cost but I think it's likely.

This totally depends on the tradeoff you want to make, usually focusing on 2 sides of the triangle is not optimum. 
Here triangle is: payload size vs speed (encoding/decoding) vs interoperability 
Focus here seems like size and speed, with big loss on interoperability

> I tested CBOR but it didn't show much improvement over JSON in terms of encoded size.

This is also little meaningless ( focusing on the one side of the triangle ), hypothetically, if you have some gains with CBOR on speed side, with that budget, you can throw in a zstd compression/decompression with custom dictionary to process, that can lead to very small encoded sizes and still keep interoperability.

Even if you don't have any gains on the speed with CBOR ( though most likely you should have ) throwing a layer of compression still can justify the computation cost with the payload size reduction gained. ( considering the network traffic vs CPU tradeoffs )

PR #1831, opened 2022-07-20
Optimize interpreter configuration, reuse it in runtime environment

### LINKED ISSUE #1823: Avoid re-creating interpreter and checker configuration

turbolent:
Closes #1823

[Functional options are slow and use more memory than a plain struct](https://www.evanjones.ca/go-functional-options-slow.html). 

Refactor the functional options of the interpreter into a configuration struct. This also allows it to be reused in the runtime environment.

______


dsainati1 (approved):
This hurts my functional programming soul, but if functional options are indeed slower then this change makes sense.

SupunS (approved):

PR #1832, opened 2022-07-20
[Test Framework] Add support for importing and initializing contracts

SupunS:

## Description

This PR adds support for importing and initializing contracts in tests.
```cadence
import FooContract from "./path/to/FooContract"

pub fun testImportingContract() {
    var foo = FooContract()
    var result = foo.sayHello()
    assert(result == "hello from Foo")
}
```

### Note:
This currently doesn't support importing a contract with imports (i.e: more than one level of imports)
______


dsainati1 (approved):

test-framework/test_framework.go
```diff
@@ -197,7 +246,37 @@ func newChecker(program *ast.Program) (*sema.Checker, error) {
...
+						Identifier:               contractDecl.Identifier.Identifier,
+						Type:                     constructorType,
+						DocString:                contractDecl.DocString,
+						Access:                   contractDecl.Access,
+						Kind:                     contractDecl.DeclarationKind(),
+						Pos:                      contractDecl.Identifier.Pos,
+						IsConstant:               true,
+						ArgumentLabels:           constructorArgumentLabels,
+						AllowOuterScopeShadowing: false,
+					})
```
turbolent, line 275: As far as I understand, the goal here is to declare a contract as a constructor *function*, instead of the default, a *singleton value*?

Instead of having the checker declare the singleton, then removing that declaration, and finally adding a function declaration back, maybe the result of declaring a contract should be configurable in the checker?

That would also avoid having to export all the implementation details (`valueActivations`, etc.).

robert-e-davidson3 (approved):

test-framework/test_framework_test.go
```diff
@@ -104,11 +115,139 @@ func TestExecuteScript(t *testing.T) {
...
+	t.Run("init no params", func(t *testing.T) {
+		t.Parallel()
+
+		code := `
+            import FooContract from "./FooContract"
+
+            pub fun test() {
+                var foo = FooContract()
+                var result = foo.sayHello()
+                assert(result == "hello from Foo")
```
turbolent, line 139: 😍

turbolent (commented):
Nice!

test-framework/test_framework.go
SupunS: That sounds cleaner. Added a checker option `WithContractVariableHandler`, a one similar to its interpreter counterpart `WithContractValueHandler`: cd433916bc25c2da5cb7322cb6925c10647afd94

runtime/interpreter/interpreter.go
```diff
@@ -1827,14 +1827,26 @@ func (interpreter *Interpreter) declareNonEnumCompositeValue(
...
-			return contract
+
+			// Under normal circumstances, a contract value is always a CompositeValue.
+			// However, in the test framework, an imported contract is constructed via a constructor function.
+			// Hence, during tests, the value is a HostFunctionValue.
+			switch valueHandler := valueHandler.(type) {
+			case *CompositeValue:
+				valueHandler.NestedVariables = nestedVariables
+			case *HostFunctionValue:
+				valueHandler.NestedVariables = nestedVariables
```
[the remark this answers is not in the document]
SupunS: added  in c28f4a8bc4dc14a15af7974960fa6d1a9aaf0f5d

turbolent (approved):
Nice work!

Issue #208, opened 2020-07-04
Storage Querying API


cybercent:
### Issue To Be Solved
A way to iterate over the storage to be able to list the resources owned by an account.

### (Optional): Suggest A Solution

- using a Cadence script:
`for entry in account.storageEntries {
log(entry)
}`
- using RPC:
` getAccount(“0x01")`
The response result can have a `storage` field with data in the cadence format (json with types)

* Details of the technical implementation
* Tradeoffs made in design decisions
* Caveats and considerations for the future

If there are multiple solutions, please present each one separately. Save comparisons for the very end.)

### (Optional): Context
I’m working on a marketplace and I was looking for a way to list all NFTs owned by an account.

cybercent:
Hi @turbolent is there an ETA for the Storage Querying API ?

turbolent:
@cybercent There's no ETA, but it's definitely one of the things we'll work on soon, along with other storage API improvements like #369 and #376.

I've added a [roadmap for Cadence](https://github.com/onflow/cadence/blob/master/ROADMAP.md) yesterday, and improving the storage API is a high-priority item.

If you're interested in helping out, let me know!

10thfloor:
Bump!

testrain:
Is this issue resolved? -> https://docs.onflow.org/faq/developers/#how-can-i-see-what-fungible-tokens-an-account-has

turbolent:
@testrain No, this issue is still open. The documentation you linked requires knowledge about the fact that the account has certain values stored. The idea for this issue is to allow developers to query what storage paths store values, and what they are, without having to prior knowledge what is stored (i.e. what storage paths are used).

We're currently working on a new dictionary data structure and will eventually use is for account storage, effectively giving us the functionality to list storage paths which store values for free.

atapin:
@turbolent any updates for this issue?

D10100111001:
This is the one thing that is highly painful and annoying when working with Flow. I think this should be prioritized. Any other blockchain allows any dev to generically get all NFTs for a given account. In the case of flow, because an NFT can be stored into any storage path, it is impossible to generically show all NFTs owned by the an account. Another problem that will need to be solved after the generic storage querying is figured out is generically loading metadata for the NFTs. The INFT interface should have the signature for a the getMetadata function so developers can generically pull it for any NFT in a collection.

dete:
> This is the one thing that is highly painful and annoying when working with Flow. 

Agreed!

[...]

PR #1851, opened 2022-07-27
Storage Iteration functions for accounts

dsainati1:
Part of https://github.com/onflow/cadence/issues/208
## Description

This adds three new functions to the `AuthAccount` object `forEachPublic`, `forEachPrivate` and `forEachStored` (as well as just `forEachPublic` to `PublicAccount`) that iterate over all the paths (and the types of the objects stored there) in the respective domains of the account. 

Each of these functions takes a function parameter of type `(X, Type): Bool` (where `X` is the type of the path being iterated over by the function in question (`PublicPath`, `PrivatePath` and `StoragePath` respectively)), which is applied to every element stored in that account under the domain in question. The first parameter holds the path of the element, and the second holds its runtime type.

These paths and types are loaded from storage one by one, to avoid loading the entirety of storage into memory in the case of large accounts. 

The specific order in which the objects are iterated over is undefined, as is the behavior when a path is added or removed from storage, whether by linking, unlinking, saving or loading. The behavior is, however, deterministic; we make no guarantees about what we do but we do the same thing every time. 
______


docs/language/accounts.mdx
```diff
@@ -121,6 +124,11 @@ to the `prepare` phase of the transaction.
       fun getLinkTarget(_ path: CapabilityPath): Path?
       fun unlink(_ path: CapabilityPath)
 
+      // Storage iteration  
+      fun forEachPublic(_ function: ((PublicPath, Type): Bool)): Void
```
dsainati1, line 128: If we don't make any guarantees about the order of iteration, is it actually useful to be able to terminate iteration using this boolean return value? What use cases would there be for wanting to get, say, 3 random NFTs out of storage without caring specifically which ones they are?

bluesign: Searching for specific `Type` can be some use case. Normally people will have one Collection Type in storage.

bluesign:
> These paths and types are loaded from storage one by one.

Why do we load full value here ? Shouldn't we load type info only ? This can cause big usability problems.

dsainati1:
We do load only the runtime type info here, the actual stored value itself is not loaded.

runtime/interpreter/interpreter.go
```diff
@@ -3673,6 +3673,50 @@ func (interpreter *Interpreter) allAccountPaths(addressValue AddressValue, getLo
...
+
+			for key, value := storageIterator.Next(); key != "" && value != nil; key, value = storageIterator.Next() {
+				pathValue := NewPathValue(invocation.Interpreter, domain, key)
+				runtimeType := NewTypeValue(invocation.Interpreter, value.StaticType(invocation.Interpreter))
+
+				subInvocation := NewInvocation(
+					interpreter,
+					nil,
+					[]Value{pathValue, runtimeType},
+					[]sema.Type{pathType, sema.MetaType},
```
SupunS, line 3699: Can `[]sema.Type{pathType, sema.MetaType}` be extracted out and re-use across iterations? I guess it is not updated in downstream codes?

docs/language/accounts.mdx
SupunS, line 128: I understand that only the `Type` is made available here to avoid loading the value. But in the implementation, the value is already loaded anyway (i.e: `storageIterator.Next()` returns the value), so wouldn't it be better to just make the value available here as well? I think loading a value doesn't load the entire value graph (i.e: lazy loading). So it should be fine?

dsainati1: The problem with making the value available is that if it is a resource, it needs to be removed and then resaved in order to avoid implicitly duplicating it (which can also re-order things in storage). By just providing the type and then making the developer handle loading resources explicitly we avoid implicit duplication.

SupunS: ah ok, that makes sense!

fxamacker (commented):
Nice work!  I focused on atree and storage related code.  I left a couple comments related to handling an edge case where `storageMap` can be nil.

runtime/tests/interpreter/account_test.go
```diff
@@ -1942,157 +1945,543 @@ func TestInterpretAccount_iteration(t *testing.T) {
...
 
-		value, err = inter.Invoke("getPublicPaths")
+			fun test(): Int {
+				account.save(S(value: 2), to: /storage/foo)
+				account.save("", to: /storage/bar)
+				account.link<&S>(/public/a, target:/storage/foo)
+				account.link<&String>(/public/b, target:/storage/bar)
+				account.link<&S>(/public/c, target:/storage/foo)
+				account.link<&S>(/public/d, target:/storage/foo)
+				account.link<&String>(/public/e, target:/storage/bar)
```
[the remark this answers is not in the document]
dsainati1: > Also, do we need to test path inside iteration functions?

Do you mean do we need to test if the path is correct? Some of these tests use `borrow` on the path value, which would only work if the path is accurate.

fxamacker (approved):
Great work! :+1:  I focused on atree and storage aspect.

runtime/interpreter/interpreter.go
```diff
@@ -3671,6 +3671,57 @@ func (interpreter *Interpreter) storageAccountPaths(addressValue AddressValue, g
 	return interpreter.accountPaths(addressValue, getLocationRange, common.PathDomainStorage, PrimitiveStaticTypeStoragePath)
 }
 
+func (interpreter *Interpreter) iterOverStorageDomain(addressValue AddressValue, domain common.PathDomain, pathType sema.Type) *HostFunctionValue {
```
turbolent, line 3674: Maybe improve naming here: starting the name of the function with a verb makes it sound like it performs the iteration, but it does not, it only returns a function. A common naming scheme is e.g.

```suggestion
func (interpreter *Interpreter) newStorageIterationFunction(addressValue AddressValue, domain common.PathDomain, pathType sema.Type) *HostFunctionValue {
```

runtime/sema/authaccount_type.go
```diff
@@ -222,6 +243,45 @@ const authAccountTypePrivatePathsFieldDocString = `
 All the private paths of an account
 `
 
+const authAccountForEachPublicDocString = `
+Iterate over all the public paths of an account.
+
+Takes two arguments: the first is the path (/domain/key) of the stored object, and the second is the runtime type of that object.
```
turbolent, line 249: The iteration functions accept one argument, the iteration function. The iteration function then is passed two arguments.

runtime/tests/checker/account_test.go
```diff
@@ -1644,3 +1644,242 @@ func TestCheckAccountPaths(t *testing.T) {
...
+			correctType string
+		}{
+			{name: "forEachPublic", correctType: "PublicPath"},
+			{name: "forEachPrivate", correctType: "PrivatePath"},
+			{name: "forEachStored", correctType: "StoragePath"},
+		}
+
+		for _, pair := range nameTypePairs {
+			t.Run("basic", func(t *testing.T) {
+				t.Parallel()
```
turbolent, line 1778: Careful: Using `t.Parallel` will pause the closure, then restart it. By the time it's restarted, the `pair` variable of the outer for-loop has been rebound. This basically results in totally random behaviour, in the worst case only  the last element of `nameTypePairs` is tested 3 times.

You can fix this by factoring the body of the loop into a helper function and then calling it, e.g. (pseudo-code)

```go
test := func(pair /* ... */) { 
	t.Run(fmt.Sprintf("basic: %s", pair.name), func(t *testing.T) {
		t.Parallel()
		// ...
	})

	// ... 
}

for _, pair := range nameTypePairs {
	test(pair)
}
```

This rebinds `pair`.

Also, the current loop runs a sub-test "basic" for each function, maybe use the pair's name in it, otherwise you get generated subtest names like `basic #0`, `basic #1`, etc.

turbolent (commented):
Nice!

runtime/interpreter/account.go
```diff
@@ -86,6 +86,15 @@ func NewAuthAccountValue(
 		sema.AuthAccountStoragePathsField: func(inter *Interpreter, getLocationRange func() LocationRange) Value {
 			return inter.storageAccountPaths(address, getLocationRange)
 		},
+		sema.AuthAccountForEachPublicField: func(inter *Interpreter, _ func() LocationRange) Value {
+			return inter.iterOverStorageDomain(address, common.PathDomainPublic, sema.PublicPathType)
```
robert-e-davidson3, line 90: The `Value` here is the `*HostValueFunction`. I'm not sure what a `HostValueFunction` is but should it be invoked here?

dsainati1: No. These define a method on the account object. They are invoked when the method is called, not when it is defined.

turbolent (approved):
Nice work! 👍

PR #1863, opened 2022-08-04
[Test Framework] Add initial support for Matchers

SupunS:

## Description

Supports:
- **Equal matcher**
  ```cadence
  // matcher with primitive type
  let matcher = Test.equal(1)

  // matcher with struct type
  let matcher = Test.equal(Foo())
  ```

- **Custom matchers**: 
  Supports defining custom matchers that deal with `AnyStruct` type.

  _e.g.(1): Customer matcher for a primitive type_
  ```cadence
  let matcher = Test.NewMatcher(fun (_ value: Int): Bool {
      return value == 7
  })

  assert(matcher.test(7))
  ```
   _e.g.(2): Customer matcher for a struct type_
  ```cadence
  let matcher = Test.NewMatcher(fun (_ value: AnyStruct): Bool {
      if !value.isInstance(Type<Int>()) {
          return false
      }

      return (value as! Int) > 5
  })

  assert(matcher.test(8))
  ```
   _e.g.(3): Customer matcher for handling **resource type**. Even though resource types are not supported, it can be made to work with a resource reference_
  ```cadence
  let matcher = Test.NewMatcher(fun (_ value: &Foo): Bool {
      return value.a == 4
  })

  let f <-create Foo(4)
  assert(matcher.test(&f as &Foo))
  ```

- **Expect function**
  ```cadence
  let matcher = Test.equal("other string")
  Test.expect("this string", matcher)
  ```
  _*This API was slightly changed compared to the proposal, to support resources. This API is not final and can be adjusted based on requirements/feedback._

______


runtime/interpreter/interpreter.go
```diff
@@ -1085,10 +1085,15 @@ func (interpreter *Interpreter) InvokeExternally(
 		preparedArguments[i] = interpreter.ConvertAndBox(getLocationRange, argument, nil, parameterType)
 	}
 
+	var self *CompositeValue
+	if boundFunc, ok := functionValue.(BoundFunctionValue); ok {
+		self = boundFunc.Self
+	}
```
dsainati1, line 1091: Does this have any effect on anything else outside of the testing framework? Seems like changing the receiver of an externally invoked function could have other consequences.

runtime/interpreter/interpreter.go
```diff
@@ -1085,10 +1085,15 @@ func (interpreter *Interpreter) InvokeExternally(
...
+
 	// NOTE: can't fill argument types, as they are unknown
 	invocation := NewInvocation(
 		interpreter,
-		nil,
+		self,
```
turbolent, line 1096: :+1:

runtime/stdlib/test.go
```diff
@@ -281,6 +442,80 @@ var testNewEmulatorBlockchainFunction = interpreter.NewUnmeteredHostFunctionValu
...
 
+// 'Test.NewMatcher' function.
+// Constructs a matcher that test only 'AnyStruct'.
+// Accepts test function that accepts subtype of 'AnyStruct'.
+//
+// Signature:
+//    Test.NewMatcher<T>(test: ((T): Bool)): AnyStruct<Test.Matcher>
+// where T is optional, and bound to 'AnyStruct'.
+//
+// Sample usage: Test.NewMatcher(fun (_value: Int: Bool) { return true})
```
turbolent, line 453:
```suggestion
// Signature:
//    fun Test.NewMatcher<T: AnyStruct>(test: ((T): Bool)): AnyStruct{Test.Matcher}
// where `T` is optional, and bound to `AnyStruct`.
//
// Sample usage: `Test.NewMatcher(fun (_ value: Int: Bool) { return true })`
```

runtime/stdlib/test.go
```diff
@@ -1016,6 +1251,112 @@ var emulatorBackendCommitBlockFunction = interpreter.NewUnmeteredHostFunctionVal
...
+			func(invocation interpreter.Invocation) interpreter.Value {
+
+				thisValue, ok := invocation.Arguments[0].(interpreter.EquatableValue)
+				if !ok {
+					panic(errors.NewUnreachableError())
+				}
+
+				equal := thisValue.Equal(
+					inter,
+					interpreter.ReturnEmptyLocationRange,
```
turbolent, line 1346:
```suggestion
					invocation.GetLocationRange,
```

runtime/stdlib/test.go
```diff
@@ -1033,3 +1374,251 @@ func (e TestFailedError) Unwrap() error {
...
+// 'Test.DefaultMatcher' struct.
+//
+// This is the default implementation of 'Test.Matcher' interface.
+// It accepts 'Any' for the test function. Hence, can be used with both structs and resources.
+// But the usage is limited by the constructor.
+// i.e: the constructor is hidden from users, to avoid misuses (see below).
+// Instead, use 'Test.NewMatcher()' to construct a matcher that test only 'AnyStruct'.
+//
+// Limitations:
+// Handling resource inside matchers is not supported yet, only exception being the built-in matchers.
```
turbolent, line 1387: Can `AnyStruct` be used instead then?

runtime/testlib_test.go
```diff
@@ -114,3 +138,697 @@ func TestExecuteScript(t *testing.T) {
...
+		script := `
+            import Test
+
+            pub fun main() {
+
+                let matcher = Test.NewMatcher(fun (_ value: Int): Bool {
+                     return value == 7
+                })
+
+                assert(matcher.test(7))
```
turbolent, line 187: Just to clarify my understanding: `matcher` has non-generic type `AnyStruct{Test.Matcher}` here, correct?

If that's the case, then `matcher.test` could be called with a non-`Int` value here – is the run-time type checked somewhere? (I can't quite see it)

runtime/testlib_test.go
```diff
@@ -114,3 +138,697 @@ func TestExecuteScript(t *testing.T) {
...
+
+            pub fun main() {
+
+                let matcher = Test.NewMatcher(fun (_ value: &Foo): Bool {
+                    return value.a == 4
+                })
+
+                let f <-create Foo(4)
+
+                assert(matcher.test(&f as &Foo))
```
turbolent, line 215: 😍

turbolent (commented):
Great work!

runtime/interpreter/interpreter.go
SupunS: It shouldn't really affect others. `InvokeExternally` is used to invoke a cadence function from Go. For bound functions (e.g: `foo.bar()`), the receiver information was missing. Perhaps it was never needed before.

runtime/testlib_test.go
SupunS: Aha, good catch! This is not handled at the moment. The test-case I had added for this is flawed (it checks equality, so it fails anyway). Need to add a runtime check.
Enforcing that statically would have been much better, but I don't see a way to do that. Do you have any ideas?

runtime/stdlib/test.go
SupunS: Yeah we could. The only reason to add `Any` here was to allow some built-in matchers like `equal`, `beNil`, etc. to work with resources. If we are to say that we aren't going to support even those with resources (custom matchers are anyway not supported), then we can just make this `AnyStruct`

runtime/testlib_test.go
SupunS: added the runtime check with 4792dd12f0565d7972d43f0fef7a50661afe69fb

runtime/stdlib/test.go
turbolent: Maybe we can use references for resources?

runtime/testlib_test.go
turbolent: Nice! Thank you for also adding a test

runtime/stdlib/test.go
SupunS: The problem with references is that equality works differently. It checks whether the reference is equal (i.e: they refer to the same resource), but not whether the two resources referred to are equal.

SupunS: yea, that makes sense

SupunS: updated in e0d14e7026c9e533dcbf95fae917631ce580e9e7 to use `AnyStruct`

turbolent (approved):
Nice!

SupunS:
@turbolent Thanks for the review! I did a quick refactor to remove the `Matcher` interface because it was added in the first place to support resources, and is no longer needed / adds complexity.

Can you have one last look? It's mostly this commit: 58f9b11908eefed4c47e71b89bd7bf529b5b620a

turbolent (approved):
Looks good!

PR #1869, opened 2022-08-09
[Test Framework] Add support for loading programs from local files

SupunS:

## Description

⚠️  _**NOTE**: Added this as an experimental feature._ 

One of the major pain points in executing transactions/script or deploying contracts during tests is, having to construct the source-code string inline. Since multi-line string is also not supported this can become a real hassle.

Hence, this PR introduces a utility method to load programs from local files. Works similar to resolving the imports from a local file, and also leverages the same file resolver as imports. The resolver has to be provided by the user of the test runner (i.e: CLI etc.).

```cadence
let script = Test.readFile("./sample/script.cdc")
```
The program/test will terminate upon an error in reading the file.

### Future Improvements:
- Maybe provide support for having placeholders in the file content, and dynamically fill them during reading the file (something similar to `sprintf()`). Can be useful to inject import addresses etc. 
______


dsainati1 (approved):

turbolent (approved):
Good idea, this is definitely useful!

PR #1870, opened 2022-08-11
Purity Enforcement

dsainati1:
Implements the enforcement for the purity analysis implemented in https://github.com/onflow/cadence/pull/1818

Currently, impure operations (those operations banned in `pure` contexts), are:
  * Calling an impure function (including any functions that modify account state or storage)
  * Writing to any resources
  * Writing to any references
  * Indexed assignment or writes to any variables not statically knowable to have been defined in the current function's scope

This is very restrictive; however given that the purpose of `pure` functions is to allow some operations to be performed in function pre and post-conditions, it is unlikely that support for a more complex analysis is worth the complexity it would require. As it stands, this allows calls to pure functions on composites, creation of composites with `pure` initializers, and any operations that don't perform an assignment. 

______


robert-e-davidson3 (approved):

runtime/sema/authaccount_type.go
```diff
@@ -367,7 +367,7 @@ var AuthAccountTypeLinkFunctionType = func() *FunctionType {
 	}
 
 	return &FunctionType{
-		Purity: PureFunction,
+		Purity: ImpureFunction,
```
turbolent, line 370: 👍

runtime/sema/check_assignment.go
```diff
@@ -107,6 +110,76 @@ func (checker *Checker) checkAssignment(
...
+	//     let b: &[Int] = [0]
+	//     let c = [a, b]
+	//     c[x][0] = 4 // we cannot know statically whether a or b receives the write here
+	// }
+	if variable == nil {
+		checker.ObserveImpureOperation(assignment)
+		return
+	}
+
+	// `self` technically exists in param scope, but should still not be writeable
```
turbolent, line 148: Does this imply that initializers assigning to fields of `self` are considered pure?

runtime/sema/check_assignment.go
```diff
@@ -107,6 +110,76 @@ func (checker *Checker) checkAssignment(
...
+	//   let b <- f
+	//   b.x = 3 // b was created in the current scope but modifies the resource value
+	//   return <-b
+	// }
+	if _, ok := variable.Type.(*ReferenceType); ok || variable.Type.IsResourceType() ||
+		// when the variable is neither a resource nor a reference, we can write to if its
+		// activation depth is greater than or equal to the depth at which the current purity
+		// scope was created; i.e. if it is a parameter to the current function or was created
+		// within it
+		checker.CurrentPurityScope().ActivationDepth > variable.ActivationDepth {
```
turbolent, line 179: Maybe improve indentation, the last case (activation depth) looks like part of the body:
```suggestion
	//
	// When the variable is neither a resource nor a reference, 
	// we can write to if its activation depth is greater than or equal to the depth 
	// at which the current purity scope was created; 
	// i.e. if it is a parameter to the current function or was created within it.
	//
	if _, ok := variable.Type.(*ReferenceType); ok || 
		variable.Type.IsResourceType() ||
		checker.CurrentPurityScope().ActivationDepth > variable.ActivationDepth {

```

runtime/sema/check_assignment.go
dsainati1: Yes, for the following reason: whenever we call a constructor from inside a pure scope, the value being constructed is local to that scope, so it is safe to create a new value from within a pure scope. This means that functions that just construct new values are technically pure (in the same way that they are in a functional programming sense), as long as they don't modify anything else while constructing those values.

turbolent: Ah, right!

runtime/tests/checker/purity_test.go
```diff
@@ -195,3 +196,1028 @@ func TestCheckPuritySubtyping(t *testing.T) {
...
+	t.Run("indeterminate write", func(t *testing.T) {
+		t.Parallel()
+		_, err := ParseAndCheck(t, `
+		let a: [Int] = []
+		pure fun foo() {
+			let b: [Int] = []
+        	let c = [a, b]
+       		c[0][0] = 4
+		}
+		`)
```
SupunS, line 636: seems there is a mix of tabs and whitespaces

runtime/tests/checker/purity_test.go
```diff
@@ -195,3 +196,1028 @@ func TestCheckPuritySubtyping(t *testing.T) {
...
+				self.x = x
+			}
+		}
+		
+		pure fun foo(_ s: &S) {
+			s.x = 3
+		}
+		`)
+
+		errs := ExpectCheckerErrors(t, err, 1)
```
SupunS, line 732: Also, add some tests for mutating references inside containers (struct field, arrays, dictionaries).
i.e: a reference somewhere in a member-access/index-access expression chain.

runtime/sema/check_assignment.go
```diff
@@ -107,6 +110,83 @@ func (checker *Checker) checkAssignment(
...
+	// if the target is not a variable (e.g. a nested index expression x[0][0], or a nested
+	// member expression x.y.z), the analysis cannot know for sure where the value being
+	// assigned to originated, so we must default to impure. Consider:
+	// -----------------------------------------------------------------------------
+	// let a: &[Int] = &[0]
+	// pure fun foo(_ x: Int) {
+	//     let b: &[Int] = &[0]
+	//     let c = [a, b]
+	//     c[x][0] = 4 // we cannot know statically whether a or b receives the write here
+	// }
```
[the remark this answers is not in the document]
dsainati1: Hmm,  I think you're right, although we also need to check whether any accessed member is a resource as well; since those are moved rather than copied.

runtime/sema/check_assignment.go
```diff
@@ -110,6 +113,95 @@ func (checker *Checker) checkAssignment(
...
+		switch targetExp := target.(type) {
+		case *ast.IdentifierExpression:
+			baseVariable = checker.valueActivations.Find(targetExp.Identifier.Identifier)
+			if baseVariable != nil {
+				accessChain = append(accessChain, baseVariable.Type)
+			}
+			inAccessChain = false
+		case *ast.IndexExpression:
+			target = targetExp.TargetExpression
+			elementType := checker.visitIndexExpression(targetExp, true)
```
SupunS, line 158: I assume the `targetExp.TargetExpression` is already visited once by the time we reach here? 

If what is needed here is only the type info, maybe look it up from `checker.Elaboration.IndexExpressionTypes[targetExp]` 

Same for member expression as well.

runtime/sema/check_assignment.go
```diff
@@ -110,6 +113,95 @@ func (checker *Checker) checkAssignment(
...
+		case *ast.MemberExpression:
+			target = targetExp.Expression
+			memberType, _, _ := checker.visitMember(targetExp)
+			accessChain = append(accessChain, memberType)
+		default:
+			inAccessChain = false
+		}
+	}
+
+	if baseVariable == nil {
```
SupunS, line 169: maybe add a brief explanation here about what is meant by `baseVariable == nil`.
I guess it's cases like `foo().something = 5`

SupunS: can we use the type of the expression to perform the check, in such cases? e.g: `foo()`s type is the return type of the function. Target type can be passed by the callee (callee should have this info)

runtime/sema/check_assignment.go
dsainati1: Member expression already has a cache check at the top of the function, so if we've already visited the member we should return early with the information we want. Good catch on the target expression though.

runtime/sema/check_assignment.go
dsainati1: It's not just function calls; we also have to consider cases like `b ? s1 : s2`, where `s1` may be local and `s2` might not be.

turbolent:
I'll have another look tomorrow

SupunS (approved):
LGTM!

runtime/sema/check_emit_statement.go
```diff
@@ -59,5 +59,10 @@ func (checker *Checker) VisitEmitStatement(statement *ast.EmitStatement) (_ stru
 		)
 	}
 
+	// emitting an event is an impure operation, but it is redundant to enforce that here:
+	// because the emit statement can only be used on events (as a result of the above checks),
+	// and because an event can only be invoked in an emit statement, every emit corresponds
+	// 1:1 to an event invocation, which is an impure function call. Thus we already report
+	// emits as impure by virtue of the necessity to create the event being emitted
```
turbolent, line 66:
```suggestion
	// Emitting an event is an impure operation, but it is redundant to enforce that here:
	// because the emit statement can only be used on events (as a result of the above checks),
	// and because an event can only be invoked in an emit statement, 
	// every emit corresponds 1:1 to an event invocation, which is an impure function call. 
	//
	// Thus we already report emits as impure by virtue of the necessity to create the event being emitted.
```

runtime/sema/checker.go
```diff
@@ -204,12 +204,12 @@ func (checker *Checker) CurrentPurityScope() PurityCheckScope {
 	return checker.purityCheckScopes[len(checker.purityCheckScopes)-1]
 }
 
-func (checker *Checker) PushNewPurityScope(enforce bool) {
+func (checker *Checker) PushNewPurityScope(enforce bool, depth int) {
 	checker.purityCheckScopes = append(
 		checker.purityCheckScopes,
 		PurityCheckScope{
```
turbolent, line 210: Didn't think of this in the previous PR: Can the purity scopes be combined with the function activations? Or are they not always mapped 1:1? That could reduce the complexity a bit

runtime/sema/checker.go
```diff
@@ -220,13 +220,14 @@ func (checker *Checker) PopPurityScope() PurityCheckScope {
...
+	}
+}
+
 func (checker *Checker) ObserveImpureOperation(operation ast.Element) {
 	scope := checker.CurrentPurityScope()
-	// purity is monotonic, if we already know this scope is impure, there's no need to continue
-	if scope.CurrentPurity != FunctionPurityView {
-		return
-	}
-	scope.CurrentPurity = FunctionPurityImpure
```
turbolent, line 229: Just curious, why was this removed?

turbolent (commented):
Great work!

runtime/sema/checker.go
dsainati1: No; they are specifically separate so we can push a pure scope when we enter a precondition or postcondition in the next PR.

runtime/sema/checker.go
dsainati1: Tracking the current purity of the scope is a relic of the original inference behavior; the purity of the scope we are in can no longer change so we don't need to update a variable here.

Issue #1805, opened 2022-07-12
Function pre- and post-conditions can have side effects


dsainati1:
### Issue To Be Solved

Currently function pre- and post-conditions can call arbitrary functions, which may have arbitrary side effects. This allows users to write underhanded code that modifies state during a function post-condition that circumvents any post-condition checks that an interface may require. 

### Suggested Solution

There are two possible solutions here: we could prevent any and all function calls in function conditions, or we could build out a type system feature to determine whether a given function is mutating or "pure".

PR #1871, opened 2022-08-11
Enforce `view` in function conditions

dsainati1:
Closes https://github.com/onflow/cadence/issues/1805 https://github.com/onflow/cadence/issues/1806

This changes Cadence to require that function preconditions and postconditions are `view`; i.e. cannot have side effects. 

______


runtime/nft_test.go
```diff
@@ -77,7 +77,7 @@ pub contract interface NonFungibleToken {
     // publish for their collection
     pub resource interface CollectionPublic {
         pub fun deposit(token: @NFT)
-        pub fun getIDs(): [UInt64]
+        pure pub fun getIDs(): [UInt64]
```
dsainati1, line 80: Note that this means the NonFungibleToken will need to be updated to make this function `pure` in order to roll out this change; and therefore that any contracts with resources implementing this interface will need to make their implementations `pure` as well. In most cases this should be a very straightforward change, however.

bluesign: @dsainati1 not better to change post condition ?

dsainati1: The whole point of this change is to allow people to call certain functions inside their conditions; so something like the postcondition here:
```
pub fun createEmptyCollection(): @Collection {
        post {
            result.getIDs().length == 0: "The created collection must be empty!"
        }
    }
}
``` 
is a pattern we would like to support.

bluesign: I meant we can change post condition in the standard, to not to use function. Changing standard, then expecting everyone to implement this change (even little) is too complicated. 

I agree supporting pure functions on post/pre is nice and very big improvement. Just I dont like unnecessary breaking change on many contracts if we can avoid.

runtime/tests/interpreter/resources_test.go
```diff
@@ -2545,48 +2545,3 @@ func TestInterpretReferenceUseAfterTransferAndDestruction(t *testing.T) {
 		assert.Equal(t, 24, invalidatedResourceErr.StartPosition().Line)
 	})
 }
-
-func TestInterpretResourceDestroyedInPreCondition(t *testing.T) {
```
turbolent, line 2549: Instead of removing the test case, adjust it to the new behaviour

runtime/tests/interpreter/resources_test.go
```diff
@@ -2545,3 +2545,46 @@ func TestInterpretReferenceUseAfterTransferAndDestruction(t *testing.T) {
 		assert.Equal(t, 24, invalidatedResourceErr.StartPosition().Line)
 	})
 }
+
+func TestInterpretResourceDestroyedInPreCondition(t *testing.T) {
```
turbolent, line 2549: The current version of this interpreter test is always a static error, so move/copy this test case to the checker tests.

Then, adjust this interpreter test case to test what was tested before: how the interpreter handles a resource being destroyed in a pre-condition. I think this requires annotating `destroyResource` with `pure`.

Also, maybe add some whitespace between the declarations.

dsainati1: Destroy is an impure operation, so destroyResource being annotated as pure would still be a static error. I am not sure how to recover the original behavior of this test to be honest, as it shouldn’t be possible anymore to destroy resources in conditions.

runtime/tests/checker/purity_test.go
```diff
@@ -1221,3 +1221,65 @@ func TestCheckContainerMethodPurity(t *testing.T) {
 		require.NoError(t, err)
 	})
 }
+
+func TestCheckConditionPurity(t *testing.T) {
```
SupunS, line 1225:
```suggestion
func TestCheckConditionPurity(t *testing.T) {
	t.Parallel()

```

docs/language/composite-types.mdx
```diff
@@ -228,6 +228,23 @@ token.id = 23
...
 Initializers do not support overloading.
 
+Initializers can also be declared with the `pure` keyword, to indicate that they do not perform any impure operations, and
+to allow them to be called from within other `pure` functions. In an initializer, writes to `self` are not considered impure
+like they are within other composite functions; as the value being constructed here is by definition local to the context 
+calling the initializer. 
+
+```cadence
+pub struct Token {
+    pub let id: Int
```
SupunS, line 238:
```suggestion
    pub var id: Int
```

runtime/nft_test.go
turbolent: @bluesign How could the condition look like instead?

bluesign: @turbolent maybe `result.ownedNFTs.keys.length` ?

dsainati1: No post condition would allow people to create "empty" collections that are not actually empty, which seems rather underhanded and might allow people to sneak malicious objects into others' accounts.

bluesign: @dsainati1 I was thinking like you before ( when I first come here ) actually, but actually I gave up on this. On Flow tendency is contract owner is the **boss**. There is nothing preventing contract owner to sneak objects to your account. 

PS:  I really support your way of thinking.

dsainati1: Hmm it looks like the typical (e.g. in TopShot) implementation for `getIDs` is to base it on the `ownedNFT`s array, so I think using `result.ownedNFTs.keys.length` would be similar enough, even if not exactly the same, to work. I would feel comfortable changing it to be that.

docs/language/composite-types.mdx
```diff
@@ -228,6 +228,23 @@ token.id = 23
 
 Initializers do not support overloading.
 
+Initializers can also be declared with the `view` keyword, to indicate that they do not perform any mutating operations, 
+and to allow them to be called from within other `view` functions. 
+In an initializer, writes to `self` are not considered non-view like they are within other composite functions,
+as the value being constructed here is by definition local to the context calling the initializer.
```
turbolent, line 234: Thanks for documenting this 👍 

Maybe invert the double negation. It is correct, but hard to read

docs/language/functions.mdx
```diff
@@ -479,6 +556,10 @@ fun incrementN() {
 }
 ```
 
+Both preconditions and postconditions are considered `view` contexts;
```
turbolent, line 559: Maybe link to the view functions section above

runtime/nft_test.go
```diff
@@ -999,13 +999,13 @@ pub contract TopShotShardedCollection {
...
-        pub fun getIDs(): [UInt64] {
+        view pub fun getIDs(): [UInt64] {
 
             var ids: [UInt64] = []
             // Concatenate IDs in all the Collections
-            for key in self.collections.keys {
+            for i, key in self.collections.keys {
                 for id in self.collections[key]?.getIDs() ?? [] {
-                    ids.append(id)
+                    ids[i] = id
```
turbolent, line 1008: This will result in an index out-of-bounds error. Why was it changed?

runtime/tests/interpreter/resources_test.go
```diff
@@ -2520,34 +2520,32 @@ func TestInterpretResourceDestroyedInPreCondition(t *testing.T) {
                 }
             }
         }
-
```
turbolent, line 2523: Maybe leave the line breaks between the declarations, otherwise the code becomes quite unreadable

runtime/missingmember_test.go
```diff
@@ -4408,9 +4408,9 @@ pub contract ExampleNFT {
 
         pub fun deposit(token: @NFT)
 
-        pub fun getIDs(): [UInt64]
+        view pub fun getIDs(): [UInt64]
 
-        pub fun idExists(id: UInt64): Bool
+        view pub fun idExists(id: UInt64): Bool
```
turbolent, line 4413: Why does this function have to be `view` now?

runtime/missingmember_test.go
```diff
@@ -1697,7 +1697,7 @@ pub contract ItemNFT: NonFungibleToken {
    pub resource interface ItemCollectionPublic {
        pub fun deposit(token: @NonFungibleToken.NFT)
        pub fun batchDeposit(tokens: @NonFungibleToken.Collection)
-       pub fun getIDs(): [UInt64]
+       view pub fun getIDs(): [UInt64]
```
turbolent, line 1700: Can the `getIDs` functions stay non-view, now that the function isn't used in a condition anymore?

turbolent (commented):
👍

runtime/nft_test.go
dsainati1: This was changed in order to make this function `view`, as `append` is a non-view function and thus not callable in a `view` context. However, if this doesn't need to be `view` then we can change it back.

runtime/missingmember_test.go
dsainati1: It was called from within a `getIds`, but can be non-view now.

runtime/nft_test.go
dsainati1: This is a limitation of the system really; because we can only annotate a function as being `view` or not `view`, there is no additional information given about what kind of mutating operation might happen inside of a non-`view` function like we would have with a true effect system. We have no way right now to say that a non-`view` method on an array only mutates the receiver, for example, so we can't express the idea that `append` is safe to use in `view` contexts when the receiver is local to that context. In order to be able to model this later, we would need to add more complexity to `view` annotations, or an additional purity keyword, to allow us to describe which kinds of mutations a function is allowed to perform, which feels like trouble than it's worth.

turbolent:
Looks like GitHub still breaks CI runs sometimes when changing branches. Pushing an empty commit should re-trigger CI

runtime/tests/interpreter/resources_test.go
```diff
@@ -2533,21 +2533,23 @@ func TestInterpretResourceDestroyedInPreCondition(t *testing.T) {
         }
 
         resource Bar  {}
-
+
```
turbolent, line 2536:
```suggestion

```

runtime/nft_test.go
turbolent: That makes sense, thank you for the explanation 👌

turbolent (approved):
Nice!

turbolent:
@SupunS ping

SupunS (approved):
Sorry, I thought I've already approved this in my last review.

LGTM!

PR #1879, opened 2022-08-16
Merging V0.25 branch back to master

SupunS:
## Description

Merging the v0.25 branch back to master, to get the branch-only changes onto the master branch.

______

### DIFF
--- a/version.go
+++ b/version.go
@@ -21,4 +21,4 @@
 
 package cadence
 
-const Version = "v0.24.1"
+const Version = "v0.25.0"
--- a/docs/language/contract-updatability.md
+++ b/docs/language/contract-updatability.md
@@ -118,7 +118,7 @@ A field may belong to a contract, struct, resource, or interface.
   // Updated contract
 
   pub contract Foo {
-      priv var a: Int   // access modifier changed to 'priv'
+      priv var a: String   // access modifier changed to 'priv'
   }
   ```
 
--- a/npm-packages/cadence-docgen/package.json
+++ b/npm-packages/cadence-docgen/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@onflow/cadence-docgen",
-  "version": "0.24.1",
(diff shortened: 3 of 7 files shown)


version.go
```diff
@@ -21,4 +21,4 @@
 
 package cadence
 
-const Version = "v0.24.1"
+const Version = "v0.25.0"
```
turbolent, line 24: This can unfortunately not get in and needs to stay as-is this time.

This file is generated and the input is the latest Git tag of master. As `v0.25.0` was tagged on a different branch, it does not appear on `master`. CI runs `go generate` for all files, checks the generated code is committed, and reports an error if there is a mismatch.

I had already tried this on `master` but had to revert it.

SupunS: oh! So that introduces a constraint that master can rely only on tags created on master, and not possible to update to 
 e.g: a patch-version created off release branches (e.g: to use a fix that went on a branch).

WDYT of updating that version lookup logic to something like:
- If the version specified in `version.go` <=  latest tag on master, then use the latest tag on master
- Otherwise, use the version specified.

The idea is to always use the latest version whenever possible.

turbolent: Yeah, we could just use the latest tag *across all branches*.  This can be done by changing https://github.com/onflow/cadence/blob/master/utils/version/main.go#L78 to call:
`git for-each-ref refs/tags --sort=-taggerdate --format='%(refname:short)' --count=1`

SupunS: my worry on that is, when we create new tags in branches (e.g: new patch releases on top of a release branch), then it'll break the master build, because now there's a new tag all of a sudden somewhere else, but master is not updated.

turbolent: Ah, I misunderstood what you had suggested! I'll open a PR for that

turbolent:
@SupunS I think once #1906 is merged, this should be mergable

turbolent (approved):

PR #1892, opened 2022-08-18
Add panic after continuing storage iteration after a storage mutation

### LINKED ISSUE #208: Storage Querying API

dsainati1:
Closes #208

Because modifying storage can arbitrarily rearrange the way that values are stored within it, we want to prevent users from iterating over a mutated storage. Thus, if users continue a storage iteration after performing a mutating operation, we will panic and abort execution. If users end iteration after their mutation however, this is safe, as no operations are performed that rely on the stored values maintaining their original, pre-mutation, arrangement. 

______


docs/language/accounts.mdx
```diff
@@ -641,8 +641,28 @@ as is the behavior when a path is added or removed from storage.
 
 <Callout type="warning">
 The order of iteration is undefined. Do not rely on any particular behaviour.
+
 Saving to or removing from storage during iteration can cause the order in which values are stored to change arbitrarily. 
-Continuing to iterate after such an operation may cause iteration to repeat some elements or skip others. 
+
+Continuing to iterate after such an operation will cause Cadence to panic and abort execution. In order to avoid such errors,
+we recommend not modifying storage during iteration. If you do, return `false` from the iteration callback to cause iteration 
+to end after the mutation like so:
```
turbolent, line 649: Breaking on logical ends makes future edits/reviews easier:

```suggestion
Continuing to iterate after such an operation will cause Cadence to panic and abort execution. 
In order to avoid such errors, we recommend not modifying storage during iteration. 
If you do, return `false` from the iteration callback to cause iteration to end after the mutation like so:
```

docs/language/accounts.mdx
```diff
@@ -641,8 +641,28 @@ as is the behavior when a path is added or removed from storage.
...
+
+```cadence
+account.save(1, to: /storage/foo1)
+account.save(2, to: /storage/foo2)
+account.save(3, to: /storage/foo3)
+account.save("qux", to: /storage/foo4)
+
+account.forEachStored(fun (path: StoragePath, type: Type): Bool {
```
turbolent, line 657: 👌

runtime/interpreter/interpreter.go
```diff
@@ -355,6 +355,8 @@ type Interpreter struct {
 	atreeValueValidationEnabled    bool
 	atreeStorageValidationEnabled  bool
 	tracingEnabled                 bool
+	inStorageIteration             bool
```
turbolent, line 358: Move both  `inStorageIteration` and `storageMutatedDuringIteration` to `sharedState`

runtime/interpreter/interpreter.go
```diff
@@ -3753,6 +3774,18 @@ func (interpreter *Interpreter) newStorageIterationFunction(addressValue Address
 				if !shouldContinue {
 					break
 				}
+
+				// it is not safe to check this at the beginning of the loop (i.e. on the next invocation of the callback)
+				// because if the mutation performed in the callback reorganized storage such that the iteration pointer is now
+				// at the end, we will not invoke the callback again but will still silently skip elements of storage. In order
+				// to be safe, we perform this check here to effectively enforce that users return `false` from their callback
+				// in all cases where storage is mutated
```
turbolent, line 3782: 👌

runtime/tests/interpreter/account_test.go
```diff
@@ -2597,3 +2588,318 @@ func TestInterpretAccount_iteration(t *testing.T) {
...
+					return true
+				})
+				account.save("baz", to: /storage/foo6)
+			}`,
+		)
+
+		_, err := inter.Invoke("test")
+		require.NoError(t, err)
+	})
+}
```
turbolent, line 2905: Could you please add a test case where an imported function modifies storage? That tests the code for the sub-interpreter

turbolent (commented):
Nice! This is going to make the iteration feature even more useful and ergonomic 👍

turbolent (approved):
👍

PR #1896, opened 2022-08-18
Remove invalid markdown in docs.

10thfloor:
For your consideration:

Closes https://github.com/onflow/next-docs-v1/issues/597

## Description

A user in Discord reported that inline Cadence snippets were rendering with some artifacts (See image below):

The new docs site has changed how inline snippets are handled, making this notation invalid.
This pr removes the artifacts. 

This might make updating these inline snippets later, if we decide to add support for inline highlighing of Cadence. 

______

### DIFF
--- a/docs/language/accounts.mdx
+++ b/docs/language/accounts.mdx
@@ -363,7 +363,7 @@ Both `StoragePath` and `CapabilityPath` are subtypes of `Path`.
 
 #### Path Functions
 
-- `cadence•fun toString(): String`
+- `fun toString(): String`
 
   Returns the string representation of the path.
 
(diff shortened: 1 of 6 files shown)


joshuahannan (commented):
I may be mistaken, but weren't the `cadence` markers here to indicate to whatever is displaying the code snippets to use cadence highlighting?  would this make that impossible?

dsainati1:
Yeah, it seems like the issue is with the new doc website not rendering the markdown properly, not with the markdown itself; we should fix it on the other end of the problem.

turbolent:
Agreed, it would be great if we could keep support for highlighting inline code.

@10thfloor can you point me to where Cadence highlighting is implemented in the new doc site?

Alternatively, we might want to switch to code blocks instead.

10thfloor:
> Agreed, it would be great if we could keep support for highlighting inline code.
> 
> @10thfloor can you point me to where Cadence highlighting is implemented in the new doc site?
> 
> Alternatively, we might want to switch to code blocks instead.

We're using Prism. We still need to develop a Prism grammar for Cadence.

https://github.com/onflow/developer-portal/blob/main/app/ui/design-system/src/lib/Components/InternalCodeblock/Code.tsx

j1010001:
As discussed, syntax highlighting should be re-added to the dev portal.

PR #1907, opened 2022-08-24
Optimize and make visitor generic

turbolent:
## Description

- Make the AST visitors generic
- Refactor the `Accept` method on `Element` to a global function, as Go does not support generic receiver methods (ideally we would like to have `Element` have a function `Accept[T](visitor Visitor[T]) T`)

______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -424,7 +424,7 @@ func (interpreter *Interpreter) Interpret() (err error) {
 	})
 
 	if interpreter.Program != nil {
-		interpreter.Program.Program.Accept(interpreter)
+		ast.Accept[any](interpreter.Program.Program, interpreter)
 	}
 
 	interpreter.interpreted = true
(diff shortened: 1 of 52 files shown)


runtime/interpreter/interpreter.go
```diff
@@ -637,7 +637,7 @@ func (interpreter *Interpreter) RecoverErrors(onError func(error)) {
 	}
 }
 
-func (interpreter *Interpreter) VisitProgram(program *ast.Program) ast.Repr {
+func (interpreter *Interpreter) VisitProgram(program *ast.Program) any {
```
dsainati1, line 640: Why are we using `any` here? Can we have these interpreter functions return `Value` like the checker ones return `Type`?

dsainati1 (commented):
This is awesome. Very excited for this!

runtime/interpreter/interpreter.go
turbolent: Unfortunately there are currently two types being returned, `Value` and `controlReturn`. 

That is why for now I kept the interpreter code as-is (`ast.Repr` was an alias for `any`). 
In a subsequent PR I was planning on refactoring this to a single "enum" `Result`, which can represent both.

dsainati1 (approved):

runtime/interpreter/interpreter.go
turbolent: Continued the improvements in #1919

dreamsmasher (approved):
LGTM, splitting the Visitor traits by target AST-node type is a nice touch.

PR #1908, opened 2022-08-24
Remove broken CI checks

schpet:
## Description

- ~Remove~ Fixes the broken "Lint Docs" check
- Remove legacy "call-refresh-doc" action

in the future, if we want to run docs/lint.js it needs to return correct exit codes (i.e. `process.exit(1)` for failure, `process.exit()` on success), and it should also output something useful instead of piping output to /dev/null or into variables


turbolent:
@schpet Can you please elaborate on how the cache refresh is deprecated? @10thfloor Do you have any details?

As for the linting: Instead of removing the check, we should fix it. The CI action should already return a status code, but you are right, it should log the errors.

10thfloor:
> @schpet Can you please elaborate on how the cache refresh is deprecated? @10thfloor Do you have any details?
> 
> As for the linting: Instead of removing the check, we should fix it. The CI action should already return a status code, but you are right, it should log the errors.

We're using a webhook based approach now. We created a github app that pings a webhook, when it sees a push from a repo we care about, with the diff, instead of using an action to send the diff. 

This way repo owners don;t need to install anything (ie GitHub Action) in their repo for changes to their docs to be reflected in the site.

schpet:
@turbolent appreciate your feedback on this: i've corrected the github action – and resolved the problem in the doc that it had identified.

turbolent:
@schpet Awesome, thank you!

@10thfloor Ah, great!

turbolent (approved):

PR #1910, opened 2022-08-24
Export environment checker and interpreter config to allow customization

turbolent:
## Description

The default checker and interpreter configurations may want to be changed, e.g. by the test framework. 
Export these two fields

______

### DIFF
--- a/runtime/environment.go
+++ b/runtime/environment.go
@@ -63,18 +63,23 @@ type Environment interface {
 }
 
 type interpreterEnvironment struct {
-	config                                Config
-	baseActivation                        *interpreter.VariableActivation
-	baseValueActivation                   *sema.VariableActivation
-	runtimeInterface                      Interface
-	storage                               *Storage
-	coverageReport                        *CoverageReport
-	codesAndPrograms                      codesAndPrograms
+	config Config
+
+	baseActivation      *interpreter.VariableActivation
+	baseValueActivation *sema.VariableActivation
+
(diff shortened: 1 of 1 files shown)


runtime/environment.go
```diff
@@ -63,18 +63,23 @@ type Environment interface {
...
 type interpreterEnvironment struct {
-	config                                Config
-	baseActivation                        *interpreter.VariableActivation
-	baseValueActivation                   *sema.VariableActivation
-	runtimeInterface                      Interface
-	storage                               *Storage
-	coverageReport                        *CoverageReport
-	codesAndPrograms                      codesAndPrograms
+	config Config
+
```
dsainati1, line 67: Can we remove the newlines here so the fields are all vertically aligned?

turbolent: Hmm, intentionally added newlines to group related fields and hopped it would help

dsainati1: I see. In that case I suppose it's fine but it is frustrating that grouping the fields this way also ruins the alignment.

turbolent: yeah, I usually have whitespace changes disabled on GitHub for that reason

dsainati1 (approved):

SupunS (approved):
For the moment, I've added this locally: https://github.com/onflow/cadence/commit/a7c0cec13e950fcd7644f2b8e9b1453e57aeebde.

Should be OK to merge this 👍

PR #1917, opened 2022-08-26
Fix ToBigEndianBytes implementations for U?Int(128|256)

### LINKED ISSUE #1913: `toBigEndianBytes` for `UInt256`, `UInt128`, and `Int128` may not result in what we want
### Issue To Be Solved
The following script shows what happened when we call `toBigEndianBytes` for `UInt256`, `UInt128`, and `Int128`. 
```cadence

pub fun main(): [UInt8] {
    let a: UInt256 = 1;
    var b = a.toBigEndianBytes();

    // output `[1]` when a is type `UInt256`, `UInt128`, and `Int128` 
    // but I think most developers want get `[0, 0, 0, 0, ... , 0, 1]`, the length of which is 32 for `UInt256` and 16 for `UInt128` and `Int128`
    return b;
}

```

When we call `toBigEndianBytes` for other types, for example, for `UInt64` we will get `[0, 0, 0, 0, 0, 0, 0, 1]`.

### Suggested Solution

Temporarily, I solved this problem manually as follows:
```cadence

// UInt128, Int128, and UInt256 need to be specially processed!
    //128
    pub fun to_be_bytes_u128(number: UInt128): [UInt8] {
        var output = number.toBigEndianBytes();
        while output.length < 16 {
            output = [UInt8(0)].concat(output);
        }

        return output;
    }

    pub fun to_be_bytes_i128(number: Int128): [UInt8] {
        var output = number.toBigEndianBytes();
        while output.length < 16 {
            output = [UInt8(0)].concat(output);
        }

        return output;
    }

    // 256
    pub fun to_be_bytes_u256(number: UInt256): [UInt8] {
        var output = number.toBigEndianBytes();
        while output.length < 32 {
            output = [UInt8(0)].concat(output);
        }

        return output;
    }

```

Are there any better or more efficient solutions?
Why the underlying implementation of `cadence` does not support this?

dreamsmasher:
Closes #1913 

## Description
.toBigEndianBytes() on Int128, Int256, UInt128, and UInt256 was returning a variable-length byte array due to their underlying representations (using golang's math/big.Int). this PR adds some fixed-size methods for outputting the binary representations of these types.

Changing the layout of these types in the future might be a good idea for performance and reducing allocations, e.g. using a fixed amount of u64's and carrying bits under arithmetic. Rust's primitive_types crate does something similar to great effect. 
______


go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/schollz/progressbar/v3 v3.8.3
 	github.com/stretchr/testify v1.7.3
 	github.com/turbolent/prettier v0.0.0-20220320183459-661cc755135d
+	github.com/ElrondNetwork/big-int-util v0.1.0
```
turbolent, line 19: This library is GPL licensed, so unfortunately cannot be used.

dreamsmasher: NOOOO

dreamsmasher: that's upsetting haha, I did notice the GPL license yesterday but a quick search told me that Apache and GPL were compatible. I guess they work together if you reverse the dependency order (GPL project depends on Apache) since Apache is less restrictive. My bad.

runtime/interpreter/big.go
```diff
@@ -57,19 +57,55 @@ func SignedBigIntToBigEndianBytes(bigInt *big.Int) []byte {
...
+			buf[i] = 255 // sign extend
+		}
+		for i := 0; i < len(buf)-offset; i++ {
+			buf[i+offset] = ^bytes[i]
+		}
+	case 0:
+		break
+	case 1:
+		bigInt.FillBytes(buf)
+	default:
```
dreamsmasher, line 80: codecov is yelling about these unreachable cases but how do we even test them? mock big.Int and add a pathological .Sign() implementation?

values_test.go
```diff
@@ -329,24 +329,24 @@ func TestToBigEndianBytes(t *testing.T) {
...
+			NewInt256(42):                 {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 42},
+			NewInt256(127):                {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 127},
+			NewInt256(128):                {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 128},
+			NewInt256(200):                {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 200},
+			NewInt256(-1):                 {255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255},
+			NewInt256(-200):               {255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 56},
+			NewInt256(-10000000000000000): {255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 220, 121, 13, 144, 63, 0, 0},
```
dreamsmasher, line 349: didn't add the min/max cases to this test suite because we can't construct an Int256 by passing in a machine int.

runtime/interpreter/big.go
```diff
@@ -30,7 +30,7 @@ func SignedBigIntToBigEndianBytes(bigInt *big.Int) []byte {
 	case -1:
 		// Encode as two's complement
 		twosComplement := new(big.Int).Neg(bigInt)
-		twosComplement.Sub(twosComplement, big.NewInt(1))
+		twosComplement.Sub(twosComplement, bigOne)
```
turbolent, line 33: 👍

runtime/interpreter/value.go
```diff
@@ -56,11 +56,9 @@ type typeConformanceResultEntry struct {
 //
 // NOTE: Do not generalize to map[interpreter.Value],
 // as not all values are Go hashable, i.e. this might lead to run-time panics
-//
```
turbolent, line 59: This is unfortunately a behaviour of Go 1.19's gofmt. We should maybe clean this up in a separate PR for the whole codebase

runtime/interpreter/value.go
```diff
@@ -6419,7 +6411,7 @@ func (Int128Value) SetMember(_ *Interpreter, _ func() LocationRange, _ string, _
 }
 
 func (v Int128Value) ToBigEndianBytes() []byte {
-	return SignedBigIntToBigEndianBytes(v.BigInt)
+	return SignedBigIntToSizedBigEndianBytes(v.BigInt, 128/8)
```
turbolent, line 6414: Maybe define constants for the magic numbers. `sema` already defines some related constants e.g see `Int128Type*`.

runtime/tests/interpreter/builtinfunctions_test.go
```diff
@@ -196,24 +196,28 @@ func TestInterpretToBigEndianBytes(t *testing.T) {
...
-			"-200":               {255, 56},
-			"-10000000000000000": {220, 121, 13, 144, 63, 0, 0},
+			"0":                  {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0},
+			"42":                 {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 42},
+			"127":                {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 127},
+			"128":                {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 128},
+			"200":                {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 200},
+			"-1":                 {255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255},
+			"-200":               {255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 56},
+			"-10000000000000000": {255, 255, 255, 255, 255, 255, 255, 255, 255, 220, 121, 13, 144, 63, 0, 0},
```
turbolent, line 206: Good idea to add more test cases! 👍 

Could you please also add similar new cases for the smaller signed integer types, i.e. `Int8|16|32|64`?

values_test.go
turbolent: You can construct those values with `Int256{...}`. `sema` defines min/max values for all types, e.g. you can use `Int256TypeMaxIntBig`

turbolent (commented):
Nice!

values.go
```diff
@@ -970,7 +969,7 @@ func (v UInt128) Big() *big.Int {
 }
 
 func (v UInt128) ToBigEndianBytes() []byte {
-	return interpreter.UnsignedBigIntToBigEndianBytes(v.Value)
+	return interpreter.UnsignedBigIntToSizedBigEndianBytes(v.Value, 128/8)
```
turbolent, line 972: Looks like this was not covered by tests, but should be

values.go
```diff
@@ -1033,7 +1032,7 @@ func (v UInt256) Big() *big.Int {
 }
 
 func (v UInt256) ToBigEndianBytes() []byte {
-	return interpreter.UnsignedBigIntToBigEndianBytes(v.Value)
+	return interpreter.UnsignedBigIntToSizedBigEndianBytes(v.Value, 256/8)
```
turbolent, line 1035: Looks like this was not covered by tests, but should be

runtime/interpreter/big.go
turbolent: No, we don't test those, they're just a fallback

runtime/tests/interpreter/builtinfunctions_test.go
```diff
@@ -364,12 +387,21 @@ func TestInterpretToBigEndianBytes(t *testing.T) {
...
 					interpreter.ByteSliceToByteArrayValue(inter, expected),
-					inter.Globals["result"].GetValue(),
+					result,
 				)
+
+				// ensure that .toBigEndianBytes() is the same size as the source type, if it's fixed-width
+				if hasSize {
+					arrayVal := result.(*interpreter.ArrayValue)
+					arraySize := uint(arrayVal.Count())
+					assert.Equalf(t, size, arraySize, "Expected %s.toBigEndianBytes() to return %d bytes, got %d", ty, size, arraySize)
```
turbolent, line 403: 👌

turbolent (commented):
Looks good, nice work!

The only remaining work left is to revert the formatting changes

runtime/interpreter/value.go
dreamsmasher: https://github.com/golang/go/issues/40028

W O W you can't even change this behaviour

dsainati1 (approved):

turbolent (commented):
Linting is still failing. Once fixed this should be good to go

turbolent (approved):

PR #1920, opened 2022-08-26
Add description to interpreter.Value

turbolent:
______

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -99,6 +99,9 @@ func (NonStorable) ChildStorables() []atree.Storable {
 //
 // To accomplish this, cadence.Value is the "user-facing" hierarchy that is easy to work with:
 // simple Go types that can be used without an interpreter or storage.
+//
+// cadence.Value can be converted to an interpreter.Value by "importing" it with importValue,
+// and interpreter.Value can be "exported" to a cadence.Value with ExportValue.
 type Value interface {
 	atree.Value
 	// Stringer provides `func String() string`
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -89,8 +89,16 @@ func (NonStorable) ChildStorables() []atree.Storable {
 	return nil
 }
 
-// Value
-
+// Value is the Cadence value hierarchy which is heavily tied to the interpreter and persistent storage,
+// and has lots of implementation details.
+//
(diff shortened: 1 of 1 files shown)


dreamsmasher (approved):
Ahh, thank you for adding these to the docs in response to my question earlier!

bluesign:
What is the suggested way to convert from interpreter.Value to cadence.Value?

turbolent:
@bluesign Good question, adding it to the comment.

runtime/interpreter/value.go
```diff
@@ -89,8 +89,16 @@ func (NonStorable) ChildStorables() []atree.Storable {
...
+// Value is the Cadence value hierarchy which is heavily tied to the interpreter and persistent storage,
+// and has lots of implementation details.
+//
+// We do not want to expose those details to users
+// (for example, Cadence is used as a library in flow-go (FVM), in the Flow Go SDK, etc.),
+// because we want to be able to change the API and implementation details;
+// nor do we want to require users to Cadence (the library) to write lots of low-level/boilerplate code (e.g. setting up storage).
+//
+// To accomplish this, cadence.Value is the "user-facing" hierarchy that is easy to work with:
+// simple Go types that can be used without an interpreter or storage.
```
turbolent, line 101:
```suggestion
// simple Go types that can be used without an interpreter or storage.
//
// cadence.Value can be converted to an interpreter.Value by "importing" it with importValue,
// and interpreter.Value can be "exported" to a cadence.Value with ExportValue.
```

bluesign:
@turbolent thanks a lot, it is maybe unrelated or I maybe missing some context, but can we say that eventually everything will be importable and exportable? 

I stuck once on this, though been very long time I am fuzzy on details, some values were not importable etc

dsainati1 (approved):

turbolent:
@bluesign it should be possible to make more value importable or exportable that are currently not yet. 

We flagged some types as not importable/exportable intentionally, just because that requires extra effort, but adding support is a low priority. For some values we decided that we do not want to make them importable/exportable yet because we still want to change details about them.

For some values it is easier to add support for importing/exporting, as their internal representation is simpler, for some it is harder. For others, like functions, it is unlikely that they will ever gain support for importing/exporting, because it is unclear how this could be implemented (how are functions serialized? as source code? what about closures?)

PR #1924, opened 2022-08-26
Export functionality required by debugger

turbolent:
## Description

During the latest refactor, some of the functionality required by the new debugger (https://github.com/onflow/flow-emulator/pull/164) accidentally got hidden.

Export the functionality again, so the debugger PR can get updated and eventually merged.

______


robert-e-davidson3 (approved):

runtime/runtime.go
```diff
@@ -203,17 +205,17 @@ func reportMetric(
 
 // interpreterRuntime is an interpreter-based version of the Flow runtime.
 type interpreterRuntime struct {
-	defaultConfig Config
+	DefaultConfig Config
```
SupunS, line 208: Does the debugger need to write this field? If not, maybe expose this via a getter? Might also help to prevent future accidental changes probably.
Fine with either way though.

SupunS (approved):

runtime/runtime.go
turbolent: Looking at https://github.com/onflow/flow-emulator/pull/164 again: It seems like it's not actually needed? I might have directly modified the config at some point, and then later introduced `SetDebugger` instead 🤔

bluesign:
Does this PR fix latest breaking changes for debugger on fvm ? Basically work with reuseablecadenceruntime stuff?

PR #1927, opened 2022-08-29
Add String.fromUTF8 for conversion from [UInt8]

### LINKED ISSUE #1894: Convert `[UInt8]` byte array to UTF-8 encoded string
### Add Function to convert byte array (`[UInt8]`) to UTF-8 encoded string.

Currently its possible to convert a UTF-8 string to a byte array via `"mystring".utf8`.  However, its not possible to go back from a byte array to a UTF-8 string.

This would also enable the conversion of a hex-encoded string to a utf-8 encoded string.

### Suggested Solution

Add a standard library function to convert `[UInt8]` to a utf8 encoded string.

dreamsmasher:
Closes #1894 

## Description

Adds a static method on the `String` type for attempting to convert byte arrays into Strings:

```cadence
fun fromUTF8(_ input: [UInt8]): String?
```

Currently, the conversion naively loads the entire input into a Go bytearray and runs UTF-8 validation. A better approach might be to stream the bytes from slab storage, and incrementally clustering bytes into codepoints with a state machine approach. This would allow for linear space usage with short-circuiting.
______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -3051,7 +3051,7 @@ var stringFunction = func() Value {
 				memoryUsage := common.NewStringMemoryUsage(len(buf))
 
 				return NewSomeValueNonCopying(
-					inter, 
+					inter,
 					NewStringValue(inter, memoryUsage, func() string {
 						return string(buf)
 					}),
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -81,6 +81,7 @@ type getterSetter struct {
 // are treated like they are returning a value.
 
 // OnEventEmittedFunc is a function that is triggered when an event is emitted by the program.
+//
 type OnEventEmittedFunc func(
 	inter *Interpreter,
 	getLocationRange func() LocationRange,
@@ -89,21 +90,25 @@ type OnEventEmittedFunc func(
 ) error
 
 // OnStatementFunc is a function that is triggered when a statement is about to be executed.
+//
 type OnStatementFunc func(
 	inter *Interpreter,
 	statement ast.Statement,
 )
 
 // OnLoopIterationFunc is a function that is triggered when a loop iteration is about to be executed.
+//
 type OnLoopIterationFunc func(
 	inter *Interpreter,
 	line int,
 )
 
 // OnFunctionInvocationFunc is a function that is triggered when a function is about to be invoked.
+//
 type OnFunctionInvocationFunc func(inter *Interpreter)
 
 // OnInvokedFunctionReturnFunc is a function that is triggered when an invoked function returned.
+//
 type OnInvokedFunctionReturnFunc func(inter *Interpreter)
 
 // OnRecordTraceFunc is a function that records a trace.
@@ -132,6 +137,7 @@ type OnMeterComputationFunc func(
 )
 
 // InjectedCompositeFieldsHandlerFunc is a function that handles storage reads.
+//
 type InjectedCompositeFieldsHandlerFunc func(
 	inter *Interpreter,
 	location common.Location,
@@ -140,6 +146,7 @@ type InjectedCompositeFieldsHandlerFunc func(
 ) map[string]Value
 
 // ContractValueHandlerFunc is a function that handles contract values.
+//
 type ContractValueHandlerFunc func(
 	inter *Interpreter,
 	compositeType *sema.CompositeType,
@@ -148,6 +155,7 @@ type ContractValueHandlerFunc func(
 ) *CompositeValue
 
 // ImportLocationHandlerFunc is a function that handles imports of locations.
+//
 type ImportLocationHandlerFunc func(
 	inter *Interpreter,
 	location common.Location,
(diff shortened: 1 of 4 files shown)


runtime/interpreter/interpreter.go
```diff
@@ -80,7 +81,6 @@ type getterSetter struct {
 // are treated like they are returning a value.
 
 // OnEventEmittedFunc is a function that is triggered when an event is emitted by the program.
-//
```
dsainati1, line 83: Why were these removed?

runtime/interpreter/interpreter.go
```diff
@@ -3026,6 +3000,58 @@ var stringFunction = func() Value {
...
+
+				if utf8.Valid(buf) {
+					return NewSomeValueNonCopying(inter, NewStringValue(inter, memoryUsage, func() string {
+						return string(buf)
+					}))
+				} else {
+					return NewNilValue(inter)
+				}
+			},
+			sema.StringTypeEncodeHexFunctionType,
```
dsainati1, line 3052: Wrong type here

runtime/interpreter/interpreter.go
```diff
@@ -3026,6 +3000,58 @@ var stringFunction = func() Value {
...
+				if !ok {
+					panic(errors.NewUnreachableError())
+				}
+
+				inter := invocation.Interpreter
+
+				// meter byte count, not utf-8 codepoints
+				memoryUsage := common.NewArrayExpressionMemoryUsage(
+					argument.Count(),
+				)
```
dsainati1, line 3017: The byte array was already metered when it was created. What we're doing here is metering the new string we are going to return, so this should be a `NewStringMemoryUsage`, rather than an array expression. Also, so that we don't incorrectly meter a string when we return `nil` from this function; put the usage calculation down in the `if` statement that actually returns the string.

runtime/interpreter/interpreter.go
```diff
@@ -3026,6 +3000,58 @@ var stringFunction = func() Value {
...
+
+				if err != nil {
+					panic(errors.NewExternalError(err))
+				}
+
+				buf := make([]byte, argument.array.Count())
+
+				idx := 0
+				// naively read the entire byte stream into a byte array
+				for {
```
dsainati1, line 3029: I'd use the `iter.Iterate` function here instead of a while loop. If you don't want to, I'd still suggest moving the checks into the loop header like `for val, err := iter.Next(); val != nil; val, err = iter.Next()` instead of having them in the body.

runtime/sema/type.go
```diff
@@ -4854,15 +4862,14 @@ func (t *AddressType) GetMembers() map[string]MemberResolver {
...
-//                        e.g: IsSubType(<<someType>>, FixedPointType)
-//
+//   - IsSubType()      - To check the assignability. e.g: Is argument type T is a sub-type
+//     of parameter type R. This is the more frequent use-case.
+//   - IsSameTypeKind() - To check if a type strictly belongs to a certain category. e.g: Is the
+//     expression type T is any of the integer types, but nothing else.
+//     Another way to check is, asking the question of "if the subType is Never,
+//     should the check still pass?". A common code-smell for potential incorrect
+//     usage is, using IsSubType() method with a constant/pre-defined superType.
+//     e.g: IsSubType(<<someType>>, FixedPointType)
```
dsainati1, line 4872: Why did this get reformatted?

runtime/tests/interpreter/string_test.go
```diff
@@ -172,6 +173,62 @@ func TestInterpretStringEncodeHex(t *testing.T) {
...
+
+	type Testcase struct {
+		expr     string
+		expected any
+	}
+
+	testCases := [...]Testcase{
+		// String.fromUTF(str.utf8) = str
+		{`"omae wa mou shindeiru".utf8`, "omae wa mou shindeiru"},
+		{`"would you still use cadence if i was a worm 🥺😳 👉👈".utf8`, "would you still use cadence if i was a worm 🥺😳 👉👈"},
```
dsainati1, line 187: subjunctive
```suggestion
		{`"would you still use cadence if i were a worm 🥺😳 👉👈".utf8`, "would you still use cadence if i were a worm 🥺😳 👉👈"},
```

dsainati1 (commented):
Nice! Needs a couple minor fixes, but otherwise good job! Also, please revert the formatting changes.

runtime/sema/type.go
dreamsmasher: unscrupulous gofmt on save

runtime/tests/interpreter/string_test.go
dreamsmasher: but the meme uses the indicative

https://knowyourmeme.com/memes/would-you-still-love-me-if-i-was-a-worm

runtime/sema/type.go
turbolent: Since Go 1.19, gofmt also re-formats doc comments: https://tip.golang.org/doc/go1.19#go-doc.

Please revert the changes for now and use Go 1.18 for formatting, and then let's have a separate PR where we migrate to Go 1.19 and re-format the doc strings manually. This automated re-format makes the comments unreadable.

dsainati1 (approved):
Nice!

turbolent:
Linting is still failing. Once fixed this should be good to go (a flaky test failed, so I re-ran the test CI job BTW)

PR #1932, opened 2022-09-02
Optimize type ID generation, remove location ID

turbolent:
______


runtime/common/addresslocation.go
```diff
@@ -80,12 +80,30 @@ func (l AddressLocation) MeteredID(memoryGauge MemoryGauge) LocationID {
...
 func (l AddressLocation) TypeID(memoryGauge MemoryGauge, qualifiedIdentifier string) TypeID {
-	return NewMeteredTypeID(
-		memoryGauge,
-		AddressLocationPrefix,
-		l.Address.Hex(),
-		qualifiedIdentifier,
-	)
+	var i int
+
+	// AddressLocationPrefix '.' hex-encoded address '.' qualifiedIdentifier
```
dreamsmasher, line 85: the TypeID impls for AddressLocation, ScriptLocation, and TransactionLocation seem almost identical save for the prefices and middle fields

dsainati1: Agree it would be nice to refactor these to use a common helper if possible.

turbolent: Good idea, refactored in 36afaf932b06df966f5ac43d8429c6fee4ce1172

dsainati1 (approved):

runtime/common/address.go
```diff
@@ -61,6 +61,12 @@ func (a Address) Hex() string {
 	return fmt.Sprintf("%x", a[:])
 }
 
+// HexInto encodes the address as a hex string representation of the address
+// into the given target byte slice.
+func (a Address) HexInto(target []byte) {
```
robert-e-davidson3, line 66: This naming is less confusing for me:

```suggestion
// ToHex encodes the address as a hex string representation of the address
// into the given target byte slice.
func (a Address) ToHex(target []byte) {
```

robert-e-davidson3 (approved):

runtime/common/addresslocation.go
```diff
@@ -80,10 +80,11 @@ func (l AddressLocation) MeteredID(memoryGauge MemoryGauge) LocationID {
 }
 
 func (l AddressLocation) TypeID(memoryGauge MemoryGauge, qualifiedIdentifier string) TypeID {
-	return NewMeteredTypeID(
+	return idLocationTypeID(
```
SupunS, line 83: Can the old method be replaced with the new one?

Is the new method, and the old one essentially produce the same output? If the old method will not be removed, then maybe add a test case to ensure both produce identical results

turbolent: It couldn't be removed yet, as REPL, string, and identifier locations were still relying on it. 
I optimized those as well in cbbad360e4d20444fe67403f5bad41f25a4bfcd5, and also removed the unnecessary location ID code along the way.

runtime/common/address.go
turbolent: This wasn't actually needed, so I removed it

PR #1933, opened 2022-09-02
Prevent keywords from being used as identifiers

### LINKED ISSUE #1929: Keywords can be used as identifiers in certain situations
### Problem

Keywords like `destroy`, `for`, `while` or `transaction` can be used as identifiers for composite, function and variable declarations. This is unnecessarily confusing, and should not be allowed. 

### Steps to Reproduce

Try the following code in the Playground or the emulator:

```cadence
resource destroy {
   fun transaction() {}
}
let f <- create destroy()
f.transaction()
destroy f
```

Note that this parses and typechecks, but it probably should fail parsing. 

### Suggested Solution

We should add a list of keywords that are not allowed to be identifiers, and when parsing an identifier to be used as the name of a composite, variable or function, we should reject these keywords in the parser.

dreamsmasher:
Closes #1929

## Description

Currently, the parser treats keywords as identifiers and allows them to be used as identifiers. This can lead to confusing situations like (stolen from #1929) 

```cadence
resource destroy {
   fun transaction() {}
}
let f <- create destroy()
f.transaction()
destroy f
```

This PR outlines a list of hard and soft keywords, and modifiers the parser to prevent such cases, erroring out if hard keywords are used in identifier position.

Soft keywords are keywords that only occur in limited contexts, and can be useful as names in some situations. The list is as follows:

- `from`: only used in import statements `import foo from ...`
- `account`: used in access modifiers `access(account) let ...`
- `set`: used in access modifier `pub(set) var globalMutVar = ...`
- `all`: used in access modifier `access(all) let ...`

This is a breaking change that will be part of Stable Cadence.

______


dreamsmasher:
Still need to rebase this against the Stable Cadence branch.

dreamsmasher:
Rebased commits onto branch feature/stable-cadence in #1937. Closing now

PR #1934, opened 2022-09-05
Implement FLIP #1071: borrow contract

bluesign:
Work towards [FLIP #1071](https://github.com/onflow/flow/pull/1071)

## Description

- Docs I will update after confirmed that I am on correct path

______


runtime/sema/authaccount_contracts.go
```diff
@@ -184,6 +191,42 @@ var AuthAccountContractsTypeGetFunctionType = &FunctionType{
 	),
 }
 
+const authAccountContractsTypeBorrowFunctionDocString = `
+Returns a reference in the given type to the contract with the given name in the account, if any.
+
+Returns nil if no contract with the given name exists in the account or not conforming to given type.
```
dsainati1, line 197:
```suggestion
Returns nil if no contract with the given name exists in the account, or if the contract does not conform to the given type.
```

dsainati1 (commented):
Looks good so far; can you add some tests to check that everything works as expected?

turbolent:
@bluesign Updated to master and added tests

turbolent:
Only thing that's left is documentation

runtime/sema/account_contracts.go
```diff
@@ -0,0 +1,88 @@
+/*
```
turbolent, line 1: There was lots of duplication between `AuthAccount` and `PublicAccount` for the common functionality (get, names, borrow), so refactored this as part of this PR, so we don't add to the duplication

turbolent:
FLIP has been approved: https://github.com/onflow/flow/pull/1071

runtime/stdlib/account.go
```diff
@@ -1267,7 +1277,85 @@ func newAccountContractsGetFunction(
...
+			}
+			if len(code) == 0 {
+				return interpreter.Nil
+			}
+
+			// Load the contract
+
+			contractLocation := common.NewAddressLocation(gauge, address, name)
+			subInterpreter := inter.EnsureLoaded(contractLocation)
+			contractValue, err := subInterpreter.GetContractComposite(contractLocation)
```
dsainati1, line 1334: For my own understanding, why do we use the subInterpreter here but not anywhere else later in this function?

bluesign: @dsainati1 I don't know really, I confess I did little cargo cult programming here  :( maybe @turbolent can explain

turbolent: We should use `subInterpreter`, I missed that. Thanks for the flag!

turbolent: Fixed in [e27fe5a](https://github.com/onflow/cadence/pull/1934/commits/e27fe5a8b768249f67650646b44b2f74e73e4c07)

bluesign: @turbolent any chance you can explain the reason? Now I looked up the place I borrowed the code [0], a bit confused. 

[0] https://github.com/onflow/cadence/blob/2926bebcebb6de2a8e3e4ddf148e4ed84967c2d6/runtime/interpreter/interpreter_import.go#L40

SupunS: I guess the reason here is `ConvertStaticToSemaType` (called deep within `IsSubTypeOfSemaType`) uses the elaboration from the interpreter to load types. So if we don't use the imported-interpreter (aka subInterpreter), It will fail to load the type of the imported/loaded contract during runtime subtype validation.

SupunS (approved):
Looks good!

dsainati1 (approved):

turbolent:
@bluesign Could you please open a follow-up PR for documenting this? Thank you!

PR #1937, opened 2022-09-06
Prevent keywords from being used as identifiers (redux)

### LINKED ISSUE #1929: Keywords can be used as identifiers in certain situations

dreamsmasher:
Closes #1929

## Description

Currently, the parser treats keywords as identifiers and allows them to be used as identifiers. This can lead to confusing situations like (stolen from #1929) 

```cadence
resource destroy {
   fun transaction() {}
}
let f <- create destroy()
f.transaction()
destroy f
```

This PR outlines a list of hard and soft keywords, and modifiers the parser to prevent such cases, erroring out if hard keywords are used in identifier position.

Soft keywords are keywords that only occur in limited contexts, and can be useful as names in some situations. The list is as follows:

- `from`: only used in import statements `import foo from ...`
- `account`: used in access modifiers `access(account) let ...`
- `set`: used in access modifier `pub(set) var globalMutVar = ...`
- `all`: used in access modifier `access(all) let ...`

This is a breaking change that will be part of Stable Cadence.

______


runtime/parser/transaction.go
```diff
@@ -100,8 +101,8 @@ func parseTransactionDeclaration(p *parser, docString string) (*ast.TransactionD
 		default:
 			return nil, p.syntaxError(
 				"unexpected identifier, expected keyword %q or %q, got %q",
-				keywordPrepare,
-				keywordExecute,
+				KeywordPrepare,
+				KeywordExecute,
```
turbolent, line 105: Could you please add test cases for the uncovered code along the way?

turbolent (commented):
Nice!

runtime/parser/declaration.go
```diff
@@ -20,12 +20,14 @@ package parser
...
 import (
 	"encoding/hex"
+	"fmt"
 	"strconv"
 	"strings"
 
 	"github.com/onflow/cadence/runtime/ast"
 	"github.com/onflow/cadence/runtime/common"
 	"github.com/onflow/cadence/runtime/errors"
+	. "github.com/onflow/cadence/runtime/parser/constants"
```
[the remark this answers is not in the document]
dreamsmasher: ah this was a holdover from when I was trying to implement this by separating keywords in the lexer. moving the keywords into a common package in `runtime/parser` was the only way i could access them from both the parser and the lexer subpackage. will revert

runtime/parser/constants/keyword.go
```diff
@@ -0,0 +1,124 @@
...
+	KeywordTransaction = "transaction"
+	KeywordPrepare     = "prepare"
+	KeywordExecute     = "execute"
+	KeywordCase        = "case"
+	KeywordSwitch      = "switch"
+	KeywordDefault     = "default"
+	KeywordEnum        = "enum"
+)
+
+var AllKeywords mapset.Set[string] = mapset.NewSet(
```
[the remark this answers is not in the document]
dreamsmasher: ah ok, I was specifically trying to avoid writing a billion `{}`'s and make intentions more clear.

runtime/parser/keyword.go
```diff
@@ -19,47 +19,116 @@
 package parser
 
 const (
-	keywordIf          = "if"
```
dsainati1, line 22: It doesn't seem like these need to be accessible outside of the `parser` module; I'd lowercase prefix all these (as well as the names of the keywords maps) so we don't expose more implementation details than is strictly necessary.

dsainati1 (approved):
LGTM

runtime/parser/declaration.go
```diff
@@ -901,7 +896,13 @@ func parseCompositeOrInterfaceDeclaration(
 			p.next()
 			continue
 		} else {
-			identifier = p.tokenToIdentifier(p.current)
+			ctx := fmt.Sprintf("following %s declaration", compositeKind.Keyword())
+			nonReserved, err := p.nonReservedIdentifier(ctx)
```
turbolent, line 900: This always formats/allocates, even if the error message is not used. Maybe make this a closure that computes the message lazily.

Issue #1949, opened 2022-09-09
[LS] Handling parser exceptions


sideninja:
We need to handle parser panics when parsing invalid programs in a graceful way, not permitting the LS to crash. The errors are currently reported in Sentry. A wrapper to the parser should be put in place recovering from the parser panic.

Example error:
```
runtime/debug.Stack()
	runtime/debug/stack.go:24 +0x65
github.com/onflow/cadence/runtime/errors.NewUnexpectedErrorFromCause(...)
```

Issue #1544, opened 2022-03-31
[LS] Crashing on referencing non-existing imported variable


sideninja:
### Problem
When referencing a variable that doesn't exist in the imported contract the LS crashes with the following stack:

Langauge Server report:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x50 pc=0x10238c724]

goroutine 27 [running]:
github.com/onflow/cadence/runtime/ast.(*CompositeDeclaration).DeclarationMembers(0x0)
	/Users/dapper/go/pkg/mod/github.com/onflow/cadence@v0.23.0/runtime/ast/composite.go:69 +0x4
github.com/onflow/cadence/languageserver/server.(*Server).convertError.func3(0x140005c1b80, 0x0)
	/Users/dapper/go/pkg/mod/github.com/onflow/cadence/languageserver@v0.18.3-0.20220202133308-207188a51831/server/server.go:2208 +0x50
github.com/onflow/cadence/languageserver/server.(*Server).maybeAddDeclarationActionsResolver.func1()
	/Users/dapper/go/pkg/mod/github.com/onflow/cadence/languageserver@v0.18.3-0.20220202133308-207188a51831/server/server.go:2630 +0x328
github.com/onflow/cadence/languageserver/server.(*Server).CodeAction(0x14000204540, {0x103789bd8, 0x14000592c98}, 0x140008539e0)
	/Users/dapper/go/pkg/mod/github.com/onflow/cadence/languageserver@v0.18.3-0.20220202133308-207188a51831/server/server.go:874 +0x244
github.com/onflow/cadence/languageserver/protocol.(*Server).handleCodeAction(0x140003c35c0, 0x14000015b78)
	/Users/dapper/go/pkg/mod/github.com/onflow/cadence/languageserver@v0.18.3-0.20220202133308-207188a51831/protocol/methods.go:103 +0xbc
github.com/onflow/cadence/languageserver/jsonrpc2.(*handler).Handle(0x14000592cb8, {0x1037806e0, 0x1400013a008}, 0x140000d5ef0, 0x1400086f4a0)
	/Users/dapper/go/pkg/mod/github.com/onflow/cadence/languageserver@v0.18.3-0.20220202133308-207188a51831/jsonrpc2/server.go:53 +0x25c
[...]

PR #1950, opened 2022-09-09
[LS] Multiple language server bugfixes

### LINKED ISSUE #1570: [LS] Better cache invalidation for imported program checkers

sideninja:
Closes #1570 #1544 #1949

## Description
This PR addresses multiple issues causing logical or execution bugs. It solves handling the unknown member crash, it resets the caches for imported programs and it wraps parsing to handle panics.

______


languageserver/server/server.go
```diff
@@ -1730,6 +1732,10 @@ func (s *Server) InlayHint(
 
 	for _, variableDeclaration := range variableDeclarations {
 		targetType := checker.Elaboration.VariableDeclarationTargetTypes[variableDeclaration]
+		if targetType == nil { // bugfix getting nil targetType
```
turbolent, line 1735: Skipping is an OK fix, but maybe add a TODO here, this should never occur

languageserver/server/server.go
```diff
@@ -2061,6 +2067,20 @@ func (s *Server) getDiagnosticsForParentError(
 
 // parse parses the given code and returns the resultant program.
 func parse(code, location string, log func(*protocol.LogMessageParams)) (*ast.Program, error) {
+	defer func() {
```
turbolent, line 2070: Nice. I guess we could send a Sentry request here in the future?

languageserver/server/server.go
```diff
@@ -2349,6 +2369,11 @@ func (s *Server) convertError(
 				err.Name,
 				func(checker *sema.Checker, isFunction bool) insertionPosition {
 					declaration := declarationGetter(checker.Elaboration)
+					if declaration == nil {
```
turbolent, line 2372: Is returning an empty insertion position the same as not recommending a code action to add the member?

Maybe also add a TODO to look up the getter in the correct checker (it might be imported one)

languageserver/server/server.go
```diff
@@ -541,6 +540,10 @@ func (s *Server) DidChangeTextDocument(
 		Version: version,
 	}
 
+	// todo implement smarter cache invalidation with dependency resolution using https://github.com/onflow/cadence/pull/1634
+	// we should build dependency tree upfront and then based on the changes only reset checkers contained in that tree
+	s.checkers = make(map[common.LocationID]*sema.Checker)
```
turbolent, line 545: It's cheaper to clear the map than to allocate a new map each time:
```suggestion
	for locationID := range s.checkers {
		delete(s.checkers, locationID)
	}
```

robert-e-davidson3 (approved):

languageserver/server/server.go
sideninja: nice

languageserver/server/server.go
sideninja: will add that in yes

languageserver/server/server.go
sideninja: Do you have a theory why this happens? You think it could be some sort of a race condition using checker? the checker might recheck during the execution of this function. Do you think we can submit stuff to Sentry so we debug this easier?

languageserver/server/server.go
sideninja: Chnaged to sentry recover with context as that is mostly what we want to do here.

languageserver/server/server.go
sideninja: Wilco.
By the way how would we access the correct checker here? we would need to know the location right? but we don't have it. How could I know which checker to get?

sideninja:
@turbolent implemented all the feedback. Let me know if it looks good.

languageserver/server/server.go
turbolent: I don't know off the top of my head, we'll need to look into it.

languageserver/server/server.go
turbolent: Nice!

languageserver/server/server.go
turbolent: The interface type / composite type has a location. We need to look up that location's checker, instead using the checker for the current document where the new/undeclared member is used (`maybeAddDeclarationActionsResolver` incorrectly passes the current document's checker to `memberInsertionPosGetter`). But let's do that in a follow-up PR.

turbolent (approved):
Nice!

SupunS (approved):

PR #1954, opened 2022-09-09
[Test Framework] Separate import resolver and file resolver

SupunS:
## Description
A test script has two things to resolve from the running environment.
- Imports of the test script (not the imports of the code that is being tested)
- Files (for `readFile`).
i.e:
```cadence
import Test
import "SampleContract"  // <-- (1) import needs to be resolved locally, either from a file or from flow.json

pub fun testScript() {
   let content = Test.readFile("./sample/script.cdc")  // <-- (2) file needs to be resolved locally
}
```

Currently, both of these were resolved using one resolver, which limits usability. This PR separates the two into two resolvers.

______


test-framework/test_runner.go
```diff
@@ -67,11 +67,27 @@ type Result struct {
...
+// Must be provided by the user of the TestRunner.
+//
+type FileResolver func(path string) (string, error)
+
 // TestRunner runs tests.
 //
 type TestRunner struct {
+
+	// importResolver is used to resolve imports of the *test script*.
+	// Note: This doesn't resolve the imports for the codes that is being tested.
```
robert-e-davidson3, line 81:
```suggestion
	// Note: This doesn't resolve the imports for the code that is being tested.
```

test-framework/test_runner.go
```diff
@@ -67,11 +67,27 @@ type Result struct {
...
+// Returns the content of the file as a string.
+// Must be provided by the user of the TestRunner.
+//
+type FileResolver func(path string) (string, error)
+
 // TestRunner runs tests.
 //
 type TestRunner struct {
+
+	// importResolver is used to resolve imports of the *test script*.
```
robert-e-davidson3, line 80: The test script here is the Test contract, right? Either way please clarify because I am unsure.

robert-e-davidson3 (approved):

test-framework/test_runner.go
SupunS: the "test script" in this case is the script with the test cases. i.e: the .cdc file you would run as tests.

turbolent (approved):

PR #1961, opened 2022-09-12
Invalidate resource references when the referenced resource is moved

SupunS:
## Description

This is the draft implementation of the FLIP: https://github.com/onflow/flow/pull/1043

The general idea is that, a resource reference is invalidated, if the pointed resource is being moved.

______

### DIFF
--- a/runtime/interpreter/errors.go
+++ b/runtime/interpreter/errors.go
@@ -865,3 +865,18 @@ func (InvalidHexLengthError) IsUserError() {}
 func (InvalidHexLengthError) Error() string {
 	return "hex string has non-even length"
 }
+
+// MovedResourceReferenceError is reported when accessing a reference value
+// that is pointing to a moved resource.
+//
+type MovedResourceReferenceError struct {
+	LocationRange
+}
+
+var _ errors.UserError = MovedResourceReferenceError{}
+
+func (MovedResourceReferenceError) IsUserError() {}
+
+func (e MovedResourceReferenceError) Error() string {
+	return "referring resource is moved"
+}
--- a/runtime/interpreter/errors.go
+++ b/runtime/interpreter/errors.go
@@ -878,5 +878,5 @@ var _ errors.UserError = MovedResourceReferenceError{}
 func (MovedResourceReferenceError) IsUserError() {}
 
 func (e MovedResourceReferenceError) Error() string {
-	return "referring resource is moved"
+	return "referenced resource has been moved after creating the reference"
 }
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -17322,9 +17322,9 @@ type EphemeralReferenceValue struct {
 	Value        Value
 	BorrowedType sema.Type
 
-	// originAddress is the address of the referencing resource value
+	// originAddress is the address of the referenced resource value
 	// at the time of the reference creation.
-	// It is always empty if the referencing value is not a resource.
+	// It is always empty if the referenced value is not a resource.
 	originAddress atree.Address
 }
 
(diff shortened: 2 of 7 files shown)


runtime/interpreter/value.go
```diff
@@ -17556,6 +17577,7 @@ func (v *EphemeralReferenceValue) RemoveKey(
 	self := *referencedValue
 
 	interpreter.checkReferencedResourceNotDestroyed(self, getLocationRange)
+	v.checkReferenceResourceNotMoved(self, getLocationRange)
```
dsainati1, line 17580: These seem to occur together all the time; can we refactor this to be one function call that does both checks?

SupunS: Yeah, should be possible. I haven't migrated the existing tests, so wasn't sure what other places need updates, etc. So kept it separated for now. But I guess we can merge the two where ever possible.

runtime/interpreter/errors.go
```diff
@@ -865,3 +865,18 @@ func (InvalidHexLengthError) IsUserError() {}
...
+type MovedResourceReferenceError struct {
+	LocationRange
+}
+
+var _ errors.UserError = MovedResourceReferenceError{}
+
+func (MovedResourceReferenceError) IsUserError() {}
+
+func (e MovedResourceReferenceError) Error() string {
+	return "referring resource is moved"
```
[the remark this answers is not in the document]
SupunS: That's a good idea 👍 I'll see if we can point to the location where it was moved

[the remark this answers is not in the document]
SupunS: I had a look into this, and it seems very difficult to keep track of the 'move' operation that causes the invalidation.
e.g: A single resource can have multiple references and multiple move operations. Also, the move operations are not aware of the existing references (we don't maintain inward references to the resource). So if we are to add this, need to maintain such mapping separately, which can be costly. That kind of bookkeeping is a bit too heavy for an error message improvement.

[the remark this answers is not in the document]
SupunS: improved the error message slightly https://github.com/onflow/cadence/pull/1961/commits/a6250a67cbcd9be89e5a7de0d21e9c2293aaa547

runtime/tests/interpreter/reference_test.go
```diff
@@ -472,23 +472,17 @@ func TestInterpretResourceReferenceAfterMove(t *testing.T) {
...
 		)
 
-		arrayRef := &interpreter.EphemeralReferenceValue{
-			Authorized: false,
-			Value:      array,
-			BorrowedType: &sema.VariableSizedType{
+		arrayRef := interpreter.NewUnmeteredEphemeralReferenceValue(
+			false,
+			array,
+			&sema.VariableSizedType{
```
dreamsmasher, line 478: maybe keep the labels here, although it's just a test case
```suggestion
		arrayRef := interpreter.NewUnmeteredEphemeralReferenceValue(
			Authorized:   false,
			Value:        array,
			BorrowedType: &sema.VariableSizedType{
```

dreamsmasher (approved):

runtime/tests/interpreter/reference_test.go
SupunS: This was changed from the struct-literal to a function call (a constructor). So the labels are gone now.

SupunS:
The behaviour of the implementation was slightly modified to invalidate the resource immediately upon a move, as described in https://github.com/onflow/flow/pull/1043#issuecomment-1247394376. I'll open a separate PR with this modified behaviour. Hence closing this one.

SupunS:
New PR: https://github.com/onflow/cadence/pull/1999

PR #1982, opened 2022-09-16
Add fromString method on numeric types

### LINKED ISSUE #1918: Parsing String into UFix64 or any other numberable type

dreamsmasher:
Closes #1918

## Description

Adds a static method to numeric types [Int{8,16,32,64,128,256}, UInt{8,16,32,64,128,256}, Word{8,16,32,64}, Fix64, UFix64]:

```
fun fromString(_ input: String): T? 
```

that attempts to parse the string into the target type, failing if the input string is an invalid serialization or if the result is out of bounds.

This PR only pertains to decimal representations, although depending on demand, it might be a good idea to accept hex, octal, and binary formats with numeric prefices.

______


runtime/sema/type.go
```diff
@@ -361,6 +361,46 @@ const toStringFunctionDocString = `
...
+	}
+
+	builder := new(strings.Builder)
+	builder.WriteString(
+		fmt.Sprintf(
+			"Attempts to parse %s from a string. Returns `nil` on overflow or invalid input. Whitespace or invalid digits will return a nil value.\n",
+			ty.String(),
+		))
+
+	if IsSameTypeKind(ty, FixedPointType) {
```
dreamsmasher, line 388: I'm taking a leap here and assuming that `IsSameTypeKind` is the correct check, instead of `IsSubType`.

runtime/interpreter/interpreter.go
```diff
@@ -2214,8 +2217,182 @@ func (interpreter *Interpreter) writeStored(
...
+	return FromStringFunctionValue{
+		parentType:   ty,
+		hostFunction: hostFunctionImpl,
+	}
+}
+
+// default implementation for parsing a given unsigned numeric type from a string.
+// the size provided by sizeInBytes is passed to strconv.ParseUint, ensuring that the parsed value fits in the target type.
+// input strings must not begin with a '+' or '-'.
+func fromStringImplUnsigned(bitSize int, convert func(uint64) Value) func(*Interpreter, string) Value {
```
SupunS, line 2249: `func(*Interpreter, string) Value` is used in multiple places, maybe define a separate type for it.

Also given that these are all returning numeric types, can use the `NumberValue` as the return type. (or should be `OptionalValue`?)

runtime/interpreter/interpreter.go
```diff
@@ -2214,8 +2217,182 @@ func (interpreter *Interpreter) writeStored(
...
+			return NewUnmeteredWord8Value(uint8(n))
+		})),
+		fromStringDeclaration(sema.Word16Type, fromStringImplUnsigned(16, func(n uint64) Value {
+			return NewUnmeteredWord16Value(uint16(n))
+		})),
+		fromStringDeclaration(sema.Word32Type, fromStringImplUnsigned(32, func(n uint64) Value {
+			return NewUnmeteredWord32Value(uint32(n))
+		})),
+		fromStringDeclaration(sema.Word64Type, fromStringImplUnsigned(64, func(n uint64) Value {
+			return NewUnmeteredWord64Value(n)
```
SupunS, line 2363: The return type for the `fromString()` methods is always `OptionalType`. Therefore, all these methods should return an `OptionalValue` (i.e: wrap with `SomeValue`)

runtime/interpreter/interpreter.go
```diff
@@ -2214,8 +2217,182 @@ func (interpreter *Interpreter) writeStored(
...
+			return NewUnmeteredInt8Value(int8(n))
+		})),
+		fromStringDeclaration(sema.Int16Type, fromStringImplSigned(16, func(n int64) Value {
+			return NewUnmeteredInt16Value(int16(n))
+		})),
+		fromStringDeclaration(sema.Int32Type, fromStringImplSigned(32, func(n int64) Value {
+			return NewUnmeteredInt32Value(int32(n))
+		})),
+		fromStringDeclaration(sema.Int64Type, fromStringImplSigned(64, func(n int64) Value {
+			return NewUnmeteredInt64Value(n)
```
SupunS, line 2317: All these methods should use the 'metered' version of the constructors. Because one can create a lot of numbers from its string representation inside a loop, and those needs to be metered.

runtime/sema/type.go
```diff
@@ -361,6 +361,46 @@ const toStringFunctionDocString = `
...
+			{
+				Label:          ArgumentLabelNotRequired,
+				Identifier:     "input",
+				TypeAnnotation: NewTypeAnnotation(StringType),
+			},
+		},
+		ReturnTypeAnnotation: NewTypeAnnotation(
+			&OptionalType{ty},
+		),
+	}
```
SupunS, line 379: Maybe split the type creation and the doc-comment creation into two methods. Currently, it seems to do two independent tasks..

runtime/tests/interpreter/integers_test.go
```diff
@@ -882,3 +883,53 @@ func TestInterpretIntegerMinMax(t *testing.T) {
...
+			high = big.NewInt(math.MaxInt64)
+		}
+
+		typeName := typ.String()
+
+		t.Run(typeName, func(t *testing.T) {
+
+			code := fmt.Sprintf(`
+				fun testFromString(_ input: String): Int? {
+					return %s.fromString(input).map(Int)
```
SupunS, line 906: Maybe remove the `.map(Int)` function, to reduce the number of things we test in a single test-case. Or make it (calling `map` on the returned value) a separate test.
```suggestion
					return %s.fromString(input)
```

runtime/tests/utils/utils.go
```diff
@@ -192,3 +192,6 @@ func AssertValueSlicesEqual(t testing.TB, inter *interpreter.Interpreter, expect
 
 	return true
 }
+
+// Noop statement for placing a breakpoint during debugging.
+func Breakpoint() {}
```
SupunS, line 197: Is this change intended?

dreamsmasher: ah yea, just a useful oneliner that wouldn't make for much of a PR on its own. i can remove it

runtime/interpreter/interpreter.go
```diff
@@ -2214,8 +2217,182 @@ func (interpreter *Interpreter) writeStored(
...
+
+		converted := convert(val)
+		common.UseMemory(interpreter, expectedUsage)
+		return NewSomeValueNonCopying(interpreter, converted)
+	}
+}
+
+func fromStringImplBig(convert func(*big.Int) (Value, bool)) func(*Interpreter, string) Value {
+	return func(interpreter *Interpreter, input string) Value {
+		val, ok := common.NewBigInt(interpreter).SetString(input, 10)
```
turbolent, line 2284: Creating the big int allocates, so needs to be metered before

runtime/tests/interpreter/fixedpoint_test.go
```diff
@@ -592,3 +594,115 @@ func TestInterpretFixedPointMinMax(t *testing.T) {
 		})
 	}
 }
+
+func TestStringFixedpointConversion(t *testing.T) {
+	type testcase struct {
```
turbolent, line 599:
```suggestion
func TestStringFixedpointConversion(t *testing.T) {

	t.Parallel()

	type testcase struct {
```

runtime/tests/interpreter/fixedpoint_test.go
```diff
@@ -592,3 +594,115 @@ func TestInterpretFixedPointMinMax(t *testing.T) {
...
+		for m = 0; cloned.Cmp(bigZero) != 0; m++ {
+			cloned.Div(cloned, bigTen)
+		}
+
+		return m
+	}
+
+	for _, testsuite := range suites {
+		t.Run(testsuite.name, func(t *testing.T) {
+			t.Parallel()
```
turbolent, line 672: Subtle bug here: The `for`-`range` loop *rebinds* `testsuite` to each suite, so by the time the sub-test closure is run, it might already have a different value assigned to it. 

You can fix this by running the tests sequentially, by removing the t.Parallel call; or by rebinding `testsuite` – the common way in the code base is to refactor the loop-body into a local helper function:

```go
test := func (testsuite testsuite) {
	t.Run(testsuite.name, func(t *testing.T) {
		t.Parallel()
		// ...
	})
}

for _, testsuite := range suites {
	test(testsuite)
}
```

runtime/tests/interpreter/integers_test.go
```diff
@@ -882,3 +883,53 @@ func TestInterpretIntegerMinMax(t *testing.T) {
 		})
 	}
 }
+
+func TestStringIntegerConversion(t *testing.T) {
+
```
turbolent, line 888:
```suggestion

		t.Parallel()

```

runtime/interpreter/interpreter.go
```diff
@@ -2475,9 +2654,16 @@ func init() {
 			continue
 		}
 
-		if _, ok := converterNames[numberType.String()]; !ok {
+		// todo use TypeID's here?
+		typeName := numberType.String()
```
turbolent, line 2658: The number type's type IDs and string representation are the same, so there isn't really any difference

runtime/interpreter/interpreter.go
```diff
@@ -2226,13 +2403,15 @@ type ValueConverterDeclaration struct {
 var ConverterDeclarations = []ValueConverterDeclaration{
 	{
 		name:         sema.IntTypeName,
+		targetType:   sema.IntType.Tag(),
```
turbolent, line 2406: Why do only `Int` and `UInt` need the `targetType`, why is it not needed for the other types?

turbolent (commented):
Nice

runtime/interpreter/interpreter.go
dreamsmasher: they're wrapped in the `fromStringImpl-` series of functions, these are just converters to avoid duplicated code

runtime/interpreter/interpreter.go
dreamsmasher: the metering is also done in the fromString- wrapper functions to avoid redundancy

runtime/interpreter/interpreter.go
dreamsmasher: Could you elaborate a bit more? It seems that `common.NewBigInt` meters an uninitialized bigInt, and `SetString` returns nil on failure. Should we meter the state of `val` after the parsing step, regardless of success?

The docs for `setString` also state that "the value of z is undefined" if parsing fails.

runtime/interpreter/interpreter.go
dreamsmasher: whoops dead code

runtime/tests/interpreter/fixedpoint_test.go
dreamsmasher: what in the actual python

runtime/interpreter/interpreter.go
turbolent: Metering, like memory metering or computation metering, must always occur *before* the operation, so that the metering to prevent the future operation if it is determined that the limit would be exceeded.

Basically, move the `common.UseMemory` up before this line and over-estimate the amount of memory based on the length of the string.

runtime/interpreter/interpreter.go
```diff
@@ -2214,6 +2217,190 @@ func (interpreter *Interpreter) writeStored(
...
+	return func(interpreter *Interpreter, input string) OptionalValue {
+		// maxVal =~ 10^|input|
+		// bytes = ceil(log256(maxVal))
+		// = ceil(log256(10 ^ |input|)) via substitution
+		// = ceil(|input| * log256(10)) via power rule
+		// = ceil(|input| * 0.415241011)
+		// there's gotta be a better way of estimating memory usage lol
+
+		estimatedSize := int(math.Ceil(float64(len(input)) * 0.415241011))
+		common.UseMemory(interpreter, common.NewCadenceBigIntMemoryUsage(estimatedSize))
```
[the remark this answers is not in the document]
dsainati1: Memory usage is deliberately an overestimate. If you're multiplying by a decimal value you can just round to the nearest whole number for convenience. In particular, I think it's fine here to just use `len(input)` as the estimated size.

runtime/tests/interpreter/integers_test.go
dreamsmasher: hmm, my reason for converting everything to an Int was to reduce the amount of manual conversions for expected values. since Int has an unbounded domain, this approach only needs to specify low and high values.

runtime/interpreter/interpreter.go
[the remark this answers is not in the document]
turbolent: Agreed, we can just overestimate, we do not need to be accurate, as long as we do not underestimate.

One problem might be that the function does not just allocate for the result, but performs additional temporary allocations. Looking at `Int.SetString` it indeed seems to gradually allocate, but I'm not sure. 
@tarakby Could you maybe help us here? 🙏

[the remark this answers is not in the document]
dreamsmasher: i'm just gonna do an integer division by 2, any objections?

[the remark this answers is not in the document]
tarakby: Some comments:

- I think `bytes = ceil(log256(maxVal))` doesn't cover all cases. Think of the case when `maxVal` is a power of 256. An easier example considering base `2` instead of base `256` : when `maxVal = 8`, the formula above gives 3 while the number of bits should be 4. 
A formula that works in all cases is `bytes = floor(log256(maxVal) + 1)`
- `math/big`  works in `word` instead of bytes. Words in our case are 64 bits (and that would be an overestimate in case we switch to a 32 bits cpu). So we would need to work in base 2^64 instead of 2^8. 
- as @Bastian mentioned, `math/big` over-allocates 4 words more than the size it needs. 
-  @dreamsmasher I agree, divisions by 2 or power of 2s are a good way to avoid float operations. We can use a similar trick and we can tighten the overestimate a bit more (to do better than 0.41 ~ 0.5)

I would suggest :
```
// maxVal = 10^l    , with l = |input|
// words = 4 + floor(log_(2^64)(maxVal) + 1)
//             = 4 +  floor(log_(2^64)(10^l) + 1) via substitution
//             = 4 +  floor(l * C) + 1   , with C = log_(2^64)(10)
//   over estimating C:  C = 0.05190512..  <= 213/4096
//. words = 5 + floor((l*213)/2^12))
//  bytes = words * 4

estimatedSize := 4 * (5+ ((len(input)*213) >> 12)
```

Please check the steps in case I missed something :)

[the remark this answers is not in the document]
dreamsmasher: seems about right, I love the degree of bikeshedding in this thread.

Seems you can optimize the steps a bit more:
```
estimatedSize 
  = 4 * (5 + ((l * 213) >> 12))
  = 20 + (4 * ((l * 213) >> 12))
  = 20 + (((l * 213) >> 12) << 2)
  = 20 + ((l * 213) >> 10)
```

https://imgur.com/a/ItClhVp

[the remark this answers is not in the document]
turbolent: @tarakby Thank you!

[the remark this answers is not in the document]
dreamsmasher: Guess this should go into OverEstimateBigIntFromString? https://github.com/onflow/cadence/blob/7a111462643d4e2f796704c2fcb8725249d33c1a/runtime/common/bigInt.go#L47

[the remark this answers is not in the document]
turbolent: @dreamsmasher We already have similar formulas in https://github.com/onflow/cadence/blob/master/runtime/common/metering.go, so that's a good spot to put it, along with a comment explaining the magic numbers

runtime/interpreter/interpreter.go
SupunS: Oh, I see. 
The 'metered' constructor guarantees consistency. i.e: all IntX types are metered in one place. So if we have to update at some point, then only have to update one place / wont miss out to update other places.

Also, the fromString- wrapper first creates the value - and then is metered. The constructor actually does it the other way around (so the limit is never exceeded).

runtime/tests/utils/utils.go
SupunS: Is this for go-debugging?
Given that is not used anywhere - and sort of not related to the objective of the PR, would be better to remove it from the PR.

runtime/interpreter/interpreter.go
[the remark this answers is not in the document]
turbolent: I was so sure we already had this function, but couldn't find it, thank you for finding it.

So we already have this estimation function, how does the new one differ? Can we just use the existing one?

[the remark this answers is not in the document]
dreamsmasher: it breaks a bunch of tests, I'll just use that function for now and then open a separate PR for the new bigInt estimation algo and tests.

runtime/interpreter/interpreter.go
```diff
@@ -2214,6 +2217,190 @@ func (interpreter *Interpreter) writeStored(
...
+		common.UseMemory(interpreter, expectedUsage)
+		return NewSomeValueNonCopying(interpreter, converted)
+	}
+}
+
+// default implementation for parsing a given signed numeric type from a string.
+// the size provided by sizeInBytes is passed to strconv.ParseUint, ensuring that the parsed value fits in the target type.
+// input strings may begin with a '+' or '-'.
+func signedIntValueParser(bitSize int, convert func(int64) Value) valueParser {
+	expectedUsage := common.NewCadenceBigIntMemoryUsage(bitSize / 8)
```
SupunS, line 2271: `NewCadenceBigIntMemoryUsage` (all metering methods/values with `Cadence` prefix) are used for Cadence external values (interpreter values uses a similar set counterparts, without that prefix)

-- sort of exactly why using the 'metered' constructor is safe 😉

runtime/interpreter/interpreter.go
SupunS: ah right, didn't notice that one 👍

runtime/interpreter/interpreter.go
dreamsmasher: after a LOT of generics abuse, i've managed to fix things to use metered constructors

runtime/interpreter/interpreter.go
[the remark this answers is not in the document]
turbolent: Agreed, use the existing function for this PR and then we can potentially open another PR to improve it. 
The discussion here is not bikeshedding, it's just solving the problem correctly, which involves figuring our how to meter memory usage, which is unfortunately not straight forward (well, we had already figured it out in the past, but forgot :D, then again it seems to have lead to an improvement, haha)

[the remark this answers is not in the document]
dreamsmasher: hahaha, I'm calling it bikeshedding because everyone got nerdsniped by this problem. I asked on twitter a few days ago and got like 30 responses too.

runtime/interpreter/interpreter.go
```diff
@@ -2214,6 +2217,188 @@ func (interpreter *Interpreter) writeStored(
 	interpreter.recordStorageMutation()
 }
 
+type fromStringFunctionValue struct {
+	receiverType sema.Type
+	hostFunction *HostFunctionValue
+}
+
+// a function that attempts to create a Cadence value from a string, e.g. parsing a number from a string
+type valueParser func(*Interpreter, string) OptionalValue
```
turbolent, line 2226: Maybe make the naming more specific, e.g. `stringValueParser`

runtime/interpreter/interpreter.go
```diff
@@ -2214,6 +2217,188 @@ func (interpreter *Interpreter) writeStored(
...
+			return fromInt64(val)
+		})
+		return NewSomeValueNonCopying(interpreter, converted)
+	}
+}
+
+// No need to use metered constructors for values represented by big.Ints, since estimation is more granular than fixed-size types.
+func bigIntValueParser(convert func(*big.Int) (Value, bool)) valueParser {
+	return func(interpreter *Interpreter, input string) OptionalValue {
+		estimatedSize := common.OverEstimateBigIntFromString(input)
```
turbolent, line 2295: 👍

runtime/interpreter/interpreter.go
```diff
@@ -2214,6 +2217,188 @@ func (interpreter *Interpreter) writeStored(
...
+var fromStringFunctionValues = func() map[string]fromStringFunctionValue {
+	u64_8 := func(n uint64) uint8 { return uint8(n) }
+	u64_16 := func(n uint64) uint16 { return uint16(n) }
+	u64_32 := func(n uint64) uint32 { return uint32(n) }
+	u64_64 := identity[uint64]
+
+	declarations := []fromStringFunctionValue{
+		// signed int values from 8 bit -> infinity
+		newFromStringFunction(sema.Int8Type, signedIntValueParser(8, NewInt8Value, func(n int64) int8 {
+			return int8(n)
```
turbolent, line 2328: Is this `int64` -> `int8` conversion here and the conversions below truncating? Do these functions need range checks?

runtime/interpreter/interpreter.go
```diff
@@ -2214,6 +2217,188 @@ func (interpreter *Interpreter) writeStored(
...
+		newFromStringFunction(sema.Int16Type, signedIntValueParser(16, NewInt16Value, func(n int64) int16 {
+			return int16(n)
+		})),
+		newFromStringFunction(sema.Int32Type, signedIntValueParser(32, NewInt32Value, func(n int64) int32 {
+			return int32(n)
+		})),
+		newFromStringFunction(sema.Int64Type, signedIntValueParser(64, NewInt64Value, identity[int64])),
+		newFromStringFunction(sema.Int128Type, bigIntValueParser(func(b *big.Int) (Value, bool) {
+			ok := inRange(b, sema.Int128TypeMinIntBig, sema.Int128TypeMaxIntBig)
+			return Int128Value{b}, ok
```
turbolent, line 2339: Use the constructor functions instead of struct literals, e.g. here `NewUnmeteredInt128Value`

runtime/interpreter/interpreter.go
dreamsmasher: the range checks are handled by ParseInt/ParseUint, hence the bitsize parameter. I should make that more clear in the comments

runtime/interpreter/interpreter.go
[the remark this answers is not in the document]
tarakby: I didn't know we had the "old" formula already. The old and new formulas seem pretty much equivalent.  They are just 2 methods of the same reasoning (one uses base 2^64 and the other base 2). I don't think the new one is an improvement (unless you see big result divergence). It is fine to just keep the old formula and not update it.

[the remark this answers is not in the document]
turbolent: Great, thank you for the analysis @tarakby

runtime/interpreter/interpreter.go
```diff
@@ -2334,33 +2334,41 @@ var fromStringFunctionValues = func() map[string]fromStringFunctionValue {
 			return int32(n)
 		})),
 		newFromStringFunction(sema.Int64Type, signedIntValueParser(64, NewInt64Value, identity[int64])),
-		newFromStringFunction(sema.Int128Type, bigIntValueParser(func(b *big.Int) (Value, bool) {
-			ok := inRange(b, sema.Int128TypeMinIntBig, sema.Int128TypeMaxIntBig)
-			return Int128Value{b}, ok
+		newFromStringFunction(sema.Int128Type, bigIntValueParser(func(b *big.Int) (v Value, ok bool) {
+			if ok = inRange(b, sema.Int128TypeMinIntBig, sema.Int128TypeMaxIntBig); ok {
+				v = NewUnmeteredInt128ValueFromBigInt(b)
+			}
```
turbolent, line 2340: Good idea to only conditionally perform the creation 👍

turbolent (approved):
Nice work!

SupunS (approved):

PR #1983, opened 2022-09-19
Publish and Claim: Inbox API for Capability Bootstrapping

### LINKED ISSUE #1951: API for bootstrapping Capabilities
### Issue To Be Solved

After https://github.com/onflow/flow/pull/945 was rejected, we still lack a clean solution for the use case of bootstrapping Capabilities (which is outlined in a comment on that FLIP). We'd like a dedicated API for creating a capability (or for a general value) and giving it to a specific user in a secure and asynchronous manner. 

The original comment (from @dete) is replicated here:

Here is the problematic use case. I'll describe possible solutions (including one using an Identity object) afterwards.

I have a project that requires some amount of human oversight, maybe something as simple as a "pause" button that lets me block activity in the smart contract temporarily in response to a bug or market panic. As such, I have an administrator capability in my account that gives me access to call the "pause" function if necessary.

In time, Bjarte joins my team, and I decide that it'd be useful if his account could also call the "pause" function in case I'm sleeping when the emergency happens. I'd like to send him a Capability for the same Admin object that I have. How do we get the Capability into his account securely?

One possible solution: Co-signing

One solution (which is possible in Cadence today) is for Bjarte and I to both sign a single transaction. That transaction can create/copy the Capability using my account, and then store it in his account.

This solution is workable, but awkward. In particular, it has a coordination problem. Both Bjarte and I have to be online at more-or-less that same time in order to both sign the shared transaction within the expiry window.

Another solution: The Dropbox

Another answer would be for me to create the Capability for Bjarte whenever it suits me and then "leave it somewhere he can get it at his convenience". But where do I leave it that doesn't allow a stranger to intervene?

Ideally, I'd have a public function on my account that returns that Capability, but only if the caller provides Bjarte's Identity.

To be clear: I don't love this use case, because once the Identity object exists people might start using it in other cases, and if it's used frequently, in a variety of ways, it could open the window for attacks. For example, perhaps a malicious actor notices that I've posted a Capability for Bjarte before Bjarte himself notices it. That attacker could offer him "1000 free FLOW!" using a different Dropbox object that asks for his Identity. But instead of returning the 1000F, the malicious Dropbox object uses Bjarte's Identity to grab the Admin Capability I was offering and transfer it to the attacker!

The problem with an Identity object in a capability-based security system like Cadence that sharing the Identity object is transitive. I'm not just proving my identity to a smart contract when I pass it my Identity, I'm temporarily allowing it to impersonate me!

We'd really need wallets to carefully restrict the use of the Identity object to scenarios where they explicitly check the type of the Resource they are passing the Identity to in the transaction itself, and ensuring that the type they see is on a short list of known-safe Dropbox implementations (ideally one!).

Unfortunately, I don't really have a better solution for bootstrapping a new Capability setup that doesn't suffer from the coordination problem of double-signing. I'd love to hear other suggestions!

dsainati1:
Closes #1951

This is an implementation of the Publish/Claim API described in this FLIP: https://github.com/onflow/flow/pull/1122

______


runtime/interpreter/value.go
```diff
@@ -16486,6 +16486,8 @@ func (v *DictionaryValue) IsResourceKinded(interpreter *Interpreter) bool {
 type OptionalValue interface {
 	Value
 	isOptionalValue()
+	iter(f func(Value))
+	fmap(inter *Interpreter, f func(Value) Value) OptionalValue
```
dreamsmasher, line 16490: now we gotta extend optionals all the way with `<*>`, `>>=`, `<|>`, and `<>`

dreamsmasher (commented):
LGTM, what's currently missing compared to a full PR?

runtime/interpreter/value.go
dsainati1: I would love to make `OptionalValue` generic in its `Value` (i.e. something like `type OptionalValue[T any] interface { ...` so that less downcasting is necessary in cases like these, but it seems like a lot of restructuring of code in order to make Go not complain about circular references in the visitor.

runtime/account_test.go
```diff
@@ -660,7 +660,7 @@ func TestRuntimeHashAlgorithm(t *testing.T) {
 
 	require.Len(t, array.Values, 3)
 
-	// Check key1
+	// `Check `key1
```
dsainati1, line 663:
```suggestion
	// Check key1
```

dsainati1:
I've added some new documentation here about the new Inbox API, but https://developers.flow.com/cadence/design-patterns#capability-receiver also needs to be updated to suggest using this API instead of `CapabilityReceiver`s, and the linked example contracts should be rewritten to use the new API as well.

runtime/stdlib/account.go
```diff
@@ -791,6 +800,245 @@ func newPublicAccountContractsValue(
...
+			inter.WriteStored(address, inboxStorageDomain, valuePath(nameValue.Str), nil)
+			inter.WriteStored(address, inboxStorageDomain, recipientPath(nameValue.Str), nil)
+
+			publishedValue = publishedValue.Transfer(
+				inter,
+				getLocationRange,
+				atree.Address(address),
+				true,
+				nil,
+			)
```
turbolent, line 937: This order of the operations, and the transfer isn't quite right. You can follow how `load` is implemented, i.e. what `authAccountReadFunction` does, see the comment there: first transfer to the stack (empty address), then write nil to storage.

runtime/stdlib/account.go
```diff
@@ -791,6 +800,245 @@ func newPublicAccountContractsValue(
...
+				getLocationRange,
+			)
+
+			publishedValue = publishedValue.Transfer(
+				inter,
+				getLocationRange,
+				atree.Address(address),
+				true,
+				nil,
+			)
```
turbolent, line 1019: Same as above: first transfer from account to to stack (empty) address, then write nil t storage.

turbolent (commented):
Great work! Looks good, just a few problems in the implementation of the storage functions

docs/language/accounts.mdx
```diff
@@ -321,6 +340,68 @@ transaction() {
...
+If the borrow type of the Capability is not a subtype of the provided type argument, the function will error at runtime. 
+
+Upon successful completion of the `claim` function, the claimed Capability is removed from the Provider's inbox. 
+Note that this means a given Capability can only be claimed once. 
+
+Calling this function emits an event, `InboxValueRemoved`, 
+that includes the address of both the Provider and the Recipient, as well as the name of the claimed Capability.
+Refer to the [`Core Events` section](core-events#inbox-value-removed) for more details on this event.
+
+#### Unpublishing a Capability
```
dsainati1, line 386:
```suggestion
### Unpublishing a Capability
```

docs/language/accounts.mdx
```diff
@@ -321,6 +340,68 @@ transaction() {
...
+
+This publishes the specified Capability using the provided string as an identifier, to be later claimed by the Recipient.
+Note, however, that until the Recipient does claim this Capability, it is stored on the Provider's account, 
+and contributes towards their Account Storage total. 
+
+Calling this function emits an event, `InboxValuePublished`, 
+that includes the address of both the Provider and the Recipient, as well as the name and the type of the published Capability.
+Refer to the [`Core Events` section](core-events#inbox-value-published) for more details on this event.
+
+#### Claiming a Capability
```
dsainati1, line 365:
```suggestion
### Claiming a Capability
```

docs/language/core-events.md
```diff
@@ -121,3 +121,51 @@ pub event AccountContractRemoved(
...
+
+To reduce the potential for spam, 
+we recommend that user agents that display events do not display this event as-is to their users, 
+and allow users to restrict whom they see events from. 
+
+### Inbox Value Removed
+
+Event that is emitted when a Capability is removed from an account. 
+
+Event name: `flow.InboxValueRemoved`
```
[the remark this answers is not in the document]
dsainati1: I don't really think there is any benefit to separating these events. In fact, I don't actually know whether the `remover` parameter is even necessary here. If we think about who would actually care about the `Removed` event, it would only be the original publisher and the intended recipient who have any reason to actually listen for this event. Because the value can only be removed if the publisher calls `unpublish` or the recipient calls `claim`, when the value is removed both parties know for certain who was responsible, since they either just submitted a transaction that resulted in the event being removed (in which case it was them), or they didn't (in which case it must have been the other party).

runtime/interpreter/primitivestatictype.go
```diff
@@ -189,8 +189,8 @@ const (
 	_
 	_
 	_
-	_
-	_
+	PrimitiveStaticTypePublicAccountInbox
+	PrimitiveStaticTypeAuthAccountInbox
```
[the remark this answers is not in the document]
dsainati1: Moving these to the end breaks a bunch of tests for reasons that are unclear to me; it seems as though this was designed to only be modified by replacing the blanks with new types.

runtime/stdlib/account.go
```diff
@@ -791,6 +800,245 @@ func newPublicAccountContractsValue(
...
+			if publishedValue == nil {
+				return interpreter.NewNilValue(gauge)
+			}
+
+			// compare the intended recipient with the caller
+			intendedRecipient, ok := inter.ReadStored(providerAddress, inboxStorageDomain, recipientPath(nameValue.Str)).(interpreter.AddressValue)
+			if !ok {
+				panic(errors.NewUnreachableError())
+			}
+			if intendedRecipient.ToAddress() != recipientValue.ToAddress() {
```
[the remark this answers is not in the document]
dsainati1: Nope! Removing these.

runtime/interpreter/value.go
```diff
@@ -18594,6 +18612,138 @@ func (v LinkValue) ChildStorables() []atree.Storable {
 	}
 }
 
+// PublishedValue
+
+type PublishedValue struct {
+	Recipient AddressValue
+	Value     *CapabilityValue
```
turbolent, line 18619: This implementation should work fore now, given that the published value is jut a Capability. For arbitrary values, which this might get refactored to in the future, the storage-related parts of the value are not quite complete e.g. transfer, deep removal, etc. are not supported.

Maybe add a TODO/NOTE warning above the`Value` field, that once the field is changed to type `Value`, `PublishedValue` needs to be extended and the remaining storage-related code must be implement.

runtime/stdlib/account.go
```diff
@@ -791,6 +800,236 @@ func newPublicAccountContractsValue(
...
+					providerValue,
+					nameValue,
+				},
+				getLocationRange,
+			)
+
+			value := publishedValue.Value.Transfer(
+				inter,
+				getLocationRange,
+				atree.Address(address),
```
turbolent, line 925: Wrong address, value is transferred from storage to the stack:

```suggestion
				atree.Address{},
```

runtime/stdlib/account.go
```diff
@@ -791,6 +800,236 @@ func newPublicAccountContractsValue(
...
+					ExpectedType:  ty,
+					ActualType:    inter.MustConvertStaticToSemaType(publishedType),
+					LocationRange: getLocationRange(),
+				})
+			}
+
+			value := publishedValue.Value.Transfer(
+				inter,
+				getLocationRange,
+				atree.Address(address),
```
turbolent, line 994: Wrong address, value is transferred from storage to stack:

```suggestion
				atree.Address{},
```

turbolent (commented):
Looks good, just a couple storage-related issues left.

Also, the documentation checkbox in the description can be marked as done, the PR contains documentation for the new feature 👍

runtime/interpreter/primitivestatictype.go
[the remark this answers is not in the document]
dsainati1: Yes, the issue was in the interpreter `encoding_test.go` where the specific tag values were being tested in one case. Updating this test should fix things though.

runtime/interpreter/encoding_test.go
```diff
@@ -3097,7 +3097,7 @@ func TestEncodeDecodeCapabilityValue(t *testing.T) {
 			// positive integer to follow
 			0x18,
 			// public account (tag)
-			0x5b,
+			0x59,
```
turbolent, line 3100: This change seems suspicious, I'll have a look

runtime/interpreter/primitivestatictype.go
```diff
@@ -189,8 +189,6 @@ const (
 	_
 	_
 	_
```
turbolent, line 191: These lines were accidentally removed and need to brought back, otherwise the values for the following constants are changed, which breaks existing data:

```suggestion
	_
	_
	_
```

runtime/interpreter/encoding_test.go
turbolent, line 3100: After fixing the primitive static type constants above, this should be reverted:

```suggestion
			0x5b,
```

dsainati1: We have a test at the end of `statictype_test.go` asserting that there are exactly 98 static types (why?), so in order to add the two new inbox types at the end, I had to move the other account static types two earlier in the list, hence this number decreasing by two. If you'd rather change that test to account for 100 types instead, then that is also an option.

runtime/interpreter/statictype_test.go
```diff
@@ -963,6 +963,11 @@ func TestRestrictedStaticType_Equal(t *testing.T) {
 func TestPrimitiveStaticTypeCount(t *testing.T) {
 	t.Parallel()
 
+	// This asserts that the total number of types in the PrimitiveStaticType enum has not changed,
+	// in order to prevent adding new types into the enum in the middle.
+	// However, it is possible to safely change the size of this enum by only appending new types the end,
+	// (before the PrimitiveStaticType_Count of course).
+	// Only update this test if you are certain your change to this enum was to append new types to the end.
```
turbolent, line 970: 👌

turbolent (approved):
Nice work!

turbolent:
This is a rather large new feature, maybe hold off with merging this until we released the final version for the spork. 
It would be good to test this more extensively, especially in terms of how this change handles existing data.

SupunS (approved):
LGTM!

runtime/interpreter/accountinbox.go
```diff
@@ -0,0 +1,68 @@
...
+
+	var str string
+	stringer := func(memoryGauge common.MemoryGauge, _ SeenReferences) string {
+		if str == "" {
+			common.UseMemory(memoryGauge, common.AuthAccountInboxStringMemoryUsage)
+			addressStr := addressValue.MeteredString(memoryGauge, SeenReferences{})
+			str = fmt.Sprintf("AuthAccount.Inbox(%s)", addressStr)
+		}
+		return str
+	}
```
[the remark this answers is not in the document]
turbolent: 👍 

Could you please also fix this for the other cases in the codebase while at it?

runtime/stdlib/account.go
```diff
@@ -791,6 +800,236 @@ func newPublicAccountContractsValue(
...
+
+			recipient := recipientValue.Transfer(
+				inter,
+				getLocationRange,
+				atree.Address(address),
+				true,
+				nil,
+			).(interpreter.AddressValue)
+
+			publishedValue := interpreter.NewPublishedValue(inter, recipient, value)
```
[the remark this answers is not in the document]
turbolent: Yes, like noted above, `Transfer` would have to handle its children, like e.g. `SomeValue.Transfer`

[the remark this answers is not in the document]
SupunS: just double-checking: was this added? couldn't find it in the newest commits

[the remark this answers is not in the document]
dsainati1: Oh; I was interpreted @turbolent's comment to mean that this would be handled in the case where we generalized `PublishedValue` to be a container for arbitrary `Value` types in the future. I can submit another PR with this fix if that wasn't the case.

PR #1992, opened 2022-09-21
add key iteration method on dictionaries

### LINKED ISSUE #1938: Add key iteration function to dictionaries
### Issue To Be Solved

Dictionaries currently have a `let keys: [K]` field (where `K` is the type of the keys), which returns all keys of the dictionary.

Dictionaries may stores a lot of elements, so the resulting keys array might be very expensive to produce.

### Suggested Solution

Similar to the new [storage iteration functions](https://github.com/onflow/cadence/pull/1851), add a new function to `Dictionary` which allows iterating over keys of the dictionary:

```cadence 
fun forEachKey(_ f: ((K): Bool))
```

The `Bool` return value of the iteration function `f` determines whether iteration continues; `true` will proceed to the next key, while `false` will terminate iteration. The specific order in which the objects are iterated over is undefined, as is the behavior when a key is added or removed from the dictionary.

dreamsmasher:
Closes #1938 

## Description

Adds a method `fun forEachKey(f: ((K): Bool)): Void` on dictionary types, that allows for iteration over their keys with early exit.
______


dsainati1:
Curious why the design for this is `fun forEachKey(f: ((K): Bool)): Void` instead of `fun forEach(f: ((K, V): Bool)): Void`? There is not much you can do with a dictionary key without looking up the value in the dictionary, so requiring users to do another dictionary lookup with `d[k]` in the function body when they want to get the value, why not just provide the value as the second function parameter?

bluesign:
> why not just provide the value as the second function parameter?

It would cost double ( or more )

turbolent:
@dsainati1 No, atree has support for just iterating just over keys, without loading any values (`IterateKeys`). This is much more efficient. Just gathering the keys of a dictionary is actually a common operation, e.g. to get all NFT IDs stored in a collection.

dsainati1:
Makes sense, thanks.

turbolent:
@dreamsmasher Could you please also update the documentation?

docs/language/control-flow.md
```diff
@@ -344,6 +344,19 @@ for key in dictionary.keys {
 // 2
 ```
 
+Alternatively, dictionaries carry a method `forEachKey` that avoids allocating an intermediate array for keys:
+
+```cadence
+let dictionary = {"one": 1, "two": 2, "three": 3}
+dictionary.forEachKey(fun(key: String): Bool {
```
turbolent, line 351:
```suggestion
dictionary.forEachKey(fun (key: String): Bool {
```

docs/language/values-and-types.mdx
```diff
@@ -1394,6 +1394,32 @@ booleans[0] = true
...
+  The order of key iteration is undefined, similar to `.keys`.
+
+  ```cadence
+  // Take in a targetKey to look for, and a dictionary to iterate through.
+  fun myContainsKey(targetKey: String, dictionary: {String: Int}) {
+    // Declare an accumulator that we'll capture inside a closure.
+    var found = false
+
+    // At each step, `key` will be bound to another key from `dictionary`.
+    dictionary.forEachKey(fun(key: String): Bool {
```
turbolent, line 1410:
```suggestion
    dictionary.forEachKey(fun (key: String): Bool {
```

docs/language/values-and-types.mdx
```diff
@@ -1394,6 +1394,32 @@ booleans[0] = true
   //
   let containsKey42 = numbers.containsKey(42)
   ```
+- `cadence•fun forEachKey(function: ((K): Bool)): Void`
```
turbolent, line 1397:
```suggestion
- `cadence•fun forEachKey(_ function: ((K): Bool)): Void`
```

dreamsmasher: Should we add labels to all the other function signatures too?

turbolent: This change effectively *removes* the argument label, as the parameter name is used as the argument label by default if none is specified. 

This is also how it is already documented, implemented, and used in the rest of the PR, e.g. https://github.com/onflow/cadence/pull/1992/files/887efe8b73cd24d620bcc6760dc36392b4639b2f#diff-c01765841e64b85306258f5ee4daff2bdc5b9e9e107167f55a6dba4162e8bdb8R4595, this was the only place where it was inconsistent.

dreamsmasher: wait what lol, this whole time I thought the default if no label was specified was ArgumentLabelNotRequired. thanks for the heads up hahaha

runtime/sema/type.go
```diff
@@ -4491,6 +4499,28 @@ func (t *DictionaryType) initializeMemberResolvers() {
 					)
 				},
 			},
+			"forEachKey": {
+				Kind: common.DeclarationKindFunction,
+				Resolve: func(memoryGauge common.MemoryGauge, identifier string, targetRange ast.Range, report func(error)) *Member {
+					if t.KeyType.IsResourceType() {
```
dreamsmasher, line 4505: actually wait, do we even allow resources to be keys?

turbolent: No, but this is good future proofing :-)

dreamsmasher:
codecov pls have mercy,,, how do i reach unreachable errors in order to appease your hunger for coverage,,,,, how much of this repo must be tested before you are finally satiated,,,,,,,,

dsainati1 (approved):

runtime/interpreter/value.go
```diff
@@ -15627,6 +15627,43 @@ func (v *DictionaryValue) Destroy(interpreter *Interpreter, getLocationRange fun
 	)
 }
 
+func (v *DictionaryValue) ForEachKey(
+	interpreter *Interpreter,
+	getLocationRange func() LocationRange,
+	procedure FunctionValue,
+) VoidValue {
```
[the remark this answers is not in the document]
dreamsmasher: ah, do you mean to make `ForEachKey` a simple procedure? that makes sense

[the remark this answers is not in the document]
SupunS: yeh, one that returns nothing

SupunS (approved):

PR #1994, opened 2022-09-22
Improve type information in interpreter errors

### LINKED ISSUE #1986: Improve error message for casts
### Issue To Be Solved

When a type casting cannot be done, the message error is not specific enough
`error: unexpectedly found non-<Type> while force-casting value`

### Suggested Solution

The error should include the type of the value that's being casted, that should make it more obvious

turbolent:
Closes #1986

## Description

- Include actual type in error
- Use qualified strings, which helps with debugging, e.g., when force-casting a stored type which has the same qualified identifier, but different location (e.g. `0x1.C.S` vs `0x2.C.S`)
- Simplify related code by using `MustConvertStaticToSemaType` helper

______

### DIFF
--- a/runtime/account_test.go
+++ b/runtime/account_test.go
@@ -192,6 +192,8 @@ func TestRuntimeAccountKeyConstructor(t *testing.T) {
 		},
 	)
 	require.Error(t, err)
+	_ = err.Error()
+
 	assert.Contains(t, err.Error(), "cannot find variable in this scope: `AccountKey`")
 }
 
@@ -230,6 +232,8 @@ func TestRuntimeStoreAccountAPITypes(t *testing.T) {
 		)
 
 		require.Error(t, err)
+		_ = err.Error()
+
 		assert.Contains(t, err.Error(), "expected `Storable`")
 	}
 }
@@ -1073,6 +1077,8 @@ func TestRuntimePublicKey(t *testing.T) {
 
 		_, err := executeScript(script, runtimeInterface)
 		require.Error(t, err)
+		_ = err.Error()
+
 		assert.Contains(t, err.Error(), "value of type `PublicKey` has no member `validate`")
 	})
 
@@ -1111,6 +1117,8 @@ func TestRuntimePublicKey(t *testing.T) {
 				assert.NoError(t, err)
 			} else {
 				assert.Nil(t, value)
+				_ = err.Error()
+
 				assert.ErrorAs(t, err, &errorToReturn)
 				assert.ErrorAs(t, err, &interpreter.InvalidPublicKeyError{})
 			}
@@ -1257,6 +1265,7 @@ func TestRuntimePublicKey(t *testing.T) {
 
 		_, err := executeScript(script, runtimeInterface)
 		require.Error(t, err)
+		_ = err.Error()
 
 		var checkerErr *sema.CheckerError
 		require.ErrorAs(t, err, &checkerErr)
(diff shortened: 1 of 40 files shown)


SupunS (approved):

dsainati1 (approved):

turbolent:
Modifying the error messages showed they and many other error's error messages were not covered by tests.

In the checker tests, we have helpers that eagerly produce the message of an error, if it occurs. We do not have / use such a helper in the interpreter and runtime tests, so I manually added lines to generate error messages to all existing tests in [bc79b03](https://github.com/onflow/cadence/pull/1994/commits/bc79b03c4f740530ad9d5c4563734caa676f32ae). We should follow this pattern in the future or introduce and use a helper instead.

runtime/account_test.go
```diff
@@ -192,6 +192,8 @@ func TestRuntimeAccountKeyConstructor(t *testing.T) {
 		},
 	)
 	require.Error(t, err)
+	_ = err.Error()
```
dreamsmasher, line 195: Is this to catch panics when calling `.Error()`? I didn't realize that `_` binds an actual name instead of being a reserved wildcard

turbolent: Correct, this checks if the error message can be produced properly. We had cases in the past where the error message function was implemented incorrectly, or the error value itself was prepared incorrectly (e.g. a field was not filled, leading to a nil-dereference in the `Error()` function).

runtime/interpreter/errors.go
```diff
@@ -349,15 +350,17 @@ func (ForceCastTypeMismatchError) IsUserError() {}
 
 func (e ForceCastTypeMismatchError) Error() string {
 	return fmt.Sprintf(
-		"unexpectedly found non-`%s` while force-casting value",
+		"failed to force-cast value: expected `%s`, got `%s`",
```
dreamsmasher, line 353:
```suggestion
		"failed to force-cast value: expected type `%s`, got `%s`",
```

runtime/interpreter/errors.go
```diff
@@ -718,7 +726,7 @@ func (e NonStorableValueError) Error() string {
 // NonStorableStaticTypeError
 //
 type NonStorableStaticTypeError struct {
```
dreamsmasher, line 728: is this still a NonStorableStaticTypeError with the change?

runtime/tests/interpreter/capability_test.go
```diff
@@ -212,6 +212,7 @@ func TestInterpretCapability_borrow(t *testing.T) {
 
 			_, err := inter.Invoke("nonExistent")
 			require.Error(t, err)
+			_ = err.Error()
```
dreamsmasher, line 215: WHAT DO THESE MEAN

dreamsmasher (commented):
Looks good, although I'm confused about all the unbound err.Error() calls. If they have some semantic meaning, it might be worth factoring that call together with require.Error.

dreamsmasher (approved):

dsainati1 (approved):

SupunS (approved):

runtime/interpreter/errors.go
turbolent: Yes, we are just using the sema type hierarchy now instead of the interpreter static type one (both represent static types).

turbolent:
@dreamsmasher I tried to explain it in the comment above yours. Agreed, it would be nice to maybe introduce a helper function for it

PR #1996, opened 2022-09-22
Fix batch script

turbolent:
## Description
- Adjust error message used to detect non-existing account
- Add flag to configure number of concurrent clients
- Fix race condition: Don't drop error

______

### DIFF
--- a/tools/batch-script/address_provider.go
+++ b/tools/batch-script/address_provider.go
@@ -39,7 +39,7 @@ type AddressProvider struct {
 	currentIndex     uint
 }
 
-const endOfAccountsError = "storage used is not initialized or not initialized correctly"
+const endOfAccountsError = "get storage used failed"
 
 const accountStorageUsageScript = `
 pub fun main(address: Address): UInt64 {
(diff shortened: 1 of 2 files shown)


tools/batch-script/address_provider.go
```diff
@@ -39,7 +39,7 @@ type AddressProvider struct {
 	currentIndex     uint
 }
 
-const endOfAccountsError = "storage used is not initialized or not initialized correctly"
+const endOfAccountsError = "get storage used failed"
```
dreamsmasher, line 42: the wording here is kinda awkward

```suggestion
const endOfAccountsError = "failed to get storage used"
```

alternatively:
```suggestion
const endOfAccountsError = "aw heck we did a lil failure while getting storage used >_< gomenasorry"
```

dreamsmasher (approved):
LGTM

tools/batch-script/address_provider.go
turbolent: Just to clarify, this error is produced by flow-go / FVM and we are searching for it, not producing it.

dsainati1 (approved):

PR #1999, opened 2022-09-23
Invalidate resource references when the referenced resource is moved

SupunS:
## Description

Closes: https://github.com/onflow/cadence/issues/2092

This is the implementation of the FLIP: https://github.com/onflow/flow/pull/1043

Adds a run-time check to invalidate references, if the pointed resource is being moved.

______


runtime/tests/interpreter/reference_test.go
```diff
@@ -696,11 +676,7 @@ func TestInterpretReferenceUseAfterShiftStatementMove(t *testing.T) {
                   }
 
                   fun borrowR2(): &R2? {
-                      let optR2 <- self.r2 <- nil
-                      let r2 <- optR2!
-                      let ref = &r2 as &R2
-                      self.r2 <-! r2
-                      return ref
+                      return &self.r2 as &R2?
```
SupunS, line 679: The test code was updated in this test-case to keep supporting the rest of the test.
A duplicate of this test is added below (`"container in account, reference in stack"`) which doesn't change the code, but instead asserts the new behaviour.

runtime/interpreter/interpreter.go
```diff
@@ -4162,6 +4162,28 @@ func (interpreter *Interpreter) checkReferencedResourceNotDestroyed(value Value,
...
+) {
+	resourceKindedValue, ok := referencedValue.(ReferenceTrackedResourceKindedValue)
+	if !ok {
+		return
+	}
+
+	if resourceKindedValue.IsDestroyed() {
+		panic(DestroyedResourceError{
+			LocationRange: locationRange,
+		})
```
turbolent, line 4177: Maybe add a test case covering this branch, in case it's missing

turbolent (approved):
Nice work!

runtime/interpreter/interpreter.go
SupunS: This was an already existing check, so should be already covered in existing tests.

Anyway added another, just be sure: 6bfa8213e775eb1e4bb8bd9bc89eccf918dfd251

runtime/interpreter/errors.go
```diff
@@ -846,3 +846,17 @@ func (InvalidHexLengthError) IsUserError() {}
...
+type InvalidatedResourceReferenceError struct {
+	LocationRange
+}
+
+var _ errors.UserError = InvalidatedResourceReferenceError{}
+
+func (InvalidatedResourceReferenceError) IsUserError() {}
+
+func (e InvalidatedResourceReferenceError) Error() string {
+	return "referenced resource has been moved or destroyed after taking the reference"
```
dreamsmasher, line 861: this message is a little vague, why not something like

```suggestion
	return "attempted to follow a reference to a resource that has been moved or destroyed"
```

runtime/interpreter/value.go
```diff
@@ -1570,13 +1571,17 @@ func (v *ArrayValue) IsImportable(inter *Interpreter) bool {
 }
 
 func (v *ArrayValue) checkInvalidatedResourceUse(interpreter *Interpreter, locationRange LocationRange) {
-	if v.isDestroyed || (v.array == nil && v.IsResourceKinded(interpreter)) {
+	if v.isDestroyed || v.IsStaleResource(interpreter) {
 		panic(InvalidatedResourceError{
 			LocationRange: locationRange,
 		})
 	}
 }
```
dreamsmasher, line 1579: seems like the definition of `checkInvalidatedResourceUse` is the same for all 3 types (`ArrayValue`, `CompositeValue`, `DictionaryValue`). this could be a free function or a method on `Interpreter`

runtime/tests/interpreter/resources_test.go
```diff
@@ -2380,7 +2384,8 @@ func TestInterpretArrayOptionalResourceReference(t *testing.T) {
 	)
 
 	_, err := inter.Invoke("test")
-	require.NoError(t, err)
+	require.Error(t, err)
+	require.ErrorAs(t, err, &interpreter.InvalidatedResourceReferenceError{})
```
dreamsmasher, line 2388: same here
```suggestion
	require.ErrorAs(t, err, &interpreter.InvalidatedResourceReferenceError{})
```

runtime/tests/interpreter/resources_test.go
```diff
@@ -2344,7 +2347,8 @@ func TestInterpretOptionalResourceReference(t *testing.T) {
...
```
dreamsmasher, line 2351: `require.Error` is redundant here since `ErrorAs` already asserts that the error matches the target

```suggestion
	require.ErrorAs(t, err, &interpreter.InvalidatedResourceReferenceError{})
```

dreamsmasher (approved):

runtime/interpreter/value.go
SupunS: They are not, actually. for e.g: they rely on the atree value local field, which is different between, say `ArrayValue` and `CompositeValue`.

runtime/tests/interpreter/resources_test.go
SupunS: hmm actually, now that master is merged to the stable cadence branch, `require.Error` should be replaced by `RequireError`, which also checks for error messages etc.

runtime/interpreter/value.go
dreamsmasher: ahh, fun stuff. could still potentially abstract it away with an anonymous interface (as ugly as that approach is)

runtime/interpreter/errors.go
SupunS: I have mixed feelings about this.
Didn't want the error message to sound too technical, considering many smart-contract engineers may not be familiar with language internals/terms (i.e: "following a reference"). But at the same time, this may not be the best wording to put it (not my first language 😄), so I'm open to suggestions

dreamsmasher: What about
```suggestion
	return "null reference error: underlying resource has been moved or destroyed"
```

turbolent: There are no "null references" (or "null"s in general) in Cadence, I'd avoid mentioning it, as it would just confuse developers.

The current wording aligns with the phrasing of the rest of the error message. We can do a pass of improving error messages in the future, this shouldn't block getting a Stable Cadence preview release out

PR #2004, opened 2022-09-26
Make Void values constructible

### LINKED ISSUE #2005: Make Void Values constructible
### Issue To Be Solved

Currently, the Void type is a little inconsistent - it serves as an implicit return value for procedures, but isn't constructible on its own despite being a first-class value. The only way to conjure up a Void value is to bind a name to a procedure call:

```cadence
let v: Void = (fun() {})()
```

In the REPL, Void values are currently printed as () (the empty tuple, similar to other languages with unit types) but writing () on its own results in a parse error. Newcomers from languages like Rust, or Swift, might be thrown off by this inconsistency given that we already allow `Void` values to be used in expression position. 

### Suggested Solution

Introduce a `()` literal that corresponds to `Void`, allowing for unit values to be constructed outside of redundant procedure calls.

dreamsmasher:
Closes #2005 

## Description

Currently, the `Void` type is a little inconsistent - it serves as an implicit return value for procedures, but isn't constructible on its own despite being a first-class value. The only way to conjure up a `Void` value is to bind a name to a procedure call:

```cadence
let v: Void = (fun() {})()
```

In the REPL, Void values are currently printed as `()` (the empty tuple, similar to other languages with unit types) but writing `()` on its own results in a parse error. This PR introduces the following changes:

1. Introduce the `()` literal, which corresponds to `Void`. This equivalence (where `Void` refers to the empty tuple) is inspired by Swift, where `Void` is just a [type alias](https://developer.apple.com/documentation/swift/void). If tuples are added to the language later on, `Void` could also become a type alias for the empty tuple, piggybacking off the same parsing mechanisms as n-ary tuples. Using these values is cheap, as all instances of `Void` point to the same singleton object. Though it's outside the scope of this PR, future optimizations could treat unit types as having zero-size and elide their creation altogether. 

2. Make `Void` values equatable, i.e.

```cadence
() == ()
```

since equality is reflexive. This is a non-breaking change, as attempting to equate `Void` values previously resulted in an `InvalidBinaryOperandsError`.
______

### DIFF
--- a/runtime/ast/expression.go
+++ b/runtime/ast/expression.go
@@ -2009,8 +2009,10 @@ func (v *VoidExpression) String() string {
 
 func (v *VoidExpression) isIfStatementTest() {}
 
+var voidExpressionDoc prettier.Doc = prettier.Text("()")
+
 func (v *VoidExpression) Doc() prettier.Doc {
-	return prettier.Text("()")
+	return voidExpressionDoc
 }
 
 func (v *VoidExpression) isExpression() {}
@@ -2024,6 +2026,7 @@ func NewVoidExpression(
 	startPos Position,
 	endPos Position,
 ) *VoidExpression {
+	common.UseMemory(gauge, common.VoidValueMemoryUsage)
 	return &VoidExpression{Range: Range{startPos, endPos}}
 }
 
--- a/runtime/ast/expression.go
+++ b/runtime/ast/expression.go
@@ -2024,7 +2024,8 @@ func NewVoidExpression(
 	startPos Position,
 	endPos Position,
 ) *VoidExpression {
-	 return &VoidExpression{Range : Range {startPos, endPos}}
+	return &VoidExpression{Range: Range{startPos, endPos}}
 }
+
 var _ Element = &VoidExpression{}
 var _ Expression = &VoidExpression{}
--- a/runtime/ast/expression.go
+++ b/runtime/ast/expression.go
@@ -1981,3 +1981,50 @@ func (e *PathExpression) MarshalJSON() ([]byte, error) {
 func (*PathExpression) precedence() precedence {
 	return precedenceLiteral
 }
+
+// ()
+type VoidExpression struct {
+	Range
+}
+
+func (v *VoidExpression) StartPosition() Position {
+	return v.StartPos
+}
+
+func (v *VoidExpression) EndPosition(memoryGauge common.MemoryGauge) Position {
+	return v.EndPos
+}
+
+func (v *VoidExpression) ElementType() ElementType {
+	return ElementTypeVoidExpression
+}
+
+func (v *VoidExpression) Walk(walkChild func(Element)) {
+	// NO-OP
+}
+
+func (v *VoidExpression) String() string {
+	return "()"
+}
+
+func (v *VoidExpression) isIfStatementTest() {}
+
+func (v *VoidExpression) Doc() prettier.Doc {
+	return prettier.Text("()")
+}
+
+func (v *VoidExpression) isExpression() {}
+
+func (v *VoidExpression) precedence() precedence {
+	return precedenceLiteral
+}
+
+func NewVoidExpression(
+	gauge common.MemoryGauge,
+	startPos Position,
+	endPos Position,
+) *VoidExpression {
+	 return &VoidExpression{Range : Range {startPos, endPos}}
+}
+var _ Element = &VoidExpression{}
+var _ Expression = &VoidExpression{}
--- a/runtime/ast/expression_extractor.go
+++ b/runtime/ast/expression_extractor.go
@@ -172,7 +172,7 @@ type ExpressionExtraction struct {
 }
 
 // utility for expressions whose rewritten form is identical, i.e. nothing to rewrite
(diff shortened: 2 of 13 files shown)


runtime/ast/expression.go
```diff
@@ -1981,3 +1981,51 @@ func (e *PathExpression) MarshalJSON() ([]byte, error) {
...
+func (v *VoidExpression) precedence() precedence {
+	return precedenceLiteral
+}
+
+func NewVoidExpression(
+	gauge common.MemoryGauge,
+	startPos Position,
+	endPos Position,
+) *VoidExpression {
+	return &VoidExpression{Range: Range{startPos, endPos}}
```
SupunS, line 2027: Need to meter the expression creation.

runtime/ast/expression.go
```diff
@@ -1981,3 +1981,51 @@ func (e *PathExpression) MarshalJSON() ([]byte, error) {
...
+}
+
+func (v *VoidExpression) String() string {
+	return "()"
+}
+
+func (v *VoidExpression) isIfStatementTest() {}
+
+func (v *VoidExpression) Doc() prettier.Doc {
+	return prettier.Text("()")
```
SupunS, line 2013: maybe extract `prettier.Text("()")` to a variable and reuse

runtime/ast/expression_extractor.go
```diff
@@ -167,6 +171,26 @@ type ExpressionExtraction struct {
 	ExtractedExpressions []ExtractedExpression
 }
 
+// utility for expressions whose rewritten form is identical, i.e. nothing to rewrite
+func RewriteAsIs(expression Expression) ExpressionExtraction {
```
turbolent, line 175: Global function's name should be more specific, e.g. `RewriteExpressionAsIs`

runtime/sema/void_type.go
```diff
@@ -28,7 +28,7 @@ var VoidType = &SimpleType{
 	IsInvalid:            false,
 	IsResource:           false,
 	Storable:             false,
-	Equatable:            false,
+	Equatable:            true,
```
turbolent, line 31: Nice, lucky that `VoidValue` already implements `EquatableValue` 👍

dreamsmasher: yeah, it's the same story as the Paths. just need to turn the feature on :)

runtime/ast/expression_extractor.go
```diff
@@ -167,6 +171,26 @@ type ExpressionExtraction struct {
 	ExtractedExpressions []ExtractedExpression
 }
 
+// utility for expressions whose rewritten form is identical, i.e. nothing to rewrite
+func rewriteAsIs(expression Expression) ExpressionExtraction {
```
turbolent, line 175: See https://github.com/onflow/cadence/pull/2004#discussion_r982692582:

```suggestion
func rewriteExpressionAsIs(expression Expression) ExpressionExtraction {
```

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -9572,6 +9572,79 @@ func TestInterpretInternalAssignment(t *testing.T) {
...
+		inter,
+		interpreter.NewArrayValue(
+			inter, interpreter.ReturnEmptyLocationRange,
+			interpreter.VariableSizedStaticType{Type: interpreter.PrimitiveStaticTypeVoid},
+			common.Address{},
+			expected...,
+		),
+		arrayVal,
+	)
+}
```
turbolent, line 9647:
```suggestion
}

```

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -9572,6 +9572,79 @@ func TestInterpretInternalAssignment(t *testing.T) {
 	)
 }
 
+func TestInterpretVoidReturn(t *testing.T) {
```
turbolent, line 9575: Good idea to generate lots of test cases, but maybe run the test cases as individual subtests. That allows them to be be debugged and resolved more easily in case of a failure

turbolent (commented):
Does this have any effect on existing code?

dreamsmasher:
> 

It shouldn't, given that we're allowing a larger subset of programs than before. Void types were already assignable, and the only place where `()` was valid was in function calls.

bluesign:
I feel something is very wrong here, but @turbolent knows better.

turbolent:
@dreamsmasher Sounds good!

turbolent:
@bluesign What do you mean? What feels wrong?

turbolent (approved):

turbolent:
Just [like in Swift](https://developer.apple.com/documentation/swift/void), `Void` can be thought of as an alias for the empty tuple `()`.

bluesign:
> Just [like in Swift](https://developer.apple.com/documentation/swift/void), `Void` can be thought of as an alias for the empty tuple `()`.

@turbolent I misread probably:

I thougt: 
```
let v: Void = (fun() {})()
```
will be
```
let v: Void = ()
```

turbolent:
Yes, those two programs will be equivalent. `(fun () {})()` is basically equivalent to ` (fun (): Void { return () })()`

PR #1995, opened 2022-09-22
Fix crash in error pretty printer - Flow CLI (REPL)

Maelkum:
## Description

This PR makes a few code changes to the Cadence error pretty printer.

Since the crash is tightly coupled with the scenario of using a REPL that does somewhat unexpected things (like inserting empty code lines) adding tests would be somewhat more involved to custom tailor this scenario (but probably possible).

It's possible to test this quickly using https://github.com/Maelkum/cadence/tree/fix-pretty-print-v0.26.0 in a `replace` directive in `go.mod` file (I've got some errors trying to build Flow CLI with latest master of Cadence).

### Crash Description

Reproducible using latest Flow CLI (master branch) targeting Cadence v0.26.0 (also reproducible on v0.26.1).
Use Flow CLI to start Cadence REPL and try to re-declare a constant.

<details>
  <summary>Stack trace</summary>

```
./flow cadence
Welcome to Cadence v0.26.0!
Type '.help' for assistance.

1> let x = 1
2> let x = 1
error: cannot redeclare constant: `x` is already declared
 --> REPL:1:4
  |
1 |
  | panic: runtime error: index out of range [0] with length 0 [recovered]
        panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/onflow/cadence/runtime/pretty.ErrorPrettyPrinter.PrettyPrintError.func1()
        /home/aco/go/pkg/mod/github.com/onflow/cadence@v0.26.0/runtime/pretty/print.go:156 +0x106
panic({0x178bc40, 0xc000044438})
        /usr/local/go/src/runtime/panic.go:838 +0x207
github.com/onflow/cadence/runtime/pretty.ErrorPrettyPrinter.writeCodeExcerpts({{0x1e42500?, 0xc000010020?}, 0xc2?}, {0xc0001e2f00, 0x2, 0x40e374?}, {0x1e51548, 0x2a39848}, {0xc0001443e0, 0xb})
        /home/aco/go/pkg/mod/github.com/onflow/cadence@v0.26.0/runtime/pretty/print.go:316 +0x7e5
github.com/onflow/cadence/runtime/pretty.ErrorPrettyPrinter.prettyPrintError({{0x1e42500?, 0xc000010020?}, 0x80?}, {0x1e37ca0?, 0xc00026c080}, {0x1e51548, 0x2a39848}, {0xc0001443e0, 0xb})
[...]

PR #2021, opened 2022-10-01
Improve REPL

turbolent:
## Description

As previously reported in #1995, the REPL crashes during error pretty printing.
To fix this problem, we need two codes:
1. The code used for parsing and type checking (`code`). 

    Like before, this should be only the code that was just entered into the REPL, as we do not want to re-check and re-run the whole program already previously entered in the REPL – the checker's and interpreter's state are kept, and they already have the previously entered declarations. 

    However, just parsing the entered code would result in an AST with wrong position information, the line number would be always 1. To adjust the line information, we prepend the new code with empty lines.

2. The code used for error pretty printing (`codes`). 

    We temporarily update the full code of the whole program to include the new code. This allows the error pretty printer to properly refer to previous code (instead of empty lines), as well as the new code.
    However, if an error occurs, we revert the addition of the new code and leave the program code as it was before.

In addition, don't increment the line number if the code was invalid. Basically we properly completely invalid code now.

Also, fix the type checker and do not redeclare variables. The idea here was to improve error messages, by "assuming" a new type for a variable, which would follow what the user "means", and thus reduce follow-up error messages. For example:

```cadence
fun foo(bar: Int): String {
    // ...
    let bar: String = "..."    // previously, we would still re-declare `bar` as `String` (we still report an invalid re-declaration)
    return bar                 // ... so that e.g. we avoid a useless/additional type-mismatch error here
}
```

This is a nice-to-have feature, but also makes it impossible for the REPL to redact incorrect redeclared variables. 
Removing the feature allows the REPL to continue after invalid declarations were entered, but will also slightly decrease error message UX for developers. 
Alternatively we could make the variable declaration a checker option, have it enabled by default, and disable it for the REPL.

______

### DIFF
--- a/runtime/repl.go
+++ b/runtime/repl.go
@@ -19,6 +19,7 @@
 package runtime
 
 import (
+	"bytes"
 	"fmt"
 	goRuntime "runtime"
 	"sort"
@@ -95,18 +96,26 @@ func NewREPL(
 	return repl, nil
 }
 
-func (r *REPL) handleCheckerError() bool {
+func (r *REPL) handleCheckerError() error {
 	err := r.checker.CheckerError()
 	if err == nil {
-		return true
+		return nil
 	}
 	if r.onError != nil {
 		r.onError(err, r.checker.Location, r.codes)
 	}
-	return false
+	return err
 }
 
-func (r *REPL) Accept(code []byte) (inputIsComplete bool) {
+var lineSep = []byte{'\n'}
+
+func (r *REPL) Accept(code []byte) (inputIsComplete bool, err error) {
+	// Append the new code to the existing code (used for error reporting),
+	// temporarily, so that errors for the new code can be reported
+
+	currentCode := r.codes[r.checker.Location]
+
+	r.codes[r.checker.Location] = append(currentCode[:], code...)
 
 	defer func() {
 		if panicResult := recover(); panicResult != nil {
@@ -129,12 +138,35 @@ func (r *REPL) Accept(code []byte) (inputIsComplete bool) {
 		}
 	}()
 
-	r.codes[r.checker.Location] = code
+	// If the new code results in a parsing or checking error,
+	// reset the code
+	defer func() {
+		if err != nil {
+			r.codes[r.checker.Location] = currentCode
+		}
+	}()
+
+	// Only parse the new code, and ignore the existing code.
+	//
+	// Prefix the new code with empty lines,
+	// so that the line number is correct in error messages
+
+	lineSepCount := bytes.Count(currentCode, lineSep)
+
+	if lineSepCount > 0 {
+		prefixedCode := make([]byte, lineSepCount+len(code))
+
+		for i := 0; i < lineSepCount; i++ {
+			prefixedCode[i] = '\n'
+		}
+		copy(prefixedCode[lineSepCount:], code)
+
+		code = prefixedCode
+	}
 
 	// TODO: detect if the input is complete
 	inputIsComplete = true
 
-	var err error
 	result, errs := parser.ParseStatements(code, nil)
 	if len(errs) > 0 {
 		err = parser.Error{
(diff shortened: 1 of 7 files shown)


bluesign:
Out of curiosity, wouldn’t introducing some ReplLocation help?  So Repl would have cells (like in jupyter notebook) where each have line numbers.

dreamsmasher:
So to confirm, we can now redeclare variables on new lines? I've also been wanting some sort of history to remain between REPL invocations, e.g. if a line induces a panic then I can restart and up-arrow to get what I last typed. That could be a separate ticket, along with catching REPL panics to avoid crashes altogether.

runtime/repl.go
```diff
@@ -93,25 +94,75 @@ func NewREPL(
...
+		prefixedCode := make([]byte, lineSepCount+len(code))
 
-	r.codes[r.checker.Location] = code
+		for i := 0; i < lineSepCount; i++ {
+			prefixedCode[i] = '\n'
+		}
+		copy(prefixedCode[lineSepCount:], code)
+
+		code = prefixedCode
+	}
```
Maelkum, line 161: Same thing but worded slightly differently, think it may be a bit cleaner (no need for the if branch or the loop):

```go
prefix := bytes.Repeat([]byte{'\n'}, lineSepCount)

var prefixedCode []byte
prefixedCode = append(prefixedCode, prefix...)
prefixedCode = append(prefixedCode, code...)

code = prefixedCode
```

Just a stylistic comment so feel free to disregard, the code works.

runtime/sema/variable_activations.go
```diff
@@ -298,15 +298,12 @@ func (a *VariableActivations) declare(declaration variableDeclaration) (variable
 			existingVariable.ActivationDepth == depth ||
 			existingVariable.ActivationDepth == 0) {
 
-		err = &RedeclarationError{
+		return nil, &RedeclarationError{
```
Maelkum, line 301: Outside of the scope of changes in this PR but it popped into my eye - this function (`declare`) now doesn't use named return values and returns explicitly. The named return values can be removed - less things to keep in your head when reading the code.

Maelkum (approved):
Looks good to me. Tested that the REPL crash does not happen, and the error is reported correctly (e.g., on const redeclaration the previous code is correctly cited).

turbolent:
@bluesign 
> Out of curiosity, wouldn’t introducing some ReplLocation help?

The REPL already uses `REPLLocation` (`REPL`) as the location for the checker and interpreter.

> Repl would have cells (like in jupyter notebook) where each cell have own line numbering.

The current REPL implements a "traditional" REPL, where lines are immutable. Creating "notebook" functionality is a great idea, but probably better done in a separate implementation.

turbolent:
@dreamsmasher 

> So to confirm, we can now redeclare variables on new lines? 

No, this change does not change anything about that, variables still cannot be redeclared. Support for that could be added in the future though. This PR just fixes handling the error that is reported for a redeclaration.

> I've also been wanting some sort of history to remain between REPL invocations, e.g. if a line induces a panic then I can restart and up-arrow to get what I last typed. That could be a separate ticket, along with catching REPL panics to avoid crashes altogether.

These are all great ideas 👍  Maybe open a feature request issue to gather such ideas for improvements to the REPL

runtime/repl.go
turbolent: Thanks for the suggestion. The code is indeed a bit more "idiomatic", though I prefer keeping the current code, as it is more efficient:
- It only allocates prefixed code if needed. In the case where there is nothing to prefix (`lineSepCount == 0`) we can use `code` as-is and do not need to 
- It allocates the full slice upfront and fills it, which avoids potential re-allocations when appending

runtime/sema/variable_activations.go
turbolent: Good point! I removed the variable names in 2fa741bbbf78f7dfe07ac5e64650324e4c8d9650

runtime/repl.go
Maelkum: Ah, okay, I didn't consider that aspect of things. That's cool then, feel free to disregard my comment :)

SupunS (approved):

turbolent:
@dsainati1 Could you please have a quick look? 🙏

PR #2022, opened 2022-10-03
Fix accounts not being returnable from scripts

dsainati1:
Adding the storage iteration functions to the accounts as field members rather than function members accidentally prevented them from being considered externally returnable. This re-labels these as function members to fix this regression. 
______

### DIFF
--- a/runtime/sema/publicaccount_type.go
+++ b/runtime/sema/publicaccount_type.go
@@ -113,7 +113,7 @@ var PublicAccountType = func() *CompositeType {
 			PublicAccountPathsType,
 			publicAccountTypePathsFieldDocString,
 		),
-		NewUnmeteredPublicConstantFieldMember(
+		NewUnmeteredPublicFunctionMember(
 			publicAccountType,
 			PublicAccountForEachPublicField,
 			PublicAccountForEachPublicFunctionType,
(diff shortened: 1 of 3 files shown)


janezpodhostnik (approved):
Code looks ok to me, but I don't know the details.

I like the new tests!

Not sure if the `AuthAccount` was returnable before... I guess it was since this fixes it.

runtime/sema/publicaccount_type.go
```diff
@@ -113,7 +113,7 @@ var PublicAccountType = func() *CompositeType {
 			PublicAccountPathsType,
 			publicAccountTypePathsFieldDocString,
 		),
-		NewUnmeteredPublicConstantFieldMember(
+		NewUnmeteredPublicFunctionMember(
```
turbolent, line 116: Just noticed: Is there any reason we do not have `PublicAccount.forEachPrivate` and `PublicAccount.forEachStored`? Those two functions would be very useful in scripts, e.g. for a wallet that is introspecting an account.

turbolent (approved):
Thank you for the quick fix, and most of all, adding a test case for returning a `PublicAccount`! Given that we now also have `getAuthAccount` in scripts, maybe we should support returning `AuthAccount` as well?

runtime/sema/publicaccount_type.go
dsainati1: These functions don't need to exist for scripts since we support `getAuthAccount` now; users can just `getAuthAccount` and call the functions directly from there.

turbolent: I see, that makes sense 👍

turbolent: Only downside is that using `getAuthAccount` instead is not possible in a transaction. We can add the functions it later if needed though

PR #2024, opened 2022-10-04
Change for-loop variables from per-loop to per-iteration

turbolent:
⚠️ Depends on #2141

## Description

We should probably change the semantics of for-loops. 
Currently Cadence introduces one variable for the value, one variable for the index (if given), for the whole loop. 
Capturing these variables has surprising behaviour. For example, this program repeatedly logs 3, instead of the items of the array, as 3 is the last value assigned to the value variable.

```cadence
let fs: [((): Int)] = []
for x in [1, 2, 3] {
    fs.append(fun (): Int {
        return x
    })
}

for f in fs {
    log(f())
}
```

This has been a problem in many other languages. Some languages and users mostly agree changing the semantics is a good idea:

- [Go is planning to change their for-loop semantics](https://github.com/golang/go/discussions/56010).
- [C# already changed their for-loop semantics which caused little problems](https://github.com/golang/go/discussions/56010#discussioncomment-3788526). 

The improvement is simple: Instead of introducing the variable(s) for the whole loop, introduce them for each loop iteration.

I quickly implemented the proposed change so it can be evaluated. 
If there's positive sentiment here, I'll open a FLIP to propose the change more formally. 

______


dsainati1:
I am in favor of this change; it makes intuitive sense, I was surprised to run into this behavior while writing Go. This definitely needs a FLIP though before we proceed.

runtime/tests/interpreter/for_test.go
```diff
@@ -219,3 +219,34 @@ func TestInterpretForStatementEmpty(t *testing.T) {
...
+	t.Parallel()
+
+	inter := parseCheckAndInterpret(t, `
+       fun test(): Int {
+           let fs: [((): Int)] = []
+           for x in [1, 2, 3] {
+               fs.append(fun (): Int {
+                   return x
+               })
+           }
```
SupunS, line 234: Also extend this test to check the same for the index-variable

SupunS (approved):
+1 for the new behaviour 👍

runtime/tests/interpreter/for_test.go
```diff
@@ -219,3 +219,34 @@ func TestInterpretForStatementEmpty(t *testing.T) {
...
+           var sum = 0
+           for f in fs {
+              sum = sum + f()
```
dreamsmasher, line 237: maybe push each function call's result to an output array, which ensures the correct ordering of captures too. this test will pass if 1 and 2 are discarded but the function returning 3 is run twice, for instance.

dreamsmasher (approved):
LGTM, I didn't even realize we had the same scoping issue for loop variables. I agree with Daniel about adding more tests.

turbolent:
Opened FLIP https://github.com/onflow/flips/pull/13

runtime/tests/interpreter/for_test.go
turbolent: Good idea! Improved the tests in [e4478d3](https://github.com/onflow/cadence/pull/2024/commits/e4478d3ac06134975b6f473edee095ae00718372)

turbolent:
Blocked on onflow/flips#49. @SupunS Could you please have a quick look?

PR #2034, opened 2022-10-05
Check errors' potential secondary message and error notes

turbolent:
## Description

In addition to checking if the message of an error can be produced without problems, also check that:
- The error's secondary message, if any, can be produced without problems
- The error's error notes' messages, if any, can be produced without problems

______


runtime/tests/utils/utils.go
```diff
@@ -192,3 +193,17 @@ func AssertValueSlicesEqual(t testing.TB, inter *interpreter.Interpreter, expect
 
 	return true
 }
+
+func CheckErrorMessage(err error) {
```
dreamsmasher, line 197: why not combine this with `require.Error`, since every invocation asserts that err isn't null?

```suggestion
func CheckErrorMessage(t *testing.T, err error) {
	require.Error(t, err)
	_ = err.Error()
```

dreamsmasher (approved):

runtime/tests/utils/utils.go
SupunS: +1
And perhaps rename `CheckErrorMessage` to something like `RequireError`

turbolent: No reason why not, good idea

turbolent: Refactored in [94a0700](https://github.com/onflow/cadence/pull/2034/commits/94a07006da695f97e4833396c806c846e9514a78)

turbolent:
Took some time and also switched the majority of the remaining `require.Error` uses to `RequireError`, as well as improved assertions in [a30a8ea](https://github.com/onflow/cadence/pull/2034/commits/a30a8ea827681ea926eebebec6eb82449d42945f)

runtime/tests/checker/access_test.go
```diff
@@ -1999,7 +1999,9 @@ func TestCheckAccessSameContractInnerStructField(t *testing.T) {
 			if expectSuccess {
 				require.NoError(t, err)
 			} else {
-				require.Error(t, err)
+				errs := ExpectCheckerErrors(t, err, 1)
+
+				assert.IsType(t, &sema.InvalidAccessError{}, errs[0])
```
dsainati1, line 2004: Why do these not use the new `RequireError` function?

SupunS (approved):

runtime/tests/checker/access_test.go
turbolent: Refactored in [0e5b5f7](https://github.com/onflow/cadence/pull/2034/commits/0e5b5f7cdcabf6e315cdea76120a3f7b7326ea93)

turbolent:
@SupunS just pushed another commit, a simple rename, could you please have another look?

dsainati1:
Why is the linter failing?

turbolent:
@dsainati1 

I was confused about it at first, but it makes sense: https://github.com/onflow/cadence/pull/2034/files#diff-c3306e7eb27241cc245335501f5b9cf9dcd45474613bc9e0c0926f7a71666973R29 adds an import of `github.com/stretchr/testify/require` to `runtime/tests/utils/utils.go` – even though it is intended as "testing utility code", it still is part of the module, as it is not testing code (`_test.go`). Unfortunately, it needs to be non-test code, because test-code cannot be imported from another package. 

Updated the go.cap file in 0e40238f94298cd9573bb3b6312f36d436a917b6

PR #2037, opened 2022-10-06
Detect invalidated references statically

SupunS:
Working towards the FLIP: https://github.com/onflow/flow/pull/1043

## Description

Statically check for invalidated references, whenever possible.

All invalidations may not be caught by the static checker. Hence the runtime validation is also added in: https://github.com/onflow/cadence/pull/1999

______


dsainati1 (commented):
I am conflicted about this. On the one hand it's always nice to catch more errors statically, but I do wonder if this might make us run afoul of the uncanny valley effect that can sometimes show up in gradually typed or partially statically typed languages. 

When one has an analysis like this that is best-effort but not completely sound, improving the coverage of the analysis without achieving complete soundness can paradoxically make the user experience worse, because without an actual guarantee of correctness the user cannot actually fully trust that the analysis is going to catch their error. The first time a user encounters a reference invalidation that we don't catch here and learns that the static checking is not able to identify every possible runtime error, they are going to lose trust in the analysis entirely. In the best case, they are simply going to learn to think about how references behave at runtime the same way they would if the analysis didn't exist, while in the worst case having a partially correct analysis is going to confuse them about the behavior more than if there were no analysis at all. 

IMO in order to evaluate this we should have a known list of cases the analysis is not capable of catching statically so we can see how often they show up in practice.

SupunS:
I had the same thought initially. But after seeing how many contracts/obvious-cases can be detected statically, I actually changed that stance, and feels like it would be a sin not to catch them statically and enforce them to fix it (vs waiting for the run-time check to catch them on the act.)

The known scenario we don't catch right now is 
1) taking storage references and changing the storage
2) moving resources after taking references with re-entrancy: not possible to detect / would require massive analysis including maintaining function states.

This change only targets ephemeral references, so it could argue that no change is done for storage references (case-1). I think it safe to say that case-2 is very rare.

On the other hand, there are similar places where the best-effort analysis is done; for eg: certain invalid casting operations can be caught statically. And then there are also ones that cannot be, and those are caught dynamically at run-time.

runtime/missingmember_test.go
```diff
@@ -1619,20 +1619,12 @@ pub contract ItemNFT: NonFungibleToken {
 
        // get a reference to the garment that item stores
        pub fun borrowGarment(): &GarmentNFT.NFT? {
-           let garmentOptional <- self.garment <- nil
-           let garment <- garmentOptional!
-           let garmentRef = &garment as auth &GarmentNFT.NFT
-           self.garment <-! garment
-           return garmentRef
+           return &self.garment as auth &GarmentNFT.NFT?
```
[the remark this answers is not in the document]
SupunS: I kept the original casting as is, and only removed the necessary move operation.

dsainati1:
I see! If the only case we can't detect statically is the storage change, then it makes sense to add this check, especially since as you pointed out this whole FLIP only proposed to change ephemeral references anyways. We should document this limitation explicitly though, so people are aware of what the boundaries of the analysis are.

On that point, the documentation boxes on this PR (and the underlying PR) are checked but I don't see any changes to the documentation in them. Are we planning to add a section to the reference documentation about this invalidation?

SupunS:
Actually, realized one more case: reference-typed fields are not possible to track. Added a test-case: https://github.com/onflow/cadence/pull/2037/commits/f023a1d37a9eb22feb62b0a765fb8b5ef4246df5

dsainati1 (approved):

runtime/sema/check_assignment.go
```diff
@@ -118,18 +120,19 @@ func (checker *Checker) checkAssignment(
...
+//	  init(x: Int) {
+//	    self.x = x
+//	  }
+//	 }
+//
+//	view fun foo(_ f: @R): @R {
+//	  let b <- f
+//	  b.x = 3 // b was created in the current scope but modifies the resource value
+//	  return <-b
+//	}
```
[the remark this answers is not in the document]
SupunS: That's actually the go 1.19 linter. It identifies it as a code block, and indents

docs/language/references.mdx
```diff
@@ -151,3 +151,39 @@ counterRef3.count  // is `44`
 
 References are ephemeral, i.e. they cannot be [stored](accounts#account-storage).
 Instead, consider [storing a capability and borrowing it](capability-based-access-control) when needed.
+
+### Reference validity
+
+Ephemeral references stay valid throughout the course of the program.
+However, **references to resources** can become invalid during a program if the underlying resource is moved
```
turbolent, line 158:
```suggestion
However, **references to resources** can become invalid during the execution of a program, if the referenced resource is moved
```

docs/language/references.mdx
```diff
@@ -151,3 +151,39 @@ counterRef3.count  // is `44`
...
+// Static error, since the referenced resource has been moved.
+ref.id = 2
+```
+
+A reference is invalidated upon the first transfer of the underlying resource, regardless of the origin and the destination.
+
+```cadence
+let ref = &r as &R
+
+// Moving the resource to a different vairable would also invalidate the reference.
```
turbolent, line 180:
```suggestion
// Moving a resource to a different variable invalidates all references of it.
```

docs/language/references.mdx
```diff
@@ -151,3 +151,39 @@ counterRef3.count  // is `44`
...
+let r2 <- r
+
+// Static error, since the referenced resource has been moved.
+ref.id = 2
+```
+
+<Callout type="info">
+Invalidation of storage references are not statically caught, but happens at run-time.
```
turbolent, line 188:
```suggestion
Invalidations of storage references are not statically caught, but only at run-time.
```

runtime/missingmember_test.go
turbolent, line 1622: Does not have to be `auth`
```suggestion
           return &self.garment as &GarmentNFT.NFT?
```

runtime/missingmember_test.go
```diff
@@ -1619,20 +1619,12 @@ pub contract ItemNFT: NonFungibleToken {
...
        }
 
        // get a reference to the material that item stores
        pub fun borrowMaterial(): &MaterialNFT.NFT?  {
-           let materialOptional <- self.material <- nil
-           let material <- materialOptional!
-           let materialRef = &material as auth &MaterialNFT.NFT
-           self.material <-! material
-           return materialRef
+           return &self.material as auth &MaterialNFT.NFT?
```
turbolent, line 1627: Does not have to be `auth`

```suggestion
           return &self.material as &MaterialNFT.NFT?
```

runtime/sema/check_variable_declaration.go
```diff
@@ -234,3 +236,119 @@ func (checker *Checker) elaborateNestedResourceMoveExpression(expression ast.Exp
...
+
+	var variableRefExpr *ast.Identifier
+
+	switch refExpression := refExpression.(type) {
+	case *ast.ReferenceExpression:
+		// If it is a reference expression, then find the root variable.
+		// i.e: root variable of `&a.b.c as &T` is `a`.
+		// This is because nested resources cannot be moved.
+		// If `c` needs to be moved, then `a` has to be moved.
+		// So tracking `a` is sufficient.
```
turbolent, line 272: If field `c` is resource-typed, it can be moved if the field is variable, e.g. through a swap.
As nested resources (here, in a nested field, or e.g. in an array) cannot be tracked, it still makes sense to at least track `a`.

Maybe

```suggestion
		// If it is a reference expression, then find the "root variable".
		// As nested resources cannot be tracked, at least track the "root" if possible.
		// For example, for an expression `&a.b.c as &T`, the "root variable" is `a`.
```

runtime/sema/check_variable_declaration.go
```diff
@@ -234,3 +236,119 @@ func (checker *Checker) elaborateNestedResourceMoveExpression(expression ast.Exp
...
+	if variableRefExpr == nil {
+		return nil
+	}
+
+	referencedVariableName := variableRefExpr.Identifier
+
+	for {
+		// If the referenced variable is again a reference,
+		// then find the variable of the root of the reference chain.
+		// e.g::
```
turbolent, line 289:
```suggestion
		// e.g.:
```

runtime/sema/check_variable_declaration.go
```diff
@@ -234,3 +236,119 @@ func (checker *Checker) elaborateNestedResourceMoveExpression(expression ast.Exp
...
+		//     ref1 = &v
+		//     ref2 = &ref1[0]
```
turbolent, line 291: Is the indexing an accident?

```suggestion
		//     ref2 = &ref1
```

runtime/sema/check_variable_declaration.go
```diff
@@ -234,3 +236,119 @@ func (checker *Checker) elaborateNestedResourceMoveExpression(expression ast.Exp
...
+	}
+
+	return checker.valueActivations.Find(referencedVariableName)
+}
+
+// referenceExpression returns the expression that return a reference.
+// There could be two types of expressions that can result in a reference:
+//  1. Expressions that create a new reference.
+//     (i.e: reference-expression)
+//  2. Expressions that returns an existing reference.
```
turbolent, line 312:
```suggestion
//  2. Expressions that return an existing reference.
```

runtime/sema/check_variable_declaration.go
```diff
@@ -234,3 +236,119 @@ func (checker *Checker) elaborateNestedResourceMoveExpression(expression ast.Exp
...
+//     (i.e: identifier-expression/member-expression/index-expression having a reference type)
+//
+// However, it is  not currently possible to track member-expressions/index-expression.
+// So this method would only return either a `reference-expression` or an `identifier-expression`.
```
turbolent, line 316:
```suggestion
// However, it is currently not possible to track member-expressions and index-expressions.
// So this method either returns a reference-expression or an identifier-expression.
```

runtime/sema/check_variable_declaration.go
```diff
@@ -234,3 +236,119 @@ func (checker *Checker) elaborateNestedResourceMoveExpression(expression ast.Exp
...
+		return expr
+	case *ast.ForceExpression:
+		return checker.referenceExpression(expr.Expression)
+	case *ast.CastingExpression:
+		return checker.referenceExpression(expr.Expression)
+	case *ast.BinaryExpression:
+		if expr.Operation != ast.OperationNilCoalesce {
+			return nil
+		}
+		return checker.referenceExpression(expr.Left)
```
turbolent, line 333: Why do we return the left expression for nil-coalescing? Is it a "best guess"?

runtime/sema/check_variable_declaration.go
```diff
@@ -234,3 +236,119 @@ func (checker *Checker) elaborateNestedResourceMoveExpression(expression ast.Exp
...
+//  2. Expressions that returns an existing reference.
+//     (i.e: identifier-expression/member-expression/index-expression having a reference type)
+//
+// However, it is  not currently possible to track member-expressions/index-expression.
+// So this method would only return either a `reference-expression` or an `identifier-expression`.
+//
+// The expression could also be hidden inside some other expression.
+// e.g(1): `&v as &T` is a casting expression, but has a hidden reference expression.
+// e.g(2): `(&v as &T?)!
+func (checker *Checker) referenceExpression(expr ast.Expression) ast.Expression {
```
turbolent, line 321: Doesn't really use the receiver (checker), maybe make a normal function

runtime/sema/check_variable_declaration.go
```diff
@@ -234,3 +236,119 @@ func (checker *Checker) elaborateNestedResourceMoveExpression(expression ast.Exp
...
+	case *ast.IdentifierExpression:
+		return expr
+	default:
+		return nil
+	}
+}
+
+// variableReferenceExpression returns the identifier expression
+// of a var-ref/member-access/index-access expression.
+func (checker *Checker) variableReferenceExpression(expr ast.Expression) *ast.Identifier {
```
turbolent, line 343: Doesn't really use the receiver (checker), maybe make a normal function

runtime/sema/checker.go
```diff
@@ -121,6 +121,9 @@ type Checker struct {
 	// memoryGauge is used for metering memory usage
 	memoryGauge  common.MemoryGauge
 	PositionInfo *PositionInfo
+
+	// Holds a mapping between references and their referenced-variable
+	references *VariableActivations
```
turbolent, line 126: Maybe compare this approach of implementing reference tracking with adding a new `referencedVariable *Variable` field to `Variable` and comparing the performance. We can also do this in a follow-up PR

runtime/sema/check_expression.go
```diff
@@ -41,16 +41,39 @@ func (checker *Checker) VisitIdentifierExpression(expression *ast.IdentifierExpr
...
+		return
+	}
+
+	resourceInfo := checker.resources.Get(Resource{Variable: referencedVar})
+	if resourceInfo.Invalidations.Size() == 0 {
+		return
+	}
+
+	checker.report(&InvalidatedResourceReferenceError{
+		Range: ast.NewRangeFromPositioned(checker.memoryGauge, hasPosition),
```
turbolent, line 71: Maybe also include the position of the invalidation to aid the developer.

See [the implementation of `ResourceUseAfterInvalidationError` on `master`](https://github.com/onflow/cadence/blob/d8b6241606b9dd78a5ca5d9b646dfa27f7842037/runtime/sema/errors.go#L1869-L1909).

turbolent (commented):
Great work!

docs/language/references.mdx
dsainati1: I think usually we say "references to a value" rather than "references of a value", so maybe "invalidates all references to it" instead?

SupunS:
```suggestion
// Moving a resource to a different variable invalidates all references to it.
```

runtime/sema/check_variable_declaration.go
SupunS: no, what I actually meant there was taking a reference to an inner element, via a reference.
But you are right, we can simplify that by simply taking a reference to a reference.

runtime/sema/check_variable_declaration.go
SupunS: Resources re not allowed in rhs of the nil-coalescing expressions. So only need to visit the lhs.
I'll add a comment mentioning this.

SupunS: ah actually, reference is still allowed. good catch!

runtime/sema/checker.go
SupunS: Seems there's no performance diff between the two. Anyway removed the bookkeeping since it makes the code simple bdde127d0906bb36a0f9dc051bf6b9a57020a229

runtime/sema/check_variable_declaration.go
SupunS: Added: https://github.com/onflow/cadence/pull/2037/commits/c6dff546a8d9a6bc35e635fa3aae09b386c8a5db

runtime/sema/check_expression.go
SupunS: Seems I need to wait for this PR https://github.com/onflow/cadence/pull/2051

turbolent: You don't have to wait for it necessarily, what I referred to was just meant as inspiration.

SupunS: I could re-use the same `ErrorNote` interface. But the API has changed a bit in the master, so the two PRs might conflict. I can wait for that to merge, or also can add this in a follow-up PR.

turbolent: Ah, I see. That's a good plan!

runtime/storage_test.go
```diff
@@ -2438,19 +2440,13 @@ func TestRuntimeReferenceOwnerAccess(t *testing.T) {
...
-				"nil",
-				"0x0000000000000001",
-				"nil",
-				"nil",
-				"0x0000000000000002",
-				"0x0000000000000002",
-			},
-			loggedMessages,
-		)
+		errors := checker.ExpectCheckerErrors(t, err, 4)
```
turbolent, line 2444: The test here and the tests below are integration tests that test if contracts' `owner` field can be accessed.
Instead of adjusting the assertions to static errors, adjust the test code so that it will actually be semantically valid. I can assist with that if you'd like.

runtime/tests/interpreter/reference_test.go
```diff
@@ -452,7 +451,7 @@ func TestInterpretResourceReferenceAfterMove(t *testing.T) {
 
 		t.Parallel()
 
-		inter := parseCheckAndInterpret(t, `
```
turbolent, line 455: Given that these are interpreter tests, we only want to test semantically valid programs, and reference use after resource moves are not semantically valid anymore, maybe just remove this whole test case (`TestInterpretResourceReferenceAfterMove`), as long as we have equivalent tests for the checker, i.e. move test cases to the checker test cases if they cover cases not yet covered by the checker tests.

Same for the other test cases that assumed use after referenced resource move are valid, e.g. `TestInterpretReferenceUseAfterShiftStatementMove`

turbolent (commented):
Looks good! Main last thing is moving/removing/adjusting test cases. I can assist with that if you'd like me to

runtime/sema/check_variable_declaration.go
```diff
@@ -240,3 +242,155 @@ func (checker *Checker) elaborateNestedResourceMoveExpression(expression ast.Exp
...
+		switch refExpr := refExpr.(type) {
+		case *ast.ReferenceExpression:
+			// If it is a reference expression, then find the "root variable".
+			// As nested resources cannot be tracked, at least track the "root" if possible.
+			// For example, for an expression `&a.b.c as &T`, the "root variable" is `a`.
+			variableRefExpr = rootVariableOfExpression(refExpr.Expression)
+		case *ast.IdentifierExpression:
+			variableRefExpr = &refExpr.Identifier
+		default:
+			continue
```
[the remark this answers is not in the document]
SupunS: There is a test-case for this one already: https://github.com/onflow/cadence/blob/dbd98a8181767957b95ba2bf02594070688e0e48/runtime/tests/checker/reference_test.go#L2269

Codecov reports false positives sometimes.

runtime/sema/check_variable_declaration.go
```diff
@@ -240,3 +242,155 @@ func (checker *Checker) elaborateNestedResourceMoveExpression(expression ast.Exp
...
+
+		rhsRef := referenceExpressions(expr.Right)
+		if rhsRef != nil {
+			refExpressions = append(refExpressions, rhsRef...)
+		}
+
+		return refExpressions
+	case *ast.IdentifierExpression:
+		return []ast.Expression{expr}
+	default:
```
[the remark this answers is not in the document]
SupunS: Ah yes! added in 1fb97f405fb8bdae0a3e0bb24ac2267c64d89808

runtime/sema/check_variable_declaration.go
```diff
@@ -240,3 +242,155 @@ func (checker *Checker) elaborateNestedResourceMoveExpression(expression ast.Exp
...
+		if !variable.Type.IsResourceType() {
+			return nil
+		}
+
+		return []*Variable{variable}
+	}
+
+	var referencedResourceVariables []*Variable
+	for _, referencedVar := range variable.referencedResourceVariables {
+		nestedReferencedVars := nestedReferencedVariables(referencedVar)
```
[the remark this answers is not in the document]
SupunS: the reference chain is a tree, and needs to visit in a depth-first manner. Writing this with a for loop would not be more optimized than the recursion, as it would need to maintain a stack.

Recursion here might not be that bad, because in practice, it is highly unlikely people would write code that has long chains of references. Maybe a depth of one or two in most cases.

runtime/tests/interpreter/reference_test.go
SupunS: Moved these to the checker tests: 55b8fa36662fbaeabee0e32b4bf6e528e07a2a8f

runtime/sema/check_variable_declaration.go
```diff
@@ -240,3 +242,169 @@ func (checker *Checker) elaborateNestedResourceMoveExpression(expression ast.Exp
...
+	case *ast.IdentifierExpression:
+		return []ast.Expression{expr}
+	default:
+		return nil
+	}
+}
+
+// rootVariableOfExpression returns the identifier expression
+// of a var-ref/member-access/index-access expression.
+func rootVariableOfExpression(expr ast.Expression) *ast.Identifier {
```
dreamsmasher, line 376: does go support tail recursion? this could be rewritten to a loop to avoid stack overflow

dreamsmasher (approved):

runtime/sema/check_expression.go
```diff
@@ -41,16 +41,39 @@ func (checker *Checker) VisitIdentifierExpression(expression *ast.IdentifierExpr
...
+	})
+}
+
 // checkSelfVariableUseInInitializer checks uses of `self` in the initializer
 // and ensures it is properly initialized
-//
```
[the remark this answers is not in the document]
SupunS: yup

runtime/tests/interpreter/dynamic_casting_test.go
```diff
@@ -2334,6 +2336,15 @@ func testReferenceCastValid(t *testing.T, types, fromType, targetType string, op
 
 	switch operation {
 	case ast.OperationFailableCast:
+		if isResource {
+			AssertValuesEqual(
+				t,
+				inter,
+				interpreter.BoolValue(true),
```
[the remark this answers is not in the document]
SupunS: Master is still not merged to the stable-cadence branch yet. We can update it once it is merged. (although this test case might not break since go does a value-equality rather than instance equality for value types)

runtime/sema/check_variable_declaration.go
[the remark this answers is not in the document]
turbolent: Ah, right. I missed that this was tree-shaped 👍

runtime/sema/check_variable_declaration.go
SupunS: refactored in 62b2e043a55cc0f15bf54632c2443494a04392ba

runtime/sema/check_expression.go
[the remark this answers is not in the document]
dreamsmasher: ah i wrote most of these review comments last week, guess I forgot to submit them until this batch. should be fine now that I upgraded things to go 1.19

runtime/sema/check_variable_declaration.go
```diff
@@ -374,16 +374,20 @@ func referenceExpressions(expr ast.Expression) []ast.Expression {
...
-	switch expr := expr.(type) {
-	case *ast.IdentifierExpression:
-		return &expr.Identifier
-	case *ast.MemberExpression:
-		return rootVariableOfExpression(expr.Expression)
-	case *ast.IndexExpression:
-		return rootVariableOfExpression(expr.TargetExpression)
-	default:
-		return nil
+	for true {
```
dreamsmasher, line 377: for

runtime/storage_test.go
```diff
@@ -2326,29 +2326,32 @@ func TestRuntimeReferenceOwnerAccess(t *testing.T) {
...
 
                   // At this point the resource is not in storage
-                  log(ref.owner?.address)
+                  log(ref1.owner?.address)
 
                   accountA.save(<-testResource, to: /storage/test)
 
                   // At this point the resource is in storage A
-                  log(ref.owner?.address)
+                  accountA.link<&TestContract.TestResource>(/public/test, target: /storage/test)
```
turbolent, line 2337: Good idea to use storage references 👍

runtime/storage_test.go
```diff
@@ -2326,29 +2326,32 @@ func TestRuntimeReferenceOwnerAccess(t *testing.T) {
...
+                  log(ref3.owner?.address)
 
                   accountB.save(<-testResource2, to: /storage/test)
 
+                  accountB.link<&TestContract.TestResource>(/public/test, target: /storage/test)
+                  let ref4 = accountB.getCapability<&TestContract.TestResource>(/public/test).borrow()!
+
                   // At this point the resource is in storage B
-                  log(ref.owner?.address)
-                  log(ref2.owner?.address)
```
turbolent, line 2351: Given that `ref2` is a storage reference, it should be possible to keep using it here

runtime/sema/check_variable_declaration.go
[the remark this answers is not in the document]
turbolent: Ah, good! 👍

runtime/tests/interpreter/reference_test.go
turbolent: Nice!

turbolent (approved):
Great work!

runtime/storage_test.go
SupunS: it would give a dereference error, since the value is moved out of the storage

Issue #1388, opened 2022-01-28
Fetch all PublicAccount Keys


lukaracki:
### Issue To Be Solved

Allow fetching all keys from an account by doing something like `PublicAccount.keys`, currently this returns `{}`. The functionality is already present in flow-view-source and flow CLI (using the command `flow accounts get {addres}`.
Currently cadence only allows fetching individual keys with `PublicAccount.keys.get(index)`.

### Suggested Solution

Add a  method to the PublicAccount.keys type which returns an array of all keyIds or key objects assigned to a PublicAccount.

turbolent:
Adding support for this should be possible. It would need some dedicated export code for `SimpleCompositeValue` where the type ID is `PublicAccount.Keys` in `runtime.exportSimpleCompositeValue`

turbolent:
In addition, it might make sense to add a `let count: Int` field to the `*Account.Keys` types

dreamsmasher:
do we want to make visible the total number of keys in an AuthAccount?

also, what should the signature of this look like? we could return an entire array or do something similar to https://github.com/onflow/cadence/issues/1938, with short-circuiting iteration.

so the first option would either be a function that collects all keys, or a field member that provides a view into an underlying keys array:

```cadence
fun {Auth,Public}Account.keys.getAll(): [AccountKey]
```

and the second would be similar to our storage path iteration API's, e.g. `PublicAccount.forEachPublic`, 

```cadence
fun {Auth,Public}Account.keys.iter(_ f: (AccountKey: Bool)): Void
```

turbolent:
@dreamsmasher 
> do we want to make visible the total number of keys in an AuthAccount?
Good idea. Maybe `let length: Int`, like on dictionaries and arrays?

> also, what should the signature of this look like? we could return an entire array or do something similar to #1938, with short-circuiting iteration.
> 
> so the first option would either be a function that collects all keys, or a field member that provides a view into an underlying keys array:
> 
> ```cadence
> fun {Auth,Public}Account.keys.getAll(): [AccountKey]
> ```
> 
> and the second would be similar to our storage path iteration API's, e.g. `PublicAccount.forEachPublic`,
> 
> ```cadence
> fun {Auth,Public}Account.keys.iter(_ f: (AccountKey: Bool)): Void
> ```

We might want have the API similar to the recently added storage path iteration functionality (see https://developers.flow.com/cadence/language/accounts#publicaccount):

[...]

PR #2038, opened 2022-10-06
Add key iteration, count methods to AccountKeyProvider interface

dreamsmasher:
Part of #1388

## Description

This PR introduces 2 new methods to the `AccountKeyProvider` interface, and by extension, the overarching `Interface` interface for interpreter runtimes:

```go
IterateKeys(address common.Address, fn func(*AccountKey) bool) error
AccountKeysCount(address common.Address) *big.Int
```

These methods are necessary in order to implement key iteration in Cadence, bindings in the interpreter are currently stubbed out pending a companion PR to flow-go.
______


runtime/interface.go
```diff
@@ -82,6 +83,8 @@ type Interface interface {
 	AddAccountKey(address Address, publicKey *PublicKey, hashAlgo HashAlgorithm, weight int) (*AccountKey, error)
 	// GetAccountKey retrieves a key from an account by index.
 	GetAccountKey(address Address, index int) (*AccountKey, error)
+	AccountKeysCount(address Address) *big.Int
```
turbolent, line 86: Return `uint64` instead. It's pretty much impossible an account will have more keys than that

runtime/interpreter/accountkeys.go
```diff
@@ -37,12 +37,18 @@ func NewAuthAccountKeysValue(
...
+	forEachFunction FunctionValue,
+	countFunction FunctionValue,
 ) Value {
 
 	fields := map[string]Value{
 		sema.AccountKeysAddFunctionName:    addFunction,
 		sema.AccountKeysGetFunctionName:    getFunction,
 		sema.AccountKeysRevokeFunctionName: revokeFunction,
+
+		// TODO add back after flow-go is updated
```
turbolent, line 49: The update path is that Cadence introduces new requirements (functions in the runtime interface), and embedders, like flow-go, need to implement the new functions. You can uncomment the lines here and below

runtime/sema/authaccount_type.go
```diff
@@ -690,6 +703,36 @@ var AccountKeysTypeGetFunctionType = &FunctionType{
 	RequiredArgumentCount: RequiredArgumentCount(1),
 }
 
+// fun forEach(_ function: (AccountKey: Bool)): Void
```
turbolent, line 706:
```suggestion
// fun forEach(_ function: ((AccountKey): Bool)): Void
```

runtime/sema/authaccount_type.go
```diff
@@ -690,6 +703,36 @@ var AccountKeysTypeGetFunctionType = &FunctionType{
...
+var AccountKeysTypeForEachFunctionType = func() *FunctionType {
+	// fun (_ key: AccountKey): Bool
+	iterFunctionType := &FunctionType{
+		Parameters: []*Parameter{
+			{
+				Identifier:     "accountKey",
```
turbolent, line 712: Maybe just remove
```suggestion
```

runtime/sema/authaccount_type.go
```diff
@@ -690,6 +703,36 @@ var AccountKeysTypeGetFunctionType = &FunctionType{
 	RequiredArgumentCount: RequiredArgumentCount(1),
 }
 
+// fun forEach(_ function: (AccountKey: Bool)): Void
+var AccountKeysTypeForEachFunctionType = func() *FunctionType {
+	// fun (_ key: AccountKey): Bool
```
turbolent, line 708:
```suggestion
	// ((AccountKey): Bool)
```

runtime/sema/authaccount_type.go
```diff
@@ -690,6 +703,36 @@ var AccountKeysTypeGetFunctionType = &FunctionType{
...
+			},
+		},
+		ReturnTypeAnnotation: NewTypeAnnotation(VoidType),
+	}
+}()
+
+var AccountKeysTypeCountFunctionType = &FunctionType{
+	Parameters:           []*Parameter{},
+	ReturnTypeAnnotation: NewTypeAnnotation(IntType),
+}
```
turbolent, line 734: This is a field, not a function

```suggestion
var AccountKeysTypeCountFieldType = IntType
```

runtime/sema/publicaccount_type.go
```diff
@@ -189,6 +189,19 @@ var PublicAccountKeysType = func() *CompositeType {
...
+		// NewUnmeteredPublicFunctionMember(
+		// 	accountKeys,
+		// 	AccountKeysForEachFunctionName,
+		// 	AccountKeysTypeForEachFunctionType,
+		// 	accountKeysTypeForEachFunctionDocString,
+		// ),
+		// NewUnmeteredPublicConstantFieldMember(
+		// 	accountKeys,
+		// 	AccountKeysCountFieldName,
+		// 	AccountKeysTypeCountFunctionType,
```
turbolent, line 202:
```suggestion
		// 	AccountKeysTypeCountFieldType,
```

runtime/interface.go
```diff
@@ -82,6 +83,8 @@ type Interface interface {
...
+	IterateKeys(address Address, fn func(*AccountKey) bool) error
```
turbolent, line 87: Use `GetAccountKey` instead

```suggestion
```

runtime/sema/publicaccount_type.go
```diff
@@ -189,6 +189,19 @@ var PublicAccountKeysType = func() *CompositeType {
 			AccountKeysTypeGetFunctionType,
 			accountKeysTypeGetFunctionDocString,
 		),
+		// TODO add back after flow-go is updated
+		// NewUnmeteredPublicFunctionMember(
+		// 	accountKeys,
+		// 	AccountKeysForEachFunctionName,
+		// 	AccountKeysTypeForEachFunctionType,
+		// 	accountKeysTypeForEachFunctionDocString,
```
turbolent, line 197: Inconsistent naming, use `AccountKeysTypeForEach` everywhere

runtime/stdlib/account.go
```diff
@@ -658,21 +668,116 @@ func newAccountKeysGetFunction(
...
 	)
 }
 
+func newAccountKeysForEachFunction(
+	gauge common.MemoryGauge,
+	provider AccountKeyProvider,
+	addressValue interpreter.AddressValue,
+) *interpreter.HostFunctionValue {
+	address := addressValue.ToAddress()
+	// the PublicKey in `forEachKey(_ f: (PublicKey): Bool): Void`
```
turbolent, line 685:
```suggestion
	// the AccountKey in `forEach(_ function: ((AccountKey): Bool)): Void`
```

turbolent (commented):
Also, please add tests and documentation

runtime/stdlib/account.go
```diff
@@ -658,21 +668,116 @@ func newAccountKeysGetFunction(
...
+					inter,
+					getLocationRange,
+					key,
+					assumePublicKeyIsValid,
+				)
+			}
+
+			var invocationErr error
+
+			err := provider.IterateKeys(address, func(key *AccountKey) bool {
```
dsainati1, line 722: This can probably be implemented by using `GetAccountKey` on each of the indices from 0 to `count` rather than relying on the FVM to implement another function, we want to keep the runtime interface as minimal as possible.

runtime/sema/authaccount_type.go
dreamsmasher: manually writing rank-2 functions as an AST is confusing

dreamsmasher:
> Also, please add tests and documentation

ah yeah sorry, probably should opened this as a draft first

turbolent:
> probably should opened this as a draft first

My bad, sorry, I didn't see this was only supposed to be a part of the issue!

runtime/stdlib/account.go
```diff
@@ -670,21 +678,117 @@ func newAccountKeysGetFunction(
...
+
+			var index uint64
+			for index = 0; index < count; index++ {
+				wrapPanic(func() {
+					accountKey, err = provider.GetAccountKey(address, int(index))
+				})
+				if err != nil {
+					// Here it is expected the host function to return a nil key, if a key is not found at the given index.
+					// This is done because, if the host function returns an error when a key is not found, then
+					// currently there's no way to distinguish between a 'key not found error' vs other internal errors.
```
turbolent, line 743: This should go above the if-statement below, like in `newAccountKeysGetFunction`

turbolent (commented):
Looks good! A test is failing, not quite sure why

runtime/stdlib/account.go
dreamsmasher: ah yeah, been trying to figure out the failing test to no avail. meeting with @dsainati1 tomorrow to discuss it.

the failure stems from `exportValueWithInterpreter`, where the runtime is attempting to resolve a `count` field on the parent AuthAccount's `inbox` type. I think scoping and/or aliasing is causing the inbox type's members to be mutated during repeated nested calls to `exportCompositeType`, but idk where exactly.

turbolent: Huh? Very odd the inbox changes related to this. I'll have a look as well

turbolent: The Account Inbox PR #1983 had a couple mistakes, I'll open a PR to fix them and add tests to prevent future similar ones

turbolent: Opened #2053 which fixes the mistakes and unblocks this PR. Let's get that one in first, and then once it is in, we can merge master into this and CI should pass (it does locally). There's nothing left to do for this PR, it should resolve itself. Please have a look at the bugfix PR though.

runtime/stdlib/account.go
```diff
@@ -670,21 +678,117 @@ func newAccountKeysGetFunction(
...
+
+			count := provider.AccountKeysCount(address)
+
+			var err error
+			var accountKey *AccountKey
+
+			var index uint64
+			for index = 0; index < count; index++ {
+				wrapPanic(func() {
+					accountKey, err = provider.GetAccountKey(address, int(index))
```
turbolent, line 738: Maybe simplify `uint64` <> `int` usage here

turbolent (approved):
LGTM! Nice

runtime/account_test.go
```diff
@@ -350,34 +344,50 @@ var revokedAccountKeyA = func() *stdlib.AccountKey {
...
+		rt,
+		rtInterface,
+	}
+}
+
 func TestRuntimeAuthAccountKeys(t *testing.T) {
 
 	t.Parallel()
 
+	initTestEnvironment := func(t *testing.T) accountTestEnvironment {
```
turbolent, line 371: Nice reactor/cleanup 👍

dsainati1 (approved):
👌

runtime/stdlib/account.go
```diff
@@ -670,21 +678,117 @@ func newAccountKeysGetFunction(
...
 }
 
+func newAccountKeysForEachFunction(
+	gauge common.MemoryGauge,
+	provider AccountKeyProvider,
+	addressValue interpreter.AddressValue,
+) *interpreter.HostFunctionValue {
+	address := addressValue.ToAddress()
+	// the AccountKey in `forEachKey(_ f: ((AccountKey): Bool)): Void`
+	callbackTypeParams := []sema.Type{sema.AccountKeyType}
```
SupunS, line 696: can make this a module level variable

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -9062,26 +9064,24 @@ func newTestAuthAccountValue(gauge common.MemoryGauge, addressValue interpreter.
...
+				panicFunctionValue,
+				panicFunctionValue,
+				interpreter.AccountKeysCountConstructor(func() interpreter.UInt64Value {
+					panic(errors.NewUnreachableError())
+				}),
 			)
 		},
 		func() interpreter.Value {
-			return interpreter.NewAuthAccountInboxValue(gauge, addressValue, panicFunction, panicFunction, panicFunction)
+			return interpreter.NewAuthAccountInboxValue(gauge, addressValue, panicFunctionValue, panicFunctionValue, panicFunctionValue)
```
SupunS, line 9077:
```suggestion
			return interpreter.NewAuthAccountInboxValue(
				gauge,
				addressValue,
				panicFunctionValue,
				panicFunctionValue,
				panicFunctionValue,
			)
```

SupunS (approved):

runtime/stdlib/account.go
dreamsmasher: hmm, this is only used in one place, don't think it's worth polluting the module namespace

turbolent: Avoids allocating a fresh slice, so maybe worth it

runtime/stdlib/account.go
```diff
@@ -670,21 +678,116 @@ func newAccountKeysGetFunction(
...
+			var accountKey *AccountKey
+
+			for index := 0; index < count; index++ {
+				wrapPanic(func() {
+					accountKey, err = provider.GetAccountKey(address, index)
+				})
+				if err != nil {
+					panic(err)
+				}
+
```
turbolent, line 742: Just noticed this while reviewing the doc additions. 

Is there a particular reason why revoked keys are filtered out?
We might want to either remove the filtering or make it more obvious in the function name.

Not sure if it already does, but would be good for AccountKey to have an index.

dreamsmasher: oh, I just made an assumption based on how I would expect key iteration to behave as a smart contract developer. probably should have checked in on that. @dsainati1 @SupunS any thoughts?

AccountKeys do have an index, btw

turbolent: I don't really know the answer here, it would be good to get some more input. I have added this as a topic to the next LDM. In the meantime we can maybe leave this PR open, there is no rush to get it in. Once a feature is shipped it is much harder to change

SupunS: hmm, given that `keys.get()` would also return revoked keys, I feel this should also include revoked keys. 
If not, like Bastian has said, we might want to rename the method to be more explicit about it.

I'm biased toward the first option because it supports both uses-cases (users can filter out if they want).

dreamsmasher: ok, posted a poll on slack also and the overwhelming consensus is for revoked keys to be included. removed the filtering

PR #2041, opened 2022-10-06
Lint with gofmt 1.19, upgrade to go 1.19

dreamsmasher:
## Description

Just a housekeeping task since we've been putting off switching to gofmt 1.19 for a while. Most formatting is preserved, save for trailing line comments. Only 1 doc comment had to be manually edited in runtime/sema/type.go (IsSubType).

______

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/onflow/cadence
 
-go 1.19
+go 1.18
 
 require (
 	github.com/bits-and-blooms/bitset v1.2.2
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/onflow/cadence
 
-go 1.18
+go 1.19
 
 require (
 	github.com/bits-and-blooms/bitset v1.2.2
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -4971,19 +4971,19 @@ func (t *AddressType) GetMembers() map[string]MemberResolver {
 //
 // The differences between these methods is as follows:
 //
-//	IsSubType()
+//   - IsSubType():
 //
-// To check the assignability, e.g: is argument type T is a sub-type
-// of parameter type R? This is the more frequent use-case.
+//     To check the assignability, e.g: is argument type T is a sub-type
+//     of parameter type R? This is the more frequent use-case.
 //
-//	IsSameTypeKind()
+//   - IsSameTypeKind():
 //
-// To check if a type strictly belongs to a certain category. e.g: Is the
-// expression type T is any of the integer types, but nothing else.
-// Another way to check is, asking the question of "if the subType is Never,
-// should the check still pass?". A common code-smell for potential incorrect
-// usage is, using IsSubType() method with a constant/pre-defined superType.
-// e.g: IsSubType(<<someType>>, FixedPointType)
+//     To check if a type strictly belongs to a certain category. e.g: Is the
+//     expression type T is any of the integer types, but nothing else.
+//     Another way to check is, asking the question of "if the subType is Never,
+//     should the check still pass?". A common code-smell for potential incorrect
+//     usage is, using IsSubType() method with a constant/pre-defined superType.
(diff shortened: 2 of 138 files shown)


runtime/sema/type.go
```diff
@@ -4982,15 +4968,22 @@ func (t *AddressType) GetMembers() map[string]MemberResolver {
...
+// of parameter type R? This is the more frequent use-case.
+//
+//	IsSameTypeKind()
+//
+// To check if a type strictly belongs to a certain category. e.g: Is the
+// expression type T is any of the integer types, but nothing else.
+// Another way to check is, asking the question of "if the subType is Never,
+// should the check still pass?". A common code-smell for potential incorrect
+// usage is, using IsSubType() method with a constant/pre-defined superType.
+// e.g: IsSubType(<<someType>>, FixedPointType)
```
dreamsmasher, line 4986: specifically here, this now renders as code blocks followed by description paragraphs.

SupunS, line 4986: The default automatic formatting for this one seems ok to me:
```suggestion
//   - IsSubType()      - To check the assignability. e.g: Is argument type T is a sub-type
//     of parameter type R. This is the more frequent use-case.
//   - IsSameTypeKind() - To check if a type strictly belongs to a certain category. e.g: Is the
//     expression type T is any of the integer types, but nothing else.
//     Another way to check is, asking the question of "if the subType is Never,
//     should the check still pass?". A common code-smell for potential incorrect
//     usage is, using IsSubType() method with a constant/pre-defined superType.
//     e.g: IsSubType(<<someType>>, FixedPointType)
```

SupunS (commented):
Thank you for cleaning these up. It was a bit annoying when gofmt updated random places in random PRs. 🙏

runtime/sema/type.go
dreamsmasher: @turbolent thoughts? this was one example you specifically cited as being broken in a previous PR formatted with gofmt 1.19

dreamsmasher:
need to upgrade to go 1.19 because of gofmt.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/onflow/cadence
 
-go 1.18
+go 1.19
```
turbolent, line 3: This indicates the minimum version of Go. We should be able to keep requiring Go 1.18

```suggestion
go 1.18
```

runtime/sema/type.go
turbolent, line 4986: Is it possible to indent the descriptions under the function names? 

Not tested, but just to show what I mean:

```suggestion
//	* IsSubType():
//
//     To check the assignability, e.g: is argument type T is a sub-type
//     of parameter type R? This is the more frequent use-case.
//
//	* IsSameTypeKind():
//
//     To check if a type strictly belongs to a certain category. e.g: Is the
//     expression type T is any of the integer types, but nothing else.
//     Another way to check is, asking the question of "if the subType is Never,
//     should the check still pass?". A common code-smell for potential incorrect
//     usage is, using IsSubType() method with a constant/pre-defined superType.
//     e.g: IsSubType(<<someType>>, FixedPointType)
```

turbolent (commented):
LGTM, just one comment we could maybe improve

turbolent:
Only CI should be updated, the packages themselves should stay at 1.18

runtime/sema/type.go
turbolent: Ah, I missed this thread while reviewing, but had also commented on it here: https://github.com/onflow/cadence/pull/2041#discussion_r989524751

go.mod
dreamsmasher: wouldn't that create issues for contributors since they'll fail ci-lint?

runtime/sema/type.go
dreamsmasher: Ah yeah, but the asterisks are replaced by hyphens after formatting.

turbolent (approved):
Nice!

SupunS (approved):

dreamsmasher:
YESSSSS

PR #2052, opened 2022-10-11
Allow export of function values

### LINKED ISSUE #2049: Extend exportable types
### Issue To Be Solved

Not sure if I called it correct, but functions/references etc cannot be exported causing some problems. 

For example NFTView, FTView on metadata, has function field, which causes script errors. 

Also I am developing some Jupyter/debugger integration for mainnet/testnet [0], it would be nice if everything can be exported, preferably can be important also in script environment. 

[0] https://www.loom.com/share/afed3b09cbd449dbb150d075de8d2087

turbolent:
Closes #2049

## Description

Event though functions cannot be serialized, we should still treat them as exportable: values are simply omitted.

This allows the export of composite values that contain function-typed fields. 
For example, this [the NFT metadata standard contains such a definition](https://github.com/onflow/flow-nft/blob/9877dccbd884e06bbd754a4f693b7169dd6dd62b/contracts/MetadataViews.cdc#L514).

Also, improve the tests by changing the static return type to `AnyStruct` and asserting the dynamic type checked for exportability.

______


runtime/sema/type.go
```diff
@@ -2718,8 +2718,10 @@ func (t *FunctionType) IsStorable(_ map[*Member]bool) bool {
 }
 
 func (t *FunctionType) IsExternallyReturnable(_ map[*Member]bool) bool {
-	// Functions cannot be exported, as they cannot be serialized
-	return false
+	// Event though function cannot be serialized,
```
SupunS, line 2721:
```suggestion
	// Even though function cannot be serialized,
```

runtime/runtime_test.go
```diff
@@ -2531,169 +2537,125 @@ func TestRuntimeScriptReturnTypeNotReturnableError(t *testing.T) {
...
 
 		test(t,
-			`
-              pub fun main(): PrivatePath {
-                  return /private/foo
-              }
-            `,
-			cadence.Path{
-				Domain:     "private",
-				Identifier: "foo",
```
SupunS, line 2662: why are these removed?

runtime/import_test.go
```diff
@@ -114,37 +114,3 @@ func TestRuntimeCyclicImport(t *testing.T) {
 
 	require.IsType(t, &sema.CyclicImportsError{}, errs[0])
 }
-
-func TestRuntimeExport(t *testing.T) {
```
turbolent, line 118: Already tested by `TestExportValue`

runtime/runtime_test.go
turbolent: I noticed that the export of paths is already tested in `TestExportValue`, so removed the duplicates here.

SupunS: ah ok 👌

SupunS (approved):

dsainati1:
Just omitting the value seems suspect to me; it gives no information about what was originally present on that field. Can we instead emit some kind of dummy value that indicates that there was a function there originally?

dsainati1:
That or some text saying "Function Value Omitted", for example

turbolent:
@dsainati1 Added a `cadence.Function` value which includes the function's type in [acaef39](https://github.com/onflow/cadence/pull/2052/commits/acaef392f5718a37721aaa7bd8f50bab3f5aeaf4)

dsainati1:
Nice. We should also update the json spec as well.

turbolent:
@dsainati1 Added documentation in ae1c38c16cb9e8aafe684f032c14b3b3a7367d28

dsainati1 (approved):
Nice, LGTM!

SupunS (approved):

PR #2053, opened 2022-10-12
Fix account inbox implementation

turbolent:
## Description

- Public accounts do not have an inbox, remove related code
- Add missing `inbox` field name to AuthAccount type
- Fix static type of AuthAccount.Inbox
- Add missing mappings between sema and primitive static types
- Add tests to avoid similar omissions for future primitive static types

______


runtime/interpreter/account.go
```diff
@@ -33,6 +33,7 @@ var authAccountFieldNames = []string{
 	sema.AuthAccountAddressField,
 	sema.AuthAccountContractsField,
 	sema.AuthAccountKeysField,
+	sema.AuthAccountInboxField,
```
dsainati1, line 36: Confused by this; what effect does this actually have given that none of the interpreter tests failed without this? I would have expected not to be able to access the `inbox` field without specifying this.

runtime/interpreter/accountinbox.go
```diff
@@ -28,7 +28,7 @@ import (
 // AuthAccountInbox
 
 var authAccountInboxTypeID = sema.AuthAccountInboxType.ID()
-var authAccountInboxStaticType StaticType = PrimitiveStaticTypeAuthAccountKeys
+var authAccountInboxStaticType StaticType = PrimitiveStaticTypeAuthAccountInbox
```
dsainati1, line 31: Oops! Thanks for catching this

dsainati1 (approved):
Sorry about this; guess I forgot to remove some of the public account stuff after we changed the proposal in the FLIP again. There are definitely benefits to implementing the FLIPs as they are being designed as a prototype of sorts, but the downside is it gets a lot harder to keep track of what actually belongs in the implementation PRs 😅 .

turbolent:
No worries! I had missed these things in the review too and it's easy to fix 🙂 
We probably need more checks to prevent such accidents, not quite sure yet how though

runtime/interpreter/account.go
turbolent: See the usage of `FieldNames` in `SimpleCompositeValue`. For example, they are used to iterate over the fields, prepare a string representation, check static type conformance, etc

runtime/convertValues.go
```diff
@@ -529,9 +529,7 @@ func exportSimpleCompositeValue(
 		structure, err := cadence.NewMeteredStruct(
 			inter,
 			len(fieldNames),
-			func() ([]cadence.Value, error) {
-				return makeFields()
-			},
+			makeFields,
```
dreamsmasher, line 532: nice eta reduce

runtime/interpreter/accountinbox.go
dreamsmasher: AAAA that's where my failing test was coming from, I knew it stemmed from static type resolution *somewhere*

dreamsmasher (approved):

PR #2064, opened 2022-10-13
Simplify parser

dreamsmasher:
## Description

Quick PR that abstracts away some common patterns in the parser, namely:

`parser.skipSpaceAndComments(skipNewline bool)` is almost always called with `true`, with only 1 exception in parsing return statements. this PR simplifies it to always skip a newline since we often want to just skip past all whitespace until the next semantic token.

`parser.next()` followed by `parser.skipSpaceAndComments(true)` shows up a lot (a few dozen times), when you want to advance past the currently-read token to the next semantically significant one. Added a new method `parser.nextSemanticToken()` which combines these and aids clarity.

I honestly feel that reading `parser.Current` and manually advancing with `parser.next()` is a bit of an antipattern:  `parser.next()` should advance to the next token and return it directly, similar to an iterator over tokens. This is more in line with the behaviour of parser combinator libraries in other languages, but that refactor would be a more involved effort.
______

### DIFF
--- a/runtime/parser/parser.go
+++ b/runtime/parser/parser.go
@@ -231,6 +231,13 @@ func (p *parser) next() {
 	}
 }
 
+// utility for the common pattern of advancing past the current token to
+// the next semantic token
+func (p *parser) nextSemanticToken() {
+	p.next()
+	p.skipSpaceAndComments()
+}
+
 func (p *parser) mustOne(tokenType lexer.TokenType) (lexer.Token, error) {
 	t := p.current
 	if !t.Is(tokenType) {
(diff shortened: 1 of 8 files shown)


runtime/parser/parser.go
```diff
@@ -231,6 +231,13 @@ func (p *parser) next() {
 	}
 }
 
+// utility for the common pattern of advancing past the current token to
+// the next semantic token
```
turbolent, line 235: By convention, Go documentation comments should start with the name of the declaration:

```suggestion
// nextSemanticToken advances past the current token to the next semantic token.
// It skips whitespace, including newlines, and comments
```

runtime/parser/parser.go
```diff
@@ -383,11 +390,12 @@ type triviaOptions struct {
 	parseDocStrings bool
 }
 
-func (p *parser) skipSpaceAndComments(skipNewlines bool) (containsNewline bool) {
+// utility for the common pattern of skipping past whitespace to the next semantic token
```
turbolent, line 393:
```suggestion
// skipSpaceAndComments skips whitespace, including newlines, and comments
```

turbolent (approved):
Nice! Thank you for cleaning this up 👌

runtime/parser/statement.go
```diff
@@ -207,7 +205,9 @@ func parseReturnStatement(p *parser) (*ast.ReturnStatement, error) {
 	endPosition := tokenRange.EndPos
 	p.next()
 
-	sawNewLine := p.skipSpaceAndComments(false)
```
dreamsmasher, line 210: this is the only place where we skip whitespace excluding newlines, so it's simpler to just inline the call here instead.

SupunS (approved):

Issue #1883, opened 2022-08-17
Cadence Interpreter Optimizations


j1010001:
This epic tracks investigations and performance improvements of the tree-walking interpreter, for example by compiling to functions. Interpreters can potentially also be reused and certain operations can be performed concurrently.

turbolent:
- #1899
- #1888
- #1831
- #1825
- #1813

turbolent:
@j1010001 I think that's all improvements we have so far. OK to close? (We can create another issue/epic when we do more)

PR #2065, opened 2022-10-13
Make the interpreter globals map allocation lazy

turbolent:
Work towards #1883 

## Description

Reduce memory usage in the interpreter.

Instead of always allocating the globals map at interpreter construction time, delay it until a global is set.
This should reduce allocations for the case where the interpreter is used without a program, e.g. only for storage operations.

______

### DIFF
--- a/runtime/interpreter/globalvariables.go
+++ b/runtime/interpreter/globalvariables.go
@@ -19,18 +19,28 @@
 package interpreter
 
 // GlobalVariables represents global variables defined in a program.
-type GlobalVariables map[string]*Variable
+type GlobalVariables struct {
+	variables map[string]*Variable
+}
 
-func (globalVars GlobalVariables) Contains(name string) bool {
-	_, ok := globalVars[name]
+func (g *GlobalVariables) Contains(name string) bool {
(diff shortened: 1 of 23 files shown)


runtime/interpreter/globalvariables.go
```diff
@@ -19,18 +19,28 @@
...
+		return nil
+	}
+	return g.variables[name]
 }
 
-func (globalVars GlobalVariables) Set(name string, variable *Variable) {
-	globalVars[name] = variable
+func (g *GlobalVariables) Set(name string, variable *Variable) {
+	if g.variables == nil {
+		g.variables = map[string]*Variable{}
```
dreamsmasher, line 43: should we preallocate some size on initialization, given that if one global variable is set, there are likely to be more?

dreamsmasher (approved):

SupunS (approved):

runtime/interpreter/globalvariables.go
turbolent: Yes, we could consider pre-allocating more than the default here. A "good" value is usually quite difficult to determine

PR #2079, opened 2022-10-19
Improve type inference for case expressions in switch statement

### LINKED ISSUE #2077: Case expressions in switch statement doesn't infer type
### Problem

```cadence
fun test(): Int {
    let x: UInt8 = 5

    switch x {
        case 1:    // error: cannot apply binary operation == to types: `UInt8`, `Int`
            return 1
        default:
            return 2
    }
}
```

Should be able to infer the type of the case-expression based on the type of the condition-expression (`x` in this case)

SupunS:
Closes #2077

## Description

Pass the type of condition-expression as the expected-type of the case-expressions.

______

### DIFF
--- a/runtime/sema/check_switch.go
+++ b/runtime/sema/check_switch.go
@@ -87,7 +87,12 @@ func (checker *Checker) checkSwitchCaseExpression(
 	testTypeIsValid bool,
 ) {
 
-	caseType := checker.VisitExpression(caseExpression, nil)
+	var caseExprExpectedType Type
+	if testTypeIsValid {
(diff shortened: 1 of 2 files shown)


turbolent (approved):
Nice! 

Only question/concern: Is this a breaking change?

dreamsmasher (approved):

SupunS:
@turbolent It shouldn't be breaking. This is more of a relaxation of existing rules. i.e: case expressions weren't supporting for subtypes (int subtypes, character, etc) - they were always treated as their supertype counterpart. But now they are treated as the subtype.

Issue #2061, opened 2022-10-13
Implement Typechecking for Attachments


dsainati1:

turbolent:
- First pass / bulk is done
- Small chance nothing else needs to be done (maybe naming)
- Needs more tests

Issue #357, opened 2020-08-28
Cadence Contract Extensibility


alxocity:
### Issue To Be Solved

Ability to extend a resource without updating the resource contract.

### Suggest A Solution

Essentially this, but in Cadence/Flow:
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods

### Context

This would allow adding Autographs to NBA TopShot moments, and generically adding resources to other resources without editing the parent resource contract. This would open up full personalization for anyone's NFTs. https://discord.gg/mCtsWyS

bjartek:
There are lots of interesting use cases that could emerge here. 

- Signing (like moment i guess)?
- Original Artist
- Suggested Original Artist resell cut (a marketplace could use this to send a small amount of cash to the oringal artist on resale)
- Ownership trail

alxocity:
Here's an example of modifying the Top Shot contract to support autographs that would be preferably done via extensions:
https://github.com/Flow-Moment-Company/flow-moment-contracts/blob/master/contracts/nba/TopShot.cdc

alxocity:
This commit in particular highlights the autographs extensibility that could instead be done outside of the Top Shot contract
https://github.com/Flow-Moment-Company/flow-moment-contracts/commit/5924d24444990e8ec06c2f9101769610dd6ded85

bjartek:
Could mixins be used to model this? If a mixin is stored in another account? And you could say i want 

`Topshot.NFT with Moment.Authograph` 

There has to be some guidelines to what could be done in a mixin then i guess. Maybe you should only get a reference to the parent.

robmyers:
An important issue with smart contract code rather than application code is that extending it can be used to introduce attacks. While extension methods cannot replace existing methods and perform a person-in-the-middle attack, they can add new functions that use existing code and member variables in ways that lead to undesirable results for the user. IMO it's important to look at what any proposal takes from the security of the system as well as what it adds to its expressiveness.

turbolent:
Thank you @alxocity and everyone else for starting this discussion! Extensions have been on our roadmap since the beginning, and we're excited to work on this.

I had a call with @alxocity last Friday, and started a new topic in the forum to discuss this further: https://forum.onflow.org/t/extensibility/622 

Please have a look and please share your thoughts! Hopefully we can discuss this together, come up with a proposal, and get it implemented 🚀 

[...]

PR #2082, opened 2022-10-20
Type Checking for Attachment Declarations

dsainati1:
Part of https://github.com/onflow/cadence/issues/357 and https://github.com/onflow/cadence/issues/2061

## Description

This PR implements type checking for attachments. Attachments are implemented as a new kind of composite type with some special additional functionality relating to its base type. In particular, in the body of an attachment's function members, we introduce a new `super` variable in addition to `self`, whose type is a reference to the base type of the attachment. 

A couple particular points to note:
* Attachments may not be interfaces
* Attachment base types may be structs, resources, struct interfaces, resource interfaces, `AnyStruct` or `AnyResource`. 
* Attachment types cannot be used directly, as attachments are not first-class values. They must be used through references, or Cadence will raise a static error
* Attachments may only be instantiated inside an `attach` expression, which will be implemented in a later PR
______


runtime/sema/check_composite_declaration.go
```diff
@@ -1938,29 +2097,42 @@ func (checker *Checker) checkCompositeFunctions(
...
+		checker.recordVariableDeclarationOccurrence(identifier, variable)
 	}
 }
 
+func (checker *Checker) declareSelfValue(selfType Type, selfDocString string) {
+	checker.declareLowerScopedValue(selfType, selfDocString, SelfIdentifier, common.DeclarationKindSelf)
+}
+
+func (checker *Checker) declareSuperValue(baseType Type, superDocString string) {
+	superType := NewReferenceType(checker.memoryGauge, baseType, false)
```
dsainati1, line 2132: References to `super` should be non-auth, as the actual base type in practice may be any number of subtypes of the annotated base type, not all of which should be available to the writer of the attachment.

robert-e-davidson3 (approved):

runtime/sema/check_composite_declaration.go
```diff
@@ -25,11 +25,74 @@ import (
 	"github.com/onflow/cadence/runtime/errors"
 )
 
+// while strictly speaking attachment declarations are not strictly the same thing as composite declarations,
+// they are subject to many of the same checks. In order to avoid duplicating these checks, we create a mapping
+// between an attachment declaration and its composite "equivalent" that we can use for these composite checks
+func (checker *Checker) attachmentAsComposite(declaration *ast.AttachmentDeclaration) *ast.CompositeDeclaration {
+	compositeDelcaration, ok := checker.Elaboration.AttachmentCompositeDeclarations[declaration]
```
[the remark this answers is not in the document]
dsainati1: There is also potential to refactor this so that the composite checking/interpreting functions operate on a supertype of both attachments and composites (perhaps an `CompositeLikeDeclaration`) type. Once these are merged I will look into a follow-up PR to refactor this, potentially improving performance.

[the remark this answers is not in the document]
dsainati1: https://github.com/onflow/cadence/pull/2148

runtime/sema/type.go
```diff
@@ -3522,6 +3528,31 @@ func (t *CompositeType) GetCompositeKind() common.CompositeKind {
 	return t.Kind
 }
 
+func (t *CompositeType) getBaseCompositeKind() common.CompositeKind {
+	if t.Kind != common.CompositeKindAttachment {
+		return t.Kind
```
SupunS, line 3533: Shouldn't this return a `common.CompositeKindUnknown`?

runtime/sema/type.go
```diff
@@ -3522,6 +3528,31 @@ func (t *CompositeType) GetCompositeKind() common.CompositeKind {
...
+	case *SimpleType:
+		if base == AnyResourceType {
+			return common.CompositeKindResource
+		} else if base == AnyStructType {
+			return common.CompositeKindStructure
+		}
+		return common.CompositeKindUnknown
+	default:
+		return common.CompositeKindUnknown
+	}
```
SupunS, line 3549:
```suggestion
	}
	
	return common.CompositeKindUnknown
```

runtime/tests/checker/attachments_test.go
```diff
@@ -0,0 +1,1486 @@
...
+	})
+
+	t.Run("conforms to base", func(t *testing.T) {
+
+		t.Parallel()
+
+		_, err := ParseAndCheck(t,
+			`
+			resource interface I {}
+			attachment A for I: I {}`,
```
SupunS, line 1021: This sounds like should be an error. Interfaces cannot conform to interfaces (yet).
Even when we add that, self-conforming would/should be rejected.

SupunS (commented):
Looks good!

runtime/tests/checker/attachments_test.go
dsainati1: This isn't saying that `I` conforms to `I`, it's saying that `A` conforms to `I` and also has a base of type `I`. So for example:
```
resource interface I {
    fun foo() {}
}
attachment A for I: I {
    fun foo() {
        base.foo() // delegate implementation to base
    }
}
```

SupunS: oh okay, I misread the code there

SupunS (approved):
I sort of skipped this part https://github.com/onnflow/cadence/pull/2082#discussion_r1000942822 since it is refactored in the other PR.

Rest of it looks good!

runtime/sema/check_composite_declaration.go
```diff
@@ -405,9 +497,26 @@ func (checker *Checker) declareNestedDeclarations(
 		nestedCompositeTypes = append(nestedCompositeTypes, nestedCompositeType)
 	}
 
+	// Declare nested attachments
+
+	for _, nestedDeclaration := range nestedAttachmentDeclaration {
+		if _, exists := nestedDeclarations[nestedDeclaration.Identifier.Identifier]; !exists {
+			nestedDeclarations[nestedDeclaration.Identifier.Identifier] = nestedDeclaration
+		}
```
turbolent, line 505: Maybe factor out the repeated `nestedDeclaration.Identifier.Identifier` into variable

runtime/sema/check_composite_declaration.go
```diff
@@ -405,9 +497,26 @@ func (checker *Checker) declareNestedDeclarations(
...
+		nestedCompositeType := checker.declareAttachmentType(nestedDeclaration)
+		nestedCompositeTypes = append(nestedCompositeTypes, nestedCompositeType)
+	}
+
 	return
 }
 
+func (checker *Checker) declareAttachmentType(declaration *ast.AttachmentDeclaration) *CompositeType {
+	composite := checker.declareCompositeType(checker.attachmentAsComposite(declaration))
+	composite.baseType = checker.convertNominalType(declaration.BaseType)
```
turbolent, line 516: How are we ensuring that the base type of the attachment declaration is always a nominal type?

runtime/sema/check_composite_declaration.go
```diff
@@ -1055,7 +1176,8 @@ func (checker *Checker) checkCompositeConformance(
 	// Ensure the composite kinds match, e.g. a structure shouldn't be able
 	// to conform to a resource interface
 
-	if interfaceType.CompositeKind != compositeType.Kind {
+	if interfaceType.CompositeKind != compositeType.Kind &&
+		interfaceType.CompositeKind != compositeType.getBaseCompositeKind() {
```
turbolent, line 1180: Should this be a logical or?

runtime/sema/check_composite_declaration.go
```diff
@@ -1410,6 +1540,37 @@ func (checker *Checker) checkTypeRequirement(
...
+	//    attachment A for RI { /* ... */ }
+	// }
+	//
+	// In this example, as long as `A` in `C` contains the expected member declarations as defined in `CI`, this is a valid
+	// implementation of the type requirement, as an `A` that can accept any `RI` as a base can clearly function for an `R` as well.
+	// It may also be helpful to conceptualize an attachment as a sort of implicit function that takes a `base` argument and returns a composite value.
+	if requiredCompositeType.Kind == common.CompositeKindAttachment && declaredCompositeType.Kind == common.CompositeKindAttachment {
+		if !IsSubType(requiredCompositeType.baseType, declaredCompositeType.baseType) {
+			checker.report(
+				&ConformanceError{
```
turbolent, line 1564: It might be good to extend `ConformanceError` and provide the reason for the non-conformance, i.e. the base type mismatch. See the other similar fields in `ConformanceError`.

A generic error that the concrete type does not conform to the type requirement is hard to debug, especially if the reason is that the base types of some attachment don't match.

runtime/sema/check_composite_declaration.go
```diff
@@ -1866,6 +2014,14 @@ func (checker *Checker) checkSpecialFunction(
 	defer checker.leaveValueScope(specialFunction.EndPosition, checkResourceLoss)
 
 	checker.declareSelfValue(containerType, containerDocString)
+	if containerType.GetCompositeKind() == common.CompositeKindAttachment {
+		// attachments cannot be interfaces, so this cast must succeed
```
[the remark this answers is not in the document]
turbolent: It might be nice to make this a semantic error instead, i.e. parse it, but report an error in the checker.

runtime/sema/check_composite_declaration.go
```diff
@@ -1937,29 +2109,42 @@ func (checker *Checker) checkCompositeFunctions(
 	}
 }
 
-func (checker *Checker) declareSelfValue(selfType Type, selfDocString string) {
-
-	// NOTE: declare `self` one depth lower ("inside" function),
+func (checker *Checker) declareLowerScopedValue(
```
turbolent, line 2112: Good idea to generalize this!

Maybe add a comment to this function explaining what "lower" means (inside), and provide example use cases (e.g. for self and base)

runtime/sema/check_composite_declaration.go
turbolent: Great comment, please add it to the code

runtime/sema/typeannotationstate.go
```diff
@@ -27,4 +27,5 @@ const (
 	TypeAnnotationStateValid
 	TypeAnnotationStateInvalidResourceAnnotation
 	TypeAnnotationStateMissingResourceAnnotation
+	TypeAnnotationStateDirectAttachmentTypeAnnotation
```
turbolent, line 30: Why is this needed and what does it mean?

runtime/sema/errors.go
```diff
@@ -3709,7 +3709,7 @@ func (e *InvalidEntryPointTypeError) Error() string {
 	)
 }
 
-// ImportedProgramError
+// ExternalMutationError
```
turbolent, line 3712: 👍

runtime/sema/errors.go
```diff
@@ -3741,3 +3741,43 @@ func (e *ExternalMutationError) SecondaryError() string {
...
+
+func (*InvalidBaseTypeError) isSemanticError() {}
+
+func (*InvalidBaseTypeError) IsUserError() {}
+
+func (e *InvalidBaseTypeError) Error() string {
+	return fmt.Sprintf(
+		"cannot use `%s` as the base type for attachment `%s`",
+		e.BaseType.String(),
+		e.Attachment.Identifier,
```
turbolent, line 3764:
```suggestion
		e.BaseType.QualifiedString(),
		e.Attachment.QualifiedString(),
```

runtime/sema/type.go
```diff
@@ -3522,6 +3528,29 @@ func (t *CompositeType) GetCompositeKind() common.CompositeKind {
...
+	case *CompositeType:
+		return base.Kind
+	case *InterfaceType:
+		return base.CompositeKind
+	case *SimpleType:
+		if base == AnyResourceType {
+			return common.CompositeKindResource
+		} else if base == AnyStructType {
+			return common.CompositeKindStructure
+		}
```
turbolent, line 3545: Maybe generalize to any simple type with 

```suggestion
		return base.CompositeKind()
```

And add a function 

```go
func (t *SimpleType) CompositeKind() common.CompositeKind {
	if t.IsResource {
		return common.CompositeKindResource
	} else {
		return common.CompositeKindStructure
	}
}
```

This will also allow us to easily extend simple types to be used for e.g. event declarations by refactoring this function, instead of having to look for all uses of all simple types and such "remote decision making"

runtime/sema/type.go
```diff
@@ -3578,7 +3607,11 @@ func (t *CompositeType) GetMembers() map[string]MemberResolver {
 }
 
 func (t *CompositeType) IsResourceType() bool {
-	return t.Kind == common.CompositeKindResource
+	return t.Kind == common.CompositeKindResource ||
+		// attachments are always the same kind as their base type
+		(t.Kind == common.CompositeKindAttachment &&
+			t.baseType != t &&
```
turbolent, line 3613: Why do we need to specifically look for this "self-reference"?

runtime/sema/type.go
```diff
@@ -4704,7 +4745,7 @@ func (*ReferenceType) IsEquatable() bool {
 	return true
 }
 
-func (*ReferenceType) TypeAnnotationState() TypeAnnotationState {
+func (t *ReferenceType) TypeAnnotationState() TypeAnnotationState {
```
turbolent, line 4748: The receiver variable isn't used in the body, maybe revert:

```suggestion
func (*ReferenceType) TypeAnnotationState() TypeAnnotationState {
```

turbolent (commented):
Awesome work! Looks good so far, still have to review `attachments_test.go`

runtime/sema/check_composite_declaration.go
dsainati1: This is enforced in the parser

runtime/sema/typeannotationstate.go
dsainati1: This is used in `checkTypeAnnotation` to check that naked attachment annotations are rejected statically. The idea is similar to the check that requires that resources have an `@` type.

runtime/sema/type.go
dsainati1: `attachment A for A {}` will cause a stack overflow otherwise, as we will continuously call into `isResourceType` over and over again.

runtime/sema/check_composite_declaration.go
dsainati1: I agree, but IMO this is a separate PR; adding reasons to conformance errors is more widespread a change than just attachments.

turbolent: Yes, in a separate PR is preferred 👍

runtime/sema/typeannotationstate.go
turbolent: Oh, I see! Could you please add the explanation and maybe an example as a comment to this new case? While at it, could you please do the same for the existing cases? 🙏 

For example: "Resource types must have a resource annotation (@). An annotation is invalid if the annotation is used for a non-resource type, e.g. `@Int`"

runtime/sema/check_composite_declaration.go
turbolent: Thank you for the explanation, I see now the cases and what is meant to be checked.

So does this check assume that for an attachment, the first part will always fail?

Maybe it's clearer to differentiate more clearly between the "traditional" case (non-attachment), and the attachment case?

runtime/sema/type.go
turbolent: I see! Could you please add this explanation and example as a comment to the code?

runtime/sema/check_composite_declaration.go
dsainati1: Yes; an attachment's `kind` is always `CompositeKindAttachment`, while the interface's should either be `CompositeKindResource` or `CompositeKindStruct`

turbolent (approved):
Great work!

PR #2090, opened 2022-10-24
Improve the definite return analysis

### LINKED ISSUE #1516: panics forced to be outside of if branch
### Problem

it seems if you have a panic in an else branch of a function and don't return in said branch, the contract cannot be deployed
this doesn't work and errors with `error: missing return statement`:
```
        pub fun mint(id: UInt64): @NFT {
            if id > 100 {
                return <- create NFT(initID: id)
            } else {
                panic("bad id")
            }
        }
```
this works:
```
        pub fun mint(id: UInt64): @NFT {
            if id > 100 {
                return <- create NFT(initID: id)
            } 
            panic("bad id")
        }
```

### Steps to Reproduce

https://play.onflow.org/f04d34f5-85e8-4101-8e3d-bb982d8bbd2a

turbolent:
Closes #1516

## Description

Improve the definite-return check by considering definite halts as definite returns.

______

### DIFF
--- a/runtime/sema/check_invocation_expression.go
+++ b/runtime/sema/check_invocation_expression.go
@@ -174,7 +174,7 @@ func (checker *Checker) checkInvocationExpression(invocationExpression *ast.Invo
 	if returnType == NeverType {
 		returnInfo := checker.functionActivations.Current().ReturnInfo
 		returnInfo.DefinitelyHalted = true
-		returnInfo.DefinitelyReturned = true
+		returnInfo.DefinitelyReturnedOrHalted = true
 	}
 
 	if isOptionalChainingResult {
--- a/runtime/sema/check_invocation_expression.go
+++ b/runtime/sema/check_invocation_expression.go
@@ -174,6 +174,7 @@ func (checker *Checker) checkInvocationExpression(invocationExpression *ast.Invo
 	if returnType == NeverType {
 		returnInfo := checker.functionActivations.Current().ReturnInfo
 		returnInfo.DefinitelyHalted = true
+		returnInfo.DefinitelyReturned = true
 	}
 
 	if isOptionalChainingResult {
--- a/runtime/sema/check_invocation_expression.go
+++ b/runtime/sema/check_invocation_expression.go
@@ -174,7 +174,7 @@ func (checker *Checker) checkInvocationExpression(invocationExpression *ast.Invo
 	if returnType == NeverType {
 		returnInfo := checker.functionActivations.Current().ReturnInfo
 		returnInfo.DefinitelyHalted = true
-		returnInfo.DefinitelyReturnedOrHalted = true
+		returnInfo.DefinitelyExited = true
 	}
 
 	if isOptionalChainingResult {
--- a/runtime/sema/return_info.go
+++ b/runtime/sema/return_info.go
@@ -41,6 +41,31 @@ type ReturnInfo struct {
 	// DefinitelyHalted indicates that (the branch of) the function
 	// contains a definite halt (a function call with a Never return type)
 	DefinitelyHalted bool
+	// DefinitelyReturnedOrHalted indicates that (the branch of)
+	// the function contains a definite return statement
+	// or a definite halt (a function call with a Never return type).
+	//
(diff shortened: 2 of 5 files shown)


runtime/sema/check_invocation_expression.go
```diff
@@ -174,6 +174,7 @@ func (checker *Checker) checkInvocationExpression(invocationExpression *ast.Invo
 	if returnType == NeverType {
 		returnInfo := checker.functionActivations.Current().ReturnInfo
 		returnInfo.DefinitelyHalted = true
```
dreamsmasher, line 176: doesn't this directly contradict the definition of `NeverType`?

dreamsmasher (approved):

runtime/sema/check_invocation_expression.go
SupunS: Instead of marking this as `returnInfo.DefinitelyReturned`, can the usages be updated to use `returnInfo.DefinitelyReturned || returnInfo.DefinitelyHalted` instead of `returnInfo.DefinitelyReturned`?

turbolent: @dreamsmasher 
> doesn't this directly contradict the definition of `NeverType`?

Can you elaborate? Not quite sure I follow

runtime/sema/check_invocation_expression.go
```diff
@@ -174,6 +174,7 @@ func (checker *Checker) checkInvocationExpression(invocationExpression *ast.Invo
...
+		returnInfo.DefinitelyReturned = true
```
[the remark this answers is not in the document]
dsainati1: Oh weird I thought I submitted this yesterday, but had not actually finished the review and left this comment on "Pending". That said, I agree with the others that this makes the concept of `DefinitelyReturned` a little confusing.

runtime/sema/check_invocation_expression.go
dreamsmasher: the wording is a bit confusing here, since ⊥ implies that it either doesn't halt or it panics. at the same time, the above comment mentions `Update the return info for invocations that do not return (i.e. have a `Never` return type)`. I agree with Daniel, we should either change the naming for these attributes or add comments explaining why we mark them as Definitely{Halted,Returned} here.

turbolent: Ah, I see how it's confusing, thank you for the explanation. I'll try tracking a combination of halted/returned, instead of treating one as resulting in the other.

turbolent:
@dsainati1 @dreamsmasher @SupunS Tried to refactor in 31131e5282de39eb407815d365b844a1bb03eb9d. Is that better?

runtime/sema/return_info.go
```diff
@@ -41,6 +41,31 @@ type ReturnInfo struct {
...
+	//       panic(...)
+	//
+	//       // DefinitelyReturned = false
+	//	     // DefinitelyHalted = true
+	//	     // DefinitelyReturnedOrHalted = true
+	//   }
+	//
+	//   // DefinitelyReturned = false
+	//   // DefinitelyHalted = false
+	//   // DefinitelyReturnedOrHalted = true
```
SupunS, line 67: the fact that `DefinitelyReturnedOrHalted != DefinitelyHalted || DefinitelyReturned` is killing me 😅

Should it be renamed to something like `DefinitelyExited`?

SupunS (approved):
Thanks for updating! LGTM!

dreamsmasher:
> @dsainati1 @dreamsmasher @SupunS Tried to refactor in 31131e5282de39eb407815d365b844a1bb03eb9d. Is that better?

LGTM! now we need a utility to truly verify whether an arbitrary program halts or not...

runtime/sema/return_info.go
dsainati1: Yea I like this idea

turbolent: Naming is hard, thank you for the suggestion! "exited" captures the idea well 👍 
Renamed in e26115ca91767fb78ca9913a4cc6137d09d45a4d

PR #2093, opened 2022-10-24
refactor AccountKeysCount to expect an error too

dreamsmasher:
Required for #1388.

## Description

There are a few places in the FVM where errors can occur while fetching the total number of account keys, but the method `AccountKeysCount` doesn't currently expect an error. This PR changes the type signature to accommodate this fact.

Currently a blocker for implementing `AccountKeysCount` in the FVM.
______

### DIFF
--- a/runtime/stdlib/account.go
+++ b/runtime/stdlib/account.go
@@ -787,7 +787,7 @@ func newAccountKeysCountGetter(
 	gauge common.MemoryGauge,
 	provider AccountKeyProvider,
 	addressValue interpreter.AddressValue,
-) interpreter.AccountKeysCountConstructor {
+) interpreter.AccountKeysCountGetter {
 	address := addressValue.ToAddress()
 
 	return func() interpreter.UInt64Value {
--- a/runtime/stdlib/account.go
+++ b/runtime/stdlib/account.go
@@ -292,7 +292,7 @@ func newAuthAccountKeysValue(
 			addressValue,
 		),
 		newAccountKeysForEachFunction(gauge, handler, addressValue),
-		newAccountKeysCountConstructor(gauge, handler, addressValue),
+		newAccountKeysCountGetter(gauge, handler, addressValue),
 	)
 }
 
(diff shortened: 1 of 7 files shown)


runtime/stdlib/account.go
```diff
@@ -728,12 +728,20 @@ func newAccountKeysForEachFunction(
...
+			})
+
+			if err != nil {
+				panic(err)
+			}
+
 			var accountKey *AccountKey
 
-			for index := 0; index < count; index++ {
+			for index := 0; index < int(count); index++ {
```
turbolent, line 744:
```suggestion
			for index := uint64(0); index < count; index++ {
```

runtime/stdlib/account.go
```diff
@@ -784,7 +792,12 @@ func newAccountKeysCountConstructor(
```
turbolent, line 792: Maybe also rename `newAccountKeysCountConstructor` to e.g. `newAccountKeysCountGetter`: it s not providing a constructor in the sense of the other constructors, e.g. like `NewAuthAccountConstructor`, but just a field getter function.

runtime/stdlib/account.go
```diff
@@ -784,7 +792,12 @@ func newAccountKeysCountConstructor(
 
 	return func() interpreter.UInt64Value {
 		return interpreter.NewUInt64Value(gauge, func() uint64 {
-			return provider.AccountKeysCount(address)
+			count, err := provider.AccountKeysCount(address)
```
turbolent, line 795: This external call also needs to be wrapped in a `wrapPanic`, like above

runtime/stdlib/account.go
```diff
@@ -784,7 +792,12 @@ func newAccountKeysCountConstructor(
...
+			if err != nil {
+				// if the FVM isn't able to fetch the number of account keys, we have a problem
```
turbolent, line 797: Maybe generalize "FVM" to "provider" – the embedder of Cadence might not be FVM, but e.g. some other blockchain implementation.

```suggestion
				// The provider might not be able to fetch the number of account keys, e.g when the account does not exist.
				// Panic with the user error
```

turbolent (commented):
Good idea to make the return values consistent 👍

runtime/stdlib/account.go
dreamsmasher: objects are just verbose lambdas 🤔

SupunS (approved):
LGTM once the suggestions are addressed

turbolent (approved):
Nice! 👍

PR #2099, opened 2022-10-26
Type Checking for attachment uses

dsainati1:
Part of https://github.com/onflow/cadence/issues/357 and https://github.com/onflow/cadence/issues/2061

This adds type checking for the uses of attachment types. In particular:
* attachments can be added to values using `attach A() to v`, which requires its first expression to be an attachment constructor for which `v` is a valid composite base type 
* attachments can be removed from values using `remove A from v`, which requires its first expression to be an attachment name for which `v` is a valid composite base type 
* attachments can be accessed on resources and structs using `v[A]`, which requires its indexing expression to be an attachment name for which `v` is a valid composite base type, and returns a value of type `&A?`
* introduces two new types: `AnyStructAttachment` and `AnyResourceAttachment`, which are the supertypes of all struct and resource attachments respectively. These are subtypes of `AnyStruct` and `AnyResource` as well. 
* Because attachments are not first-class values, they can only be used in annotations when inside of a reference type

______


runtime/sema/type_tags.go
```diff
@@ -624,6 +639,16 @@ func findSuperTypeFromUpperMask(joinedTypeTag TypeTag, types []Type) Type {
...
+	case anyResourceAttachmentMask:
+		if allTypesAreAttachments(types) {
+			return AnyResourceAttachmentType
+		}
+		return commonSuperTypeOfHeterogeneousTypes(types)
+	case anyStructAttachmentMask:
+		if allTypesAreAttachments(types) {
+			return AnyStructAttachmentType
+		}
+		return commonSuperTypeOfHeterogeneousTypes(types)
```
dsainati1, line 651: cc @SupunS i wasn't 100% sure if this is correct, can you please take a look and let me know?

runtime/sema/type.go
```diff
@@ -5521,6 +5766,12 @@ func checkSubTypeWithoutEquality(subType Type, superType Type) bool {
 			return typedSubType.ExplicitInterfaceConformanceSet().
 				Contains(typedSuperType)
 
+		// An interface type is a supertype of a restricted type if the restricted set contains
+		// that explicit interface type. Once interfaces can conform to interfaces, this should instead
+		// check that at least one value in the restriction set is a subtype of the interface supertype
+		case *RestrictedType:
+			return typedSubType.RestrictionSet().Contains(typedSuperType)
```
dsainati1, line 5773: This is necessary in order to check cases where an attachment is declared `attachment A for I {}` for some interface `I`, and then a value `v` of type `{ T1, T2, ... }` appears in the attach expression `attach A() to v`. `v` conforms to `I` as long as at least one of restrictions in its set is `I`. However, this will need to be updated when interfaces can conform to other interfaces to be a subtype check: i.e. `T1 <: I || T2 >: I || ..` cc @SupunS

runtime/sema/type_tags.go
SupunS: Does each attachment have its own type? or do they all have one type:  `AnyResourceAttachmentType`/ `anyStructAttachmentType`?
If they all have one type, then can simply return that from here. e.g:
```go
case anyResourceAttachmentMask:
	return AnyResourceAttachmentType
case anyStructAttachmentMask:
	return anyStructAttachmentType
```

(Explanation: here if the `joinedTypeTag` is `anyStructAttachmentMask`, that means all the types are struct attachments, so can return that as the supertype. If there are mixed types, then `joinedTypeTag` would be different hence will exit from default-case, and will eventually go to `commonSuperTypeOfHeterogeneousTypes()`)

dsainati1: In specifically the `AnyResource` case, because `anyResourceMask` is only a lower tag and `anyResourceAttachmentTag`, the switch over just the upper tag here will cause us to hit this case even if the two types that created the tag are `AnyResource` and `AnyResourceAttachment`. Hence we need to actually verify that all of the types in the list are actually attachments before we can say that the supertype is actually `AnyResourceAttachment`.

SupunS: ah yes, I noticed that too, it's actually a bug in the parent method and added a fix locally. Will send it today in a separate PR.

SupunS: Added in https://github.com/onflow/cadence/pull/2103. The two cases should be able to simplify as I suggested in the previous comment. let me know if still has problems.

runtime/ast/expression_extractor.go
```diff
@@ -781,3 +786,35 @@ func (extractor *ExpressionExtractor) VisitPathExpression(expression *PathExpres
...
+
+	if extractor.AttachExtractor != nil {
+		return extractor.AttachExtractor.ExtractAttach(extractor, expression)
+	}
+	return extractor.ExtractAttach(expression)
+}
+
+func (extractor *ExpressionExtractor) ExtractAttach(expression *AttachExpression) ExpressionExtraction {
+	// copy the expression
+	newExpression := *expression
```
robert-e-davidson3, line 803: I think it would be clearer if this happened after setting `rewrittenExpressions`. I'd also find it clearer if `newExpression` were created whole-cloth instead of copying it then rewriting 2/3 of its fields.

runtime/sema/check_assignment.go
```diff
@@ -136,6 +136,9 @@ func (checker *Checker) accessedSelfMember(expression ast.Expression) *Member {
 		members = containerType.Members
 	case *TransactionType:
 		members = containerType.Members
+	case *ReferenceType:
+		// self can only be a reference type if the container is an attachment, which is a composite
```
robert-e-davidson3, line 140: Is this guaranteed to always be true? Or should there be a switch to potentially return an UnreachableError?

runtime/sema/type.go
robert-e-davidson3: Maybe include this paragraph as a comment in the code? Or make it into an issue and include the URL to that issue in the code comment?

robert-e-davidson3 (approved):

runtime/sema/check_assignment.go
dsainati1: Yes, it is always the case.

runtime/ast/expression_extractor.go
dsainati1: This is the style the rest of the functions in this file use, so I just copied from that. cc @turbolent is there a reason we do things this way?

runtime/sema/check_attach_expression.go
```diff
@@ -0,0 +1,84 @@
...
+func (checker *Checker) VisitAttachExpression(expression *ast.AttachExpression) Type {
+	attachment := expression.Attachment
+
+	ty := checker.checkInvocationExpression(attachment)
+	if ty.IsInvalidType() {
+		return ty
+	}
+
+	attachmentType, baseIsCompositeType := ty.(*CompositeType)
+	if !baseIsCompositeType || attachmentType.Kind != common.CompositeKindAttachment {
```
dreamsmasher, line 35:
```suggestion
	if !(baseIsCompositeType && attachmentType.Kind == common.CompositeKindAttachment) {
```

de morgan

runtime/sema/check_attach_expression.go
```diff
@@ -0,0 +1,84 @@
...
+		return InvalidType
+	}
+
+	// if the annotatedBaseType is a specific interface or composite, then the above code will already have
+	// checked that the type of the base expression is also a composite. However, if the annotatedBaseType is
+	// anyresource/anystruct, we need to enforce that actualBaseType is a resource or struct
+	if _, annotatedIsCompositeType := annotatedBaseType.(CompositeKindedType); !annotatedIsCompositeType {
+		switch baseType := actualBaseType.(type) {
+		case CompositeKindedType:
+			if baseType.GetCompositeKind() != common.CompositeKindResource && baseType.GetCompositeKind() != common.CompositeKindStructure {
```
dreamsmasher, line 72:
```suggestion
			compositeKind := baseType.GetCompositeKind()
			if !(compositeKind == common.CompositeKindResource || compositeKind == common.CompositeKindStructure) {
```

CSE, de morgan

runtime/sema/check_expression.go
```diff
@@ -256,45 +256,98 @@ func (checker *Checker) visitIndexExpression(
...
-				Range: ast.NewRangeFromPositioned(checker.memoryGauge, targetExpression),
-			},
+	// Check if the type instance is actually indexable. For most types (e.g. arrays and dictionaries)
+	// this is known statically (in the sense of this host language (Go), not the implemented language),
+	// i.e. a Go type switch would be sufficient.
+	// However, for some types (e.g. reference types) this depends on what type is referenced
+	// we cannot use a switch here because in the reference type case we need to be able to "fallthrough",
+	// and Go does not allow fallthroughs in typed switches
+	valueIndexedType, isValueIndexableType := targetType.(ValueIndexableType)
+	typeIndexedType, isTypeIndexableType := targetType.(TypeIndexableType)
```
dreamsmasher, line 275: you should be able to move this cast to line 301, saving the overhead if `targetType` is value indexable

runtime/sema/check_remove_statement.go
```diff
@@ -0,0 +1,81 @@
...
+			},
+		)
+		return
+	}
+
+	// the actual base type must be a composite that can receive an attachment,
+	// and it must also be a valid subtype of the declared base type
+	switch baseType := base.(type) {
+	case *CompositeType:
+		if (baseType.Kind != common.CompositeKindResource && baseType.Kind != common.CompositeKindStructure) ||
```
dreamsmasher, line 50: actually this check on `CompositeKind` to verify that it's a valid attachment base seems to pop up in a few places, might be worth abstracting out for clarity

dreamsmasher: also de morgan's law again

runtime/sema/type.go
```diff
@@ -3779,6 +3827,137 @@ func (t *CompositeType) GetNestedTypes() *StringTypeOrderedMap {
 	return t.NestedTypes
 }
 
+func (t *CompositeType) isTypeIndexableType() bool {
+	// resources and structs only can be indexed for attachments
+	return t.Kind == common.CompositeKindResource || t.Kind == common.CompositeKindStructure
+}
```
dreamsmasher, line 3833: ah, this could be a method on `CompositeKind` actually, then you could use it in Checker.VisitAttachExpression and Checker.VisitRemoveStatement

runtime/sema/type.go
```diff
@@ -3779,6 +3827,137 @@ func (t *CompositeType) GetNestedTypes() *StringTypeOrderedMap {
...
+		IsSubType(t, attachmentType.baseType) &&
+		attachmentType.IsResourceType() == t.IsResourceType()
+}
+
+const compositeForEachAttachmentFunctionName = "forEachAttachment"
+
+const compositeForEachAttachmentFunctionDocString = `
+Iterates over the attachments present on the receiver, applying the function argument to each.
+The order of iteration is undefined. If a type argument is provided, only attachments that conform
+to the specified interface are iterated on, the others are filtered out.
```
dreamsmasher, line 3855: should we have the function return a `Bool` like the other `forEach` functions? that way we have a more unified API in case we decide to generalize `forEach` to any collection type in the future

runtime/sema/type.go
```diff
@@ -4775,6 +4972,30 @@ func (t *ReferenceType) isValueIndexableType() bool {
 	return referencedType.isValueIndexableType()
 }
 
+func (t *ReferenceType) isTypeIndexableType() bool {
+	referencedType, ok := t.Type.(TypeIndexableType)
+	if !ok {
+		return false
+	}
+	return referencedType.isTypeIndexableType()
```
dreamsmasher, line 4980:
```suggestion
	return ok && referencedType.isTypeIndexableType()
```

runtime/sema/type.go
```diff
@@ -4775,6 +4972,30 @@ func (t *ReferenceType) isValueIndexableType() bool {
...
+	}
+	return referencedType.TypeIndexingElementType(indexingType)
+}
+
+func (t *ReferenceType) IsValidIndexingType(ty Type) bool {
+	referencedType, ok := t.Type.(TypeIndexableType)
+	if !ok {
+		return false
+	}
+	return referencedType.IsValidIndexingType(ty)
```
dreamsmasher, line 4996:
```suggestion
	return ok && referencedType.IsValidIndexingType(ty)
```

runtime/sema/type_tags.go
```diff
@@ -613,6 +619,15 @@ func findSuperTypeFromLowerMask(joinedTypeTag TypeTag, types []Type) Type {
 	}
 }
 
+func allTypesAreAttachments(types []Type) bool {
+	for _, ty := range types {
+		if !isAttachmentType(ty) {
+			return false
```
dreamsmasher, line 625: if only golang had iterators. or folds. or any abstraction facilities whatsoever :'(

dreamsmasher (commented):
LGTM so far, just some minor nitpicks and microoptimizations. it's interesting seeing how restricted types interact with variance in our current type system

runtime/sema/check_remove_statement.go
```diff
@@ -0,0 +1,81 @@
...
+		)
+		return
+	}
+
+	// the actual base type must be a composite that can receive an attachment,
+	// and it must also be a valid subtype of the declared base type
+	switch baseType := base.(type) {
+	case *CompositeType:
+		if !baseType.Kind.SupportsAttachments() ||
+			!IsSubType(baseType, attachmentType.baseType) {
```
dreamsmasher, line 51:
```suggestion
		if !(baseType.Kind.SupportsAttachments() &&
			IsSubType(baseType, attachmentType.baseType)) {
```

de morgan

runtime/sema/type.go
```diff
@@ -5502,6 +5506,19 @@ func checkSubTypeWithoutEquality(subType Type, superType Type) bool {
...
+		case *InterfaceType:
+			switch typedInnerSubType := typedSubType.Type.(type) {
+			case *CompositeType:
+				return typedInnerSubType.ExplicitInterfaceConformanceSet().Contains(typedInnerSuperType)
+			// An interface type is a supertype of a restricted type if the restricted set contains
+			// that explicit interface type. Once interfaces can conform to interfaces, this should instead
+			// check that at least one value in the restriction set is a subtype of the interface supertype
+			case *RestrictedType:
+				return typedInnerSubType.RestrictionSet().Contains(typedInnerSuperType)
+			}
```
dsainati1, line 5520: This is needed to make `&{I}` a subtype of `&I`, and thus to allow an attachment for `I` to be accessed on a reference restricted to `{I}`

runtime/sema/anyattachment_types.go
```diff
@@ -0,0 +1,51 @@
...
+	IsResource:    true,
+	// The actual storability of a value is checked at run-time
+	Storable:  true,
+	Equatable: false,
+	// The actual returnability of a value is checked at run-time
+	ExternallyReturnable: true,
+	Importable:           false,
+}
+
+// AnyStructAttachmentType represents the top type of all resource attachment types
```
dsainati1, line 37:
```suggestion
// AnyStructAttachmentType represents the top type of all struct attachment types
```

runtime/sema/type.go
dsainati1: Maybe promote the `I` to a restricted type instead.

runtime/sema/type.go
```diff
@@ -3779,6 +3827,137 @@ func (t *CompositeType) GetNestedTypes() *StringTypeOrderedMap {
...
+					Type: &GenericType{
+						TypeParameter: typeParameter,
+					},
+				},
+			},
+		),
+	}
+}
+
+const attachmentGetFunctionFunctionName = "getFunction"
```
dsainati1, line 3926: Might want this to be an `invokeFunction` instead, which would take the name and a variadic number of arguments depending on the type parameter, and return an optional return type. This would prevent issues with bound resource functions.

runtime/sema/type.go
dsainati1: After some thought, requiring manual conversion from an interface to a singleton restricted set every time the base type of an attachment is used is pretty fragile; forgetting to perform the conversion in some case could easily cause a missed error. I'm pretty confident this logic is correct and feel like adding support here is better than requiring all potential inputs to this function to be converted.

dsainati1:
The reflection and iteration API features need some more discussion, but we don't want to block development of the basic parts of this feature on those other parts. So pending discussion on https://github.com/onflow/flips/pull/50 I've removed the parts of typechecking that pertained to these features, and will add them back in future PRs once we come to a decision about how they should work.

runtime/sema/check_expression.go
```diff
@@ -256,45 +256,101 @@ func (checker *Checker) visitIndexExpression(
...
-		checker.report(
-			&NotIndexingAssignableTypeError{
-				Type:  indexedType,
-				Range: ast.NewRangeFromPositioned(checker.memoryGauge, targetExpression),
-			},
+	// Check if the type instance is actually indexable. For most types (e.g. arrays and dictionaries)
+	// this is known statically (in the sense of this host language (Go), not the implemented language),
+	// i.e. a Go type switch would be sufficient.
+	// However, for some types (e.g. reference types) this depends on what type is referenced
+	// we cannot use a switch here because in the reference type case we need to be able to "fallthrough",
```
SupunS, line 272: Why a fallthrough will be needed? Is it because some types can belong to both `ValueIndexableType` and `TypeIndexableType`?

runtime/sema/check_expression.go
```diff
@@ -256,45 +256,101 @@ func (checker *Checker) visitIndexExpression(
...
+	} else {
+		reportNonIndexable(targetType)
+		return InvalidType
+	}
+}
+
+func (checker *Checker) checkTypeIndexingExpression(
+	base TypeIndexableType,
+	reportError func(indexedType Type),
+	indexingExpression ast.Expression,
```
SupunS, line 333: maybe swap the order of these last two params, so the base and index are positioned together (just for better readability)

runtime/sema/check_remove_statement.go
```diff
@@ -0,0 +1,81 @@
...
+
+	if nominalType == InvalidType {
+		return
+	}
+
+	attachmentType, isCompositeType := nominalType.(*CompositeType)
+
+	if !isCompositeType || attachmentType.Kind != common.CompositeKindAttachment {
+		checker.report(
+			&RemoveFromInvalidTypeError{
```
SupunS, line 38: Should this be a different error compared to the below ones? i.e: Here it is the attachment type that is wrong. More like "removing a non-attachment type error"

runtime/sema/check_expression.go
dsainati1: Yes; `ReferenceType` is statically both a `ValueIndexableType` and a `TypeIndexableType`, but in practice the return of `isValueIndexableType` and `isTypeIndexableType` depend on the specific borrow type.

runtime/sema/check_remove_statement.go
```diff
@@ -0,0 +1,81 @@
...
+package sema
+
+import (
+	"github.com/onflow/cadence/runtime/ast"
+	"github.com/onflow/cadence/runtime/common"
+)
+
+func (checker *Checker) VisitRemoveStatement(statement *ast.RemoveStatement) (_ struct{}) {
+	nominalType := checker.convertNominalType(statement.Attachment)
+	base := checker.VisitExpression(statement.Value, nil)
```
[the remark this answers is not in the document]
dsainati1: Maybe? `Value` is consistent with how other statements (`ForStatement` and `AssignmentStatement`) are named, but we could also change those to be more descriptive as well potentially).

runtime/sema/check_remove_statement.go
dsainati1: This is what the error message looks like in practice, but I will rename the error type to be more representative of the fact that it handles both cases.

runtime/sema/check_expression.go
SupunS: makes sense. Thanks for the explanation!

SupunS (approved):
LGTM!

runtime/sema/check_expression.go
```diff
@@ -47,6 +48,29 @@ func (checker *Checker) VisitIdentifierExpression(expression *ast.IdentifierExpr
...
 
+// checkSelfVariableUseInInitializer checks uses of `base` in the initializer
+// and ensures it is not used outside of a member expression
+func (checker *Checker) checkBaseVariableUseInAttachment(variable *Variable, expression *ast.IdentifierExpression) {
+	// Is this a use of `base`?
+
+	if variable.Identifier == "base" {
+		y := 3
+		_ = y + 1
+	}
```
dsainati1, line 59:
```suggestion
```

runtime/sema/type.go
```diff
@@ -3543,6 +3556,13 @@ func (t *CompositeType) getBaseCompositeKind() common.CompositeKind {
 	return common.CompositeKindUnknown
 }
 
+func isAttachmentType(t Type) bool {
```
turbolent, line 3559: Maybe double check this function is used everywhere necessary, not just the first case

runtime/sema/type.go
```diff
@@ -4793,14 +4857,30 @@ func (t *ReferenceType) IndexingType() Type {
 	return referencedType.IndexingType()
 }
 
-func (*ReferenceType) Unify(_ Type, _ *TypeParameterTypeOrderedMap, _ func(err error), _ ast.Range) bool {
-	// TODO:
-	return false
+func (t *ReferenceType) Unify(
```
turbolent, line 4860: Nice work implementing Unify/Resolve. Is this needed for the attachments feature?

turbolent (commented):
Great work! 

Reviewed everything except `sema/type.go`, need to review the new subtyping rules in detail.

runtime/sema/check_attach_expression.go
```diff
@@ -0,0 +1,97 @@
...
+		checker.report(
+			&AttachNonAttachmentError{
+				Type:  ty,
+				Range: ast.NewRangeFromPositioned(checker.memoryGauge, attachment),
+			},
+		)
+		return InvalidType
+	}
+
+	annotatedBaseType := attachmentType.baseType
```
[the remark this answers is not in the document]
dsainati1: The name here is meant to communicate that this is the base type that was annotated in the declaration of the attachment type: i.e. `I` in `attachment A for I {}`, specifically to contrast with the specific `baseType` as obtained from typechecking the right hand side of `attach`.

runtime/sema/type.go
dsainati1: Doesn't actually seem like it; I thought it would be because of how I was planning on using references, but in actuality it doesn't seem necessary. If you think it doesn't belong in this PR I can revert the change here.

runtime/sema/check_attach_expression.go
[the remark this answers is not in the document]
turbolent: Right, that was also my understanding. I just didn't quite know what "annotated" / "annotation" was referring to. Maybe "declaration" is better?

runtime/sema/type.go
turbolent: No, it's great it got implemented, thank you! Please make sure we have tests for it as well, or maybe submit it as a separate PR, so this PR doesn't keep growing.

turbolent:
@dsainati1 Points 5 and 6 in the description are not part of the PR anymore, right?

dsainati1:
Ah yea good point

runtime/sema/type.go
```diff
@@ -5261,6 +5347,19 @@ func checkSubTypeWithoutEquality(subType Type, superType Type) bool {
 			//
 			// The holder of the reference may not gain more permissions or knowledge.
 
+			return false
+
+		case *InterfaceType:
+			switch typedInnerSubType := typedSubType.Type.(type) {
+			case *CompositeType:
```
turbolent, line 5354: Please also add a comment like below. 

This seems to implement `&T <: &U`, where `T` is a concrete type and `U` is an interface type.
This is similar to the rule `&T <: &{U}`. 

Maybe point that out and include a comment similar to the one for that rule.

runtime/sema/type.go
```diff
@@ -5284,18 +5383,20 @@ func checkSubTypeWithoutEquality(subType Type, superType Type) bool {
...
 			case *CompositeType:
-				return typedInnerSubType.Kind == common.CompositeKindResource
+				return typedInnerSubType.IsResourceType()
 			}
 
 		case AnyStructType:
 			// `&T <: &AnyStruct` iff `T <: AnyStruct`
 			return IsSubType(typedSubType.Type, typedSuperType.Type)
+		case AnyResourceAttachmentType, AnyStructAttachmentType:
+			return IsSubType(typedSubType.Type, typedSuperType.Type)
```
turbolent, line 5399: Please add comments explaining these rules

runtime/sema/type.go
```diff
@@ -5519,6 +5620,19 @@ func checkSubTypeWithoutEquality(subType Type, superType Type) bool {
...
+		// check that at least one value in the restriction set is a subtype of the interface supertype
+
+		// This particular case comes up when checking attachment access; enabling the following expression to typechecking:
+		// resource interface I { /* ... */ }
+		// attachment A for I { /* ... */ }
+		//
+		// let i : &{I} = ... // some operation constructing `i`
+		// let a = i[A] // must here check that `i`'s type is a subtype of `A`'s base type, or that {I} <: I
+		case *RestrictedType:
+			return typedSubType.RestrictionSet().Contains(typedSuperType)
```
turbolent, line 5634: Same as above: Maybe this can be a separate function that is only used for this purpose.

The `IsSubType` function is already very complex, and so far assumes / could rely on the fact that interfaces only appear in restricted types. I'm worried the rules / this function becomes incomplete.

runtime/sema/check_attach_expression.go
```diff
@@ -0,0 +1,97 @@
...
+			},
+		)
+		return InvalidType
+	}
+
+	// if the annotatedBaseType is a specific interface or composite, then the above code will already have
+	// checked that the type of the base expression is also a composite. However, if the annotatedBaseType is
+	// anyresource/anystruct, we need to enforce that baseType is a resource or struct, to prevent
+	// permitting code like `attach A to 4`, if `A` was declared for `AnyStruct`
+	if _, annotatedIsCompositeType := declaredBaseType.(CompositeKindedType); !annotatedIsCompositeType {
```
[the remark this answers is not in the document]
dsainati1: No; references are type indexable, but it shouldn't be possible to attach to them, even if the attachment was declared for `AnyStruct`.

runtime/sema/type.go
dsainati1: > The implementation ignores the restricted type (T in &T{Us}), does it need to be checked too?

We are deliberately excluding this check: 

```
struct R: I, J {}
struct interface I {}
struct interface J {}
attachment A for I {}
pub fun foo(r: R{J}) {
	r[A]
}
```

We would like the above code to fail to typecheck; although `R` is a subtype of `I`, when it is restricted to `J` we don't want people to be able to access or add attachments to it that are only intended to work with `I` values.

runtime/sema/type.go
dsainati1: https://github.com/onflow/cadence/pull/2099#discussion_r1042362870

runtime/sema/check_attach_expression.go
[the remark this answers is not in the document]
turbolent: That makes sense, thanks for the explanation! Could you please add this as a comment?

runtime/sema/type.go
```diff
@@ -5591,6 +5709,19 @@ func checkSubTypeWithoutEquality(subType Type, superType Type) bool {
...
 
+		// An interface type is a supertype of a restricted type if the restricted set contains
+		// that explicit interface type. Once interfaces can conform to interfaces, this should instead
+		// check that at least one value in the restriction set is a subtype of the interface supertype
+
+		// This particular case comes up when checking attachment access; enabling the following expression to typechecking:
+		// resource interface I { /* ... */ }
+		// attachment A for I { /* ... */ }
+		//
+		// let i : &{I} = ... // some operation constructing `i`
```
turbolent, line 5720: I think the the rule is actually for `{I}`, not `&{I}`:

```suggestion
		// let i : {I} = ... // some operation constructing `i`
```

runtime/sema/type.go
```diff
@@ -5591,6 +5709,19 @@ func checkSubTypeWithoutEquality(subType Type, superType Type) bool {
 			return typedSubType.ExplicitInterfaceConformanceSet().
 				Contains(typedSuperType)
 
+		// An interface type is a supertype of a restricted type if the restricted set contains
+		// that explicit interface type. Once interfaces can conform to interfaces, this should instead
+		// check that at least one value in the restriction set is a subtype of the interface supertype
+
```
turbolent, line 5715:
```suggestion
		//
```

runtime/sema/type.go
```diff
@@ -5333,6 +5419,34 @@ func checkSubTypeWithoutEquality(subType Type, superType Type) bool {
...
+				return typedInnerSubType.ExplicitInterfaceConformanceSet().Contains(typedInnerSuperType)
+			// An unauthorized reference to an interface type `&T`
+			// is a supertype of a reference to a restricted type `&{U}`:
+			// if the restriction set contains that explicit interface type.
+			//
+			// Note that this does not check whether the restricted type's restricted type conforms to the interface;
+			// i.e. whether in `&R{X} <: &I`, `R <: I`. This is intentional;
+			// when checking whether an attachment declared for `I` is accessible on a value of type `&R{X}`,
+			// even if `R <: I`, we only want to allow access if `X <: I`
+			//
```
turbolent, line 5442: Maybe add the great example for the non-reference case you have below here too:


```suggestion
			//
			// This particular case comes up when checking attachment access;
			// resource interface I { /* ... */ }
			// attachment A for I { /* ... */ }
			//
			// let i : &{I} = ... // some operation constructing `i`
			// let a = i[A] // must here check that `i`'s type is a subtype of `A`'s base type, or that &{I} <: &I
```

runtime/sema/type.go
```diff
@@ -5261,6 +5347,19 @@ func checkSubTypeWithoutEquality(subType Type, superType Type) bool {
 			//
 			// The holder of the reference may not gain more permissions or knowledge.
 
+			return false
+
+		case *InterfaceType:
```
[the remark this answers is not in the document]
turbolent: Thank you for explaining the options and their tradeoffs 👏

turbolent (approved):
Great work! 👏 

Just some more suggestions for improving the comments even further, but finally ready to go! 🎉 
Sorry this too so long, but this PR touches a lot of pieces

PR #2100, opened 2022-10-26
Allow embedders to reuse interpreter shared state

### LINKED ISSUE #2048: Reuse interpreted programs within a transaction
### Issue To Be Solved

This relates to [this](https://dapperlabs.slack.com/archives/CG0B7CJAJ/p1664459408641099) slack discussion.

Exploring the profile of execution an empty transaction (transaction that does basically nothing) I noticed that EsureLoaded is called multiple times during one transaction and is using a significant amount of time. Most of the time within EsureLoaded is spent inside Interpreter.Interpret building interpretations.

image

Ensure loaded is used in the three different parts of the trasanction:
- invoke users transaction
- deduct fees
- check accounts storage limits

In the scenario profiled above, all programs should be in the programs cache, very little time is needed to retrieve a program from the programs cache.

**Deduct fees** and **check accounts storage limits share** a lot of contracts (FlowToken, FungibleToken, FlowServiceAccount) and since those are commonly used **invoke users transaction** might have needed them as well. So It would be beneficial to reuse the interpretation between these three parts of the trasnaction.

The reuse would happen only within a transaction, and would be dropped if a revert happened (failed fee deduction or storage check or the invocation itself).

After some more digging I managed to create a PoC:
- https://github.com/onflow/cadence/compare/v0.27.0...janez/poc-interpreter-shared-state-reuse
- https://github.com/onflow/flow-go/compare/tonyz/fvm_setting_prefetch...janez/poc-interpreter-shared-state-reuse

... and it looks like this might lead to a 15% speedup in execution times.

### Suggested Solution

There are two different use cases for [NewInterpreter](https://github.com/onflow/cadence/blob/01000ccddec4448e9cd1e7476671e1c509edd46a/runtime/interpreter/interpreter.go#L304)
One  is when program is not nil (transactions, scripts) and one where it is (contract function invocation, get stored value). I think those two cases should be treated separately on the [Environment](https://github.com/onflow/cadence/blob/a3b018786063df6bcc2308b833fa56d5155c754e/runtime/environment.go#L37) interface.
With the case where the program is nil, you should be able to specify an existing interpreter (or better yet an `interpreter.sharedState`) when you have one. The FVM could grab the interpreter (or shared state) created during transaction invocation, and use it for calls to  contract function invocations and get stored value.

The problem is that the interpreter (and also shared state) is not stateless, and I am not sure if calling `CommitStorage` on it multiple times is safe.

turbolent:
Closes #2048 

## Description

Based on @janezpodhostnik's idea and [proof-of-concept](https://github.com/onflow/cadence/commit/4ac0d1bb126a14b889c319a36e990489bf2e3628):
Allow the embedder, e.g. FVM, to reuse the shared state across multiple runtime top-level entry calls (e.g. transaction invocation, contract function invocation).

If the embedder, through the new interface functions `SetInterpreterSharedState` and `GetInterpreterSharedState`, saves the state and re-supplies it in a subsequent top-level entry call, interpreters for imported contracts are re-used – contract data does not need to be re-read from storage, and contract values do not need to be recreated by the interpreter.

When enabling this optimization for `BenchmarkRuntimeFungibleTokenTransfer`, this optimization improves performance and allocations by ca ~16%:

```
6878	    172776 ns/op	  106226 B/op	    2002 allocs/op
8194	    145198 ns/op	   81168 B/op	    1645 allocs/op
```
______

### DIFF
--- a/runtime/runtime_test.go
+++ b/runtime/runtime_test.go
@@ -140,7 +140,7 @@ type testRuntimeInterface struct {
 	getProgram                func(Location) (*interpreter.Program, error)
 	setProgram                func(Location, *interpreter.Program) error
 	setInterpreterSharedState func(state *interpreter.SharedState)
-	getInterpreterShareState  func() *interpreter.SharedState
+	getInterpreterSharedState func() *interpreter.SharedState
 	storage                   testLedger
 	createAccount             func(payer Address) (address Address, err error)
 	addEncodedAccountKey      func(address Address, publicKey []byte) error
@@ -252,11 +252,11 @@ func (i *testRuntimeInterface) SetInterpreterSharedState(state *interpreter.Shar
 }
 
 func (i *testRuntimeInterface) GetInterpreterSharedState() *interpreter.SharedState {
-	if i.getInterpreterShareState == nil {
+	if i.getInterpreterSharedState == nil {
 		return nil
 	}
 
-	return i.getInterpreterShareState()
+	return i.getInterpreterSharedState()
 }
 
 func (i *testRuntimeInterface) ValueExists(owner, key []byte) (exists bool, err error) {
--- a/runtime/runtime_test.go
+++ b/runtime/runtime_test.go
@@ -135,15 +135,17 @@ func newTestInterpreterRuntime() Runtime {
 }
 
 type testRuntimeInterface struct {
-	resolveLocation         func(identifiers []Identifier, location Location) ([]ResolvedLocation, error)
-	getCode                 func(_ Location) ([]byte, error)
-	getProgram              func(Location) (*interpreter.Program, error)
-	setProgram              func(Location, *interpreter.Program) error
-	storage                 testLedger
-	createAccount           func(payer Address) (address Address, err error)
-	addEncodedAccountKey    func(address Address, publicKey []byte) error
-	removeEncodedAccountKey func(address Address, index int) (publicKey []byte, err error)
-	addAccountKey           func(
+	resolveLocation           func(identifiers []Identifier, location Location) ([]ResolvedLocation, error)
+	getCode                   func(_ Location) ([]byte, error)
+	getProgram                func(Location) (*interpreter.Program, error)
+	setProgram                func(Location, *interpreter.Program) error
+	setInterpreterSharedState func(state *interpreter.SharedState)
+	getInterpreterShareState  func() *interpreter.SharedState
+	storage                   testLedger
+	createAccount             func(payer Address) (address Address, err error)
+	addEncodedAccountKey      func(address Address, publicKey []byte) error
+	removeEncodedAccountKey   func(address Address, index int) (publicKey []byte, err error)
+	addAccountKey             func(
 		address Address,
 		publicKey *stdlib.PublicKey,
 		hashAlgo HashAlgorithm,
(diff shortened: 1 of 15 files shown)


bluesign:
Very nice

Does this support rollback on state? Can I use this with debugger to rollback state for example? Like time machine. 

Is this state is serializable?

runtime/runtime_test.go
```diff
@@ -135,15 +135,17 @@ func newTestInterpreterRuntime() Runtime {
...
-	createAccount           func(payer Address) (address Address, err error)
-	addEncodedAccountKey    func(address Address, publicKey []byte) error
-	removeEncodedAccountKey func(address Address, index int) (publicKey []byte, err error)
-	addAccountKey           func(
+	resolveLocation           func(identifiers []Identifier, location Location) ([]ResolvedLocation, error)
+	getCode                   func(_ Location) ([]byte, error)
+	getProgram                func(Location) (*interpreter.Program, error)
+	setProgram                func(Location, *interpreter.Program) error
+	setInterpreterSharedState func(state *interpreter.SharedState)
+	getInterpreterShareState  func() *interpreter.SharedState
```
dreamsmasher, line 143:
```suggestion
	getInterpreterSharedState  func() *interpreter.SharedState
```

runtime/runtime_test.go
```diff
@@ -241,6 +243,22 @@ func (i *testRuntimeInterface) SetProgram(location Location, program *interprete
...
+func (i *testRuntimeInterface) SetInterpreterSharedState(state *interpreter.SharedState) {
+	if i.setInterpreterSharedState == nil {
+		return
+	}
+
+	i.setInterpreterSharedState(state)
+}
+
+func (i *testRuntimeInterface) GetInterpreterSharedState() *interpreter.SharedState {
+	if i.getInterpreterShareState == nil {
```
dreamsmasher, line 255:
```suggestion
	if i.getInterpreterSharedState == nil {
```

runtime/runtime_test.go
```diff
@@ -241,6 +243,22 @@ func (i *testRuntimeInterface) SetProgram(location Location, program *interprete
...
+		return nil
+	}
```
dreamsmasher, line 257: maybe panic here like the other testRuntimeInterface methods, since returning nil doesn't signify whether the operation failed or if the interface just doesn't have a fn pointer

runtime/sharedstate_test.go
```diff
@@ -0,0 +1,259 @@
...
+			loggedMessages = append(loggedMessages, message)
+		},
+		emitEvent: func(event cadence.Event) error {
+			events = append(events, event)
+			return nil
+		},
+		setInterpreterSharedState: func(state *interpreter.SharedState) {
+			interpreterState = state
+		},
+		getInterpreterShareState: func() *interpreter.SharedState {
```
dreamsmasher, line 138:
```suggestion
		getInterpreterSharedState: func() *interpreter.SharedState {
```

dreamsmasher (approved):

dsainati1 (approved):

runtime/runtime_test.go
turbolent: Fixed in 3d44ae4a28b25764b4d9b146e25db6f8d280f6a4

runtime/runtime_test.go
turbolent: Fixed in 3d44ae4a28b25764b4d9b146e25db6f8d280f6a4

runtime/sharedstate_test.go
turbolent: Fixed in 3d44ae4a28b25764b4d9b146e25db6f8d280f6a4

runtime/runtime_test.go
turbolent: `GetInterpreterSharedState` is an optional function, the embedder does not need to implement it. The runtime still works if this function is not implemented

SupunS (approved):

turbolent:
@bluesign 

> Does this support rollback on state? Can I use this with debugger to rollback state for example? Like time machine.
> Is this state is serializable?

The word "state" here refers to the in-memory data structures that the interpreter builds up for executing programs. For example, there is one `Interpreter` instance per location (e.g. contract), and one interpreter value representing a contract object (a `CompositeValue`). The state also e.g. contains the call stack.

This interpreter state is unrelated to the "chain execution state", which is called "storage" in Cadence.

The interpreter shared state is not currently serializable.

PR #2103, opened 2022-10-27
Improve common supertype calculation

SupunS:
Required for https://github.com/onflow/cadence/pull/2099

## Description

Improve the supertype calculation to make it easy to add new types in the 'upper mask' section.

This does not change the behaviour for the existing types. Therefore not a breaking change.

______

### DIFF
--- a/runtime/sema/type_tags.go
+++ b/runtime/sema/type_tags.go
@@ -369,6 +369,25 @@ var (
 // Methods
 
 func LeastCommonSuperType(types ...Type) Type {
+
+	superType := leastCommonSuperType(types...)
+	if superType == InvalidType {
+		return superType
+	}
+
+	// Do a sanity check to see if all types are
+	// actually subtypes of the inferred supertype.
+	for _, typ := range types {
+		if !IsSubType(typ, superType) {
+			return InvalidType
+		}
+
+	}
+
+	return superType
+}
+
+func leastCommonSuperType(types ...Type) Type {
 	join := NoTypeTag
 
 	for _, typ := range types {
@@ -404,10 +423,16 @@ func findCommonSuperType(joinedTypeTag TypeTag, types ...Type) Type {
 	// finding super type for the rest of the types is sufficient.
 	joinedTypeTag = joinedTypeTag.And(notNeverType)
 
-	if joinedTypeTag.upperMask != 0 {
-		superType = findSuperTypeFromUpperMask(joinedTypeTag, types)
-	} else {
+	if joinedTypeTag == NoTypeTag {
+		return InvalidType
+	}
+
+	// If both masks are on, then the types are heterogeneous.
+	// So skip the optimization and find the supertype in the hard way.
+	if joinedTypeTag.lowerMask != 0 && joinedTypeTag.upperMask == 0 {
 		superType = findSuperTypeFromLowerMask(joinedTypeTag, types)
+	} else if joinedTypeTag.lowerMask == 0 && joinedTypeTag.upperMask != 0 {
+		superType = findSuperTypeFromUpperMask(joinedTypeTag, types)
 	}
 
 	if superType != nil {
@@ -423,7 +448,7 @@ func findCommonSuperType(joinedTypeTag TypeTag, types ...Type) Type {
 		// Get the types without the optionals
 		unwrappedTypes, levels := unwrapOptionals(types)
 
-		superType := findCommonSuperType(joinedTypeTag, unwrappedTypes...)
+		superType = findCommonSuperType(joinedTypeTag, unwrappedTypes...)
 
 		// If the common supertype of the rest of types contain nil (e.g: AnyStruct),
 		// then do not wrap with optional again.
@@ -680,7 +705,7 @@ func commonSuperTypeOfVariableSizedArrays(types []Type) Type {
 		elementTypes = append(elementTypes, arrayType.ElementType(false))
 	}
 
-	elementSuperType := LeastCommonSuperType(elementTypes...)
+	elementSuperType := leastCommonSuperType(elementTypes...)
 
 	if elementSuperType == InvalidType {
 		return InvalidType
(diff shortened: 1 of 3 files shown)


runtime/sema/type_tags.go
```diff
@@ -404,10 +423,16 @@ func findCommonSuperType(joinedTypeTag TypeTag, types ...Type) Type {
...
+	if joinedTypeTag == NoTypeTag {
+		return InvalidType
+	}
+
+	// If both masks are on, then the types are heterogeneous.
+	// So skip the optimization and find the supertype in the hard way.
+	if joinedTypeTag.lowerMask != 0 && joinedTypeTag.upperMask == 0 {
 		superType = findSuperTypeFromLowerMask(joinedTypeTag, types)
+	} else if joinedTypeTag.lowerMask == 0 && joinedTypeTag.upperMask != 0 {
+		superType = findSuperTypeFromUpperMask(joinedTypeTag, types)
```
SupunS, line 435: The actual change/fix is here.
Earlier it was assumed that checking the existence of 'upper-mask' is sufficient, because the upper mask only had three types, and for those three types `'findSuperTypeFromUpperMask' also handled heterogeneous types.
However, when adding more types to upper-masks, this duplicates the work as they all needed to handle heterogeneous types for newly added types as well.

Instead, if both upper and lower masks are present, fall back to the default case here, so `findSuperTypeFromUpperMask` don't have to worry about heterogeneous types. Heterogeneous types are handled outside of this 'if' condition already.

runtime/sema/type_tags.go
```diff
@@ -369,6 +369,25 @@ var (
...
+	if superType == InvalidType {
+		return superType
+	}
+
+	// Do a sanity check to see if all types are
+	// actually subtypes of the inferred supertype.
+	for _, typ := range types {
+		if !IsSubType(typ, superType) {
+			return InvalidType
+		}
```
dreamsmasher, line 383: you could add this sanity check at the end of `leastCommonSuperType`, and export that directly. the recursion could be accomplished with an internal closure or a fixedpoint combinator

dreamsmasher: wait no, you can't have recursive nested functions can you... maybe rename `leastCommonSuperType` to make it more apparent that this is just a wrapper around a recursive worker fn

dreamsmasher (approved):

runtime/sema/type_tags.go
```diff
@@ -336,39 +336,58 @@ var (
...
```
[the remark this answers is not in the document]
dsainati1: Do we really want this sanity checking running on actual contracts? Or is it possible only to enable this in the language server or linter? Perhaps when `extendedElaboration` is enabled?

dsainati1 (approved):

runtime/sema/type_tags.go
SupunS: yeah, `LeastCommonSuperType` is the public function (exported) to be used everywhere, and is just a wrapper around the `leastCommonSuperType` which is a local function. We have used this naming pattern elsewhere too to distinguish between a local function vs a global/exported function.

runtime/sema/type_tags.go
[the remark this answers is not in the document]
SupunS: We need this by default to correct any bugs in the implementation/logic of the super-type inference, if ever there were/will be any. 
e.g: adding a new type can unlock new paths in the logic (like the one we just had), which may have not been thought of before, and as a result, if a wrong type is inferred, then that should be prevented. Otherwise, it could cause type issues at runtime.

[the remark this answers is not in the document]
SupunS: It sort of ensures that the supertype inference is compliant with (i.e: inverse of) `IsSubType` check

turbolent (approved):
Nice! Good idea to add a defensive check 👍

turbolent:
Maybe add a benchmark which exercises `LeastCommonSuperType` and compare to before, then add a comment that gives a rough estimate for the overhead of the defensive check.

SupunS:
Okayyy this sounds bad 
(well, this is a micro-benchmark and the times are in `ns`, so this wouldn't really be visible in the overall performance. But still..)
```
name                             old time/op    new time/op    delta
SuperTypeInference/integers-8      16.1ns ± 1%    65.7ns ± 1%  +308.53%  (p=0.016 n=4+5)
SuperTypeInference/arrays-8         164ns ±15%     225ns ± 2%   +36.98%  (p=0.008 n=5+5)
SuperTypeInference/composites-8    77.5ns ±12%    90.1ns ± 0%   +16.30%  (p=0.016 n=5+4)

name                             old alloc/op   new alloc/op   delta
SuperTypeInference/integers-8       0.00B          0.00B           ~     (all equal)
SuperTypeInference/arrays-8         96.0B ± 0%     96.0B ± 0%      ~     (all equal)
SuperTypeInference/composites-8     0.00B          0.00B           ~     (all equal)

name                             old allocs/op  new allocs/op  delta
SuperTypeInference/integers-8        0.00           0.00           ~     (all equal)
SuperTypeInference/arrays-8          3.00 ± 0%      3.00 ± 0%      ~     (all equal)
SuperTypeInference/composites-8      0.00           0.00           ~     (all equal)
```

Having the defensive check means if the type is inferred, the `IsSubType` check is run for the 'inferred' vs 'actual' types, just as if the user has declared the type (in that case 'expected' vs 'actual'). So that overhead would be there even if the user declared the type.

What's interesting is, inferring the type is faster than performing the subtype types check, for some cases, like integers 🤔
(test is here: [9994d8b](https://github.com/onflow/cadence/pull/2103/commits/9994d8ba1c02066d0a250b5a856e2e868f8e45c2))
```
SuperTypeInference/integers-8      16.1ns ± 1%    65.7ns ± 1%  +308.53% 
```
The overhead os subtype check is larger than the actual inference.

```cadence
let x = [1, 2, 3]      // type is inferred, so its fast
```

```cadence
let x: Int = [1, 2, 3]  // IsSubType runs, so its slow
```

SupunS:
Offline meeting conclusion: The check happens at checking time, and also the overhead is in nano-seconds. So this doesn't really impact the type-checking performance visibly. At the same time, assuring this safety using a test case is also impossible, since there can be infinite number of combinations of types. So the defensive check is worth having.

PR #2108, opened 2022-10-28
Fix self declaration

turbolent:
## Description

Identified by @dsainati1: Invocations declared `self` even when not necessary.

In f63134aef91af5dfab0c84645f8f40d64230c4aa I had changed the type of `Self` from a concrete `*CompositeValue` to `MemberAccessibleValue`, but kept storing `nil` in it, leading to the "typed nil" Go footgun.

______


SupunS (approved):

runtime/contract_function_executor.go
```diff
@@ -183,10 +183,12 @@ func (executor *interpreterContractFunctionExecutor) execute() (val cadence.Valu
 		return nil, newError(err, location, codesAndPrograms)
 	}
 
+	var self interpreter.MemberAccessibleValue = contractValue
+
 	// prepare invocation
 	invocation := interpreter.NewInvocation(
 		inter,
-		contractValue,
+		&self,
```
dsainati1, line 191: So am I correct then in understanding that this would now be a double-pointer `**CompositeValue` type in the non-`nil` case for `contractValue`? Getting flashbacks to C++ here

dsainati1 (approved):
Thank you for fixing this!

runtime/contract_function_executor.go
turbolent: Correct

PR #2110, opened 2022-10-29
Meter big int allocations in parser, improve over-estimation

turbolent:
Closes dapperlabs/cadence-private-issues#41

## Description

- Meter big int allocations in parser
- Improve over-estimate of big int allocation for a string
- Move estimation tests, fix mixed-case filename

______


runtime/common/bigint.go
```diff
@@ -54,5 +56,9 @@ func OverEstimateBigIntFromString(s string) int {
 	// First convert to word, and then find the size,
 	// to be consistent with `common.BigIntByteLength`
 	wordLen := (bitLen + 63) / 64
-	return (wordLen + 4) * BigIntWordSize
+	result := wordLen * BigIntWordSize
```
tarakby, line 59: We can make this check higher in the code :
```
func OverEstimateBigIntFromString(s string) int {
	l := len(s) 
	if l <= 2*BigIntWordSize   { // up to 16 hex characters fit into a 64-bits word
              return BigIntWordSize
         }
```
More details here: https://github.com/dapperlabs/cadence-private-issues/issues/41#issuecomment-1295666060

runtime/parser/expression.go
```diff
@@ -1726,6 +1726,9 @@ func parseIntegerLiteral(p *parser, literal, text []byte, kind IntegerLiteralKin
 		if withoutUnderscores == "" {
 			report(InvalidNumberLiteralKindMissingDigits)
 		} else {
+			estimatedSize := common.OverEstimateBigIntFromString(withoutUnderscores)
+			common.UseMemory(p.memoryGauge, common.NewBigIntMemoryUsage(estimatedSize))
```
SupunS, line 1730: now that the bigInt is metered here, we can remove the metering from the interpreter:
https://github.com/onflow/cadence/blob/ab15f9a9bea9a7e71502514df3d54553c70c1fb5/runtime/interpreter/interpreter_expression.go#L560-L565

SupunS: not for all of them, but for the ones that uses the bigInt as is.
i.e: `Int`, `UIntType`, `Int128Type `, `Int256Type`, `UInt128Type `, `UInt256Type`,  `Integer`, `SignedInteger`

runtime/common/bigint.go
tarakby: @turbolent oh you're right, my bad I was assuming inputs are in base 16.

> Integer literals may have different bases (we support 2, 8, 10, 16)

The function implementation so far only considers base `10`. In general it would be great to cover other bases too, maybe 2 and 16.

runtime/common/bigint.go
```diff
@@ -54,5 +56,9 @@ func OverEstimateBigIntFromString(s string) int {
...
+	if result < minBigIntLength {
+		return minBigIntLength
+	}
+	return result
```
tarakby, line 63: Now that I've been reminded inputs are base 10 and not base 16 :)  I think we can be simplified to something like:
```
        if wordLen == 1 {
		return BigIntWordSize
	}
	return (wordLen + 4) * BigIntWordSize
```

runtime/common/integerliteralkind.go
```diff
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-package parser
+package common
```
SupunS, line 19: Moved this to `common` to eliminate the circular dependency.

runtime/parser/expression.go
SupunS: Removed in https://github.com/onflow/cadence/pull/2110/commits/4cecda419e9b624a0dad3dc4f84913b3c23d5d3e

runtime/common/bigint.go
SupunS: @tarakby  added the support for handling other bases too. Can you please check the equations above?

Also, how the ["test to check if the string fits in one word only"](https://github.com/onflow/cadence/pull/2110/files#r1008597788)" can be added now? I guess this is now needed to be handled separately for each base type?

runtime/common/bigint.go
SupunS: Same question as above. How can this be changed to match the different bases (given that it now supports bases 2, 8, 10, and 16)?

runtime/common/bigint.go
```diff
@@ -42,17 +44,46 @@ func NewBigIntFromAbsoluteValue(gauge MemoryGauge, value *big.Int) *big.Int {
 	return new(big.Int).Abs(value)
 }
 
+const minBigIntLength = 4 * BigIntWordSize
+
 // OverEstimateBigIntFromString is an approximate inverse of `interpreter.OverEstimateBigIntStringLength`.
 // Returns the estimated size in bytes.
-func OverEstimateBigIntFromString(s string) int {
+func OverEstimateBigIntFromString(s string, literalKind IntegerLiteralKind) int {
```
tarakby, line 51: Is it possible that the input `s` has leading zeros? If yes, it would be great to work with `l` as the non-zero length instead of `len(s)`.

runtime/common/bigint.go
```diff
@@ -42,17 +44,46 @@ func NewBigIntFromAbsoluteValue(gauge MemoryGauge, value *big.Int) *big.Int {
...
 	l := len(s)
 
-	// Use 6804/2028 to over estimate log_2(10)
-	bitLen := (l*6804)>>11 + 1
+	// By definition: log_b(v) = log_2(v) / log_2(b)
+	// i.e: log_2(v) = log_b(v) * log_2(b)
+
+	// Each digit in base 'b' requires 'log_2(b)' number of digits to represent in binary.
+	// Therefore, to get the number of digits in base 'b', we could multiply by 'log_2(b)'.
```
tarakby, line 58: Since `log_2(b)` is not necessarily an integer, I suggest to change the wording a bit. 

```suggestion
        // an integer `v` requires `ceiling(log_b(v))` digits in base `b`, and `ceiling(log_2(v))` digits in base `2`. 
        
        // By definition: log_b(v) = log_2(v) / log_2(b)
	// i.e: log_2(v) = log_b(v) * log_2(b)
	// We can therefore overestimate `ceiling(log_2(v))` by `ceiling(log_b(v)) * ceiling(log_b(v))`.
```

runtime/common/bigint.go
```diff
@@ -42,17 +44,46 @@ func NewBigIntFromAbsoluteValue(gauge MemoryGauge, value *big.Int) *big.Int {
...
+		// bitLen = l * log_2(16)
+		bitLen = l * 4
+	default:
+		panic(errors.NewUnreachableError())
+	}
 
 	// Calculate amount of bytes.
 	// First convert to word, and then find the size,
 	// to be consistent with `common.BigIntByteLength`
 	wordLen := (bitLen + 63) / 64
```
tarakby, line 83: I just noticed that we're hardcoding `64` but it's in reality related to `BigIntWordSize`. 
For a cleaner code, I suggest defining another "global" constant `BigIntWordBitSize = BigIntWordSize * 8`. 

And then back to this function `wordLen := (bitLen + BigIntWordBitSize - 1) / BigIntWordBitSize`

runtime/common/bigint.go
tarakby: It's same as the code I wrote above. If `wordLen` is `1`, then `BigIntWordSize` are enough memory, otherwise it is `(wordLen + 4) * BigIntWordSize`. 
I don't think the constant `minBigIntLength` is needed 🤔

runtime/common/bigint.go
SupunS: Good idea!

runtime/common/bigint.go
SupunS: That's a good catch! yes, it is possible. Handled this ind6cc134ad53db865bf3572bb65e60f544d1040a2

tarakby (approved):
Looks good 👌🏼

SupunS (approved):

turbolent (commented):
Nice work! (I can't approve my own PR)

Issue #1956, opened 2022-09-09
Interface Implementation Requirements


turbolent:
### Description

An earlier version of the language reference specified an unimplemented feature: Interface implementation requirements:

> Interfaces can require implementing types to also implement other interfaces of the same kind.
> Interface implementation requirements can be declared by following the interface name with a colon (`:`)
> and one or more names of interfaces of the same kind, separated by commas.
> 
> ```cadence
> // Declare a structure interface named `Shape`.
> //
> pub struct interface Shape {}
> 
> // Declare a structure interface named `Polygon`.
> // Require implementing types to also implement the structure interface `Shape`.
> //
> pub struct interface Polygon: Shape {}
> 
> // Declare a structure named `Hexagon` that implements the `Polygon` interface.
> // This also is required to implement the `Shape` interface,
> // because the `Polygon` interface requires it.
> //
> pub struct Hexagon: Polygon, Shape {}
> ```
> 

This feature is needed for the new FT / NFT standards.

turbolent:
cc @joshuahannan @dsainati1

joshuahannan:
Thank you! I'm excited to have this feature! :)

SupunS:
Added with https://github.com/onflow/cadence/pull/2112

turbolent:
FLIP: https://github.com/onflow/flips/pull/40

PR #2112, opened 2022-11-01
Add interface inheritance

SupunS:
Closes https://github.com/onflow/cadence/issues/1956

## Description

Add support for interfaces to inherit other interfaces.

This is the reference implementation of https://github.com/onflow/flips/pull/40.

📖  Documentation: https://github.com/onflow/cadence/pull/2473

______


runtime/tests/interpreter/interface_test.go
```diff
@@ -241,3 +242,425 @@ func TestInterpretInterfaceDefaultImplementationWhenOverriden(t *testing.T) {
...
+		// Implementation should satisfy both inherited conditions
+
+		_, err = inter.Invoke("main", interpreter.NewUnmeteredIntValueFromInt64(5))
+		utils.RequireError(t, err)
+		assert.ErrorAs(t, err, &interpreter.ConditionError{})
+
+		_, err = inter.Invoke("main", interpreter.NewUnmeteredIntValueFromInt64(25))
+		utils.RequireError(t, err)
+		assert.ErrorAs(t, err, &interpreter.ConditionError{})
+	})
```
turbolent, line 665: Re: https://github.com/onflow/flips/pull/40/files#r1178153289

In what order are pre/post conditions run if there is complex DAG of interfaces?

SupunS: Currently, it is a depth-first traversal. I'll add a test case for that 👍

SupunS: Added here: 9b184ac14bdbbb51f24eca9b09a4a1629c60b337

dsainati1: One potential reason to delay this until Stable Cadence: until the `view` changes restrict mutations in conditions, this may present an attack vector because people will be able to execute arbitrary code in conditions and thus get around the prevention of overriding default implementations.

runtime/tests/checker/interface_test.go
```diff
@@ -2689,3 +2689,1069 @@ func TestCheckBadStructInterface(t *testing.T) {
...
+                pub fun hello() {
+                    var a = 1
+                }
+            }
+        `)
+
+		require.NoError(t, err)
+	})
+
+	t.Run("default impl in both", func(t *testing.T) {
```
dsainati1, line 3263: Can we get a test for `A` with a default impl, `B` inheriting from `A` with no conflict, and then `C` inheriting from `B` that has another default impl to conflict with `A`?

runtime/tests/checker/interface_test.go
```diff
@@ -2689,3 +2689,1069 @@ func TestCheckBadStructInterface(t *testing.T) {
...
+                    }
+                }
+            }
+
+            contract interface C: A, B {}
+
+            contract X: C {
+                struct ANested {
+                    pub fun test(): Int {
+                        return 3
```
dsainati1, line 3528: How exactly do these implementations interact with the default implementation check? Do we need to prevent `C` from declaring a different implementation for `ANested.test()`? If so we should add tests for that.

runtime/tests/checker/interface_test.go
```diff
@@ -2689,3 +2689,1069 @@ func TestCheckBadStructInterface(t *testing.T) {
...
+		assert.Equal(t, common.DeclarationKindStructure, memberConflictError.ConflictingMemberKind)
+	})
+
+	t.Run("nested resource interface conflicting", func(t *testing.T) {
+
+		t.Parallel()
+
+		_, err := ParseAndCheck(t, `
+            contract interface A {
+                resource interface Nested {
```
dsainati1, line 3582: We should add tests for when a nested type interface inherits from another interface, particularly one nested inside a type that is also present inside the containing contract's type heirarchy. E.g.:

```
contract interface A {
    resource interface X: B.Y {}
}

contract interface B: A {
    resource interface Y {}
    resource interface X: Y {}
}
```

Should this count as a cyclical conformance?

runtime/tests/checker/interface_test.go
```diff
@@ -2689,3 +2689,1069 @@ func TestCheckBadStructInterface(t *testing.T) {
...
+                }
+            }
+        `)
+
+		errs := RequireCheckerErrors(t, err, 1)
+
+		memberConflictError := &sema.InterfaceMemberConflictError{}
+		require.ErrorAs(t, errs[0], &memberConflictError)
+		assert.Equal(t, "hello", memberConflictError.MemberName)
+		assert.Equal(t, "A", memberConflictError.ConflictingInterfaceType.QualifiedIdentifier())
```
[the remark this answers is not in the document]
SupunS: yes, correct.

runtime/tests/checker/interface_test.go
```diff
@@ -2689,3 +2689,1069 @@ func TestCheckBadStructInterface(t *testing.T) {
...
+                    pre { true }
+                }
+            }
+
+            struct interface C: A, B {}
+
+            struct D: C {}
+        `)
+
+		// The interface `C` allows to have a default implementation coming from one path,
```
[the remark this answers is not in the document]
SupunS: hmm, I need to look a bit more into this. It's been a while since I've added this, can't exactly remember. But it does sound off a bit. Good catch

runtime/tests/checker/interface_test.go
SupunS: There was no change to the behavior for type requirements as well.

Here the conformance-error is not with the default implementations, but rather because of the missing/unimplemented type requirement. I'll improve the asserts to also check for the details of the error.

runtime/tests/checker/interface_test.go
SupunS: Added in c535fb201ed7837b5b9b65bc852a796b801fc8da

runtime/tests/checker/interface_test.go
SupunS: Improved the asserts in c535fb201ed7837b5b9b65bc852a796b801fc8da

runtime/tests/checker/interface_test.go
SupunS: This is not cyclic, because `B.Y` is just a qualified name (a grouping), and neither `A` nor `A.X` directly depends on `B`.

Anyway, that example is a static error, because there's a conflict in type requirements: both `A` and `B` define their own `X` requirements. Can have only one of them.

Added a test with a simplified version of the example in c535fb201ed7837b5b9b65bc852a796b801fc8da.

runtime/tests/checker/interface_test.go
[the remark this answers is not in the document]
SupunS: OK, so this is also the current behavior. A simplified version is:

```cadence
struct interface A {
    // Have default impl
    pub fun hello() {
        var a = 1
    }
}
            
struct interface B {
    // Declaration only
    pub fun hello()
}

struct C: A, B {}
```

I agree this can be further improved. But given that it's a change to the existing behavior, this is going to be a breaking change in theory if we do this. So I think it should be done in a separate PR, and preferably in stable-cadence.
i.e: no need to block supporting interface inheritance, for an improvement in default functions.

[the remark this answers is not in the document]
SupunS: Created an issue: https://github.com/onflow/cadence/issues/2471

runtime/tests/interpreter/interface_test.go
SupunS: I think it's possible even today, without inheritance. One can invoke arbitrary functions in conditions of two interfaces implemented by a concrete type, where having two default implementations is not allowed.

Interface inheritance in a way only 'flattens' a hierarchy, and is more of a desugar for explicitly writing them down in the concrete implementation.
e.g:
```cadence
struct Foo: A {}

struct interface A : B {}
struct interface B : C {}
struct interface C {}
```

has the same effect as:
```cadence
struct Foo: A, B, C {}

struct interface A {}
struct interface B {}
struct interface C {}
```

runtime/sema/type.go
```diff
@@ -5581,7 +5659,7 @@ func checkSubTypeWithoutEquality(subType Type, superType Type) bool {
 					// TODO: once interfaces can conform to interfaces, include
```
turbolent, line 5659:
```suggestion
```

There are some other "once interfaces can conform to interfaces" comments in the code-base, maybe double-check they are addressed and remove

runtime/sema/type.go
```diff
@@ -5860,7 +5938,7 @@ func checkSubTypeWithoutEquality(subType Type, superType Type) bool {
...
```
turbolent, line 5938:
```suggestion
```

runtime/sema/type.go
```diff
@@ -5966,7 +6044,7 @@ func checkSubTypeWithoutEquality(subType Type, superType Type) bool {
 			}
 
 			// TODO: once interfaces can conform to interfaces, include
```
turbolent, line 6046:
```suggestion
```

runtime/ast/composite.go
```diff
@@ -37,6 +37,7 @@ type CompositeLikeDeclaration interface {
 	Statement
 	Kind() common.CompositeKind
 	ConformanceList() []*NominalType
+	IsInterface() bool
```
turbolent, line 40: This seems to be added because with `InterfaceDeclaration` gaining `func ConformanceList() []*NominalType`, it suddenly conforms to `CompositeLikeDeclaration`.

OK if I push up a commit that removes this function and improves the AST interface conformance?

SupunS: yeah, please do!

turbolent: Pushed up 54e0ce34a714f0aef4d02408d7a8cd01b48108fc

runtime/sema/errors.go
```diff
@@ -1390,11 +1399,22 @@ func (n MemberMismatchNote) Message() string {
...
+func (n NestedConformanceMismatchNote) Message() string {
+	return fmt.Sprintf("does not conform to nested interface requirement `%s`", n.nestedInterfaceType)
+}
+
 // DuplicateConformanceError
 //
 // TODO: just make this a warning?
 type DuplicateConformanceError struct {
-	CompositeType *CompositeType
+	CompositeType CompositeKindedType
```
turbolent, line 1417: Maybe rename `CompositeType` to `CompositeKindedType`, because it may now also be an `InterfaceType`.
Same for the other errors below

runtime/sema/type.go
```diff
@@ -4542,6 +4563,63 @@ func (t *InterfaceType) FieldPosition(name string, declaration *ast.InterfaceDec
...
+
+// distinctConformances recursively visit conformances and their conformances,
+// and return all the distinct conformances as an array.
+func distinctConformances(
+	conformances []*InterfaceType,
+	parent *InterfaceType,
+	seenConformances map[*InterfaceType]struct{},
+) []Conformance {
+
+	collectedConformances := make([]Conformance, 0)
```
turbolent, line 4586: Maybe avoid allocating empty slices and pre-allocate with capacity:

```suggestion
	if len(conformances) == 0 {
		return nil
	}

	collectedConformances := make([]Conformance, 0, len(conformances))
```

runtime/sema/type.go
```diff
@@ -3734,23 +3741,33 @@ func (t *CompositeType) Tag() TypeTag {
...
-		t.explicitInterfaceConformanceSet = NewInterfaceSet()
-		for _, conformance := range t.ExplicitInterfaceConformances {
-			t.explicitInterfaceConformanceSet.Add(conformance)
+		for _, conformance := range t.EffectiveInterfaceConformances() {
+			t.effectiveInterfaceConformanceSet.Add(conformance.InterfaceType)
 		}
 	})
 }
 
+func (t *CompositeType) EffectiveInterfaceConformances() []Conformance {
```
turbolent, line 3759: Just curious: Is it necessary we keep both a set and a slice of the effective interface conformances?

runtime/sema/checker.go
```diff
@@ -341,11 +341,22 @@ func (checker *Checker) CheckProgram(program *ast.Program) {
 		}
 	}
 
+	// NOTE: Resolving interface conformances and registering types in elaboration
```
turbolent, line 344: 👍

runtime/sema/type.go
SupunS: This is pretty much the same as what's there currently:
```go
explicitInterfaceConformanceSet     *InterfaceSet
ExplicitInterfaceConformances       []*InterfaceType
```

We could construct the slice the first time, but that would undo some other optimizations, like unnecessarily constructing a set, when it is not actually needed, etc. Not sure which is more expensive tbh, i.e: storing the interim result (the slice) vs always constructing the set.

runtime/sema/type.go
SupunS: Added in https://github.com/onflow/cadence/pull/2112/commits/c1486dbc47b13a5181c1a94e44767ad0a6a61afa. However, can't set the array capacity, since this is a recursive call and the size is unknown.

runtime/tests/checker/interface_test.go
```diff
@@ -3988,4 +4030,62 @@ func TestCheckInheritedInterfacesSubtyping(t *testing.T) {
...
+
+            struct S: B, C {}
+
+            // Case I: &S{B, C} is a subtype of &S{B}
+            fun foo(): &S{B} {
+                var s: S{B, C} = S()
+                return &s as &S{B, C}
+            }
+
+            // Case II: &S{B} is a subtype of &S{A}
```
dsainati1, line 4082: Can we get a test for the combined case where `{B, C}` is a subtype of `{A}`?

SupunS: Added in fe0e730efa45ca8795fa6d5e7299ceb2d9dacb99

runtime/sema/check_composite_declaration.go
```diff
@@ -1416,6 +1466,14 @@ func (checker *Checker) memberSatisfied(compositeMember, interfaceMember *Member
...
 			}
 
+		case common.DeclarationKindStructure,
+			common.DeclarationKindResource,
+			common.DeclarationKindEnum:
+			// Interfaces and their conformances cannot have nested composite declarations
+			// with conflicting names (i.e: no type requirements for interfaces).
+			if _, isInterface := compositeType.(*InterfaceType); isInterface {
+				return false
+			}
```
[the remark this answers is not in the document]
SupunS: Good catch, this is not actually needed. This constraint is checked separately in `checkInterfaceConformance` function.
Removed and added a test: https://github.com/onflow/cadence/pull/2112/commits/ff5976edcebc5b6915400e4f5a8ae1359c291752

runtime/tests/checker/interface_test.go
```diff
@@ -2689,3 +2689,1459 @@ func TestCheckBadStructInterface(t *testing.T) {
...
+		_, err := ParseAndCheck(t, `
+            struct interface A {
+                pub fun hello()
+            }
+
+            struct interface B: A {
+                pub fun hello() {
+                    var a = 1
+                }
+            }
```
turbolent, line 3301: Currently, it is an error when a concrete type conforms to interfaces where there is a requirement and a default function (#2471), e.g. having just `struct interface B` and adding `struct C: A, B` here.

Should this also be enforced in the interface inheritance case?

turbolent (approved):
Great work!

runtime/tests/checker/interface_test.go
SupunS: I'll also add the same restriction here, and probably document this as unimplemented/deviation from the FLIP. 👍

turbolent: I see, I had not realized that the FLIP specified a different behaviour. Though maybe different than from the FLIP, it might be better to be consistent and avoid confusion for developers.

Like discussed and suggested by you, we can then maybe later put out a FLIP and propose to lift the restrictions everywhere, keeping a consistent behaviour.

SupunS: Added the restriction here: ed64d6f2a611ea5280862a2bf6fb85e4f84afdcd

SupunS:
@turbolent @dsainati1 can you please have a look at the last commit ed64d6f2a611ea5280862a2bf6fb85e4f84afdcd 🙏 

This is basically adding the restriction described in https://github.com/onflow/cadence/issues/2471, for interfaces as well, to make 'interface inheritance' and 'concrete type interface conformance' behave the same way. This would be relaxed back after https://github.com/onflow/flips/pull/83.

turbolent (approved):
Nice! Thanks for the follow-up work 🙏

dsainati1 (approved):

PR #2113, opened 2022-11-01
Interpreting for basic attachment operations

dsainati1:
Part of https://github.com/onflow/cadence/issues/357 and https://github.com/onflow/cadence/issues/2062

This implements runtime semantics for attaching, accessing and removing attachments

______


robert-e-davidson3 (approved):

runtime/interpreter/value.go
```diff
@@ -15023,6 +15194,88 @@ func NewEnumCaseValue(
...
+}
+
+func (v *CompositeValue) forEachAttachment(interpreter *Interpreter, locationRange LocationRange, f func(*CompositeValue)) {
+	iterator, err := v.dictionary.Iterator()
+	if err != nil {
+		panic(errors.NewExternalError(err))
+	}
+
+	interpreter.SharedState.inAttachmentIteration[v] = struct{}{}
+	defer delete(interpreter.SharedState.inAttachmentIteration, v)
```
dsainati1, line 15222: Every time we enter an iteration over a composite's attachment we store that composite in this set, and remove it when we're done. This can't just be a `bool` like the storage iteration because we only want to restrict mutations of this specific composite's attachments, not all attachments generally.

runtime/interpreter/interpreter_expression.go
```diff
@@ -1142,7 +1183,50 @@ func (interpreter *Interpreter) VisitPathExpression(expression *ast.PathExpressi
...
+		false,
+		base,
+		interpreter.MustSemaTypeOfValue(base).(*sema.CompositeType),
+	)
+	// technically this makes `super` available to the arguments to the attachment constructor, but we enforce statically
+	// that it cannot appear there, so it's ok
+	interpreter.declareVariable(sema.SuperIdentifier, baseValue)
+
+	attachment, ok := interpreter.VisitInvocationExpression(attachExpression.Attachment).(*CompositeValue)
+	interpreter.activations.Remove(sema.SuperIdentifier)
```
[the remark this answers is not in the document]
dsainati1: I have since determined that this is actually not safe; if `base` is already in scope (i.e. if we are inside an attachment body) this will shadow the existing `base` value too early. Working on a new approach for this now.

dsainati1:
The reflection and iteration API features need some more discussion, but we don't want to block development of the basic parts of this feature on those other parts. So pending discussion on https://github.com/onflow/flips/pull/50 I've removed the parts of interpreting that pertained to these features, and will add them back in future PRs once we come to a decision about how they should work.

runtime/interpreter/interpreter_expression.go
```diff
@@ -1142,7 +1184,66 @@ func (interpreter *Interpreter) VisitPathExpression(expression *ast.PathExpressi
...
+		nil,
+	).(*CompositeValue)
+
+	// we enforce this in the checker
+	if !ok {
+		panic(errors.NewUnreachableError())
+	}
+
+	// when `v[A]` is executed, we set `A`'s base to `&v`
+	attachment.setBaseValue(interpreter, base)
```
SupunS, line 1239: Can't we directly pass the `baseValue` here,  instead of passing `base` and creating another `EphemeralReferenceValue` inside it?

runtime/interpreter/interpreter_statement.go
```diff
@@ -401,8 +401,46 @@ func (interpreter *Interpreter) VisitEmitStatement(statement *ast.EmitStatement)
...
+	}
+
+	attachment, ok := removed.(*CompositeValue)
+	// we enforce this in the checker
+	if !ok {
+		panic(errors.NewUnreachableError())
+	}
+
+	if attachment.IsResourceKinded(interpreter) {
+		// this attachment is no longer attached to its base, but `super` is still available in the destructor
```
SupunS, line 439:
```suggestion
		// this attachment is no longer attached to its base, but `base` variable is still available in the destructor
```

runtime/interpreter/value.go
```diff
@@ -14038,6 +14058,14 @@ func (v *CompositeValue) Destroy(interpreter *Interpreter, locationRange Locatio
 		}()
 	}
 
+	// if this type has attachments, destroy all of them before invoking the destructor
+	v.forEachAttachment(interpreter, locationRange, func(attachment *CompositeValue) {
+		// an attachment's destructor may make reference to `base`, so we must set the base value
+		// for the attachment before invoking its destructor
+		attachment.setBaseValue(interpreter, v)
```
SupunS, line 14065: Why do we need to explicitly set the base here? Wouldn't the destructor work similarly to any other function defined on the attachment?

runtime/interpreter/value.go
```diff
@@ -15023,6 +15096,97 @@ func NewEnumCaseValue(
...
+	return v.GetMember(interpreter, locationRange, attachmentNamePrefix+string(ty.ID()))
+}
+
+func (v *CompositeValue) forEachAttachment(interpreter *Interpreter, locationRange LocationRange, f func(*CompositeValue)) {
+	iterator, err := v.dictionary.Iterator()
+	if err != nil {
+		panic(errors.NewExternalError(err))
+	}
+
+	old_shared_state := interpreter.SharedState.inAttachmentIteration[v]
```
SupunS, line 15131:
```suggestion
	oldSharedState := interpreter.SharedState.inAttachmentIteration[v]
```

runtime/interpreter/value.go
```diff
@@ -16889,16 +17063,7 @@ func (v *StorageReferenceValue) RemoveMember(
...
-	if referencedValue == nil {
-		panic(DereferenceError{
-			LocationRange: locationRange,
-		})
-	}
-
-	self := *referencedValue
-
-	interpreter.checkReferencedResourceNotDestroyed(self, locationRange)
+	self := v.mustReferencedValue(interpreter, locationRange)
```
SupunS, line 17066: 👌

runtime/interpreter/value.go
dsainati1: Any other function defined on the attachment must be accessed via an access expression `v[A]` on the composite value `v`. This access expression is what populates the `base` field on `A`. Destroying `v` will invoke the destructor on `A` though without going through an access expression, so we need to populate the field before calling the function.

runtime/interpreter/interpreter_expression.go
```diff
@@ -37,6 +38,9 @@ func (interpreter *Interpreter) assignmentGetterSetter(expression ast.Expression
 		return interpreter.identifierExpressionGetterSetter(expression)
 
 	case *ast.IndexExpression:
+		if attachmentType, ok := interpreter.Program.Elaboration.AttachmentAccessTypes[expression]; ok {
```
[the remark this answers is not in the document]
dsainati1: As before, the issue here is `*EphemeralReferenceValue`, which implements both interfaces, but is not necessarily a value-indexable value or a type-indexable value. Put another way, presence in the `AttachmentAccessTypes` implies that a value is a `TypeIndexableValue`, but being a `TypeIndexableValue` doesn't guarantee that there will be an attachment in that map, or even that the value isn't a `ValueIndexableValue`.

runtime/interpreter/interpreter_expression.go
dsainati1: This function is used elsewhere and only takes a `CompositeValue`. To change this to take a reference value would require all the call sites to construct their own reference values instead of having the logic encapsulated inside this function.

runtime/interpreter/interpreter_expression.go
```diff
@@ -1142,7 +1184,66 @@ func (interpreter *Interpreter) VisitPathExpression(expression *ast.PathExpressi
...
+	// constructor to make it available
+	baseValue := NewEphemeralReferenceValue(
+		interpreter,
+		false,
+		base,
+		interpreter.MustSemaTypeOfValue(base).(*sema.CompositeType),
+	)
+	interpreter.trackReferencedResourceKindedValue(base.StorageID(), base)
+
+	interpreter.SharedState.deferredBaseValue = baseValue
```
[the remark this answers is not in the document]
dsainati1: Great catch! Fixed

runtime/interpreter/interpreter_expression.go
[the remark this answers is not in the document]
SupunS: ah ok, that makes sense. Maybe add this as a comment

runtime/interpreter/value.go
SupunS: Gotcha. This would also be good to include as a comment in the code

SupunS (approved):

runtime/interpreter/sharedstate.go
```diff
@@ -49,6 +51,7 @@ func NewSharedState(config *Config) *SharedState {
 		},
 		inStorageIteration:             false,
 		storageMutatedDuringIteration:  false,
+		inAttachmentIteration:          map[*CompositeValue]bool{},
```
turbolent, line 54: Maybe initialize this lazily, so we do not have the pay the cost of allocating it even when the program involves no attachments.

runtime/interpreter/value.go
```diff
@@ -14169,8 +14218,14 @@ func (v *CompositeValue) GetMember(interpreter *Interpreter, locationRange Locat
 
 	function, ok := v.Functions[name]
 	if ok {
+		var base *EphemeralReferenceValue
 		var self MemberAccessibleValue = v
-		return NewBoundFunctionValue(interpreter, function, &self)
+		if v.Kind == common.CompositeKindAttachment {
+			base = v.getBaseValue(interpreter, locationRange)
+			// in attachment functions, self is a reference value
+			self = NewEphemeralReferenceValue(interpreter, false, v, interpreter.MustConvertStaticToSemaType(v.StaticType(interpreter)))
```
turbolent, line 14226: Can use the new helper function for getting the sema type for the value.

runtime/interpreter/value.go
```diff
@@ -15023,6 +15100,97 @@ func NewEnumCaseValue(
 	return v
 }
 
+func (v *CompositeValue) getBaseValue(interpreter *Interpreter, locationRange LocationRange) *EphemeralReferenceValue {
+	return v.base
+}
+
+func (v *CompositeValue) setBaseValue(interpreter *Interpreter, base *CompositeValue) {
+	attachmentType, ok := interpreter.MustConvertStaticToSemaType(v.StaticType(interpreter)).(*sema.CompositeType)
```
turbolent, line 15108: Can use the new helper function for getting the sema type for the value.

runtime/interpreter/value.go
```diff
@@ -14047,11 +14079,18 @@ func (v *CompositeValue) Destroy(interpreter *Interpreter, locationRange Locatio
 
 	destructor := v.Destructor
 
+	var base *EphemeralReferenceValue
+	var self MemberAccessibleValue = v
+	if v.Kind == common.CompositeKindAttachment {
+		base = v.getBaseValue(interpreter, locationRange)
+		self = NewEphemeralReferenceValue(interpreter, false, v, interpreter.MustConvertStaticToSemaType(v.StaticType(interpreter)))
```
turbolent, line 14086: Can use the new helper function for getting the sema type for the value.

runtime/interpreter/value.go
```diff
@@ -15023,6 +15100,97 @@ func NewEnumCaseValue(
...
+		baseType, _ = ty.RewriteWithRestrictedTypes()
+	default:
+		baseType = ty
+	}
+
+	// the base reference can only be borrowed with the declared type of the attachment's base
+	v.base = NewEphemeralReferenceValue(interpreter, false, base, baseType)
+}
+
+func (v *CompositeValue) getAttachmentValue(interpreter *Interpreter, locationRange LocationRange, ty sema.Type) Value {
```
turbolent, line 15125: Callers cast to `*CompositeValue` immediately. Maybe do the cast in a centralized place here, so the assumption is in one place.

runtime/interpreter/value.go
```diff
@@ -15023,6 +15100,97 @@ func NewEnumCaseValue(
...
+	return NewSomeValueNonCopying(interpreter, NewEphemeralReferenceValue(interpreter, false, attachment, ty))
+}
+
+func (v *CompositeValue) SetTypeKey(
+	interpreter *Interpreter,
+	locationRange LocationRange,
+	attachmentType sema.Type,
+	attachment Value,
+) {
+	if v.SetMember(interpreter, locationRange, attachmentNamePrefix+string(attachmentType.ID()), attachment) {
```
turbolent, line 15177: Maybe refactor `attachmentNamePrefix+` into a function. See alsp other uses of generating them member name for the attachment, like `getAttachmentValue` above and `RemoveTypeKey` below

runtime/interpreter/value.go
```diff
@@ -15310,10 +15478,10 @@ func (v *DictionaryValue) ForEachKey(
 	keyType := v.SemaType(interpreter).KeyType
 
 	iterationInvocation := func(key Value) Invocation {
-		var self MemberAccessibleValue = v
 		return NewInvocation(
 			interpreter,
-			&self,
+			nil,
+			nil,
```
turbolent, line 15484: This looks like a "bug-fix", i.e. is the reason for the change that the key-iteration callback does not need to have the `self` value declared?

runtime/interpreter/value.go
```diff
@@ -16865,10 +17035,9 @@ func (v *StorageReferenceValue) ReferencedValue(interpreter *Interpreter) *Value
 	return value
 }
 
-func (v *StorageReferenceValue) GetMember(
+func (v *StorageReferenceValue) mustReferencedValue(
```
turbolent, line 17038: Nice cleanup 👍

runtime/sema/checker.go
```diff
@@ -1762,8 +1762,8 @@ func (checker *Checker) checkCharacterLiteral(expression *ast.StringExpression)
 	)
 }
 
-func (checker *Checker) isReadableAccess(access ast.Access) bool {
-	switch checker.Config.AccessCheckMode {
+func (mode AccessCheckMode) IsReadableAccess(access ast.Access) bool {
```
turbolent, line 1765: Good idea to decouple this from `Checker` by changing the receiver. Please move it to `AccessCheckMode` (`accesscheckmode.go`). Same for `IsWriteableAccess` below.

turbolent (commented):
Great work!

runtime/interpreter/interpreter_expression.go
```diff
@@ -1142,7 +1184,66 @@ func (interpreter *Interpreter) VisitPathExpression(expression *ast.PathExpressi
...
+		Location:    interpreter.Location,
+		HasPosition: attachExpression,
+	}
+
+	if inIteration := interpreter.SharedState.inAttachmentIteration[base]; inIteration {
+		panic(AttachmentIterationMutationError{
+			Value:         base,
+			LocationRange: locationRange,
+		})
+	}
```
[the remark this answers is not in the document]
SupunS: oh, this is no longer valid. I was under the assumption that iteration was/will be removed altogether until the reflection APIs are finalized, which is not the case.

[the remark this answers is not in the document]
turbolent: Ah, I see, this is a left-over from the iteration API. Agreed, maybe remove this and other iteration-related code that is still in the current PRs.

runtime/interpreter/value.go
dsainati1: Yes, the key iteration function should not have a `self` value declared. I don't think this was abusable because use of `self` would have failed at checking time but we were allocating a variable unnecessarily here.

runtime/interpreter/interpreter_expression.go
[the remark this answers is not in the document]
dsainati1: This check is actually still necessary. Although the ability for users to explicitly iterate over attachments with `forEachAttachment` has been removed from this PR, implicit iteration can still happen when a resource containing attachments is `destroy`ed: we iterate over all the attachments on that resource and destroy them one by one. We still need to make sure attachments are not added or removed from that resource during that iteration, as it could cause the order to change and thus one of the attachments (and potential nested resource) to not be destroyed.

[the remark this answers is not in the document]
dsainati1: Added a test demonstrating this case in https://github.com/onflow/cadence/pull/2113/commits/8499a7c79e1bfd5af982a6fe55952d3b74d6861b

runtime/interpreter/sharedstate.go
```diff
@@ -51,8 +51,15 @@ func NewSharedState(config *Config) *SharedState {
...
 		resourceVariables:              map[ResourceKindedValue]*Variable{},
 	}
 }
+
+func (s *SharedState) inAttachmentIteration(base *CompositeValue) bool {
+	if s.attachmentIterationMap == nil {
+		s.attachmentIterationMap = map[*CompositeValue]bool{}
+		return false
+	}
+	return s.attachmentIterationMap[base]
```
turbolent, line 63: Instead of making the getter allocate, move the allocation to the setter

runtime/interpreter/interpreter.go
```diff
@@ -1151,6 +1187,14 @@ func (interpreter *Interpreter) declareNonEnumCompositeValue(
 				value.Destructor = destructorFunction
 
 				var self MemberAccessibleValue = value
+				if declaration.CompositeKind == common.CompositeKindAttachment {
+					self = NewEphemeralReferenceValue(interpreter, false, value, interpreter.MustSemaTypeOfValue(value))
+					// set the base to the implicitly provided value, and remove this implicit argument from the list
+					implicitArgumentPos := len(invocation.Arguments) - 1
+					invocation.Base = invocation.Arguments[implicitArgumentPos].(*EphemeralReferenceValue)
+					invocation.Arguments = invocation.Arguments[:implicitArgumentPos]
+					invocation.ArgumentTypes = invocation.ArgumentTypes[:implicitArgumentPos]
```
turbolent, line 1196: Also, clear out the underlying pointer to the argument value and argument type:

```suggestion
					invocation.Arguments[:implicitArgumentPos] = nil
					invocation.Arguments = invocation.Arguments[:implicitArgumentPos]
					
					invocation.ArgumentTypes[:implicitArgumentPos] = nil
					invocation.ArgumentTypes = invocation.ArgumentTypes[:implicitArgumentPos]
```

turbolent:
CI is broken because of the automatic base branch change (GitHub bug), merging master should fix it and retrigger CI

runtime/interpreter/value.go
```diff
@@ -14186,8 +14234,14 @@ func (v *CompositeValue) GetMember(interpreter *Interpreter, locationRange Locat
...
 	function, ok := v.Functions[name]
 	if ok {
+		var base *EphemeralReferenceValue
 		var self MemberAccessibleValue = v
-		return NewBoundFunctionValue(interpreter, function, &self)
+		if v.Kind == common.CompositeKindAttachment {
+			base = v.getBaseValue(interpreter, locationRange)
+			// in attachment functions, self is a reference value
+			self = NewEphemeralReferenceValue(interpreter, false, v, interpreter.MustSemaTypeOfValue(v))
+		}
```
turbolent, line 14243: Maybe refactor into a function, and also use above https://github.com/onflow/cadence/pull/2113/files#diff-424575c293dc2bc3a28aab845829d7656fa83a7be4e026de8906935e23bbd52eR14100-R14103

runtime/interpreter/value.go
```diff
@@ -15040,6 +15116,104 @@ func NewEnumCaseValue(
...
+
+func (v *CompositeValue) forEachAttachment(interpreter *Interpreter, locationRange LocationRange, f func(*CompositeValue)) {
+	iterator, err := v.dictionary.Iterator()
+	if err != nil {
+		panic(errors.NewExternalError(err))
+	}
+
+	oldSharedState := interpreter.SharedState.inAttachmentIteration(v)
+	interpreter.SharedState.setAttachmentIteration(v, true)
+	defer func() { interpreter.SharedState.setAttachmentIteration(v, oldSharedState) }()
```
turbolent, line 15160:
```suggestion
	defer func() {
		interpreter.SharedState.setAttachmentIteration(v, oldSharedState)
	}()
```

runtime/interpreter/value.go
```diff
@@ -15040,6 +15116,104 @@ func NewEnumCaseValue(
...
+}
+
+func (v *CompositeValue) getAttachmentValue(interpreter *Interpreter, locationRange LocationRange, ty sema.Type) *CompositeValue {
+	if attachment := v.GetMember(interpreter, locationRange, attachmentMemberName(ty)); attachment != nil {
+		return attachment.(*CompositeValue)
+	}
+	return nil
+}
+
+func (v *CompositeValue) forEachAttachment(interpreter *Interpreter, locationRange LocationRange, f func(*CompositeValue)) {
```
turbolent, line 15152: Maybe remove iteration-related code from PR for now

runtime/interpreter/value.go
```diff
@@ -15040,6 +15116,104 @@ func NewEnumCaseValue(
...
+		}
+		if key == nil {
+			break
+		}
+		if strings.HasPrefix(string(key.(StringAtreeValue)), attachmentNamePrefix) {
+			attachment, ok := MustConvertStoredValue(interpreter, value).(*CompositeValue)
+			if !ok {
+				panic(errors.NewExternalError(err))
+			}
+			f(attachment)
```
turbolent, line 15175: Should this be a reference?

runtime/interpreter/sharedstate.go
```diff
@@ -53,3 +54,14 @@ func NewSharedState(config *Config) *SharedState {
...
+func (s *SharedState) inAttachmentIteration(base *CompositeValue) bool {
+	return s.attachmentIterationMap[base]
+}
+
+func (s *SharedState) setAttachmentIteration(base *CompositeValue, b bool) {
+	if s.attachmentIterationMap == nil {
+		s.attachmentIterationMap = map[*CompositeValue]bool{}
+	}
+	s.attachmentIterationMap[base] = b
+}
```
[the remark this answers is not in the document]
dsainati1: As mentioned here (https://github.com/onflow/cadence/pull/2113#discussion_r1038252290), it's not unused

runtime/interpreter/value.go
dsainati1: As mentioned here (https://github.com/onflow/cadence/pull/2113#discussion_r1038252290), it's not unused

runtime/interpreter/value.go
dsainati1: No. In this PR the only place this is used is to iterate over the attachments on a resource when `destroy`ing it; and as such the attachments are passed directly to the function so they can be `destroy`ed themselves. In prior versions of this PR when this function was also used to implement explicit iteration via `forEachAttachment`, the `f` provided would convert the composite to a reference before calling the user-provided callback argument.

runtime/interpreter/interpreter_expression.go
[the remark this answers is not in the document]
turbolent: Ah, right! 👍

runtime/interpreter/value.go
turbolent: Makes sense! Maybe add a comment explaining this

turbolent (approved):
Great work! 👏

PR #2115, opened 2022-11-01
Fix whitespace parsing between expressions

### LINKED ISSUE #1119: Incorrect parsing (as function call) when something in braces is in line next to expression
### Instructions

### Problem

This code

```
pre {
    (royaltyReceiver == nil) == (royaltyPercent == nil)
    (royaltyPercent == nil ) || (royaltyPercent! <= self.maxRoyaltyPercent)
}
```

raises error:

```
error: cannot call type: `Bool`
E             --> f8d6e0586b0a20c7.XXX:57:13
E              |
E           57 |             (royaltyReceiver == nil) == (royaltyPercent == nil)
E              |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

While other code with same meaning works without interpreter errors:

```
 pre {
    payment.balance == self.priceFor(tokenID: tokenID)
    self.isNFTStored(tokenID: tokenID)
}
```

### Steps to Reproduce

Version of emulator used:

```
flow version  
Version: v0.25.0
Commit: 9afd7938ebf603611d7b159278a259c850e0adda
```

### Acceptance Criteria

To me it looks like parsing ambiguity bug, I think code should be accepted.

turbolent:
Closes #1119 

## Description

Fix the parsing of whitespace/newlines between expressions. Even though we had logic in place to prevent parsing when a newline is encountered, it did not handle all cases and was also unnecessarily complicated.
The issue was that the space token was always consumed, but should only be conditionally skipped.

While at it:
- Improve the parsing of array and dictionary expressions
- Improve the AST explorer:
  - Update the dependencies
  - Switch from using a WASM binary of the parser to a Go HTTP API, which allows easier iteration: after an update of the parser, only the API has to be restarted, which is fast and can be done automatically using a file change watcher

This is pretty much a bugfix, and existing code should keep parsing as-is. I will verify this by validating the Mainnet contracts.
However, given that parsing behaviour changed (for the better), this change must theoretically still be seen as a breaking change, as existing code could depend on the old parsing behaviour.
We might need to/want to make this a Stable Cadence change.

______

### DIFF
--- a/tools/astexplorer/src/index.tsx
+++ b/tools/astexplorer/src/index.tsx
@@ -1,7 +1,7 @@
 /*
  * Cadence - The resource-oriented smart contract programming language
  *
- * Copyright 2019-2021 Dapper Labs, Inc.
+ * Copyright 2019-2022 Dapper Labs, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,13 +19,11 @@
 
 import * as monaco from "monaco-editor";
 import configureCadence, {CADENCE_LANGUAGE_ID} from "./cadence";
-import {CadenceParser} from "@onflow/cadence-parser"
-
 import * as React from "react"
-import * as ReactDOM from "react-dom"
-import {TreeView} from "./tree"
+import { Data, TreeView } from "./tree";
+import { createRoot } from 'react-dom/client';
 
-const code = `
+const defaultCode = `
 pub contract C {
     pub resource R {}
     pub fun createR(): @R {
@@ -33,6 +31,10 @@ pub contract C {
     }
 }
 `
+
+function Error({ children }: {children?: React.ReactNode[]}) {
+  return <div className='error'>{children}</div>;
+}
 
 interface Position {
   Offset: number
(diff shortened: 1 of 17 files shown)


robert-e-davidson3 (approved):

tools/astexplorer/src/index.tsx
```diff
@@ -75,20 +79,53 @@ document.addEventListener('DOMContentLoaded', async () => {
     }
   );
 
-  const parser = await CadenceParser.create("cadence-parser.wasm")
+  const root = createRoot(astElement)
+  const render = root.render.bind(root)
```
dreamsmasher, line 83: you can use `root::render` with @babel/plugin-proposal-function-bind. but we also don't use babel for the ast explorer 😅

dreamsmasher (approved):

SupunS (approved):
LGTM!

turbolent:
Follow-up: I analyzed all Mainnet contracts and noticed some existing contracts broke. This is fixed with #2136. With the fix, all contracts that currently parse on Mainnet still do, and parsing results are the same for them after this PR. 

TLDR: Theoretically a breaking change, but no impact on Mainnet contracts.

PR #2117, opened 2022-11-02
Improve view modifier parsing and printing

turbolent:
## Description

Noticed this while working on related code:

- Forbid `view` modifier for pragmas
- Ensure `view` keyword is declared as such
- Update pretty-printing of function declarations, expressions, and types to include `view` modifier

______

### DIFF
--- a/runtime/parser/declaration.go
+++ b/runtime/parser/declaration.go
@@ -75,6 +75,9 @@ func parseDeclaration(p *parser, docString string) (ast.Declaration, error) {
 
 		switch p.current.Type {
 		case lexer.TokenPragma:
+			if purity != ast.FunctionPurityUnspecified {
+				return nil, p.syntaxError("invalid view modifier for pragma")
+			}
 			if access != ast.AccessNotSpecified {
 				return nil, p.syntaxError("invalid access modifier for pragma")
 			}
--- a/runtime/parser/declaration.go
+++ b/runtime/parser/declaration.go
@@ -76,17 +76,17 @@ func parseDeclaration(p *parser, docString string) (ast.Declaration, error) {
 		switch p.current.Type {
 		case lexer.TokenPragma:
 			if purity != ast.FunctionPurityUnspecified {
-				return nil, p.syntaxError("invalid view modifier for pragma")
+				return nil, NewSyntaxError(*purityPos, "invalid view modifier for pragma")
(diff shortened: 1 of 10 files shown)


runtime/parser/declaration.go
```diff
@@ -75,6 +75,9 @@ func parseDeclaration(p *parser, docString string) (ast.Declaration, error) {
```
turbolent, line 75: @dsainati1 Do we want to maybe enforce that the order of the access modifier and the view modifier is access then view, and forbid view then access?

dsainati1: Is there a good reason to do so? In the long run if we add more keywords I don’t think we’d want to enforce that they appear in a specific order.

dsainati1 (approved):

runtime/parser/declaration.go
turbolent: Mainly to enforce consistency and improve readability. For example, imagine a case:

```cadence
resource R {
  ...
  
  pub view fun foo() { ... }

  pub fun bar() { ... }

  view pub fun baz() { ... }

  pub fun blub() { ... }

}
```

IMHO it's hard to see the third function is also public. 

It might better to only have one way, so there isn't even a question if order matters.
Then again I don't know if any other language enforces such an order.

turbolent:
@dsainati1 Also added missing pretty printing in [188d81d](https://github.com/onflow/cadence/pull/2117/commits/188d81d0ab7c6ceb815809127679418623f11e11). Could you please have another look?

runtime/parser/declaration.go
turbolent: Also, just noticed that the access modifier position is assumed the start position of the function declaration at the moment

dsainati1: If we were to enforce an order I think it might make sense to require the access modifier first, and then any other annotation keywords in any order after that. Iirc Java requires `public` or `private` or such to come first, and then allows modifiers like `static`, `final` or `volatile` to come after.

turbolent: Agreed

dreamsmasher (approved):

dreamsmasher:
LGTM, let's get this merged soon - I'm going to base my work for #2120 on these commits for the time being.

PR #2119, opened 2022-11-03
Add static and native modifiers for fields and functions

turbolent:
## Description

Currently the majority of the built-in and standard library types of Cadence are implemented in Go.
This requires a lot of boilerplate and related definitions are hard to find.

It would be great if we would could declare built-in and standard library types in Cadence, like we already do for e.g. the `Crypto` contract.

However, built-in and standard library types, as well as their functionality, cannot be expressed using the current syntax and functionality available in Cadence.

To allow the declaration in Cadence, add two new declaration modifiers, for now syntactically allowed for fields and functions, and for now semantically rejected for any declaration. These modifiers will only be available for checking of built-in and standard library types shipped with Cadence, not for on-chain Cadence programs, such as transactions/scripts and deployed contracts.

- `native`: Expresses the notion that the function body is implemented natively, i.e. as an interpreter host function; and for a field, that its value does not have to be initialized, and accesses are implemented natively, i.e. using computed fields.  In a future PR, native functions will be required to have an empty body and checks for function bodies, like they have a return statement, will be disabled; native fields will not be required to get initialized, etc. 

- `static`: Expresses the notion that the field/function is available on the type itself, e.g. `String.fromUTF8`.  In a future PR, static fields/functions will be defined on the type.

This first PR only adds support for parsing and keeping the modifiers in the AST. For now, any usage will be semantically rejected. Future PRs will enable functionality as described above. 

This will eventually enable the new built-in and standard library types to be defined using this mechanism, and existing types to be refactored to using this mechanism.

______


runtime/sema/check_composite_declaration.go
```diff
@@ -1044,7 +1043,7 @@ func (checker *Checker) checkCompositeConformance(
 	compositeKindMismatchIdentifier ast.Identifier,
 	options compositeConformanceCheckOptions,
 	inheritedMembers map[string]struct{},
-	// type requirement name -> inherited members
+// type requirement name -> inherited members
```
turbolent, line 1046:
```suggestion
	// type requirement name -> inherited members
```

dsainati1:
Has this been proposed in a FLIP? Adding new keywords like this feels like the kind of thing we should propose first to the community. 

Also I think we discussed this here https://github.com/onflow/cadence/pull/2117 but how do these new modifiers interact with `view`?

turbolent:
I see this mostly as an internal language feature, an addition to the language that implementors can use, but has no impact on users of the language, so IDK if we need a FLIP for it

dsainati1:
I agree that it has no current impact on users, but it does restrict the potential future direction of development. I.e. if we want to later add `static` functions as a first-class language feature that users can use, we are effectively locked in to the implementation here because it already exists for builtin values. IMO for that reason we should get buy in from users to make sure that if we do open these up to users at some point in the future for any reason, it's with a design that everyone is on board with.

runtime/ast/composite.go
```diff
@@ -257,10 +257,20 @@ func CompositeDocument(
...
 }
 
+// FieldDeclarationFlags
+
+type FieldDeclarationFlags uint8
+
+const (
+	FieldDeclarationFlagsIsStatic FieldDeclarationFlags = 1 << iota
+	FieldDeclarationFlagsIsNative
+)
```
SupunS, line 267: Good idea to have these as flags

runtime/sema/checker.go
```diff
@@ -2337,3 +2337,23 @@ func wrapWithOptionalIfNotNil(typ Type) Type {
...
+		checker.report(
+			&InvalidStaticModifierError{
+				Range: ast.NewRangeFromPositioned(checker.memoryGauge, position),
+			},
+		)
+	}
+}
+
+func (checker *Checker) checkNativeModifier(isNative bool, position ast.HasPosition) {
+	if isNative && !checker.Config.AllowStaticFunctions {
```
SupunS, line 2352:
```suggestion
	if isNative && !checker.Config.AllowNativeFunctions {
```

SupunS (approved):

SupunS:
Here we are only sort of reserving the keyword, so I feel it doesn't really need to have a FLIP. But when we actually implement the support for native and static fields/methods, I agree a FLIP makes more sense.

SupunS:
BTW, shouldn't this go to the stable-cadence branch? since introducing keywords can break existing codes

runtime/sema/checker.go
SupunS: ah just noticed, maybe rename this to `AllowNativeDeclarations` since it checks both fields and functions. Same for the other one too.

turbolent:
@SupunS I was hoping to use it for the new Capability Controller code, which will be implemented on master.
I guess if we make the keywords soft for now the additions would not be breaking changes?

runtime/sema/checker.go
turbolent: Good catch! Fixed use and improved naming in a409088c56b1adc81f4964047521b3ca6937ee7f

dsainati1:
I believe the change preventing the use of keywords as identifiers is on the Stable Cadence branch anyways, so this would not be breaking either way. That said, these should still be soft keywords, especially given that `view` is now also a soft keyword.

dsainati1 (approved):

SupunS (approved):

turbolent:
@dsainati1 @SupunS How about we make the new rules optional in the parser, and only enable them when parsing code for built-ins and the standard library? That way it would have absolutely no impact on user code

dsainati1:
Sure! If we decide we want to extend these modifiers for user code later that will make things pretty easy.

SupunS:
Yeah, sounds good 👌

turbolent:
@onflow/cadence In [399b75a](https://github.com/onflow/cadence/pull/2119/commits/399b75ac12d3b50377840bed3e1e5c7b64387f0f) I gated the new modifiers behind configuration options, which are disabled by default. That way the changes have no influence on user-code and we can start using the modifiers as discussed in the sync.

dsainati1 (approved):

SupunS (approved):

PR #2121, opened 2022-11-03
Make view a soft keyword

### LINKED ISSUE #2120: Make `view` a soft keyword
### Issue To Be Solved

Function purity annotations were added in #1870 for Stable Cadence, allowing referentially-transparent functions to be annotated as such with the new `view` keyword. 

We currently treat `view` as a hard keyword, which prevents it from being used as an identifier in any context. This is a problem, as the NFT contracts and many other on-chain contracts use it for arguments and variables, which breaks their code.

### Suggested Solution

Make `view` a soft keyword in identifier position. This will require 1 token of lookahead in the parser when a `view` token is encountered - if it's followed by `fun`, then we know we're encountering a view function declaration or expression. Otherwise, it can be treated as an identifier. The list of soft keywords in the Stable Cadence changelog will also need to be updated.

Tasks:
- [x] Extend the parser
- [x] Update documentation and changelogs

dreamsmasher:
Closes #2120

## Description

Modifies the parser to allow the `view` keyword to be used as an identifier. The keyword has been added to the `SoftKeywords` set in `runtime/parser/keyword.go`.
______

### DIFF
--- a/runtime/parser/declaration.go
+++ b/runtime/parser/declaration.go
@@ -62,7 +62,7 @@ func parseDeclarations(p *parser, endTokenType lexer.TokenType) (declarations []
 	}
 }
 
-func parseDeclaration(p *parser, docString string) (ast.Declaration, error) {
+func parseDeclaration(p *parser, docString string) (declaration ast.Declaration, err error) {
 
 	access := ast.AccessNotSpecified
 	var accessPos *ast.Position
@@ -78,7 +78,11 @@ func parseDeclaration(p *parser, docString string) (ast.Declaration, error) {
 
 	defer func() {
 		if noDeclarationFound {
-			p.replayBuffered()
+			bufferErr := p.replayBuffered()
+			if bufferErr != nil && err == nil {
+				declaration = nil
+				err = bufferErr
+			}
 		} else {
 			p.acceptBuffered()
 		}
@@ -151,10 +155,9 @@ func parseDeclaration(p *parser, docString string) (ast.Declaration, error) {
 				}
 				pos := p.current.StartPos
 				accessPos = &pos
-				var err error
 				access, err = parseAccess(p)
 				if err != nil {
-					return nil, err
+					return
 				}
 
 				continue
@@ -162,7 +165,7 @@ func parseDeclaration(p *parser, docString string) (ast.Declaration, error) {
 		}
 
 		noDeclarationFound = true
-		return nil, nil
+		return
 	}
 }
 
--- a/runtime/parser/declaration.go
+++ b/runtime/parser/declaration.go
@@ -70,7 +70,7 @@ func parseDeclaration(p *parser, docString string) (ast.Declaration, error) {
 	purity := ast.FunctionPurityUnspecified
 	var purityPos *ast.Position
 
-	// access modifiers/purity annotations are soft keywords
+	// some access modifiers/purity annotations are soft keywords
 	// in which case, parsing them as semantic tokens might mangle the
 	// parser state before they can parsed as identifiers
 	noDeclarationFound := false
--- a/runtime/parser/declaration.go
+++ b/runtime/parser/declaration.go
@@ -70,6 +70,20 @@ func parseDeclaration(p *parser, docString string) (ast.Declaration, error) {
 	purity := ast.FunctionPurityUnspecified
 	var purityPos *ast.Position
 
+	// access modifiers/purity annotations are soft keywords
+	// in which case, parsing them as semantic tokens might mangle the
+	// parser state before they can parsed as identifiers
+	noDeclarationFound := false
+	p.startBuffering()
+
+	defer func() {
+		if noDeclarationFound {
+			p.replayBuffered()
+		} else {
+			p.acceptBuffered()
+		}
+	}()
+
 	for {
 		p.skipSpaceAndComments()
 
@@ -125,6 +139,7 @@ func parseDeclaration(p *parser, docString string) (ast.Declaration, error) {
 				if purity != ast.FunctionPurityUnspecified {
 					return nil, p.syntaxError("invalid second view modifier")
 				}
+
 				pos := p.current.StartPos
 				purityPos = &pos
 				purity = parsePurityAnnotation(p)
@@ -146,6 +161,7 @@ func parseDeclaration(p *parser, docString string) (ast.Declaration, error) {
 			}
 		}
 
+		noDeclarationFound = true
 		return nil, nil
 	}
 }
--- a/runtime/parser/declaration.go
+++ b/runtime/parser/declaration.go
@@ -62,31 +62,13 @@ func parseDeclarations(p *parser, endTokenType lexer.TokenType) (declarations []
 	}
 }
 
-func parseDeclaration(p *parser, docString string) (declaration ast.Declaration, err error) {
+func parseDeclaration(p *parser, docString string) (ast.Declaration, error) {
 
 	access := ast.AccessNotSpecified
 	var accessPos *ast.Position
 
 	purity := ast.FunctionPurityUnspecified
 	var purityPos *ast.Position
-
-	// some access modifiers/purity annotations are soft keywords
-	// in which case, parsing them as semantic tokens might mangle the
-	// parser state before they can parsed as identifiers
-	noDeclarationFound := false
-	p.startBuffering()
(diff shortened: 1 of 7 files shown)


runtime/parser/expression.go
```diff
@@ -815,15 +815,18 @@ func defineIdentifierExpression() {
 				), nil
 
 			case keywordView:
```
turbolent, line 817: Nice! Needs to be "duplicated" in `parseStatement`

dreamsmasher: ah damn you're right, this bit me last time too

runtime/parser/expression.go
```diff
@@ -815,15 +815,18 @@ func defineIdentifierExpression() {
...
-				if !p.isToken(p.current, lexer.TokenIdentifier, keywordFun) {
-					return nil, p.syntaxError(
-						"expected fun keyword, but got %s",
-						string(p.tokenSource(p.current)),
-					)
+				// if `view` is followed by `fun`, then it denotes a pure function expression
```
dsainati1, line 818:
```suggestion
				// if `view` is followed by `fun`, then it denotes a view function expression
```

dsainati1 (approved):

runtime/parser/declaration.go
```diff
@@ -62,14 +62,32 @@ func parseDeclarations(p *parser, endTokenType lexer.TokenType) (declarations []
...
-func parseDeclaration(p *parser, docString string) (ast.Declaration, error) {
+func parseDeclaration(p *parser, docString string) (declaration ast.Declaration, err error) {
 
 	access := ast.AccessNotSpecified
 	var accessPos *ast.Position
 
 	purity := ast.FunctionPurityUnspecified
 	var purityPos *ast.Position
 
+	// some access modifiers/purity annotations are soft keywords
```
turbolent, line 73: Is there maybe a way we could avoid buffering and replay?

runtime/parser/expression.go
```diff
@@ -815,15 +815,18 @@ func defineIdentifierExpression() {
...
+					return parseFunctionExpression(p, token, ast.FunctionPurityView)
 				}
-				p.nextSemanticToken()
 
-				return parseFunctionExpression(p, token, ast.FunctionPurityView)
+				// otherwise, we treat it as an identifier called "view"
+				return ast.NewIdentifierExpression(
+					p.memoryGauge,
+					p.tokenToIdentifier(token),
+				), nil
```
turbolent, line 828: Duplicate of the default, maybe just remove this duplicated code and move the default case of the switch after the switch

runtime/parser/expression_test.go
```diff
@@ -5889,6 +5889,18 @@ func TestParseCasting(t *testing.T) {
 	)
 }
 
+func testParseIdentifiersWith(t *testing.T, identifiers []string, condition func(*testing.T, string, error)) {
+	for _, name := range identifiers {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
```
turbolent, line 5895: This use of t.Parallel is broken, this will have non-deterministic behaviour (`name` will be bound to some arbitrary value in the test case)

runtime/parser/statement_test.go
```diff
@@ -2637,3 +2638,54 @@ func TestParseReferenceExpressionStatement(t *testing.T) {
...
+		offsetPos := len(name) + offset
+		return ast.Position{
+			Line:   1,
+			Offset: offsetPos,
+			Column: offsetPos,
+		}
+	}
+
+	for keyword := range softKeywords {
+		// haha scoping
```
turbolent, line 2655: Please improve this comment or remove it

runtime/parser/expression_test.go
```diff
@@ -5901,35 +5913,45 @@ func TestParseIdentifiers(t *testing.T) {
...
-func TestParseReservedIdent(t *testing.T) {
+func TestParseHardKeywords(t *testing.T) {
 	t.Parallel()
 
-	for keyword := range hardKeywords {
-		code := fmt.Sprintf(`let %s = 0`, keyword)
-		_, err := testParseProgram(code)
-		upcast := err.(Error)
-		errs := upcast.Errors
+	// seriously why doesn't golang have iterators
```
turbolent, line 5925: Please don't leave rants about the language in the comments, they are not useful

runtime/parser/expression_test.go
```diff
@@ -5901,35 +5913,45 @@ func TestParseIdentifiers(t *testing.T) {
...
+	hardKeywordList := make([]string, len(hardKeywords))
```
turbolent, line 5926: You probably meant to pre-allocate, currently the loop will be twice as long and check lots of empty strings

```suggestion
	hardKeywordList := make([]string, 0, len(hardKeywords))
```

runtime/parser/expression_test.go
```diff
@@ -5901,35 +5913,45 @@ func TestParseIdentifiers(t *testing.T) {
...
-			errs,
+			err.(Error).Errors,
 		)
+	})
+}
+
+func TestParseSoftKeywords(t *testing.T) {
+	t.Parallel()
+
+	softKeywordList := make([]string, len(softKeywords))
```
turbolent, line 5947: Same as above, you probably just want to preallocate the capacity:

```suggestion
	softKeywordList := make([]string, 0, len(softKeywords))
```

runtime/parser/statement_test.go
```diff
@@ -2637,3 +2638,54 @@ func TestParseReferenceExpressionStatement(t *testing.T) {
...
+		name := keyword
```
turbolent, line 2656: Maybe factor the subtest into a helper function

runtime/parser/statement.go
```diff
@@ -96,31 +99,41 @@ func parseStatement(p *parser) (ast.Statement, error) {
 			return parseEmitStatement(p)
 		case keywordView:
 			purityPos := p.current.StartPos
+			// look ahead 1 token for the `fun` keyword
+			p.startBuffering()
```
turbolent, line 103: Given that we're not parsing a complex rule, but just want a couple tokens lookahead, you can simplify to saving and restoring the token stream cursor and current token:

```suggestion
			current := p.current
			cursor := p.tokens.Cursor()
```

This should be a bit more efficient than buffering.

runtime/parser/statement.go
```diff
@@ -96,31 +99,41 @@ func parseStatement(p *parser) (ast.Statement, error) {
...
+
 			p.nextSemanticToken()
-			if !p.isToken(p.current, lexer.TokenIdentifier, keywordFun) {
-				return nil, p.syntaxError(
-					"expected fun keyword, but got %s",
-					string(p.tokenSource(p.current)),
-				)
+			if p.isToken(p.current, lexer.TokenIdentifier, keywordFun) {
+				p.acceptBuffered()
```
turbolent, line 107: Continuing simplification from above:

```suggestion
```

runtime/parser/statement.go
```diff
@@ -96,31 +99,41 @@ func parseStatement(p *parser) (ast.Statement, error) {
...
+
+				return parseFunctionDeclarationOrFunctionExpressionStatement(p, ast.FunctionPurityView, &purityPos)
 			}
-			return parseFunctionDeclarationOrFunctionExpressionStatement(p, ast.FunctionPurityView, &purityPos)
+			// no `fun` :( revert back to previous lexer state and treat it as an identifier
+			err := p.replayBuffered()
+			if err != nil {
+				return nil, err
+			}
```
turbolent, line 115: Continuing simplification from above:

```suggestion
			p.tokens.Revert(cursor)
			p.current = current
```

runtime/parser/statement.go
```diff
@@ -96,31 +99,41 @@ func parseStatement(p *parser) (ast.Statement, error) {
...
 	}
 
-	// If it is not a keyword for a statement,
-	// it might start with a keyword for a declaration
-
-	declaration, err := parseDeclaration(p, "")
-	if err != nil {
-		return nil, err
-	}
+	if !tokenIsIdentifier {
```
turbolent, line 126: Great idea! 👌

turbolent (commented):
Nice! Just one last suggestion regarding the remaining case of lookahead/buffering

SupunS (approved):

PR #2126, opened 2022-11-08
Clean up function types and type annotations

turbolent:
## Description

- Introduce `NewSimpleFunctionType` and use it where possible
- Fix some incorrect function types
- Create "constants" for common type annotations

______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -41,12 +41,13 @@ import (
 
 //
 
-var emptyImpureFunctionType = &sema.FunctionType{
-	Purity: sema.FunctionPurityImpure,
-	ReturnTypeAnnotation: &sema.TypeAnnotation{
+var emptyImpureFunctionType = sema.NewSimpleFunctionType(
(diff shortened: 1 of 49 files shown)


runtime/interpreter/interpreter.go
```diff
@@ -2618,65 +2618,15 @@ func init() {
 		BaseActivation,
 		"DictionaryType",
 		NewUnmeteredHostFunctionValue(
-			func(invocation Invocation) Value {
```
turbolent, line 2621: Just extracted as global functions to make the outer function shorter / easier to reason about

runtime/interpreter/value.go
```diff
@@ -2745,12 +2745,7 @@ func getNumberValueMember(interpreter *Interpreter, v NumberValue, name string,
 					other,
 				)
 			},
-			&sema.FunctionType{
```
turbolent, line 2748: Function types were wrong, the parameters were missing. Reuse the functions type that were already created in the checker

PR #2125, opened 2022-11-07
Add more purity annotations to built-in and standard library function types

turbolent:
Work towards #2084 

## Description

Updating the NFT Storefront contracts identified that some more built-in and standard library functions are impure, but could be view: https://github.com/onflow/nft-storefront/pull/74#discussion_r1015633981

Add explicit purity annotations to more built-in and standard library function types, make functions view when possible.
 

______

- [ ] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #2130, opened 2022-11-09
Fix test matcher

turbolent:
Depends on #2129

## Description

In #2125 I broke the test matcher function, but did not realize at the time, as tests are located in the `cadence-tools` repo, so only realized when updating the test framework to Stable Cadence.

In #2129 I moved some of the tests to Cadence to avoid similar breakage in the future. 
When merged into Stable Cadence, the tests fail, showing the test framework is broken, the creation and invocation of matchers results in erroneous run-time type mismatch errors.

This PR fixes the matcher creation and invocation.

______

### DIFF
--- a/runtime/stdlib/test.go
+++ b/runtime/stdlib/test.go
@@ -54,7 +54,7 @@ const transactionArgsFieldName = "arguments"
 
 const accountAddressFieldName = "address"
 
-const matcherTestFunctionName = "test"
+const matcherTestFieldName = "test"
 
 const addressesFieldName = "addresses"
 
@@ -185,6 +185,17 @@ var matcherType = func() *sema.CompositeType {
 
 	return compositeType
 }()
+
+var matcherTestFieldType = compositeFunctionType(matcherType, matcherTestFieldName)
+
+func compositeFunctionType(parent *sema.CompositeType, funcName string) *sema.FunctionType {
+	testFunc, ok := parent.Members.Get(funcName)
+	if !ok {
+		panic(memberNotFoundError(parent.Identifier, funcName))
+	}
+
+	return getFunctionTypeFromMember(testFunc, funcName)
+}
 
 func interfaceFunctionType(parent *sema.InterfaceType, funcName string) *sema.FunctionType {
 	testFunc, ok := parent.Members.Get(funcName)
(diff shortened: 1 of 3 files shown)


runtime/stdlib/test.go
```diff
@@ -1517,7 +1527,7 @@ func newMatcherWithGenericTestFunction(
 
 			return value
 		},
-		matcherTestFunctionType,
+		matcherTestFieldType,
```
turbolent, line 1530: Fix: Create a function that has the type of the `Matcher.test` field, `((AnyStruct): Bool)`, so the initialization of the field in `Matcher.init` works.

runtime/stdlib/test.go
```diff
@@ -1472,12 +1483,12 @@ func newMatcherWithGenericTestFunction(
...
 
-	staticType, ok := testFunc.StaticType(inter).(interpreter.FunctionStaticType)
-	if !ok {
+	typeParameterPair := invocation.TypeParameterTypes.Oldest()
+	if typeParameterPair == nil {
 		panic(errors.NewUnreachableError())
 	}
 
-	parameters := staticType.Type.Parameters
+	parameterType := typeParameterPair.Value
```
turbolent, line 1491: Fix: Use the type argument for type parameter `T` of `Test.newMatcher<T>`: The wrapper should check if the run-time type of argument passed to the test function (`Matcher.test` has type  `((AnyStruct): Bool)`) matches the type argument `T`

runtime/stdlib/test.go
```diff
@@ -1472,12 +1483,12 @@ func newMatcherWithGenericTestFunction(
 
 	inter := invocation.Interpreter
 
-	staticType, ok := testFunc.StaticType(inter).(interpreter.FunctionStaticType)
-	if !ok {
+	typeParameterPair := invocation.TypeParameterTypes.Oldest()
+	if typeParameterPair == nil {
```
SupunS, line 1487: What happens if the type parameter is not provided? does it still get resolved and would be not-nil?
e.g: `newMatcher(fun (_ value: AnyStruct): Bool { ... }`

SupunS (approved):

runtime/stdlib/test.go
turbolent: Correct! 

For example, here's a test case for it https://github.com/onflow/cadence/pull/2130/files#diff-74ad93a1435e307865ff8474567598f410879038788126ec263e27c9244edf98R152-R154 (already existed, was just moved from the test framework to the Cadence repo in #2129)

The invocation gets the types from the elaboration, we keep them there for each invocation, and type parameters bindings are inferred from uses (here, the parameter of the test function expression passed to `newMatcher`).

PR #2136, opened 2022-11-15
Fix parsing of array and dictionary expressions

turbolent:
Related to #1119

## Description

#2115 fixed parsing of whitespace in expressions.

Analyzing all Mainnet contracts revealed that parsing of array and dictionary expressions broke.

Fix parsing by consuming whitespace again.

Also, add tool that can be used to detect such regressions. It would be nice to run this in CI at some point.

______


dsainati1 (approved):

tools/compare-parsing.py
```diff
@@ -0,0 +1,47 @@
+# compare-parsing.py parses all programs in a CSV files with header location,code
```
SupunS, line 1: Would it be possible to migrate this logic to Go?
I'm a bit worried about using multiple languages in a single repo, as it would become harder to maintain. i.e: Need expertise in multiple languages.
(Code base is mainly Go, and there are some minor usages of Typescript etc. as well. Having another language would add one more to the technology stack)

turbolent: Good point! Ported the tool to Go in [d6bc6fa](https://github.com/onflow/cadence/pull/2136/commits/d6bc6fad287e55be5b5d9e215b2975ffc45ede6e)

tools/compare-parsing/main.go
```diff
@@ -0,0 +1,116 @@
+package main
+
+// compare-parsing.py parses all programs in a CSV files with header location,code
```
SupunS, line 3:
```suggestion
// Parses all programs in a CSV file with header location,code
```

SupunS (approved):
Thanks! LGTM!

PR #2138, opened 2022-11-15
Improve error messages for values/types that cannot be imported or exported

turbolent:
## Description

Some values cannot be exported, like `Block`, and some values cannot be imported, like functions.

Report errors for exports of non-exportable values, and errors for imports of non-importable values as user errors, instead of implementation errors.

Also, simplify the export of composite and simple composite values, and naming.

______

### DIFF
--- a/runtime/convertValues.go
+++ b/runtime/convertValues.go
@@ -196,7 +196,7 @@ func exportValueWithInterpreter(
 			seenReferences,
 		)
 	case *interpreter.SimpleCompositeValue:
-		return exportSimpleCompositeValue(
+		return exportCompositeValue(
 			v,
 			inter,
 			locationRange,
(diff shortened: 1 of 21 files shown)


SupunS (approved):

runtime/convertValues.go
```diff
@@ -361,10 +371,23 @@ func exportCompositeValue(
...
+				computeField := v.ComputeField
+
+				if fieldValue == nil && computeField != nil {
+					fieldValue = computeField(fieldName, inter, locationRange)
+				}
+
+			case *interpreter.CompositeValue:
+				fieldValue = v.GetField(inter, locationRange, fieldName)
+				if fieldValue == nil && v.ComputedFields != nil {
+					if computedField, ok := v.ComputedFields[fieldName]; ok {
```
dsainati1, line 388: What's going on here with this change? Seems like more code is duplicated now?

turbolent: No, this is the result of deduplicating `exportSimpleCompositeValue` below, which was identical to this function, `exportCompositeValue`, except for the logic to get the value's fields.

dsainati1 (approved):

PR #2140, opened 2022-11-16
Implement FLIP #43 (fn type syntax change)

### LINKED ISSUE #2094: Change syntax for function types
### Issue To Be Solved

Currently, function types are printed with confusing syntax. For instance, the type of the function

```cadence
fun foo(n: Int, s: String): Bool
```

is rendered as

```cadence
foo: ((Int, String): Bool)
```

This inconsistency is more apparent when writing higher-order functions:

```cadence
fun forEach(xs: [Int], f: ((Int): Bool)): Void
```

which harms readability.

### Suggested Solution

Function types should use the `fun` keyword so their types are closer to the way they're defined. For the examples above, their types should be rendered as

```cadence
foo: fun(Int, String): Bool
forEach: fun([Int], fun(Int): Bool): Void
```

To avoid breaking existing contracts, this could be implemented as a non-breaking change by allowing the old syntax but updating the parser to also accept the `fun` keyword as a prefix. The stringification of function types should be changed to include the `fun` keyword. 

[See FLIP](https://github.com/onflow/flips/pull/43)

dreamsmasher:
Closes #2094, [FLIP #43](https://github.com/onflow/flips/pull/43)

## Description

Changes function types to make use of the `fun` keyword. For example:

```cadence
fun foo(x: Int8): Int16 {...}

// old syntax
let f: ((Int8): Int16) = foo

// new syntax
let g: fun(Int8): Int16 = foo
```

The old syntax is no longer supported, and pretty-printing will now make use of the new syntax. This PR also extends the parser to allow parenthesized types: for a type `T`, `(T)` will simply parse the inner type. This was previously restricted only to function types, but users can now parenthesize anything. ٩(◕‿◕｡)۶

______


runtime/ast/type.go
```diff
@@ -462,19 +462,28 @@ func (t *FunctionType) String() string {
...
+
+	if t.PurityAnnotation != FunctionPurityUnspecified {
+		result = append(
+			result,
+			prettier.Text(t.PurityAnnotation.Keyword()),
+			prettier.Space,
+		)
 	}
 
+	result = append(result, functionTypeKeywordDoc)
```
turbolent, line 485: It would be good to add a space after the keyword, like for all other keywords

turbolent:
Please also add a label. For documentation, we probably do not want to switch to the new syntax yet once it is available. Could you please create a follow-up issue to update the documentation later?

runtime/parser/expression.go
```diff
@@ -816,20 +816,20 @@ func defineIdentifierExpression() {
...
-						p.current = current
-					}
+				cursor := p.tokens.Cursor()
+
+				p.skipSpaceAndComments()
+
+				if p.isToken(p.current, lexer.TokenIdentifier, KeywordFun) {
+					// skip the `fun` keyword
+					p.nextSemanticToken()
+					return parseFunctionExpression(p, token, ast.FunctionPurityView)
```
dreamsmasher, line 827: noticed a bug when resolving the merge conflict - by only checking for a space after the `view` keyword, then expressions like

```cadence
let x = view /* UwU */ fun() {}
```

fails to parse.

runtime/parser/type.go
```diff
@@ -148,39 +148,9 @@ func init() {
...
-					right,
-					token.StartPos,
-				), nil
-
-			default:
-				return parseNominalTypeRemainder(p, token)
-			}
-		},
-	)
+	defineOldSyntaxFunctionType()
```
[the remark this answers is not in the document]
dreamsmasher: ah ok, like allow the new syntax in `master` then deprecate it for stable cadence? this PR does change the pretty-printing of function types too, i'm worried that might be a breaking change. thoughts?

runtime/parser/expression.go
turbolent: Good catch! Could you please open this fix as a separate PR?

runtime/parser/type.go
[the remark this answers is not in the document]
turbolent: The easiest is to just target Stable Cadence and remove support for the old syntax.

turbolent:
@dreamsmasher Still missing documentation updates

turbolent (approved):

turbolent:
@SupunS @dsainati1 could you please have a look so we can get this merged? Thank you

SupunS (approved):

Issue #2050, opened 2022-10-11
Improve error messages


turbolent:
### Issue To Be Solved

Error messages are often sparse and unhelpful.

### Suggested Solution

- Extend error messages with more detail
- Provide example errors and explain them
- Provide suggestions for how to fix the error
- Link to documentation
- Link to page with more detail for given error

High priority errors:
- `InvalidTransactionAuthorizerCountError`
- `ConformanceError`
- *Your unintelligible error here*

Inspiration:
- https://elm-lang.org/news/compiler-errors-for-humans
- https://typescript.tv/errors/

SeanRobb:
This would be tremendous!

Frequently what part of the stack has the error lies and options for a solution are difficult to find.

Adding more detailed error messages with links to documentation with even more detail about what causes the error and potential solutions would benefit the whole ecosystem.

PR #2173, opened 2022-12-01
Suggest lexically closest member in cases of NotDeclaredMemberError

dsainati1:

When users attempt to access a member (field or function) on a type, if that member does not exist, the checker will raise an error. This adds some additional info to the error suggesting a different field they may have meant, which is particularly helpful in cases of typos or spelling differences between different languages/dialects. 

______


go.mod
```diff
@@ -19,13 +19,15 @@ require (
...
-	golang.org/x/text v0.3.7
-	golang.org/x/tools v0.1.11
+	golang.org/x/crypto v0.1.0
+	golang.org/x/mod v0.6.0
+	golang.org/x/text v0.4.0
+	golang.org/x/tools v0.2.0
 	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1
 )
 
+require github.com/texttheater/golang-levenshtein/levenshtein v0.0.0-20200805054039-cae8b0eaed6c
```
dsainati1, line 29: Is there a specific thing I should be looking for in libraries like this to check whether it's okay for us to rely on them?

turbolent (approved):
Nice!

Maybe this should only be done in "development mode", e.g. for the language server?

runtime/sema/errors.go
```diff
@@ -909,9 +911,34 @@ func (e *NotDeclaredMemberError) SecondaryError() string {
...
+// findClosestMember searches the names of the members on the accessed type,
+// and finds the name with the smallest edit distance from the member the user
+// tried to access. In cases of typos, this should provide a helpful hint.
+
+func (e *NotDeclaredMemberError) findClosestMember() string {
+	members := maps.Keys(e.Type.GetMembers())
+	closestDistance := len(e.Name)
+	var closestMember string
+	// TODO: when attachments are merged to master, discount any members prefixed by the attachment prefix
+	for _, member := range members {
```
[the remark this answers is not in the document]
dsainati1: Oh good point I was thinking of their runtime implementation, rather than their static implementation. This should actually work fine.

runtime/tests/checker/member_test.go
```diff
@@ -296,3 +296,90 @@ func TestCheckFunctionTypeReceiverType(t *testing.T) {
...
+
+            let test: Test = Test()
+            let x = test.foop()
+        `)
+
+		errs := RequireCheckerErrors(t, err, 1)
+
+		assert.IsType(t, &sema.NotDeclaredMemberError{}, errs[0])
+		memberErr := errs[0].(*sema.NotDeclaredMemberError)
+		assert.Equal(t, "unknown member", memberErr.SecondaryError())
```
turbolent, line 363:
```suggestion
		var memberErr *sema.NotDeclaredMemberError
		require.ErrorAs(t, errs[0], &memberErr)
		assert.Equal(t, "unknown member", memberErr.SecondaryError())
```

runtime/tests/checker/member_test.go
```diff
@@ -296,3 +296,90 @@ func TestCheckFunctionTypeReceiverType(t *testing.T) {
...
```
turbolent, line 383:
```suggestion
		var memberErr *sema.NotDeclaredMemberError
		require.ErrorAs(t, errs[0], &memberErr)
		assert.Equal(t, "unknown member", memberErr.SecondaryError())
```

dsainati1:
> Maybe this should only be done in "development mode", e.g. for the language server?

Do we currently have a flag like this for dev mode vs. production, or something similar? Either way, I don't think this is going to cause much, if any, performance overhead because the additional computation is only done when the error is printed, and most contracts and transactions being successfully executed on the network are already going to typecheck.

SupunS (approved):
Nice idea!

+1 to have this is in 'developer mode' only. Maybe add this to the linter?

turbolent:
@dsainati1 We do not have a configuration for such a mode. Even though it does not add much overhead, but it is theoretically unnecessary, and it adds up with all transactions. 

Similar, we recently added a "error-short circuiting" mode, which stops reporting more errors after the first (`ErrorShortCircuitingEnabled`), as on-chain only differentiating between an error or success matters, and "as many errors as possible" only matters  off-chain / for developers. 

Maybe we can have a `suggestionsEnabled` configuration option? We could also gate the existing suggestion before the one you added, which suggests to use optional-chaining when the member does not exist, but does on the unwrapped type.

turbolent (approved):

PR #2174, opened 2022-12-02
Add positions to numerical runtime errors

dsainati1:

`OverflowError`, `UnderflowError` and `DivisionByZeroError` were being reported with no information in the error about the location of the erroring operation. This adds `LocationRange`s to their structs, and refactors a large amount of code to pass the appropriate locations through to the error sites. 
______


runtime/interpreter/numerical_value.go
```diff
@@ -0,0 +1,11236 @@
...
+	}
+
+	valueGetter := func() int8 {
+		return int8(v % o)
+	}
+
+	return NewInt8Value(interpreter, valueGetter)
+}
+
+func (v Int8Value) Mul(interpreter *Interpreter, other NumberValue, locationRange LocationRange) NumberValue {
```
dsainati1, line 1005: While modifying all these functions, I noticed that they are all fairly redundant in terms of their content and implementation; mostly we just compare values against type-dependent maximums and minimums, and overflow/underflow as necessary, performing the arithmetic operation otherwise. It might be nice to rewrite these implementations to use generic functions. Does this seem worth it or would we rather keep these as they are for performance?

turbolent: Yes, if generics can be used to reduce code duplication, while not regressing on performance too much, we should try it 👍

turbolent:
Nice, thank you for adding this!

I understand the desire to split `value.go`: it is massive and we should definitely split it into smaller files. However, currently this will cause lots of conflicts when we merge such file refactors into the Stable Cadence feature branch. Could you please move the changes back for now? Once Stable Cadence is released, we can address such technical debt more easily.

runtime/interpreter/numerical_value.go
dsainati1: Nice, I'll look into a separate PR for this next week.

turbolent (approved):
👍

SupunS (approved):

PR #2180, opened 2022-12-06
Add get or load program to the interface

janezpodhostnik:
Related to https://github.com/onflow/flow-go/issues/3015

## Description

First step of changing the dangerous `GetProgram`, `SetProgram` to a `GetOrLoadProgram` that takes a closure doing the actual loading.

## Some concerns

1. I was not able to remove `SetProgram` due to some cases where set is used without first calling get (which I don't fully understand, but I think it relates to the non address location programs)

2. I duplicated some code from `parseAndCheckProgram`. This wouldn't have been needed if `runtimeInterface.SetProgram` could be removed from the function. But that is not possible due to the previous point.
______

### DIFF
--- a/runtime/contract_update_test.go
+++ b/runtime/contract_update_test.go
@@ -45,6 +45,18 @@ func TestContractUpdateWithDependencies(t *testing.T) {
 
 	var checkGetSetProgram, getProgramCalled, setProgramCalled bool
 
+	setProgram := func(location Location, program *interpreter.Program) error {
+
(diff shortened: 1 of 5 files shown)


turbolent:
@janezpodhostnik Great work, thank you again for opening this PR and making the fixing of this problem a priority 🙏 

I fixed the outstanding case of SetProgram calls without GetProgram calls in #2249, and then used this PR to open #2254. 

Those two PRs sort of replace this PR

janezpodhostnik:
Ok, lets close this one then.

PR #2181, opened 2022-12-06
Genericize NumberValue code to remove redundancies

dsainati1:
As part of earlier improvements to `NumberValue`s, thought this could use some cleanup

______


dsainati1:
-3000 LoC 😄

runtime/interpreter/value.go
```diff
@@ -2695,247 +2692,1468 @@ type NumberValue interface {
...
 
-// BigNumberValue is a number value with an integer value outside the range of int64
-type BigNumberValue interface {
+type BoundedValue[T Numerical] interface {
 	NumberValue
-	ByteLength() int
-	ToBigInt(memoryGauge common.MemoryGauge) *big.Int
+	MaxValue() T
+	Underlying() T
+	Constructor(common.MemoryGauge, func() T) NumberValue
```
turbolent, line 2715: `Constructor` sounds like this function returns a constructor (function). Maybe rename to a verb, e.g. `Construct`.

This and the other two functions `MinValue` and `MaxValue` are really don't really need the value as a receiver. Maybe there is a more efficient way to dispatch just on the type instead of the value, i.e. have something like a static function, but I can't think of a way to do that in Go 🤔

turbolent (commented):
Great work! This cleans up the code a lot 👌 

Thank you for refactoring in separate PRs, that helps with the review of this large PR. 

Have you had a look at the performance impact of the changes?

dsainati1:
> Have you had a look at the performance impact of the changes?

Yeah I looked at the benchmark results on this PR and they seem fine, but let me know if there's any other tests you think I should run. 

I also had another idea to improve this further so I'll try implementing that as well here.

runtime/interpreter/value.go
```diff
@@ -2695,6 +2692,57 @@ type NumberValue interface {
...
+	NumberValue
+	MaxValue() T
+	MinValue() T
+	Underlying() T
+	Constructor(common.MemoryGauge, func() T) NumberValue
+}
+
+func negate[T constraints.Signed](interpreter *Interpreter, v BoundedSignedValue[T], locationRange LocationRange) NumberValue {
+	// INT32-C
+	underlying := v.Underlying()
```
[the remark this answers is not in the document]
dsainati1: Unfortunately no; despite the fact that, say, a `BoundedSignedValue[int32]` is really just an `int32` under the hood, Go doesn't know that statically so we can't just cast like this.

turbolent:
@dsainati1 The CI benchmarks are pretty unreliable, could you please try benchmarking locally? Might also be helpful to add a few smaller arithmetic benchmarks that exercise the refactored code more.

dsainati1:
I ran an arithmetic benchmark for these changes (added in a new commit in this PR). The results for the pre-change code:

```
BenchmarkValueArithmetic/Int8Value-10         	 50.97 ns/op	       0 B/op	       0 allocs/op
BenchmarkValueArithmetic/Int16Value-10        	 68.20 ns/op	       4 B/op	       2 allocs/op
BenchmarkValueArithmetic/Int32Value-10        	 81.35 ns/op	      12 B/op	       3 allocs/op
BenchmarkValueArithmetic/Int64Value-10        	 91.12 ns/op	      40 B/op	       5 allocs/op
BenchmarkValueArithmetic/Fix64Value-10        	 456.6 ns/op	     336 B/op	      22 allocs/op
BenchmarkValueArithmetic/Int128Value-10       	 350.6 ns/op	     336 B/op	      14 allocs/op
BenchmarkValueArithmetic/Int256Value-10       	 62.1 ns/op	     336 B/op	      14 allocs/op
BenchmarkValueArithmetic/UInt8Value-10        	 48.33 ns/op	       0 B/op	       0 allocs/op
BenchmarkValueArithmetic/UInt16Value-10       	 59.01 ns/op	       2 B/op	       1 allocs/op
BenchmarkValueArithmetic/UInt32Value-10       	 68.61 ns/op	       8 B/op	       2 allocs/op
BenchmarkValueArithmetic/UInt64Value-10       	 78.14 ns/op	      32 B/op	       4 allocs/op
BenchmarkValueArithmetic/UFix64Value-10       	 373.2 ns/op	     328 B/op	      21 allocs/op
BenchmarkValueArithmetic/UInt128Value-10      	 317.4 ns/op	     296 B/op	      12 allocs/op
BenchmarkValueArithmetic/UInt256Value-10      	 321.0 ns/op	     296 B/op	      12 allocs/op
BenchmarkValueArithmetic/Word8Value-10        	 44.46 ns/op	       0 B/op	       0 allocs/op
BenchmarkValueArithmetic/Word16Value-10       	 57.94 ns/op	       2 B/op	       1 allocs/op
BenchmarkValueArithmetic/Word32Value-10         67.36 ns/op	       8 B/op	       2 allocs/op
BenchmarkValueArithmetic/Word64Value-10         77.11 ns/op	      32 B/op	       4 allocs/op
```

and the post-change code:

```
BenchmarkValueArithmetic/Int8Value-10        	 221.4 ns/op	     304 B/op	       6 allocs/op
BenchmarkValueArithmetic/Int16Value-10        	 234.1 ns/op	     308 B/op	       8 allocs/op
BenchmarkValueArithmetic/Int32Value-10             248.0 ns/op	     320 B/op	      10 allocs/op
BenchmarkValueArithmetic/Int64Value-10        	 279.0 ns/op	     424 B/op	      12 allocs/op
BenchmarkValueArithmetic/Fix64Value-10        	 564.6 ns/op	     608 B/op	      27 allocs/op
BenchmarkValueArithmetic/Int128Value-10       	 544.3 ns/op	     736 B/op	      20 allocs/op
BenchmarkValueArithmetic/Int256Value-10       	 545.8 ns/op	     736 B/op	      20 allocs/op
BenchmarkValueArithmetic/UInt8Value-10        	 185.8 ns/op	     280 B/op	       5 allocs/op
BenchmarkValueArithmetic/UInt16Value-10       	 194.3 ns/op	     282 B/op	       6 allocs/op
BenchmarkValueArithmetic/UInt32Value-10       	 209.8 ns/op	     304 B/op	       7 allocs/op
BenchmarkValueArithmetic/UInt64Value-10       	 240.0 ns/op	     368 B/op	       9 allocs/op
BenchmarkValueArithmetic/UFix64Value-10       	 481.7 ns/op	     568 B/op	      24 allocs/op
BenchmarkValueArithmetic/UInt128Value-10      	 486.4 ns/op	     664 B/op	      17 allocs/op
BenchmarkValueArithmetic/UInt256Value-10      	 492.8 ns/op	     664 B/op	      17 allocs/op
BenchmarkValueArithmetic/Word8Value-10        	 190.6 ns/op	     280 B/op	       5 allocs/op
BenchmarkValueArithmetic/Word16Value-10       	 201.8 ns/op	     282 B/op	       6 allocs/op
BenchmarkValueArithmetic/Word32Value-10         218.5 ns/op	     304 B/op	       7 allocs/op
BenchmarkValueArithmetic/Word64Value-10         244.3 ns/op	     368 B/op	       9 allocs/op
```

This is a fairly significant increase actually, in terms of raw percentage. It's not a big change relative to the amount of time we spend doing other things in the code though. Not sure what to do here.

turbolent:
Hmm, yeah, the numbers don't look good. I wonder why there's more memory being allocated now, will investigate

turbolent:
@dsainati1 could you please push up the benchmark, it looks like there was no new commit yet

dsainati1:
Ok I added the benchmark in b3574e8ae8e8458d2f7982093e1504dcf3627499

runtime/interpreter/number_test.go
```diff
@@ -166,3 +167,89 @@ func TestOverEstimateFixedPointStringLength(t *testing.T) {
...
+
+	isNegatable := func(v NumberValue) bool {
+		switch v.(type) {
+		case IntValue, Int8Value, Int16Value,
+			Int32Value, Int64Value, Int128Value,
+			Int256Value, Fix64Value:
+			return true
+		}
+		return false
+	}
```
dreamsmasher, line 183: i think you can replace this with a check if v is a subtype of `SignedNumberValue`

dreamsmasher (approved):

runtime/interpreter/number_test.go
```diff
@@ -166,3 +167,89 @@ func TestOverEstimateFixedPointStringLength(t *testing.T) {
...
+
+	isSaturating := func(v NumberValue) bool {
+		switch v.(type) {
+		case Word8Value, Word16Value, Word32Value, Word64Value:
+			return false
+		}
+		return false
```
turbolent, line 190: Function always returns false

runtime/interpreter/number_test.go
dsainati1: What type is this? I don't think this actually exists in the interpreter.

turbolent:
Another idea: What if we keep the generic code, but try to "manually" monomorphize by using an approach similar to #2278, i.e. parse the Go code using [`go/parser`](https://pkg.go.dev/go/parser), then "instantiate" for specific type arguments, and print out the resulting Go code? This should also result in the same Go code that we have manually written, but won't have any generics overhead

j1010001:
this is causing performance issues (slowing down execution)

PR #2183, opened 2022-12-06
Initialize the Crypto contract lazily

turbolent:
## Description

Instead of eagerly initializing the `Crypto` contract, initialize it lazily.

This is especially useful for development: Currently, when modifying the parser or checker, the eager initialization of the contract exercises the parser and checker before the test case runs.

______

### DIFF
--- a/runtime/cmd/cmd.go
+++ b/runtime/cmd/cmd.go
@@ -99,9 +99,10 @@ func DefaultCheckerConfig(
 			importedLocation common.Location,
 			_ ast.Range,
 		) (sema.Import, error) {
-			if importedLocation == stdlib.CryptoChecker.Location {
+			cryptoChecker := stdlib.CryptoChecker()
+			if importedLocation == cryptoChecker.Location {
 				return sema.ElaborationImport{
-					Elaboration: stdlib.CryptoChecker.Elaboration,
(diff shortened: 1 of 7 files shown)


runtime/cmd/cmd.go
```diff
@@ -99,9 +99,10 @@ func DefaultCheckerConfig(
 			importedLocation common.Location,
 			_ ast.Range,
 		) (sema.Import, error) {
-			if importedLocation == stdlib.CryptoChecker.Location {
+			cryptoChecker := stdlib.CryptoChecker()
+			if importedLocation == cryptoChecker.Location {
```
SupunS, line 103: Here and in other places, shouldn't `cryptoChecker := stdlib.CryptoChecker()` be moved inside the if condition?
Otherwise, the crypto contract would be loaded even if `importedLocation != cryptoChecker.Location`

turbolent: Great point! Refactored in 1d54a7106c0769a67459b7960f17b1485bdd8f58 by adding a constant for the location of the Crypto contract and using it for the check, then only creating the checker if necessary

dsainati1 (approved):
oh amazing thanks for this 🙏

SupunS (approved):

PR #2185, opened 2022-12-07
Pretty print interpreter errors

dsainati1:

We included `LocationRange` info in some interpreter errors, but didn't actually print that information. This should ensure that the location is included. 

______

### DIFF
--- a/runtime/error_test.go
+++ b/runtime/error_test.go
@@ -140,6 +140,44 @@ func TestRuntimeError(t *testing.T) {
 		)
 	})
 
+	t.Run("execution error with position", func(t *testing.T) {
+
+		t.Parallel()
+
+		runtime := newTestInterpreterRuntime()
+
+		script := []byte(`
+			pub fun main() {
+				let x: AnyStruct? = nil
+				let y = x!
+			}
+        `)
+
+		runtimeInterface := &testRuntimeInterface{}
+
+		location := common.ScriptLocation{0x1}
+
+		_, err := runtime.ExecuteScript(
+			Script{
+				Source: script,
+			},
+			Context{
+				Interface: runtimeInterface,
+				Location:  location,
+			},
+		)
+		require.EqualError(
+			t,
+			err,
+			"Execution failed:\n"+
+				"error: unexpectedly found nil while forcing an Optional value\n"+
+				" --> 0100000000000000000000000000000000000000000000000000000000000000:4:12\n"+
+				"  |\n"+
+				"4 | 				let y = x!\n"+
+				"  | 				        ^^\n",
+		)
+	})
+
 	t.Run("parse error in import", func(t *testing.T) {
 
 		t.Parallel()
--- a/runtime/interpreter/errors.go
+++ b/runtime/interpreter/errors.go
@@ -25,6 +25,7 @@ import (
 	"github.com/onflow/cadence/runtime/ast"
 	"github.com/onflow/cadence/runtime/common"
 	"github.com/onflow/cadence/runtime/errors"
+	"github.com/onflow/cadence/runtime/pretty"
 	"github.com/onflow/cadence/runtime/sema"
 )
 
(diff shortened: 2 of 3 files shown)


runtime/interpreter/errors.go
```diff
@@ -60,7 +61,14 @@ func (e Error) Unwrap() error {
...
 func (e Error) Error() string {
-	return e.Err.Error()
+	var sb strings.Builder
+	sb.WriteString("Execution failed:\n")
+	printErr := pretty.NewErrorPrettyPrinter(&sb, false).
+		PrettyPrintError(e.Err, e.Location, map[common.Location][]byte{})
+	if printErr != nil {
+		panic(printErr)
+	}
+	return sb.String()
```
SupunS, line 71: Would this not duplicate the error message/location info? 
Printing the root error with `PrettyPrintError` recursively visits child errors and prints them with locations. There, `error.Error()` is used to get the 'message' only https://github.com/onflow/cadence/blob/65fd955ef61ee1cf4653acaa4f59837ae69993f7/runtime/pretty/print.go#L221 and the location info is printed separately. Adding it here might duplicate those in the output.

Maybe add a test case to validate the output

dsainati1: No; we don't include the location when we print interpreter error messages right now. 

You can see this if you try to run 

```
pub fun main() {
     let x: AnyStruct? = nil
    let y = x!
}
```

on the Flow Playground. This is an interpreter error with a `LocationRange`, but no location information is actually printed. 

I've added a test for this though.

SupunS: not sure if it's just a playground thingy.

All runtime errors are returned to FVM with a wrapped `runtime.Error`. e.g:
https://github.com/onflow/cadence/blob/65fd955ef61ee1cf4653acaa4f59837ae69993f7/runtime/script_executor.go#L190

and the `runtime.Error.Error()` already use PrettyPrinting: https://github.com/onflow/cadence/blob/65fd955ef61ee1cf4653acaa4f59837ae69993f7/runtime/errors.go#L53-L62

turbolent: So far it has been up to whatever uses the interpreter, e.g. the runtime, REPL, etc. to take care of error printing.

You can see that location information for interpreter errors is e.g. printed for interpreter errors:

```
$ go run ./runtime/cmd/main
Welcome to Cadence v0.29.0!
Type '.help' for assistance.

1>
2> let x: Int? = nil
3> x!
error: unexpectedly found nil while forcing an Optional value
 --> REPL:3:0
  |
3 | x!
  | ^^
```

I think it might be OK to improve the implementation if `interpreter.Error.Error()` by using the error pretty-printer by default, so that if someone calls the function they get a good default behaviour.
Like @SupunS mentioned though, we have to make sure that this change doe not end up printing the error multiple times.

dsainati1: Added an additional test for this in fb656ae572283f5c78e82d01eb2781fa073d8707

It looks like given that all the tests in this file continued working as expected; this just kind of works out of the box.

SupunS: Oh, I see why it's not duplicating: for `errors.ParentError`s the message is never printed via the pretty printer (only for leaf-errors the message gets printed). So this would only be used if someone does` interpreter.Error.Error()` 👍

SupunS: Thanks for adding a test!

SupunS (approved):

runtime/interpreter/errors.go
dsainati1: > So this would only be used if someone does `interpreter.Error.Error()`

What does the playground do to print errors?

runtime/error_test.go
```diff
@@ -140,6 +140,44 @@ func TestRuntimeError(t *testing.T) {
...
+		_, err := runtime.ExecuteScript(
+			Script{
+				Source: script,
+			},
+			Context{
+				Interface: runtimeInterface,
+				Location:  location,
+			},
+		)
+		require.EqualError(
```
turbolent, line 169: Was the output different before improving `interpreter.Error.Error()`?

runtime/interpreter/errors_test.go
```diff
@@ -40,3 +42,22 @@ func TestOverwriteError_Error(t *testing.T) {
 		"failed to save object: path /storage/test in account 0x0000000000000001 already stores an object",
 	)
 }
+
+func TestErrorOutputIncludesLocationRage(t *testing.T) {
+
```
turbolent, line 47:
```suggestion
	t.Parallel()
```

turbolent (approved):

runtime/error_test.go
dsainati1: Yep, previously this just printed the `unexpectedly found nil while forcing an Optional value` message without including any of the location/position data

PR #2194, opened 2022-12-13
improve wording and classification of some interpreter errors

dsainati1:

Improves wording and description of some user errors, and reclassify some errors as internal errors since they should not be possible to see in practice. 

______


runtime/interpreter/errors.go
```diff
@@ -643,9 +646,9 @@ type MemberAccessTypeError struct {
 	LocationRange
 }
 
-var _ errors.UserError = MemberAccessTypeError{}
+var _ errors.InternalError = MemberAccessTypeError{}
 
-func (MemberAccessTypeError) IsUserError() {}
+func (MemberAccessTypeError) IsInternalError() {}
```
dsainati1, line 651: This cannot occur at runtime outside of an implementation mistake afaik.

runtime/interpreter/errors.go
```diff
@@ -662,9 +665,9 @@ type ValueTransferTypeError struct {
 	LocationRange
 }
 
-var _ errors.UserError = ValueTransferTypeError{}
+var _ errors.InternalError = ValueTransferTypeError{}
 
-func (ValueTransferTypeError) IsUserError() {}
+func (ValueTransferTypeError) IsInternalError() {}
```
dsainati1, line 670: This cannot occur at runtime outside of an implementation mistake afaik.

runtime/interpreter/errors.go
```diff
@@ -250,6 +250,10 @@ func (e DereferenceError) Error() string {
 	return "dereference failed"
 }
 
+func (e DereferenceError) SecondaryError() string {
+	return "the value being referenced has been destroyed or moved"
```
turbolent, line 254: This error is also reported when a storage reference is dereferenced and the value stored at the path has the wrong type, or there is no value stored at the path

runtime/interpreter/errors.go
```diff
@@ -685,9 +688,9 @@ type ResourceConstructionError struct {
 	LocationRange
 }
 
-var _ errors.UserError = ResourceConstructionError{}
+var _ errors.InternalError = ResourceConstructionError{}
 
-func (ResourceConstructionError) IsUserError() {}
+func (ResourceConstructionError) IsInternalError() {}
```
turbolent, line 693: This is not an internal error, but is reported e.g. when a user tries to call the resource constructor function outside of the declaration location.

Maybe add a comment to the error type declaration instead which clarifies this.

dsainati1: This is a static error isn't it? 

IIRC the runtime version is more of a defensive check against any potential implementation errors we may have made in the static check. The only test that we have for this error requires creating a function value in Go that cannot be constructed in Cadence; if that's the only way we have to trigger an error it seems to me that it shouldn't be considered a user error since there is no way for a user to cause it.

dsainati1:
cc @SupunS can you take a look at this?

runtime/interpreter/errors.go
SupunS: yeah, I think this is just the defensive check for the interpreter. Ideally should be prevented at the checker, so this is OK to be an internal error.

SupunS (approved):

runtime/interpreter/errors.go
```diff
@@ -239,6 +239,9 @@ func (e RedeclarationError) Error() string {
 // DereferenceError
 
 type DereferenceError struct {
+	Cause        string
+	ExpectedType *sema.Type
+	ActualType   *sema.Type
```
turbolent, line 244: These fields can be non-pointers (`sema.Type` is an interface)

runtime/interpreter/value.go
```diff
@@ -17180,7 +17198,9 @@ func (v *EphemeralReferenceValue) MeteredString(memoryGauge common.MemoryGauge,
 func (v *EphemeralReferenceValue) StaticType(inter *Interpreter) StaticType {
 	referencedValue := v.ReferencedValue(inter, EmptyLocationRange)
 	if referencedValue == nil {
-		panic(DereferenceError{})
+		panic(DereferenceError{
+			Cause: "the value being referenced has been destroyed or moved",
```
turbolent, line 17202: In a follow-up PR we can maybe pass along the location range, but that requires passing it to `Value.StaticType()`

turbolent (approved):
Nice! Just one comment regarding the pointer-typed fields

runtime/interpreter/errors.go
dsainati1: I made these pointer types because these fields are not always included in the error struct; i.e. they can be `nil`. If these aren't pointer types won't we run afoul of the dynamic-`nil` type behavior that has caused issues in the past?

turbolent: No, not in this case. We do not have typed-nil values for `sema.Type`, and also use `Cause` to decide if `ExpectedType` and `ActualType` should be used / are set, or if not; instead of checking if those fields are  `nil`

runtime/interpreter/errors.go
```diff
@@ -586,7 +605,7 @@ var _ errors.UserError = UUIDUnavailableError{}
 func (UUIDUnavailableError) IsUserError() {}
 
 func (e UUIDUnavailableError) Error() string {
-	return "cannot get UUID: unavailable"
+	return "cannot get UUID: UUID acces is unavailable in this configuration of Cadence"
```
dreamsmasher, line 608:
```suggestion
	return "cannot get UUID: UUID access is unavailable in this configuration of Cadence"
```

dreamsmasher (approved):

PR #2197, opened 2022-12-15
Allow errors to pretty-print excerpts over multiple lines

dsainati1:

Multi-line errors print confusingly, as if the error occurs on the last line, only the first is shown. This changes pretty printing to print all the lines of the error. 
______


dsainati1:
cc @turbolent @SupunS can you take a look at this?

SupunS (approved):
Nice!

dsainati1:
@turbolent bump

runtime/pretty/print.go
```diff
@@ -293,19 +287,35 @@ func (p ErrorPrettyPrinter) writeCodeExcerpts(
...
-			line := lines[excerpt.startPos.Line-1]
-			if len(line) > maxLineLength {
-				p.writeString(line[:maxLineLength])
-				p.writeString(excerptDots)
-			} else {
-				p.writeString(line)
-			}
+				// if the line number increases in digit length during the error,
+				// fill the extra space with blank spaces
+				if i = 0; i < lineNumberLength-len(plainLineNumberString) {
```
turbolent, line 296: `i` is the index of the excerpts, see https://github.com/onflow/cadence/pull/2197/files#diff-2c07386513ebe19503a3f5f086fbd480960c3d2a441c68f2a2fc29ec46008a63R254

Maybe rename that index to something more descriptive, like `excerptIndex` and then introduce a separate index variable here, instead of assigning/overwriting.

runtime/pretty/print.go
```diff
@@ -257,14 +257,8 @@ func (p ErrorPrettyPrinter) writeCodeExcerpts(
 		lineNumberLength := 0
 		if excerpt.startPos != nil {
 
-			plainLineNumberString := strconv.Itoa(excerpt.startPos.Line)
+			plainLineNumberString := strconv.Itoa(excerpt.endPos.Line)
```
turbolent, line 260: Why did this change from start to end position?

runtime/runtime_test.go
```diff
@@ -7706,3 +7706,111 @@ func TestRuntimeTypeMismatchErrorMessage(t *testing.T) {
...
+		},
+	)
+	require.Error(t, err)
+
+	errorString := `Execution failed:
+error: unexpectedly found nil while forcing an Optional value
+  --> 0000000000000000000000000000000000000000000000000000000000000000:9:15
+   |
+ 9 |         return a
+10 |             .firstIndex(of: 5)!
```
turbolent, line 7756: 👌

runtime/runtime_test.go
```diff
@@ -7706,3 +7706,111 @@ func TestRuntimeTypeMismatchErrorMessage(t *testing.T) {
...
+   |                ^^^^^^^^^^^^^^^^
```
turbolent, line 7757: From what I can gather, the indicator still starts at the start pos column (`a`). We can maybe improve that in a subsequent PR, e.g. by rendering indicators for each line

turbolent (commented):
Nice! Sorry for the delay. Just one issue and a question

runtime/pretty/print.go
dsainati1: This is used to compute the "length" (number of digits) in the line number string so we know how many spaces we need to allocate for the left side of the `|` in the error excerpt. In the case where number of digits increases, e.g. in an error between lines 8-12, we need to allocate space for the largest line number in the excerpt, which will always be the `endPos`.

turbolent: Thank you for the explanation, that makes sense! Could you please add a comment?

turbolent (approved):
👌

PR #2202, opened 2022-12-20
Skip broken values during storage iteration

### LINKED ISSUE #2200: Skip broken values during storage iteration
### Issue To Be Solved

As described in https://github.com/onflow/cadence/issues/2170, some storage values could be broken/fail to load, due to outdated contracts. This becomes problematic in cases like storage iteration etc.

### Suggested Solution

Improve the storage iteration API to skip broken values

SupunS:
Closes #2200

## Description

The storage iteration API will now skip any broken values in the storage.

______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -3239,7 +3239,12 @@ func (interpreter *Interpreter) recordStorageMutation() {
 	}
 }
 
-func (interpreter *Interpreter) newStorageIterationFunction(addressValue AddressValue, domain common.PathDomain, pathType sema.Type) *HostFunctionValue {
+func (interpreter *Interpreter) newStorageIterationFunction(
+	addressValue AddressValue,
+	domain common.PathDomain,
+	pathType sema.Type,
+) *HostFunctionValue {
+
 	address := addressValue.ToAddress()
 	config := interpreter.SharedState.Config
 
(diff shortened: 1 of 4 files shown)


runtime/interpreter/interpreter.go
```diff
@@ -3271,8 +3276,31 @@ func (interpreter *Interpreter) newStorageIterationFunction(addressValue Address
...
+							switch r := r.(type) {
+							case errors.UserError, errors.ExternalError:
+								typeError = r.(error)
+							default:
+								panic(r)
+							}
+						}
+					}()
+					_, typeError = inter.ConvertStaticToSemaType(staticType)
+				}()
```
dsainati1, line 3296: is it necessary to have this double nested function here? What errors are we throwing inside `ConvertStaticToSemaType` that aren't being returned in the `err` value?

SupunS: Yup, the nested function is needed to recover errors thrown from  `ConvertStaticToSemaType`. For example, If the import handler fails to load the program (contract), it panics:
https://github.com/onflow/cadence/blob/4ab56fd102820c614d48748dbcfedd9d933344ed/runtime/environment.go#L807-L815

The two tests added cover these two cases; one where the error is returned, and the other for the case where the error is thrown.

SupunS: Extracted the type loading function out, so its no longer a lambda: c236cd4

dsainati1 (approved):

turbolent (approved):
Nice!

PR #2203, opened 2022-12-21
Add equal method for cadence external types

### LINKED ISSUE #2199: Add `Type.Equal(Type) bool` function to compare Cadence types efficiently
### Issue To Be Solved

Checking if `cadence.Type` are equal is currently done by comparing `Type.ID()`.  It should be possible to provide a more efficient  way to do this comparison in Go.

### Suggested Solution

One solution is to add `Type.Equal(Type) bool` function to `cadence.Type` interface.
  
During brief chat about this on Slack, @turbolent confirmed we don't currently have a utility function to do this comparison and also mentioned it should be possible to do recursive comparison without allocating.

SupunS:
Closes #2199

## Description

Adds `Type.Equal(Type) bool` method for cadence external types, since `Type.ID()` cannot be used to compare some types (e.g: restricted types). This newly added `Equal()` method compares go structs recursively and hence doesn't do allocations (except for the restricted types).

______

### DIFF
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -2688,12 +2688,6 @@ func (t *FunctionType) Equal(other Type) bool {
 		return false
 	}
 
-	// constructors
-
-	if t.IsConstructor != otherFunction.IsConstructor {
-		return false
-	}
-
 	return true
 }
 
--- a/types.go
+++ b/types.go
@@ -20,6 +20,7 @@ package cadence
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/onflow/cadence/runtime/common"
 )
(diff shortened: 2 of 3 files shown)


runtime/sema/type.go
```diff
@@ -2688,12 +2688,6 @@ func (t *FunctionType) Equal(other Type) bool {
 		return false
 	}
 
-	// constructors
-
-	if t.IsConstructor != otherFunction.IsConstructor {
-		return false
-	}
-
```
SupunS, line 2696: A code cleanup; This check is already done above

turbolent (commented):
Looks good so far 👍

SupunS:
Oh, this should be ready to review. Work here is done, kept it as a draft until the CI passes, and then forgot to change the state.

dsainati1 (approved):

turbolent (approved):

PR #2225, opened 2023-01-06
Refactor external types to instantiate only once whenever possible

### LINKED ISSUE #2167: Add static type for Cadence external `PathLink` value
### Issue To Be Solved

`cadence.PathLink` doesn't have a static type.  It is needed because:
- [CCF specs (draft)](https://github.com/fxamacker/ccf_draft) needs to include `cadence.PathLink` static type.
- CCF codec (WIP) needs all Cadence external values to have a static type.

Updates #2157 

### Suggested Solution

Maybe add `PathLinkType` and return `PathLinkType` from `PathLink.Type()`.

EDIT: Renamed `Link` to its new name `PathLink`.

SupunS:
Depends on: #2248

## Description

New instances of Cadence external types are created every time during decoding, type conversions, etc.
However, some of them (simple types) can be instantiated only once and can be re-used, which also eliminates the need for memory metering. Hence refactored such types by introducing and re-using global instances.

Also noticed that `cadence.ExportType` method is a duplicate of `cadence.ExportMeteredType` with the only difference being the memory gauge is `nil`. Refactored that as well, to re-use one method for both scenarios.
______


runtime/convertTypes.go
```diff
@@ -29,156 +29,8 @@ import (
 )
 
 // ExportType converts a runtime type to its corresponding Go representation.
-func ExportType(
-	t sema.Type,
-	results map[sema.TypeID]cadence.Type,
-) cadence.Type {
```
SupunS, line 35: This method was a duplicate of `ExportMeteredType(nil, t, results)`

fxamacker (commented):
Thanks for implementing `PathLinkType` and great idea to refactor simple types as singletons.  :+1: 

I have a few questions:
- Should we support encoding and decoding `PathLinkType` as type value?
- Should we make new singleton functions such as `NewAnyType()` unexported to encourage reusing singleton?
- Should we use pointer for simple type singletons?

SupunS:
Thank you for having a look @fxamacker!

> Should we support encoding and decoding PathLinkType as type value?

Currently, a PathLinkType cannot exist at Cadence runtime (there is no run-time/internal type counterpart for external PathLinkType). So even if we add encoding/decoding, it will not be used.
I still need to add a test for this; will be adding it with the rest of the tests 👍 
 
> Should we make new singleton functions such as NewAnyType() unexported to encourage reusing singleton?

I also had the same thought, but eventually decided to keep it as is, so that any downstream dependencies (e.g: go-sdk, any community project, etc.) would not be broken by the API change. I did remove the metered constructor though, since it is supposed to be used internally only.

> Should we use pointer for simple type singletons?

Yeah, the real use of singleton comes if we actually re-use the same memory location. But I'm not sure how will it impact perf (GC in particular). i.e: Go copying in stack vs GC-ing a heap object/pointer. Do you think a pointer would be better performant/memory efficient?
I guess the gain we get from the current refactoring is the removal of metering overhead.

encoding/json/decode.go
```diff
@@ -1142,107 +1142,109 @@ func (d *Decoder) decodeType(valueJSON any, results typeDecodingResults) cadence
...
+		return cadence.TheAnyType
 	case "AnyStruct":
-		return cadence.NewMeteredAnyStructType(d.gauge)
+		return cadence.TheAnyStructType
 	case "AnyResource":
-		return cadence.NewMeteredAnyResourceType(d.gauge)
+		return cadence.TheAnyResourceType
 	case "Type":
-		return cadence.NewMeteredMetaType(d.gauge)
+		return cadence.TheMetaType
```
dreamsmasher, line 1151: "yeah I worked at facebook"

"oh? i never took you for the meta type"

dreamsmasher (approved):

fxamacker:
@SupunS

> Currently, a PathLinkType cannot exist at Cadence runtime (there is no run-time/internal type counterpart for external PathLinkType). So even if we add encoding/decoding, it will not be used.

Ah, that makes sense.  Maybe a comment can be added to document this.  

> Yeah, the real use of singleton comes if we actually re-use the same memory location. But I'm not sure how will it impact perf (GC in particular). i.e: Go copying in stack vs GC-ing a heap object/pointer. Do you think a pointer would be better performant/memory efficient?

Actually,  escape analysis with go1.19.4 shows both non-pointer singleton and pointer singleton escape to the heap.  Pointer singleton is automatically escaped to the heap.  Non-pointer singleton is escaped because of how it is used by value object.  So either approach is OK.

```
var TheVoidType = NewVoidType()

func NewVoidType() *VoidType {
	return &VoidType{}
}
```
```
var TheVoidType = NewVoidType()

func NewVoidType() VoidType {
	return VoidType{}
}
```

> I guess the gain we get from the current refactoring is the removal of metering overhead.

Another benefit of using singletons is memory optimization by reducing heap use.  For example, the current implementation of `Type()` creates a new heap object.  So using singleton will avoid dynamic allocations that just create duplicates on the heap. 

```
func (Void) Type() Type {
	return NewVoidType()
}

```

SupunS:
I can also make the constructor return the singleton (instead of removing the constructor), so we can achieve both at once: reducing memory allocations while keeping the API backward compatible.
```cadence
var TheVoidType = VoidType{}

func NewVoidType() VoidType {
	return TheVoidType
}
```

> Actually, escape analysis with go1.19.4 shows both non-pointer singleton and pointer singleton escape to the heap.

Oh, didn't know that! I was assuming otherwise.

types.go
```diff
@@ -51,13 +51,10 @@ func (t TypeID) Equal(other Type) bool {
 
 type AnyType struct{}
 
-func NewAnyType() AnyType {
-	return AnyType{}
-}
+var TheAnyType = AnyType{}
```
turbolent, line 54: Probably want to box as `Type` here and for the other declarations:
```suggestion
var TheAnyType Type = AnyType{}
```

turbolent (approved):
👍

types.go
SupunS: That would be problematic for the constructors: they return the concrete type
e.g:
```cadence
func NewAnyType() AnyType
```

Issue #2177, opened 2022-12-02
Add support for extended transactions


SupunS:
### Issue To Be Solved

This issue is to track the implementation of FLIP: https://github.com/onflow/flips/pull/41

Assuming the 'role block' syntax would be used, the following needs to be implemented: 
- [ ] Support parsing role blocks/declarations in the parser
  - Add a new AST node to represent role blocks
  - Update the grammar
- [ ] Type check role declarations
  -  Scoping rules would seamlessly work since it will be yet another block.
- [ ] Interpret role block
  - Ensure order: `prepare`, `post` x and other similar blocks must be executed in the same order as the role definitions.
- [ ] Add documentation

TBD:
- [ ] Finalize the syntax and semantics for accessing cross-role variables. i.e: accessing variables belonging to role `A` from outside of the role `A`.

turbolent:
Support for transaction roles is done now. Remaining work is 
- Maybe add support for pre and post conditions to role blocks. This has not been decided yet, so holding off for now
- Documenting the functionality, but blocked until the FLIP has been approved

PR #2234, opened 2023-01-11
Parse transaction role declarations

turbolent:
Work towards #2177

## Description

- Add AST element for transaction role declaration
- Parse transaction role declarations
- Also: Simplify original transaction declaration parsing, remove unnecessary duplicated execute block parsing

______


runtime/parser/transaction.go
```diff
@@ -71,76 +72,65 @@ func parseTransactionDeclaration(p *parser, docString string) (*ast.TransactionD
...
 
-	// Prepare (optional) or execute (optional)
+	// Prepare (optional)
 
 	var prepare *ast.SpecialFunctionDeclaration
-	var execute *ast.SpecialFunctionDeclaration
 
 	p.skipSpaceAndComments()
-	if p.current.Is(lexer.TokenIdentifier) {
+	if p.isToken(p.current, lexer.TokenIdentifier, keywordPrepare) {
```
SupunS, line 80: Nice, this is much cleaner: the order of blocks is much more intuitive now

runtime/parser/declaration_test.go
```diff
@@ -3731,6 +3731,144 @@ func TestParseTransactionDeclaration(t *testing.T) {
...
+		result, errs := testParseDeclarations(`
+          transaction {
+            role buyer {
+              var x: Int
+
+		      prepare(signer: AuthAccount) {
+	            x = 0
+			  }
+            }
+          }
```
SupunS, line 3748:
```suggestion
          transaction {
            role buyer {
              var x: Int

              prepare(signer: AuthAccount) {
                x = 0
              }
            }
          }
```

runtime/parser/declaration_test.go
```diff
@@ -3731,6 +3731,144 @@ func TestParseTransactionDeclaration(t *testing.T) {
 			result.Declarations(),
 		)
 	})
+
+	t.Run("role", func(t *testing.T) {
```
SupunS, line 3735: maybe also add some tests for role blocks without prepare, empty role blocks, etc.

SupunS (approved):

runtime/parser/declaration_test.go
dsainati1: Yea; would like to see tests for the failure cases here. In addition to what Supun mentioned above, would be good to have tests for nested role blocks, role blocks with conditions, and multiple role blocks in a transaction.

turbolent: Added more tests in 67105eb570ec74e27c045912b4aa3ea780a4e179.

Nesting and pre/post conditions are not supported

SupunS (approved):

PR #2241, opened 2023-01-12
[v0.31] Ensure SetProgram is called after GetProgram, even if there is an error

turbolent:
## Description

Calls to `runtime.Interface.GetProgram` and `SetProgram` need to be balanced.
So far this was not the case when an error occurred after `GetProgram` was called: 

A subsequent error, e.g. a parsing or type checking error, was bubbled up and lead to the failure of the whole program execution (e.g. when a static import fails).

However, since #2202, storage iteration skips type loading errors, which results in `GetProgram` calls without subsequent `SetProgram` calls.

Ensure `SetProgram` is always called after a call to `GetProgram`, even when an error occurs.

Test the pairing of `GetProgram`/`SetProgram` calls by re-implementing the defensive check FVM implements: Keep a stack of calls and ensure there are no outstanding calls to `SetProgram` at the end of execution.

We should maybe enable this check for all runtime tests in a subsequent PR.

This is a fix of the current runtime interface. We should aim to land #2180 next which should reduce/remove the possibility of incorrect uses of the get/set program functions.

______

### DIFF
--- a/runtime/environment.go
+++ b/runtime/environment.go
@@ -382,11 +382,7 @@ func (e *interpreterEnvironment) parseAndCheckProgram(
 		wrapPanic(func() {
 			err = e.runtimeInterface.SetProgram(location, program)
 		})
-		if err != nil {
-			return err
-		}
-
-		return nil
+		return err
 	}
 
 	wrapParsingCheckingError := func(err error) error {
--- a/runtime/environment.go
+++ b/runtime/environment.go
@@ -348,6 +348,8 @@ func (e *interpreterEnvironment) ParseAndCheckProgram(
 	)
 }
 
+// parseAndCheckProgram parses and checks the given program.
+// If storeProgram is true, it calls Interface.SetProgram.
 func (e *interpreterEnvironment) parseAndCheckProgram(
 	code []byte,
 	location common.Location,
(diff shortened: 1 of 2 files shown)


dsainati1 (approved):

runtime/environment.go
```diff
@@ -357,7 +359,37 @@ func (e *interpreterEnvironment) parseAndCheckProgram(
...
+
+		var err error
+		wrapPanic(func() {
+			err = e.runtimeInterface.SetProgram(location, program)
+		})
+		if err != nil {
+			return err
+		}
+
+		return nil
```
SupunS, line 389: same as:
```suggestion
		return err
```

SupunS (approved):
Looks fine to me. I'll have another round of looks tomorrow.

runtime/environment.go
turbolent: Simplified in [f8f740d](https://github.com/onflow/cadence/pull/2241/commits/f8f740dd98f2012e3da0c5c70edd3d942d605073)

bluesign:
Is this something like borrowProgram? I read the related code, but why is the balance required?

janezpodhostnik (approved):

turbolent:
@bluesign programs here mean: parsing result (AST) + type checking result (elaboration). The `GetProgram` / `SetProgram` calls are used by Cadence during type checking and interpretation to refer to the same programs again if they are used in multiple places. E.g imagine a program A that imports B. If C also imports B, then the same B must be returned that was the result of importing B from A.

In addition, these calls are then also further used in FVM to cache programs across multiple scripts/transactions.

janezpodhostnik:
@bluesign 
GetProgram calls the FVM programs cache, and if it is a miss, the FVM assumes that the program is going to get loaded by cadence, so it pushes a View on the stack to track all the register touches that are going to happen during contract loading. 
SetProgram should pop this view. If the view isn't popped this leads to problems.

This get changed to a closure in the future, so this scenario cant happen.

PR #2242, opened 2023-01-13
Add publicTypes method to DeployedContract for introspection

### LINKED ISSUE #2232: Add computed field/method on `Contract`s to view all declarations
### Issue To Be Solved

Part of #2201.

As part of the NFT V2 spec, we need a way to view all declarations within a given contract.
### Suggested Solution

This could consist of a single computed field or getter on the `Contract` type that returns an array of `Type`s: 

```cadence
pub fun Contract.publicDeclarations(): [Type]
```

for further granularity, we could introduce separate getters for each kind of declaration: function, struct, resource, interface, etc.

dreamsmasher:
Closes #2232 

## Description

______
This PR adds a method to `DeployedContract`:

```cadence
fun DeployedContract.publicTypes(): [Type]
```

to enable smart contracts to view all public type declarations in a contract. This enables one to filter through all the declared types in a contract, e.g. to find all the types that implement a given interface.

As an MVP, this PR only adds the one method, which returns both struct and resource declarations. If needed, more granular methods can be implemented specifically for the struct and resource declarations in a contract.

### DIFF
--- a/docs/language/accounts.mdx
+++ b/docs/language/accounts.mdx
@@ -211,6 +211,9 @@ to the `prepare` phase of the transaction.
   struct DeployedContract {
       let name: String
       let code: [UInt8]
+
+      // Returns an array of `Type` objects representing all the public struct and resource declarations in this contract
+      fun publicTypes(): [Type]
   }
   ```
 
--- a/docs/language/accounts.mdx
+++ b/docs/language/accounts.mdx
@@ -212,7 +212,17 @@ to the `prepare` phase of the transaction.
       let name: String
       let code: [UInt8]
 
-      // Returns an array of `Type` objects representing all the public struct and resource declarations in this contract
+      // Returns an array of `Type` objects representing all the public type declarations in this contract (e.g. structs, resources, enums)
+      // 
+      // For example, given a contract
+      // ```
+      // contract Foo {
+      //       pub struct Bar {...}
+      //       pub resource Qux {...}
+      // }
+      // ```
+      // then `.publicTypes()` will return an array equivalent to the expression `[Type<Bar>(), Type<Qux>()]`
+    
       fun publicTypes(): [Type]
   }
   ```
--- a/runtime/deployedcontract_test.go
+++ b/runtime/deployedcontract_test.go
@@ -1,0 +1,107 @@
+package runtime
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/onflow/cadence"
+	"github.com/onflow/cadence/runtime/common"
+	"github.com/onflow/cadence/runtime/tests/utils"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDeployedContracts(t *testing.T) {
+	t.Parallel()
+
+	contractCode := `
+		pub contract Test {
+			pub struct A {}
+			pub struct B {}
+			pub struct C {}
+
+			init() {}
+		}
+	`
+
+	script :=
+		`
+		transaction {
+			prepare(signer: AuthAccount) {
+				let deployedContract = signer.contracts.get(name: "Test")
+				log(signer.contracts.names)
+				assert(deployedContract!.name == "Test")
+
+				let types = deployedContract!.publicTypes()
+				let identifiers: [String] = []
+			}
+		}
+		`
+
+	rt := newTestInterpreterRuntime()
+	accountCodes := map[Location][]byte{}
+
+	runtimeInterface := &testRuntimeInterface{
+		getCode: func(location Location) (bytes []byte, err error) {
+			return accountCodes[location], nil
+		},
+		getSigningAccounts: func() ([]Address, error) {
+			return []Address{{42}}, nil
+		},
+		getAccountContractCode: func(address Address, name string) ([]byte, error) {
+			location := common.AddressLocation{
+				Address: address, Name: name,
+			}
+			return accountCodes[location], nil
+		},
+		getAccountContractNames: func(_ Address) ([]string, error) {
+			names := make([]string, 0, len(accountCodes))
+			for location, _ := range accountCodes {
+				names = append(names, location.String())
+			}
+			return names, nil
+		},
+		emitEvent: func(event cadence.Event) error {
+			fmt.Println(event.String())
+			return nil
+		},
+		updateAccountContractCode: func(address common.Address, name string, code []byte) error {
+			location := common.AddressLocation{
+				Address: address, Name: name,
+			}
+			accountCodes[location] = code
+			fmt.Println(accountCodes)
+			return nil
+		},
+		log: func(msg string) {
+			fmt.Println(msg)
+		},
+		storage: newTestLedger(nil, nil),
+	}
+
+	nextTransactionLocation := newTransactionLocationGenerator()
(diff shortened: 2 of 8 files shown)


turbolent:
Given that this is not a breaking change and just a feature addition, we can target `master` instead of the Stable Cadence

turbolent:
Please add a label. The checkbox is already checked, but there's none

runtime/deployedcontract_test.go
```diff
@@ -0,0 +1,114 @@
+package runtime
```
turbolent, line 1: Needs license header

runtime/deployedcontract_test.go
```diff
@@ -0,0 +1,114 @@
...
+		},
+		updateAccountContractCode: func(address common.Address, name string, code []byte) error {
+			location := common.AddressLocation{
+				Address: address, Name: name,
+			}
+			accountCodes[location] = code
+			return nil
+		},
+		log: func(msg string) {
+			fmt.Println(msg)
```
turbolent, line 83: Don't println in tests. Instead, just ignore if there are no assertions about logged messages

runtime/interpreter/deployedcontract.go
```diff
@@ -37,18 +41,62 @@ func NewDeployedContractValue(
...
 		},
 		nil,
 		nil,
 		nil,
 	)
 }
+
+func newPublicTypesFunctionValue(inter *Interpreter, addressValue AddressValue, name *StringValue) FunctionValue {
+	// public types only need to be computed once per contract
+	var once sync.Once
```
turbolent, line 64: No need to make this thread-safe. Maybe just replace by initializing when `publicTypes == nil`

runtime/interpreter/deployedcontract.go
```diff
@@ -37,18 +41,62 @@ func NewDeployedContractValue(
...
+				panic(err)
+			}
+
+			nestedTypes := compositeType.NestedTypes
+			pair := nestedTypes.Oldest()
+			yieldNext := func() Value {
+				if pair == nil {
+					return nil
+				}
+				typeValue := TypeValue{Type: ConvertSemaToStaticType(innerInter, pair.Value)}
```
turbolent, line 86: Use the metering `NewTypeValue` constructor instead of the struct literal

runtime/interpreter/deployedcontract.go
```diff
@@ -37,18 +41,62 @@ func NewDeployedContractValue(
...
+				pair = pair.Next()
+				return typeValue
+			}
+
+			publicTypes = NewArrayValueWithIterator(
+				innerInter,
+				NewVariableSizedStaticType(innerInter, PrimitiveStaticTypeMetaType),
+				common.Address{0x0},
```
turbolent, line 94: Can just use the default value of `Address` (an `[8]byte`)

```suggestion
				common.Address{},
```

runtime/sema/deployed_contract.go
```diff
@@ -95,3 +107,18 @@ const DeployedContractTypeCodeFieldName = "code"
...
 `
+
+const DeployedContractTypePublicTypesFunctionName = "publicTypes"
+const DeployedContractTypePublicTypesFunctionDocString = `
+	Returns an array of Type objects representing all public type declarations in this contract.
+`
+
+var DeployedContractTypePublicTypesFunctionType = NewSimpleFunctionType(
+	FunctionPurityView,
+	[]Parameter{},
```
turbolent, line 118: No need to allocate a slice, can just be `nil`

```suggestion
	nil,
```

runtime/tests/checker/account_test.go
```diff
@@ -1454,6 +1454,7 @@ func TestAuthAccountContracts(t *testing.T) {
 	t.Parallel()
 
 	t.Run("contracts type", func(t *testing.T) {
+		t.Parallel()
```
turbolent, line 1457: 👍

turbolent (commented):
Looks good, just a few minor changes needed

runtime/sema/deployed_contract.go
```diff
@@ -95,3 +107,18 @@ const DeployedContractTypeCodeFieldName = "code"
 const deployedContractTypeCodeFieldDocString = `
 The code of the contract
 `
+
+const DeployedContractTypePublicTypesFunctionName = "publicTypes"
+const DeployedContractTypePublicTypesFunctionDocString = `
+	Returns an array of Type objects representing all public type declarations in this contract.
```
SupunS, line 113:
```suggestion
Returns an array of Type objects representing all public type declarations in this contract.
```

runtime/deployedcontract_test.go
```diff
@@ -0,0 +1,114 @@
...
+				let expected: {String: Void} =  
+					{ "A.2a00000000000000.Test.A": ()
+					, "A.2a00000000000000.Test.B": ()
+					, "A.2a00000000000000.Test.C": ()
+					}
+				let types = deployedContract!.publicTypes()
+				assert(types.length == 3)
+
+				for type in types {
+					assert(expected[type.identifier] != nil, message: type.identifier)
```
SupunS, line 43: maybe make this a script, return the types from the script, and assert them at Go level.  It gives more control over the assertion: e.g: Can check for the types-kind, names, etc.

dreamsmasher: ah yeah, that was my original idea but faced some roadblocks when trying to access the contract itself from the script. 

when we deploy the contracts to the runtime manually using `rt.ExecuteTransaction`, are they stored under the signer's address? what address should we be passing to `getAccount` inside Cadence?

SupunS: > when we deploy the contracts to the runtime manually using rt.ExecuteTransaction, are they stored under the signer's address? what address should we be passing to `getAccount` inside Cadence?

Yes. Should be able to get the account using the same address.

Alternatively, can always return the same contract code inside `GetAccountContractCode` without really worrying about the address, since there's only one contract anyway.

docs/language/accounts.mdx
```diff
@@ -211,6 +211,9 @@ to the `prepare` phase of the transaction.
   struct DeployedContract {
       let name: String
       let code: [UInt8]
+
+      // Returns an array of `Type` objects representing all the public struct and resource declarations in this contract
```
turbolent, line 215: Is it specifically only structs and resources? It probably also returns enums? Maybe turn the kinds into an example list:
```suggestion
      // Returns an array of `Type` objects representing all the public declarations in this contract (e.g. structs, resources)
```

Maybe also update the docstring accordingly (examples are useful!)

turbolent (approved):
Nice!

turbolent:
@dsainati1 Could you also please have a look? 🙏

runtime/deployedcontract_test.go
```diff
@@ -0,0 +1,126 @@
+/*
+ * Cadence - The resource-oriented smart contract programming language
+ *
+ * Copyright 2019-2022 Dapper Labs, Inc.
```
SupunS, line 4:
```suggestion
 * Copyright Dapper Labs, Inc.
```

SupunS (approved):

PR #2249, opened 2023-01-14
Ensure GetProgram calls before SetProgram

turbolent:
Work towards #2247 

## Description

- [Add missing program stack push/pop to storage iteration skip test case](https://github.com/onflow/cadence/commit/23f4bc427aa8abbedee915fb51885bb9251a0a43)
- [Fix program stack assertions in storage iteration skip test cases](https://github.com/onflow/cadence/commit/12f982b57fe6829f31c17409d2cd823e2b6eb03f)
- [Ensure `ParseAndCheckProgram` calls `GetProgram` before `SetProgram`](https://github.com/onflow/cadence/commit/396a5f131a3b1977afc3ee2de2601f1f3a12fe9a): Previously, `GetProgram` was not called for scripts and transactions. This reactor is the required step to get #2180 in, and finally replace `GetProgram`/`GetOrLoadProgram` and `SetProgram` with a single function call (make set implicit as a result of the get or load)
- [Require `ScriptLocation` for `ExecuteScript`, and `TransactionLocation` for `ExecuteTransaction`](https://github.com/onflow/cadence/commit/c5a37f2896fd4f94101e7db502ed1ec56caf7ae4): Previously, many of the context locations had incorrect types and were also re-used across multiple executions, which results in incorrect caching

______

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #2254, opened 2023-01-16
Replace interface GetProgram and SetProgram with GetAndSetProgram

### LINKED ISSUE #2247: Runtime interface GetProgram and SetProgram calls must be in pairs
### Current Behavior

After calls to `runtime.Interface.GetProgram` calls to `SetProgram` are missing when an error is encountered. 

### Expected Behavior

`runtime.Interface` calls to `GetProgram` and `SetProgram` must be done in pairs, even if an error is encountered. 

### Steps To Reproduce

https://www.notion.so/dapperlabs/Storage-Iteration-with-broken-contracts-bug-91d98f9c23604ee7b585fba54a6d18a9

### Environment

```markdown
- Cadence version: v0.31.2
- Network: Testnet
```

turbolent:
⚠️ Depends on #2249
Closes #2247
Closes #2180 (replacement)

## Description

Bases on the great start by @janezpodhostnik  in #2180 🙏 

With #2249, all `GetProgram` and `SetProgram` calls are now balanced.
Finally, replace both functions with a single function `GetAndSetProgram`.

______

### DIFF
--- a/runtime/contract_update_test.go
+++ b/runtime/contract_update_test.go
@@ -42,8 +42,14 @@ func TestContractUpdateWithDependencies(t *testing.T) {
 		Address: signerAccount,
 		Name:    "Foo",
 	}
+	var checkGetAndSetProgram, getProgramCalled bool
 
-	var checkGetSetProgram, getProgramCalled, setProgramCalled bool
+	programs := map[Location]*interpreter.Program{}
+	clearPrograms := func() {
+		for l := range programs {
+			delete(programs, l)
+		}
+	}
 
 	runtimeInterface := &testRuntimeInterface{
 		getCode: func(location Location) (bytes []byte, err error) {
@@ -75,24 +81,33 @@ func TestContractUpdateWithDependencies(t *testing.T) {
 		decodeArgument: func(b []byte, t cadence.Type) (value cadence.Value, err error) {
 			return json.Decode(nil, b)
 		},
-		getProgram: func(location Location) (*interpreter.Program, error) {
+		getAndSetProgram: func(
+			location Location,
+			load func() (*interpreter.Program, error),
+		) (
+			program *interpreter.Program,
+			err error,
+		) {
 			_, isTransactionLocation := location.(common.TransactionLocation)
-			if checkGetSetProgram && !isTransactionLocation {
+			if checkGetAndSetProgram && !isTransactionLocation {
 				require.Equal(t, location, fooLocation)
 				require.False(t, getProgramCalled)
-				getProgramCalled = true
 			}
-			// Always force to get the old program from the source during the update.
-			return nil, nil
-		},
-		setProgram: func(location Location, program *interpreter.Program) error {
-			_, isTransactionLocation := location.(common.TransactionLocation)
-			if checkGetSetProgram && !isTransactionLocation {
-				require.Equal(t, location, fooLocation)
-				require.False(t, setProgramCalled)
-				setProgramCalled = true
+
+			var ok bool
+			program, ok = programs[location]
+			if ok {
+				return
 			}
-			return nil
+
+			program, err = load()
+
+			// NOTE: important: still set empty program,
+			// even if error occurred
+
+			programs[location] = program
+
+			return
 		},
 	}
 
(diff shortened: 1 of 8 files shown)


runtime/contract_update_test.go
```diff
@@ -75,24 +81,33 @@ func TestContractUpdateWithDependencies(t *testing.T) {
...
+		) {
 			_, isTransactionLocation := location.(common.TransactionLocation)
-			if checkGetSetProgram && !isTransactionLocation {
+			if checkGetAndSetProgram && !isTransactionLocation {
 				require.Equal(t, location, fooLocation)
 				require.False(t, getProgramCalled)
-				getProgramCalled = true
 			}
-			// Always force to get the old program from the source during the update.
-			return nil, nil
```
turbolent, line 86: This was an incorrect implementation of `GetProgram`. Instead, it should return what it loaded. Also, programs are only effective during the duration of execution (e.g. of a transaction or script), not across them, so clear all programs. This is because contract updates are delayed, and the test runtime interface does not properly commit/overwrite the new programs

runtime/deployment_test.go
```diff
@@ -283,7 +283,7 @@ func TestRuntimeTransactionWithContractDeployment(t *testing.T) {
 					"  |\n"+
 					"4 |               fun testCase() {}\n"+
 					"  |               ^\n",
-				1,
+				2,
```
turbolent, line 286: `codesAndPrograms` has now more information, which is better

runtime/environment.go
```diff
@@ -355,103 +355,42 @@ func (e *interpreterEnvironment) ParseAndCheckProgram(
 func (e *interpreterEnvironment) parseAndCheckProgram(
```
turbolent, line 355: `parseAndCheckProgram` no longer has the responsibility of calling `SetProgram` if `storeProgram` is true

runtime/storage_test.go
```diff
@@ -3403,34 +3403,13 @@ func TestRuntimeStorageIteration(t *testing.T) {
...
-
-			runtimeInterface := &testRuntimeInterface{
+		newRuntimeInterface := func() Interface {
+			return &testRuntimeInterface{
 				storage: ledger,
 				getSigningAccounts: func() ([]Address, error) {
 					return []Address{address}, nil
 				},
 				resolveLocation: singleIdentifierLocationResolver(t),
-				getProgram: func(location Location) (*interpreter.Program, error) {
```
turbolent, line 3417: The `getProgram` and `setProgram` really only checked the balancing of the `GetProgram` and `SetProgram` calls. Given that this is no longer possible, remove those assertions

runtime/environment.go
```diff
@@ -577,48 +516,51 @@ func (e *interpreterEnvironment) getProgram(
...
 
-	e.codesAndPrograms.setProgram(location, program.Program)
+		return &interpreter.Program{
+			Program:     parsedProgram,
+			Elaboration: elaboration,
+		}, nil
+	}
 
-	return program, nil
+	if getAndSetProgram {
```
SupunS, line 545: Can invert the condition to reduce the nesting here
e.g:
```go
if !getAndSetProgram {
    return load()
}

wrapPanic(func() {
...
}

**Review by SupunS — APPROVED** [review]

File: runtime/interface.go [hunk 6]
```diff
@@ -34,36 +34,25 @@ type Interface interface {
...
-	//   As SetProgram is called when GetProgram is nil, then SetProgram MUST also be called when
-	//   GetProgram returns a program. This prevents nondeterministic behaviour
-	//
-	// Deprecated: This function should be refactored to ensure that SetProgram is always called
-	//
-	GetProgram(Location) (*interpreter.Program, error)
-	// SetProgram sets the program for the given location.
-	SetProgram(Location, *interpreter.Program) error
+	// - This function MUST also return what was previously loaded,
+	//   it may NOT return something different or nothing/nil (!) after load was called.
```
janezpodhostnik, line 50: what about the case where load itself returns:
- nil
- or an error

janezpodhostnik (approved):

runtime/interface.go
turbolent: Tried to improve the comment further in [e3cfb8e](https://github.com/onflow/cadence/pull/2254/commits/e3cfb8eee8c43a54d5ae2c8eb400e650324fe9a9)

PR #2256, opened 2023-01-16
Replace empty common.Address{} with singleton

dreamsmasher:
## Description

Introduces a global singleton `common.ZeroAddress` to be used in place of the empty constructor `common.Address{}` that pops up a bunch. This helps with code clarity and robustness in case we change the internal implementation of `Address`.
______

### DIFF
--- a/runtime/common/address.go
+++ b/runtime/common/address.go
@@ -32,7 +32,7 @@ const AddressLength = 8
 type Address [AddressLength]byte
 
 // Singleton for address 0x0. This should be used in place of an empty initializer `Address{}`
-var NilAddress Address = Address{}
+var ZeroAddress Address = Address{}
 
 // MustBytesToAddress returns Address with value b.
 //
--- a/runtime/common/address.go
+++ b/runtime/common/address.go
@@ -31,6 +31,9 @@ const AddressLength = 8
 
 type Address [AddressLength]byte
 
+// Singleton for address 0x0. This should be used in place of an empty initializer `Address{}`
+var NilAddress Address = Address{}
+
 // MustBytesToAddress returns Address with value b.
 //
 // If the address is too large, then the function panics.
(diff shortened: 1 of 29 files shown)


dreamsmasher:
lol whoops git was named after me

runtime/common/address.go
```diff
@@ -31,6 +31,9 @@ const AddressLength = 8
 
 type Address [AddressLength]byte
 
+// Singleton for address 0x0. This should be used in place of an empty initializer `Address{}`
+var NilAddress Address = Address{}
```
dsainati1, line 35: Nitpick: this is more of a "ZeroAddress" than a "NilAddress", can we use that name instead?

turbolent:
Is there any difference between the two? Is the main advantage using a specific name/variable?

dreamsmasher:
> Is there any difference between the two? Is the main advantage using a specific name/variable?

Yes, there are over 100 places where we use empty addresses without any clear indication why. It's mostly a stylistic choice to refactor to a singleton given that addresses are 64-bit, but I'd argue this is more in line with the other singleton refactors that have been implemented lately

turbolent:
I don't have any preference regarding style, but just to clarify, the other singleton refactors where for performance / memory allocation reasons: In cases where there's an interface and multiple concrete types conforming to them (e.g. `sema.Type`), it makes sense to box once and keep it as a singleton (e.g. for `sema.AddressType`, `sema.TheAddressType`), to avoid allocation and boxing in cases where the interface type is needed. This isn't the case here

turbolent (approved):

runtime/common/address.go
bluesign: As 0x0 is invalid address, I think NilAddress makes more sense here. 

Though I believe `Address{}` was also clear, I think we can maybe also alias `Stack` here. Most of the code usage of `Address{}` is creating something on stack I believe.

SupunS (approved):

PR #2257, opened 2023-01-18
Type check transaction role declarations

turbolent:
Work towards #2177

______

### DIFF
--- a/runtime/sema/check_transaction_declaration.go
+++ b/runtime/sema/check_transaction_declaration.go
@@ -31,10 +31,8 @@ func (checker *Checker) VisitTransactionDeclaration(declaration *ast.Transaction
 		panic(errors.NewUnreachableError())
 	}
 
-	checker.containerTypes[transactionType] = true
-	defer func() {
-		checker.containerTypes[transactionType] = false
-	}()
+	checker.containerTypes[transactionType] = struct{}{}
+	defer delete(checker.containerTypes, transactionType)
 
 	fields := declaration.Fields
 	fieldMembers := orderedmap.New[MemberFieldDeclarationOrderedMap](len(fields))
--- a/runtime/sema/check_transaction_declaration.go
+++ b/runtime/sema/check_transaction_declaration.go
@@ -319,54 +319,48 @@ func (checker *Checker) declareTransactionDeclaration(declaration *ast.Transacti
 		transactionType.PrepareParameters = checker.parameters(parameterList)
 	}
 
-	roles := map[string]struct{}{}
 	for _, roleDeclaration := range declaration.Roles {
 		transactionRoleType := checker.transactionRoleType(roleDeclaration)
 		checker.Elaboration.SetTransactionRoleDeclarationType(roleDeclaration, transactionRoleType)
 
-		// Ensure roles are not duplicated
-		roleName := roleDeclaration.Identifier.Identifier
-		if _, ok := roles[roleName]; ok {
-			checker.report(
-				&DuplicateTransactionRoleError{
-					Name: roleName,
-					Range: ast.NewRangeFromPositioned(
-						checker.memoryGauge,
-						roleDeclaration.Identifier,
-					),
-				},
-			)
-			continue
-		} else {
-			roles[roleName] = struct{}{}
-		}
-
 		// Ensure roles and fields do not clash
+		roleName := roleDeclaration.Identifier.Identifier
 		if member, ok := members.Get(roleName); ok {
-			if member != nil {
+
+			identifierRange := ast.NewRangeFromPositioned(
+				checker.memoryGauge,
+				roleDeclaration.Identifier,
+			)
+
+			if _, ok := member.TypeAnnotation.Type.(*TransactionRoleType); ok {
+				checker.report(
+					&DuplicateTransactionRoleError{
+						Name:  roleName,
+						Range: identifierRange,
+					},
+				)
+			} else {
 				checker.report(
 					&TransactionRoleWithFieldNameError{
 						FieldIdentifier: member.Identifier,
-						Range: ast.NewRangeFromPositioned(
-							checker.memoryGauge,
-							roleDeclaration.Identifier,
-						),
+						Range:           identifierRange,
 					},
 				)
 			}
-		} else {
-			members.Set(
-				roleName,
-				&Member{
-					ContainerType:   transactionType,
-					Identifier:      roleDeclaration.Identifier,
-					DeclarationKind: common.DeclarationKindTransactionRole,
-					VariableKind:    ast.VariableKindConstant,
-					TypeAnnotation:  NewTypeAnnotation(transactionRoleType),
-					DocString:       roleDeclaration.DocString,
-				},
-			)
+			continue
 		}
+
+		members.Set(
+			roleName,
+			&Member{
+				ContainerType:   transactionType,
+				Identifier:      roleDeclaration.Identifier,
+				DeclarationKind: common.DeclarationKindTransactionRole,
+				VariableKind:    ast.VariableKindConstant,
+				TypeAnnotation:  NewTypeAnnotation(transactionRoleType),
+				DocString:       roleDeclaration.DocString,
+			},
+		)
 	}
 
 	checker.Elaboration.SetTransactionDeclarationType(declaration, transactionType)
--- a/runtime/sema/check_transaction_declaration.go
+++ b/runtime/sema/check_transaction_declaration.go
@@ -256,18 +256,19 @@ func (checker *Checker) declareTransactionDeclaration(declaration *ast.Transacti
 	}
 
 	roles := &orderedmap.OrderedMap[string, *TransactionRoleType]{}
-	for _, role := range declaration.Roles {
-		transactionRoleType := checker.transactionRoleType(role)
+	for _, roleDeclaration := range declaration.Roles {
+		transactionRoleType := checker.transactionRoleType(roleDeclaration)
+		checker.Elaboration.SetTransactionRoleDeclarationType(roleDeclaration, transactionRoleType)
 
 		// Ensure roles are not duplicated
-		roleName := role.Identifier.Identifier
+		roleName := roleDeclaration.Identifier.Identifier
 		if _, ok := roles.Get(roleName); ok {
 			checker.report(
 				&DuplicateTransactionRoleError{
 					Name: roleName,
 					Range: ast.NewRangeFromPositioned(
 						checker.memoryGauge,
-						role.Identifier,
+						roleDeclaration.Identifier,
 					),
 				},
 			)
@@ -281,7 +282,7 @@ func (checker *Checker) declareTransactionDeclaration(declaration *ast.Transacti
 					Name: roleName,
 					Range: ast.NewRangeFromPositioned(
 						checker.memoryGauge,
-						role.Identifier,
+						roleDeclaration.Identifier,
 					),
 				},
 			)
@@ -293,11 +294,11 @@ func (checker *Checker) declareTransactionDeclaration(declaration *ast.Transacti
 			roleName,
 			&Member{
 				ContainerType:   transactionType,
-				Identifier:      role.Identifier,
+				Identifier:      roleDeclaration.Identifier,
 				DeclarationKind: common.DeclarationKindTransactionRole,
 				VariableKind:    ast.VariableKindConstant,
 				TypeAnnotation:  NewTypeAnnotation(transactionRoleType),
-				DocString:       role.DocString,
+				DocString:       roleDeclaration.DocString,
 			},
 		)
 	}
--- a/runtime/sema/check_transaction_declaration.go
+++ b/runtime/sema/check_transaction_declaration.go
@@ -278,9 +278,10 @@ func (checker *Checker) declareTransactionDeclaration(declaration *ast.Transacti
 
 		// Ensure roles and fields do not clash
 		if _, ok := members.Get(roleName); ok {
+			field := declaration.DeclarationMembers().FieldsByIdentifier()[roleName]
 			checker.report(
 				&TransactionRoleWithFieldNameError{
-					Name: roleName,
+					FieldIdentifier: field.Identifier,
 					Range: ast.NewRangeFromPositioned(
 						checker.memoryGauge,
 						roleDeclaration.Identifier,
--- a/runtime/sema/check_transaction_declaration.go
+++ b/runtime/sema/check_transaction_declaration.go
@@ -332,7 +332,7 @@ func (checker *Checker) declareTransactionDeclaration(declaration *ast.Transacti
 				roleDeclaration.Identifier,
 			)
 
-			if _, ok := member.TypeAnnotation.Type.(*TransactionRoleType); ok {
+			if member.DeclarationKind == common.DeclarationKindTransactionRole {
 				checker.report(
 					&DuplicateTransactionRoleError{
 						Name:  roleName,
--- a/runtime/sema/check_transaction_declaration.go
+++ b/runtime/sema/check_transaction_declaration.go
@@ -42,7 +42,9 @@ func (checker *Checker) VisitTransactionDeclaration(declaration *ast.Transaction
 	fieldMembers := getFieldMembers(members, fields)
 
 	checker.checkTransactionFields(fields)
-	checker.checkPrepareExists(declaration.Prepare, fields)
+
+	prepareFunctionDeclaration := declaration.Prepare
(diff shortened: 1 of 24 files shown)


runtime/sema/type.go
```diff
@@ -5758,6 +5761,105 @@ func (t *TransactionType) Resolve(_ *TypeParameterTypeOrderedMap) Type {
...
+func (*TransactionRoleType) String() string {
+	return "TransactionRole"
+}
+
+func (*TransactionRoleType) QualifiedString() string {
+	return "TransactionRole"
+}
+
+func (*TransactionRoleType) ID() TypeID {
+	return "TransactionRole"
```
dreamsmasher, line 5797: maybe extract these into a constant

runtime/tests/checker/transactions_test.go
```diff
@@ -348,6 +365,206 @@ func TestCheckTransactions(t *testing.T) {
...
+                      let yang: String
+
+                      prepare() {
+                          self.yang = "2"
+                      }
+                  }
+
+                  execute {
+                      let bar: Int = self.foo.bar
+                      let yang: String = self.ying.yang
```
dreamsmasher, line 542: can we not trivialize chinese philosophy for test cases? also it's yin not ying

dreamsmasher (approved):
LGTM, just some nitpicks

runtime/tests/checker/transactions_test.go
turbolent: Huh? I used it because I thought it was nice, I wasn't intentionally trying to trivialize it. 

Changed it in 557080cf6c8b51db54f226980eaab66befb20a14

runtime/sema/type.go
turbolent: Refactored in [a5ca45a](https://github.com/onflow/cadence/pull/2257/commits/a5ca45aba8ec70e8cb41a4bfe7e9dcbd36c6e581)

runtime/sema/check_transaction_declaration.go
```diff
@@ -248,36 +239,168 @@ func (checker *Checker) declareTransactionDeclaration(declaration *ast.Transacti
...
+
+	if declaration.Prepare != nil {
+		parameterList := declaration.Prepare.FunctionDeclaration.ParameterList
+		transactionType.PrepareParameters = checker.parameters(parameterList)
+	}
+
+	roles := &orderedmap.OrderedMap[string, *TransactionRoleType]{}
+	for _, roleDeclaration := range declaration.Roles {
+		transactionRoleType := checker.transactionRoleType(roleDeclaration)
+		checker.Elaboration.SetTransactionRoleDeclarationType(roleDeclaration, transactionRoleType)
```
dsainati1, line 262: Should we move this after the duplicate/field name clash checking? It seems a waste to compute the type if we are just going to discard it when we `continue`.

runtime/sema/check_transaction_declaration.go
```diff
@@ -248,36 +239,168 @@ func (checker *Checker) declareTransactionDeclaration(declaration *ast.Transacti
...
+		fieldName := field.Identifier.Identifier
+		// Fields were previously declared in declareTransactionDeclaration
+		member, ok := transactionRoleType.Members.Get(fieldName)
+
+		if !ok {
+			panic(errors.NewUnreachableError())
+		}
+
+		fieldMembers.Set(member, field)
 	}
```
dsainati1, line 387: Code is shared between `VisitTransactionRoleDeclaration` and `VisitTransactionDeclaration`; refactor to function?

runtime/sema/errors.go
```diff
@@ -3009,57 +2974,67 @@ func (e *InvalidNonImportableTransactionParameterTypeError) Error() string {
...
+		e.Name,
+	)
 }
 
-// InvalidTransactionPrepareParameterTypeError
+// TransactionRoleWithFieldNameError
 
-type InvalidTransactionPrepareParameterTypeError struct {
-	Type Type
+type TransactionRoleWithFieldNameError struct {
```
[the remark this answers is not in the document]
turbolent: Good idea! Added in abf74c30acb6385b022af0c2b7328394630121d2

runtime/tests/checker/transactions_test.go
```diff
@@ -348,6 +365,206 @@ func TestCheckTransactions(t *testing.T) {
...
+		)
+	})
+
+	t.Run("multiple roles, one field each, execute", func(t *testing.T) {
+		test(
+			t,
+			`
+              transaction {
+
+                  role foo {
```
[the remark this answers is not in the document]
turbolent: I felt like there was not enough feedback yet to indicate of it is needed, and if so, how, so I left it out for now. We can add support for it in subsequent PRs. That also keeps this PR small

runtime/sema/errors.go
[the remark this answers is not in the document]
turbolent: Simplified in f2396c7c0d34a69f2f92940ac67f59a782d4108a

runtime/sema/check_transaction_declaration.go
turbolent: We always want to report as much as possible. Just because the field is duplicated or there is a clash with a field should not prevent us from also checking the transaction role itself later on.

runtime/sema/check_transaction_declaration.go
```diff
@@ -248,36 +239,168 @@ func (checker *Checker) declareTransactionDeclaration(declaration *ast.Transacti
...
-		declaration.DeclarationKind(),
+		declarationKind,
 	)
+}
 
-	transactionType.Members = members
-	transactionType.Fields = fields
-	if checker.PositionInfo != nil {
-		checker.PositionInfo.recordMemberOrigins(transactionType, origins)
+func (checker *Checker) VisitTransactionRoleDeclaration(declaration *ast.TransactionRoleDeclaration) (_ struct{}) {
```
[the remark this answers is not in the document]
turbolent: Same as above: I felt like there was not enough feedback yet to indicate of it is needed, so I left it out for now. We can add support for it in subsequent PRs. That also keeps this PR small

runtime/sema/check_transaction_declaration.go
dsainati1: Agreed, but unless I am mistaken because we have `continue` statements in the conditionals that check for name conflicts, we are not actually adding the newly computed `role` type to the transaction's `role` fields at the end of the loop, so we aren't going to typecheck it regardless.

runtime/sema/check_transaction_declaration.go
turbolent: Refactored in ea613440bb9bab17c5cdcf329accb550a6fce97d. Turns out this was also duplicated in the composite declaration checking function

dsainati1:
https://github.com/onflow/cadence/pull/2257#discussion_r1073952338 linking this in case it got missed

runtime/tests/checker/transactions_test.go
```diff
@@ -348,6 +365,206 @@ func TestCheckTransactions(t *testing.T) {
...
+                      let blub: String
+
+                      prepare() {
+                          self.blub = "2"
+                      }
+                  }
+
+                  execute {
+                      let bar: Int = self.foo.bar
+                      let blub: String = self.baz.blub
```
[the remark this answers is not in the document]
turbolent: Good idea! Added in c768ac4257c72e2bb5898eeca2ec4b28496fb088. Required extending the check for non-first class-typed values, from just identifier expressions to member expressions.

turbolent:
Realized there's a problem with resource-kinded fields in roles: https://github.com/onflow/flips/pull/41#discussion_r1080571013.

Does not block this PR, but necessitates a follow-up PR

runtime/sema/check_transaction_declaration.go
turbolent: Even if roles are duplicated or clash with fields, still all AST role declarations are iterated and checked (see https://github.com/onflow/cadence/pull/2257/files/a5ca45aba8ec70e8cb41a4bfe7e9dcbd36c6e581#diff-3ee2ecce28d4ab69f08348e38d2c38de4610023f91f00b769787df1d5a14e07aR73-R75 and in turn `VisitTransactionRoleDeclaration`), which assumes that the transaction role type was defined.

You can check this for yourself by attempting the proposed change, moving the transaction type creation and the association of the AST role declaration with the role type in the elaboration below the continue statements of the duplication/clash checks.

turbolent:
@dsainati1 Sorry I missed that comment, thanks for bumping it 👍

runtime/sema/check_transaction_declaration.go
turbolent: I still improved it a bit in 149085186089642b1f395c39ca7465a422be4c7f by replacing the continue statements with separate checks for duplication and field clashing, and in each case, performing the operation (associating the role if not a duplicate, and declaring a member if no field clash) separately if possible.

runtime/sema/check_transaction_declaration.go
```diff
@@ -248,36 +303,175 @@ func (checker *Checker) declareTransactionDeclaration(declaration *ast.Transacti
...
+					),
+				},
+			)
+		} else {
+			roles.Set(roleName, transactionRoleType)
+		}
+
+		// Ensure roles and fields do not clash
+		if _, ok := members.Get(roleName); ok {
+			field := fieldDeclarationByIdentifier(roleName)
```
SupunS, line 357: Seems the `field` is only used to get the field's Identifier. The same info should be available via `member.Identifier`

SupunS: With that, I think it is not needed to maintain a separate `roles` mapping, given that roles also are added to members, "duplicate roles" also can be checked here along with fields. The reported error can be switched based on the `member.DeclarationKind`

runtime/tests/checker/transactions_test.go
```diff
@@ -398,82 +727,171 @@ func TestCheckInvalidTransactionSelfMoveToFunction(t *testing.T) {
...
+              }
+          }
+        `)
+
+		errs := RequireCheckerErrors(t, err, 1)
+
+		assert.IsType(t, &sema.InvalidMoveError{}, errs[0])
+	})
+
+	t.Run("role in execute", func(t *testing.T) {
```
SupunS, line 878: The `checkValueMove` is only get called only for a limited set of expressions. Maybe add some tests for other use-cases, such as calling a function on a role, etc.

runtime/sema/check_transaction_declaration.go
```diff
@@ -84,6 +94,73 @@ func (checker *Checker) VisitTransactionDeclaration(declaration *ast.Transaction
...
+					ExpectedType: transactionPrepareParameter.TypeAnnotation.Type,
+					ActualType:   transactionRolePrepareParameter.TypeAnnotation.Type,
+					Range: ast.NewRangeFromPositioned(
+						checker.memoryGauge,
+						parameter.TypeAnnotation,
+					),
+				},
+			)
+		}
+	}
```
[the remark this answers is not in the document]
turbolent: An equality check would also suffice to determine if the parameter lists match, but I wanted to provide better diagnostics by reporting errors where they mismatch.

runtime/sema/check_transaction_declaration.go
```diff
@@ -248,36 +303,175 @@ func (checker *Checker) declareTransactionDeclaration(declaration *ast.Transacti
...
+					Identifier:      roleDeclaration.Identifier,
+					DeclarationKind: common.DeclarationKindTransactionRole,
+					VariableKind:    ast.VariableKindConstant,
+					TypeAnnotation:  NewTypeAnnotation(transactionRoleType),
+					DocString:       roleDeclaration.DocString,
+				},
+			)
+		}
+	}
+	transactionType.Roles = roles
```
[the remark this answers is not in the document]
turbolent: Because it is later needed to iterate over all the rules and check if their resource-kinded fields were invalidated. It is true that this information is also available in `transactionType.Members`, but that also contains fields, so needs filtering. I'll refactor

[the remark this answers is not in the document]
turbolent: Refactored in 67c69228316b47dde4f8599674a319bc9154ba99

runtime/sema/check_transaction_declaration.go
turbolent: Good idea! Refactored in 170f96aaa17becb8b72d5f3c0e63748d0b513028

turbolent: And further refactored in 20d50449f2a77150143b620b2cc9e1db60814601

runtime/tests/checker/transactions_test.go
turbolent: Improved the tests in 36f2848657ff94ea58f1770d221a85232b66fb31

runtime/sema/check_transaction_declaration.go
```diff
@@ -248,36 +302,157 @@ func (checker *Checker) declareTransactionDeclaration(declaration *ast.Transacti
...
+		// Ensure roles and fields do not clash
+		roleName := roleDeclaration.Identifier.Identifier
+		if member, ok := members.Get(roleName); ok {
+
+			identifierRange := ast.NewRangeFromPositioned(
+				checker.memoryGauge,
+				roleDeclaration.Identifier,
+			)
+
+			if _, ok := member.TypeAnnotation.Type.(*TransactionRoleType); ok {
```
SupunS, line 335: not really a big deal: we can also check for `member.DeclarationKind == common.DeclarationKindTransactionRole` instead of the type cast?

runtime/tests/checker/transactions_test.go
```diff
@@ -398,82 +727,292 @@ func TestCheckInvalidTransactionSelfMoveToFunction(t *testing.T) {
...
+                  }
+              }
+            `)
+
+			errs := RequireCheckerErrors(t, err, 1)
+
+			assert.IsType(t, &sema.InvalidMoveError{}, errs[0])
+		})
+
+	})
```
SupunS, line 1017: I was trying more ways to break the type checking, and seems the following does not give any checker errors:
```cadence
fun test(_ x: AnyStruct) {}

transaction {

    role buyer {}

    execute {
        test(&self.buyer as &AnyStruct)  // role inside a casting expression
    }
}
```

While it may be harmless, I think we should also reject such cases (thinks it's the same reason as https://github.com/onflow/cadence/pull/2257#discussion_r1083138732).

SupunS (commented):
Looks good! Thanks for the refactoring!

Just have one question regarding the checker/test.

runtime/sema/check_transaction_declaration.go
turbolent: That works too, changed in bb6bebaa137b4c24ccf78c2ac526db9345d565c0

runtime/tests/checker/transactions_test.go
turbolent: We could, but we currently don't reject references to non-first class values anywhere else.

References to roles are not useful, but why would we reject it?

SupunS: > We could, but we currently don't reject references to non-first class values anywhere else.
>
> References to roles are not useful, but why would we reject it?

I think 'not useful / not being able to use anywhere' is a good enough reason to reject it. By statically rejecting, a developer would know that they are writing some wrong code, as early as possible.

Having said that, I also just realized, whether we are rejecting it or not, may also depend on how we 'explain' roles to users: If we are going to say a role is a 'value' inside a transaction, being able to take references is OK I guess. But if we are pitching roles as just a named-scope, then taking a reference to a scope sounds odd (I'm looking at roles from this second perspective so far, hence the suggestion). I mean, we may implement it as a value under the hood, but just wanted to say that the validations may depend on the story we want to tell to the users.

SupunS: anyway, it shouldn't be a blocker for the PR

SupunS (approved):

runtime/tests/checker/transactions_test.go
turbolent: @SupunS Great points! It's easy to add, so lets do it. I'll make the check recursive, unless you think it should not be.

SupunS: should be on creating references?
i.e: roles can only be used in member access expressions - to access its members?

turbolent: Sounds good, I'll add that in a follow-up PR, this PR is already large and there are already two PRs stacked on this one

turbolent: Opened #2268

PR #2259, opened 2023-01-19
Interpret transaction role declarations

turbolent:
Work towards #2177

## Description

Interpret transaction roles. Call each role's prepare block just like the transaction's prepare block before.

______

### DIFF
--- a/runtime/interpreter/interpreter_transaction.go
+++ b/runtime/interpreter/interpreter_transaction.go
@@ -93,7 +93,7 @@ func (interpreter *Interpreter) declareTransactionEntryPoint(declaration *ast.Tr
 
 	transactionFunction := &HostFunctionValue{
 		Function: func(invocation Invocation) Value {
-			interpreter.activations.PushNewWithParent(lexicalScope)
+			transactionScope := interpreter.activations.PushNewWithParent(lexicalScope)
 			defer interpreter.activations.Pop()
 
 			self := MemberAccessibleValue(transactionValue)
@@ -113,10 +113,6 @@ func (interpreter *Interpreter) declareTransactionEntryPoint(declaration *ast.Tr
 				interpreter.bindParameterArguments(declaration.ParameterList, transactionArguments)
 				invocation.Arguments = prepareArguments
 			}
-
-			// NOTE: get current scope instead of using `lexicalScope`,
-			// because current scope has `self` declared
-			transactionScope := interpreter.activations.CurrentOrNew()
 
 			if prepareFunction != nil {
 				prepare := interpreter.functionDeclarationValue(
@@ -219,14 +215,12 @@ func (interpreter *Interpreter) declareTransactionRole(
 
 	roleFunction = &HostFunctionValue{
 		Function: func(invocation Invocation) Value {
-			interpreter.activations.PushNewWithCurrent()
+			transactionRoleScope := interpreter.activations.PushNewWithCurrent()
 			defer interpreter.activations.Pop()
 
 			self := MemberAccessibleValue(roleValue)
 			invocation.Self = &self
 			interpreter.declareVariable(sema.SelfIdentifier, self)
-
-			transactionRoleScope := interpreter.activations.CurrentOrNew()
 
 			if prepareFunction != nil {
 				prepare := interpreter.functionDeclarationValue(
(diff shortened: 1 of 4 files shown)


dsainati1 (commented):
Would be nice to see some tests for the multi-signer use case discussed in the FLIP, where each signer is a different role.

runtime/tests/interpreter/transactions_test.go
```diff
@@ -318,3 +321,121 @@ func TestInterpretTransactions(t *testing.T) {
 		)
 	})
 }
+
+func TestInterpretTransactionRoles(t *testing.T) {
+
+	t.Parallel()
+
+	t.Run("NoPrepareFunction", func(t *testing.T) {
```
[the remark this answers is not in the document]
turbolent: Copy-and-paste mistake. Improved the name here and for the transaction tests in 1fd08112589d9f752be8d6702cf15711dc6af177

turbolent:
Added tests for multiple roles and multiple signers

runtime/interpreter/interpreter_transaction.go
```diff
@@ -148,6 +173,74 @@ func (interpreter *Interpreter) declareTransactionEntryPoint(declaration *ast.Tr
...
+	//
+	// This host function value is an internally created and used function,
+	// and can never be passed around as a value.
+	// Hence, the type is not required.
+
+	common.UseMemory(interpreter, common.HostFunctionValueMemoryUsage)
+
+	prepareFunctionValue = &HostFunctionValue{
+		Function: func(invocation Invocation) Value {
+			interpreter.activations.PushNewWithCurrent()
```
SupunS, line 222: I guess the activation retrieved below (i.e: `interpreter.activations.CurrentOrNew()`) is always the one that is pushed here?
So maybe directly use the returned:
```go
transactionRoleScope := interpreter.activations.PushNewWithCurrent()
```

runtime/interpreter/interpreter_transaction.go
```diff
@@ -148,6 +173,74 @@ func (interpreter *Interpreter) declareTransactionEntryPoint(declaration *ast.Tr
...
+			defer interpreter.activations.Pop()
+
+			self := MemberAccessibleValue(roleValue)
+			invocation.Self = &self
+			interpreter.declareVariable(sema.SelfIdentifier, self)
+
+			transactionRoleScope := interpreter.activations.CurrentOrNew()
+
+			if prepareFunction != nil {
```
SupunS, line 231: Would it be more efficient to early-exit at the top of this function? It might save additional activation push/pop, and self-var declarations.

SupunS (approved):
👌

runtime/interpreter/interpreter_transaction.go
turbolent: I left it in the same format as the interpretation of transactions because it is the same as for transactions. Currently, role blocks only have a prepare, but they'll likely get support for conditions like in transactions. f3f15b714e45bfc9cca9814e5e6e3054230de236 improves naming a bit

runtime/interpreter/interpreter_transaction.go
turbolent: Good point, improved in 55a572f54713dc2abc638ce34b2966ce497f43db

turbolent:
@dsainati1 Could you please have a look? That would unblock getting the PR stack / this feature in. Thanks!

runtime/tests/interpreter/transactions_test.go
```diff
@@ -318,3 +321,345 @@ func TestInterpretTransactions(t *testing.T) {
...
+			`
+              transaction {
+
+                  prepare(signer1: AuthAccount, signer2: AuthAccount) {}
+
+                  role role1 {
+                      let signer1Address: String
+                      let signer2Address: String
+
+                      prepare(signer1: AuthAccount, signer2: AuthAccount) {
```
dsainati1, line 612: Should this be allowed? If each signer should have 1 role, we probably need to enforce that each role has one signer and also there are an equal number of roles to signers.

dsainati1 (approved):
Discussed with Bastian, this will be addressed in a followup PR. For now we can accept this and fix it later.

PR #2262, opened 2023-01-19
Add support for resource-kinded fields in transaction roles

turbolent:
Work towards #2177

## Description

Transaction roles may have resource-kinded fields.

Add dedicated support for allowing resource-kinded fields to get invalidated in a transaction's execute block, even though such invalidations are nested, which are not allowed in other contexts (composite destructors).

______

### DIFF
--- a/runtime/sema/check_transaction_declaration.go
+++ b/runtime/sema/check_transaction_declaration.go
@@ -85,6 +85,14 @@ func (checker *Checker) VisitTransactionDeclaration(declaration *ast.Transaction
 		func() {
 			checker.withResourceFieldInvalidationAllowed(
 				func(expression *ast.MemberExpression) *Member {
+					// In addition to the behaviour in composite destructors,
+					// where a self member may be invalidated,
+					// also allow invalidation of role members
+					memberInfo, _ := checker.Elaboration.MemberExpressionMemberInfo(expression)
+					if _, ok := memberInfo.AccessedType.(*TransactionRoleType); ok {
+						return memberInfo.Member
+					}
+
 					return checker.accessedSelfMember(expression)
 				},
 				func() {
@@ -94,7 +102,11 @@ func (checker *Checker) VisitTransactionDeclaration(declaration *ast.Transaction
 		},
 	)
 
-	checker.checkResourceFieldsInvalidated(transactionType, transactionType.Members)
+	checker.checkResourceFieldsInvalidated(transactionType, members)
+
+	transactionType.Roles.Foreach(func(_ string, roleTye *TransactionRoleType) {
+		checker.checkResourceFieldsInvalidated(roleTye, roleTye.Members)
+	})
 
 	return
 }
--- a/runtime/sema/check_transaction_declaration.go
+++ b/runtime/sema/check_transaction_declaration.go
@@ -83,9 +83,14 @@ func (checker *Checker) VisitTransactionDeclaration(declaration *ast.Transaction
 		declaration.PostConditions,
 		VoidType,
 		func() {
-			checker.withSelfResourceInvalidationAllowed(func() {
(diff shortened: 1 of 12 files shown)


runtime/sema/checker.go
```diff
@@ -1366,32 +1366,13 @@ func (checker *Checker) recordResourceInvalidation(
 		)
 	}
```
turbolent, line 1368: The refactor of this function is probably best viewed in split-mode. Also, I first refactored the function a bit to simplify it in 7abf054ce0c4cadfe092d9736e173e18eadfb87d and to allow customization of the check what is allowed to a function in 48247ab22059162d0d97db8c0ca4ce715c2e1a90.

runtime/sema/check_transaction_declaration.go
```diff
@@ -83,13 +83,30 @@ func (checker *Checker) VisitTransactionDeclaration(declaration *ast.Transaction
...
+				func(expression *ast.MemberExpression) *Member {
+					// In addition to the behaviour in composite destructors,
+					// where a self member may be invalidated,
+					// also allow invalidation of role members
+					memberInfo, _ := checker.Elaboration.MemberExpressionMemberInfo(expression)
+					if _, ok := memberInfo.AccessedType.(*TransactionRoleType); ok {
+						return memberInfo.Member
+					}
+
+					return checker.accessedSelfMember(expression)
```
turbolent, line 96: The default, like in https://github.com/onflow/cadence/pull/2262/files#diff-c34bee2125d8f3673160cc7b3c912540ad811b4622a66c3a2ed6c25fc94e1cb0R176

dsainati1 (approved):

runtime/sema/checker.go
```diff
@@ -104,15 +104,15 @@ type Checker struct {
...
-	allowSelfResourceFieldInvalidation bool
-	inAssignment                       bool
-	inInvocation                       bool
-	inCreate                           bool
-	isChecked                          bool
+	_beforeExtractor                 *BeforeExtractor
+	errors                           []error
+	functionActivations              *FunctionActivations
+	inCondition                      bool
+	resourceFieldInvalidationAllowed func(*ast.MemberExpression) *Member
```
dreamsmasher, line 111: I'd run a field alignment check here, since golang doesn't optimize field ordering. having the bool before a function pointer would add a fair bit of padding

dreamsmasher (approved):

runtime/sema/checker.go
turbolent: Good point! Improved here and elsewhere in 8a955179e8972d5df05de0cdd4023808e1bde5ea

runtime/tests/checker/transactions_test.go
```diff
@@ -661,6 +661,90 @@ func TestCheckTransactionRoles(t *testing.T) {
...
+                  role role1 {
+                      var r: @R
+
+                      prepare() {
+                          self.r <- create R()
+                      }
+                  }
+
+                  execute {
+                      absorb(<-self.role1.r)
```
SupunS, line 685: Can you also add a test to ensure that only the immediate members of a role can be moved, but not nested members?
i.e:
```cadence
absorb(<-self.role1.r.field)  // must give errors
```

turbolent: @SupunS Good idea! Added tests in [a91d406](https://github.com/onflow/cadence/pull/2262/commits/a91d4063d5eb9282966da44d6a7c737de94ac426) 

Also, along the way, improved the error message for the error, which helps users and allows better assertions in tests: [1a90380](https://github.com/onflow/cadence/pull/2262/commits/1a9038051d6850751994804c164c5440c5d58386)

turbolent:
@SupunS pushed some more tests and improvements, could you please have a final look? Thank you!

SupunS (approved):

PR #2282, opened 2023-01-31
Use bitmasks for checking declaration kinds of top-level declarations

dreamsmasher:
## Description

Draft PR to run benchstat and compare performance against master. This is a small microoptimization that switches the checker over to using a bitset instead of a hashset when checking the kinds of top-level declarations.
______

### DIFF
--- a/runtime/sema/checker.go
+++ b/runtime/sema/checker.go
@@ -348,20 +348,20 @@ func (checker *Checker) checkTopLevelDeclarationValidity(declarations []ast.Decl
 		return
 	}
 
-	validDeclarationKinds := map[common.DeclarationKind]bool{}
-
 	validTopLevelDeclarations := validTopLevelDeclarationsHandler(checker.Location)
 	if validTopLevelDeclarations == nil {
 		return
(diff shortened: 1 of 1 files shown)


dreamsmasher:
ah damn, nvm. thought there'd be at least some change over baseline but maybe our test cases don't have enough top-level decls for any meaningful difference.

runtime/sema/checker.go
```diff
@@ -348,20 +348,20 @@ func (checker *Checker) checkTopLevelDeclarationValidity(declarations []ast.Decl
 		return
 	}
 
-	validDeclarationKinds := map[common.DeclarationKind]bool{}
-
 	validTopLevelDeclarations := validTopLevelDeclarationsHandler(checker.Location)
```
turbolent, line 351: Maybe have `ValidTopLevelDeclarationsHandlerFunc` return a mask

turbolent (commented):
Good idea! 

Maybe add a specific benchmark that has e.g. a script and a struct declaration

PR #2284, opened 2023-01-31
Transaction role prepare blocks take one signer each

turbolent:
Work towards #2177

## Description

As clarified in the [FLIP](https://github.com/onflow/flips/pull/41), a prepare block in a transaction role may have one parameter, and each role is passed one signer, sequentially from the full list of all signers.

______

### DIFF
--- a/runtime/sema/check_transaction_declaration.go
+++ b/runtime/sema/check_transaction_declaration.go
@@ -74,6 +74,28 @@ func (checker *Checker) VisitTransactionDeclaration(declaration *ast.Transaction
 		checker.checkRolePrepareParameterList(roleDeclaration, transactionType)
 	}
 
+	transactionPrepareParameterCount := len(transactionType.PrepareParameters)
+
+	// If the transaction has roles and the prepare block has parameters,
+	// then there should be exactly one role for each parameter
+	roleCount := len(declaration.Roles)
+	if roleCount > 0 &&
+		prepareFunctionDeclaration != nil &&
+		transactionPrepareParameterCount > 0 &&
+		roleCount != transactionPrepareParameterCount {
+
+		checker.report(
+			&RoleCountMismatchError{
+				ActualCount:   roleCount,
+				ExpectedCount: transactionPrepareParameterCount,
+				Range: ast.NewRangeFromPositioned(
+					checker.memoryGauge,
+					prepareFunctionDeclaration.FunctionDeclaration.ParameterList,
+				),
+			},
+		)
+	}
+
 	if declaration.PreConditions != nil {
 		checker.visitConditions(*declaration.PreConditions)
 	}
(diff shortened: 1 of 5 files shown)


runtime/sema/check_transaction_declaration.go
```diff
@@ -132,52 +154,46 @@ func (checker *Checker) checkRolePrepareParameterList(
...
-	for prepareParameterIndex := 0; prepareParameterIndex < minPrepareParameterCount; prepareParameterIndex++ {
-		transactionPrepareParameter := transactionType.PrepareParameters[prepareParameterIndex]
-		transactionRolePrepareParameter := transactionRoleType.PrepareParameters[prepareParameterIndex]
-
-		if !transactionRolePrepareParameter.TypeAnnotation.
-			Equal(transactionPrepareParameter.TypeAnnotation) {
-
-			parameter := roleDeclaration.Prepare.FunctionDeclaration.ParameterList.Parameters[prepareParameterIndex]
+	} else {
+		// If the transaction declaration's prepare block has no parameters,
```
dsainati1, line 174:
```suggestion
		// If the transaction declaration's prepare block has parameters,
```

runtime/tests/checker/transactions_test.go
```diff
@@ -552,10 +552,72 @@ func TestCheckTransactionRoles(t *testing.T) {
...
+
+              prepare(first: AuthAccount, second: AuthAccount) {}
+
+              role role1 {
+                  prepare(signer: AuthAccount) {}
+              }
+
+              role role2 {
+                  prepare(signer: AuthAccount) {}
+              }
```
dsainati1, line 616: I still think this feature would benefit from some way to encode in Cadence which of the signers in the outer `prepare` is tied to which role; but that may be beyond the scope of this PR and might just be another discussion point for the FLIP.

dsainati1 (approved):

turbolent:
@SupunS Could you please have a look? It's (currently) the last PR for the feature, and would allow us to get a preview build out. Thanks!

runtime/sema/check_transaction_declaration.go
```diff
@@ -74,6 +74,28 @@ func (checker *Checker) VisitTransactionDeclaration(declaration *ast.Transaction
...
 
+	transactionPrepareParameterCount := len(transactionType.PrepareParameters)
+
+	// If the transaction has roles and the prepare block has parameters,
+	// then there should be exactly one role for each parameter
+	roleCount := len(declaration.Roles)
+	if roleCount > 0 &&
+		prepareFunctionDeclaration != nil &&
+		transactionPrepareParameterCount > 0 &&
+		roleCount != transactionPrepareParameterCount {
```
SupunS, line 85: Do they have to be equal? `roleCount <= transactionPrepareParameterCount` should be also fine?
i.e: There can be accounts used by the transaction's prepare only, but not by any roles.

But maybe this should be also something to discuss on the FLIP

SupunS (approved):

PR #2288, opened 2023-02-01
Make compatibility checker workflow work with old tags

SupunS:
## Description

Instead of checking out the current/base branch/tag, just update the tool to use that particular version of cadence.
That way, the tool always stays the same (old tags don't need to have the tool), and only the cadence version is bumped.
Should have done it this way in the first place.
______

### DIFF
--- a/.github/workflows/compatibility-check-template.yml
+++ b/.github/workflows/compatibility-check-template.yml
@@ -31,7 +31,9 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
-          ref: ${{ inputs.current-branch }}
+          # Checkout the branch where the workflow is available.
+          # Usually this is the master branch.
+          ref: ${{ github.ref_name }}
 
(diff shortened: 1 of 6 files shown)


.github/workflows/compatibility-check-template.yml
```diff
@@ -75,21 +77,22 @@ jobs:
...
         run: |
+          go get github.com/onflow/cadence@${{ inputs.current-branch }}
+          go mod tidy
           go run ./cmd/check_contracts/main.go ../../tmp/contracts.csv ../../tmp/output-new.txt
 
       # Check contracts using base branch
 
-      - name: Checkout base branch
-        run: git checkout ${{ inputs.base-branch }}
-
```
SupunS, line 84: Instead of checking out the old branch, just do a cadence version bump on the tool

turbolent (approved):
Good idea!

dsainati1 (approved):

PR #2290, opened 2023-02-01
Add Vec type internally for abstracting over array operations

dreamsmasher:
## Description

This PR adds a new internal library that exposes a `Vec` API, similar to that of Rust. Golang's built-in array operations are lacking, and leads to code bloat from having to manually repeat patterns like folds, maps, and initialization. 

The `Vec` library uses generics for type safety, and exposes a simple newtype wrapper over a slice. This type is coercible to a slice and should be preferred over go slices for new code. Refactoring all our existing code is possible, but would be a large undertaking.

Examples:

```go
// initialization of array with capacity 16, length 0
array  := make([]Type, 0, 16)
vector := vec.New[Type](16)

// initialization of array containing 16 zeroed elements
array  := make([]Type, 16)
vector := vec.Zeroed[Type](16)

// array literal
array  := []int{1, 2, 3, 4}
vector := vec.New(1, 2, 3, 4)
```

```go
// map a function over an array
var buf []int = ...
f := func(x int) string {...}

// with slices
mapped := make([]string, 0, len(buf))
for i, x := range buf {
  mapped[i] = f(x)
}

// with Vec
var vector Vec[int] = ...
mapped := vec.Map(vector, f)
```

```go
// check if any elements in a slice satisfy a predicate
func canBork(d doggo) bool {...}

func anyBorksBuf(doggos []doggo) bool {
  for _, doge := range doggos {
    if canBork(doge) {
      return true
    }
  }
  return false
}

func anyBorksVec(doggos vec.Vec[doggo]) bool {
  return doggos.Any(canBork)
}
```

______


bluesign:
why not expose those with something like `sliceutils` instead of Vec type?

runtime/common/vec/vec_test.go
```diff
@@ -0,0 +1,333 @@
...
+	ys := Map(xs, func(x int) string {
+		return fmt.Sprint(x)
+	})
+
+	// should be pure
+	require.Equal(t, New(1, 2, 3), xs)
+	require.Equal(t, New("1", "2", "3"), ys)
+}
+
+func TestVecFlatMap(t *testing.T) {
```
SupunS, line 118: How is the performance of these operations compared to using slices? Go generics tend to add some significant overhead. e.g: https://github.com/onflow/cadence/pull/2181#issuecomment-1361824071. 

Might be useful to add some benchmarks too.

SupunS (commented):
What do you think of making this a separate open-source library, so anyone out there can use it? :)

runtime/common/vec/vec_test.go
dreamsmasher: ugh i didn't even realize generics came with performance overhead. that tracks given the lack of type erasure, I'll write some benchmarks to investigate

dreamsmasher: https://planetscale.com/blog/generics-can-make-your-go-code-slower

Interesting, seems like most of the generics overhead actually comes from vtable passing and indirections. The only code that uses parametric polymorphism are the numeric folds, `Sum()` and `Product()`, whereas the other methods are built on top of slice operations. This might hurt inlining but i'll post an update after benchmarking

dreamsmasher: Ok, results aren't promising. Here are the results:

```
BenchmarkVecAll
BenchmarkVecAll/array_any
BenchmarkVecAll/array_any-8             1000000000               0.0000342 ns/op               0 B/op          0 allocs/op
BenchmarkVecAll/vec_any
BenchmarkVecAll/vec_any-8               1000000000               0.0002208 ns/op               0 B/op          0 allocs/op
```

given the following benchmark

```go

func BenchmarkVecAll(b *testing.B) {
	size := 100000
	array := make([]int, size + 1)
	array[size] = 1
	vec := Vec[int](array)
	for i := 0; i < b.N; i++ {

		b.Run("array any", func(b *testing.B) {
			res := false
			for _, n := range array {
				if n != 0 {
					res = true
					break
				}
			}
			require.True(b, res)
		})

		b.Run("vec any", func(b *testing.B) {
			res := vec.Any(func(x int) bool {
				return x != 0
			})
			require.True(b, res)
		})

	}
}
```

It seems that functional patterns incur a ton of overhead due to a lack of inlining at the call site, I'm not sure how much generics factor into this. But the almost 10x slowdown is alarming

bluesign: @dreamsmasher this shouldn't be this big

bluesign:
```
goos: darwin
goarch: arm64
pkg: vextest/vec
BenchmarkVecAll/array_any-10         	1000000000	         0.003678 ns/op
BenchmarkVecAll/vec_any-10           	1000000000	         0.003284 ns/op
PASS
ok  	vextest/vec	0.166s
```

bluesign:
```
goos: linux
goarch: amd64
pkg: vextest/vec
cpu: Common KVM processor
BenchmarkVecAll/array_any         	1000000000	         0.002960 ns/op
BenchmarkVecAll/vec_any           	1000000000	         0.01795 ns/op
PASS
ok  	vextest/vec	0.137s
```

bluesign: I maybe totally missing something as a go noob

j1010001:
causing performance issues (slowing down execution), not worth the code reusability improvements for now.

PR #2296, opened 2023-02-03
Improve code coverage functionality

m-Peter:
Work towards onflow/developer-grants#132

**Complementary PRs:**
- https://github.com/onflow/flow-cli/pull/827
- https://github.com/onflow/cadence-tools/pull/78

## Description

Add support for:
- Inspecting program statements and their lines
- Calculating covered statements percentage
- Tracking covered lines
- Tracking missed lines
- Excluding locations when collecting code coverage
______

```bash
go test --cover -coverprofile=coverage.out -timeout 30s -run '^(TestNewLocationCoverage|TestLocationCoverageAddLineHit|TestLocationCoverageCoveredLines|TestLocationCoverageMissedLines|TestLocationCoveragePercentage|TestNewCoverageReport|TestCoverageReportExcludeLocation|TestCoverageReportInspectProgram|TestCoverageReportInspectProgramForExcludedLocation|TestCoverageReportAddLineHit|TestCoverageReportAddLineHitForExcludedLocation|TestCoverageReportAddLineHitForNonInspectedProgram|TestRuntimeCoverage|TestRuntimeCoverageWithExcludedLocation)$' github.com/onflow/cadence/runtime

go tool cover -func=coverage.out | grep 'coverage.go'

github.com/onflow/cadence/runtime/coverage.go:40:			AddLineHit				100.0%
github.com/onflow/cadence/runtime/coverage.go:47:			Percentage				100.0%
github.com/onflow/cadence/runtime/coverage.go:69:			CoveredLines				100.0%
github.com/onflow/cadence/runtime/coverage.go:82:			MissedLines				100.0%
github.com/onflow/cadence/runtime/coverage.go:95:			NewLocationCoverage			100.0%
github.com/onflow/cadence/runtime/coverage.go:116:			ExcludeLocation				100.0%
github.com/onflow/cadence/runtime/coverage.go:122:			IsLocationExcluded			100.0%
github.com/onflow/cadence/runtime/coverage.go:131:			AddLineHit				100.0%
github.com/onflow/cadence/runtime/coverage.go:148:			InspectProgram				100.0%
github.com/onflow/cadence/runtime/coverage.go:189:			IsProgramInspected			100.0%
github.com/onflow/cadence/runtime/coverage.go:197:			CoveredStatementsPercentage		100.0%
github.com/onflow/cadence/runtime/coverage.go:212:			NewCoverageReport			100.0%
github.com/onflow/cadence/runtime/coverage.go:222:			MarshalJSON				100.0%
```

### DIFF
--- a/runtime/coverage.go
+++ b/runtime/coverage.go
@@ -100,27 +100,20 @@ type CoverageReport struct {
 	// Contains an *ast.Program per location.
 	Programs map[common.Location]*ast.Program `json:"-"`
 	// Contains locations excluded from coverage collection.
-	ExcludedLocations []common.Location `json:"-"`
+	ExcludedLocations map[common.Location]struct{} `json:"-"`
 }
 
 // ExcludeLocation adds the given location to the array of excluded
 // locations.
 func (r *CoverageReport) ExcludeLocation(location Location) {
-	if r.IsLocationExcluded(location) {
-		return
-	}
-	r.ExcludedLocations = append(r.ExcludedLocations, location)
+	r.ExcludedLocations[location] = struct{}{}
 }
 
 // IsLocationExcluded checks whether the given location is excluded
 // or not, from coverage collection.
 func (r *CoverageReport) IsLocationExcluded(location Location) bool {
-	for _, excludedLocation := range r.ExcludedLocations {
-		if excludedLocation == location {
-			return true
-		}
-	}
-	return false
+	_, ok := r.ExcludedLocations[location]
+	return ok
 }
 
 // AddLineHit increments the hit count for the given line, on the given
@@ -210,8 +203,9 @@ func (r *CoverageReport) CoveredStatementsPercentage() string {
 // NewCoverageReport creates and returns a *CoverageReport.
 func NewCoverageReport() *CoverageReport {
 	return &CoverageReport{
-		Coverage: map[common.Location]*LocationCoverage{},
-		Programs: map[common.Location]*ast.Program{},
+		Coverage:          map[common.Location]*LocationCoverage{},
+		Programs:          map[common.Location]*ast.Program{},
+		ExcludedLocations: map[common.Location]struct{}{},
 	}
 }
 
--- a/runtime/coverage.go
+++ b/runtime/coverage.go
@@ -20,56 +20,229 @@ package runtime
 
 import (
 	"encoding/json"
+	"fmt"
+	"sort"
 
+	"github.com/onflow/cadence/runtime/ast"
 	"github.com/onflow/cadence/runtime/common"
 )
 
-// LocationCoverage records coverage information for a location
+// LocationCoverage records coverage information for a location.
 type LocationCoverage struct {
-	LineHits map[int]int `json:"line_hits"`
+	// Contains hit count for each line on a given location.
+	// A hit count of 0 means the line was not covered.
+	LineHits map[int]int
+	// Total number of statements on a given location.
+	Statements int
 }
 
+// AddLineHit increments the hit count for the given line.
 func (c *LocationCoverage) AddLineHit(line int) {
 	c.LineHits[line]++
 }
 
-func NewLocationCoverage() *LocationCoverage {
+// Percentage returns a string representation of the covered
+// statements percentage. It is defined as the ratio of covered
+// lines over the total statements for a given location.
+func (c *LocationCoverage) Percentage() string {
+	coveredLines := c.CoveredLines()
+	if coveredLines > c.Statements {
+		// We saturate the percentage at 100%, when the inspector
+		// fails to correctly count all statements for a given
+		// location.
+		coveredLines = c.Statements
+	}
+	percentage := 100 * float64(coveredLines) / float64(c.Statements)
+	return fmt.Sprintf("%0.1f%%", percentage)
+}
+
+// CoveredLines returns the count of covered lines for a given location.
+// This is the number of lines with a hit count > 0.
+func (c *LocationCoverage) CoveredLines() int {
+	coveredLines := 0
+	for _, hits := range c.LineHits { // nolint:maprange
+		if hits > 0 {
+			coveredLines += 1
+		}
+	}
+	return coveredLines
+}
+
+// MissedLines returns an array with the missed lines for a given location.
+// These are all the lines with a hit count == 0. The resulting array is
+// sorted in ascending order.
+func (c *LocationCoverage) MissedLines() []int {
+	missedLines := make([]int, 0)
+	for line, hits := range c.LineHits { // nolint:maprange
+		if hits == 0 {
+			missedLines = append(missedLines, line)
+		}
+	}
+	sort.Ints(missedLines)
+	return missedLines
+}
+
+// NewLocationCoverage creates and returns a *LocationCoverage with the
+// given lineHits map.
+func NewLocationCoverage(lineHits map[int]int) *LocationCoverage {
 	return &LocationCoverage{
-		LineHits: map[int]int{},
+		LineHits:   lineHits,
+		Statements: len(lineHits),
 	}
 }
 
-// CoverageReport is a collection of coverage per location
+// CoverageReport collects coverage information per location.
+// It keeps track of inspected programs per location, and can
+// also exclude locations from coverage collection.
 type CoverageReport struct {
+	// Contains a *LocationCoverage per location.
 	Coverage map[common.Location]*LocationCoverage `json:"-"`
+	// Contains an *ast.Program per location.
+	Programs map[common.Location]*ast.Program `json:"-"`
+	// Contains locations excluded from coverage collection.
+	ExcludedLocations []common.Location `json:"-"`
 }
 
-func (r *CoverageReport) AddLineHit(location Location, line int) {
-	locationCoverage := r.Coverage[location]
-	if locationCoverage == nil {
-		locationCoverage = NewLocationCoverage()
-		r.Coverage[location] = locationCoverage
+// ExcludeLocation adds the given location to the array of excluded
+// locations.
+func (r *CoverageReport) ExcludeLocation(location Location) {
+	if r.IsLocationExcluded(location) {
+		return
 	}
+	r.ExcludedLocations = append(r.ExcludedLocations, location)
+}
+
+// IsLocationExcluded checks whether the given location is excluded
+// or not, from coverage collection.
+func (r *CoverageReport) IsLocationExcluded(location Location) bool {
+	for _, excludedLocation := range r.ExcludedLocations {
+		if excludedLocation == location {
+			return true
+		}
+	}
+	return false
+}
+
+// AddLineHit increments the hit count for the given line, on the given
+// location. The method call is a NO-OP in two cases:
+// - If the location is excluded from coverage collection
+// - If the location's *ast.Program, has not been inspected
+func (r *CoverageReport) AddLineHit(location Location, line int) {
+	if r.IsLocationExcluded(location) {
+		return
+	}
+
+	if !r.IsProgramInspected(location) {
+		return
+	}
+
+	locationCoverage := r.Coverage[location]
 	locationCoverage.AddLineHit(line)
 }
 
+// InspectProgram inspects the elements of the given *ast.Program, and counts its
+// statements. If inspection is successful, the *ast.Program is marked as inspected.
+// If the given location is excluded from coverage collection, the method call
+// results in a NO-OP.
+func (r *CoverageReport) InspectProgram(location Location, program *ast.Program) {
+	if r.IsLocationExcluded(location) {
+		return
+	}
+	r.Programs[location] = program
+	lineHits := make(map[int]int, 0)
+	var depth int
+
+	inspector := ast.NewInspector(program)
+	inspector.Elements(
+		nil, func(element ast.Element, push bool) bool {
+			if push {
+				depth++
+
+				_, isStatement := element.(ast.Statement)
+				_, isDeclaration := element.(ast.Declaration)
+				_, isVariableDeclaration := element.(*ast.VariableDeclaration)
+
+				// Track only the statements that are not declarations, such as:
+				// - *ast.CompositeDeclaration
+				// - *ast.SpecialFunctionDeclaration
+				// - *ast.FunctionDeclaration
+				// However, also track local (i.e. non-top level) variable declarations.
+				if (isStatement && !isDeclaration) ||
+					(isVariableDeclaration && depth > 2) {
+					line := element.StartPosition().Line
+					lineHits[line] = 0
+				}
+			} else {
+				depth--
+			}
+
+			return true
+		})
+
+	locationCoverage := NewLocationCoverage(lineHits)
+	r.Coverage[location] = locationCoverage
+}
+
+// IsProgramInspected checks whether the *ast.Program on the given
+// location, has been inspected or not.
+func (r *CoverageReport) IsProgramInspected(location Location) bool {
+	_, isInspected := r.Programs[location]
+	return isInspected
+}
+
+// CoveredStatementsPercentage returns a string representation of
+// the covered statements percentage. It is defined as the ratio
+// of total covered lines over total statements, for all locations.
+func (r *CoverageReport) CoveredStatementsPercentage() string {
+	totalStatements := 0
+	totalCoveredLines := 0
+	for _, locationCoverage := range r.Coverage { // nolint:maprange
+		totalStatements += locationCoverage.Statements
+		totalCoveredLines += locationCoverage.CoveredLines()
+	}
+	percentage := fmt.Sprintf(
+		"%0.1f%%",
+		100*float64(totalCoveredLines)/float64(totalStatements),
+	)
+	return fmt.Sprintf("Coverage: %v of statements", percentage)
+}
+
+// NewCoverageReport creates and returns a *CoverageReport.
 func NewCoverageReport() *CoverageReport {
 	return &CoverageReport{
 		Coverage: map[common.Location]*LocationCoverage{},
+		Programs: map[common.Location]*ast.Program{},
 	}
 }
 
+// MarshalJSON serializes each common.Location/*LocationCoverage
+// key/value pair on the *CoverageReport.Coverage map.
 func (r *CoverageReport) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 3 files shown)


turbolent:
Great work! Thank you for improving this feature and contributing 👍

runtime/coverage_test.go
```diff
@@ -28,8 +28,280 @@ import (
...
+
+	locationCoverage.AddLineHit(3)
+	locationCoverage.AddLineHit(4)
+	locationCoverage.AddLineHit(5)
+	locationCoverage.AddLineHit(15)
+
+	// One side effect of counting line hits that are not included
+	// in the statementLines array, is a percentage > 100.
+	// TBD
+	assert.Equal(t, "133.3%", locationCoverage.Percentage)
```
turbolent, line 140: Maybe just saturate at 100%?

runtime/environment.go
```diff
@@ -620,8 +620,21 @@ func (e *interpreterEnvironment) newOnStatementHandler() interpreter.OnStatement
 
 	return func(inter *interpreter.Interpreter, statement ast.Statement) {
 		location := inter.Location
+		if !e.coverageReport.IsProgramInspected(location) {
+			program := inter.Program.Program
+			err := e.coverageReport.InspectProgram(location, program)
+			if err != nil {
+				// TODO: Log something
+				return
+			}
```
turbolent, line 629: `InspectProgram` always returns `nil`, so maybe just remove the error return value from `InspectProgram` and then there is no need for error handling here

runtime/coverage.go
```diff
@@ -20,56 +20,179 @@ package runtime
...
+	LineHits   map[int]int
+	Statements int
+	Percentage string
 }
 
 func (c *LocationCoverage) AddLineHit(line int) {
 	c.LineHits[line]++
+	coveredLines := c.CoveredLines()
+	percentage := 100 * float64(coveredLines) / float64(c.Statements)
+	c.Percentage = fmt.Sprintf("%0.1f%%", percentage)
```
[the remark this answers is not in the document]
m-Peter: Good point. I encountered the same dilemma with the tracking of missed lines, and in the end I went with the method approach. Fixed in https://github.com/onflow/cadence/pull/2296/commits/de9dcd5af2ae13b5015b6565b77a49bb5992f96f

runtime/coverage_test.go
m-Peter: That makes sense! Fixed in https://github.com/onflow/cadence/pull/2296/commits/4bc05a38b54b77a7b76ff78c3db9968856f62051

runtime/coverage.go
```diff
@@ -20,56 +20,179 @@ package runtime
...
+	return nil
 }
 
-func (r *CoverageReport) AddLineHit(location Location, line int) {
+func (r *CoverageReport) InspectProgram(location Location, program *ast.Program) error {
+	if r.IsLocationExcluded(location) {
+		return fmt.Errorf("location is excluded from coverage: %v", location)
+	}
+	r.Programs[location] = program
+	var statements []int
```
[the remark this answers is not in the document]
m-Peter: That is indeed more straightforward :rocket: . Fixed in https://github.com/onflow/cadence/pull/2296/commits/45c587a076073103cac86fd10f06f62bd340b776

runtime/coverage.go
```diff
@@ -20,56 +20,179 @@ package runtime
...
+		if hits > 0 {
+			coveredLines += 1
+		}
+	}
+	return coveredLines
+}
+
+func (c *LocationCoverage) MissedLines() []int {
+	missedLines := make([]int, 0)
+	for line, hits := range c.LineHits { // nolint:maprange
```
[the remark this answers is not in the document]
m-Peter: Good catch :100: Fixed in https://github.com/onflow/cadence/pull/2296/commits/04adad648d14bded771bf438f5bcb93f46c82054

runtime/environment.go
m-Peter: Hmm, maybe I'm missing something here, but:
```go
if r.IsLocationExcluded(location) {
	return fmt.Errorf("location is excluded from coverage: %v", location)
}
```
`InspectProgram` will return early if the `location` is excluded. Side note: one use for excluding locations, is to not generate coverage for Cadence test scripts.

I could move this check in `runtime/environment.go`:

```go
return func(inter *interpreter.Interpreter, statement ast.Statement) {
	location := inter.Location
	if !e.coverageReport.IsProgramInspected(location) {
		program := inter.Program.Program
		err := e.coverageReport.InspectProgram(location, program)
		if err != nil {
			// TODO: Log something
			return
		}
	}

	...
}
```

However, I thought that it is best if the coverage package accounts for this.

runtime/coverage.go
```diff
@@ -20,56 +20,179 @@ package runtime
...
+			(*ast.BreakStatement)(nil),
+			(*ast.ContinueStatement)(nil),
+			(*ast.IfStatement)(nil),
+			(*ast.WhileStatement)(nil),
+			(*ast.ForStatement)(nil),
+			(*ast.EmitStatement)(nil),
+			(*ast.AssignmentStatement)(nil),
+			(*ast.SwapStatement)(nil),
+			(*ast.ExpressionStatement)(nil),
+			(*ast.VariableDeclaration)(nil),
```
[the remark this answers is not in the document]
m-Peter: There is indeed a reason for filtering specific AST element types, but correct if I am wrong.

The [newOnStatementHandler](https://github.com/onflow/cadence/blob/master/runtime/environment.go#L613-L618) is not called with all `ast.Statement` elements found on a program/location.

Hence, we will often have missed lines for statements that are not meant to be executed.

I have created a [gist](https://gist.github.com/m-Peter/980bbe50f235b89e89a027f8a2ec9475) which attempts to show what  I explained above.

The `"missed_lines": [1, 4, 13, 17]` from the `coverage.json` file, correspond to the following `ast.Statement`:
- ElementType: `ElementTypeCompositeDeclaration`, Statement: `*ast.CompositeDeclaration`
- ElementType: `ElementTypeSpecialFunctionDeclaration`, Statement: `*ast.SpecialFunctionDeclaration`
- ElementType: `ElementTypeFunctionDeclaration`, Statement: `*ast.FunctionDeclaration`
- ElementType: `ElementTypeFunctionDeclaration`, Statement: `*ast.FunctionDeclaration`

Is there any other way around this?

m-Peter:
> Great work! Thank you for improving this feature and contributing +1

Your comment is much appreciated :bow: 

I have addressed/answered all of the comments.
In addition, I've opened and linked the complementary PRs, to assist with the review process. I hope it will give a better perspective to the reviewers, about the end result.
Looking forward to make this feature available to the developer community :pray:

runtime/coverage.go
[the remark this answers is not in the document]
turbolent: Thank you for the detailed explanation, it totally makes sense now what the "problem" is: some `ast.Statement` are `ast.Declaration`, e.g `ast.FunctionDeclaration`, because they may appear as a "statement" inside of a function body – but we do not expect all of them to be executed/interpreted (and get a statement handler callback), but only some (e.g. `ast.VariableDeclaration`).

Manually specifying this particular set here is brittle, as additional statements might get added in the future, which means this set has to be maintained.

It might be better to look for any `ast.Statement`, so that all statements are handled, but also filter ones that are `ast.Declaration` – except for one  (for now) particular declaration, which is executed/interpreted, `VariableDeclaration`.

runtime/environment.go
turbolent: If a location is excluded from being inspected, is that really an error? I assumed `InspectProgram` can return early in that case.

runtime/coverage.go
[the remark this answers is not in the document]
m-Peter: Now it's quite clear :pray: I also observed this with some trial & error. For example:

```cadence
pub let specialNumbers: {Int: String} = {
    1729: "Harshad",
    8128: "Harmonic",
    41041: "Carmichael"
}
```
would not be executed/interpreted, even though it is an `ast.VariableDeclaration`. That's because it is declared on the top-level.

However,

```cadence
for input in testInputs.keys {
    let result = getIntegerTrait(input)
    ...
}
```

this would be executed/interpreted, because it is declared inside the `main()` function.

I have updated the the logic of the `InspectProgram` method here: https://github.com/onflow/cadence/pull/2296/commits/4cc2c02e21bb7491ca708d4b75c03579a8e86c87.
Let me know if I got the logic straight, and if there's a more elegant way to track which `ast.VariableDeclaration` are locally declared.

runtime/environment.go
m-Peter: Fair point, it's not really an error. I have updated the methods to perform a NO-OP when a location is excluded. See https://github.com/onflow/cadence/pull/2296/commits/4cc2c02e21bb7491ca708d4b75c03579a8e86c87

runtime/coverage.go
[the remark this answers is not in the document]
turbolent: Nice refactor, the `isStatement && !isDeclaration` logic is a bit more future proof 👍 

Regarding detecting non-top level variable declarations: Using the start position of the variable declaration does not handle a program which does not use indentation (bad style, but still valid). 

You can improve this by using the `push` parameter to keep a depth variable. At depth == 2 (program + variable declaration), you know you are inspecting a top-level declaration vs a local variable declaration:

```go
func TestInspectorNonTopLevelVariableDeclarations(t *testing.T) {

	t.Parallel()

	const code = `
      let x = 1

      fun f() {
	      let y = 2
      }

      let z = 3
    `

	program, err := parser.ParseProgram(
		nil,
		[]byte(code),
		parser.Config{},
	)
	require.NoError(t, err)

	var depth int
	var topLevelVariables int
	var localVariables int

	inspector := ast.NewInspector(program)
	inspector.Elements(
		nil,
		func(element ast.Element, push bool) bool {
			if push {
				depth++

				if _, ok := element.(*ast.VariableDeclaration); ok {
					// 2: program + variable declaration
					atTopLevel := depth == 2

					if atTopLevel {
						topLevelVariables += 1
					} else {
						localVariables += 1
					}
				}
			} else {
				depth--
			}

			return true
		},
	)

	assert.Equal(t, 2, topLevelVariables)
	assert.Equal(t, 1, localVariables)
}
```

runtime/coverage.go
```diff
@@ -20,56 +20,237 @@ package runtime
...
+					}
+					if isLocalVariable {
+						line := element.StartPosition().Line
+						lineHits[line] = 0
+					}
+				}
+			}
+			return true
+		})
+
```
turbolent, line 189:
```suggestion
	var depth int

	inspector := ast.NewInspector(program)
	inspector.Elements(
		nil,
		func(element ast.Element, push bool) bool {
			if push {
				depth++

				_, isStatement := element.(ast.Statement)
				_, isDeclaration := element.(ast.Declaration)
				_, isVariableDeclaration := element.(*ast.VariableDeclaration)

				// Track only the statements that are not declarations, such as:
				// - *ast.CompositeDeclaration
				// - *ast.SpecialFunctionDeclaration
				// - *ast.FunctionDeclaration
				//
				// However, we do care for local variable declarations,
				// such as inside a function's body.

				if (isStatement && !isDeclaration) || 
					(isVariableDeclaration && depth > 2) {

					line := element.StartPosition().Line
					lineHits[line] = 0
				}
			} else {
				depth--
			}

			return true
		},
	)
```

runtime/coverage.go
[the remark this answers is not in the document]
m-Peter: Arigato, sensei :bow:
That was a perfect explanation on the defects of using the start position :100:
Not sure if it would cross my mind :sweat_smile:

runtime/coverage.go
turbolent: Nice!

runtime/coverage.go
```diff
@@ -20,56 +20,228 @@ package runtime
...
+				_, isStatement := element.(ast.Statement)
+				_, isDeclaration := element.(ast.Declaration)
+				_, isVariableDeclaration := element.(*ast.VariableDeclaration)
+
+				// Track only the statements that are not declarations, such as:
+				// - *ast.CompositeDeclaration
+				// - *ast.SpecialFunctionDeclaration
+				// - *ast.FunctionDeclaration
+				if (isStatement && !isDeclaration) ||
+					(isVariableDeclaration && depth > 2) {
```
turbolent, line 170:
```suggestion
				// - *ast.FunctionDeclaration
				// However, also track local (i.e. non-top level) variable declarations
				if (isStatement && !isDeclaration) ||
					(isVariableDeclaration && depth > 2) {

```

m-Peter: Missed that one :sweat_smile: Added in https://github.com/onflow/cadence/pull/2296/commits/84b077202f093176e82931cffe2d1b2e19f0246a#diff-dd50d117359d58b9ec20ba8e56dc4b6b224a84ca819d943436617baec41d34d5R169

turbolent (approved):
Great work! 👏 

@dsainati1 @SupunS could you please also have a look?

m-Peter:
I have updated the supplementary PRs as well, they are out of the draft status now, and ready for review.

For the record, the process I followed for building & testing this end to end, looks something like this:

```bash
# `flow-cli` uses this specific version
git checkout tags/v0.31.0 -b v0.31.0-branch

Apply changes to the three corresponding files:
- runtime/coverage.go
- runtime/coverage_test.go
- runtime/environment.go

cadence-tools/test/go.mod (Build-Squad:enable-coverage-for-test-runner branch):
- replace github.com/onflow/cadence => relative path to local cadence repo

flow-cli/go.mod (Build-Squad:test-command-add-coverage-flags branch):
- replace github.com/onflow/cadence => relative path to local cadence repo
- replace github.com/onflow/cadence-tools/test => relative path to local cadence-tools/test repo

flow-cli/pkg/flowkit/go.mod (Build-Squad:test-command-add-coverage-flags branch):
- replace github.com/onflow/cadence => relative path to local cadence repo
- replace github.com/onflow/cadence-tools/test => relative path to local cadence-tools/test repo
```

Although it works, not sure if this is the best way to go about it :sweat_smile:

turbolent:
@onflow/cadence could you please also have a look?

runtime/coverage.go
```diff
@@ -20,56 +20,229 @@ package runtime
...
+// IsLocationExcluded checks whether the given location is excluded
+// or not, from coverage collection.
+func (r *CoverageReport) IsLocationExcluded(location Location) bool {
+	for _, excludedLocation := range r.ExcludedLocations {
+		if excludedLocation == location {
+			return true
+		}
+	}
+	return false
+}
```
SupunS, line 124: Given that `ExcludedLocations` is used only to perform this 'existence' check, maybe use a `map[common.Location]bool` for `ExcludedLocations` instead of an array. Then this lookup would be a lot simpler and faster.
e.g:
```go
func (r *CoverageReport) IsLocationExcluded(location Location) bool {
    return r.ExcludedLocations[location]
} 
```

runtime/coverage.go
```diff
@@ -20,56 +20,229 @@ package runtime
...
+// Percentage returns a string representation of the covered
+// statements percentage. It is defined as the ratio of covered
+// lines over the total statements for a given location.
+func (c *LocationCoverage) Percentage() string {
+	coveredLines := c.CoveredLines()
+	if coveredLines > c.Statements {
+		// We saturate the percentage at 100%, when the inspector
+		// fails to correctly count all statements for a given
+		// location.
+		coveredLines = c.Statements
```
SupunS, line 53: I was wondering how it can go beyond 100% threshold... Is it because we filter a certain set of statements when we initialize the `LocationCoverage`? i.e: here: https://github.com/onflow/cadence/blob/84b077202f093176e82931cffe2d1b2e19f0246a/runtime/coverage.go#L170-L174

and then when marking a line as covered, we don't filter them out:
https://github.com/onflow/cadence/blob/84b077202f093176e82931cffe2d1b2e19f0246a/runtime/coverage.go#L126-L130

What if we add a check to see if a given line is also excluded from the coverage (other than the location and program)? something like:
```go
func (c *LocationCoverage) AddLineHit(line int) {
	if _, included := c.LineHits[line]; included {
		c.LineHits[line]++
	}
}
```

Would that solve it?

SupunS (commented):
Nice!

runtime/coverage.go
turbolent: 👍 

Even further, maybe just use `map[common.Location]struct{}`, which is effectively the idiomatic way to express a set in Go.

runtime/coverage.go
turbolent: It seems like if we get line hits (`AddLineHit`), then those lines should have also been found / included by `InspectProgram`?

SupunS: Yes, that's my understanding. I think it should always hold that informal agreement when `AddLineHit` is invoked by the interpreter. Maybe this is just a problem for tests (when `AddLineHit` is called from outside/manually)

runtime/coverage.go
m-Peter: Great suggestion, it certainly saved a few lines :100:
Fixed in https://github.com/onflow/cadence/pull/2296/commits/a3d875ede68bcef2b210bfc5b648e94cc0dfaf68

runtime/coverage.go
m-Peter: My reasoning here is that the ground truth is the [newOnStatementHandler](https://github.com/onflow/cadence/blob/master/runtime/environment.go#L621-L625), which means that every call of `AddLineHit` from the callback, should be respected. The `InspectProgram` may have failed, for whatever reason, to find that specific line. But this is a good insight to solidify its implementation, and debug the inspection failure. After @turbolent suggestion's, the implementation of `InspectProgram` seems more bullet proof, than my original version, which means that the case I described above maybe does not hold anymore, and it can just manifest in tests.
How do you think we should proceed with this?

SupunS: hmm yeah, I too think that it'll never go into this condition. Thanks for the explanation.
Maybe just add a note/comment mentioning this reasoning and that ideally, it would never reach there. Then we can just leave it as is (as a fail-safe mechanism)?

m-Peter: Good point :+1: 
Updated in https://github.com/onflow/cadence/pull/2296/commits/cd4f4df3fa20b841afef4cf477f28d0508244555

SupunS (approved):
LGTM!

PR #2302, opened 2023-02-04
Fix docstring parsing for functions

turbolent:
## Description

Currently, docstrings for member declarations after functions are not retained correctly when the function does not have a return type annotation or body, because the parsing function always skips trivia (whitespace and comments).

Fix this by performing a lookahead for the optional return type start, and for the (potentially) optional block by performing a look.

______

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #2303, opened 2023-02-04
Retain docstrings for special function declarations, remove unnecessary return type annotations

turbolent:
## Description

While fixing missing docstrings for special function declarations (e.g. initializers and destructors), just like for functions in #2302, I noticed the special function declarations where parsed differently from normal function declarations.

Instead of duplicating the fix for special functions, use the parsing function for parameters lists / optional return type annotations / (potentially) optional blocks used for normal functions, which was already fixed in #2303. 

This resolves an old TODO, improves the error message produced when a special function declaration has a return type annotation, and reduces allocations.

______


runtime/parser/declaration.go
```diff
@@ -1410,6 +1397,21 @@ func parseSpecialFunctionDeclaration(
 		declarationKind = common.DeclarationKindPrepare
 	}
 
+	if returnTypeAnnotation != nil {
```
dreamsmasher, line 1400: Do we want to explicitly disallow return type annotations for special functions? I don't think it hurts readability to add a `: Void`, especially if `Void` return annotations are used elsewhere in a codebase.

dreamsmasher (approved):

runtime/parser/declaration.go
turbolent: We already currently reject explicit type annotations for special functions, but do so implicitly by not parsing any / looking for one: The old parsing code only parses the parameter list and the (potentially) optional body.

we could allow an explicit `Void` return type annotation in the future, but I do not see any value in it. A user learning the language might assume that it could return something other than `Void`, which is not supported or makes sense for any of the special functions (`init`, `destroy`, etc.)

dreamsmasher: Hmm, seems that we should move this validation to the checker instead then. Even if a user assumes special functions can have different return types, they'll immediately get feedback in the form of a type error (which we can specialize to that scenario too, e.g. `Wrong type for init() function, expected Void`). It's not like special functions have any other signifiers other than the docs anyway

SupunS (approved):

dsainati1 (approved):

PR #2307, opened 2023-02-06
Handle cyclic imports in analyzer tool

SupunS:
## Description

Handles cyclic imports in the analyzer tool.

However, the analyzer tool resides within the cadence package, so when the compatibility checker switch to an older version of cadence, it won't have this fix. So the compatibility checker would only be able to handle this from the next release onwards. One solution would be to convert the analysis package to its own repo, but don't think it is worth the hassle.

______


turbolent (approved):
Nice fix!

SupunS:
@dsainati1 @dreamsmasher Can you please have a 👀?

Need to include this for the release.

dsainati1 (approved):

PR #2308, opened 2023-02-06
Reject cyclic imports during contract deployment

### LINKED ISSUE #2304: Reject contracts with cyclic dependencies during deployment
### Issue to be solved

Currently, cyclic imports are not rejected during contract updates, but are only rejected once the contract is deployed and is imported/used by another program (contract/script/transaction).

For e.g: Assume the is a `Bar` contract which imports `Foo`
```cadence
// foo contract
contract Foo {}

// bar contract
import Foo from 0x1234
contract Bar {}
```

It is possible to update `Foo` contract to also import `Bar` creating an cyclic import.
```cadence
// foo contract
import Bar from 0x1234
contract Foo {}

// bar contract
import Foo from 0x1234
contract Bar {}
```

However, cyclic imports are validated and rejected when a third-party program imports any of these contracts, making both of them unusable.

### Suggested Solution

It is preferable to reject the creation of cyclic imports, rather than waiting until the usage, so the developers would know early on.

SupunS:
Closes #2304

## Description

Currently, cyclic imports are not rejected during contract updates, but are only rejected once the contract is deployed and is imported/used by another program (contract/script/transaction).

This PR rejects creating cyclic contracts early on, at the time of deployment. i.e: Fails early.

______

### DIFF
--- a/runtime/environment.go
+++ b/runtime/environment.go
@@ -336,7 +336,11 @@ func (e *interpreterEnvironment) ParseAndCheckProgram(
 		code,
 		location,
 		storeProgram,
-		importResolutionResults{},
+		importResolutionResults{
+			// Current program is already in check.
+			// So mark it also as 'already seen'.
+			location: true,
+		},
 	)
 }
 
--- a/runtime/import_test.go
+++ b/runtime/import_test.go
@@ -25,6 +25,8 @@ import (
 
 	"github.com/stretchr/testify/require"
 
+	"github.com/onflow/cadence"
+	"github.com/onflow/cadence/encoding/json"
 	"github.com/onflow/cadence/runtime/common"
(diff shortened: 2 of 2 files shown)


turbolent (approved):

runtime/environment.go
```diff
@@ -336,7 +336,11 @@ func (e *interpreterEnvironment) ParseAndCheckProgram(
 		code,
 		location,
 		storeProgram,
-		importResolutionResults{},
+		importResolutionResults{
+			// Current program is already in check.
+			// So mark it also as 'already seen'.
+			location: true,
+		},
```
turbolent, line 343: Nice! Love how simple the fix is 👍

dreamsmasher (approved):

PR #2326, opened 2023-02-10
Cache external Cadence type IDs

### LINKED ISSUE #2325: Cache type ID of Cadence external types
### Issue to be solved

There is a speed and memory penalty from not caching type ID of Cadence external types.

### Suggested Solution

Speed and memory can be improved by caching type ID of Cadence external types on their first use.

### Special Thanks to Supun 👍 

While working on CCF specs and CCF codec, I wanted to cache type ID of Cadence external types.  However, none of the several approaches (that were fast enough) avoided breaking existing Cadence tests.

@SupunS responded to my Slack message asking the Cadence team about tests getting broken if we add a cache.  We discussed various ways and decided to proceed with Supun's approach.  Breaking tests couldn't be avoided (with efficient approaches) but @SupunS offered to refactor tests that would get broken by this!

fxamacker:
## Description

Cache external Cadence type IDs on the first use of ID() to avoid performance penalty.

Add utility test functions to cache type IDs.

This commit breaks some existing tests.

Fix all JSON-Cadence Data Interchange encoding tests broken by this commit.

Other tests broken by this commit will need to be refactored in subsequent commits.  They can be refactored to use utility functions (provided by this commit) that recursively caches type IDs.

Closes #2325 

______

### DIFF
--- a/types.go
+++ b/types.go
@@ -1056,7 +1056,7 @@ func (*StructType) isType() {}
 
 func (t *StructType) ID() string {
 	if len(t.typeID) == 0 {
-		t.typeID = generateID(t.Location, t.QualifiedIdentifier)
+		t.typeID = generateTypeID(t.Location, t.QualifiedIdentifier)
 	}
 	return t.typeID
 }
@@ -1132,7 +1132,7 @@ func (*ResourceType) isType() {}
 
 func (t *ResourceType) ID() string {
 	if len(t.typeID) == 0 {
-		t.typeID = generateID(t.Location, t.QualifiedIdentifier)
+		t.typeID = generateTypeID(t.Location, t.QualifiedIdentifier)
 	}
 	return t.typeID
 }
@@ -1208,7 +1208,7 @@ func (*EventType) isType() {}
 
 func (t *EventType) ID() string {
 	if len(t.typeID) == 0 {
-		t.typeID = generateID(t.Location, t.QualifiedIdentifier)
+		t.typeID = generateTypeID(t.Location, t.QualifiedIdentifier)
 	}
 	return t.typeID
 }
@@ -1284,7 +1284,7 @@ func (*ContractType) isType() {}
 
 func (t *ContractType) ID() string {
 	if len(t.typeID) == 0 {
-		t.typeID = generateID(t.Location, t.QualifiedIdentifier)
+		t.typeID = generateTypeID(t.Location, t.QualifiedIdentifier)
 	}
 	return t.typeID
 }
@@ -1372,7 +1372,7 @@ func (*StructInterfaceType) isType() {}
 
 func (t *StructInterfaceType) ID() string {
 	if len(t.typeID) == 0 {
-		t.typeID = generateID(t.Location, t.QualifiedIdentifier)
+		t.typeID = generateTypeID(t.Location, t.QualifiedIdentifier)
 	}
 	return t.typeID
 }
@@ -1448,7 +1448,7 @@ func (*ResourceInterfaceType) isType() {}
 
 func (t *ResourceInterfaceType) ID() string {
 	if len(t.typeID) == 0 {
-		t.typeID = generateID(t.Location, t.QualifiedIdentifier)
+		t.typeID = generateTypeID(t.Location, t.QualifiedIdentifier)
 	}
 	return t.typeID
 }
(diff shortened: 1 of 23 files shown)


fxamacker:
Hey @SupunS, nice work!  Thanks again for [suggesting a great approach and generously offering to fix tests that would get broken](https://github.com/onflow/cadence/issues/2325)!  :+1:

types.go
```diff
@@ -2125,3 +2150,41 @@ func (AccountKeyType) ID() string {
...
+		return identifier
+	}
+
+	return string(location.TypeID(nil, identifier))
+}
+
+// TypeWithCachedTypeID recursively caches type ID of type t.
+// This is needed because each type ID is lazily cached on
+// its first use in ID() to avoid performance penalty.
+func TypeWithCachedTypeID(t Type) Type {
```
SupunS, line 2165: _Note: This method is only used for tests, but is added here to avoid circular dependencies._

values.go
```diff
@@ -1974,3 +1974,63 @@ func (v Function) String() string {
 	// TODO: include function type
 	return format.Function("(...)")
 }
+
+// ValueWithCachedTypeID recursively caches type ID of value v's type.
+// This is needed because each type ID is lazily cached on
+// its first use in ID() to avoid performance penalty.
+func ValueWithCachedTypeID[T Value](value T) T {
```
SupunS, line 1981: _Note: This method is only used for tests, but is added here to avoid circular dependencies._

SupunS (approved):

types.go
```diff
@@ -125,7 +126,10 @@ func NewMeteredOptionalType(gauge common.MemoryGauge, typ Type) OptionalType {
 func (OptionalType) isType() {}
 
 func (t OptionalType) ID() string {
-	return fmt.Sprintf("%s?", t.Type.ID())
+	if len(t.typeID) == 0 {
+		t.typeID = fmt.Sprintf("%s?", t.Type.ID())
```
turbolent, line 130: Here, and for other non-pointer receiver `ID` functions: The receiver is mutated here, but because it is not a pointer, this mutation is "lost" after the function returns.

To effectively cache IDs, we'll likely need to turn all current non-pointer types conforming to `cadence.Type` to pointer types.

types.go
```diff
@@ -1636,11 +1653,14 @@ func NewMeteredReferenceType(
...
 
 func (t ReferenceType) ID() string {
-	id := fmt.Sprintf("&%s", t.Type.ID())
-	if t.Authorized {
-		id = "auth" + id
+	if len(t.typeID) == 0 {
+		id := fmt.Sprintf("&%s", t.Type.ID())
+		if t.Authorized {
+			id = "auth" + id
+		}
```
turbolent, line 1660:
```suggestion
		var prefix string
		if t.Authorized {
			prefix = "auth"
		}
		id := fmt.Sprintf("%s&%s", prefix, t.Type.ID())
```

types.go
```diff
@@ -2125,3 +2150,41 @@ func (AccountKeyType) ID() string {
 func (t AccountKeyType) Equal(other Type) bool {
 	return t == other
 }
+
+func generateID(location common.Location, identifier string) string {
```
turbolent, line 2154: Maybe make the name more specific, e.g. `generateTypeID`

turbolent (approved):
Looks good!

I think we can merge this as it is, but might want to consider refactoring the types as mentioned in a comment. If we do, we should do it in a separate PR

types.go
```diff
@@ -125,7 +126,10 @@ func NewMeteredOptionalType(gauge common.MemoryGauge, typ Type) OptionalType {
 func (OptionalType) isType() {}
 
 func (t OptionalType) ID() string {
-	return fmt.Sprintf("%s?", t.Type.ID())
+	if len(t.typeID) == 0 {
```
dreamsmasher, line 129: I'm not sure what the convention is for checking empty strings. It seems we have 40 instances of `len\(.*== *0` and 41 instances of `== *""` in the repo, but the first regex also includes arrays/slices.

dreamsmasher (approved):

types.go
SupunS: Good catch, added with https://github.com/onflow/cadence/pull/2336

SupunS:
Refactored the types in a follow-up PR: https://github.com/onflow/cadence/pull/2336

types.go
dreamsmasher: hmm, they're used interchangeably in the go stdlib and produce identical assembly: https://godbolt.org/z/fib1x1

entirely just a style choice, nothing big

PR #2327, opened 2023-02-10
Add Capability Controller types and values

### LINKED ISSUE #2292: Add new `StorageCapabilityController` and `AccountCapabilityController` types
## Context

The [Capability Controllers FLIP](https://github.com/onflow/flow/pull/798) proposes the new types `StorageCapabilityController` and `AccountCapabilityController`.

## Definition of Done

- [x] Add the type definitions for `StorageCapabilityController` and `AccountCapabilityController` to `sema` package, and add it to the set of built-in types.
- [x] Add a value implementation of `StorageCapabilityController` and `AccountCapabilityController` to the `interpreter` package.

dreamsmasher:
Closes #2292

## Description

Progress toward implementing [FLIP #798](https://github.com/onflow/flow/blob/d3bfc4a4de6e981c573b71983690797e87e63087/flips/20220203-capability-controllers.md).

Defines the `StorageCapabilityController` and `AccountCapabilityController` type and associated value constructors. Account integration to come in a subsequent PR

______


runtime/interpreter/capability_controller.go
```diff
@@ -0,0 +1,125 @@
...
+import (
+	"fmt"
+
+	"github.com/onflow/cadence/runtime/common"
+	"github.com/onflow/cadence/runtime/errors"
+	"github.com/onflow/cadence/runtime/format"
+	"github.com/onflow/cadence/runtime/sema"
+)
+
+type CapabilityControllerValue struct {
```
dsainati1, line 30: Maybe I am missing something, but where are these values actually constructed? It seems like the `NewCapabilityControllerValue` returns a `SimpleCompositeValue`, not a `CapabilityControllerValue`?

turbolent:
Given that the Capability Controller feature will require many PRs and we want to keep main branches like the Stable Cadence branch releasable, please target a new feature branch, e.g. `feature/capability-controllers`, branched off of `feature/stable-cadence`, instead of targeting it directly.

runtime/sema/capability_controller.cdc
```diff
@@ -0,0 +1,26 @@
...
+    /// Returns the targeted storage path of the capability.
+    pub fun target(): StoragePath {}
+   
+    /// Revoke the capability making it no longer usable.
+    /// When borrowing from a revoked capability the borrow returns nil.
+    pub fun revoke() {}
+   
+    /// Retarget the capability.
+    /// This moves the CapCon from one CapCon array to another.
+    pub fun retarget(target: StoragePath) {}
```
turbolent, line 25: Function declarations should not need empty bodies anymore, given that #2302 landed. If it isn't on `feature/stable-cadence` yet, it needs to be synced with `master`.

runtime/format/capability.go
```diff
@@ -42,3 +42,16 @@ func AccountCapability(address string) string {
 		address,
 	)
 }
+
+func CapabilityController(borrowType string, capabilityID string) string {
+	var typeArgument string
+	if borrowType != "" {
+		typeArgument = fmt.Sprintf("<%s>", borrowType)
+	}
```
turbolent, line 50: The borrow type is a member of the value, not a type argument

runtime/interpreter/capability_controller.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	var str string
+	stringer := func(memoryGauge common.MemoryGauge, seenReferences SeenReferences) string {
+		if str == "" {
+			common.UseMemory(memoryGauge, common.CapabilityControllerStringMemoryUsage)
+			borrowTypeStr := borrowType.MeteredString(memoryGauge)
+
+			idStr := fmt.Sprint(capabilityID) // probably better to take the log10(capabilityID) first
+			common.UseMemory(memoryGauge, common.NewStringMemoryUsage(len(idStr)))
```
turbolent, line 116: Memory needs to be metered before allocating it

runtime/interpreter/capability_controller.go
```diff
@@ -0,0 +1,125 @@
...
+
+			str = format.CapabilityController(borrowTypeStr, idStr)
+		}
+
+		return str
+	}
+
+	return NewSimpleCompositeValue(gauge, sema.CapabilityControllerType.ID(), staticType, capabilityControllerFieldNames, fields, computeField, nil, stringer)
```
turbolent, line 124: Refactor long argument lists to one argument per line

runtime/interpreter/statictype.go
```diff
@@ -945,3 +945,39 @@ func (p TypeParameter) String() string {
 	}
 	return builder.String()
 }
+
+type CapabilityControllerStaticType struct {
+	BorrowType StaticType
```
turbolent, line 950: The type `CapabilityController` itself has no type parameters (at the moment). The borrow type of a capability controller is a run-time type, it is a member of the `CapabilityController` value.

runtime/interpreter/capability_controller.go
dreamsmasher: ah yeah whoops, i originally implemented things in terms of `CapabilityControllerValue` but switched over to simple composite value. forgot to remove the actual struct definition and visitor code

runtime/parser/declaration.go
```diff
@@ -1384,6 +1384,7 @@ func parseMemberOrNestedDeclaration(p *parser, docString string) (ast.Declaratio
 				staticPos,
 				nativePos,
 				identifier,
+				docString,
```
turbolent, line 1387: This and the changes below are from another PR. To avoid a re-review, please update the target branch first to include these changes (i.e. merge `feature/stable-cadence` into `feature/capability-controllers`)

runtime/sema/capability_controller.cdc
```diff
@@ -0,0 +1,35 @@
...
+    /// The Type of the capability, i.e.: the T in Capability<T>.
+    pub let borrowType: Type
+   
+    /// The id of the related capability.
+    /// This is the UUID of the created capability.
+    /// All copies of the same capability will have the same UUID
+    pub let capabilityID: UInt64
+   
+    /// Is the capability revoked.
+    pub fun isRevoked(): Bool
```
turbolent, line 14: With `revoke` having changed to `delete`, this function got removed as well

runtime/interpreter/primitivestatictype.go
```diff
@@ -200,6 +200,7 @@ const (
 	PrimitiveStaticTypePublicAccountKeys
 	PrimitiveStaticTypeAccountKey
 	PrimitiveStaticTypeAuthAccountInbox
+	PrimitiveStaticTypeCapabilityController
```
turbolent, line 203: Also make sure to handle this new type in relevant places, e.g. `PrimitiveStaticType.elementType()`, `PrimitiveStaticType.SemaType()`, etc.

runtime/interpreter/capability_controller.go
```diff
@@ -0,0 +1,130 @@
...
+			return NewHostFunctionValue(inter, func(invocation Invocation) Value {
+				return AsBoolValue(isRevoked)
+			}, sema.CapabilityControllerTypeIsRevokedFunctionType)
+
+		case sema.CapabilityControllerTypeTargetFunctionName:
+			return NewHostFunctionValue(gauge, func(invocation Invocation) Value {
+				return targetPath
+			}, sema.CapabilityControllerTypeTargetFunctionType)
+
+		case sema.CapabilityControllerTypeRevokeFunctionName:
```
turbolent, line 70: The revoke function was removed from the type and replaced with the delete function

runtime/interpreter/capability_controller.go
```diff
@@ -0,0 +1,130 @@
...
+			return capabilityID
+		}),
+	}
+
+	computeField := func(name string, inter *Interpreter, locationRange LocationRange) Value {
+		switch name {
+		case sema.CapabilityControllerTypeIsRevokedFunctionName:
+			return NewHostFunctionValue(inter, func(invocation Invocation) Value {
+				return AsBoolValue(isRevoked)
+			}, sema.CapabilityControllerTypeIsRevokedFunctionType)
```
turbolent, line 63: As mentioned above, the `isRevoked` function should be removed

runtime/interpreter/capability_controller.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func NewCapabilityControllerValue(
+	gauge common.MemoryGauge,
+	issueHeight uint64,
+	capabilityID uint64,
+	targetPath PathValue,
+	borrowType StaticType,
+	isRevoked bool,
+	revoke func() error,
```
turbolent, line 43:
```suggestion
	delete func() error,
```

runtime/interpreter/primitivestatictype.go
dreamsmasher: sorry, do you mean `elementSize`? can't find any other match with `rg -i 'primitivestatic.*element'`

turbolent: Yes, `elementSize`

dreamsmasher: done

runtime/interpreter/capability_controller.go
```diff
@@ -0,0 +1,123 @@
...
+				if !ok {
+					panic(errors.NewUnreachableError())
+				}
+
+				err := retarget(newTarget)
+				if !ok {
+					panic(err)
+				}
+
+				targetPath = newTarget
```
turbolent, line 86: It might be good to avoid the local state here and instead have `targetPath` a getter function as well: The API will allow users to have multiple controller instances for a single capability, so if we return a new `CapabilityControllerValue` each time, then a retarget on one will not be reflected in other instances. Alternative we would need to ensure that there is only exactly one `CapabilityControllerValue` for a capability and we always return the same one

runtime/interpreter/capability_controller.go
```diff
@@ -0,0 +1,123 @@
...
+
+		return nil
+	}
+
+	var str string
+	stringer := func(memoryGauge common.MemoryGauge, seenReferences SeenReferences) string {
+		if str == "" {
+			common.UseMemory(memoryGauge, common.CapabilityControllerStringMemoryUsage)
+
+			// "Type<T>()"
```
turbolent, line 99: What does this comment mean?

runtime/sema/type_tags.go
```diff
@@ -218,6 +218,7 @@ const (
 	transactionTypeMask
 
 	invalidTypeMask
+	capabilityControllerTypeMask
```
turbolent, line 221: Doesn't really matter much, but maybe but this after `transactionTypeMask` (before the gap) – it looks like the gap is there to separate the invalid type mask from the rest. (Not sure, @SupunS might know)

runtime/sema/capability_controller.cdc
```diff
@@ -0,0 +1,35 @@
+pub struct CapabilityController {
```
turbolent, line 1: Please add some checker tests for this new type

turbolent (commented):
Looks good! Just one issue left re: future usage of the new value

runtime/interpreter/capability_controller.go
dreamsmasher: ah whoops, I was laying out the expected output in comments and forgot to remove. Was talking about meteredstring output for typevalues

runtime/sema/capability_controller.cdc
dreamsmasher: hmm, added some rudimentary ones asserting the type exists in scope and conforms to the spec. will add some proper tests for the account capabilities methods when this PR is in, since those involve non-trivial generics.

runtime/tests/checker/capability_controller_test.go
```diff
@@ -0,0 +1,63 @@
...
+		code,
+		ParseAndCheckOptions{
+			Config: &sema.Config{
+				BaseValueActivation: baseActivation,
+			},
+		},
+	)
+}
+
+func TestCapconNonEquatable(t *testing.T) {
```
turbolent, line 49:
```suggestion
func TestCheckCapconNonEquatable(t *testing.T) {
	t.Parallel()

```

runtime/tests/checker/capability_controller_test.go
```diff
@@ -0,0 +1,63 @@
...
+	_, err := ParseAndCheckCapcon(t, `
+		let kaboom: Bool = controller == controller
+	`)
+
+	errs := RequireCheckerErrors(t, err, 1)
+	require.IsType(t, &sema.InvalidBinaryOperandsError{}, errs[0])
+}
+
+func TestCapconTypeInScope(t *testing.T) {
```
turbolent, line 58:
```suggestion
func TestCheckCapconTypeInScope(t *testing.T) {
	t.Parallel()

```

runtime/tests/checker/capability_controller_test.go
```diff
@@ -0,0 +1,63 @@
+/*
```
turbolent, line 1: Could you please also add a test which asserts types of the type's members? e.g. for `issueHeight` it could be something simple as `let issueHeight: UInt64 = controller.issueHeight`

turbolent (commented):
It would be great to have a few more tests for the type definition. That would ensure we don't break the definitions at some point (e.g. we can refactor/optimize fearlessly).

Also, the generated code needs to be updated, and the code needs a gofmt run to fix imports.

turbolent:
Replaced by #2413

PR #2328, opened 2023-02-13
Refactor tests to cache typeIDs during assertions

SupunS:
Work towards #2325

## Description

Use the utility method introduced in https://github.com/onflow/cadence/pull/2326 in tests to cache the typeID, so the value assertions are not broken by the new cached typeID field.

______

### DIFF
--- a/encoding/json/encoding_test.go
+++ b/encoding/json/encoding_test.go
@@ -2990,8 +2990,8 @@ func testDecode(t *testing.T, actualJSON string, expectedVal cadence.Value, opti
 
 	assert.Equal(
 		t,
-		valueWithCachedTypeID(expectedVal),
-		valueWithCachedTypeID(decodedVal),
+		cadence.ValueWithCachedTypeID(expectedVal),
+		cadence.ValueWithCachedTypeID(decodedVal),
 	)
 }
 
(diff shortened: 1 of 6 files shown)


encoding/json/encoding_test.go
```diff
@@ -3221,91 +3221,3 @@ func TestExportFunctionValue(t *testing.T) {
         `,
 	)
 }
-
-// valueWithCachedTypeID recursively caches type ID of value v's type.
-// This is needed because each type ID is lazily cached on
-// its first use in ID() to avoid performance penalty.
-func valueWithCachedTypeID(v cadence.Value) cadence.Value {
```
SupunS, line 3228: I moved this and the below method to `values.go` and `types.go` files respectively, to be able to use in other tests, and to avoid circular dependency.

types_test.go
```diff
@@ -116,38 +116,114 @@ func TestType_ID(t *testing.T) {
...
+				QualifiedIdentifier: "BarI",
 			},
-			"S.test.FooI",
+			"S.test.BarI",
 		},
 		{
 			(&RestrictedType{}).WithID("S.test.Foo{S.test.FooI}"),
 			"S.test.Foo{S.test.FooI}",
 		},
+		{
```
SupunS, line 175: Existing tests did not cover some of the types, so added those.

fxamacker (approved):

PR #2339, opened 2023-02-17
Add metering for encoded values

SupunS:
Work towards https://github.com/dapperlabs/cadence-private-issues/issues/64

## Description

Adds memory metering for encoded values.

Also moves the computation metering of [ValueStore.SetValue()](https://github.com/onflow/flow-go/blob/fca02b0ce396a561f9a893520352ce320675aa74/fvm/environment/value_store.go#L160) to the Cadence side. This now happens during the value encoding.

______


runtime/common/computationkind.go
```diff
@@ -109,7 +109,7 @@ const (
 	_
 	_
 	_
-	_
+	ComputationKindEncodeValue
```
SupunS, line 112: I added this new computation kind here. Let me know if this needs to change. We could remove the computation for `ComputationKindSetValue` from FVM once this is merged.
cc: @janezpodhostnik

runtime/common/metering.go
```diff
@@ -31,6 +31,8 @@ type MemoryUsage struct {
 	Amount uint64
 }
 
+// MemoryGauge may be used in parallel. e.g: During value encoding.
+// Hence, any implementation of this interface must be thread-safe.
```
turbolent, line 35: We should maybe check if is the case in FVM

turbolent (approved):
Nice!

@fxamacker Could we please tag a new release for atree?

runtime/common/metering.go
janezpodhostnik: Its not, it uses a map! But it can be made that way. This will have to be a FVM change.

janezpodhostnik: If metering is multi-threaded is determinism still guaranteed? 

Is it possible that because things will be metered in random order during encoding that and execution fork will happen because some transactions will fail differently then others? We should make sure that is not the case.

SupunS: hmm, that's a good point. Yes, the order would not be deterministic. All forks will fail eventually, but may not be at the same exact moment. Is that still going to be a problem?

janezpodhostnik: for memory measurements no, because the only relevant factor there is if it reaches the limit or not, and that is not dependent on the order. Same for interaction measurements.

Computation measurements are different, as those get computed into fees. 

To illustrate the problem, lets say there are 2 values to encode V1 and V2 at the end of the transaction. both use so much memory that the transaction will fail, but V1 uses 100 computation while v2 uses 200 computation. If V1 is encoded first the transaction will fail and the fees will be `F(100)`, but if V2 is encoded first the transaction will still fail but the fees will be `F(200)` which would cause an execution fork.

SupunS: As discussed, moved the metering to "before" encoding:https://github.com/onflow/cadence/pull/2339/commits/0cbfe2e89a0786b2cbb27c8d976a899857f8a44f

This also made it simpler, and also the memory meter doesn't need to be thread-safe anymore 👍

runtime/runtime_memory_metering_test.go
```diff
@@ -1071,3 +1071,150 @@ func TestMemoryMeteringErrors(t *testing.T) {
...
+
+		text := "A quick brown fox jumps over the lazy dog"
+
+		err := rt.ExecuteTransaction(
+			Script{
+				Source: []byte(fmt.Sprintf(`
+                transaction() {
+                    prepare(acc: AuthAccount) {
+                        var s = "%s"
+                        acc.save(s, to:StoragePath(identifier: "some_path")!)
```
turbolent, line 1107: Maybe just use a path literal:

```suggestion
                        acc.save(s, to: /storage/some_path)
```

runtime/runtime_memory_metering_test.go
```diff
@@ -1071,3 +1071,150 @@ func TestMemoryMeteringErrors(t *testing.T) {
...
+		address := common.MustBytesToAddress([]byte{0x1})
+		storage := newTestLedger(nil, nil)
+		meter := newTestMemoryGauge()
+
+		runtimeInterface := &testRuntimeInterface{
+			storage: storage,
+			getSigningAccounts: func() ([]Address, error) {
+				return []Address{address}, nil
+			},
+			log:         func(message string) {},
```
turbolent, line 1139: Looks like it's not used, maybe remove
```suggestion
```

runtime/runtime_memory_metering_test.go
```diff
@@ -1071,3 +1071,150 @@ func TestMemoryMeteringErrors(t *testing.T) {
...
+		err := rt.ExecuteTransaction(
+			Script{
+				Source: []byte(fmt.Sprintf(`
+                transaction() {
+                    prepare(acc: AuthAccount) {
+                        var i = 0
+                        var s = "%s"
+                        while i<1000 {
+                            acc.save(s, to:StoragePath(identifier: "i".concat(i.toString()))!)
+                            i=i+1
```
turbolent, line 1154:
```suggestion
                        while i < 1000 {
                        	let path = StoragePath(identifier: "i".concat(i.toString()))!
                            acc.save(s, to: path)
                            i = i + 1
```

runtime/runtime_memory_metering_test.go
```diff
@@ -1071,3 +1071,150 @@ func TestMemoryMeteringErrors(t *testing.T) {
...
+		_, err := rt.ExecuteScript(
+			Script{
+				Source: []byte(`
+                pub fun main() {
+                    let acc = getAuthAccount(0x02)
+                    var i = 0
+                    var f = Foo()
+                    while i<1000 {
+                        acc.save(f, to:StoragePath(identifier: "i".concat(i.toString()))!)
+                        i=i+1
```
turbolent, line 1200:
```suggestion
                    while i < 1000 {
                        let path = StoragePath(identifier: "i".concat(i.toString()))!
                        acc.save(f, to: path)
                        i = i + 1
```

runtime/runtime_test.go
```diff
@@ -7548,24 +7550,36 @@ func TestRuntimeComputationMetring(t *testing.T) {
...
-			ok:          true,
-			expCompUsed: 5,
+			ok:        true,
+			hits:      5,
+			intensity: 6,
+		},
+		{
+			name: "statement + (functionInvocation x 2) + encoding",
+			code: `
+          acc.save("A quick brown fox jumps over the lazy dog", to:StoragePath(identifier: "some_path")!)
```
turbolent, line 7578: Maybe just use a path literal

```suggestion
          acc.save("A quick brown fox jumps over the lazy dog", to: /storage/some_path)
```

runtime/storage.go
```diff
@@ -232,6 +232,14 @@ func (s *Storage) Commit(inter *interpreter.Interpreter, commitContractUpdates b
 	deltas := s.PersistentSlabStorage.DeltasWithoutTempAddresses()
 	common.UseMemory(s.memoryGauge, common.NewAtreeEncodedSlabMemoryUsage(deltas))
 
+	size := s.PersistentSlabStorage.DeltasSizeWithoutTempAddresses()
+	if size > 0 {
+		inter.ReportComputation(common.ComputationKindEncodeValue, uint(size))
+
+		usage := common.NewBytesMemoryUsage(int(size))
+		common.UseMemory(s.memoryGauge, usage)
+	}
```
turbolent, line 241: Maybe move this above the introduction of `deltas`.

Love how simple the PR got! 👌

go.mod
```diff
@@ -11,10 +11,10 @@ require (
 	github.com/k0kubun/pp v3.0.1+incompatible
 	github.com/leanovate/gopter v0.2.9
 	github.com/logrusorgru/aurora v0.0.0-20200102142835-e9ef32dff381
-	github.com/onflow/atree v0.4.0
+	github.com/onflow/atree v0.4.1-0.20230228213801-3c7c4769dae0
```
turbolent, line 14: Could we maybe release a new version of atree @fxamacker?

turbolent (approved):
Nice! 

Just a couple minor suggestions

go.mod
fxamacker: @bastian @SupunS

This looks like a green light, so I merged Atree PR [290](https://github.com/onflow/atree/pull/290) to add DeltasSizeWithoutTempAddresses to return uncommitted slab size.

>> Could we maybe release a new version of atree @fxamacker?

Yes, unless it's needed today, I plan to tag a new version after wrapping up these items:

- Merge Supun's [PR 291](https://github.com/onflow/atree/pull/291) (currently draft) to revert PR [287](https://github.com/onflow/atree/pull/287) adding a callback to the Encoder.
- Refactor atree error handling .  I'm currently working on this so it'll wrap up soon (tomorrow or Monday if there are no surprises).  I was focused on [adding CCF codec](https://github.com/onflow/cadence/pull/2364) to onflow/cadence until yesterday.

janezpodhostnik (approved):

go.mod
turbolent: @SupunS https://github.com/onflow/atree/releases/tag/v0.5.0 is ready now. Could you please update the PR?

SupunS: Updated: 4470df8a9e587ff3fb662da4674ce599106d2f1d

go.mod
```diff
@@ -11,10 +11,10 @@ require (
 	github.com/k0kubun/pp v3.0.1+incompatible
 	github.com/leanovate/gopter v0.2.9
 	github.com/logrusorgru/aurora v0.0.0-20200102142835-e9ef32dff381
-	github.com/onflow/atree v0.4.0
+	github.com/onflow/atree v0.4.1-0.20230217210622-1ca0edf9d31f
```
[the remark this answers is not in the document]
SupunS: updated : 4470df8a9e587ff3fb662da4674ce599106d2f1d

PR #2341, opened 2023-02-23
Convert reference values

turbolent:
Port of https://github.com/dapperlabs/cadence-internal/pull/113 and https://github.com/dapperlabs/cadence-internal/pull/114

Cf: https://github.com/dapperlabs/cadence-internal/compare/v0.31.3...v0.31-internal

## Description

When transferring a reference value, convert it to the target type.

______

- [x] Targeted PR against ~`master`~ release branch

### DIFF
--- a/version.go
+++ b/version.go
@@ -21,4 +21,4 @@
 
 package cadence
 
-const Version = "v0.31.4"
+const Version = "v0.31.5"
--- a/version.go
+++ b/version.go
@@ -21,4 +21,4 @@
 
 package cadence
 
-const Version = "v0.31.3"
+const Version = "v0.31.4"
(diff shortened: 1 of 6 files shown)


dsainati1 (approved):

version.go
```diff
@@ -21,4 +21,4 @@
 
 package cadence
 
-const Version = "v0.31.3"
+const Version = "v0.31.5"
```
SupunS, line 24: So the next tag going to be `v0.31.6`?

SupunS (approved):

version.go
turbolent: Yes, if we release another patch after this. This PR is just making public what was already released, so no need to bump the version here one more time.

PR #2346, opened 2023-02-24
Refactor interpreter `convert()` function

SupunS:
## Description

Move the common `if valueType.Equal(unwrappedTargetType)` check to the top.

______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -1620,150 +1620,94 @@ func (interpreter *Interpreter) convert(value Value, valueType, targetType sema.
 
 	unwrappedTargetType := sema.UnwrapOptionalType(targetType)
 
+	if valueType.Equal(unwrappedTargetType) {
+		return value
+	}
+
 	switch unwrappedTargetType {
 	case sema.IntType:
-		if !valueType.Equal(unwrappedTargetType) {
-			return ConvertInt(interpreter, value, locationRange)
-		}
-
+		return ConvertInt(interpreter, value, locationRange)
 	case sema.UIntType:
-		if !valueType.Equal(unwrappedTargetType) {
-			return ConvertUInt(interpreter, value, locationRange)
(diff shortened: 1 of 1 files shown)


turbolent (commented):
This is definitely cleaner, but the reasoning for performing the equality tests in the branches was that this avoids unnecessary equality tests for types not handled by the switch.

Benchmarks do not report any difference, but that is probably because we do not have good benchmarks for this code, not tests it seems.

Could you please add some benchmarks to ensure this does not regress performance?

SupunS:
Yeah, that's a fair point. In that case, we can selectively do the type-equality check: d6b546cf2ff068d0eba54081dc4f2ca2e13474c2

runtime/interpreter/interpreter.go
```diff
@@ -1620,150 +1620,104 @@ func (interpreter *Interpreter) convert(value Value, valueType, targetType sema.
...
+				unwrappedTargetType.Type,
+			)
 
-			default:
-				panic(errors.NewUnexpectedError("unsupported reference value: %T", ref))
-			}
+		default:
+			panic(errors.NewUnexpectedError("unsupported reference value: %T", ref))
 		}
 	}
```
turbolent, line 1722: Maybe panic with an unexpected error here: The new check at the top filters which types need conversion – if it allowed a type, and we reached this place, the type's conversion isn't implemented properly (feel free to even add this as a comment)

turbolent (approved):
Looks good! Just one minor suggestion

runtime/interpreter/interpreter.go
SupunS: While adding this, it made me realize that certain values that were allowed from the above check don't need conversion. e.g: Numeric super types.
Now I'm having second thoughts about this check. Having the checks in two places can easily lead to inconsistencies. 
I now think it maybe good to keep it as is.

PR #2349, opened 2023-02-27
fixed formating of codeblocks in lists and remove outdated callout 

bjartek:
Remove comment that default implementations will be available later. They are here today

### DIFF
--- a/docs/language/interfaces.mdx
+++ b/docs/language/interfaces.mdx
@@ -470,10 +470,6 @@ struct SomeInner: OuterInterface.InnerInterface {}
 
 ## Interface Default Functions
 
-<Callout type="info">
-🚧 Status: This feature will be available in a future spork.
-</Callout>
-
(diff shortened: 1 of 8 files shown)


dsainati1 (approved):

bjartek:
I dont know why ci fails here, but it appears that it does not work well on PRs cross forks.

SupunS:
Unfortunately, the `BackwardCompatibilityCheck` won't work across forks (limitation with `go get`). Since this is just a doc update, we can ignore that check.

docs/language/interfaces.mdx
```diff
@@ -470,10 +470,6 @@ struct SomeInner: OuterInterface.InnerInterface {}
 
 ## Interface Default Functions
 
-<Callout type="info">
```
turbolent, line 473: 👍

turbolent (approved):

PR #2355, opened 2023-02-28
Require #allowAccountLinking pragma to allow use of account linking function

### LINKED ISSUE #2353: Gate account linking behind pragma
### Issue to be solved

Currently, the [account capabilities / linking](https://github.com/onflow/flips/pull/53) feature is gated behind a configuration option of the checker. This option is constantly enabled by FVM on all networks but Mainnet.

Sentiment on the [account capabilities FLIP](https://github.com/onflow/flips/pull/53) has been positive, but there has also been a lot of feedback, requesting additional safe guards to make abuse of the feature harder.

A long term solution is being discussed in https://forum.onflow.org/t/super-user-account/4088. However, it will likely only be released with Stable Cadence.

As we would like to not block the release of the feature until the release of Stable Cadence, we need a temporary solution. 

### Suggested Solution

Introduce a new `#allowAccountLinking` pragma.

In the Cadence implementation, when the type checker encounters the pragma, enable the account linking feature.

This solution is both simple to add and remove, once migrating to the long-term solution.

It also allows wallets to detect relatively easily if the account linking feature is used in a transaction, and potentially prevent it.

turbolent:
Closes #2353 

## Description

Even when account linking is enable, require the program to have a pragma declaration `#allowAccountLinking` to make the `AuthAccount.linkAccount` available.

The pragma must appear at the top-level (i.e. may not be nested), and must appear before all other declarations.

______

### DIFF
--- a/runtime/sema/checker.go
+++ b/runtime/sema/checker.go
@@ -112,6 +112,7 @@ type Checker struct {
 	inInvocation                       bool
 	inCreate                           bool
 	isChecked                          bool
+	allowAccountLinking                bool
 }
 
 var _ ast.DeclarationVisitor[struct{}] = &Checker{}
@@ -329,7 +330,31 @@ func (checker *Checker) CheckProgram(program *ast.Program) {
 
 	checker.checkTopLevelDeclarationValidity(declarations)
 
+	var rejectAllowAccountLinkingPragma bool
+
 	for _, declaration := range declarations {
+
+		// A pragma declaration #allowAccountLinking determines
+		// if the program is allowed to use the account linking.
+		//
+		// It must appear as a top-level declaration (i.e. not nested in the program),
+		// and must appear before all other declarations (i.e. at the top of the program).
+		//
+		// This is a temporary feature, which is planned to get replaced
+		// by capability controllers, a new Account type, and account entitlements.
+
+		if pragmaDeclaration, isPragma := declaration.(*ast.PragmaDeclaration); isPragma {
+			if IsAllowAccountLinkingPragma(pragmaDeclaration) {
+				if rejectAllowAccountLinkingPragma {
+					checker.reportInvalidNonHeaderPragma(pragmaDeclaration)
+				} else {
+					checker.allowAccountLinking = true
+				}
+				continue
+			}
+		}
+
+		rejectAllowAccountLinkingPragma = true
 
 		// Skip import declarations, they are already handled above
 		if _, isImport := declaration.(*ast.ImportDeclaration); isImport {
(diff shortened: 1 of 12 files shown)


dsainati1 (approved):

runtime/sema/checker.go
```diff
@@ -360,6 +385,10 @@ func (checker *Checker) checkTopLevelDeclarationValidity(declarations []ast.Decl
 	}
 
 	for _, declaration := range declarations {
+		if _, ok := declaration.(*ast.PragmaDeclaration); ok {
+			continue
+		}
+
```
SupunS, line 391: Instead of special-casing it here, maybe add pragma to [validTopLevelDeclarationsInTransaction](https://github.com/onflow/cadence/blob/bd4d65b982a5a6cb3d75d33715b5b1db90e47da4/runtime/runtime.go#L167)?

runtime/tests/checker/pragma_test.go
```diff
@@ -79,21 +79,63 @@ func TestCheckPragmaInvalidInvocationExprNonStringExprArgument(t *testing.T) {
...
+
+		_, err := ParseAndCheck(t, `
+          #allowAccountLinking
+
+          let x = 1
+        `)
+		require.NoError(t, err)
+	})
+
+	t.Run("top-level, before other declarations", func(t *testing.T) {
```
SupunS, line 116:
```suggestion
	t.Run("top-level, after other declarations", func(t *testing.T) {
```

runtime/sema/checker.go
turbolent: Good idea, I didn't realize this exists. I added it in [6f3c9cf](https://github.com/onflow/cadence/pull/2355/commits/6f3c9cf196ecc6bb0caa5b032a5b3f1af1e9ad5a), and then noticed how this top-level declaration validity checking is quite wasteful and allocates maps, so removed the pragma-special casing and optimized it in [6f7cd51](https://github.com/onflow/cadence/pull/2355/commits/6f7cd512e6dccffa3cea217d4a86ec13224f3ebf)

runtime/sema/checker.go
```diff
@@ -341,51 +366,50 @@ func (checker *Checker) CheckProgram(program *ast.Program) {
...
 
-func (checker *Checker) checkTopLevelDeclarationValidity(declarations []ast.Declaration) {
+func (checker *Checker) checkTopLevelDeclarationsValidity(declarations []ast.Declaration) {
 	validTopLevelDeclarationsHandler := checker.Config.ValidTopLevelDeclarationsHandler
 
 	if validTopLevelDeclarationsHandler == nil {
 		return
 	}
 
-	validDeclarationKinds := map[common.DeclarationKind]bool{}
```
turbolent, line 351: We were previously allocating a map for each check

dsainati1 (approved):

runtime/sema/checker.go
SupunS: Nice optimization 👌

SupunS (approved):

Issue #2350, opened 2023-02-27
Add defensive checks for adding and removing attachments


dsainati1:
### Issue to be solved

To improve security, it would be ideal to have additional defensive checks in the interpreter that ensure the static checks for attachments are not circumvented.

### Suggested Solution

- [x] Add some runtime checks to the `attach` and `remove` operations, that assert at run-time that the target of these operations (i.e. the composite having its attachment added or removed) is a resource type, and not a reference type.

PR #2357, opened 2023-02-28
Add defensive check for adding and removing attachments

dsainati1:
Closes https://github.com/onflow/cadence/issues/2350

We enforce that `attach` and `remove` for attachments can only target owned resource and struct objects statically in the type checker. However, we can also check defensively in the interpreter that this invariant holds. 

______

### DIFF
--- a/runtime/interpreter/errors.go
+++ b/runtime/interpreter/errors.go
@@ -929,3 +929,20 @@ func (e AttachmentIterationMutationError) Error() string {
 		e.Value.QualifiedIdentifier,
 	)
 }
+
+// InvalidAttachmentOperationTargetError
+type InvalidAttachmentOperationTargetError struct {
+	Value Value
+	LocationRange
+}
+
+var _ errors.InternalError = InvalidAttachmentOperationTargetError{}
+
+func (InvalidAttachmentOperationTargetError) IsInternalError() {}
+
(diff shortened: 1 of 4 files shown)


SupunS (approved):

turbolent (commented):
Nice! 

Could you please try to add test cases? You can do this by using `parseCheckAndInterpretWithOptions` and passing `ParseCheckAndInterpretOptions.HandleCheckerError`, expecting/ignoring the static checking error.

runtime/interpreter/errors.go
```diff
@@ -929,3 +929,20 @@ func (e AttachmentIterationMutationError) Error() string {
...
+}
+
+var _ errors.InternalError = InvalidAttachmentOperationTargetError{}
+
+func (InvalidAttachmentOperationTargetError) IsInternalError() {}
+
+func (e InvalidAttachmentOperationTargetError) Error() string {
+	return fmt.Sprintf(
+		"cannot add or remove attachment with non-owned value %s",
+		e.Value.String(),
```
turbolent, line 946: This might be quite computational and memory intensive, maybe just print the Go type in the error message:

```suggestion
		"cannot add or remove attachment with non-owned value (%T)",
		e.Value,
```

turbolent (approved):
Nice work! Just one last suggestion for the error

PR #2363, opened 2023-03-01
Parsing for Entitlements

dsainati1:
Begins implementation of https://github.com/onflow/flips/pull/54, part of https://github.com/onflow/cadence/issues/2412

Implements parsing for entitlement declarations, entitlement mapping delcarations, `access(X)` modifiers, and `auth(X)` reference types
______


turbolent:
Great start! Before we merge, we should target a feature branch, so we don't block the Stable Cadence release

runtime/ast/access.go
```diff
@@ -20,38 +20,156 @@ package ast
...
+func (EntitlementAccess) isAccess() {}
+
+func (EntitlementAccess) Description() string {
+	return "entitled access"
+}
+
+func (e EntitlementAccess) entitlementsString() string {
+	str := strings.Builder{}
+	for i, entitlement := range e.EntitlementSet.Entitlements() {
+		str.Write([]byte(entitlement.String()))
```
turbolent, line 95: Here and below:

```suggestion
		str.WriteString(entitlement.String())
```

runtime/ast/access.go
```diff
@@ -20,38 +20,156 @@ package ast
...
+		if i < len(e.EntitlementSet.Entitlements())-1 {
+			str.Write([]byte(e.EntitlementSet.Separator()))
+		}
+	}
+	return str.String()
+}
+
+func (e EntitlementAccess) String() string {
+	return "ConjunctiveEntitlementAccess " + e.entitlementsString()
```
turbolent, line 104: Maybe create a string builder here, write the prefix to it, and pass it to `entitlementsString` (maybe as a `StringWriter`)

runtime/ast/access.go
```diff
@@ -20,38 +20,156 @@ package ast
...
+}
+
+func (e EntitlementAccess) Keyword() string {
+	return "access(" + e.entitlementsString() + ")"
```
turbolent, line 108: Same as above, maybe create a string builder here, write the prefix to it, and pass it to `entitlementsString`

runtime/ast/access.go
```diff
@@ -20,38 +20,156 @@ package ast
...
+}
+
+func (e EntitlementAccess) MarshalJSON() ([]byte, error) {
+	return json.Marshal(e.String())
+}
+
+func (e EntitlementAccess) subset(other EntitlementAccess) bool {
+	otherSet := make(map[*NominalType]struct{})
+	for _, entitlement := range other.EntitlementSet.Entitlements() {
```
turbolent, line 117: Pre-allocate to the known final size of the map:

```suggestion
	otherEntitlements := other.EntitlementSet.Entitlements()
	otherSet := make(map[*NominalType]struct{}, len(otherEntitlements))
	for _, entitlement := range otherEntitlements {
```

Also, maybe the map should be cached in the `EntitlementAccess`? Maybe `ConjunctiveEntitlementSet.Elements` should be an `OrderedSet`?

runtime/ast/type.go
```diff
@@ -570,13 +574,24 @@ func (t *ReferenceType) EndPosition(memoryGauge common.MemoryGauge) Position {
...
+	if t.Authorization != nil {
+		doc = append(doc, referenceTypeAuthKeywordDoc)
+		if t.Authorization.EntitlementSet != nil && len(t.Authorization.EntitlementSet.Entitlements()) > 0 {
+			doc = append(doc, prettier.Text("("))
+			for i, entitlement := range t.Authorization.EntitlementSet.Entitlements() {
+				doc = append(doc, entitlement.Doc())
+				if i < len(t.Authorization.EntitlementSet.Entitlements())-1 {
+					doc = append(doc, prettier.Text(t.Authorization.EntitlementSet.Separator()), prettier.Space)
+				}
+			}
```
turbolent, line 591: Maybe refactor the repeated `t.Authorization.EntitlementSet.Entitlements()` expressions into a variable.

runtime/parser/declaration.go
```diff
@@ -274,7 +284,7 @@ var enumeratedAccessModifierKeywords = common.EnumerateWords(
 //	access
 //	    : 'priv'
 //	    | 'pub' ( '(' 'set' ')' )?
-//	    | 'access' '(' ( 'self' | 'contract' | 'account' | 'all' ) ')'
+//	    | 'access' '(' ( 'self' | 'contract' | 'account' | 'all' | '<ENTITLEMENT LIST>' ) ')'
```
turbolent, line 287:
```suggestion
//	    | 'access' '(' ( 'self' | 'contract' | 'account' | 'all' | entitlementList ) ')'
```

runtime/ast/entitlement_declaration.go
```diff
@@ -0,0 +1,282 @@
...
+func (d *EntitlementMappingDeclaration) DeclarationIdentifier() *Identifier {
+	return &d.Identifier
+}
+
+func (d *EntitlementMappingDeclaration) DeclarationAccess() Access {
+	return d.Access
+}
+
+func (d *EntitlementMappingDeclaration) DeclarationKind() common.DeclarationKind {
+	return common.DeclarationKindEntitlement
```
turbolent, line 211: Probably should have its own declaration kind, `DeclarationKindEntitlementMapping`

runtime/parser/declaration.go
```diff
@@ -339,32 +349,81 @@ func parseAccess(p *parser) (ast.Access, error) {
...
+			p.nextSemanticToken()
 
 		default:
-			return ast.AccessNotSpecified, p.syntaxError(
-				"expected keyword %s, got %q",
-				enumeratedAccessModifierKeywords,
-				keyword,
-			)
-		}
+			firstTy, err := parseNominalType(p, lowestBindingPower, true)
```
turbolent, line 377: Maybe refactor the entitlement list parsing into a function (the code in this switch's default case)

runtime/parser/parser.go
```diff
@@ -179,8 +179,7 @@ func ParseTokenStream[T any](
 	result, err := parse(p)
 	if err != nil {
 		p.report(err)
-		var zero T
-		return zero, p.errors
+		return result, p.errors
```
turbolent, line 182: Was there a particular reason for changing the behaviour here and returning the result even if there is an error?

runtime/parser/type.go
```diff
@@ -544,10 +545,36 @@ func defineRestrictedOrDictionaryType() {
 	)
 }
 
-// parseNominalTypes parses zero or more nominal types separated by comma.
+func parseNominalType(
+	p *parser,
+	rightBindingPower int,
+	rejectAccessKeywords bool,
```
turbolent, line 551: Maybe instead of adding specific "access keyword rejection" support to the nominal type parsing here and to `parseNominalTypes`, maybe just parse all identifiers as nominal types, and then reject nominal type lists that have nominal types which are e.g. access keywords in the case of parsing entitlements.

Also/alternatively, given that this feature goes into Stable Cadence: Naomi had worked on rejecting identifiers which are also keywords in other places, see e.g. `nonReservedIdentifier`, so maybe we can just use that here and generalize from specifically rejecting access-related keywords to all hard keywords.

runtime/parser/type.go
```diff
@@ -937,6 +959,60 @@ func defineIdentifierTypes() {
 			case KeywordAuth:
 				p.skipSpaceAndComments()
 
+				var authorization ast.Authorization
```
turbolent, line 962: It looks like the code here shares a lot with the code in `parseAccess`

tools/update/package-lock.json
turbolent, line 1: This change seems unrelated and was probably committed by accident, maybe undo and open as a separate PR

turbolent (commented):
Nice!

runtime/parser/declaration.go
```diff
@@ -957,6 +1025,174 @@ func parseFieldWithVariableKind(
...
+	inputNominalType, ok := inputType.(*ast.NominalType)
+	if !ok {
+		p.reportSyntaxError(
+			"expected nominal type, got %s",
+			inputType,
+		)
+	}
+	if err != nil {
+		return nil, err
+	}
```
SupunS, line 1042: Shouldn't this check be moved to right after parsing the type?

```suggestion
	inputType, err := parseType(p, lowestBindingPower)
	if err != nil {
		return nil, err
	}

	inputNominalType, ok := inputType.(*ast.NominalType)
	if !ok {
		p.reportSyntaxError(
			"expected nominal type, got %s",
			inputType,
		)
	}
```

SupunS, line 1042: Shouldn't the error-check be moved to right after parsing the type?

```suggestion
	inputType, err := parseType(p, lowestBindingPower)
	if err != nil {
		return nil, err
	}

	inputNominalType, ok := inputType.(*ast.NominalType)
	if !ok {
		p.reportSyntaxError(
			"expected nominal type, got %s",
			inputType,
		)
	}
```

runtime/parser/declaration.go
```diff
@@ -957,6 +1025,174 @@ func parseFieldWithVariableKind(
...
+	outputNominalType, ok := outputType.(*ast.NominalType)
+	if !ok {
+		p.reportSyntaxError(
+			"expected nominal type, got %s",
+			outputType,
+		)
+	}
+	if err != nil {
+		return nil, err
+	}
```
SupunS, line 1063: Same question here

runtime/parser/declaration.go
```diff
@@ -957,6 +1025,174 @@ func parseFieldWithVariableKind(
...
+
+		default:
+			mapping, err := parseEntitlementMapping(p, docString)
+			if err != nil {
+				return nil, err
+			}
+
+			if mapping == nil {
+				return mappings, nil
+			}
```
SupunS, line 1095: Can `mapping` be `nil`?

runtime/parser/declaration.go
```diff
@@ -1108,6 +1344,50 @@ func parseCompositeOrInterfaceDeclaration(
...
+
+	// skip the `entitlement` keyword
+	p.nextSemanticToken()
+
+	nominalType, err := parseNominalType(p, lowestBindingPower, true)
+	if err != nil {
+		return nil, err
+	}
+
+	return nominalType, nil
```
SupunS, line 1373:
```suggestion
	return parseNominalType(p, lowestBindingPower, true)
```

runtime/parser/type.go
```diff
@@ -544,10 +545,36 @@ func defineRestrictedOrDictionaryType() {
...
+		return nil, p.syntaxError("unexpected non-nominal type: %s", ty)
+	}
+	if rejectAccessKeywords &&
+		nominalType.Identifier.Identifier == KeywordAll ||
+		nominalType.Identifier.Identifier == KeywordAccess ||
+		nominalType.Identifier.Identifier == KeywordAccount ||
+		nominalType.Identifier.Identifier == KeywordSelf {
+		return nil, p.syntaxError("unexpected non-nominal type: %s", ty)
+	}
+	return nominalType, nil
```
SupunS, line 568: maybe use a switch-case here:
```suggestion
	if rejectAccessKeywords {
		switch nominalType.Identifier.Identifier {
		case KeywordAll, KeywordAccess, KeywordAccount, KeywordSelf
			return nil, p.syntaxError("unexpected non-nominal type: %s", ty)
		}
	}
	return nominalType, nil
```

runtime/parser/declaration.go
dsainati1: No guess not

SupunS (approved):
Looks good!

I think it is better to merge this to a separate branch, so that it'll allow us to do a preview build of stable cadence before merging all entitlements-related PRs, if needed.

runtime/ast/access.go
```diff
@@ -20,38 +20,167 @@ package ast
...
+	Keyword() string
+	Description() string
+	String() string
+	MarshalJSON() ([]byte, error)
+	IsLessPermissiveThan(Access) bool
+}
+
+type EntitlementSet interface {
+	Entitlements() []*NominalType
+	Separator() string
```
SupunS, line 41: Noticed while checking the follow-up PRs: maybe return an enum (iota) for the separator, rather than a string.
Could avoid any inconsistencies, e.g: the `DisjunctiveEntitlementSet` separator has a preceding whitespace, but the `ConjunctiveEntitlementSet` does not.

runtime/parser/type.go
turbolent: Right, I just meant that this rejection functionality is currently "pushed down" all the way into the nominal type parsing function(s), though the callers of those parsing functions can handle it themselves: if they parse `false` for `rejectAccessKeywords`, then no rejection is needed; if they pass `true`, then the callr can also just reject the result of the call.

tools/update/package-lock.json
turbolent: Are the changes to this file related to this PR?

runtime/ast/access.go
```diff
@@ -20,38 +20,167 @@ package ast
...
+
+func (EntitlementAccess) Description() string {
+	return "entitled access"
+}
+
+func (e EntitlementAccess) entitlementsString(prefix strings.Builder) strings.Builder {
+	for i, entitlement := range e.EntitlementSet.Entitlements() {
+		prefix.WriteString(entitlement.String())
+		if i < len(e.EntitlementSet.Entitlements())-1 {
+			prefix.Write([]byte(e.EntitlementSet.Separator()))
```
turbolent, line 100:
```suggestion
			prefix.WriteString(e.EntitlementSet.Separator())
```

runtime/ast/access.go
```diff
@@ -20,38 +20,167 @@ package ast
...
+	return EntitlementAccess{EntitlementSet: entitlements}
+}
+
+func (EntitlementAccess) isAccess() {}
+
+func (EntitlementAccess) Description() string {
+	return "entitled access"
+}
+
+func (e EntitlementAccess) entitlementsString(prefix strings.Builder) strings.Builder {
```
turbolent, line 96: Maybe just pass by pointer here and avoid the return type, and having to remember to use the result at the call site
```suggestion
func (e EntitlementAccess) entitlementsString(prefix *strings.Builder) {
```

runtime/parser/keyword.go
```diff
@@ -120,6 +125,7 @@ var allKeywords = []string{
 	KeywordDefault,
 	KeywordEnum,
 	KeywordView,
+	KeywordWith,
```
turbolent, line 128: Do `KeywordMapping` and `KeywordRequire` have to be added here?

runtime/parser/declaration.go
```diff
@@ -269,12 +279,57 @@ var enumeratedAccessModifierKeywords = common.EnumerateWords(
...
+		// Luckily, however, the former two are just equivalent, and the latter we can disambiguate in the type checker.
+		return ast.NewConjunctiveEntitlementSet(entitlements), nil
+	default:
+		return nil, p.syntaxError(
+			"unexpected entitlement separator %s",
+			p.current.Type.String(),
+		)
+	}
+
+	if separator != lexer.TokenError {
```
turbolent, line 307: Is this condition always true?

runtime/ast/access.go
```diff
@@ -20,38 +20,173 @@ package ast
...
+	Description() string
+	String() string
+	MarshalJSON() ([]byte, error)
+	IsLessPermissiveThan(Access) bool
+}
+
+type Separator = string
+
+const (
+	Disjunction Separator = " |"
```
turbolent, line 42: Is the space a typo?

turbolent (commented):
Looks good!

Not a deal-breaker, but would be nice to keep access-related code out of the type parsing code.

runtime/parser/type.go
dsainati1: Extracted this out into a wrapper function.

runtime/ast/access.go
dsainati1: I do actually use it later on in one of the type checking PRs, so having it be a separate type is probably safer.

turbolent: In that case maybe it's best to keep the type definition, but remove the values, and define additional constants for the documents as described above and use them based on the constant

runtime/parser/declaration.go
```diff
@@ -269,12 +279,75 @@ var enumeratedAccessModifierKeywords = common.EnumerateWords(
 	"or",
 )
 
+func rejectAccessKeywords(p *parser, produceNominalType func() (*ast.NominalType, error)) (*ast.NominalType, error) {
```
turbolent, line 282: 👌

runtime/parser/expression.go
```diff
@@ -965,12 +965,41 @@ func parseAttachExpressionRemainder(p *parser, token lexer.Token) (*ast.AttachEx
...
+		if err != nil {
+			return nil, err
+		}
+
+		entitlements, _, err = parseNominalTypes(p, lexer.TokenParenClose, lexer.TokenComma)
+		for _, entitlement := range entitlements {
+			switch entitlement.Identifier.Identifier {
+			case KeywordAll, KeywordAccess, KeywordAccount, KeywordSelf:
+				return nil, p.syntaxError("unexpected non-nominal type: %s", entitlement)
+			}
```
turbolent, line 989: Can `rejectAccessKeywords` be used here?

runtime/ast/type.go
```diff
@@ -570,13 +574,26 @@ func (t *ReferenceType) EndPosition(memoryGauge common.MemoryGauge) Position {
...
 func (t *ReferenceType) Doc() prettier.Doc {
 	var doc prettier.Concat
-	if t.Authorized {
-		doc = append(doc, referenceTypeAuthKeywordSpaceDoc)
+	if t.Authorization != nil {
+		doc = append(doc, referenceTypeAuthKeywordDoc)
+		entitlementSet := t.Authorization.EntitlementSet
+		if entitlementSet != nil && len(entitlementSet.Entitlements()) > 0 {
+			entitlements := entitlementSet.Entitlements()
+			doc = append(doc, prettier.Text("("))
```
turbolent, line 587:
```suggestion
			// TODO: add indentation, improve separators. follow e.g. ParameterList.Doc()
			doc = append(doc, prettier.Text("("))
```

turbolent (approved):
Great work!

runtime/parser/expression.go
dsainati1: Yep!

turbolent (approved):

runtime/parser/keyword.go
```diff
@@ -129,6 +141,7 @@ var softKeywords = []string{
 	KeywordSet,
 	KeywordAll,
 	KeywordView,
+	keywordTo,
```
turbolent, line 144:
```suggestion
	keywordRemove,
```

runtime/parser/keyword.go
```diff
@@ -120,6 +125,13 @@ var allKeywords = []string{
 	KeywordDefault,
 	KeywordEnum,
 	KeywordView,
+	KeywordWith,
+	KeywordMapping,
+	KeywordRequire,
+	keywordRemove,
```
turbolent, line 131:
```suggestion
```

turbolent (commented):
Looks like `remove` should be made a soft keyword

runtime/parser/keyword.go
dsainati1: Hmm, I'm not sure it can be actually, because the soft keyword test then fails. If we see a `remove = 3` assignment, the parser starts parsing this as a `remove` statement rather than as an assignment. Unless I am misunderstanding then we need to make `remove` a hard keyword, and thus technically attachments would be a breaking change?

turbolent: Ah, I see. I'll refamiliarize myself with how we handle hard vs soft keywords at the moment and will see how we can handle these cases. We might have to improve the parsing function for the remove statement a bit

dsainati1: Going to handle this in a different PR, as it's not related to entitlements.

turbolent: We will address the attachment related code changes in a separate PR

turbolent:
@dsainati1 just one last thing: linting is reporting a missing check for an error return value

dsainati1:
Thanks, fixed

turbolent:
CI is passing so you can do the honours @dsainati1 :-)

Issue #2157, opened 2022-11-21
Cadence Compact Format (CCF) Specification RC and CCF Codec (fully self-describing mode)


fxamacker:
## Problem

Currently, Cadence external values (such as events) are encoded using JSON-Cadence Data Interchange Format. The JSON-based format is a human-readable text encoding that is verbose, inefficient, and doesn't define deterministic encoding (canonical format).

## Proposed Solution

Design and specify Cadence Compact Format (CCF) as a more compact, efficient, and deterministic alternative to the JSON-based format.

We can leverage an existing data format's data model to reduce complexity, cost, and risks to CCF specs and codecs.

Additionally, we can design and implement a CCF codec by using an existing codec (e.g. the same way COSE codec uses CBOR codec under the hood).

For more info about CCF requirements or design considerations, see
- [Compact Cadence Value External Encoding Requirements (notion)](https://www.notion.so/dapperlabs/Compact-Cadence-Value-External-Encoding-fe380872bc84434b875489de1b0ad407)
- [Draft of Cadence Compact Format (fxamacker/ccf_draft)](https://github.com/fxamacker/ccf_draft)
- [Cadence Compact Format (onflow/ccf)](https://github.com/onflow/ccf)

## Scope

CCF Specifications only specifies Cadence Compact Format.

It is outside the scope of CCF Specifications to specify individual CCF-based formats or protocols (e.g. events).

## Introduction

CCF is a data format that allows compact, efficient, and deterministic encoding of Cadence external values.

Cadence external values (e.g. events, transaction arguments, etc.) have been encoded using JSON-CDC, which is inefficient, verbose, and doesn't define deterministic encoding.

The same `FeesDeducted` event on the Flow blockchain can encode to:
- 298 bytes in JSON-CDC (minified).
- 118 bytes in CCF (fully self-describing mode).
- &nbsp;20 bytes in CCF (partially self-describing mode).

CCF defines all requirements for deterministic encoding (sort orders, smallest encoded forms, and Cadence-specific requirements) to allow CCF codecs implemented in different programming languages to produce the same deterministic encodings.

Some requirements (such as "Deterministic CCF Encoding Requirements") are defined as optional.  Each CCF-based format or protocol can have its specification state how CCF options are used.  This allows each protocol to balance tradeoffs such as compatibility, determinism, speed, encoded data size, etc.

[...]

PR #2258, opened 2023-01-19
Remove copyright years from file headers, update years in NOTICE

turbolent:
## Description

See https://github.com/curl/curl/pull/10205#issuecomment-1368971786 and https://dapperlabs.slack.com/archives/C02FFE3JCTA/p1673571497147069

______

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #2364, opened 2023-03-01
Add CCF codec implementing CCF Specification (RC1)

fxamacker:
## Description

Updates #2157 

[Cadence Compact Format (CCF)](https://github.com/fxamacker/ccf_draft) is a data format designed for compact, efficient, and deterministic encoding of Cadence external values.

CCF obsoletes [JSON-Cadence Data Interchange Format](https://developers.flow.com/cadence/json-cadence-spec) (JCDIF) for use cases that do not require JSON.

Unlike JCDIF, etc. CCF Specifications explicitly defines requirements for:
- well-formed encodings
- valid encodings
- deterministic encodings

CCF is hybrid data format. CCF-based messages can be:
- fully self-describing or
- partially self-describing.

Both CCF modes are more compact than JSON-based messages. CCF-based protocols can send Cadence metadata just once for all messages of that type. Malformed data can be detected without Cadence metadata and without creating Cadence objects.

CCF codec implements [CCF Specification (RC1)](https://github.com/fxamacker/ccf_draft/commit/4e8f42db29925bd15481301729eca1b52852dcb4), which is temporarily at [github.com/fxamacker/ccf_draft](https://github.com/fxamacker/ccf_draft). The CCF specs
will be hosted under [github.com/onflow](https://github.com/onflow/cadence) after it is updated, cleaned up, and reformatted.

### API

CCF codec provides the same API as the existing JSON-Cadence Data Interchange codec.

Given this, integration effort to replace the old codec should be minimal.

### PRELIMINARY COMPARISONS

Comparisons used 48,309 events from a single mainnet transaction.

There were 9 event types. To simplify benchmark code, the first event's value in each of the 9 event types was used.

CCF's partially self-describing mode (aka "detached" mode) would be even smaller than this (e.g. maybe less than 1/4 the size of JSON when Flow eventually supports detached mode).

#### SIZE COMPARISON

```
Encoding | Num Events | Encoded size | Comments
-------- | ---------- | ------------ | --------
JSON     |     48,309 |   13,858,836 | JSON-Cadence Data Interchange
CCF      |     48,309 |    6,159,931 | CCF in fully self-describing mode
```

#### CCF SPEED AND MEMORY COMPARISONS

This isn't apples to apples comparison.   JSON data isn't sorted, etc.
- CCF encoder sorts data to encode event data deterministically.
- CCF decoder also verifies event data is sorted, well-formed, and valid.
- CCF encoding is less than 1/2 size of JSON-Cadence Data Interchange.

```
$ benchstat bench_json_events_48k.log bench_ccf_events_48k.log 
goos: linux
goarch: amd64
pkg: github.com/onflow/cadence/encoding/ccf
cpu: 13th Gen Intel(R) Core(TM) i5-13600K
                     │ bench_json_events_48k.log │      bench_ccf_events_48k.log       │
                     │          sec/op           │   sec/op     vs base                │
EncodeBatchEvents-20                 96.61m ± 4%   70.73m ± 3%  -26.79% (p=0.000 n=10)
DecodeBatchEvents-20                 647.7m ± 3%   157.5m ± 3%  -75.68% (p=0.000 n=10)
geomean                              250.1m        105.5m       -57.81%

                     │ bench_json_events_48k.log │       bench_ccf_events_48k.log       │
                     │           B/op            │     B/op      vs base                │
EncodeBatchEvents-20                32.45Mi ± 0%   25.82Mi ± 0%  -20.45% (p=0.000 n=10)
DecodeBatchEvents-20               234.97Mi ± 0%   56.16Mi ± 0%  -76.10% (p=0.000 n=10)
geomean                             87.32Mi        38.08Mi       -56.39%

                     │ bench_json_events_48k.log │      bench_ccf_events_48k.log       │
                     │         allocs/op         │  allocs/op   vs base                │
EncodeBatchEvents-20                 756.6k ± 0%   370.4k ± 0%  -51.05% (p=0.000 n=10)
DecodeBatchEvents-20                 4.746M ± 0%   1.288M ± 0%  -72.86% (p=0.000 n=10)
geomean                              1.895M        690.7k       -63.55%
```

Benchmarked using Go 1.19.6, linux_amd64, i5-13600k.  Results are subject to change because CCF codec reference implementation in Go has not yet been reviewed.

### NEXT STEPS

:point_right: After this PR is merged, there's additional work that should be completed before using or deploying this codec.

- [x] Add more tests to  handle edge cases (as time allows).  Currently, `go test -cover` reports 77.3% for CCF codec.

- [x] Add and run fuzz tests.  It would be very unusual for fuzzing to not find any problems with a brand new codec for a new data format.  I think Cadence Team is best qualified for this task (i.e. expert in Cadence types & values).

- [ ] Integration with FVM to use CCF codec for events.

- [ ] Integration work related to converting CCF encodings to JSON-Cadence Data Interchange.

- [ ] Add integration tests in relevant projects.

EDIT: updated benchmarks to match commit f911063ca19956976ac3d9f15bc5a35bfaa0cf91

______


encoding/ccf/encode.go
```diff
@@ -0,0 +1,1751 @@
...
+//	/ simple-value
+//	/ optional-value
+//	/ array-value
+//	/ dict-value
+//	/ composite-value
+//	/ path-value
+//	/ capability-value
+//	/ function-value
+//	/ type-value
+//
```
fxamacker, line 284: The CCF codec includes code comments that describe parts of [CCF Specification (RC1)](https://github.com/fxamacker/ccf_draft/commit/4e8f42db29925bd15481301729eca1b52852dcb4).  Two notations from RFC 8610 are used:

- [Concise Data Definition Language (CDDL)](https://www.rfc-editor.org/rfc/rfc8610.html) notation is used to document encoding/decoding functions and can help keep the codec and specs in sync.

- [Extended Diagnostic Notation (EDN)](https://www.rfc-editor.org/rfc/rfc8610.html#appendix-G) is used to document parts of unit tests.

encoding/ccf/ccf_test.go
```diff
@@ -0,0 +1,8208 @@
...
+			// array, 2 items follow
+			0x82,
+			// tag
+			0xd8, ccf.CBORTagSimpleType,
+			// void type ID (50)
+			0x18, 0x32,
+			// nil
+			0xf6,
+		},
+	)
```
fxamacker, line 72: All relevant tests from JSON-Cadence Data Interchange codec were ported to CCF codec.

All round-trip tests compare generated CCF encoding to expected binary result.  Each binary result is documented with matching JSON encoding as well as CCF encoding represented in Extended Diagnostic Notation.

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1751 @@
...
+// An Encoder converts Cadence values into CCF-encoded bytes.
+type Encoder struct {
+	// CCF codec uses CBOR codec under the hood.
+	enc *cbor.StreamEncoder
+}
+
+// Encode returns the CCF-encoded representation of the given value.
+//
+// This function returns an error if the Cadence value cannot be represented in CCF.
+func Encode(value cadence.Value) ([]byte, error) {
```
fxamacker, line 58: CCF encoder uses the same API as the encoder for JSON-Cadence Data Interchange:
- `Encode(cadence.Value)`
- `MustEncode(cadence.Value)`
- `NewEncoder(io.Writer)`
- `Encoder.Encode(cadence.Value)`

encoding/ccf/decode.go
```diff
@@ -0,0 +1,1879 @@
...
+	// CCF codec uses CBOR codec under the hood.
+	dec   *cbor.StreamDecoder
+	gauge common.MemoryGauge
+}
+
+// Decode returns a Cadence value decoded from its CCF-encoded representation.
+//
+// This function returns an error if the bytes represent CCF that is malformed,
+// invalid, or does not comply with requirements in the CCF specification.
+func Decode(gauge common.MemoryGauge, b []byte) (cadence.Value, error) {
```
fxamacker, line 61: CCF decoder uses the same API as the decoder for JSON-Cadence Data Interchange:
- `Decode(common.MemoryGauge, []byte)`
- `NewDecoder(common.MemoryGauge, []byte)`
- `Decoder.Decode()`

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+	dictionaryType := v.DictionaryType.(*cadence.DictionaryType)
+	if dictionaryType == nil {
+		return fmt.Errorf("unexpected dictionary type %s", v.DictionaryType.ID())
+	}
+
+	staticKeyType := dictionaryType.KeyType
+	staticElementType := dictionaryType.ElementType
+
+	// Encode array head with array size of 2 * number of pairs.
+	err := e.enc.EncodeArrayHead(uint64(len(v.Pairs) * 2))
```
turbolent, line 720:
```suggestion
	err := e.enc.EncodeArrayHead(uint64(len(v.Pairs)) * 2)
```

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+	defer putBuffer(buf)
+
+	// Encode and sort key value pairs.
+	sortedPairs, err := encodeAndSortKeyValuePairs(buf, v, tids)
+	if err != nil {
+		return err
+	}
+
+	// Encode array head with 2 * number of pairs.
+	err = e.enc.EncodeArrayHead(uint64(len(v.Pairs) * 2))
```
turbolent, line 763:
```suggestion
	err = e.enc.EncodeArrayHead(uint64(len(v.Pairs)) * 2)
```

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+	if err != nil {
+		return err
+	}
+
+	for i := 0; i < len(sortedPairs); i++ {
+		// Encode key and value.
+		err = e.enc.EncodeRawBytes(sortedPairs[i].encodedPair)
```
turbolent, line 770:
```suggestion
	for _, sortedPair := range sortedPairs {
		// Encode key and value.
		err = e.enc.EncodeRawBytes(sortedPair.encodedPair)
```

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+	staticElementType := dictionaryType.ElementType
+
+	// Encode array head with array size of 2 * number of pairs.
+	err := e.enc.EncodeArrayHead(uint64(len(v.Pairs) * 2))
+	if err != nil {
+		return err
+	}
+
+	for i := 0; i < len(v.Pairs); i++ {
+		pair := v.Pairs[i]
```
turbolent, line 726:
```suggestion
	for _, pair := range v.Pairs {
```

encoding/ccf/bench_test.go
```diff
@@ -0,0 +1,215 @@
+/*
+ * Cadence - The resource-oriented smart contract programming language
+ *
+ * Copyright 2022-2023 Dapper Labs, Inc.
```
SupunS, line 4: We decided to omit the copyright year in the license text (and instead already added it to the NOTICE file): https://github.com/onflow/cadence/pull/2258

```suggestion
 * Copyright Dapper Labs, Inc.
```

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+			len(staticFieldTypes),
+		))
+	}
+
+	// Encode array head with number of fields.
+	err := e.enc.EncodeArrayHead(uint64(len(fields)))
+	if err != nil {
+		return err
+	}
+
```
turbolent, line 904: Maybe also short-circuit for `len(fields) == 0`, i.e. just return. Maybe just `switch` on `len(fields)`

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+		return e.encodeCapabilityTypeValue(typ, visited)
+
+	case *cadence.EnumType:
+		return e.encodeEnumTypeValue(typ, visited)
+
+	case nil:
+		return e.encodeNilTypeValue()
+
+	default:
+		panic(fmt.Errorf("unsupported type value %s (%T)", typ.ID(), typ))
```
turbolent, line 1123:
```suggestion
		panic(errors.NewUnexpectedError("unsupported type value %s (%T)", typ.ID(), typ))
```

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+}
+
+// encodeFieldTypeValues encodes composite field types as
+// language=CDDL
+//
+//	fields: [
+//	    + [
+//	        name: tstr,
+//	        type: type-value
+//	    ]
```
turbolent, line 1513: Same here: Should this be "zero or more"? Maybe already a mistake in the spec

```suggestion
//	    * [
//	        name: tstr,
//	        type: type-value
//	    ]
```

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+//	]
+func (e *Encoder) encodeFieldTypeValues(fieldTypes []cadence.Field, visited ccfTypeIDByCadenceType) error {
+	// Encode CBOR array head with number of field types.
+	err := e.enc.EncodeArrayHead(uint64(len(fieldTypes)))
+	if err != nil {
+		return err
+	}
+
```
turbolent, line 1521: Maybe also short-circuit for `len(fieldTypes) == 0`, i.e. just return (and avoid allocation)

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+//	        type: type-value
+//	    ]
+//	]
+func (e *Encoder) encodeParameterTypeValues(parameterTypes []cadence.Parameter, visited ccfTypeIDByCadenceType) error {
+	// Encode CBOR array head with number of parameter types.
+	err := e.enc.EncodeArrayHead(uint64(len(parameterTypes)))
+	if err != nil {
+		return err
+	}
+
```
turbolent, line 1626: Maybe short-circuit for `len(parameterTypes) == 0`, i.e. just return (and avoid allocation for sorting)

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+		return false
+	}
+	// Here, static type is different from runtime type.
+	// Handle special case of runtime type being OptionalType{NeverType}.
+	if _, ok := staticType.(*cadence.OptionalType); ok {
+		if isOptionalNeverType(runtimeType) {
+			return false
+		}
+	}
+	return true
```
turbolent, line 1732: Not sure, but why do we have to handle `Never?` specially?

encoding/ccf/ccf_type_id.go
```diff
@@ -0,0 +1,77 @@
...
+	"math/big"
+
+	"github.com/onflow/cadence"
+)
+
+// ccfTypeID represents CCF type ID.
+type ccfTypeID uint64
+
+func newCCFTypeID(b []byte) ccfTypeID {
+	return ccfTypeID(new(big.Int).SetBytes(b).Uint64())
```
SupunS, line 32: What if we use `binary.BigEndian.Uint64()` here? would that save extra allocations of  `new(big.Int)`.
Same question on the `Bytes()` method.

encoding/ccf/decode_type.go
```diff
@@ -0,0 +1,615 @@
...
+		}
+
+		// "Valid CCF Encoding Requirements" in CCF specs:
+		//
+		//   "Elements MUST be unique in restricted-type or restricted-type-value."
+		if _, ok := restrictionTypeIDs[restrictedType.ID()]; ok {
+			return nil, fmt.Errorf("found duplicate restricted type %s", restrictedType.ID())
+		}
+
+		// "Deterministic CCF Encoding Requirements" in CCF specs:
```
turbolent, line 544: For a follow-up PR, in general: Maybe we can make the deterministic encoding checking in the decoder optional? Most of the time we do not need, i.e. we could avoid the associated overhead

encoding/ccf/decode.go
```diff
@@ -0,0 +1,1887 @@
...
+//
+// This function returns an error if the bytes represent CCF that is malformed,
+// invalid, or does not comply with requirements in the CCF specification.
+func (d *Decoder) Decode() (value cadence.Value, err error) {
+	// Capture panics that occur during decoding.
+	defer func() {
+		// Recover panic error if there is any.
+		if r := recover(); r != nil {
+			// Don't recover Go errors.
+			goErr, ok := r.(goRuntime.Error)
```
turbolent, line 94: Also handle `InternalError`

encoding/ccf/decode.go
```diff
@@ -0,0 +1,1887 @@
...
+func (d *Decoder) decodeInt8() (cadence.Value, error) {
+	i, err := d.dec.DecodeInt64()
+	if err != nil {
+		return nil, err
+	}
+	if i < math.MinInt8 || i > math.MaxInt8 {
+		return nil, fmt.Errorf(
+			"encoded int8-value %d is outside range of Int8 [%d, %d]",
+			i,
+			math.MaxInt8,
```
turbolent, line 479:
```suggestion
			math.MinInt8,
```

encoding/ccf/decode.go
```diff
@@ -0,0 +1,1887 @@
...
+// int128-value = bigint
+func (d *Decoder) decodeInt128() (cadence.Value, error) {
+	bigInt, err := d.dec.DecodeBigInt()
+	if err != nil {
+		return nil, err
+	}
+	return cadence.NewMeteredInt128FromBig(
+		d.gauge,
+		func() *big.Int {
+			return bigInt
```
turbolent, line 546: Here and elsewhere: The decoding above should happen inside this function, so the memory is metered before the decoding

encoding/ccf/decode.go
```diff
@@ -0,0 +1,1887 @@
...
+func (d *Decoder) decodeUInt() (cadence.Value, error) {
+	bigInt, err := d.dec.DecodeBigInt()
+	if err != nil {
+		return nil, err
+	}
+	if bigInt.Sign() < 0 {
+		return nil, fmt.Errorf(
+			"encoded uint-value %s is negative",
+			bigInt.String(),
+		)
```
turbolent, line 579: Here and below: The value might be very large, so formatting it as a string might allocate a large amount of memory:

```suggestion
		return nil, errors.New("encoded uint-value is negative")
```

encoding/ccf/decode_type.go
```diff
@@ -0,0 +1,615 @@
...
+		// Decode restriction.
+		restrictedType, err := decodeTypeFn(types)
+		if err != nil {
+			return nil, err
+		}
+
+		// "Valid CCF Encoding Requirements" in CCF specs:
+		//
+		//   "Elements MUST be unique in restricted-type or restricted-type-value."
+		if _, ok := restrictionTypeIDs[restrictedType.ID()]; ok {
```
SupunS, line 540: Maybe extract `restrictedType.ID()` to a variable and re-use it here and below. Could save a few nil checks (for cached types IDs)

encoding/ccf/decode.go
```diff
@@ -0,0 +1,1887 @@
...
+// language=CDDL
+// array-value = [* value]
+func (d *Decoder) decodeArray(typ cadence.ArrayType, hasKnownSize bool, knownSize uint, types cadenceTypeByCCFTypeID) (cadence.Value, error) {
+	// Decode array length.
+	n, err := d.dec.DecodeArrayHead()
+	if err != nil {
+		return nil, err
+	}
+
+	if hasKnownSize && uint64(knownSize) != n {
```
turbolent, line 827: Maybe just change the parameter to `knownSize uint64`

```suggestion
	if hasKnownSize && knownSize != n {
```

encoding/ccf/decode.go
```diff
@@ -0,0 +1,1887 @@
...
+		func() ([]cadence.KeyValuePair, error) {
+			// "Valid CCF Encoding Requirements" in CCF specs:
+			//
+			//   "Keys MUST be unique in dict-value. Decoders are not always required to check
+			//   for duplicate dictionary keys. In some cases, checking for duplicate dictionary
+			//   key is not necessary or it may be delegated to the application."
+			//
+			// Here, decoder doesn't check uniqueness of dictionary keys
+			// because checking is delegated (entrusted) to Cadence runtime.
+			return pairs, nil
```
turbolent, line 930: The allocation of `[]cadence.KeyValuePair` should occur in this closure, so that memory is metered before the memory is actually allocated

encoding/ccf/decode.go
```diff
@@ -0,0 +1,1887 @@
...
+		}
+		fieldValues[i] = field
+	}
+
+	return fieldValues, nil
+}
+
+// decodeStruct decodes encoded composite-value as
+// language=CDDL
+// composite-value = [+ (field: value)]
```
turbolent, line 973: Here and for the other composite value decoding functions: Should this be "zero or more"? Maybe also check encoding function. Might also be an issue in the spec 

```suggestion
// composite-value = [* (field: value)]
```

turbolent (commented):
Awesome work! 👏 

I reviewed more or less everything except for the tests. Will do more review passes after the first set of feedback has been addressed.

encoding/ccf/decode_type.go
```diff
@@ -0,0 +1,615 @@
...
+		return d.decodeRestrictedType(types, d.decodeInlineType)
+
+	case CBORTagCapabilityType:
+		return d.decodeCapabilityType(types, d.decodeInlineType)
+
+	case CBORTagTypeRef:
+		return d.decodeTypeRef(types)
+
+	default:
+		return nil, fmt.Errorf("unsupported encoded inline type with CBOR tag number %d", tagNum)
```
SupunS, line 84: Maybe use `errors.NewDefaultUserError()` for all decoding errors. I think it's safe to categorize any invalid input as a user error (from the decoder's POV).

SupunS (commented):
Reviewed up to decoding. Will review the remaining (encoding) on the second batch.

encoding/ccf/bench_test.go
fxamacker: Done in commit [82e1812](https://github.com/onflow/cadence/pull/2364/commits/82e1812f11a327cea381fb37e155726ee4a41f3c).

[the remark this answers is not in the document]
fxamacker: Done in commit [82e1812](https://github.com/onflow/cadence/pull/2364/commits/82e1812f11a327cea381fb37e155726ee4a41f3c).

encoding/ccf/encode_typedef.go
```diff
@@ -0,0 +1,252 @@
...
+		return e.encodeCompositeTypeField(fieldTypes[0], tids)
+	}
+
+	// "Deterministic CCF Encoding Requirements" in CCF specs:
+	//
+	//   "composite-type.fields MUST be sorted by name"
+	sortedIndexes := getSortedFieldIndex(typ)
+
+	for i := 0; i < len(sortedIndexes); i++ {
+		index := sortedIndexes[i]
```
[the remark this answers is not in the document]
fxamacker: Done in commit 2ce376a2b6267226e8ae337c5f0f22e5b7a18dc2.

encoding/ccf/traverse_value.go
```diff
@@ -0,0 +1,229 @@
...
+		// No need to reassign ccf id, nor sort types.
+		return ct.types, ct.ids
+	}
+
+	// Sort Cadence types by Cadence type ID.
+	sort.Sort(bytewiseCadenceTypeInPlaceSorter(ct.types))
+
+	// Assign sorted array index as local ccf ID.
+	for i := 0; i < len(ct.types); i++ {
+		ct.ids[ct.types[i].ID()] = ccfTypeID(i)
```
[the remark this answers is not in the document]
fxamacker: Done in commit 2ce376a2b6267226e8ae337c5f0f22e5b7a18dc2.

encoding/ccf/traverse_value.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// Traverse v's elements for runtime types.
+	switch x := v.(type) {
+
+	case cadence.Optional:
+		ct.traverseValue(x.Value)
+
+	case cadence.Array:
+		for i := 0; i < len(x.Values); i++ {
+			ct.traverseValue(x.Values[i])
```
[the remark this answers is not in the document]
fxamacker: Done in commit 2ce376a2b6267226e8ae337c5f0f22e5b7a18dc2.

encoding/ccf/encode_type.go
```diff
@@ -0,0 +1,512 @@
...
+	// "Deterministic CCF Encoding Requirements" in CCF specs:
+	//
+	//   "restricted-type.restrictions MUST be sorted by restriction's cadence-type-id"
+	//   "restricted-type-value.restrictions MUST be sorted by restriction's cadence-type-id."
+	sorter := newBytewiseCadenceTypeSorter(restrictions)
+
+	sort.Sort(sorter)
+
+	for i := 0; i < len(sorter.indexes); i++ {
+		index := sorter.indexes[i]
```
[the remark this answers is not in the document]
fxamacker: Done in commit 2ce376a2b6267226e8ae337c5f0f22e5b7a18dc2.

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+//	  )]
+func (e *Encoder) encodeTypeDefs(types []cadence.Type, tids ccfTypeIDByCadenceType) error {
+	// Encode array head with number of type definitions.
+	err := e.enc.EncodeArrayHead(uint64(len(types)))
+	if err != nil {
+		return err
+	}
+
+	for i := 0; i < len(types); i++ {
+		typ := types[i]
```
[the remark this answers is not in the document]
fxamacker: Done in commit 2ce376a2b6267226e8ae337c5f0f22e5b7a18dc2.

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+
+	// Encode array head with number of array elements.
+	err := e.enc.EncodeArrayHead(uint64(len(v.Values)))
+	if err != nil {
+		return err
+	}
+
+	for i := 0; i < len(v.Values); i++ {
+		// Encode element as value.
+		err = e.encodeValue(v.Values[i], staticElementType, tids)
```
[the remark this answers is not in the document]
fxamacker: Done in commit 2ce376a2b6267226e8ae337c5f0f22e5b7a18dc2.

encoding/ccf/encode.go
fxamacker: Done in commit 2ce376a2b6267226e8ae337c5f0f22e5b7a18dc2.

encoding/ccf/encode.go
fxamacker: Done in commit 2ce376a2b6267226e8ae337c5f0f22e5b7a18dc2.

encoding/ccf/decode.go
```diff
@@ -0,0 +1,1887 @@
...
+	}
+
+	switch tagNum {
+	case CBORTagTypeDefAndValue:
+		// Decode ccf-type-and-value-message.
+		return d.decodeTypeDefAndValue()
+
+	case CBORTagTypeAndValue:
+		// Decode ccf-typedef-and-value-message.
+		return d.decodeTypeAndValue(cadenceTypeByCCFTypeID{})
```
turbolent, line 126:
```suggestion
	case CBORTagTypeDefAndValue:
		// Decode ccf-typedef-and-value-message.
		return d.decodeTypeDefAndValue()

	case CBORTagTypeAndValue:
		// Decode ccf-type-and-value-message.
		return d.decodeTypeAndValue(cadenceTypeByCCFTypeID{})
```

encoding/ccf/encode_type.go
```diff
@@ -0,0 +1,512 @@
...
+
+	// element 2: restrictions as array.
+
+	// Encode array head with number of restrictions.
+	restrictions := typ.Restrictions
+	err = e.enc.EncodeArrayHead(uint64(len(restrictions)))
+	if err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
fxamacker: Done in commit 266fb6a280996e376da1e56ac228bd006ee5c2b0 to match today's change to CCF specs to allow "zero or more" (was "one or more").

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+//
+//	id: id,
+//	cadence-type-id: cadence-type-id,
+//	; type is only used by enum type value
+//	type: nil / type-value,
+//	fields: [
+//	    + [
+//	        name: tstr,
+//	        type: type-value
+//	    ]
```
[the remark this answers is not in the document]
fxamacker: Done in commits f89eb0615b95c5bb5deeeed7966fa0d6fa98e9d0 and 42d865792c152b0af33991e4ad5d1b6adf796249.  

Codec now matches today's updates to CCF specs.  Changes to CCF specs include:
- https://github.com/fxamacker/ccf_draft/pull/66
- https://github.com/fxamacker/ccf_draft/pull/67

encoding/ccf/encode.go
fxamacker: Done in commit 266fb6a280996e376da1e56ac228bd006ee5c2b0 to match today's change to CCF specs to allow "zero or more" (was "one or more").

encoding/ccf/encode.go
fxamacker: Done in commits f89eb0615b95c5bb5deeeed7966fa0d6fa98e9d0 and 42d865792c152b0af33991e4ad5d1b6adf796249.  

Codec now matches today's updates to CCF specs.  Changes to CCF specs include:
- https://github.com/fxamacker/ccf_draft/pull/66
- https://github.com/fxamacker/ccf_draft/pull/67

encoding/ccf/encode.go
fxamacker: Done in commit 266fb6a280996e376da1e56ac228bd006ee5c2b0 to match today's change to CCF specs to allow "zero or more" (was "one or more").

encoding/ccf/encode.go
fxamacker: Done in commit 266fb6a280996e376da1e56ac228bd006ee5c2b0.

encoding/ccf/decode.go
fxamacker: Done in commits f89eb0615b95c5bb5deeeed7966fa0d6fa98e9d0 and 42d865792c152b0af33991e4ad5d1b6adf796249.  

Codec now matches today's updates to CCF specs.  Changes to CCF specs include:
- https://github.com/fxamacker/ccf_draft/pull/66
- https://github.com/fxamacker/ccf_draft/pull/67

encoding/ccf/decode_type.go
fxamacker: Yes, I agree.

We don't need to do it here because `Decoder.Decode()` already wraps returned error inside `DefaultUserError` like this:

https://github.com/onflow/cadence/blob/266fb6a280996e376da1e56ac228bd006ee5c2b0/encoding/ccf/decode.go#L108-L110

encoding/ccf/encode_type.go
```diff
@@ -0,0 +1,512 @@
...
+	case *cadence.ConstantSizedArrayType:
+		return e.encodeConstantSizedArrayType(typ, tids)
+
+	case *cadence.DictionaryType:
+		return e.encodeDictType(typ, tids)
+
+	case cadence.CompositeType, cadence.InterfaceType:
+		id, err := tids.id(typ)
+		if err != nil {
+			panic(err)
```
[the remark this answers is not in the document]
fxamacker: Yes, done in commit 4aa5f6b302b01140ec109f5c4eb5745137c1a2b4.  

I also updated all encoder's internal panics to throw `UnexpectedError`.

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+//
+// This function returns an error if the given value's type is not supported
+// by the encoder.
+func (e *Encoder) Encode(value cadence.Value) (err error) {
+	// capture panics
+	defer func() {
+		// Recover panic error if there is any.
+		if r := recover(); r != nil {
+			// Don't recover Go errors.
+			goErr, ok := r.(goRuntime.Error)
```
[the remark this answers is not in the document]
fxamacker: Done in commit 4aa5f6b302b01140ec109f5c4eb5745137c1a2b4.

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+			}
+
+			err = panicErr
+		}
+
+		// Add context to error if there is any.
+		if err != nil {
+			err = fmt.Errorf(
+				"ccf: failed to encode value %q (type %q): %s",
+				value.String(),
```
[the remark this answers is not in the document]
fxamacker: Good point!  Done in commit 4aa5f6b302b01140ec109f5c4eb5745137c1a2b4.

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+	v cadence.Value,
+	staticType cadence.Type,
+	tids ccfTypeIDByCadenceType,
+) error {
+
+	runtimeType := v.Type()
+
+	// CCF requires value to have non-nil type.
+	if runtimeType == nil {
+		panic(errors.New("value has nil type"))
```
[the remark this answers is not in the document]
fxamacker: Done in commit 4aa5f6b302b01140ec109f5c4eb5745137c1a2b4.

encoding/ccf/encode.go
fxamacker: Done in commit 4aa5f6b302b01140ec109f5c4eb5745137c1a2b4.

encoding/ccf/decode.go
fxamacker: Good catch!  Done in bc220dce48e3811da284fd94fd47e5be21c810de.

encoding/ccf/decode.go
fxamacker: Good point!  Done in bc220dce48e3811da284fd94fd47e5be21c810de.

I also updated all related error messages for `big.Int`.

encoding/ccf/consts.go
```diff
@@ -0,0 +1,144 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ccf
+
+// CCF uses CBOR tag numbers 128-255, which are unassigned by [IANA]
+// (https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml).
+
+const (
```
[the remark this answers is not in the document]
fxamacker: Good idea!  Done in commit cae3f98820310ae8761975694a938f51fe155919 with some changes and additions.

encoding/ccf/decode_type.go
fxamacker: Good point!  Done in commit ef5f6d95271ef9fbd10a86e7019690660944ae00.

encoding/ccf/bench_test.go
```diff
@@ -0,0 +1,215 @@
...
+		for _, bm := range batchBenchmarks {
+			for i := 0; i < bm.count; i++ {
+				encoded, err = json.Encode(bm.value)
+				encodedSize += len(encoded)
+			}
+		}
+	}
+	require.NoError(b, err)
+
+	// fmt.Printf("Batch events encoded in JSON are %d bytes\n", encodedSize)
```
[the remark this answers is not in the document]
fxamacker: Done in commit 8791232f56fcd97771f3d9a255b524e73624fc9f.

encoding/ccf/encode_typedef.go
```diff
@@ -0,0 +1,252 @@
...
+		return err
+	}
+
+	// Encode array head with length 2.
+	err = e.enc.EncodeArrayHead(2)
+	if err != nil {
+		return err
+	}
+
+	// element 0: CCf type ID
```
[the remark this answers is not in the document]
fxamacker: Done in commit 8791232f56fcd97771f3d9a255b524e73624fc9f.

encoding/ccf/encode.go
fxamacker: Done in commit 8791232f56fcd97771f3d9a255b524e73624fc9f.

encoding/ccf/encode.go
fxamacker: Done in commit 8791232f56fcd97771f3d9a255b524e73624fc9f.

encoding/ccf/decode.go
fxamacker: Done in commit 8791232f56fcd97771f3d9a255b524e73624fc9f.

encoding/ccf/decode.go
fxamacker: Great catch!  Done in commit 8791232f56fcd97771f3d9a255b524e73624fc9f.

encoding/ccf/decode.go
fxamacker: Good catch!  Done in commit 588594e75d0d5a0632cf3d18d395041ceb171f91 for here and elsewhere.

encoding/ccf/decode.go
fxamacker: Good catch!  Done in commit 2e6e4a4cd53ad48abd3ae5f12080656cceca4c9b.

encoding/ccf/cache.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"sort"
+	"sync"
+
+	"github.com/onflow/cadence"
+)
+
+// cachedSortedFieldIndex contains sorted field index of Cadence composite types.
+var cachedSortedFieldIndex sync.Map // key: cadence.CompositeType, value: []int
```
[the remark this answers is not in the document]
fxamacker: @turbolent `cachedSortedFieldIndex` is global so it can be reused across multiple CCF encodings.  However, `cadence.CompositeType` may not be the best choice as map key.

Do you think it is useful to have a global cache of sorted composite field indexes across multiple CCF encodings?  If so, does composite type (with the same Cadence type ID) always have the same unsorted field order?

encoding/ccf/ccf_type_id.go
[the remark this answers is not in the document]
fxamacker: There are two reasons for using `big.Int` under the hood even though `ccfTypeID` is `uint64`:
- `ccfTypeID` can be changed to `big.Int` type if we decide to use hash digest as CCF ID (less important).
- using `big.Int` allows smaller encoded size vs `binary.BigEndian` which is always 8 bytes (more important).

Also `new(big.Int)` here and in `Bytes()` don't escape to heap.

I tried a custom function without `big.Int` to encode to same small size, but there was no difference in memory and allocations.  Speed was also comparable as well.

fxamacker: Bastian asked a similar question. :smile:

There are two reasons for using `big.Int` under the hood even though `ccfTypeID` is `uint64`:
- `ccfTypeID` can be changed to `big.Int` type if we decide to use hash digest as CCF ID (less important).
- using `big.Int` allows smaller encoded size vs `binary.BigEndian` which is always 8 bytes (more important).

Also `new(big.Int)` here and in `Bytes()` don't escape to heap.

I tried a custom function without `big.Int` to encode to same small size, but there was no difference in memory and allocations.  Speed was also comparable as well.

encoding/ccf/ccf_type_id.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (types ccfTypeIDByCadenceType) id(t cadence.Type) (ccfTypeID, error) {
+	id, ok := types[t.ID()]
+	if !ok {
+		return 0, fmt.Errorf("CCF type ID not found for type %s", t.ID())
+	}
+	return id, nil
+}
+
+type cadenceTypeByCCFTypeID map[ccfTypeID]cadence.Type
```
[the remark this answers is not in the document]
fxamacker: Yes, we can use a slice here for sequentially generated CCF ID.  Unfortunately, when we switch to UUID as CCF ID, we need to use a map.  So I thought this can save some work down the road.

encoding/ccf/encode_typedef.go
```diff
@@ -0,0 +1,252 @@
...
+//	cadence-type-id: cadence-type-id,
+//	fields: [
+//	    + [
+//	        field-name: tstr,
+//	        field-type: inline-type
+//	    ]
+//	]
+//
+// ]
+func (e *Encoder) encodeCompositeType(typ cadence.CompositeType, tids ccfTypeIDByCadenceType) error {
```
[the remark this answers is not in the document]
fxamacker: I thought the same too and previously tried to include `tids ccfTypeIDByCadenceType` as part of encoder.  

However, it didn't fit in neatly with encoder API because:
- CCF IDs in `ccfTypeIDByCadenceType` is currently sequentially generated and is specific to encoded value,
- `Encoder` API allows encoding multiple values though `encoder.Encode()`.

So I made `tids ccfTypeIDByCadenceType` specific to encoded value, instead of encoder.

encoding/ccf/sort.go
```diff
@@ -0,0 +1,207 @@
...
+// bytewiseFieldSorter is used to sort fields by identifier.
+// NOTE: in order to avoid making a copy of fields for sorting,
+//   - create sorter by calling newBytewiseFieldIdentifierSorter()
+//   - sort by calling sort.Sort(sorter)
+//   - iterate sorted fields by
+//     for _, index := range sorter.indexes {
+//     // process sorted field at fields[index]
+//     }
+type bytewiseFieldSorter struct {
+	// NOTE: DON'T sort fields because it isn't a copy.
```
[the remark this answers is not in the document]
fxamacker: Done in commit 9d172c996d5dc692ba465690a1744428a6806c3f here and in several other places.

Basically, `bytewiseFieldSorter` shouldn't sort fields **in place** because sorter doesn't make a copy of fields.  Instead it should sort `indexes` by field identifier.  Caller should iterate over sorted `indexes`.

```Go
sorter := newBytewiseFieldIdentifierSorter(fields)
sort.Sort(sorter)
for _, index := range sorter.indexes {
        // process sorted field at fields[index]
}
```

encoding/ccf/encode_type.go
```diff
@@ -0,0 +1,512 @@
...
+}
+
+// encodeSimpleType encodes cadence simple type as
+// language=CDDL
+// simple-type =
+//
+//	; cbor-tag-simple-type
+//	#6.137(simple-type-id)
+func (e *Encoder) encodeSimpleType(id uint64) error {
+	rawTagNum := []byte{0xd8, CBORTagSimpleType}
```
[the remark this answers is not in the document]
fxamacker: Go 1.19.7 optimizes this to not allocate because `rawTagNum` doesn't escape.

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+		return e.encodeEvent(x, tids)
+
+	case cadence.Contract:
+		return e.encodeContract(x, tids)
+
+	case cadence.Path:
+		return e.encodePath(x)
+
+	case cadence.TypeValue:
+		return e.encodeTypeValue(x.StaticType, ccfTypeIDByCadenceType{})
```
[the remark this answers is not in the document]
fxamacker: Done in commit 36b52d7949f6ab2557f4132490d468b8478d6477.

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+
+	case cadence.StorageCapability:
+		return e.encodeCapability(x)
+
+	case cadence.Enum:
+		return e.encodeEnum(x, tids)
+
+	case cadence.Function:
+		return e.encodeFunction(x.FunctionType, ccfTypeIDByCadenceType{})
```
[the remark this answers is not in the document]
fxamacker: Yes, added comments in commit 36b52d7949f6ab2557f4132490d468b8478d6477.

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+// encodeDictionary encodes cadence.Dictionary as
+// language=CDDL
+// dict-value = [* (key: value, value: value)]
+func (e *Encoder) encodeDictionary(v cadence.Dictionary, tids ccfTypeIDByCadenceType) error {
+	if len(v.Pairs) > 1 {
+		return e.encodeSortedDictionary(v, tids)
+	}
+
+	dictionaryType := v.DictionaryType.(*cadence.DictionaryType)
+	if dictionaryType == nil {
```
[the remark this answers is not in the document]
fxamacker: Opened PR #2388 to fix some old issues outside the codec:
- change type of `cadence.Dictionary.DictionaryType` from `Type` to `*cadence.DictionaryType`
- changed `Type()` for 8 external values to check and return nil Type, instead of returning Type referencing nil pointer

I'll update this block of code after that PR is merged.

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1766 @@
...
+	return e.enc.EncodeNil()
+}
+
+// encodeOptional encodes cadence.Optional as
+// language=CDDL
+// optional-value = nil / value
+func (e *Encoder) encodeOptional(v cadence.Optional, tids ccfTypeIDByCadenceType) error {
+	switch innerValue := v.Value.(type) {
+	case cadence.Optional:
+		return e.encodeOptional(innerValue, tids)
```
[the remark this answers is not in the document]
fxamacker: Good catch!  I simplified `encodeOptional` in commit 705fdd57ffb42d9f68fe3904e2fbc0aace362751.

encoding/ccf/encode.go
fxamacker: We handle special case of `Optional{nil}` because its runtime type is `OptionalType{NeverType}` while its static type can be different, such as `OptionalType{AddressType}`.

In this [test example](https://github.com/onflow/cadence/blob/705fdd57ffb42d9f68fe3904e2fbc0aace362751/encoding/ccf/ccf_test.go#L7925-L7939), field `to`'s type is `OptionalType{AddressType}` and its value is nil with runtime type `OptionalType{NeverType}`.  Even though runtime type is different from static type (field type), encoder encodes `nil` value without encoding its runtime type.

encoding/ccf/decode_type.go
fxamacker: Yeah, I agree.  I previously looked into this and there were some complications.  We can tackle it in a follow-up PR as suggested.

encoding/ccf/cache.go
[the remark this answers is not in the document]
fxamacker: Moved `cachedSortedFieldIndex` from global into `Encoder` with commit cd55481beedde1b8da789e485fad97adc8dd1f88.  

This change causes extra memory allocation.  I can optimize this in later PR.

encoding/ccf/ccf_type_id.go
```diff
@@ -0,0 +1,77 @@
...
+// IMPORTANT: Don't use cadence.Type as map key because all Cadence composite/interface
+// types are pointers, and different instance of the same type will be treated as
+// different map key.
+type ccfTypeIDByCadenceType map[string]ccfTypeID
+
+func (types ccfTypeIDByCadenceType) id(t cadence.Type) (ccfTypeID, error) {
+	id, ok := types[t.ID()]
+	if !ok {
+		return 0, fmt.Errorf("CCF type ID not found for type %s", t.ID())
+	}
```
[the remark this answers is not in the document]
fxamacker: Done in commit 33505965e31b3eb97a7eef705db64b3b9e0b7ebb.

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1808 @@
...
+		// encodeFunction() receives a new ccfTypeIDByCadenceType to deduplicate
+		// composite type values within the same CCF function value encoding.
+		// For example, when a composite type appears more than once
+		// (recursive or repeated as nested type) within the same function value,
+		// it is only encoded once and is subsequently represented by its CCF ID.
+		// For function value encoding, CCF type ID is sequentially generated by
+		// traversal order of sorted parameters and return type.
+		return e.encodeFunction(x.FunctionType, ccfTypeIDByCadenceType{})
+
+	default:
```
SupunS, line 503: There was a new Cadence value (and a corresponding type) introduced recently called `Attachment`, which may need handling.
e.g:
 https://github.com/onflow/cadence/blob/2e801370afbd4cc24f1e56c613257b49e97608bf/values.go#L1650-L1653

SupunS: Corresponding type:
https://github.com/onflow/cadence/blob/2e801370afbd4cc24f1e56c613257b49e97608bf/types.go#L1221-L1227

encoding/ccf/simple_type_utils.go
```diff
@@ -0,0 +1,247 @@
...
+	TypeNever
+	TypeNumber
+	TypeSignedNumber
+	TypeInteger
+	TypeSignedInteger
+	TypeFixedPoint
+	TypeSignedFixedPoint
+	TypeBytes
+	TypeVoid
+	TypeFunction
```
SupunS, line 83: Similarly, along with `AttachmentType`, two new simple types were also introduced, namely `AnyStructAttachmentType` and `TheAnyResourceAttachmentType`. i.e:

https://github.com/onflow/cadence/blob/2e801370afbd4cc24f1e56c613257b49e97608bf/types.go#L92

and

https://github.com/onflow/cadence/blob/2e801370afbd4cc24f1e56c613257b49e97608bf/types.go#L132

SupunS (commented):
Noticed that we may need to add support for the newly added attachment type and value.

fxamacker:
> Noticed that we may need to add support for the newly added attachment type and value.

I agree. There is an issue #2379 to add support for recently added types (e.g. `AttachmentType`).  I plan to add support in a followup PR (this one is already very large).

@SupunS, is there a way for me to get automatically notified when external types & values gets added or modified?

SupunS:
oh okay, didn't know there was an issue already 👍 I Just ran onto this while fuzzing. Agree that it could fit into a separate PR.

> is there a way for me to get automatically notified when external types & values gets added or modified?

Unfortunately no. Adding new types / modifying happens very rarely, and would be even less frequent (ideally zero) once stable-cadence is released. Currently, we manually ensure all components are updated when the type is added/modified. e.g: One who add/modify a type would create an issue to update the encoder (or even include the changes, if feasible).
But on this occasion, I guess it's a temporary situation where the encoder and attachments feature were developed parallelly, and hence run into this problem where two components are not in sync.

encoding/ccf/cache.go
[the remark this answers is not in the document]
turbolent: Nice! We can potentially reuse the information across multiple executions as an internal optimization, it's just nice to have the library be stateless 👍

encoding/ccf/ccf_type_id.go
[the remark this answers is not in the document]
turbolent: That makes sense! Thanks for investigating the overhead 👌

encoding/ccf/ccf_type_id.go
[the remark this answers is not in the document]
turbolent: I see, that's a good point 👍

turbolent:
Great work on addressing all the feedback @fxamacker (and sorry it is so much – simply due to the nature of being a large PR)! I'll try to do another full review pass soon

encoding/ccf/sort.go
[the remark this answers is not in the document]
turbolent: When I asked this question initially, I was sort of getting the how, but was mostly wondering about the why. It seems like the answer is simply that we do not want to have the input being mutated, which makes sense 👍

encoding/ccf/encode_typedef.go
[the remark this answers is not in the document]
turbolent: Good point! 👌

encoding/ccf/encode.go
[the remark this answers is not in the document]
fxamacker: Done in commit 24dcca667b4c930aad03fade3087ab9a05fa0a02.

Since all `cadence.Value`'s runtime type is checked to be non-nil in `encodeValue()`, `v.DictionaryType` is not nil in `encodeDictionary()`.  So I also updated code to use `v.DictionaryType` directly.

encoding/ccf/encode.go
fxamacker: Added comment in commit 49d23b175ccf510f635c9e8fe8cdc68eefbb649a.

fxamacker:
> Great work on addressing all the feedback @fxamacker (and sorry it is so much – simply due to the nature of being a large PR)! I'll try to do another full review pass soon

@turbolent No worries!  Actually very relieved to see reviews are not cursory.  Thanks for taking time to review this!  :+1:  

I'll continue to add more tests.

types.go
```diff
@@ -1727,9 +1727,29 @@ func NewMeteredFunctionType(
...
+	case 1:
+		return "((" + t.Parameters[0].Type.ID() + "):" + returnTypeID + ")"
+	default:
+		params := make([]string, len(t.Parameters))
+		for i, param := range t.Parameters {
+			params[i] = param.Type.ID()
+		}
+		return "((" + strings.Join(params, ", ") + "):" + returnTypeID + ")"
+	}
+}
```
turbolent, line 1751: I think it is better to leave this out. We should rather consider removing `FunctionType.typeID` (in a separate issue/PR)

fxamacker: @turbolent 

CCF codec doesn't encode type ID for `FunctionType` anymore, so the decoded Cadence `FunctionType` wouldn't have type ID if it can't be generated.

If we leave `id()` out, would that cause any problems when we decode CCF encoded data into Cadence values/types, and then encode those values/types in JSON-CDC?

types.go
```diff
@@ -1843,9 +1844,29 @@ func NewMeteredRestrictedType(
...
+	case 1:
+		return typeID + "{" + t.Restrictions[0].ID() + "}"
+	default:
+		restrictions := make([]string, len(t.Restrictions))
+		for i, restriction := range t.Restrictions {
+			restrictions[i] = restriction.ID()
+		}
+		return typeID + "{" + strings.Join(restrictions, ", ") + "}"
+	}
+}
```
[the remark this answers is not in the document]
fxamacker: Done in commit d78f2f79d8a8376506c1503f6506b5b86ee085b4.  See previous comment for details.

encoding/ccf/decode_typedef.go
```diff
@@ -0,0 +1,477 @@
...
+		return ccfTypeID(0), cadenceTypeID(""), err
+	}
+
+	// element 2: fields
+	rawField, err := d.dec.DecodeRawBytes()
+	if err != nil {
+		return ccfTypeID(0), cadenceTypeID(""), err
+	}
+
+	_ = types.add(ccfID, constructor(location, identifier))
```
turbolent, line 342:
```suggestion
	// The return value can be ignored, because its non-existence was already checked above
	_ = types.add(ccfID, constructor(location, identifier))
```

encoding/ccf/decode_typedef.go
```diff
@@ -0,0 +1,477 @@
...
+		return ccfTypeID(0), cadenceTypeID(""), fmt.Errorf("found duplicate CCF type ID %d in interface-type", ccfID)
+	}
+
+	// element 1: cadence-type-id
+	cadenceID, location, identifier, err := d.decodeCadenceTypeID()
+	if err != nil {
+		return ccfTypeID(0), cadenceTypeID(""), err
+	}
+
+	_ = types.add(ccfID, constructor(location, identifier))
```
turbolent, line 475:
```suggestion
	// The return value can be ignored, because its non-existence was already checked above
	_ = types.add(ccfID, constructor(location, identifier))
```

encoding/ccf/decode_type.go
```diff
@@ -0,0 +1,660 @@
...
+	//   "type-ref.id MUST refer to composite-type.id."
+	//   "type-value-ref.id MUST refer to composite-type-value.id in the same composite-type-value data item."
+	t, err := types.typ(id)
+	if err != nil {
+		return nil, err
+	}
+
+	// Track referenced type definition so the decoder can detect
+	// encoded but not referenced type definition (extraneous data).
+	types.reference(id)
```
turbolent, line 657: 👌

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1900 @@
...
+
+	// element 1: value
+	return e.encodeValue(value, runtimeType, tids)
+}
+
+// encodeTypeDefs encodes composite/interface type definitions as
+// language=CDDL
+// composite-typedef = [
+//
+//	+(
```
turbolent, line 232:
```suggestion
//  ; one-or-more instead of zero-or-more because:
//  ; - when encoding a primitive type, such as boolean or string, `ccf-type-and-value-message` is used (no `composite-typedef` at all)
//  ; - when encoding a composite type, such as event, `ccf-typedef-and-value-message` is used, which encodes at least one `composite-typedef`
//	+ (
```

encoding/ccf/traverse_value.go
```diff
@@ -0,0 +1,232 @@
...
+
+	if !checkRuntimeType {
+		// Return without traversing value for runtime types.
+		return
+	}
+
+	// Traverse v's elements for runtime types.
+	// Note: don't need to traverse fields of cadence.Enum
+	// because enum's field is an integer subtype.
+	switch x := v.(type) {
```
turbolent: Can shadow the variable name here:

```suggestion
	switch v := v.(type) {
```

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1900 @@
...
+func (e *Encoder) encodeTypeDefs(types []cadence.Type, tids ccfTypeIDByCadenceType) error {
+	// Encode array head with number of type definitions.
+	err := e.enc.EncodeArrayHead(uint64(len(types)))
+	if err != nil {
+		return err
+	}
+
+	for _, typ := range types {
+
+		switch x := typ.(type) {
```
turbolent, line 251: Can shadow the variable name here:

```suggestion
		switch typ := typ.(type) {
```

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1900 @@
...
+		// element 0: type as inline-type
+		err = e.encodeInlineType(inlineType, tids)
+		if err != nil {
+			return err
+		}
+
+		// element 1: value
+	}
+
+	switch x := v.(type) {
```
turbolent, line 374: Can shadow the variable name here:
```suggestion
	switch v := v.(type) {
```

encoding/ccf/encode.go
```diff
@@ -0,0 +1,1900 @@
...
+
+// encodeOptionalTypeValue encodes cadence.OptionalType as
+// language=CDDL
+// optional-type-value =
+//
+//	; cbor-tag-optional-type-value
+//	#6.186(type-value)
+func (e *Encoder) encodeOptionalTypeValue(typ *cadence.OptionalType, visited ccfTypeIDByCadenceType) error {
+	rawTagNum := []byte{0xd8, CBORTagOptionalTypeValue}
+	return e.encodeOptionalTypeWithRawTag(
```
turbolent, line 1183: Nice work abstracting the encoding and decoding of the two different type hierarchies 👏

encoding/ccf/decode.go
```diff
@@ -0,0 +1,2039 @@
...
+// CBORDecMode
+//
+// See https://github.com/fxamacker/cbor:
+// "For best performance, reuse EncMode and DecMode after creating them."
+var CBORDecMode = func() cbor.DecMode {
+	decMode, err := cbor.DecOptions{
+		IntDec:           cbor.IntDecConvertNone,
+		MaxArrayElements: math.MaxInt64,
+		MaxMapPairs:      math.MaxInt64,
+		MaxNestedLevels:  math.MaxInt16,
```
turbolent, line 44: I wonder if we should document this somehow. Theoretically the spec does not define such limits. Though we also want to limit the amount of decoded data to some sensible limit, so this is great.

Maybe we should add these limits also to the spec; or alternatively just document this for the codec, e.g. add a comment noting it on e.g. `Decoder`

encoding/ccf/decode.go
```diff
@@ -0,0 +1,2039 @@
...
+	case CBORTagTypeDefAndValue:
+		// Decode ccf-typedef-and-value-message.
+		return d.decodeTypeDefAndValue()
+
+	case CBORTagTypeAndValue:
+		// Decode ccf-type-and-value-message.
+		return d.decodeTypeAndValue(newCadenceTypeByCCFTypeID())
+
+	default:
+		return nil, fmt.Errorf(
```
turbolent, line 131: It would be good to add tess for the uncovered error cases here and below (could be in a follow-up PR)

encoding/ccf/decode.go
```diff
@@ -0,0 +1,2039 @@
...
+	path, err := d.decodePath()
+	if err != nil {
+		return nil, err
+	}
+
+	return cadence.NewMeteredStorageCapability(
+		d.gauge,
+		path.(cadence.Path),
+		address.(cadence.Address),
+		typ.BorrowType), nil
```
turbolent, line 1195:
```suggestion
		typ.BorrowType,
	), nil
```

encoding/ccf/decode.go
```diff
@@ -0,0 +1,2039 @@
...
+//	/ enum-type-value
+//	/ struct-interface-type-value
+//	/ resource-interface-type-value
+//	/ contract-interface-type-value
+//	/ function-type-value
+//	/ reference-type-value
+//	/ restricted-type-value
+//	/ capability-type-value
+//	/ type-value-ref
+func (d *Decoder) _decodeTypeValue(visited *cadenceTypeByCCFTypeID) (cadence.Type, error) {
```
turbolent, line 1229: Why does this function and `_decodeNullableTypeValue` have an underscore prefix?

turbolent (commented):
Fantastic work! 👏 

I reviewed everything except the test cases again and it's looking great!

Except for some minor suggestions, the only work remaining I can see is adding more tests for the unhappy paths of the decoder, but maybe it's better to get this PR in and add those in a follow-up PR, because this one is already very large.

One part I'm still trying to wrap my head around is the special handling of optional and reference types (`needToEncodeRuntimeType`, `getTypeToEncodeAsCCFInlineType`, `getOptionalInnerTypeToEncodeAsCCFInlineType`). Maybe we can have a follow-up sync around this next week.

encoding/ccf/consts.go
```diff
@@ -0,0 +1,159 @@
...
+// CCF uses CBOR tag numbers 128-255, which are unassigned by [IANA]
+// (https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml).
+//
+// !!! *WARNING* !!!
+//
+// CCF Codec *MUST* comply with CCF Specifications.  Relevant changes
+// must be in sync between codec and specifications.
+//
+// Only add new tag number by:
+// - replacing existing placeholders (`_`) with new tag number
```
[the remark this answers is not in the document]
fxamacker: By not appending to the end, we can have larger block of unused tag numbers if we ever need it.

Tag numbers in 128-255 are unassigned in CBOR, and we currently use 128-231.  Since each group of tags in this range have reserved space available, there is no need to append new tag numbers in 232-255.

encoding/ccf/ccf_type_id.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (id ccfTypeID) Bytes() []byte {
+	return new(big.Int).SetUint64(uint64(id)).Bytes()
+}
+
+func (id ccfTypeID) Equal(other ccfTypeID) bool {
+	return id == other
+}
+
+// IMPORTANT: Don't use cadence.Type as map key because all Cadence composite/interface
```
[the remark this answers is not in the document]
fxamacker: Done in commit 19f829fd4ca3401237844db355ed247b6321d166.

encoding/ccf/decode_typedef.go
```diff
@@ -0,0 +1,477 @@
...
+
+	"github.com/onflow/cadence"
+	"github.com/onflow/cadence/runtime/common"
+)
+
+// decodeTypeDefs decodes composite/interface type definitions as
+// language=CDDL
+// composite-typedef = [
+//
+//	+(
```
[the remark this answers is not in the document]
fxamacker: Done in commit 19f829fd4ca3401237844db355ed247b6321d166.

encoding/ccf/decode_typedef.go
fxamacker: Done in commit 19f829fd4ca3401237844db355ed247b6321d166.

encoding/ccf/decode_typedef.go
fxamacker: Done in commit 19f829fd4ca3401237844db355ed247b6321d166.

encoding/ccf/ccf_type_id_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/onflow/cadence"
+	"github.com/onflow/cadence/runtime/tests/utils"
+)
+
+func TestCCFTypeID(t *testing.T) {
+	testCases := []struct {
```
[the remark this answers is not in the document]
fxamacker: Done in commit 804555889ba00a8b6399280e57fddd677f30a4f5.

values.go
```diff
@@ -1267,6 +1267,11 @@ func NewMeteredFix64(gauge common.MemoryGauge, constructor func() (string, error
 	return NewFix64(value)
 }
 
+func NewMeteredFix64FromInt64(gauge common.MemoryGauge, n int64) (Fix64, error) {
```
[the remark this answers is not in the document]
fxamacker: Done in commit dbbbbd9927652d2c628a05868fee0f7db865d205.

encoding/ccf/decode.go
fxamacker: The helper function (with underscore prefix) was called by function with non-prefixed name.

I simplified this in commit dbbbbd9927652d2c628a05868fee0f7db865d205 by inlining.

encoding/ccf/traverse_value.go
fxamacker: Done in commit 926661cbc672ec8711015c5a097586a81e2f3d2b.

While at it, also applied this suggestion to other places.

encoding/ccf/encode.go
fxamacker: Done in commit 926661cbc672ec8711015c5a097586a81e2f3d2b.

While at it, also applied this suggestion to other places.

encoding/ccf/encode.go
fxamacker: Done in commit 926661cbc672ec8711015c5a097586a81e2f3d2b.

While at it, also applied this suggestion to other places.

encoding/ccf/decode.go
fxamacker: Done in commit 926661cbc672ec8711015c5a097586a81e2f3d2b.

encoding/ccf/decode_typedef.go
```diff
@@ -0,0 +1,477 @@
...
+				string(cadenceID),
+			)
+		}
+
+		cadenceTypeIDs[cadenceID] = struct{}{}
+		previousCadenceID = cadenceID
+	}
+
+	// Decode fields after all high-level type definitions are resolved.
+	for id, raw := range rawFields { //nolint:maprange
```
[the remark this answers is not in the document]
fxamacker: Good point! Done in commit f6f08eed16c991f320a671801762f8f79ba30b29.

- replaced map with slice to avoid future problems when metering is added to the function.
- renamed `rawFields` to `rawFieldsOfTypes`
- renamed `raw` to `rawFields`

encoding/ccf/decode_typedef.go
```diff
@@ -0,0 +1,477 @@
...
+//	  )]
+func (d *Decoder) decodeTypeDefs() (*cadenceTypeByCCFTypeID, error) {
+	// Decode number of type definitions.
+	count, err := d.dec.DecodeArrayHead()
+	if err != nil {
+		return nil, err
+	}
+
+	if count == 0 {
+		return nil, errors.New("found 0 type definition in composite-typedef (expected at least 1 type definition)")
```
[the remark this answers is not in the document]
fxamacker: Yes, more tests should be added in follow-up PRs to cover unhappy paths.

encoding/ccf/decode.go
fxamacker: Yes, more tests can be added in subsequent PRs for unhappy paths.

encoding/ccf/decode.go
fxamacker: > Theoretically the spec does not define such limits. Though we also want to limit the amount of decoded data to some sensible limit, so this is great.

Yeah, that is exactly my thoughts too.  The CCF specs doesn't set these limits and it currently refers to CBOR RFC 8949 for security considerations (which mentions but doesn't specify CBOR decoder limits).

- 565f84e6576a6148f6b76d77cd2e5f68b752bd57 - Reduced max element limits for arrays and maps.
- 12b7116000c8f9f6900f009d775c6dfa636e8f6e - Reject indefinite length CBOR data items in CCF decoder (spotted this while rereading section 10 of RFC 8949).

There's an [issue](https://github.com/fxamacker/ccf_draft/issues/64) opened last month for CCF Specs to update it's section on Security Considerations, so I'll go ahead and do that.

encoding/ccf/decode.go
```diff
@@ -345,46 +358,56 @@ func (d *Decoder) decodeValue(t cadence.Type, types *cadenceTypeByCCFTypeID) (ca
...
+	case cadence.PublicPathType:
+		return d.decodePath()
 
-	case cadence.PathType:
+	case cadence.PrivatePathType:
 		return d.decodePath()
 
 	case cadence.MetaType:
-		return d.decodeNullableTypeValue()
+		typeValue, err := d.decodeNullableTypeValue(newCadenceTypeByCCFTypeID())
```
turbolent, line 391: Can we pass `types` instead of constructing a new `cadenceTypeByCCFTypeID`?

turbolent (approved):
Recent changes look great! 👌

SupunS (approved):
🎉

encoding/ccf/decode.go
fxamacker: Unfortunately, no.  

`cadenceTypeByCCFTypeID` uses a map with CCF type ID as keys.  CCF type ID can collide if we reuse type definition's `types` for type values because:
- CCF type IDs are zero-based for composite `TypeValue`.
- CCF type IDs are zero-based for composite type definitions.

turbolent (approved):
Fantastic work @fxamacker! 👏👏

encoding/ccf/decode.go
turbolent: Ah, right. I had the same question for encoding I think. 

Could you maybe add this explanation as a comment?

encoding/ccf/consts.go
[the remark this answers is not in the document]
turbolent: I see, that makes sense. Could you please add that explanation as a comment to the code?

fxamacker:
@turbolent @SupunS Thanks for reviews, fuzzing, and meetings!  After merging this, I'll open followup PR to add API to allow some default CBOR limits to be overridden (e.g. `NewDecoderWithOptions()`).

I spotted an edge case where it was possible for encoder to needlessly encode redundant data, related to reference type.  

Can you take a look at commit 51552f153326be5a493682fcd68ae381d9899013 as a sanity check before I merge this PR?  :pray:

encoding/ccf/decode.go
fxamacker: Done in commit 4a6aa577f9b69074f1dfc0e20743031e73e1486a.

encoding/ccf/consts.go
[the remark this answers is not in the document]
fxamacker: Done in commit 4a6aa577f9b69074f1dfc0e20743031e73e1486a.

Issue #2192, opened 2022-12-09
Child Accounts


j1010001:
Epic to track work on child accounts.

turbolent:
Currently discussing https://forum.onflow.org/t/super-user-account/4088

PR #2366, opened 2023-03-02
Enforce account links to be private

turbolent:
Work towards #2192 

## Description

The [Account linking FLIP](https://github.com/onflow/flips/pull/53) received some great feedback, suggesting that account links should not be public.

Enforce this by changing the type of the path parameter of the `linkAccount` function from `CapabilityPath` to `PrivatePath`.

______


dsainati1 (approved):

runtime/account_test.go
```diff
@@ -2595,111 +2595,6 @@ func TestRuntimeAccountLink(t *testing.T) {
 		assert.ErrorContains(t, err, "value of type `AuthAccount` has no member `linkAccount`")
 	})
 
-	t.Run("enabled, pragma", func(t *testing.T) {
```
SupunS, line 2598: Should we keep this test, and only change the linking path from `public` to `private`?

turbolent: I started on it, then realized that the access part, which uses `getCapability`, would also have to be adjusted, i.e. the after the linking the capability must be stored somewhere, then retrieved in the access part.

After the refactor I realized that the "publish and claim" test case below does exactly that, so there was no need to keep this one here.

SupunS (approved):

PR #2368, opened 2023-03-02
Emit an event when an account is linked

turbolent:
Work towards #2192

## Description

When an account is linked, emit the new built-in event `flow.AccountLinked(address: Address, path: PrivatePath)`

______

### DIFF
--- a/runtime/stdlib/flow.go
+++ b/runtime/stdlib/flow.go
@@ -272,13 +272,11 @@ var AccountInboxClaimedEventType = newFlowEventType(
 	AccountEventNameParameter,
 )
 
-var AccountEventPathParameter = sema.Parameter{
-	Identifier:     "path",
-	TypeAnnotation: sema.NewTypeAnnotation(sema.PrivatePathType),
-}
-
 var AccountLinkedEventType = newFlowEventType(
 	"AccountLinked",
 	AccountEventAddressParameter,
-	AccountEventPathParameter,
+	sema.Parameter{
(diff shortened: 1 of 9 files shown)


dsainati1 (approved):

turbolent:
@SupunS Could you please have a look?

runtime/stdlib/flow.go
```diff
@@ -271,3 +271,12 @@ var AccountInboxClaimedEventType = newFlowEventType(
 	AccountEventRecipientParameter,
 	AccountEventNameParameter,
 )
+
+var AccountLinkedEventType = newFlowEventType(
+	"AccountLinked",
+	AccountEventAddressParameter,
+	sema.Parameter{
+		Identifier:     "path",
+		TypeAnnotation: sema.NewTypeAnnotation(sema.PrivatePathType),
```
SupunS, line 280: Is there a way to ensure the type used here is the same as the param type in the [actual account link function](https://github.com/onflow/cadence/blob/1775ebd3bfa90a29653c16d99d4a0253ee86f3b7/runtime/sema/authaccount_type.go#L83)? Just to be sure that we won't update one and forget to update the other.

SupunS (approved):

runtime/stdlib/flow.go
turbolent: That's a great idea! Refactored to a common value in 7c092a8a04e132a12b6fb299f31446120bb3c98a

Issue #2361, opened 2023-03-01
Code coverage report for an address that contains only a contract interface


m-Peter:
### Current Behavior

Generating code coverage report for an address/location that contains only a contract interface, or generally Cadence source code that is not meant to be interpreted/executed, returns a `percentage` of `"NaN%"`. E.g:

![coverage-for-resource-interface](https://user-images.githubusercontent.com/1778965/222155196-7298190f-3844-4d02-9c9a-f41ea3f11f49.png)

### Expected Behavior

How do we handle such a case? Should we show a percentage of `100%` ?

### Steps To Reproduce

This happens when adding code coverage in `flow-emulator`, and expose it in the REST API.

### Environment

```markdown
- Cadence version: v0.34.0
- Network: Emulator
```

bluesign:
@m-Peter 

Those are pre/post conditions ( `ast.Condition` ), I think you should add that to program inspector. 

Conditions are taking `expression` instead of statement.

m-Peter:
@bluesign Great catch :100: :100: :100: 

I checked every line hit, and indeed 7/8 are in fact pre/post conditions. However, the line `0` which, has `14` hits, remains a mystery.

bluesign:
@m-Peter that one is a bug I guess, I ran into that when improving debugger too.

m-Peter:
I see, I wasn't able to identify where it comes from. However, I did find a way to include pre/post conditions for all functions, whether they are defined in contract interfaces or in concrete contracts.

![Screenshot from 2023-03-04 14-39-58](https://user-images.githubusercontent.com/1778965/222903516-b773ff0f-04bb-4fef-9403-c0ac0e5160ce.png)

bluesign:
you can ignore probably line 0 anyway. normally line number starts with 1 I guess

m-Peter:
That makes total sense :+1:

turbolent:
@m-Peter Do you have an example program where you see reports of line 0?

bluesign:
> @m-Peter Do you have an example program where you see reports of line 0?

beforeStatements iirc

m-Peter:
I have a branch on `flow-cli`, where I am adding support for code coverage there as well. This peculiar case manifests only for the `FungibleToken` contract interface, and it is visible only by checking the location key format.

m-Peter:
```go
if location.String() == "ee82856bf20e2aa6.FungibleToken" {
	if statement.StartPosition().Line == 0 {
[...]

PR #2371, opened 2023-03-04
Code coverage support for pre/post conditions inside functions

m-Peter:
Closes https://github.com/onflow/cadence/issues/2361

## Description

When inspecting a program, now we also count `pre/post` conditions inside functions. We do this by checking for:

```go
functionBlock, isFunctionBlock := element.(*ast.FunctionBlock)
```
and adding the corresponding lines of expressions contained in the conditions.
______

### DIFF
--- a/runtime/coverage.go
+++ b/runtime/coverage.go
@@ -155,6 +155,10 @@ func (r *CoverageReport) InspectProgram(location Location, program *ast.Program)
 	}
 	r.Programs[location] = program
 	lineHits := make(map[int]int, 0)
+	recordLine := func(hasPosition ast.HasPosition) {
+		line := hasPosition.StartPosition().Line
+		lineHits[line] = 0
+	}
 	var depth int
 
 	inspector := ast.NewInspector(program)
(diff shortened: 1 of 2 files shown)


runtime/coverage.go
```diff
@@ -173,6 +177,22 @@ func (r *CoverageReport) InspectProgram(location Location, program *ast.Program)
...
+					if functionBlock.PreConditions != nil {
+						conditions = append(conditions, *functionBlock.PreConditions...)
+					}
+					if functionBlock.PostConditions != nil {
+						conditions = append(conditions, *functionBlock.PostConditions...)
+					}
+					for _, condition := range conditions {
+						line := condition.Test.StartPosition().Line
+						lineHits[line] = 0
+					}
```
turbolent, line 194: Maybe avoid the allocation of a temporary slice here, and avoid the code duplication of getting the start position line and recording it in the line hit map. 

You can do this e.g. by defining a helper function, after `lineHits` is introduced (see above, line 154):

```go
  	recordLine := func(hasPosition ast.HasPosition) {
		line := hasPosition.StartPosition().Line
		lineHits[line] = 0
	} 
```

Replace the recording for the statement above (line 173):

```go
					recordLine(element)
```

Then use it for the conditions:

```suggestion
					if functionBlock.PreConditions != nil {
						for _, condition := range *functionBlock.PreConditions {
							recordLine(condition.Test)
						}
					}

					if functionBlock.PostConditions != nil {
						for _, condition := range *functionBlock.PostConditions {
							recordLine(condition.Test)
						}
					}
```

turbolent (commented):
Good catch!

runtime/coverage.go
m-Peter: Great suggestion :clap:
Updated in https://github.com/onflow/cadence/pull/2371/commits/bb43f13aba26f9916436ea96ae31c9dc30ab7327

turbolent (approved):
Nice!

Issue #2360, opened 2023-03-01
Allow multiple contracts per flow address in code coverage report


m-Peter:
### Issue to be solved

When generating a coverage report, the current location key format (`A.0ae53cb6e3f42a79`) hides all contracts under one Flow address. E.g:

![current-location-format](https://user-images.githubusercontent.com/1778965/222152740-a9d6f013-cda2-4eb8-b34c-90b6239b423d.png)

### Suggested Solution

Include the flow address and the contract name (`0ae53cb6e3f42a79.FlowToken`), in the location key format. E.g:

![desired-location-format](https://user-images.githubusercontent.com/1778965/222153219-1904fad1-70d5-4469-ad55-b467d3df6c70.png)

PR #2372, opened 2023-03-06
Improve contract address for code coverage report

m-Peter:
Closes https://github.com/onflow/cadence/issues/2360

## Description

The location key format looks now like this: `A.f8d6e0586b0a20c7.FlowEpoch` . This will allow us to differentiate between contracts that are deployed in the same Flow address. This is useful when using the code coverage feature outside a testing context, such as flow emulator. For example:

Screenshot from 2023-03-06 10-12-46
______

### DIFF
--- a/runtime/common/addresslocation.go
+++ b/runtime/common/addresslocation.go
@@ -85,6 +85,10 @@ func (l AddressLocation) Description() string {
 	)
 }
 
+func (l AddressLocation) ID() string {
+	return fmt.Sprintf("%s.%s", AddressLocationPrefix, l)
+}
+
 func (l AddressLocation) MarshalJSON() ([]byte, error) {
 	return json.Marshal(&struct {
 		Type    string
--- a/runtime/common/addresslocation_test.go
+++ b/runtime/common/addresslocation_test.go
@@ -65,6 +65,19 @@ func TestAddressLocationTypeID(t *testing.T) {
 	)
 }
 
+func TestAddressLocationID(t *testing.T) {
+
+	t.Parallel()
+
+	location, _, err := decodeAddressLocationTypeID(nil, "A.0000000000000001.Bar.Baz")
+	require.NoError(t, err)
+
+	assert.Equal(t,
+		"A.0000000000000001.Bar",
+		location.ID(),
+	)
+}
+
 func TestDecodeAddressLocationTypeID(t *testing.T) {
 
 	t.Parallel()
--- a/runtime/coverage.go
+++ b/runtime/coverage.go
@@ -256,9 +256,7 @@ func (r *CoverageReport) MarshalJSON() ([]byte, error) {
 
 	coverage := make(map[string]LC, len(r.Coverage))
 	for location, locationCoverage := range r.Coverage { // nolint:maprange
-		typeID := location.TypeID(nil, "")
-		locationID := typeID[:len(typeID)-1]
-		coverage[string(locationID)] = LC{
+		coverage[location.ID()] = LC{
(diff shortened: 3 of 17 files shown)


runtime/coverage.go
```diff
@@ -235,8 +235,8 @@ func (r *CoverageReport) MarshalJSON() ([]byte, error) {
 	coverage := make(map[string]LC, len(r.Coverage))
 	for location, locationCoverage := range r.Coverage { // nolint:maprange
 		typeID := location.TypeID(nil, "")
-		locationID := typeID[:len(typeID)-1]
-		coverage[string(locationID)] = LC{
+		locationID := fmt.Sprintf("%v%v", typeID[:2], location.String())
```
turbolent, line 238: It took me a while to understand what was needed here and how it was generated.

Basically, we want a string representation of the location, including the location prefix, so it can be differentiated. However, `Location` `String()` implementations do not include the prefix.

Instead of preparing this location ID here, it might be better to just introduce a new method for `Location`, e.g. `ID()`. This also addresses the issue that a location prefix might be longer than one character.

Also, could you please add some tests? Given that the format changes here, I was expecting some tests to have to be adjusted as well.

turbolent (commented):
👍

runtime/coverage.go
m-Peter: > Also, could you please add some tests? Given that the format changes here, I was expecting some tests to have to be adjusted as well.

Will do. I was also surprised to see the tests passing without any change. I guess because on emulator we are dealing with `AddressLocation`, and on tests with `StringLocation` & `ScriptLocation`. I will check it though :+1:

m-Peter: I have updated all the `common.Location` implementations, to include an `ID` method. I have also added tests both in the coverage package, and in each specific implementation of the `common.Location` interface :pray:

runtime/common/addresslocation.go
```diff
@@ -85,6 +85,10 @@ func (l AddressLocation) Description() string {
 	)
 }
 
+func (l AddressLocation) ID() string {
+	return fmt.Sprintf("%v.%v", AddressLocationPrefix, l.String())
```
turbolent, line 89: Can use string formatting here and for the other cases:

```suggestion
	return fmt.Sprintf("%s.%s", AddressLocationPrefix, l)
```

runtime/common/addresslocation_test.go
```diff
@@ -65,6 +65,18 @@ func TestAddressLocationTypeID(t *testing.T) {
 	)
 }
 
+func TestAddressLocationID(t *testing.T) {
+
+	t.Parallel()
+
+	location, _, _ := decodeAddressLocationTypeID(nil, "A.0000000000000001.Bar.Baz")
```
turbolent, line 72: Maybe also assert there is no error here and below:

```suggestion
	location, _, err := decodeAddressLocationTypeID(nil, "A.0000000000000001.Bar.Baz")
	require.NoError(t, err)
```

turbolent (commented):
Looks good! Just a couple minor suggestions

runtime/coverage.go
turbolent: Nice!

runtime/common/addresslocation_test.go
m-Peter: Good idea!
Fixed in https://github.com/onflow/cadence/pull/2372/commits/4643d37ca193b6a6de72e576aa7505fe401c02b1

turbolent (approved):
Nice! Thank you

PR #2380, opened 2023-03-10
Change account linking pragma to run-time configuration of execution

turbolent:
### Problem

#2353 introduced a pragma, which determines if the account linking function (`AuthAccount.linkAccount`) is available. 

This is a static check: If the pragma is declared in the same program, the function is available, if the pragma is not declared, the function is not available.

This works well to allow wallets to detect transactions which attempt to use the account linking function, and potentially reject them. However, if the function is used in a contract, the current implementation requires the *contract* to declare the pragma, and a transaction which calls into the transaction does not need to declare the pragma. This prevents wallets from easily detecting a potential use of the account linking function.

### Solution

Change the behaviour of the account linking pragma `#allowAccountLinking` from just a static check, to both a static check (as before, ensure the pragma is declared in the header of a program), and a dynamic check when the account linking function is called. The dynamic check requires that the pragma was declared.

### Changes

- Still statically check the pragma occurs at the top level of a program, and not nested or after other non-pragma declarations
- Change the availability of the `AuthAccount.linkAccount` function, from being only defined when the pragma is defined in the same program, to being always available, independent of the pragma
- Add configuration option to the interpreter which determines if the account linking function may be called
- Add a function to the interpreter which configures the interpreter's configuration option based on the pragma declaration
- Perform the pragma-based configuration of the account linking availability for transactions and scripts

______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -4755,19 +4755,15 @@ func (interpreter *Interpreter) ConfigureAccountLinkingAllowed() {
 
 	config.AccountLinkingAllowed = false
 
-	var rejectAllowAccountLinkingPragma bool
-
-	for _, declaration := range interpreter.Program.Program.Declarations() {
-
-		if pragmaDeclaration, isPragma := declaration.(*ast.PragmaDeclaration); isPragma {
-			if sema.IsAllowAccountLinkingPragma(pragmaDeclaration) {
-				if !rejectAllowAccountLinkingPragma {
-					config.AccountLinkingAllowed = true
-				}
-				continue
-			}
-		}
-
-		rejectAllowAccountLinkingPragma = true
+	declarations := interpreter.Program.Program.Declarations()
+	if len(declarations) < 1 {
+		return
 	}
+
+	pragmaDeclaration, isPragma := declarations[0].(*ast.PragmaDeclaration)
+	if !isPragma || !sema.IsAllowAccountLinkingPragma(pragmaDeclaration) {
+		return
+	}
+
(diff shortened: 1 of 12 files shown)


dsainati1 (approved):

runtime/interpreter/interpreter.go
```diff
@@ -4733,3 +4739,35 @@ func (interpreter *Interpreter) DecodeTypeInfo(decoder *cbor.StreamDecoder) (atr
...
+	config.AccountLinkingAllowed = false
+
+	var rejectAllowAccountLinkingPragma bool
+
+	for _, declaration := range interpreter.Program.Program.Declarations() {
+
+		if pragmaDeclaration, isPragma := declaration.(*ast.PragmaDeclaration); isPragma {
+			if sema.IsAllowAccountLinkingPragma(pragmaDeclaration) {
+				if !rejectAllowAccountLinkingPragma {
+					config.AccountLinkingAllowed = true
```
SupunS, line 4765: It seems the loop can be terminated on the first occurrence of the pragma. That would also allow removing the `rejectAllowAccountLinkingPragma` variable.

turbolent: Good idea! I had pretty much just copied the logic from the checker, but he loop is unnecessary.

In [`c5a4e71` (#2380)](https://github.com/onflow/cadence/pull/2380/commits/c5a4e71957c99f4b3a7071cd044df078bcc62114) I made the function as strict as the checker, and simplified it, by just looking at the first pragma. In addition, I added some more checker and interpreter tests.

SupunS (approved):

PR #2388, opened 2023-03-15
Fix Value.Type() returning Type referencing nil ptr for eight external values

### LINKED ISSUE #2387: Value.Type() can return a reference to a nil pointer
### Issue to be solved

Current implementations of `Value.Type()` can return a Type referencing a nil pointer.

This can make callers deference the nil pointer.

### Suggested Solution

Implementations of `Value.Type()` should return `nil` instead of a reference to a nil pointer.

fxamacker:
Changes include:
- Changed type of `Dictionary.DictionaryType` from `Type` to `*DictionaryType`.
- Changed `Type()` to check and return nil `Type`, instead of `Type` referencing nil pointer for these external values:
  - `Dictionary`
  - `Struct`
  - `Resource`
  - `Attachment`
  - `Event`
  - `Contract`
  - `Enum`
  - `Function`

Closes #2387

______

### DIFF
--- a/runtime/format/capability.go
+++ b/runtime/format/capability.go
@@ -35,10 +35,3 @@ func StorageCapability(borrowType string, address string, path string) string {
 		path,
 	)
 }
-
-func AccountCapability(address string) string {
-	return fmt.Sprintf(
-		"Capability(address: %s)",
-		address,
-	)
-}
--- a/values.go
+++ b/values.go
@@ -1426,7 +1426,7 @@ func (v Array) String() string {
 // Dictionary
 
 type Dictionary struct {
-	DictionaryType Type
+	DictionaryType *DictionaryType
 	Pairs          []KeyValuePair
 }
 
@@ -1453,6 +1453,11 @@ func NewMeteredDictionary(
 func (Dictionary) isValue() {}
 
 func (v Dictionary) Type() Type {
+	if v.DictionaryType == nil {
+		// Return nil Type instead of Type referencing nil *DictionaryType,
+		// so caller can check if v's type is nil and also prevent nil pointer dereference.
+		return nil
+	}
 	return v.DictionaryType
 }
 
@@ -1541,6 +1546,11 @@ func NewMeteredStruct(
 func (Struct) isValue() {}
 
 func (v Struct) Type() Type {
+	if v.StructType == nil {
+		// Return nil Type instead of Type referencing nil *StructType,
+		// so caller can check if v's type is nil and also prevent nil pointer dereference.
+		return nil
+	}
 	return v.StructType
 }
 
@@ -1619,6 +1629,11 @@ func NewMeteredResource(
 func (Resource) isValue() {}
 
 func (v Resource) Type() Type {
+	if v.ResourceType == nil {
+		// Return nil Type instead of Type referencing nil *ResourceType,
+		// so caller can check if v's type is nil and also prevent nil pointer dereference.
+		return nil
+	}
 	return v.ResourceType
 }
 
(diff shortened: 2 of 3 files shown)


values.go
```diff
@@ -1453,6 +1453,11 @@ func NewMeteredDictionary(
 func (Dictionary) isValue() {}
 
 func (v Dictionary) Type() Type {
+	if v.DictionaryType == nil {
```
turbolent, line 1456: Took me a moment to realize why this is needed: Go creates a typed nil without it. (Typed nil strikes again!) Too bad this is necessary, but seems right, given that allows us to keep the `Type() Type` function.

turbolent (commented):
Thanks for fixing this!

Unfortunately, introducing typed nils is too easy. 

We should maybe add tests that avoid regressions for these functions, and in addition, prevent incorrect implementations in the future (e.g. by creating an empty instance of each `Type` implementation, and then, as each nested type is automatically `nil`, the a call to `Type()` should be equal to the untyped nil (`== nil`).

Let me try adding that.

values_test.go
```diff
@@ -635,6 +635,8 @@ func TestGetType(t *testing.T) {
 			reflect.TypeOf(Event{}):      {},
 			reflect.TypeOf(Resource{}):   {},
 			reflect.TypeOf(Struct{}):     {},
+			reflect.TypeOf(Enum{}):       {},
+			reflect.TypeOf(Attachment{}): {},
```
turbolent, line 639: So while extending these tests I noticed that the existing code relying on the `reflect.typelinks` function doesn't include these two types for whatever reason. I saw the code was taken from https://stackoverflow.com/questions/32132064/how-to-discover-all-package-types-at-runtime.

What is so special about `Enum` that it isn't included, but e.g. `Struct` is? 🤔

SupunS: Maybe because this file doesn't instantiate any `Enum` instances? (should have been a test case there for enums!)

SupunS: oh well, that looks like a problem with this test. It only can find types that are actually used inside this file 🥴

turbolent: So I tried to look a bit more into it and finally found and confirmed the problem:

`typenames` says:

```
// Only types we might want to look up are included:
// pointers, channels, maps, slices, and arrays.
```

If I add a simple statement like this, which converts to `any`, i.e. creates a pointer, `Enum` gets included:

```golang
fmt.Printf("%v\n", Enum{})
```

I can't find any alternative, so for now I'll remove the code relying on internals, and just manually list the `Value` implementations. This won't let us automatically handle new cases, but the current code may not even handle existing cases.

runtime/format/capability.go
```diff
@@ -35,10 +35,3 @@ func StorageCapability(borrowType string, address string, path string) string {
 		path,
 	)
 }
-
-func AccountCapability(address string) string {
```
turbolent, line 39: Dead code

turbolent:
Refactored the existing tests and unified them with similar tests for `String()`, as well as added further missing test cases (e.g. `Function` and `Character`) in fcb25195cb486033ea78794f3e0fc7354b25cee3

SupunS (approved):

values.go
```diff
@@ -2102,6 +2102,11 @@ func (v Function) Type() Type {
 	return v.FunctionType
 }
 
+func (v Function) WithType(functionType *FunctionType) Function {
+	v.FunctionType = functionType
+	return v
+}
+
```
fxamacker, line 2109: @turbolent  Thanks for adding new tests and refactoring existing tests!  Great work!  :+1: 

Only item I spotted is that I think adding `Function.WithType()` isn't necessary because `Function` is created with `FunctionType`.

```
func NewFunction(functionType *FunctionType) Function {
	return Function{
		FunctionType: functionType,
	}
}
```

Also, test data for `Function` may need to be adjusted for the same reason.

turbolent: Good point, reverted adding it in f45e8d26d073e07b3238c961493126b518e1cd9a

turbolent (approved):

SupunS (approved):

PR #2391, opened 2023-03-17
Improvements to `CoverageReport` public API

### LINKED ISSUE #2390: Improvements to `CoverageReport` API
### Issue to be solved

The `CoverageReport` type should provide a more feature-rich public API, in order to empower developers to build tools around it, and gather meaningful insights. The idea is to provide the necessary information, for producing the kind of outcome that can be seen in the `Codecov` GitHub plugin. For example:

Screenshot from 2023-03-17 11-05-08

### Suggested Solution

The `CoverageReport` type should expose more public methods, and provide developers with a flexible API. Including methods to:
- Get the total percentage of covered statements, as a raw value
- Merge `CoverageReport` objects
- Unmarshal previously marshaled `CoverageReport` objects
- Calculate a `CoverageReport` summary and incremental diff between `CoverageReport` objects

m-Peter:
⚠️ Depends on #2396
Closes #2390

## Description

The `CoverageReport` type should expose more public methods, and provide developers with a flexible API. Including methods to:

- Get the total percentage of covered statements, as a raw value (now it is returned only through a human-friendly message)
- Merge `CoverageReport` objects
- Calculate a `CoverageReport` summary and incremental diff between `CoverageReport` objects
- Unmarshal previously marshaled `CoverageReport` objects

The ideal end result should contain the necessary information to build a tool like this:

Screenshot from 2023-03-17 11-05-08

Feedback is more than welcome! :pray: 

______

### DIFF
--- a/runtime/coverage.go
+++ b/runtime/coverage.go
@@ -202,8 +202,7 @@ func (r *CoverageReport) InspectProgram(location Location, program *ast.Program)
 			return true
 		})
 
-	locationCoverage := NewLocationCoverage(lineHits)
-	r.Coverage[location] = locationCoverage
+	r.Coverage[location] = NewLocationCoverage(lineHits)
 }
 
 // IsLocationInspected checks whether the given location,
--- a/runtime/coverage.go
+++ b/runtime/coverage.go
@@ -104,13 +104,13 @@ func NewLocationCoverage(lineHits map[int]int) *LocationCoverage {
 }
 
 // CoverageReport collects coverage information per location.
-// It keeps track of inspected programs per location, and can
-// also exclude locations from coverage collection.
+// It keeps track of inspected locations, and can also exclude
+// locations from coverage collection.
 type CoverageReport struct {
 	// Contains a *LocationCoverage per location.
 	Coverage map[common.Location]*LocationCoverage `json:"-"`
-	// Contains an *ast.Program per location.
-	Programs map[common.Location]*ast.Program `json:"-"`
+	// Contains locations whose programs are already inspected.
+	Locations map[common.Location]struct{} `json:"-"`
 	// Contains locations excluded from coverage collection.
 	ExcludedLocations map[common.Location]struct{} `json:"-"`
 }
@@ -131,13 +131,13 @@ func (r *CoverageReport) IsLocationExcluded(location Location) bool {
 // AddLineHit increments the hit count for the given line, on the given
 // location. The method call is a NO-OP in two cases:
 // - If the location is excluded from coverage collection
-// - If the location's *ast.Program, has not been inspected
+// - If the location has not been inspected for its statements
 func (r *CoverageReport) AddLineHit(location Location, line int) {
 	if r.IsLocationExcluded(location) {
 		return
 	}
 
-	if !r.IsProgramInspected(location) {
+	if !r.IsLocationInspected(location) {
 		return
 	}
 
@@ -146,14 +146,14 @@ func (r *CoverageReport) AddLineHit(location Location, line int) {
 }
 
 // InspectProgram inspects the elements of the given *ast.Program, and counts its
-// statements. If inspection is successful, the *ast.Program is marked as inspected.
+// statements. If inspection is successful, the location is marked as inspected.
 // If the given location is excluded from coverage collection, the method call
 // results in a NO-OP.
 func (r *CoverageReport) InspectProgram(location Location, program *ast.Program) {
 	if r.IsLocationExcluded(location) {
 		return
 	}
-	r.Programs[location] = program
+	r.Locations[location] = struct{}{}
 	lineHits := make(map[int]int, 0)
 	recordLine := func(hasPosition ast.HasPosition) {
 		line := hasPosition.StartPosition().Line
@@ -206,10 +206,10 @@ func (r *CoverageReport) InspectProgram(location Location, program *ast.Program)
 	r.Coverage[location] = locationCoverage
 }
 
-// IsProgramInspected checks whether the *ast.Program on the given
-// location, has been inspected or not.
-func (r *CoverageReport) IsProgramInspected(location Location) bool {
-	_, isInspected := r.Programs[location]
+// IsLocationInspected checks whether the given location,
+// has been inspected or not.
+func (r *CoverageReport) IsLocationInspected(location Location) bool {
+	_, isInspected := r.Locations[location]
 	return isInspected
 }
 
@@ -232,13 +232,13 @@ func (r *CoverageReport) String() string {
 }
 
 // Reset flushes the collected coverage information for all locations
-// and inspected programs. Excluded locations remain intact.
+// and inspected locations. Excluded locations remain intact.
 func (r *CoverageReport) Reset() {
 	for location := range r.Coverage { // nolint:maprange
 		delete(r.Coverage, location)
 	}
-	for location := range r.Programs { // nolint:maprange
-		delete(r.Programs, location)
+	for location := range r.Locations { // nolint:maprange
+		delete(r.Locations, location)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


runtime/coverage_test.go
```diff
@@ -658,6 +659,295 @@ func TestCoverageReportAddLineHitForNonInspectedProgram(t *testing.T) {
...
+
+	actual, err = json.Marshal(diff)
+	require.NoError(t, err)
+
+	expected = `
+	  {
+	    "coverage": "100.0%",
+	    "hits": 2,
+	    "locations": 0,
+	    "misses": -2,
```
m-Peter, line 823: Note: The negative value here represents the incremental diff, in comparison to the previous coverage report.

runtime/coverage.go
```diff
@@ -108,23 +107,23 @@ func NewLocationCoverage(lineHits map[int]int) *LocationCoverage {
 // also exclude locations from coverage collection.
 type CoverageReport struct {
 	// Contains a *LocationCoverage per location.
-	Coverage map[common.Location]*LocationCoverage `json:"-"`
+	Coverage map[string]*LocationCoverage `json:"-"`
```
m-Peter, line 110: Note: I have changed the type here, from `common.Location` to `string`, in order to facilitate the JSON unmarshal. I am not sure if there's any method available, that can convert a string to its respective `common.Location` implementation, e.g `AddressLocation`, `ScriptLocation`, `TransactionLocation` or `StringLocation`. The public API still accepts `common.Location` arguments, and uses the newly-added `ID` method, to convert it to a string.

runtime/coverage.go
```diff
@@ -108,23 +107,23 @@ func NewLocationCoverage(lineHits map[int]int) *LocationCoverage {
...
 	// Contains an *ast.Program per location.
-	Programs map[common.Location]*ast.Program `json:"-"`
+	Programs map[string]struct{} `json:"-"`
```
m-Peter, line 112: Note: Here I have entirely removed the `*ast.Program` value, because it was not utilized anywhere. This is reduce allocations, and also allow to properly setup a `CoverageReport` after JSON unmarshal, meaning that we can setup which locations/programs where inspected.

runtime/coverage.go
```diff
@@ -203,31 +202,32 @@ func (r *CoverageReport) InspectProgram(location Location, program *ast.Program)
...
-	totalCoveredLines := 0
-	for _, locationCoverage := range r.Coverage { // nolint:maprange
-		totalStatements += locationCoverage.Statements
-		totalCoveredLines += locationCoverage.CoveredLines()
-	}
-	percentage := fmt.Sprintf(
+// Percentage returns a string representation of the covered statements
+// percentage. It is defined as the ratio of total covered lines over
+// total statements, for all locations.
+func (r *CoverageReport) Percentage() string {
```
m-Peter, line 218: This method could be utilized [here](https://github.com/onflow/flow-cli/pull/827/files#diff-b46e7eeb2d9c6cf9e1f6d217bb509d83d98f83bea1eaac7b7c99fcbf534fa261R132), instead of using the result from `CoveredStatementsPercentage`.

runtime/coverage.go
```diff
@@ -236,49 +236,162 @@ func (r *CoverageReport) Reset() {
...
 			MissedLines: locationCoverage.MissedLines(),
 			Statements:  locationCoverage.Statements,
 			Percentage:  locationCoverage.Percentage(),
 		}
 	}
 	return json.Marshal(&struct {
-		Coverage map[string]LC `json:"coverage"`
-		*Alias
+		Coverage          map[string]lcAlias `json:"coverage"`
+		ExcludedLocations []string           `json:"excluded_locations"`
```
m-Peter, line 360: I have also added the `excluded_locations` field in the JSON marshal/unmarshal methods. Previously we didn't include this information, I thought it would be useful.

runtime/coverage.go
```diff
@@ -203,31 +202,32 @@ func (r *CoverageReport) InspectProgram(location Location, program *ast.Program)
...
+	return fmt.Sprintf(
 		"%0.1f%%",
 		100*float64(totalCoveredLines)/float64(totalStatements),
 	)
-	return fmt.Sprintf("Coverage: %v of statements", percentage)
+}
+
+// CoveredStatementsPercentage returns a human-friendly message
+// for the covered statements percentage.
+func (r *CoverageReport) CoveredStatementsPercentage() string {
```
turbolent, line 229: Maybe just implement `Stringer`, i.e. rename this function to `String()`?

turbolent (commented):
Nice! This is going to be super useful for CI 👌

runtime/coverage.go
m-Peter: Good idea :+1: 
Updated in https://github.com/onflow/cadence/pull/2391/commits/34e424953113deeb0a5179863c9c119aefaaf1ff

runtime/coverage.go
m-Peter: Thanks for pointing this out :clap: I was not aware of it.
My main concern is that the `ID` method truncates some information from locations (e.g qualified identifiers), which means that we are not able to map it directly with a `common.Location` implementation. To instantiate them, there are checks in place, such as:

```go
pieceCount := len(parts)
if pieceCount == 1 {
	return newError("missing qualified identifier")
}
```

which prevent us from creating the proper `common.Location` implementation, from its `ID`. Is there any workaround to this? Should we update the `ID` method to also include any qualified identifiers, thus enabling us to instantiate it afterwards, by the respective decoder (e.g `decodeFlowLocationTypeID`)?

m-Peter: As a starting point, I have added a first implementation on https://github.com/onflow/cadence/pull/2391/commits/e9110b9f1041e300481361f948570c47ac711e3e, so we have something to refine. I am open to better alternatives :bow:

turbolent: Ah, you are right, I didn't think about the decoding part.

Basically, so far we had a type ID, which is location + qualified identifier. Type IDs can be decoded using `DecodeTypeID`. 

Now we added a location ID, and we also want a decoding function `DecodeLocation`.

Thanks for getting started on an implementation, I'll have a look 👌 We might want to move it to `common` (maybe as `DecodeLocationID`?), and generalize it like in `DecodeTypeID`, so that we can add further location kinds in the future.

turbolent: I'll open a separate PR to rework `DecodeTypeID` to make the missing qualified identifier part a non-error, so the function can be used to also just decode location IDs.

m-Peter: That would be great :pray:

turbolent: Opened #2396

runtime/coverage.go
```diff
@@ -236,49 +240,214 @@ func (r *CoverageReport) Reset() {
...
+	coverage := make(map[string]lcAlias, 0)
+	cr := &struct {
+		Coverage          map[string]lcAlias `json:"coverage"`
+		ExcludedLocations []string           `json:"excluded_locations"`
+		*crAlias
+	}{
+		Coverage:          coverage,
+		ExcludedLocations: r.ExcludedLocationIDs(),
+		crAlias:           (*crAlias)(r),
+	}
```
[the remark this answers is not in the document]
m-Peter: Ah, good catch :+1: Fixed in https://github.com/onflow/cadence/pull/2391/files#diff-dd50d117359d58b9ec20ba8e56dc4b6b224a84ca819d943436617baec41d34d5R422-R428

runtime/coverage.go
m-Peter: PR is rebased on master, after https://github.com/onflow/cadence/pull/2396 was merged.

runtime/coverage.go
```diff
@@ -236,49 +236,162 @@ func (r *CoverageReport) Reset() {
...
-		delete(r.Programs, program)
+	for location := range r.Programs { // nolint:maprange
+		delete(r.Programs, location)
+	}
+}
+
+// Merge adds all the collected coverage information to the
+// calling object. Excluded locations are also taken into
+// account.
+func (r *CoverageReport) Merge(other CoverageReport) {
```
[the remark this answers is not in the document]
turbolent: I would leave it away, as it seems unexpected that there is also such a side effect on the argument.

runtime/coverage.go
```diff
@@ -202,83 +202,252 @@ func (r *CoverageReport) InspectProgram(location Location, program *ast.Program)
...
+	}{
+		crAlias: (*crAlias)(r),
+	}
+
+	if err := json.Unmarshal(data, cr); err != nil {
+		return err
+	}
+
+	for locationID, locationCoverage := range cr.Coverage { // nolint:maprange
+		location, _, _ := common.DecodeTypeID(nil, locationID)
```
turbolent, line 434: Maybe also handle the returned error here, and below

turbolent (commented):
Looks good! Just one last comment (handling errors) and then this PR should be ready to get merged 👍

runtime/coverage.go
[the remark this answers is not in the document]
m-Peter: Good!

turbolent (approved):
Great work! 👏

Issue #2215, opened 2023-01-04
Lazily instantiated stdlib values related to Test contract 


turbolent:
### Issue To Be Solved

Currently the code related to the `Test` contract in https://github.com/onflow/cadence/blob/master/runtime/stdlib/test.go is eagerly instantiated, even when it is not used.

### Suggested Solution

Initialize the values for the `Test` contract, like the checker, lazily.

This was already done previously for the `Crypto` contract in #2183.

turbolent:
cc @SupunS

PR #2403, opened 2023-03-28
Initialize the Test contract lazily

m-Peter:
Closes https://github.com/onflow/cadence/issues/2215

## Description

Lazily instantiated `stdlib` values related to `Test` contract.
Since we have an accepted proposal (https://github.com/onflow/developer-grants/issues/148), regarding the Cadence testing framework, we thought it would be a good first issue to tackle, in order to get acquainted with the codebase :pray: 
______


SupunS (approved):
Nice! Thank you for adding this!

runtime/stdlib/test.go
```diff
@@ -153,7 +176,7 @@ func memberNotFoundError(parentType, member string) error {
 }
 
 var blockchainBackendInterfaceType = func() *sema.InterfaceType {
-	typ, ok := testContractType.NestedTypes.Get(blockchainBackendTypeName)
+	typ, ok := TestContractType().NestedTypes.Get(blockchainBackendTypeName)
```
turbolent, line 179: Refactoring `TestContractType` to be initialized lazily is a good start, but the call here and below will cause it to immediately initialized.

To prevent this, the rest of the file must also be refactored in a similar fashion.

turbolent (commented):
Great start 👏

runtime/stdlib/test.go
m-Peter: Ah, I see. My guide here was the comparison with the PR that did the same for the `Crypto` contract. I will update the rest of the values, to be initialized lazily :+1:

bluesign:
why this is in cadence ? shouldn't be in https://github.com/onflow/cadence-tools/test ?

SupunS:
> why this is in cadence ?

To avoid circular dependency. Because the test module depends on cadence go dependency already. Moving this to test would cause cadence to also depends on the test module, creating the cycle.
We could eventually move all standard-libs outside of cadence, but that going to need some refactoring and a dependency injection mechanism, so that stdlibs can be provided to the typechecker/runtime, without a hard dependency.

bluesign:
> > why this is in cadence ?
> 
> To avoid circular dependency. Because the test module depends on cadence already (at go level). Moving this to test would cause cadence to also depends on the test module, creating the cycle. We could eventually move all standard-libs outside of cadence, but that going to need some refactoring and a dependency injection mechanism, so that stdlibs can be provided to the typechecker/runtime, without a hard dependency.

can't we simple expose `Test` from rest runner? If test depends on cadence, but cadence doesn't depend on test.

turbolent:
@bluesign The current idea is that Cadence itself exposes a test framework in its standard library. `cadence-tools/test` is just one possible concrete test runner that uses the FVM/Emulator. However, one could imagine other implementations for Flow, and even implementations for other chains. 

That is similar to e.g. Rust having testing support in the language and the cargo test runner tool.

That does not mean that it has to stay that way of course, we could consider refactoring the test framework out of Cadence. What do you think would be the advantages of that?

SupunS:
> can't we simple expose Test from test runner? If test depends on cadence, but cadence doesn't depend on test.

Cadence needs to know about the `Test` contract (i.e: stdlib) statically. So moving it to the test module will create the cycle.

bluesign:
> That does not mean that it has to stay that way of course, we could consider refactoring the test framework out of Cadence. What do you think would be the advantages of that?

I don't know if any advantage tbh; but I always thought, host can expose stuff ( like flow-go was exposing some account freezing before ) to Cadence. As we don't use test on mainnet/testnet, I thought that would be the way.

turbolent:
@bluesign Sounds good. Maybe open a new issue for that, as this PR only really is an optimization of the existing code

runtime/stdlib/test.go
m-Peter: I found some time to get back to this. It needed quite some rework, to make it truly lazily. But hopefully I got it right this time :pray:

m-Peter: The three exported functions that can trigger the initialization of `stdlib` values related to the `Test` contract, are:
- `TestContractChecker()`
- `TestContractType()`
- `TestContractInitializerTypes()` (this is called in `NewTestContract` exported function)

runtime/stdlib/test.go
```diff
@@ -779,9 +886,45 @@ var newMatcherFunction = interpreter.NewUnmeteredHostFunctionValue(
 
 const emulatorBackendTypeName = "EmulatorBackend"
 
-var EmulatorBackendType = func() *sema.CompositeType {
```
m-Peter, line 782: For some reason, the `EmulatorBackendType` var was exported, however it is not used anywhere outside this package, nor in the `cadence-tools/test` repository. That's why I changed it to not be exported (https://github.com/onflow/cadence/pull/2403/files#diff-4709f76365b48c2caca87b859afcc573afb62cb4e6186b9431d2b540972d402fR111)

turbolent: Yeah, looks like it isn't used anywhere else either: https://github.com/search?q=language%3AGo+EmulatorBackendType&type=code&ref=advsearch

m-Peter:
@turbolent If this is in good shape, could we get it merged? I want to open another PR (for https://github.com/onflow/cadence-tools/issues/108), however, I have branched-off of this one, to avoid conflicts :pray:

turbolent:
@m-Peter Nice work, I'll review tomorrow

runtime/stdlib/test.go
```diff
@@ -85,147 +172,107 @@ var TestContractChecker = func() *sema.Checker {
...
-	)
-	if err != nil {
-		return nil, err
+	testContractInitializerTypes = make([]sema.Type, len(testContractType.ConstructorParameters))
+	for i, parameter := range testContractType.ConstructorParameters {
+		testContractInitializerTypes[i] = parameter.TypeAnnotation.Type
 	}
 
-	compositeValue := value.(*interpreter.CompositeValue)
+	initBlockchainBackendInterfaceType()
```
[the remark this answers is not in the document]
m-Peter: You are right, it has become unwieldy, to be honest. I will attempt to reduce most of them, as you suggested :+1:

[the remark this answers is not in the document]
m-Peter: I have removed quite many of them, however some are not possible to be removed, since they need to be accessed in the `NewTestContract` exported function. If you have any workaround in mind, please let me know :pray:

turbolent (commented):
Thanks Ardit for continuing to improve this!

It would be nice if this could be refactored more to avoid all the global variables. 

In the current version, the main initialization function, `initTest`, triggers the initialization of all those globals in an imperative way, which is then often followed by a use of the global for further initialization, and later to also construct a value. Between the globals there are many inter-dependencies.

For example, `initEmulatorBackendType` initializes the global `emulatorBackendType`, but it actually depends on the function globals, i.e. `initEmulatorBackendFunctions` having being called and e.g. the global `emulatorBackendExecuteScriptFunctionType` having been initialized. Finally, `emulatorBackendType` exists because it is needed in `newEmulatorBackend`.

Instead, maybe the initialization could be done in a more "functional" way without side-effects / global initialization. This could be done e.g. by introducing a type `TestEmulatorBackendType`, which holds the composite type and all function types. The type is then returned from `initEmulatorBackendType`, instead of it mutating a global. `initTest` could then be the place where the whole group is initialized.

I'll try to push a commit illustrating it and we can then refactor the other types too.

runtime/stdlib/test.go
```diff
@@ -349,33 +263,186 @@ func init() {
...
+		panic(errors.NewUnexpectedError(
+			"invalid type for '%s'. expected interface",
+			blockchainBackendTypeName,
+		))
+	}
+	return blockchainBackendInterfaceType
+}
+
+func initMatcherType() *sema.CompositeType {
+	typ, ok := testContractType.NestedTypes.Get(matcherTypeName)
```
turbolent, line 304: Here and elsewhere: Instead of accessing the global here (in this case, `testContractType`), pass it as a parameter

m-Peter:
Thanks for the heavy lifting :raised_hands: :100: 
This does look way much better now :tada: 
I guess there is no remainder here for refactoring, right?

turbolent (approved):

turbolent:
@m-Peter No, that should be all now. We'll need to update some downstream dependencies, but the required changes are very small.

@dsainati1 @SupunS Could you please review?

runtime/stdlib/test_contract.go
SupunS, line 1: Just to be sure: this is only refactoring, and no change to the functionality, correct?

SupunS (approved):

runtime/stdlib/test_contract.go
m-Peter: Yes, exactly :+1:

dsainati1 (approved):

runtime/stdlib/test_contract.go
turbolent: Yes, I only moved things around and did not change any logic

PR #2404, opened 2023-03-29
Type checking for entitlement and entitlement mapping declarations

dsainati1:
Part of the implementation of https://github.com/onflow/flips/pull/54

This implements type checking for the declaration of entitlements and entitlement mappings, interface conformance checking, and relative permission comparison. 
______


runtime/sema/access.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+}
+
+func (EntitlementAccess) isAccess() {}
+
+func (a EntitlementAccess) Access() ast.Access {
+	return a.astAccess
+}
+
+func (e EntitlementAccess) PermitsAccess(other Access) bool {
```
dsainati1, line 72: This function essentially implements the core logic for comparing entitlements:
* If an entitlement set `X` permits access of an entitlement set `Y`, that effectively means that a reference authorized to `Y` is a subtype of a reference authorized to `X`. 
* Equivalently, `X` permitting `Y` means that a `Y`-entitled reference is permitted to access an `X`-access member on a composite. 
* Equivalently, an access modifier of `X` on a composite is a valid implementation of the same member declared with `Y` access on an interface (access is contravariant). 

There are 4 essential combinations of checks we might need to perform when determining whether one set permits another depending on whether the two sets are conjunctions, disjunctions, or one of each.

PR #2406, opened 2023-03-30
Remove parameter name from API documentation when argument label is given

MaxStalker:
## Description
Documentation for `type` method have argument name specified as `at path` (space included). While it's correct Cadence it's a bit misleading for users of said method, since they care more about how they need to call it instead of how arg is named in the body of the function.

______


turbolent (approved):
Thank you for bringing this up! Agreed that we should only document external-facing API, as this is what consumers of the API will use -- the parameter name is only used internally, so is useless and potentially confusing.

Could you maybe also do a quick scan through the rest of the docs and see if we have any similar cases like this?

turbolent:
Adjusted some more cases I could find

SupunS (approved):

bluesign:
I learn something new everyday. I am tempted to say this is ( looks like ) bug more than feature though.

turbolent:
@bluesign This feature is available in many languages: Swift, which was inspired by Objective-C; Objective-C, which was inspired by Smalltalk. Smalltalk was released 50 (!) years ago

bluesign:
@turbolent yeah bug was a bit of a wrong statement, maybe let's say unintuitive. Problem is not the label part, somehow defining the label

let me explain quickly: In the grammar we have label can be ( identifier or _ ) but also it can be derived from variable name 

```
pub fun f1(a: String) 
pub fun f2(_ a: String)
pub fun f3(a a: String)
pub fun f4(a b: String)
```

except f2 all has label `a`, I think this is the confusing part of the syntax. ( f1 vs others ) 

we defined: `functionName ( label [variableName]: type ) ` where variable name is optional. But instead if we had their label is optional, things would make more sense in my opinion. `pub fun f1(a: String) == pub fun f1(_ a: String) `

I am not sure if I can explain well enough ( probably not )

PR #2414, opened 2023-04-02
Allow equality comparisons for dictionaries

### LINKED ISSUE #2275: Allow equality comparisons for dictionaries
### Issue to be solved

Similar to #2235, we currently don't allow dictionaries to be tested for equality. Logically, two dictionaries of type `{K: V}` are equal if they contain the same set of key-value pairs, and both `K` and `V` types can be tested for equality.

### Suggested Solution

This feature is [currently implemented](https://github.com/onflow/cadence/blob/5ffebbd88976ea141ef9c460da076e3f857d7e12/runtime/interpreter/value.go#L15921) but is disabled in the checker. The work is basically identical to PR #2271.

Documentation and tests will need to be updated to reflect the new behaviour. This is a non-breaking change, as this adds new functionality to the language.

darkdrag00nv2:
Closes #2275

## Description

The functionality was already implemented. Just enabled it in the checker.

______

### DIFF
--- a/runtime/tests/checker/arrays_dictionaries_test.go
+++ b/runtime/tests/checker/arrays_dictionaries_test.go
@@ -317,6 +317,91 @@ func TestCheckDictionaryValues(t *testing.T) {
 	)
 }
 
+func TestCheckDictionaryEqual(t *testing.T) {
+	t.Parallel()
+
+	testValid := func(name, code string) {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			_, err := ParseAndCheck(t, code)
+			require.NoError(t, err)
+		})
+	}
+
+	assertInvalid := func(name, code string) {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			_, err := ParseAndCheck(t, code)
+			errs := RequireCheckerErrors(t, err, 1)
+			assert.IsType(t, &sema.InvalidBinaryOperandsError{}, errs[0])
+		})
+	}
+
+	for _, opStr := range []string{"==", "!="} {
+		testValid("self_dict_equality", fmt.Sprintf(`
+			fun test(): Bool {
+				let d = {"abc": 1, "def": 2}
+				return d %s d
+			}`, opStr))
+
+		testValid("self_dict_equality_nested_1", fmt.Sprintf(`
+			fun test(): Bool {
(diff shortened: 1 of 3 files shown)


runtime/tests/checker/arrays_dictionaries_test.go
```diff
@@ -317,6 +317,91 @@ func TestCheckDictionaryValues(t *testing.T) {
...
+			assert.IsType(t, &sema.InvalidBinaryOperandsError{}, errs[0])
+		})
+	}
+
+	for _, opStr := range []string{"==", "!="} {
+		testValid("self_dict_equality", fmt.Sprintf(`
+			fun test(): Bool {
+				let d = {"abc": 1, "def": 2}
+				return d %s d
+			}`, opStr))
```
turbolent, line 347: Minor: Could you please chop down the argument lists? That would make the code a bit more readable.
For example here:

```suggestion
		testValid(
			"self_dict_equality", 
			fmt.Sprintf(
				`
					fun test(): Bool {
						let d = {"abc": 1, "def": 2}
						return d %s d
					}
				`,
				opStr,
			),
		)
```

turbolent (approved):
Nice! Thank you

runtime/tests/checker/arrays_dictionaries_test.go
```diff
@@ -317,6 +317,91 @@ func TestCheckDictionaryValues(t *testing.T) {
...
+				return d %s d2
+			}`, opStr))
+
+		assertInvalid("dict_equality_invalid_nested", fmt.Sprintf(`
+			fun test(): Bool {
+				let d = {"abc": {1: {"a": 1000}, 2: {"b": 2000}}, "def": {4: {"c": 1000}, 5: {"d": 2000}}}
+				let d2 = {"abc": {1: {1000: "a"}, 2: {2000: "b"}}, "def": {4: {1000: "c"}, 5: {2000: "d"}}}
+				return d %s d2
+			}`, opStr))
+	}
```
turbolent, line 402: Could you please add test cases where left and right side are the same type, but that type is unequatable? For example, functions are not equatable, so you could use a type like `{String: ((): Void)}`

runtime/tests/checker/arrays_dictionaries_test.go
darkdrag00nv2: Done for all the tests.

dsainati1 (approved):

Issue #2415, opened 2023-04-03
Introduce a `CoverageReport` field on `runtime.Config` type


m-Peter:
### Issue to be solved

As of now, the `runtime.Config` type, has a `boolean` field (`CoverageReportingEnabled`) for enabling/disabling coverage reporting. However, in most places, the actual `CoverageReport` object, has to be injected somehow.

### Suggested Solution

Introduce a `CoverageReport` field, which will hold the actual object. This can also deprecate the usage of `CoverageReportingEnabled`, by checking whether the `CoverageReport` is `nil` or not. If it's present, then coverage reporting should be enabled.

PR #2416, opened 2023-04-03
Introduce a `CoverageReport` field on `runtime.Config` type

m-Peter:
Closes https://github.com/onflow/cadence/issues/2415

## Description

Introduce a `CoverageReport` field, which will hold the actual object. This can also deprecate the usage of `CoverageReportingEnabled`, by checking whether the `CoverageReport` is `nil` or not. If it's present, then coverage reporting should be enabled.
______

### DIFF
--- a/runtime/config.go
+++ b/runtime/config.go
@@ -34,7 +34,10 @@ type Config struct {
 	// ResourceOwnerChangeCallbackEnabled configures if the resource owner change callback is enabled
 	ResourceOwnerChangeHandlerEnabled bool
 	// CoverageReportingEnabled configures if coverage reporting is enabled
+	// Deprecated: Use the `CoverageReport` field instead
 	CoverageReportingEnabled bool
+	// CoverageReport enables and collects coverage reporting metrics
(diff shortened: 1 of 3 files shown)


runtime/config.go
```diff
@@ -34,7 +34,10 @@ type Config struct {
 	// ResourceOwnerChangeCallbackEnabled configures if the resource owner change callback is enabled
 	ResourceOwnerChangeHandlerEnabled bool
 	// CoverageReportingEnabled configures if coverage reporting is enabled
+	// Deprecated: Use the `CoverageReport` field instead
 	CoverageReportingEnabled bool
```
turbolent, line 38: Given that the field is not used anymore, just remove it
```suggestion
```

turbolent (approved):
👍

runtime/config.go
m-Peter: Good point :+1:  I have removed it entirely in this commit https://github.com/onflow/cadence/pull/2416/commits/cc5644b671468ee11075b1a81cd1d7797a895b2a

Issue #1889, opened 2022-08-18
Testing Framework Improvements


SupunS:
## Issue To Be Solved

https://github.com/onflow/cadence/issues/331 proposed a testing framework for cadence. An initial version of this proposal is now implemented, and supports basic functionalities to write tests.

However, there is still some room for improvement, and not all functionalities suggested in https://github.com/onflow/cadence/issues/331 are implemented.

Below is a set of improvements remaining to be done:

## Assertions
- [x] #1959
  `fun fail(message: string)`
   Currently there's a `fail()` function that takes no arguments. Remove the existing implementation, and implement this natively, to accept an optional 'message' argument. 
  Since the argument is optional, this is a backward compatible change 

## Matchers
https://github.com/onflow/cadence/pull/1863 laid the foundation for implementing matches. It also added `equal()` matcher that can be used to test the equality of values. 

The proposal suggests some more convenient functions for creating most commonly used matchers. Many of these might be able to implement in cadence (some may have to be implemented natively).

- Matcher functions:
  - [ ]  `fun haveElementCount(_ count: Int): Matcher`:
    returns a matcher that succeeds if the tested value is an array or dictionary, and has the given number of elements
  - [ ] `fun beEmpty(): Matcher`:
    returns a matcher that succeeds if the tested value is an array or dictionary, and the tested value contains no elements
  - [ ] `fun beNil(): Matcher`:
    returns a matcher that succeeds if the tested value is nil
  - [ ] `fun contain(_ element: AnyStruct): Matcher`:
    returns a matcher that succeeds if the tested value is an array that contains a value that is equal to the given value, or the tested value is a dictionary that contains an entry where the value is equal to the given value
  - [ ] `fun beGreaterThan(_ value: Number): Matcher`
    returns a matcher that succeeds if the tested value is a number and greater than the given number
  - [ ] `fun beLessThan(_ value: Number): Matcher`
    returns a matcher that succeeds if the tested value is a number and less than the given number
  - [ ] `fun beSucceeded(): Matcher`
    returns a matcher that succeeds if the tested value is a transaction result or a script result, and the result has a succeeded status
  - [ ] `fun beFailed(): Matcher`
    returns a matcher that succeeds if the tested value is a transaction result or script result, and the result has a failed status
- Built-in matcher combinators:
  - [ ] `fun not(matcher: Matcher): Matcher`:
    returns a matcher that negates the given matcher

## Unit Testing
[...]

PR #2420, opened 2023-04-04
Add more built-in matchers to the Cadence Testing Framework

m-Peter:
**Work towards:** https://github.com/onflow/developer-grants/issues/148

## Description

Implements the matchers described in https://github.com/onflow/cadence/issues/1889
______

### DIFF
--- a/runtime/stdlib/contracts/test.cdc
+++ b/runtime/stdlib/contracts/test.cdc
@@ -270,7 +270,7 @@ pub contract Test {
             } else if let result = value as? ScriptResult {
                 return result.status == ResultStatus.succeeded
             } else {
-                panic("Only TransactionResult & ScriptResult types are supported.")
+                panic("expected TransactionResult or ScriptResult argument")
             }
             return false
         })
@@ -287,7 +287,7 @@ pub contract Test {
             } else if let result = value as? ScriptResult {
                 return result.status == ResultStatus.failed
             } else {
-                panic("Only TransactionResult & ScriptResult types are supported.")
+                panic("expected TransactionResult or ScriptResult argument")
             }
             return false
         })
--- a/runtime/stdlib/contracts/test.cdc
+++ b/runtime/stdlib/contracts/test.cdc
@@ -250,4 +250,13 @@ pub contract Test {
         ///
         pub fun useConfiguration(_ configuration: Configuration)
     }
+
+    /// Returns a new matcher that negates the test of the given matcher.
+    ///
+    pub fun not(_ matcher: Matcher): Matcher {
+        return Matcher(test: fun (value: AnyStruct): Bool {
+            return !matcher.test(value)
+        })
+    }
+
 }
--- a/runtime/stdlib/contracts/test.cdc
+++ b/runtime/stdlib/contracts/test.cdc
@@ -291,4 +291,12 @@ pub contract Test {
         })
     }
 
+    /// Returns a new matcher that checks if the given test value is nil.
+    ///
+    pub fun beNil(): Matcher {
+        return Matcher(test: fun (value: AnyStruct): Bool {
+            return value == nil
+        })
+    }
+
 }
--- a/runtime/stdlib/contracts/test.cdc
+++ b/runtime/stdlib/contracts/test.cdc
@@ -132,9 +132,18 @@ pub contract Test {
         pub case failed
     }
 
+    /// Result is the interface to be implemented by the various execution
+    /// operations, such as transactions and scripts.
+    ///
+    pub struct interface Result {
+        /// The resulted status of an executed operation.
+        ///
+        pub let status: ResultStatus
+    }
+
     /// The result of a transaction execution.
     ///
-    pub struct TransactionResult {
+    pub struct TransactionResult: Result {
         pub let status: ResultStatus
         pub let error: Error?
 
@@ -146,7 +155,7 @@ pub contract Test {
 
     /// The result of a script execution.
     ///
-    pub struct ScriptResult {
+    pub struct ScriptResult: Result {
         pub let status: ResultStatus
         pub let returnValue: AnyStruct?
         pub let error: Error?
@@ -265,14 +274,7 @@ pub contract Test {
     ///
     pub fun beSucceeded(): Matcher {
         return Matcher(test: fun (value: AnyStruct): Bool {
-            if let result = value as? TransactionResult {
-                return result.status == ResultStatus.succeeded
-            } else if let result = value as? ScriptResult {
-                return result.status == ResultStatus.succeeded
-            } else {
-                panic("expected TransactionResult or ScriptResult argument")
-            }
-            return false
+            return (value as! {Result}).status == ResultStatus.succeeded
         })
     }
 
@@ -282,14 +284,7 @@ pub contract Test {
     ///
     pub fun beFailed(): Matcher {
         return Matcher(test: fun (value: AnyStruct): Bool {
-            if let result = value as? TransactionResult {
-                return result.status == ResultStatus.failed
-            } else if let result = value as? ScriptResult {
-                return result.status == ResultStatus.failed
-            } else {
-                panic("expected TransactionResult or ScriptResult argument")
-            }
-            return false
+            return (value as! {Result}).status == ResultStatus.failed
         })
     }
 
--- a/runtime/stdlib/contracts/test.cdc
+++ b/runtime/stdlib/contracts/test.cdc
@@ -275,4 +275,20 @@ pub contract Test {
         })
     }
 
+    /// Returns a new matcher that checks if the given test value is either
+    /// a ScriptResult or TransactionResult and the ResultStatus is failed.
+    /// Returns false in any other case.
+    ///
+    pub fun beFailed(): Matcher {
+        return Matcher(test: fun (value: AnyStruct): Bool {
+            if let result = value as? TransactionResult {
+                return result.status == ResultStatus.failed
+            } else if let result = value as? ScriptResult {
+                return result.status == ResultStatus.failed
+            } else {
+                return false
+            }
+        })
+    }
+
 }
--- a/runtime/stdlib/contracts/test.cdc
+++ b/runtime/stdlib/contracts/test.cdc
@@ -138,7 +138,7 @@ pub contract Test {
         pub let status: ResultStatus
         pub let error: Error?
 
-        init(status: ResultStatus, error: Error) {
+        init(status: ResultStatus, error: Error?) {
             self.status = status
             self.error = error
         }
@@ -256,6 +256,22 @@ pub contract Test {
     pub fun not(_ matcher: Matcher): Matcher {
         return Matcher(test: fun (value: AnyStruct): Bool {
             return !matcher.test(value)
+        })
+    }
+
+    /// Returns a new matcher that checks if the given test value is either
+    /// a ScriptResult or TransactionResult and the ResultStatus is succeeded.
+    /// Returns false in any other case.
+    ///
+    pub fun beSucceeded(): Matcher {
+        return Matcher(test: fun (value: AnyStruct): Bool {
+            if let result = value as? TransactionResult {
+                return result.status == ResultStatus.succeeded
+            } else if let result = value as? ScriptResult {
+                return result.status == ResultStatus.succeeded
+            } else {
+                return false
+            }
         })
     }
 
--- a/runtime/stdlib/contracts/test.cdc
+++ b/runtime/stdlib/contracts/test.cdc
@@ -270,8 +270,9 @@ pub contract Test {
             } else if let result = value as? ScriptResult {
                 return result.status == ResultStatus.succeeded
             } else {
-                return false
+                panic("Only TransactionResult & ScriptResult types are supported.")
             }
+            return false
         })
     }
 
@@ -286,8 +287,9 @@ pub contract Test {
             } else if let result = value as? ScriptResult {
                 return result.status == ResultStatus.failed
             } else {
-                return false
+                panic("Only TransactionResult & ScriptResult types are supported.")
             }
+            return false
         })
     }
 
--- a/runtime/stdlib/test.go
+++ b/runtime/stdlib/test.go
@@ -124,6 +124,7 @@ func NewTestContract(
 	// Inject natively implemented matchers
 	compositeValue.Functions[newMatcherFunctionName] = newMatcherFunction
 	compositeValue.Functions[equalMatcherFunctionName] = equalMatcherFunction
+	compositeValue.Functions[beEmptyMatcherFunctionName] = beEmptyMatcherFunction
 
 	return compositeValue, nil
 }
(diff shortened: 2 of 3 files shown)


runtime/stdlib/contracts/test.cdc
```diff
@@ -123,6 +123,14 @@ pub contract Test {
                 return self.test(value) || other.test(value)
             })
         }
+
+        /// Returns a new matcher that negates the test of this matcher.
+        ///
+        pub fun not(): Matcher {
```
m-Peter, line 129: In the original [issue](https://github.com/onflow/cadence/issues/1889), the method signature is:

```cadence
pub fun not(matcher: Matcher): Matcher {
    ...
}
```

However, I don't see how the parameter is of any value here. `not` is a unary operator, unlike the `and` & `or` operators. Correct me If I understood this in the wrong way.

turbolent: I think the idea is that the function is not defined on `Matcher`, but is a global function, so one could say `not(someMatcher)`

runtime/stdlib/contracts/test.cdc
```diff
@@ -250,4 +258,49 @@ pub contract Test {
...
+    pub fun beSucceeded(): Matcher {
+        return Matcher(test: fun (value: AnyStruct): Bool {
+            if value.getType() == Type<TransactionResult>() {
+                let result = value as! TransactionResult
+                return result.status == ResultStatus.succeeded
+            } else if value.getType() == Type<ScriptResult>() {
+                let result = value as! ScriptResult
+                return result.status == ResultStatus.succeeded
+            } else {
+                return false
```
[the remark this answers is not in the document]
turbolent: In `TestContractChecker`, `panic` needs to be declared in the base activation of the checker.

After that, it must be also declared in the base activation of interpreters using the `Test` contract, e.g. in tests in `newTestContractInterpreter`. I think the test runner which uses this test framework, in `github.comcadence-tools/test`, already injects the `panic` function, but we'll need to see.

runtime/stdlib/contracts/test.cdc
m-Peter: That makes sense, indeed. However, that would mean it's no longer a matcher combinator. I can make the change easily, I was just wondering what was the desired functionality here.

runtime/stdlib/contracts/test.cdc
```diff
@@ -250,4 +258,49 @@ pub contract Test {
...
+        })
+    }
+
+    /// Returns a new matcher that checks if the given test value is either
+    /// a ScriptResult or TransactionResult and the ResultStatus is failed.
+    /// Returns false in any other case.
+    ///
+    pub fun beFailed(): Matcher {
+        return Matcher(test: fun (value: AnyStruct): Bool {
+            if value.getType() == Type<TransactionResult>() {
```
turbolent, line 286: Same as above, use failable downcasting (`as?`) instead

runtime/stdlib/test.go
```diff
@@ -288,6 +293,61 @@ func init() {
 		),
 	)
 
+	// Test.beEmpty()
```
turbolent, line 296: Could these new functions also just be declared in Cadence (`test.cdc`)? Ideally we would implement as much as possible in Cadence itself.

Not sure why `equal` was already implemented in Go, it looks implementable in Cadence, too.

turbolent (commented):
Nice!

runtime/stdlib/test.go
SupunS: Not all types are equatable at the Cadence level. So `equal` method had to be implemented in Go.

But yeah, +1 for implementing these matchers as much as possible in Cadence, if possible.

runtime/stdlib/contracts/test.cdc
m-Peter: Fixed in https://github.com/onflow/cadence/pull/2420/commits/3b32a0ca3d9850a32cbe1f9d213c0f20ce908580

runtime/stdlib/contracts/test.cdc
```diff
@@ -250,4 +258,49 @@ pub contract Test {
...
+
+    /// Returns a new matcher that checks if the given test value is either
+    /// a ScriptResult or TransactionResult and the ResultStatus is succeeded.
+    /// Returns false in any other case.
+    ///
+    pub fun beSucceeded(): Matcher {
+        return Matcher(test: fun (value: AnyStruct): Bool {
+            if value.getType() == Type<TransactionResult>() {
+                let result = value as! TransactionResult
+                return result.status == ResultStatus.succeeded
```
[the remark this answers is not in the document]
m-Peter: Nice one :100: Fixed in https://github.com/onflow/cadence/pull/2420/commits/1530e25bb4975c79fc13c449d1ee6515a22a9b97

runtime/stdlib/contracts/test.cdc
m-Peter: Fixed in https://github.com/onflow/cadence/pull/2420/commits/27bed60c70d2e52ee2aa0c4fdd60df8c03a993db

runtime/stdlib/test.go
m-Peter: I made a first attempt to port all of the 5 remaining matchers in Cadence:

```cadence
pub fun beLessThan(_ value: Number): Matcher {
    return Matcher(test: fun (testValue: AnyStruct): Bool {
        if let otherValue = testValue as? Number {
            return otherValue < value
        }
        return false
    })
}
```
**fails with:**
```bash
panic: Checking failed:
error: cannot apply binary operation < to types: `Number`, `Number`
```
---

```cadence
pub fun beGreaterThan(_ value: Number): Matcher {
    return Matcher(test: fun (testValue: AnyStruct): Bool {
        if let otherValue = testValue as? Number {
            return otherValue > value
        }
        return false
    })
}
```
**fails with:**
```bash
panic: Checking failed:
error: cannot apply binary operation > to types: `Number`, `Number`
```
---

```cadence
pub fun beEmpty(): Matcher {
    return Matcher(test: fun (testValue: AnyStruct): Bool {
        if let arr = testValue as? [AnyStruct] {
            return arr.length == 0
        } else if let dict = testValue as? {AnyStruct: AnyStruct} {
            return dict.length == 0
        }
        return false
    })
}
```
**fails with:**
```bash
panic: Checking failed:
error: cannot use type as dictionary key type: `AnyStruct`
```
---

```cadence
pub fun haveElementCount(_ count: Int): Matcher {
    return Matcher(test: fun (testValue: AnyStruct): Bool {
        if let arr = testValue as? [AnyStruct] {
            return arr.length == count
        } else if let dict = testValue as? {AnyStruct: AnyStruct} {
            return dict.length == count
        }
        return false
    })
}
```
**fails with:**
```bash
panic: Checking failed:
error: cannot use type as dictionary key type: `AnyStruct`
```
---

```cadence
pub fun contain(_ element: AnyStruct): Matcher {
    return Matcher(test: fun (testValue: AnyStruct): Bool {
        if let arr = testValue as? [AnyStruct] {
            return arr.contains(element)
        } else if let dict = testValue as? {AnyStruct: AnyStruct} {
            return dict.containsKey(element)
        }
        return false
    })
}
```
**fails with:**
```bash
panic: Checking failed:
error: cannot compare value which has type: `AnyStruct`

error: cannot use type as dictionary key type: `AnyStruct`
```

Any workarounds for these errors?

runtime/stdlib/contracts/test.cdc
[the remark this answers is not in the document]
m-Peter: Thanks for the hint :pray: I have added these commit (https://github.com/onflow/cadence/pull/2420/commits/0fea290478be8409ad57597d3690815f46ab754d, https://github.com/onflow/cadence/pull/2420/commits/d1fbaa23bd7d4edddf4084cba9b00755e3533bea).
However, I think we should probably add all these built-in functions (https://github.com/onflow/cadence/blob/master/runtime/stdlib/builtin.go#L36-L52).
I was trying to write some tests for this contract (https://github.com/justjoolz/PRNG/blob/master/cadence/contracts/PRNG.cdc#L78), and it failed when it reached the `getBlock(at: blockHeight)!.id` function call.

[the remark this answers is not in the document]
turbolent: Yeah, that might be a good idea, maybe as a separate PR.

The language server is doing something like that already, and maybe we can just refactor both at the same time: https://github.com/onflow/cadence-tools/blob/988ec8bbef8b63adb43debf901ec41df8065ff50/languageserver/server/stdlib.go

Basically, we want to use `DefaultStandardLibraryValues`/`DefaultScriptStandardLibraryValues`, and have an implementation of the handler that panics when the functions are used, because they are unavailable.

runtime/stdlib/test.go
turbolent: Thank you for giving it a shot! 👏 

- Supporting comparisons of `Number`s, i.e. potentially different kinds of numbers, is probably not something we would want to add. You could extend the implementation of the matcher to try and downcast to each concrete number type (e.g. cast LHS and RHS to `UInt8` and compare, then to `UInt16`, and so on), but there are a lot of number types.
- Currently there is no type exposed to Cadence programs which represents the type that can be used as a `Dictionary` key type. Basically, we would want to introduce a "`HashableStruct`" interface, and have all types mentioned in `IsValidDictionaryKeyType` conform to it.

In general, given the Cadence implementation are not (yet) possible, let's just keep the Go implementations for now 👍

runtime/stdlib/contracts/test.cdc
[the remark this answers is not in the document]
m-Peter: Sounds good :+1:

runtime/stdlib/contracts/test.cdc
```diff
@@ -250,4 +250,55 @@ pub contract Test {
         ///
         pub fun useConfiguration(_ configuration: Configuration)
     }
+
+    /// Returns a new matcher that checks if the given test value is either
+    /// a ScriptResult or TransactionResult and the ResultStatus is succeeded.
+    /// Returns false in any other case.
+    ///
+    pub fun beSucceeded(): Matcher {
+        return Matcher(test: fun (value: AnyStruct): Bool {
```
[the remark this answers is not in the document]
m-Peter: Nice one :100:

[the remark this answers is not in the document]
m-Peter: @SupunS I was able to get rid of these `if-else` statements, however, changing the matcher function to accept a `{Result}` value type, resulted in a type mismatch. Which makes sense, given that the type of the `test` function is:

```cadence
pub let test: ((AnyStruct): Bool)
```
Correct me if I'm wrong though :pray:

runtime/stdlib/contracts/test.cdc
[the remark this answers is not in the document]
m-Peter: Opened this [issue](https://github.com/onflow/cadence-tools/issues/96), after your comment.

runtime/stdlib/test.go
```diff
@@ -1367,6 +1428,280 @@ var equalMatcherFunction = interpreter.NewUnmeteredHostFunctionValue(
...
+			matcherTestFunctionType,
+			func(invocation interpreter.Invocation) interpreter.Value {
+				var isEmpty bool
+				switch value := invocation.Arguments[0].(type) {
+				case *interpreter.ArrayValue:
+					isEmpty = value.Count() == 0
+				case *interpreter.DictionaryValue:
+					isEmpty = value.Count() == 0
+				default:
+					panic(errors.NewUnexpectedError("expected Array or Dictionary argument"))
```
turbolent, line 1461: "Unexpected" for errors means that an *implementation* error occurred, i.e. execution should have never reached this place, and if so, that indicates a bug in the Cadence implementation itself.

However, in this case here, it is possible that this place is reached, as the static function type is `AnyStruct`, and so any value other than an array or dictionary will reach this default case.

This is actually a *user* error, i.e. the user used the function in a wrong way, which indicates that there is a bug in the Cadence program being executed.

```suggestion
					panic(errors.NewDefaultUserError("expected Array or Dictionary argument"))
```

The other panics below should also be changed to user errors.

runtime/stdlib/contracts/test.cdc
```diff
@@ -132,9 +132,18 @@ pub contract Test {
         pub case failed
     }
 
+    /// Result is the interface to be implemented by the various execution
+    /// operations, such as transactions and scripts.
+    ///
+    pub struct interface Result {
```
turbolent, line 138: Nice! 👌

runtime/stdlib/test.go
m-Peter: Thanks for clarifying this part :pray: 
I have updated the panic errors in https://github.com/onflow/cadence/pull/2420/commits/3428082350091c7f1dd0caddd98689332ba05ab1.

runtime/stdlib/contracts/test.cdc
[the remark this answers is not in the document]
SupunS: oh I see. Makes sense. Thanks for trying that out. I think being able to remove the if-else statements is already a good solution 👌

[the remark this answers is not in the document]
m-Peter: Yes, indeed, that was quite a nice improvement in itself. Thanks for the suggestion :100:

turbolent (approved):
Great work! 👌

runtime/stdlib/test.go
```diff
@@ -1367,6 +1428,280 @@ var equalMatcherFunction = interpreter.NewUnmeteredHostFunctionValue(
...
+		return newMatcherWithGenericTestFunction(invocation, haveElementCountTestFunc)
+	},
+)
+
+const containMatcherFunctionName = "contain"
+
+const containMatcherFunctionDocString = `
+Returns a matcher that succeeds if the tested value is an array that contains
+a value that is equal to the given value, or the tested value is a dictionary
+that contains an entry where the value is equal to the given value.
```
SupunS, line 1531: For dictionaries, the method description says "..contains an entry where the value is equal to the given value", whereas the implementation checks for the existence of the key (i.e: `value.ContainsKey()` ). Maybe update the description to be consistent with the implementation.

```suggestion
that contains an entry where the key is equal to the given value.
```

SupunS (approved):

runtime/stdlib/test.go
m-Peter: Good catch :100: . This got me thinking, which functionality would be more valuable to developers, checking for a `key` or `value`, or both? What is your opinion about it?

SupunS: Checking for the existence of keys is the most common, I guess. So the current implementation makes sense to me.

If someone wants to check for a value, they can get the values as an array using `value.values` and then pass it to the matcher.

m-Peter: Perfect, thanks!

runtime/stdlib/test.go
```diff
@@ -1367,6 +1428,280 @@ var equalMatcherFunction = interpreter.NewUnmeteredHostFunctionValue(
...
+						inter,
+						invocation.LocationRange,
+						element,
+					)
+				case *interpreter.DictionaryValue:
+					elementFound = value.ContainsKey(
+						inter,
+						invocation.LocationRange,
+						element,
+					)
```
turbolent, line 1578: Maybe this should check both keys and values? Then again, e.g. [testify](https://github.com/stretchr/testify)'s `assert.Contains` also only checks map keys 🤔

Issue #2412, opened 2023-03-31
Entitlements


dsainati1:
### Issue to be solved

Implement the feature as proposed in this FLIP: https://github.com/onflow/flips/pull/54

dsainati1:
Remaining work on this:
* https://github.com/onflow/cadence/pull/2563 to merge Stable Cadence to the entitlements branch
* https://github.com/onflow/cadence/pull/2564 to add tests for entitlements interaction with interface inheritance
* https://github.com/onflow/cadence/pull/2567 to fix the runtime conversion for reference-containing types

Once these three are merged we should be good to merge the entitlements feature branch into Stable Cadence, and then close this issue.

PR #2421, opened 2023-04-04
Runtime behavior for entitlements

dsainati1:
Part of the implementation of https://github.com/onflow/flips/pull/54, part of https://github.com/onflow/cadence/issues/2412

Implements the runtime behavior for entitlements at runtime

______


runtime/interpreter/interpreter.go
```diff
@@ -4431,6 +4508,62 @@ func (interpreter *Interpreter) ReportComputation(compKind common.ComputationKin
...
+
+func (interpreter *Interpreter) mapMemberValueAuthorization(self Value, memberAccess *sema.Access, resultValue Value) Value {
+	if memberAccess == nil {
+		return resultValue
+	}
+	if mappedAccess, isMappedAccess := (*memberAccess).(sema.EntitlementMapAccess); isMappedAccess {
+		var auth EntitlementSetAuthorization
+		switch selfValue := self.(type) {
+		case AuthorizedValue:
+			imageAccess, err := mappedAccess.Image(interpreter.MustConvertStaticAuthorizationToSemaAccess(selfValue.GetAuthorization()), ast.EmptyRange)
```
SupunS, line 4535: could break this into two lines
```suggestion
			selfAccess := interpreter.MustConvertStaticAuthorizationToSemaAccess(selfValue.GetAuthorization())
			imageAccess, err := mappedAccess.Image(selfAccess, ast.EmptyRange)
```

SupunS (commented):
Reviewed up to tests. Will check the remaining in a second batch.

runtime/tests/checker/entitlements_test.go
```diff
@@ -128,9 +129,10 @@ func TestCheckBasicEntitlementMappingNonEntitlements(t *testing.T) {
 			}
 		`)
 
-		errs := RequireCheckerErrors(t, err, 1)
+		errs := RequireCheckerErrors(t, err, 2)
 
-		require.IsType(t, &sema.InvalidNonEntitlementTypeInMapError{}, errs[0])
+		require.IsType(t, &sema.NotDeclaredError{}, errs[0])
```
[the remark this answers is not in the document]
dsainati1: Basically because the entitlement mappings have to be processed by the checker before struct definitions (so they can be used in struct definitions), meaning that if someone puts a struct type inside an entitlement mapping (which doesn't work anyways) the checker doesn't know about that struct yet.

runtime/interpreter/statictype.go
```diff
@@ -511,41 +512,47 @@ func NewEntitlementSetAuthorization(memoryGauge common.MemoryGauge, entitlements
...
+			if !slices.Contains(auth.Entitlements, entitlement) {
 				return false
 			}
 		}
-		return e.Kind == auth.Kind
+		for _, entitlement := range auth.Entitlements {
+			if !slices.Contains(e.Entitlements, entitlement) {
+				return false
+			}
+		}
```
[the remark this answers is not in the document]
dsainati1: In the general case that's probably true, but I expect that these entitlement sets will be very small in practice (2-3 elements usually) that the overhead of a map is unlikely to be worth it.

runtime/interpreter/value.go
```diff
@@ -14962,12 +14963,6 @@ func (v *CompositeValue) Transfer(
 				// and does not need to be converted or copied
 
 				value := MustConvertStoredValue(interpreter, atreeValue)
-				// the base of an attachment is not stored in the atree, so in order to make the
-				// transfer happen properly, we set the base value here if this field is an attachment
-				if compositeValue, ok := value.(*CompositeValue); ok && compositeValue.Kind == common.CompositeKindAttachment {
-					compositeValue.setBaseValue(interpreter, v)
-				}
-
```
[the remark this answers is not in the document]
dsainati1: I don't think it ever was tbh, I realized during some debugging that the base value was always already set by the time this check was reached, so I just removed it.

runtime/interpreter/statictype.go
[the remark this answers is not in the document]
dsainati1: Discussed this in a meeting, will replace this slice with an ordered set.

SupunS (approved):

encoding/ccf/encode_type.go
```diff
@@ -318,6 +318,13 @@ func (e *Encoder) encodeReferenceType(
 	)
 }
 
+func (e *Encoder) encodeAuthorization(
+	auth cadence.Authorization,
+) error {
+	// TODO: implement this
```
turbolent, line 324: Is this implemented in a subsequent PR?

runtime/interpreter/statictype.go
```diff
@@ -605,8 +613,6 @@ func (e EntitlementMapAuthorization) Equal(auth Authorization) bool {
 // ReferenceStaticType
 
 type ReferenceStaticType struct {
-	// BorrowedType is the type of the usage (T in &T)
-	BorrowedType StaticType
 	// ReferencedType is type of the referenced value (the type of the target)
 	ReferencedType StaticType
```
turbolent, line 617: Why was the borrowed type removed? Isn't it still needed?

runtime/sema/access.go
```diff
@@ -296,7 +296,7 @@ func (e EntitlementMapAccess) Domain() EntitlementSetAccess {
 	for _, relation := range e.Type.Relations {
 		domain[relation.Input] = struct{}{}
 	}
-	e.domain = NewEntitlementSetAccess(maps.Keys(domain), Disjunction)
+	e.domain = NewEntitlementSetAccess(maps.Keys(domain), Conjunction)
```
turbolent, line 299: Why did this change?

runtime/storage_test.go
```diff
@@ -1556,9 +1556,7 @@ func TestRuntimeStorageReferenceCast(t *testing.T) {
 		},
 	)
 
-	RequireError(t, err)
-
-	require.ErrorAs(t, err, &interpreter.ForceCastTypeMismatchError{})
+	require.NoError(t, err)
```
turbolent, line 1559: Maybe add a test case that ensures that the type cannot be downcast with additional entitlements

runtime/interpreter/encoding_test.go
```diff
@@ -3536,9 +3531,9 @@ func TestEncodeDecodePathLinkValue(t *testing.T) {
 			// authorization
 			// tag
 			0xd8, CBORTagEntitlementSetStaticAuthorization,
-			// array, 2 items follow
+			// array, length 2
 			0x82,
-			// Disjunction
+			// conjunction
```
turbolent, line 3536: Accident?

runtime/tests/interpreter/runtimetype_test.go
```diff
@@ -505,24 +505,22 @@ func TestInterpretReferenceType(t *testing.T) {
       struct S {}
 	  entitlement X
 
-      let a = ReferenceType(authorized: true, type: Type<@R>())
-      let b = ReferenceType(authorized: false, type: Type<String>())
-      let c = ReferenceType(authorized: true, type: Type<S>())
+      let a = ReferenceType(entitlements: ["S.test.X"], type: Type<@R>())!
```
turbolent, line 508: Should the entitlements set maybe be `[Type]` instead of `[String]`?

encoding/ccf/encode_type.go
dsainati1: No, I spoke to @fxamacker about this and added this TODO (and created an issue) for her to track it; she wants to implement this herself in the future.

dsainati1: No, I spoke to @fxamacker about this and added this TODO (and created an issue) for her to track it; she wants to implement this herself in the future.

runtime/tests/interpreter/runtimetype_test.go
dsainati1: No, there is no runtime function to create a runtime entitlement type, and IMO there should not be. They aren't really types, and it would be confusing for you to be able to create a `Type` value for an entitlement (especially because they are "erased" at runtime).

runtime/interpreter/statictype.go
dsainati1: No, the "BorrowedType" was previously used to check what the "visible" type of the reference was, for things like casting where the type behavior of the reference was not necessarily the same as the type behavior of its underlying type. This has been simplified, a reference's borrow type is the same as the type of its underlying value at runtime (i.e. downcasting is possible), so we don't need this anymore.

turbolent: I see, that makes sense. My confusion probably stemmed from the fact that I assumed we would keep `BorrowedType`, because that was the actual "static type", and remove `ReferencedType` as that was the "dynamic type". 

It would be good to keep the comment though: `ReferencedType is the type T in &T`. The current comment, "referenced value (the type of the target)" meant that it was actually the concrete type of the value that the reference value was pointing to. That is not the case anymore (I assume?)

runtime/sema/access.go
turbolent: Sounds good, just wanted to make sure it was not an accident

runtime/interpreter/value.go
```diff
@@ -17374,6 +17435,15 @@ func (v *StorageReferenceValue) GetTypeKey(
 ) Value {
 	self := v.mustReferencedValue(interpreter, locationRange)
 
+	if selfComposite, isComposite := self.(*CompositeValue); isComposite {
```
turbolent, line 17438: Why is this special case for `*CompositeValue` needed?

runtime/interpreter/statictype.go
turbolent: Could we remove `ReferencedType` instead of `BorrowedType`? I have a bad feeling we currently use `BorrowedType` somewhere, initialize `ReferencedType` to a different value, and then the PR switches the use to `ReferencedType` and suddenly the semantics are wrong. It would reduce the number of changes in the PR

runtime/interpreter/interpreter_expression.go
```diff
@@ -1076,14 +1077,12 @@ func (interpreter *Interpreter) VisitCastingExpression(expression *ast.CastingEx
 			}
 
 			// The failable cast may upcast to an optional type, e.g. `1 as? Int?`, so box
-			value = interpreter.BoxOptional(locationRange, value, expectedType)
+			value = interpreter.ConvertAndBox(locationRange, value, valueSemaType, expectedType)
```
turbolent, line 1080: I guess this change is unrelated to entitlements? (vaguely remember we discussed this before?)

runtime/interpreter/interpreter_expression.go
```diff
@@ -1140,6 +1139,25 @@ func (interpreter *Interpreter) VisitReferenceExpression(referenceExpression *as
 
 	interpreter.maybeTrackReferencedResourceKindedValue(result)
 
+	makeReference := func(value Value, typ *sema.ReferenceType) *EphemeralReferenceValue {
```
turbolent, line 1142: Nice refactor!

runtime/sema/check_function.go
```diff
@@ -393,7 +393,7 @@ func (checker *Checker) visitWithPostConditions(postConditions *ast.Conditions,
 			// here the `result` value in the `post` block will have type `auth(E, X, Y) &R`
 			if entitlementSupportingType, ok := innerType.(EntitlementSupportingType); ok {
 				supportedEntitlements := entitlementSupportingType.SupportedEntitlements()
-				if supportedEntitlements == nil || supportedEntitlements.Len() > 0 {
+				if supportedEntitlements != nil && supportedEntitlements.Len() > 0 {
```
turbolent, line 396: Why did this change? Do we have a test case?

turbolent (commented):
Great work! Pretty close now 👏

runtime/interpreter/value.go
dsainati1: The implementation of `GetTypeKey` for `CompositeValue`s is written under the assumption that it is called with an owned value, and calls into `getTypeKey` with the full domain of the attachment's entitlement map as the attachment's access. This is not the behavior we want when we are accessing an attachment on a reference to a composite, so we compute the entitlement set of the reference and call `getTypeKey` with that access instead.

runtime/sema/check_function.go
dsainati1: This I think is left over from a previous PR; `SupportedEntitlements` used to return `nil` and now it cannot, so these two cases are actually equivalent. I just removed the first condition to simplify.

runtime/interpreter/interpreter_expression.go
dsainati1: It is related yes; the reason for this change is that we want casting an entitled reference to always "convert" the set of entitlements on the runtime value to the target type of the cast; i.e. it shouldn't be possible to "hide" an entitlement at runtime. It's the difference between this code having `downDownRef` be `nil` and non-`nil`:

```
let ref = &1 as auth(X) &Int
let anyStruct = ref as AnyStruct
let downRef = (anyStruct as? &Int)!
let downDownRef = downRef as? auth(X) &Int
```

runtime/interpreter/statictype.go
dsainati1: The comment as written is accurate; `ReferencedType` is the type of the referenced value. The type of the `T` that appears in the `&T` is no longer part of the static type, as it is not necessary, as we can now make all up and downcasting decisions based on the runtime type of the referenced value, and do not need to consider the type it was originally borrowed with. The exception to this are the entitlements of the reference, which are tracked separately. No other static types keep a converted copy of the original sema type around; references only did this because previously the runtime value was only allowed to be used in ways that reflected its original borrow type. 

The only places in the code where this actually comes up are in the interpreter encoding (which used to encode the `BorrowedType` and now uses the `ReferencedType`), in `IsSubtypeOfSemaType` in the interpreter (which was changed before this field was removed, and was the impetus for removing it in the first place), and in the creation of sema types from static types for references (which is also intentional).

turbolent: Thank you for the great explanation! 👍

turbolent (approved):
👏 🚀

PR #2425, opened 2023-04-05
Add mapping function for types

dsainati1:
Required for https://github.com/onflow/cadence/pull/2421

This implements a function that maps type annotations according to an input function Type -> Type
______


runtime/sema/type.go
```diff
@@ -6295,6 +6392,10 @@ func (t *RestrictedType) RewriteWithRestrictedTypes() (Type, bool) {
 	return t, false
 }
 
+func (t *RestrictedType) Map(gauge common.MemoryGauge, f func(Type) Type) Type {
+	return f(NewRestrictedType(gauge, t.Type.Map(gauge, f), t.Restrictions))
```
SupunS, line 6396: Might want to apply the mapping to the `t.Restrictions` as well

runtime/sema/type.go
```diff
@@ -6676,6 +6777,10 @@ const capabilityTypeAddressFieldDocString = `
 The address of the capability
 `
 
+func (t *CapabilityType) Map(gauge common.MemoryGauge, f func(Type) Type) Type {
+	return f(NewCapabilityType(gauge, t.BorrowType.Map(gauge, f)))
```
SupunS, line 6781: Would need to handle the case where `t.BorrowType == nil`

runtime/sema/type.go
```diff
@@ -859,6 +872,10 @@ func (t *GenericType) Resolve(typeArguments *TypeParameterTypeOrderedMap) Type {
 	return ty
 }
 
+func (t *GenericType) Map(_ common.MemoryGauge, f func(Type) Type) Type {
+	return f(t)
```
SupunS, line 876: Don't we need to create a new Generic type here?
Also, for completion, might want want to map `GenericType.TypeParameter.TypeBound`?

SupunS (approved):

runtime/sema/type.go
```diff
@@ -859,6 +872,18 @@ func (t *GenericType) Resolve(typeArguments *TypeParameterTypeOrderedMap) Type {
 	return ty
 }
 
+func (t *GenericType) Map(gauge common.MemoryGauge, f func(Type) Type) Type {
+	typeParameter := &TypeParameter{
+		Name:      t.TypeParameter.Name,
+		Optional:  t.TypeParameter.Optional,
+		TypeBound: t.TypeParameter.TypeBound.Map(gauge, f),
+	}
```
turbolent, line 880: I'm not quite sure if returning a new/different type parameter here is not going to work: function types refer to their type parameters and generic type occurrences in the type (e.g. parameter's type) refer to them by pointer.

runtime/sema/type.go
```diff
@@ -3180,6 +3221,42 @@ func (t *FunctionType) Resolve(typeArguments *TypeParameterTypeOrderedMap) Type
...
+	for _, parameter := range t.TypeParameters {
+		newTypeParameterTypeBound := parameter.TypeBound.Map(gauge, f)
+
+		newTypeParameters = append(
+			newTypeParameters,
+			&TypeParameter{
+				Name:      parameter.Name,
+				Optional:  parameter.Optional,
+				TypeBound: newTypeParameterTypeBound,
+			},
```
turbolent, line 3251: These new type parameters, or actually, the mapping of old to new type parameter, should probably be passed along in the recursive map calls, so that generic type occurrences can be "fixed up"

turbolent (commented):
Nice! It's great how small and self-contained this PR is! 👌

runtime/sema/type.go
```diff
@@ -5192,6 +5302,10 @@ func (t *ReferenceType) RewriteWithRestrictedTypes() (Type, bool) {
 	}
 }
 
+func (t *ReferenceType) Map(gauge common.MemoryGauge, typeParamMap map[*TypeParameter]*TypeParameter, f func(Type) Type) Type {
+	return f(NewReferenceType(gauge, t.Type.Map(gauge, typeParamMap, f), t.Authorization))
```
turbolent, line 5306: There's a lot of nesting going on here, maybe break this down a bit to make it easier to read:
```suggestion
	mappedType := t.Type.Map(gauge, typeParamMap, f)
	return f(NewReferenceType(
		gauge, 
		mappedType, 
		t.Authorization,
	))
```

runtime/sema/type.go
```diff
@@ -6317,6 +6439,17 @@ func (t *RestrictedType) RewriteWithRestrictedTypes() (Type, bool) {
 	return t, false
 }
 
+func (t *RestrictedType) Map(gauge common.MemoryGauge, typeParamMap map[*TypeParameter]*TypeParameter, f func(Type) Type) Type {
+	restrictions := make([]*InterfaceType, 0, len(t.Restrictions))
+	for _, restriction := range t.Restrictions {
+		if mappedRestriction, isRestriction := restriction.Map(gauge, typeParamMap, f).(*InterfaceType); isRestriction {
+			restrictions = append(restrictions, mappedRestriction)
+		}
```
turbolent, line 6447: `RestrictedType.Restrictions` is `[]*InterfaceType`, so `restriction` is `*InterfaceType`, and therefore `Map` should always return the same type:

```suggestion
		mappedRestriction, ok := restriction.Map(gauge, typeParamMap, f).(*InterfaceType)
		if !ok {
			panic(errors.NewUnreachableError())
		}
		restrictions = append(restrictions, mappedRestriction)
```

(It would be great if we could express this in Go directly, but its generic system has no notion of a "self type")

runtime/sema/type.go
```diff
@@ -6317,6 +6439,17 @@ func (t *RestrictedType) RewriteWithRestrictedTypes() (Type, bool) {
...
+	}
+
+	return f(NewRestrictedType(gauge, t.Type.Map(gauge, typeParamMap, f), restrictions))
```
turbolent, line 6450:
```suggestion
	mappedType := t.Type.Map(gauge, typeParamMap, f)

	return f(NewRestrictedType(
		gauge, 
		mappedType, 
		restrictions,
	))
```

runtime/sema/type.go
```diff
@@ -3180,6 +3229,51 @@ func (t *FunctionType) Resolve(typeArguments *TypeParameterTypeOrderedMap) Type
...
+		}
+		typeParamMap[parameter] = newParam
+
+		newTypeParameters = append(
+			newTypeParameters,
+			newParam,
+		)
+	}
+
+	var newParameters []Parameter = make([]Parameter, 0, len(t.Parameters))
```
turbolent, line 3256:
```suggestion
	newParameters := make([]Parameter, 0, len(t.Parameters))
```

runtime/sema/type.go
turbolent: Ah, I see. It might be better to just panic in the case where the requirement that restrictions need to be interfaces does not hold, instead of just silently ignoring the result of the mapping function

runtime/sema/type.go
```diff
@@ -3146,7 +3184,7 @@ func (t *FunctionType) Resolve(typeArguments *TypeParameterTypeOrderedMap) Type
 
 	// parameters
 
-	var newParameters []Parameter
+	var newParameters []Parameter = make([]Parameter, 0, len(t.Parameters))
```
turbolent, line 3187: We shouldn't unnecessarily allocate an empty slice here if there are no parameters (`len(t.Parameters)`).

Maybe change the code here and elsewhere in the PR where `make` is called to the pattern we use elsewhere:
- Define the slice variable (causes no allocation)
- Check the length of the input. If positive:
  -  Pre-allocate
  - Perform operation

For example: https://github.com/onflow/cadence/blob/2081a601106baaf6ae695e3f2a84613160bb2166/runtime/convertTypes.go#L411-L420

turbolent (approved):
Nice! 👏

Issue #2276, opened 2023-01-25
Extend ordering comparisons (< , <=, etc) to support container types


dreamsmasher:
### Issue to be solved

Currently, we only support ordering comparisons for numeric types. This could be extended to other built-in types like strings, booleans, and containers whose inner types are comparable. This is particularly useful for common operations like sorting arrays of strings, binary search, and implementing dynamic lookup tables.

### Suggested Solution

While equality for values is defined by the `Equatable` interface in the interpreter, The comparison operators `Less`, `LessEqual`, `Greater`, and `GreaterEqual` are defined as part of the [`NumberValue` interface](https://github.com/onflow/cadence/blob/0354a413c16f3844b6f0bff42292a00f3121ab62/runtime/interpreter/value.go#L2691). Opening up more Cadence types for comparison would require these ops to be factored out into a new interface, `Orderable`, from which `NumberValue` can inherit.

As part of an MVP, ordering should be defined for the following types:
- `Bool`: `false < true`
- `[T]` if T is orderable: lexicographic ordering, e.g. `[1, 2] < [1, 2, 3]`, `[1] < [2]`
- `[T; N]` if T is orderable: lexicographic ordering. Both the left and right-hand expressions in the comparison must have the same lengths, otherwise they're different types.
- `String`: lexicographic ordering according to grapheme values, e.g. `"OwO" < "UwU"`, `"😎" < "🥺"`
- `T?` if T is orderable: lexicographic ordering. Let the expression `Some(x: T)` denote that `x` is not `nil`:
  - `nil == nil`
  - `nil < Some(x)`
  - `Some(x) < Some(y)` = `x < y`

Documentation and tests should be updated accordingly to reflect the new behaviour.

## Open questions
- Should `Void` support comparisons?
  - We already allow `() == ()`, which is true.
  - Allowing the other comparison ops, while not particularly useful, would be logically consistent with the other changes this issue proposes.

- Should comparing `Optional`s be polymorphic?
  - Equality comparisons for `Optional` values of different nesting levels are already allowed as long as the base type is the same, e.g. 
  - ```cadence
      let x: Int?? = 1
      let y: Int? = 1
      let z: Int = 1
      x == y // true
      x == z // true

      let w: Int????????? = nil
      w == x // false
    ```
  - This is consistent with the behaviour of `Optional` values in Swift.
  - However, Swift doesn't allow for optionals to be ordered.
[...]

PR #2430, opened 2023-04-09
Extend ordering comparisons (< , <=, etc) to Bool & Character

darkdrag00nv2:
Work towards #2276 

## Description
This PR does the following:
- Extracts comparison related interface methods to a new interface `ComparableValue`
- Adds `IsComparable` to the type interface
- Updates the checker to use the `IsComparable` method to detemine comparability
- Implements the comparison support for `Bool` and `Character`

Support for  `String` can be done as a follow up.

TODOs before this is ready for review:
- [x] Fix tests for numbers
- [x] Add tests for bool & character
- [x] Update documentation
______

### DIFF
--- a/runtime/interpreter/interpreter_expression.go
+++ b/runtime/interpreter/interpreter_expression.go
@@ -383,37 +383,11 @@ func (interpreter *Interpreter) VisitBinaryExpression(expression *ast.BinaryExpr
 		}
 		return left.BitwiseRightShift(interpreter, right, locationRange)
 
-	case ast.OperationLess:
-		left, leftOk := leftValue.(NumberValue)
-		right, rightOk := rightValue().(NumberValue)
-		if !leftOk || !rightOk {
-			error(right)
-		}
-		return left.Less(interpreter, right, locationRange)
-
-	case ast.OperationLessEqual:
-		left, leftOk := leftValue.(NumberValue)
-		right, rightOk := rightValue().(NumberValue)
-		if !leftOk || !rightOk {
-			error(right)
-		}
-		return left.LessEqual(interpreter, right, locationRange)
-
-	case ast.OperationGreater:
-		left, leftOk := leftValue.(NumberValue)
-		right, rightOk := rightValue().(NumberValue)
-		if !leftOk || !rightOk {
-			error(right)
-		}
-		return left.Greater(interpreter, right, locationRange)
-
-	case ast.OperationGreaterEqual:
-		left, leftOk := leftValue.(NumberValue)
-		right, rightOk := rightValue().(NumberValue)
-		if !leftOk || !rightOk {
-			error(right)
-		}
-		return left.GreaterEqual(interpreter, right, locationRange)
+	case ast.OperationLess,
+		ast.OperationLessEqual,
+		ast.OperationGreater,
+		ast.OperationGreaterEqual:
+		return interpreter.testComparison(leftValue, rightValue(), expression)
 
 	case ast.OperationEqual:
 		return interpreter.testEqual(leftValue, rightValue(), expression)
@@ -521,6 +495,83 @@ func (interpreter *Interpreter) testEqual(left, right Value, expression *ast.Bin
 			right,
 		),
 	)
+}
+
+func (interpreter *Interpreter) testComparison(left, right Value, expression *ast.BinaryExpression) BoolValue {
+	left = interpreter.Unbox(
+		LocationRange{
+			Location:    interpreter.Location,
+			HasPosition: expression.Left,
+		},
+		left,
+	)
+
+	right = interpreter.Unbox(
+		LocationRange{
+			Location:    interpreter.Location,
+			HasPosition: expression.Right,
+		},
+		right,
+	)
+
+	leftComparable, ok := left.(ComparableValue)
+	if !ok {
+		return FalseValue
+	}
+
+	rightComparable, ok := right.(ComparableValue)
+	if !ok {
+		return FalseValue
+	}
+
+	switch expression.Operation {
+	case ast.OperationLess:
+		return leftComparable.Less(
+			interpreter,
+			rightComparable,
+			LocationRange{
+				Location:    interpreter.Location,
+				HasPosition: expression,
+			},
+		)
+
+	case ast.OperationLessEqual:
+		return leftComparable.LessEqual(
+			interpreter,
+			rightComparable,
+			LocationRange{
+				Location:    interpreter.Location,
+				HasPosition: expression,
+			},
+		)
+
+	case ast.OperationGreater:
+		return leftComparable.Greater(
+			interpreter,
+			rightComparable,
+			LocationRange{
+				Location:    interpreter.Location,
+				HasPosition: expression,
+			},
+		)
+
+	case ast.OperationGreaterEqual:
+		return leftComparable.GreaterEqual(
+			interpreter,
+			rightComparable,
+			LocationRange{
+				Location:    interpreter.Location,
+				HasPosition: expression,
+			},
+		)
+
+	default:
+		panic(&unsupportedOperation{
+			kind:      common.OperationKindBinary,
+			operation: expression.Operation,
+			Range:     ast.NewUnmeteredRangeFromPositioned(expression),
+		})
+	}
 }
 
 func (interpreter *Interpreter) VisitUnaryExpression(expression *ast.UnaryExpression) Value {
--- a/runtime/interpreter/interpreter_expression.go
+++ b/runtime/interpreter/interpreter_expression.go
@@ -498,22 +498,6 @@ func (interpreter *Interpreter) testEqual(left, right Value, expression *ast.Bin
 }
 
 func (interpreter *Interpreter) testComparison(left, right Value, expression *ast.BinaryExpression) BoolValue {
-	left = interpreter.Unbox(
-		LocationRange{
-			Location:    interpreter.Location,
-			HasPosition: expression.Left,
-		},
-		left,
-	)
-
-	right = interpreter.Unbox(
-		LocationRange{
-			Location:    interpreter.Location,
-			HasPosition: expression.Right,
-		},
-		right,
-	)
-
 	locationRange := LocationRange{
 		Location:    interpreter.Location,
 		HasPosition: expression,
--- a/runtime/interpreter/interpreter_expression.go
+++ b/runtime/interpreter/interpreter_expression.go
@@ -514,19 +514,21 @@ func (interpreter *Interpreter) testComparison(left, right Value, expression *as
 		right,
 	)
 
-	leftComparable, ok := left.(ComparableValue)
-	if !ok {
-		return FalseValue
-	}
-
-	rightComparable, ok := right.(ComparableValue)
-	if !ok {
-		return FalseValue
-	}
-
 	locationRange := LocationRange{
 		Location:    interpreter.Location,
 		HasPosition: expression,
+	}
+
+	leftComparable, leftOk := left.(ComparableValue)
+	rightComparable, rightOk := right.(ComparableValue)
+
+	if !leftOk || !rightOk {
+		panic(InvalidOperandsError{
+			Operation:     expression.Operation,
+			LeftType:      left.StaticType(interpreter),
+			RightType:     right.StaticType(interpreter),
+			LocationRange: locationRange,
+		})
 	}
 
 	switch expression.Operation {
--- a/runtime/interpreter/interpreter_expression.go
+++ b/runtime/interpreter/interpreter_expression.go
@@ -524,45 +524,38 @@ func (interpreter *Interpreter) testComparison(left, right Value, expression *as
 		return FalseValue
 	}
 
+	locationRange := LocationRange{
+		Location:    interpreter.Location,
+		HasPosition: expression,
+	}
+
 	switch expression.Operation {
 	case ast.OperationLess:
 		return leftComparable.Less(
 			interpreter,
 			rightComparable,
-			LocationRange{
-				Location:    interpreter.Location,
-				HasPosition: expression,
-			},
+			locationRange,
 		)
 
 	case ast.OperationLessEqual:
 		return leftComparable.LessEqual(
 			interpreter,
 			rightComparable,
-			LocationRange{
-				Location:    interpreter.Location,
-				HasPosition: expression,
-			},
+			locationRange,
 		)
 
 	case ast.OperationGreater:
 		return leftComparable.Greater(
 			interpreter,
 			rightComparable,
-			LocationRange{
-				Location:    interpreter.Location,
-				HasPosition: expression,
-			},
+			locationRange,
 		)
 
 	case ast.OperationGreaterEqual:
 		return leftComparable.GreaterEqual(
 			interpreter,
 			rightComparable,
-			LocationRange{
-				Location:    interpreter.Location,
-				HasPosition: expression,
-			},
+			locationRange,
 		)
 
 	default:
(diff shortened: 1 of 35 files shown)


bluesign:
does comparable make sense for Bool ? I don't have much CS background but kind of makes no sense to me.

darkdrag00nv2:
@bluesign Yeah, I am not 100% on its usefulness. However, Kotlin does support comparison for booleans. See the discussion on the issue #2276.

bluesign:
I think it is a bit problematic; ( more headache ) 

```
a>true -> meaningless  ( prone to mistake ) 
a<false -> meaningless ( prone to mistake ) 
a>=true -> a==true
a<=false -> a==false 
a>false  -> a==true
a<true  -> a==false 
```

basically anything we compare with a constant, should be handled by `==`

Some can say maybe it can be useful for sorting array of anystruct or something, but I think gain is minimal there. ( as we will not be able to compare false to 42 for example )

darkdrag00nv2:
@turbolent Thoughts on the above comment?

runtime/interpreter/interpreter_expression.go
```diff
@@ -523,6 +497,83 @@ func (interpreter *Interpreter) testEqual(left, right Value, expression *ast.Bin
...
+		left,
+	)
+
+	right = interpreter.Unbox(
+		LocationRange{
+			Location:    interpreter.Location,
+			HasPosition: expression.Right,
+		},
+		right,
+	)
```
turbolent, line 515: As mentioned in https://github.com/onflow/cadence/issues/2276#issuecomment-1496280606, we should probably keep the current behaviour of not automatically unboxing for non-equality comparisons

runtime/interpreter/interpreter_expression.go
```diff
@@ -523,6 +497,83 @@ func (interpreter *Interpreter) testEqual(left, right Value, expression *ast.Bin
...
+
+	leftComparable, ok := left.(ComparableValue)
+	if !ok {
+		return FalseValue
```
turbolent, line 519: This was previously a panic with `InvalidOperandsError`, see the local `error` function in the calling function: https://github.com/onflow/cadence/pull/2430/files#diff-484f7e68f1a1b9f96531632abacc84c131416a281cb29a502ac3b98c041ddb78R296-R303

runtime/interpreter/interpreter_expression.go
```diff
@@ -523,6 +497,83 @@ func (interpreter *Interpreter) testEqual(left, right Value, expression *ast.Bin
...
+	}
+
+	rightComparable, ok := right.(ComparableValue)
+	if !ok {
+		return FalseValue
```
turbolent, line 524: Same as above, should probably stay an error

runtime/interpreter/interpreter_expression.go
```diff
@@ -523,6 +497,83 @@ func (interpreter *Interpreter) testEqual(left, right Value, expression *ast.Bin
...
+
+	switch expression.Operation {
+	case ast.OperationLess:
+		return leftComparable.Less(
+			interpreter,
+			rightComparable,
+			LocationRange{
+				Location:    interpreter.Location,
+				HasPosition: expression,
+			},
```
turbolent, line 535: Maybe refactor the common location range construction here and in the other cases into a variable declaration before the switch

runtime/interpreter/value.go
```diff
@@ -186,6 +186,15 @@ func newValueComparator(interpreter *Interpreter, locationRange LocationRange) a
 	}
 }
 
+// ComparableValue
+type ComparableValue interface {
+	Value
```
turbolent, line 191: Maybe not necessary, but might prevent some invalid implementations:

```suggestion
	EquatableValue
```

runtime/interpreter/value.go
```diff
@@ -2690,6 +2800,7 @@ func (v *ArrayValue) Slice(
 // NumberValue
 type NumberValue interface {
 	EquatableValue
```
turbolent, line 2802: Can be removed once `ComparableValue` requires `EquatableValue`

runtime/sema/check_binary_expression.go
```diff
@@ -277,8 +282,7 @@ func (checker *Checker) checkBinaryExpressionArithmeticOrNonEqualityComparisonOr
 			return true
 		}
 
-		// Arithmetic, bitwise and non-equality comparison operators
-		// are not supported for numeric supertypes.
+		// Arithmetic, bitwise are not supported for numeric supertypes.
```
turbolent, line 285:
```suggestion
		// Arithmetic and bitwise operations are not supported for numeric supertypes.
```

runtime/sema/check_binary_expression.go
```diff
@@ -293,18 +297,31 @@ func (checker *Checker) checkBinaryExpressionArithmeticOrNonEqualityComparisonOr
...
+			checker.report(
+				&InvalidBinaryOperandsError{
+					Operation: operation,
+					LeftType:  leftType,
+					RightType: rightType,
+					Range:     ast.NewRangeFromPositioned(checker.memoryGauge, expression),
+				},
+			)
+		}
 	}
```
turbolent, line 322: Simplify by unnesting

```suggestion
	if !(leftType.Equal(rightType) && leftType.IsComparable() && rightType.IsComparable()) && !anyInvalid {
		checker.report(
			&InvalidBinaryOperandsError{
				Operation: operation,
				LeftType:  leftType,
				RightType: rightType,
				Range:     ast.NewRangeFromPositioned(checker.memoryGauge, expression),
			},
		)
	}
```

turbolent (commented):
Nice!

runtime/interpreter/interpreter_expression.go
darkdrag00nv2: Done.

runtime/interpreter/interpreter_expression.go
darkdrag00nv2: Thanks for the pointer. Updated the logic.

runtime/interpreter/interpreter_expression.go
darkdrag00nv2: Ok, I've removed the unboxing logic. Please let me know if there is anything else that needs to be done for this.

runtime/interpreter/value.go
```diff
@@ -639,6 +648,74 @@ func (v BoolValue) Equal(_ *Interpreter, _ LocationRange, other Value) bool {
...
+			LeftType:  v.StaticType(interpreter),
+			RightType: other.StaticType(interpreter),
+		})
+	}
+
+	if v {
+		return false
+	} else {
+		return o
+	}
```
dsainati1, line 665:
```suggestion
	return !v && o
```

runtime/interpreter/value.go
```diff
@@ -639,6 +648,74 @@ func (v BoolValue) Equal(_ *Interpreter, _ LocationRange, other Value) bool {
...
+			LeftType:  v.StaticType(interpreter),
+			RightType: other.StaticType(interpreter),
+		})
+	}
+
+	if v {
+		return o
+	} else {
+		return true
+	}
```
dsainati1, line 682:
```suggestion
	return !v || o
```

runtime/interpreter/value.go
```diff
@@ -639,6 +648,74 @@ func (v BoolValue) Equal(_ *Interpreter, _ LocationRange, other Value) bool {
...
+			LeftType:  v.StaticType(interpreter),
+			RightType: other.StaticType(interpreter),
+		})
+	}
+
+	if v {
+		return !o
+	} else {
+		return false
+	}
```
dsainati1, line 699:
```suggestion
	return v && !o
```

runtime/interpreter/value.go
```diff
@@ -639,6 +648,74 @@ func (v BoolValue) Equal(_ *Interpreter, _ LocationRange, other Value) bool {
...
+			LeftType:  v.StaticType(interpreter),
+			RightType: other.StaticType(interpreter),
+		})
+	}
+
+	if v {
+		return true
+	} else {
+		return !o
+	}
```
dsainati1, line 716:
```suggestion
	return v || !o
```

runtime/interpreter/value.go
```diff
@@ -795,6 +873,38 @@ func (v CharacterValue) Equal(_ *Interpreter, _ LocationRange, other Value) bool
 	return v.NormalForm() == otherChar.NormalForm()
 }
 
+func (v CharacterValue) Less(interpreter *Interpreter, other ComparableValue, locationRange LocationRange) BoolValue {
+	otherChar, ok := other.(CharacterValue)
+	if !ok {
+		return false
+	}
```
dsainati1, line 880: Behavior here is inconsistent with `BoolValue` and `IntValue`; if the operands are not both `CharacterValue`s this should produce an `InvalidOperandsError`. Same goes for all the other comparison functions on this type.

runtime/sema/check_binary_expression.go
```diff
@@ -293,18 +297,29 @@ func (checker *Checker) checkBinaryExpressionArithmeticOrNonEqualityComparisonOr
...
+	expression *ast.BinaryExpression,
+	operation ast.Operation,
+	leftType, rightType Type,
+	anyInvalid bool,
+) (resultType Type) {
+	resultType = BoolType
 
-	default:
-		panic(errors.NewUnreachableError())
+	if !(leftType.Equal(rightType) && leftType.IsComparable() && rightType.IsComparable()) && !anyInvalid {
```
dsainati1, line 311: Does the requirement that `leftType.Equal(rightType)` here not prevent users from comparing, say, `Int16` values with `Int8` values? This seems like a valuable ability to have.

runtime/interpreter/value.go
```diff
@@ -639,6 +648,74 @@ func (v BoolValue) Equal(_ *Interpreter, _ LocationRange, other Value) bool {
...
 }
 
+func (v BoolValue) Less(interpreter *Interpreter, other ComparableValue, locationRange LocationRange) BoolValue {
+	o, ok := other.(BoolValue)
+	if !ok {
+		panic(InvalidOperandsError{
+			Operation: ast.OperationLess,
+			LeftType:  v.StaticType(interpreter),
+			RightType: other.StaticType(interpreter),
+		})
```
turbolent, line 658: Here and below in the other new comparison functions for `BoolValue`:

For numbers, we statically allow comparisons of different concrete number types, which is why there are run-time checks in the comparison functions.

However, comparisons of booleans and other types should be statically rejected, and this check here should be an implementation (unexpected) error instead.

runtime/interpreter/value.go
```diff
@@ -639,6 +648,74 @@ func (v BoolValue) Equal(_ *Interpreter, _ LocationRange, other Value) bool {
...
+	}
+
+	if v {
```
turbolent, line 661: As you mentioned in the description/TODOs, the PR still needs interpreter test cases

runtime/sema/check_binary_expression.go
turbolent: Yes, comparisons, just like e.g arithmetic and bitwise operations do not implicitly convert value types, as the rules are often surprising and lead to bugs.

For example, in C/C++, when comparing signed with unsigned, it is important the signed value is positive to avoid unexpected conversions and results.

turbolent: Yes, there are conversion functions to convert between numeric types

runtime/interpreter/value.go
darkdrag00nv2: Thank you. For some reason I didn't think of this :)

runtime/interpreter/value.go
darkdrag00nv2: Added them in `TestInterpretComparison` in `runtime/tests/interpreter/interpreter_test.go`.

runtime/interpreter/value.go
darkdrag00nv2: Thanks, I've updated the code to `panic(errors.NewUnreachableError())`.

runtime/interpreter/value.go
darkdrag00nv2: Makes sense. Updated it to `panic(errors.NewUnreachableError())`.

darkdrag00nv2:
Update on the PR:
- Resolved the review suggestions
- Added interpreter tests for char and bool
- Included upper and lowercase character comparison example in the documentation

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -4681,6 +4681,73 @@ func TestInterpretDictionaryEquality(t *testing.T) {
...
+		{"a >= a", true, "let left: Character = \"a\";\nlet right: Character = \"a\"; return left >= right"},
+		{"A >= a", false, "let left: Character = \"A\";\nlet right: Character = \"a\"; return left >= right"},
+	}
+
+	for _, test := range tests {
+		test := test // capture range variable
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+			runBooleanTest(t, test.name, test.expected, test.inner)
+		})
```
turbolent, line 4747:
```suggestion
		runBooleanTest(t, test.name, test.expected, test.inner)
```

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -4681,6 +4681,69 @@ func TestInterpretDictionaryEquality(t *testing.T) {
...
+
+			inter := parseCheckAndInterpret(t, code)
+			res, err := inter.Invoke("test")
+
+			require.NoError(t, err)
+
+			boolVal, ok := res.(interpreter.BoolValue)
+			require.True(t, ok)
+
+			require.Equal(t, bool(boolVal), expected)
```
turbolent, line 4701:
```suggestion
			require.Equal(t, expected, bool(boolVal))
```

turbolent (approved):
Nice!

turbolent:
@SupunS Could you also please have a look when you have time? Thanks!

darkdrag00nv2:
@SupunS Gentle reminder :slightly_smiling_face:

turbolent:
@dsainati1 Could you also please review the latest changes?

runtime/sema/check_binary_expression.go
SupunS, line 311: Maybe extract this condition to a util function (with early-exit pattern), just to make it a bit more readable..

runtime/sema/gen/testdata/comparable.golden.go
```diff
@@ -0,0 +1,35 @@
...
+	QualifiedName: TestTypeName,
+	TypeID:        TestTypeName,
+	tag:           TestTypeTag,
+	IsResource:    false,
+	Storable:      false,
+	Equatable:     false,
+	Comparable:    true,
+	Exportable:    false,
+	Importable:    false,
+}
```
SupunS, line 35: Is this generated test type being asserted anywhere? (i.e: to see if the generated type has the 'Comparable' field set to `true`)

SupunS (approved):
Looks good! Just one question and a nitpick.

Sorry for taking so long for the review! Appreciate the patience 🙏

runtime/sema/gen/testdata/comparable.golden.go
turbolent: The `.golden.go` files are actually the expectations for the respective `.cdc` test cases.

dsainati1 (approved):

runtime/sema/gen/testdata/comparable.golden.go
SupunS: ah ok, got it!

runtime/sema/check_binary_expression.go
turbolent: Refactored into a variable in 213c48038b8d24e076f22de3746a452f286c4a7e to improve readability

PR #2440, opened 2023-04-13
Attachment Iteration

dsainati1:
Implements iteration on attachments 
______


runtime/interpreter/interpreter.go
```diff
@@ -4534,7 +4534,7 @@ func (interpreter *Interpreter) mapMemberValueAuthorization(self Value, memberAc
 }
 
 func (interpreter *Interpreter) getMemberWithAuthMapping(self Value, locationRange LocationRange, identifier string) Value {
-	result := interpreter.getMember(self, locationRange, identifier)
+	result := interpreter.getMember(self, locationRange, identifier, nil)
```
SupunS, line 4537: Should we pass `UnauthorizedAccess` here instead of `nil`?

runtime/interpreter/value.go
```diff
@@ -15157,6 +15161,51 @@ func (v *CompositeValue) GetAttachments(interpreter *Interpreter, locationRange
 	return attachments
 }
 
+func (v *CompositeValue) forEachAttachmentFunction(interpreter *Interpreter, baseAuthorization Authorization, locationRange LocationRange) Value {
```
SupunS, line 15164: `baseAuthorization` seems to be unused here. Is it needed for any follow-up PRs?

dsainati1: Nope, that's a leftover from before when entitlements had runtime behavior. Wish Go's unused variable lints also applied to named function parameters.

runtime/interpreter/interpreter.go
dsainati1: This argument isn't actually needed anymore (see comment below), so I just removed it.

SupunS (approved):

runtime/interpreter/value.go
```diff
@@ -15304,6 +15308,51 @@ func (v *CompositeValue) GetAttachments(interpreter *Interpreter, locationRange
...
+				attachmentReference := NewEphemeralReferenceValue(
+					interpreter,
+					attachmentReferenceAuth,
+					attachment,
+					attachmentType,
+				)
+
+				invocation := NewInvocation(
+					interpreter,
+					nil,
```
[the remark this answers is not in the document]
dsainati1: No, we don't want `self` to be visible in the body of this function unless it's already visible. I.e. if you call this outside of a composite, we shouldn't make `self` available inside this invocation.

runtime/sema/type.go
```diff
@@ -4205,6 +4205,45 @@ func (t *CompositeType) IsValidIndexingType(ty Type) bool {
...
+									},
+								),
+							},
+						},
+						ReturnTypeAnnotation: NewTypeAnnotation(VoidType),
+					},
+				),
+			},
+		},
+		ReturnTypeAnnotation: NewTypeAnnotation(VoidType),
```
turbolent, line 4243:
```suggestion
						ReturnTypeAnnotation: VoidTypeAnnotation,
					},
				),
			},
		},
		ReturnTypeAnnotation: VoidTypeAnnotation,
```

runtime/tests/interpreter/attachments_test.go
```diff
@@ -1947,3 +1947,397 @@ func TestInterpretAttachmentDefensiveCheck(t *testing.T) {
...
+            attachment B for S {
+                fun foo(): Int { return 10 }
+            }
+            attachment C for S {
+                fun foo(_ x: Int): Int { return 8 + x }
+            }
+            fun test(): Int {
+                var s = attach C() to attach B() to attach A() to S()
+                var i = 0
+                s.forEachAttachment(fun(attachment: &AnyStructAttachment) {
```
turbolent, line 2024: Nice!

runtime/interpreter/value.go
```diff
@@ -15304,6 +15308,51 @@ func (v *CompositeValue) GetAttachments(interpreter *Interpreter, locationRange
...
+					[]Value{attachmentReference},
+					[]sema.Type{sema.NewReferenceType(interpreter, attachmentType, sema.UnauthorizedAccess)},
+					nil,
+					locationRange,
+				)
+				functionValue.invoke(invocation)
+			}
+
+			v.forEachAttachment(interpreter, locationRange, fn)
```
turbolent, line 15350: Do we also need mutation prevention during iteration here?

turbolent (commented):
Nice!

runtime/interpreter/value.go
dsainati1: Yeah, the `forEachAttachment` function handles the state here, and then during adding and removing of attachments we check for the state (see `VisitAttachExpression` and `VisitRemoveStatement`) respectively.

turbolent (approved):
Great to see that we can finally enable this feature! 👏 🎉

PR #2444, opened 2023-04-13
Fix built-in `result` variable for optional resource typed returns

### LINKED ISSUE #2443: Built-in `result` variable has incorrect type when the return type is an optional-resource
### Current Behavior

If the return type of a function is an optional-resource (say `T?`), then the built-in `result` variable has the type `&(T?)`. This makes it impossible to force-inwrap or nil-coalesce on the `result` variable

```cadence
pub fun main(): @Foo?  {
  post {
    result!.id == nil: "invalid id"  // Error: "value of type `&Foo?` has no member `id`. unknown member"
  }
  return <- create Foo()
}

pub resource Foo {
  pub let id: Int
  init() {
    self.id = 1
  }
}
```

### Expected Behavior

In the above case, the type of the variable should be an optional-reference `(&T)?`, instead of a reference to an optional `&(T?)`

### Steps To Reproduce

See above.

### Environment

```markdown
- Cadence version: latest/master
```

SupunS:
Closes #2443

## Description

Make the `result` variable an optional-reference, instead of a reference to an optional.

______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -755,7 +755,25 @@ func (interpreter *Interpreter) visitFunctionBody(
 	if returnType != sema.VoidType {
 		var resultValue Value
 		if returnType.IsResourceType() {
-			resultValue = NewEphemeralReferenceValue(interpreter, false, returnValue, returnType)
+			switch returnValue := returnValue.(type) {
+			// If this value is an optional value (T?), then transform it into an optional reference (&T)?.
+			case *SomeValue:
+				optionalType, ok := returnType.(*sema.OptionalType)
+				if !ok {
+					panic(errors.NewUnreachableError())
+				}
+				innerValue := NewEphemeralReferenceValue(
+					interpreter,
+					false,
+					returnValue.value,
+					optionalType.Type,
+				)
+				resultValue = NewSomeValueNonCopying(interpreter, innerValue)
+			case NilValue:
+				resultValue = NilValue{}
+			default:
+				resultValue = NewEphemeralReferenceValue(interpreter, false, returnValue, returnType)
+			}
 		} else {
 			resultValue = returnValue
 		}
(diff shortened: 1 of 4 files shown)


turbolent (commented):
Nice, good catch!

I guess this is mostly a UX issue where currently the result cannot be accessed

This logic should have probably changed back when we changed from reference-to-optional to optional-reference (e.g. in reference expressions).

Should this maybe be recursively defined, instead of just unwrapping one level? For example, given a (unusual) result type `R??` (where `R` is resource-kinded), we probably want to get `(&R)??`.

Also, I'm not quite sure, but we might have missed the resource-kinded reference tracking for this reference value creation here and should add it in case it is missing. We want to maybe also double check if there are other cases where we create references to resource-kinded values and are missing the tracking.

SupunS:
> This logic should have probably changed back when we changed from reference-to-optional to optional-reference (e.g. in > reference expressions).
> 
> Should this maybe be recursively defined, instead of just unwrapping one level? For example, given a (unusual) result type > R?? (where R is resource-kinded), we probably want to get (&R)??.

Yeah, I changed this to be aligned with the optional reference change (https://github.com/onflow/cadence/pull/1303). There also we only unwrap it just one level (see below). Not sure if we need to do the same for nested optionals as well. Need to think about it a bit more. But I think that should probably be a separate change (and maybe breaking).

https://github.com/onflow/cadence/blob/a87f3634b5d760caa20d28e9be3f95332374ea30/runtime/sema/check_reference_expression.go#L45-L50

turbolent:
As discussed earlier, agreed on aligning with the behaviour of reference expressions, i.e. not unwrapping nested optionals, leaving the current PR's behaviour 👍 

Though maybe unlikely to be a real problem currently, it would still be great to add the reference tracking sure for consistency reasons here, and maybe elsewhere where missing. It might be good to do that in a separate PR.

turbolent (approved):

PR #2447, opened 2023-04-14
Add reference tracking for the result variable

SupunS:
## Description

Follow-up on https://github.com/onflow/cadence/pull/2444.

______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -786,12 +786,15 @@ func (interpreter *Interpreter) resultValue(returnValue Value, returnType sema.T
 				returnValue.value,
 				optionalType.Type,
 			)
+
+			interpreter.maybeTrackReferencedResourceKindedValue(returnValue.value)
 			return NewSomeValueNonCopying(interpreter, innerValue)
 		case NilValue:
 			return NilValue{}
 		}
 	}
 
+	interpreter.maybeTrackReferencedResourceKindedValue(returnValue)
 	return NewEphemeralReferenceValue(interpreter, false, returnValue, returnType)
 }
 
@@ -4602,6 +4605,12 @@ func (interpreter *Interpreter) ValidateAtreeValue(value atree.Value) {
 				panic(errors.NewExternalError(err))
 			}
 		}
+	}
+}
+
+func (interpreter *Interpreter) maybeTrackReferencedResourceKindedValue(value Value) {
+	if value, ok := value.(ReferenceTrackedResourceKindedValue); ok {
+		interpreter.trackReferencedResourceKindedValue(value.StorageID(), value)
 	}
 }
 
--- a/runtime/tests/interpreter/attachments_test.go
+++ b/runtime/tests/interpreter/attachments_test.go
@@ -1599,24 +1599,37 @@ func TestInterpretAttachmentResourceReferenceInvalidation(t *testing.T) {
 		inter, _ := testAccount(t, address, true, `
             resource R {}
             attachment A for R {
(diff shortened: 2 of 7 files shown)


SupunS:
I checked for other places where references are created internally but not tracked; Seems the only remaining ones are related to attachments. I'm not really sure which one of those actually needs to be tracked, etc. 

@dsainati1 Can you maybe have a look? Just look for the usages of `NewEphemeralReferenceValue` in attachment contexts (I've checked all the other places/usages)

runtime/interpreter/interpreter.go
```diff
@@ -4605,6 +4608,12 @@ func (interpreter *Interpreter) ValidateAtreeValue(value atree.Value) {
 	}
 }
 
+func (interpreter *Interpreter) maybeTrackReferencedResourceKindedValue(value Value) {
```
turbolent, line 4611: 👍

turbolent (approved):

dsainati1:
Should this be merged into the `feature/stable-cadence` branch? I thought this reference tracking change was a Stable Cadence change. 

> Can you maybe have a look? Just look for the usages of `NewEphemeralReferenceValue` in attachment contexts

I took a look when I merged attachments into Stable Cadence. The only reference we need to track there is the references created via attachment access `v[T]`, which I believe we do. The references created to `self` and `base` are only "live" for the duration of the functions that enclose them, so we don't need to track them I believe.

SupunS:
> Should this be merged into the feature/stable-cadence branch? I thought this reference tracking change was a Stable Cadence change.

Reference tracking was already there in master/previous releases, but used for a different purpose: to keep the reference valid even after the resource is moved (i.e: to make it point to the correct resource instance). It was only re-purposed in Stable Cadence, to use it for invalidation.

> The references created to self and base are only "live" for the duration of the functions that enclose them, so we don't need to track them I believe.

Similar to the example [here](https://github.com/onflow/cadence/blob/53dad69e5eda2847db807b37e5138604d307b1ad/runtime/tests/interpreter/function_test.go#L182-L208), wouldn't it be possible to store the reference to the base/self temporarily, and return the reference later (in the same execution), but after the refrenced resource (attachment) is moved?

runtime/tests/interpreter/attachments_test.go
```diff
@@ -1599,24 +1599,37 @@ func TestInterpretAttachmentResourceReferenceInvalidation(t *testing.T) {
...
+                    self.id = 1
+                }
             }
-            fun test() {
+            fun test(): UInt8 {
                 let r <- create R()
                 let r2 <- attach A() to <-r
                 let a = r2[A]!
-                authAccount.save(<-r2, to: /storage/foo)
-                let i = a.foo()
```
SupunS, line 1609: There was a small catch with this test. When the resource is moved, the underlying atree value of the previous instance of the resource gets cleared and is used to distinguish between valid vs obsolete resources. Here in this case, since the function `a.foo()` doesn't access the atree value, no error is thrown here, even though the reference is no longer pointing to the correct resource.

I updated this test to access a field `a.id` instead of the function call, so it also accesses the internal atree value. then it started failing. But is now fixed once the resource reference tracking is added.

SupunS:
I also added the reference tracking for attachments (https://github.com/onflow/cadence/pull/2447/commits/6c22a63c5d73b5e2959833d931bfb4e6fe05c1c5). Can you have a nother look please?

SupunS:
@turbolent @dsainati1 Can you have one final look, please?

turbolent (approved):
Great work! Thank you for fixing all the cases 👍

PR #2452, opened 2023-04-18
Add Address.fromBytes to create Address from [UInt8]

### LINKED ISSUE #2433: Language support to create Address from [UInt8] byte array
### Issue to be solved

Support for initializing addresses from `[UInt8]`

### Background

The `Address` type has a `toBytes()` method, but doesn't support creating an `Address` from a `[UInt8]`.

### Suggested Solution

Add a `Address` constructor which accepts `[UInt8]`:

```cadence
let bytes = [50, ... 54]
let addrFromBytes:Address = Address(bytes) // panic if constructor argument is not convertible into a valid address
```

darkdrag00nv2:
Closes #2433 

## Description
This PR adds a new method `Address.fromBytes` to create an address from a byte array `[UInt8]`.

______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -2492,8 +2492,11 @@ type ValueConverterDeclaration struct {
 	max             Value
 	convert         func(*Interpreter, Value, LocationRange) Value
 	functionType    *sema.FunctionType
-	nestedVariables map[string]Value
-	name            string
+	nestedVariables []struct {
+		Name  string
+		Value Value
+	}
+	name string
 }
 
 // It would be nice if return types in Go's function types would be covariant
@@ -2681,12 +2684,16 @@ var ConverterDeclarations = []ValueConverterDeclaration{
 		convert: func(interpreter *Interpreter, value Value, locationRange LocationRange) Value {
 			return ConvertAddress(interpreter, value, locationRange)
 		},
-		nestedVariables: map[string]Value{
-			sema.AddressTypeFromBytesFunctionName: NewUnmeteredHostFunctionValue(
+		nestedVariables: []struct {
+			Name  string
+			Value Value
+		}{{
+			Name: sema.AddressTypeFromBytesFunctionName,
+			Value: NewUnmeteredHostFunctionValue(
 				sema.AddressConversionFunctionType,
 				AddressFromBytes,
 			),
-		},
+		}},
 	},
 	{
 		name:         sema.PublicPathType.Name,
@@ -3064,8 +3071,8 @@ var converterFunctionValues = func() []converterFunction {
 		addMember(sema.FromStringFunctionName, fromStringVal.hostFunction)
 
 		if declaration.nestedVariables != nil {
-			for name, variable := range declaration.nestedVariables {
-				addMember(name, variable)
+			for _, variable := range declaration.nestedVariables {
+				addMember(variable.Name, variable.Value)
 			}
 		}
 
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -2488,11 +2488,12 @@ var fromStringFunctionValues = func() map[string]fromStringFunctionValue {
 }()
 
 type ValueConverterDeclaration struct {
-	min          Value
-	max          Value
-	convert      func(*Interpreter, Value, LocationRange) Value
-	functionType *sema.FunctionType
-	name         string
+	min             Value
+	max             Value
+	convert         func(*Interpreter, Value, LocationRange) Value
+	functionType    *sema.FunctionType
+	nestedVariables map[string]Value
+	name            string
 }
 
 // It would be nice if return types in Go's function types would be covariant
@@ -2679,6 +2680,12 @@ var ConverterDeclarations = []ValueConverterDeclaration{
 		functionType: sema.AddressConversionFunctionType,
 		convert: func(interpreter *Interpreter, value Value, locationRange LocationRange) Value {
 			return ConvertAddress(interpreter, value, locationRange)
+		},
+		nestedVariables: map[string]Value{
+			sema.AddressTypeFromBytesFunctionName: NewUnmeteredHostFunctionValue(
+				sema.AddressConversionFunctionType,
+				AddressFromBytes,
+			),
 		},
 	},
 	{
@@ -3055,6 +3062,12 @@ var converterFunctionValues = func() []converterFunction {
 		fromStringVal := fromStringFunctionValues[declaration.name]
 
 		addMember(sema.FromStringFunctionName, fromStringVal.hostFunction)
+
+		if declaration.nestedVariables != nil {
+			for name, variable := range declaration.nestedVariables {
+				addMember(name, variable)
+			}
+		}
 
 		converterFuncValues[index] = converterFunction{
 			name:      declaration.name,
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -3366,6 +3366,22 @@ var AddressConversionFunctionType = &FunctionType{
 	},
 }
 
+const AddressTypeFromBytesFunctionName = "fromBytes"
+const AddressTypeFromBytesFunctionDocString = `
+Returns an Address from the given byte array
+`
+
+var AddressTypeFromBytesFunctionType = &FunctionType{
+	Parameters: []Parameter{
+		{
+			Label:          ArgumentLabelNotRequired,
+			Identifier:     "bytes",
+			TypeAnnotation: NewTypeAnnotation(ByteArrayType),
+		},
+	},
+	ReturnTypeAnnotation: NewTypeAnnotation(TheAddressType),
+}
+
 func init() {
 	// Declare a conversion function for the address type
 
@@ -3377,11 +3393,31 @@ func init() {
 		panic(errors.NewUnreachableError())
 	}
 
+	functionType := AddressConversionFunctionType
+
+	addMember := func(member *Member) {
+		if functionType.Members == nil {
+			functionType.Members = &StringMemberOrderedMap{}
+		}
+		name := member.Identifier.Identifier
+		if functionType.Members.Contains(name) {
+			panic(errors.NewUnreachableError())
+		}
+		functionType.Members.Set(name, member)
+	}
+
+	addMember(NewUnmeteredPublicFunctionMember(
+		functionType,
+		AddressTypeFromBytesFunctionName,
+		AddressTypeFromBytesFunctionType,
+		AddressTypeFromBytesFunctionDocString,
+	))
+
 	BaseValueActivation.Set(
 		typeName,
 		baseFunctionVariable(
 			typeName,
-			AddressConversionFunctionType,
+			functionType,
 			numberConversionDocString("an address"),
 		),
 	)
--- a/runtime/tests/checker/builtinfunctions_test.go
+++ b/runtime/tests/checker/builtinfunctions_test.go
@@ -85,6 +85,23 @@ func TestCheckToBytes(t *testing.T) {
 	})
 }
 
+func TestCheckAddressFromBytes(t *testing.T) {
+	t.Parallel()
+
+	checker, err := ParseAndCheck(t, `
+		let address = Address.fromBytes([1])
+	`)
+
+	require.NoError(t, err)
+
+	resType := RequireGlobalValue(t, checker.Elaboration, "address")
+
+	assert.Equal(t,
+		sema.TheAddressType,
+		resType,
+	)
+}
+
 func TestCheckToBigEndianBytes(t *testing.T) {
 
 	t.Parallel()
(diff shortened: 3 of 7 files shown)


darkdrag00nv2:
@turbolent Can you please take a quick look? Just want to verify that I am on the right track before I add tests and update documentation. Thanks!

turbolent (commented):
Looks good! 👍

darkdrag00nv2:
@turbolent This is ready for review now. Had to force push because I messed up my git history locally somehow.

runtime/interpreter/interpreter.go
```diff
@@ -3056,6 +3063,12 @@ var converterFunctionValues = func() []converterFunction {
 
 		addMember(sema.FromStringFunctionName, fromStringVal.hostFunction)
 
+		if declaration.nestedVariables != nil {
+			for name, variable := range declaration.nestedVariables {
```
darkdrag00nv2, line 3067: Not sure about the lint error upon range statement over map. Do you have suggestions?

turbolent: When iterating with a `range` over a Go `map`, the iteration order is non-deterministic. That may cause problems in Cadence, if the loop body has side-effects that may impact the execution result (execution results must be fully deterministic).

Though in this case the body is OK (it only mutates another map), and we could ignore the lint error here, it might be better to just ignore the non-determinism of the loop all-together. You could make the iteration order deterministic by just changing the type of field `nestedVariables` to `[]struct{Name string; Value Value}`, i.e. a slice of key-value pairs.

runtime/tests/interpreter/builtinfunctions_test.go
```diff
@@ -147,6 +149,53 @@ func TestInterpretToBytes(t *testing.T) {
...
+			require.Equal(t, expected, addressVal.ToAddress().Bytes())
+		})
+	}
+
+	runAddressTest(t, []byte{1}, "[1]")
+	runAddressTest(t, []byte{12, 34, 56}, "[12, 34, 56]")
+	runAddressTest(t, []byte{67, 97, 100, 101, 110, 99, 101, 33}, "[67, 97, 100, 101, 110, 99, 101, 33]")
+}
+
+func TestInterpretAddressFromBytesInvalid(t *testing.T) {
```
turbolent, line 179: Maybe group this and the test above into one test function with multiple subtests (`t.Run`).

Also, could you please add a test case which checks that `Address.fromButes(someAddress.toBytes()) == someAddress` holds for various `someAddress`?

runtime/tests/interpreter/builtinfunctions_test.go
```diff
@@ -147,6 +149,53 @@ func TestInterpretToBytes(t *testing.T) {
...
+
+	t.Parallel()
+
+	runAddressTest := func(t *testing.T, innerCode string) {
+		t.Run(innerCode, func(t *testing.T) {
+			t.Parallel()
+
+			code := fmt.Sprintf("fun test(): Address { \n return Address.fromBytes(%s) \n }", innerCode)
```
turbolent, line 187: Here and above: Using Go raw string literals (`) makes the code a bit more readable:

```suggestion
			code := fmt.Sprintf(`
                  fun test(): Address {
                      return Address.fromBytes(%s)
                  }
            	`,
				innerCode,
			)
```

turbolent (commented):
Nice! Thank you for adding this

runtime/interpreter/interpreter.go
darkdrag00nv2: Thanks, updated!

runtime/tests/interpreter/builtinfunctions_test.go
darkdrag00nv2: Added these cases.

runtime/tests/checker/builtinfunctions_test.go
```diff
@@ -85,6 +85,23 @@ func TestCheckToBytes(t *testing.T) {
 	})
 }
 
+func TestCheckAddressFromBytes(t *testing.T) {
```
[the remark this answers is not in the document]
darkdrag00nv2: Added test cases.

I do have one question regarding labelled arguments. I have declared the function with `ArgumentLabelNotRequired`. My assumption was that the label would be optional but user would still be able to provide that if they want.

Based on the tests, I've observed this to not be the case. Even if the user provides the correct label (`bytes` for this case), we seem to be throwing an error - https://github.com/onflow/cadence/blob/2abe1cb05f6ac90f148ab90b4f492486c318f5ea/runtime/sema/check_invocation_expression.go#L326-L341. Am I missing something or is that by design?

darkdrag00nv2:
@turbolent @dsainati1 This is ready for another round of reviews. Thanks!

runtime/sema/type.go
```diff
@@ -3366,6 +3366,22 @@ var AddressConversionFunctionType = &FunctionType{
 	},
 }
 
+const AddressTypeFromBytesFunctionName = "fromBytes"
+const AddressTypeFromBytesFunctionDocString = `
+Returns an Address from the given byte array
+`
+
+var AddressTypeFromBytesFunctionType = &FunctionType{
+	Parameters: []Parameter{
```
SupunS, line 3375: Future flag (no actions needed now): Need to add purity annotations to this function type, once the stable-cadence branch is merged to master. Created issue #2466 to track this.

runtime/tests/checker/builtinfunctions_test.go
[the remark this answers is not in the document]
SupunS: The argument label is not optional, either the label is mandatory (if the label is specified), or shouldn't have at all (if `ArgumentLabelNotRequired`)

runtime/tests/interpreter/builtinfunctions_test.go
```diff
@@ -147,6 +149,89 @@ func TestInterpretToBytes(t *testing.T) {
...
+
+	runValidCase(t, []byte{1}, "[1]")
+	runValidCase(t, []byte{12, 34, 56}, "[12, 34, 56]")
+	runValidCase(t, []byte{67, 97, 100, 101, 110, 99, 101, 33}, "[67, 97, 100, 101, 110, 99, 101, 33]")
+
+	runValidRoundTripCase(t, "0x436164656E636521")
+	runValidRoundTripCase(t, "0x01")
+	runValidRoundTripCase(t, "0x46716465AE633188")
+
+	runInvalidCase(t, "[12, 34, 56, 11, 22, 33, 44, 55, 66, 77, 88, 99, 111]")
```
SupunS, line 232: Can you also add a test case for the empty array case?

runtime/tests/interpreter/builtinfunctions_test.go
```diff
@@ -147,6 +149,89 @@ func TestInterpretToBytes(t *testing.T) {
...
+                      return Address.fromBytes(%s)
+                  }
+            	`,
+				innerCode,
+			)
+
+			inter := parseCheckAndInterpret(t, code)
+			_, err := inter.Invoke("test")
+
+			RequireError(t, err)
```
SupunS, line 220: Maybe also assert the specific expected error

runtime/tests/checker/builtinfunctions_test.go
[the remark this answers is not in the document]
bluesign: > The argument label is not optional, either the label is mandatory (if the label is specified), or shouldn't have at all (if ArgumentLabelNotRequired)

@SupunS  is this introduced recently ?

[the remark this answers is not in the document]
turbolent: When a parameter is declared with `_`, that declares that a call _must not_ provide a label for the parameter. For example, a function declaration `fun double(_ n: Int): Int` must be called like `double(1)` and cannot be called like `double(n: 1)`.

When a parameter is declared with an explicit argument label, that declares that a call _must_ provide the label. For example, a function declaration `fun double(number n: Int): Int` must be called like `double(number: 1)` and cannot be called like `double(1)` or `double(n: 1)`.

When a parameter is declared without an argument label, that declares that the parameter name is also the argument label and that a call _must_ provide the parameter name as the label. For example, a function declaration `fun double(number: Int): Int` must be called like `double(number: 1)` and cannot be called like `double(1)`.

This feature has been there since the inception of Cadence, and is directly taken from Swift: https://docs.swift.org/swift-book/documentation/the-swift-programming-language/functions/#Function-Argument-Labels-and-Parameter-Names

Unfortunately, this behaviour was not correctly implemented for calls of nested members. We fixed this a while ago in https://github.com/onflow/cadence/pull/1717. Because the fix is effectively a breaking change, we will only release it as part of the Stable Cadence milestone. See "Missing or Incorrect Argument Labels Not Reported" in https://forum.onflow.org/t/another-update-on-stable-cadence/3715.

runtime/tests/interpreter/builtinfunctions_test.go
```diff
@@ -147,6 +149,89 @@ func TestInterpretToBytes(t *testing.T) {
...
+		})
+	}
+
+	runValidCase(t, []byte{1}, "[1]")
+	runValidCase(t, []byte{12, 34, 56}, "[12, 34, 56]")
+	runValidCase(t, []byte{67, 97, 100, 101, 110, 99, 101, 33}, "[67, 97, 100, 101, 110, 99, 101, 33]")
+
+	runValidRoundTripCase(t, "0x436164656E636521")
+	runValidRoundTripCase(t, "0x01")
+	runValidRoundTripCase(t, "0x46716465AE633188")
```
turbolent, line 230:
```suggestion
	runValidCase(t, []byte{}, "[]")
	runValidCase(t, []byte{1}, "[1]")
	runValidCase(t, []byte{12, 34, 56}, "[12, 34, 56]")
	runValidCase(t, []byte{67, 97, 100, 101, 110, 99, 101, 33}, "[67, 97, 100, 101, 110, 99, 101, 33]")

	runValidRoundTripCase(t, "0x0")
	runValidRoundTripCase(t, "0x01")
	runValidRoundTripCase(t, "0x436164656E636521")
	runValidRoundTripCase(t, "0x46716465AE633188")
```

runtime/tests/interpreter/builtinfunctions_test.go
turbolent: Added in 177844b

runtime/tests/interpreter/builtinfunctions_test.go
```diff
@@ -147,6 +149,91 @@ func TestInterpretToBytes(t *testing.T) {
...
+	runValidCase(t, []byte{12, 34, 56}, "[12, 34, 56]")
+	runValidCase(t, []byte{67, 97, 100, 101, 110, 99, 101, 33}, "[67, 97, 100, 101, 110, 99, 101, 33]")
+
+	runValidRoundTripCase(t, "0x0")
+	runValidRoundTripCase(t, "0x01")
+	runValidRoundTripCase(t, "0x436164656E636521")
+	runValidRoundTripCase(t, "0x46716465AE633188")
+
+	runInvalidCase(t, "[12, 34, 56, 11, 22, 33, 44, 55, 66, 77, 88, 99, 111]")
+}
```
turbolent, line 235: Just an idea, does not have to be done in this PR: We could maybe use [Go's property testing feature](https://pkg.go.dev/testing/quick) here to generate random test cases

runtime/tests/checker/builtinfunctions_test.go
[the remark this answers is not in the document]
bluesign: thanks a lot @turbolent

runtime/tests/interpreter/builtinfunctions_test.go
darkdrag00nv2: Asserted the error as well.

runtime/tests/interpreter/builtinfunctions_test.go
darkdrag00nv2: Sure, we can look into in a future PR.

darkdrag00nv2:
@turbolent @SupunS PR suggestions incorporated, thanks!

SupunS (approved):
LGTM!

runtime/tests/interpreter/builtinfunctions_test.go
```diff
@@ -147,6 +149,92 @@ func TestInterpretToBytes(t *testing.T) {
...
+                  }
+            	`,
+				innerCode,
+			)
+
+			inter := parseCheckAndInterpret(t, code)
+			_, err := inter.Invoke("test")
+
+			RequireError(t, err)
+			require.ErrorAs(t, err, &common.AddressOverflowError)
```
turbolent, line 221: `ErrorAs` writes into the target pointer. Better to use `ErrorIs` for the equality check:

```suggestion
			require.ErrorIs(t, err, common.AddressOverflowError)
```

turbolent (approved):
Nice, thank you!

Issue #2091, opened 2022-10-24
Implement Updated Capability Controllers API


j1010001:
### Issue To Be Solved

Implement Capability Controllers API as proposed in https://github.com/onflow/flow/issues/798

```[tasklist]
### `Capability`
- [x] #2423
- [ ] https://github.com/onflow/cadence/issues/2459
```

```[tasklist]
### `Auth/PublicAccount`
- [x] `Auth/PublicAccount.Capabilities.get`
- [x] `Auth/PublicAccount.Capabilities.borrow`
- [x] `AuthAccount.Capabilities.publish`
- [x] `AuthAccount.Capabilities.unpublish`
```

```[tasklist]
### `AuthAccount.StorageCapabilities`
- [x] `issue`
- [x] `getController`
- [x] `getControllers`
- [x] `forEachController`
```

```[tasklist]
### `StorageCapabilityController`
- [x] `borrowType`
- [x] `capabilityID`
- [x] `delete`
- [x] `target`
- [x] `retarget`
- [x] `tag`
```

```[tasklist]
### `AuthAccount.AccountCapabilities`
- [x] `issue`
- [x] `getController`
- [x] `getControllers`
- [x] `forEachController`
``` 

```[tasklist]
### `AccountCapabilityController`
- [x] `borrowType`
- [x] `capabilityID`
- [x] `delete`
- [x] `tag`
```

```[tasklist]
### Other
- [x] Feature flag
- [x] Return `IDCapabilityValue` for `Auth/PublicAccount.getCapability`, when published
- [x] Migration functions (unlink -> issue storage CapCon, unlinkAccount -> issue account CapCon). https://github.com/onflow/flow/pull/798#issuecomment-1542632876
```

j1010001:
Closing - migration tracked separately in https://github.com/onflow/cadence/issues/2875

PR #2454, opened 2023-04-19
Implement Auth/PublicAccount.Capabilities functions

turbolent:
Work towards #2091 

## Description

Implement
- `Auth/PublicAccount.Capabilities.get`
- `Auth/PublicAccount.Capabilities.borrow`
- `AuthAccount.Capabilities.publish`
- `AuthAccount.Capabilities.unpublish`

I have test cases, but will open them in a separate PR, as they depend on issuing, which is currently stubbed out.

______

- [x] Targeted PR against ~`master`~ feature branch

### DIFF
--- a/runtime/stdlib/account.go
+++ b/runtime/stdlib/account.go
@@ -2217,8 +2217,8 @@ func newAuthAccountCapabilitiesValue(
 	return interpreter.NewAuthAccountCapabilitiesValue(
 		gauge,
 		addressValue,
-		newAccountCapabilitiesGetFunction(gauge, addressValue, sema.AuthAccountCapabilitiesTypeGetFunctionType),
-		nil,
+		newAccountCapabilitiesGetFunction(gauge, addressValue, sema.AuthAccountCapabilitiesTypeGetFunctionType, false),
+		newAccountCapabilitiesGetFunction(gauge, addressValue, sema.AuthAccountCapabilitiesTypeBorrowFunctionType, true),
 		newAuthAccountCapabilitiesPublishFunction(gauge, addressValue),
 		newAuthAccountCapabilitiesUnpublishFunction(gauge, addressValue),
 		func() interpreter.Value {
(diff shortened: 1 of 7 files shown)


dsainati1 (approved):

runtime/stdlib/account.go
```diff
@@ -2242,12 +2242,163 @@ func newAuthAccountCapabilitiesValue(
...
+				interpreter.TodoCapabilityID,
+				addressValue,
+				pathValue,
+				borrowStaticType,
+			)
+		},
+	)
+}
+
+func newAuthAccountCapabilitiesPublishFunction(
```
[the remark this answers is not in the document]
turbolent: For now the PRs implement what is proposed in the FLIP. 

I saw and replied to your feedback there, so far there has not been any further feedback.

I'm not a native speaker, but IMHO "publish" (to make/prepare something for public distribution) makes a lot of sense here, probably even more so than for the capability inbox, where something is only made available for a particular receiver. 

If you still feel like this is an issue, please comment on the FLIP, ideally with a suggestion of what it should be instead. Thanks!

runtime/stdlib/account.go
```diff
@@ -2256,7 +2407,109 @@ func newPublicAccountCapabilitiesValue(
...
+		newAccountCapabilitiesGetFunction(gauge, addressValue, sema.PublicAccountCapabilitiesTypeGetFunctionType, false),
+		newAccountCapabilitiesGetFunction(gauge, addressValue, sema.PublicAccountCapabilitiesTypeBorrowFunctionType, true),
+	)
+}
+
+func newAccountCapabilitiesGetFunction(
+	gauge common.MemoryGauge,
+	addressValue interpreter.AddressValue,
+	funcType *sema.FunctionType,
+	borrow bool,
```
SupunS, line 2419: Having to pass both the condition ('borrow' flag) and the type can be a little error-prone. Instead, maybe, either:
 - Use the `funcType` to determine whether it's a borrow or a get.
 - Or, maybe even better, accept only the 'borrow' flag, and pick the function type depending on the flag
   ```go
   if borrow {
       type = sema.PublicAccountCapabilitiesTypeBorrowFunctionType
   } else {
       type = sema.PublicAccountCapabilitiesTypeGetFunctionType
   }
   ```

turbolent: Good idea, improved it in af6282365b4e90f5c6b7622d91ff6139927ab3a1

SupunS (approved):

PR #2460, opened 2023-04-21
Invalidate references to nested resources upon the move of the outer resource

### LINKED ISSUE #2458: Reference invalidation doesn't track inner resource moves
### Current Behavior

Imagine a nested resource like:
```cadence
resource Foo {
    let bar: @Bar
}

resource Bar {
    let id: UInt8
}
```

Then get a reference to the inner field, and move the outer resource:

```cadence
// Get a reference to the inner resource
var barRef = &foo.bar as &Bar

// Move the resource
var foo2 <- foo

// Access the moved resource through the reference
barRef.id    // <-- This should panic, but currently doesn't
```

### Expected Behavior

See above

### Steps To Reproduce

See above

### Environment

```markdown
- Cadence version: Stable Cadence (not applicable for the master branch or any released versions)
```

SupunS:
Closes #2458

## Description

When a resource is moved, invalidate references taken not only to the moved resource, but also the references taken to the nested resources of that moved resource.

Also refactored the code to move all the invalidation logic to a single place, which was scattered in the Transfer/Destroy methods of each value type.
______


runtime/tests/interpreter/reference_test.go
```diff
@@ -1147,6 +1147,180 @@ func TestInterpretResourceReferenceInvalidationOnMove(t *testing.T) {
 		RequireError(t, err)
 		require.ErrorAs(t, err, &interpreter.InvalidatedResourceReferenceError{})
 	})
+
+	t.Run("nested resource in composite", func(t *testing.T) {
```
dsainati1, line 1151: I'm 99% sure the attachment equivalent of this will just work out of the box, since they are implemented the same way as nested resource fields, but just to double check, can you add another test here for the same behavior, but where `Baz` is an attachment for `Bar`?

SupunS: [This test ](https://github.com/onflow/cadence/blob/1a818a0d4a2bb36cfbc2d160dcb5b849c960bcd2/runtime/attachments_test.go#L149) already covers that path, so I didn't add anything new.

dsainati1: That test was already failing before this PR with a different error. The following did not previously fail on `stable-cadence`, so I'd like to make sure that it fails properly after this change:

```
t.Parallel()

inter := parseCheckAndInterpret(t, `
          var ref: &A? = nil
          resource R {}
          attachment A for R {
              let id: Int
              fun storeSelf() {
                  ref = self
              }  
              init() {
                  self.id = 3
              }
          }
          pub fun main(): @R? {
              var r <- attach A() to <- create R()
              r[A]!.storeSelf()
              return <- r
          }
          pub fun getRef(): Int {
              return ref!.id
          }
      `,
)

require.NoError(t, err)

_, err = inter.Invoke("getRef")
require.NoError(t, err)
```

SupunS: This one is also already there: https://github.com/onflow/cadence/blob/supun/sync-stable-cadence/runtime/tests/interpreter/attachments_test.go#L1754
It is now correctly failing once the master was merged to stable-cadence.

dsainati1: Oh perfect thanks

dsainati1 (approved):

turbolent:
I'll review this on Monday

SupunS:
@turbolent can you maybe have a look when you got some time?

turbolent (commented):
Great catch and work @SupunS! Sorry I missed this

Do you think there is maybe a way to implement this without having to traverse the whole object graph? The traversal can get quickly very expensive

fxamacker:
> The traversal can get quickly very expensive

Yeah, I agree.  While traversing whole object graph, additional register reads may be needed if they are not loaded already.

turbolent:
https://github.com/onflow/atree/pull/311 is ready now. Next step is to update to atree version which includes feature and using new iteration function

SupunS:
Updated to use the new atree APIs: f21739f5930bf5b85e2201bad0941ce6f8f887d5
Depends on the [feature/stable-cadence](https://github.com/onflow/atree/tree/feature/stable-cadence) branch of atree.

runtime/interpreter/interpreter.go
```diff
@@ -5544,25 +5544,66 @@ func (interpreter *Interpreter) trackReferencedResourceKindedValue(
...
+			interpreter.invalidateReferencedResources(element, destroyed)
+			return true
+		})
+		storageID = value.StorageID()
+	case *SomeValue:
+		interpreter.invalidateReferencedResources(value.value, destroyed)
+		return
+	default:
+		// skip non-container typed values.
+		return
```
turbolent, line 5578: Maybe add a test case here, e.g. referencing an `Int`

turbolent (approved):
Nice!

turbolent:
@SupunS Could you please add a benchmark that exercises this a bit? The current benchmarks don't, and it would be good to know what kind of performance impact the reference invalidation will have

runtime/interpreter/interpreter.go
SupunS: This line would not be reachable, because we already check for `!value.IsResourceKinded(interpreter)` at the start of this function.
Added the `default` case as a backup.

SupunS:
Added a benchmark in e90ddca948729bcfdd14e339d91e6fcae3f975a3. The results are as follows:

Master:
```
BenchmarkRuntimeResourceTracking-8   	1000000000	         0.005154 ns/op	       0 B/op	       0 allocs/op
```

Stable-Cadence (current):
```
BenchmarkRuntimeResourceTracking-8   	1000000000	         0.005317 ns/op	       0 B/op	       0 allocs/op
```

This PR:
```
BenchmarkRuntimeResourceTracking-8   	1000000000	         0.008469 ns/op	       0 B/op	       0 allocs/op
```

This test assumes the worst-case: i.e. All elements are loaded onto the memory (of an array of size 1000). 
So there is roughly a 59% increase (`(0.008469-0.005317)/0.005317*100`) for 6 move-operations, for a resource array with 1000 elements.

On a side note, I noticed that when the resource array is loaded from storage, all elements seem to be also 'loaded' every time (so they get iterated every time, during resource tracking). Not sure if it is due to the mocked 'test ledger' we are using for unit tests, or the expected behavior. cc: @fxamacker

turbolent:
Thank you for adding that benchmark @SupunS! I think the worst-case decrease is acceptable, if it becomes an issue we can look into optimizing it.

> On a side note, I noticed that when the resource array is loaded from storage, all elements seem to be also 'loaded' every time (so they get iterated every time, during resource tracking). Not sure if it is due to the mocked 'test ledger' we are using for unit tests, or the expected behavior. 

Hmm, that seems odd. Could you please open a separate issue (bug?) for that? We should look into it, normally this should not happen. It doesn't block this PR.

turbolent (approved):
Great work!

PR #2461, opened 2023-04-21
Implement StorageCapabilityController

turbolent:
⚠️ Depends on #2454
Work towards #2091 

## Description

- `StorageCapabilityController` was implemented as a `SimpleCompositeValue`. However, it needs to be storage, so refactor to a dedicated type
- Generate capability ID by introducing a new account ID generation function in the runtime interface.
  The function is not capability specific, so that it may used for other per-account ID generation needs that may arise in the future.
- Currently, storage maps may only be keyed by strings (hard-coded to `StringAtreeValue`). For example, storing a value under a storage path (e.g. `/storage/foo`), the path domain  (`storage`) is used to get the storage map, and the path identifier is used as the storage map key.
  
  However, we need to store capability controllers by capability ID.
  
  Refactor storage map keys and generalize them to support other non-string types. Introduce a `Uint64AtreeValue` similar to `StringAtreeValue`, which can be used to store values in storage maps keyed by `uint64`.

- Implement storage capability controller issuing. Store capability controller in two new storage maps (modeling and naming improvement suggestions very welcome): 
  - `cap_con`: Stores capability controllers. Maps capability ID (`uint64`) to capability controller (`CapabilityControllerValue`)
  - `path_cap`: Maps paths (really, their identifier, `StoragePath.identifier`) to a set of capability IDs. The capability ID set is implemented as a `interpreter.DictionaryValue`, which maps capability ID (`UInt64Value`) to `nil` (reduces storage size, encoded as just a single byte in CBOR). Using a dictionary allows easy iteration and non-linear insertion/removal. It would be maybe nice to have an `atree.OrderedSet` (cc @fxamacker).

- Implement `StorageCapbilityController` functionality (`id` field, `borrowType` field, `target` function, `retarget` function)

______


runtime/interpreter/storagemapkey.go
```diff
@@ -0,0 +1,75 @@
...
+
+package interpreter
+
+import "github.com/onflow/atree"
+
+type StorageMapKey interface {
+	isStorageMapKey()
+	AtreeValue() atree.Value
+	AtreeValueHashInput(v atree.Value, _ []byte) ([]byte, error)
+	AtreeValueCompare(storage atree.SlabStorage, value atree.Value, otherStorable atree.Storable) (bool, error)
```
turbolent, line 27: These two functions are really just "static" functions of an implementation of this interface, not instance methods: For a particular type that implements the `StorageMapKey` interface, we want to get the type's hash input function and value comparison functions.

Unfortunately, I didn't quite come up with a better way of expressing/implementing that.

runtime/capabilitycontrollers_test.go
```diff
@@ -0,0 +1,771 @@
...
+                              prepare(signer: AuthAccount) {
+                                  let storagePath = /storage/r
+                                  let publicPath = /public/r
+                                  let expectedCapID: UInt64 = 1
+                                  let resourceID = 42
+
+                                  // Arrange
+                                  Test.createAndSaveR(id: resourceID, storagePath: storagePath)
+                                  let issuedCap: Capability<&Test.R{}> =
+                                      signer.capabilities.storage.issue<&Test.R{}>(storagePath)
```
dsainati1, line 232: Perhaps it is too late in the process for this feedback, but I have to say that this feels really unwieldy for what seems like a fairly common operation; we're going from a single method call on the `AuthAccount` to a triple-nested call.

runtime/interpreter/statictype.go
```diff
@@ -387,6 +387,10 @@ func (t OptionalStaticType) Equal(other StaticType) bool {
 	return t.Type.Equal(otherOptionalType.Type)
 }
 
+var NilStaticType = OptionalStaticType{
+	Type: PrimitiveStaticTypeNever,
+}
```
dsainati1, line 392: Why is this defined this way?

runtime/interpreter/value_storagecapabilitycontroller.go
```diff
@@ -19,115 +19,212 @@
...
+	isCapabilityControllerValue()
+}
+
+// StorageCapabilityControllerValue
+
+type StorageCapabilityControllerValue struct {
+	BorrowType       StaticType
+	TargetPath       PathValue
+	CapabilityID     UInt64Value
+	RetargetFunction FunctionValue
```
dsainati1, line 41: How does the `RetargetFunction` part of this get stored? Or are we not storing the function?

runtime/interpreter/value_storagecapabilitycontroller.go
```diff
@@ -19,115 +19,212 @@
...
+
+func (v *StorageCapabilityControllerValue) Equal(interpreter *Interpreter, locationRange LocationRange, other Value) bool {
+	otherController, ok := other.(*StorageCapabilityControllerValue)
+	if !ok {
+		return false
+	}
+
+	return otherController.TargetPath.Equal(interpreter, locationRange, v.TargetPath) &&
+		otherController.BorrowType.Equal(v.BorrowType) &&
+		otherController.CapabilityID.Equal(interpreter, locationRange, v.CapabilityID)
```
dsainati1, line 136: Is it not necessary to compare the `RetargetFunction`?

runtime/stdlib/account.go
```diff
@@ -2272,22 +2344,220 @@ func newAuthAccountStorageCapabilitiesIssueFunction(
...
+				targetPathValue,
 				borrowStaticType,
 			)
 		},
 	)
 }
 
+// CapabilityControllerStorageDomain is the storage domain which stores
+// capability controllers by capability ID
+const CapabilityControllerStorageDomain = "cap_con"
```
dsainati1, line 2385: If these are new domains, is it possible to create a `Path` value with this domain? We should have some tests to make sure these can't "escape" into user code.

runtime/capabilitycontrollers_test.go
turbolent: This is good feedback, but please leave it on the FLIP instead, where we discuss the design (I had the same concerns, BTW).

runtime/interpreter/value_storagecapabilitycontroller.go
turbolent: So far, everywhere else we only compare inner values, not functions. Functions are also not comparable.

Here, once could even argue that two controllers are equal if there ID is equal.

runtime/stdlib/account.go
turbolent: Good point! We already have other storage domains which should not be user accessible, like for the capability inbox. I'll add tests if they don't already exist.

turbolent: Added tests in the PR feedback PR: [`fd60f78` (#2495)](https://github.com/onflow/cadence/pull/2495/commits/fd60f78532e32bfd38d7b343ac50f40c0f71971d)

dsainati1 (approved):

runtime/interpreter/interpreter.go
```diff
@@ -3292,22 +3288,24 @@ func (interpreter *Interpreter) IsSubTypeOfSemaType(subType StaticType, superTyp
...
+	var paths []Value
 
 	count := storageMap.Count()
 	if count > 0 {
-		values = make([]Value, 0, count)
-		for key := iterator.NextKey(); key != ""; key = iterator.NextKey() {
-			values = append(values, NewPathValue(interpreter, domain, key))
+		paths = make([]Value, 0, count)
+		for key := iterator.NextKey(); key != nil; key = iterator.NextKey() {
+			// TODO: unfortunately, the iterator only returns an atree.Value, not a StorageMapKey
```
SupunS, line 3302: Is there anything to do here in the future, that is worth creating an issue? or is it already handled in the follow-up PRs?

SupunS (approved):
LGTM!

runtime/interpreter/interpreter.go
turbolent: It would be nice to figure out a way to implement `StorageMapKey` more elegantly. The current approach works, it's just that it isn't elegant – e.g. here we have to cast. 

I left this TODO just as a general note that it would be nice to figure out a way to make it more elegant, it isn't addressed in a follow-up PR, nor does it warrant an issue.

PR #2462, opened 2023-04-21
Implement AuthAccount.StorageCapabilities.getControllers and forEachController

turbolent:
Work towards #2091 

## Description

Complete the implementation of `AuthAccount.StorageCapabilities`

______


runtime/stdlib/account.go
```diff
@@ -2297,20 +2297,156 @@ func newAuthAccountStorageCapabilitiesGetControllerFunction(
...
 
-			return interpreter.NewSomeValueNonCopying(inter, referenceValue)
+// the AccountKey in ` forEachController(forPath: StoragePath, _ function: ((&StorageCapabilityController): Bool))`
+var authAccountStorageCapabilitiesForEachControllerCallbackTypeParams = []sema.Type{
+	&sema.ReferenceType{
+		Type: sema.StorageCapabilityControllerType,
+	},
+}
+
+func newAuthAccountStorageCapabilitiesForEachControllerFunction(
```
dsainati1, line 2379: What is the behavior if a controller is added or removed during iteration? IMO the same model we use for storage iteration (and will use for attachment iteration) would be fine, i.e. to panic if mutation occurs. Either way we should have some tests for it.

turbolent: That's a good question/point and was also brought up by Janez in the PR review session yesterday.

I'm currently working on improving the behaviour to allow mutation while iterating, i.e. issuing new controllers and deleting existing controllers, and adding tests. As discussed yesterday, it probably is too limiting to reject mutation while iterating, as that is probably a common use case for the iteration function in the first place.

dsainati1: If it is possible to allow mutation during capcon iteration it is probably also possible to allow mutation during regular storage iteration as well then, considering they both iterate over a storage domain (if i understand correctly). We should look into relaxing that restriction if we can get this to work.

turbolent: I asked for some feedback on the desired behaviour on the FLIP: https://github.com/onflow/flow/pull/798#issuecomment-1542632876

turbolent: > If it is possible to allow mutation during capcon iteration it is probably also possible to allow mutation during regular storage iteration as well then, considering they both iterate over a storage domain (if i understand correctly).

It is possible, by taking a snapshot of all keys. This is potentially very (or even prohibitively) expensive when there are many entries (here, many controllers). As the point of the iteration functions is to avoid such expensive allocations, still snapshotting "under-the-hood" is probably not an option, but I've asked for feedback on it.

SupunS (approved):

PR #2463, opened 2023-04-23
Add support for generic function declarations

turbolent:
## Description

It turns out that adding basic support for generic functions does not take much effort, as we already have support for generic built-in functions.

- Remove the type parameter lists feature flag in the parser, enabling them constantly
- Add support function declarations with type parameter lists.

Even with the following limitations, this feature should already make some programs safer and allow developer to express their intent better:

- For now, type parameters may only occur in the parameter list and the return type.
 
   We may add support for type parameter occurrences in the function body later, which would require some work in the interpreter.

- Interface function declarations may not be generic (yet).
 
  

______

…clarations

### DIFF
--- a/runtime/tests/checker/genericfunction_test.go
+++ b/runtime/tests/checker/genericfunction_test.go
@@ -50,7 +50,7 @@ func parseAndCheckWithTestValue(t *testing.T, code string, ty sema.Type) (*sema.
 	)
 }
 
-func TestCheckGenericFunction(t *testing.T) {
+func TestCheckGenericFunctionInvocation(t *testing.T) {
 
 	t.Parallel()
 
@@ -925,7 +925,7 @@ func TestCheckBorrowOfCapabilityWithoutTypeArgument(t *testing.T) {
 	require.NoError(t, err)
 }
 
-func TestCheckUnparameterizedTypeInstantiationE(t *testing.T) {
+func TestCheckUnparameterizedTypeInstantiation(t *testing.T) {
 
 	t.Parallel()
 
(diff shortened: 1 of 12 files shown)


runtime/tests/checker/genericfunction_test.go
```diff
@@ -939,3 +939,167 @@ func TestCheckUnparameterizedTypeInstantiationE(t *testing.T) {
...
+	t.Run("global, struct", func(t *testing.T) {
+
+		t.Parallel()
+
+		_, err := ParseAndCheck(t, `
+          fun head<T: AnyStruct>(_ items: [T]): T? {
+              if items.length < 1 {
+                  return nil
+              }
+              return items[0]
```
dsainati1, line 956: I think we should have some more tests for these; in particular I'd like to see tests for when a generically-typed value "escapes" the scope of its function into an external/global variable. E.g. if I have a function like this that instead stores `items[0]` in a global array of type `[AnyStruct] vs one of type `[String]`, and how these subtyping checks interact with the type parameter bound.

dsainati1 (commented):
This is cool! Generics is usually a pretty complicated feature in how it interacts with the rest of a language's features though, so even if it's simple to enable here I think we want to make sure we test it thoroughly before enabling it. 

I also think we do need to add some interpreter tests to this PR. Even though we don't add support here for using the type variable inside the body of the function (like in a cast), it's still possible to downcast a value that is typed with a variable. E.g. if you have some parameter `v: T` where `T` is a type variable, you might have a later expression `v as? X` where `X` is some other type. IMO it's worth adding tests cases to the interpreter for these sorts of uses. Most particularly we should test the cases where `T` is a reference type, even if it's declared with an `AnyStruct` type bound we want to make sure that the previously identified bug with casting between `AnyStruct` and reference types doesn't remanifest itself here.

bluesign:
This will be super useful to write clean code. While this allow multiple Type parameters? 
something like: `pub fun cast<T, T2>(x: T): T2` (totally made up example, probably not makes sense here)

PR #2467, opened 2023-04-24
Fix race conditions in CCF tests caused by shared usage of types

turbolent:
## Description

Some getter functions of `cadence.Type`, like `ID()`, might mutate the receiver.

Some of the CCF tests currently use the same `cadence.Type`. As some of the tests run in parallel.

Detected on CI in an unrelated PR: 
https://github.com/onflow/cadence/actions/runs/4789029008/jobs/8516406084#step:6:11

Refactor the test cases so each test case creates individual `cadence.Type` instances that are not shared with other tests.

While at it, also parallelize some more tests.

Best viewed with whitespace changes disabled.

______

### DIFF
--- a/encoding/ccf/ccf_test.go
+++ b/encoding/ccf/ccf_test.go
@@ -13748,6 +13748,7 @@ func TestDecodeInvalidData(t *testing.T) {
 
 	test := func(tc testCase) {
 		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
 			decodedVal, err := ccf.Decode(nil, tc.data)
 			require.Nil(t, decodedVal)
 			require.Error(t, err)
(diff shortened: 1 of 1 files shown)


encoding/ccf/ccf_test.go
```diff
@@ -13421,17 +13746,22 @@ func TestDecodeInvalidData(t *testing.T) {
 		},
 	}
 
-	for _, tc := range testCases {
+	test := func(tc testCase) {
 		t.Run(tc.name, func(t *testing.T) {
```
SupunS, line 13750: Could be parallelized too?
```suggestion
		t.Run(tc.name, func(t *testing.T) {
			t.Parallel()
```

SupunS (approved):

PR #2470, opened 2023-04-25
Extend ordering comparison (<, <=, >, >=) to String

### LINKED ISSUE #2276: Extend ordering comparisons (< , <=, etc) to support container types

darkdrag00nv2:
Closes #2276 

## Description
Adds comparison support for `String`
______

### DIFF
--- a/docs/language/operators.md
+++ b/docs/language/operators.md
@@ -427,7 +427,7 @@ Comparison operators work with boolean and integer values.
   d3 != d4 // is `false`
   ```
 
-- Less than: `<`, for integers, booleans and characters
+- Less than: `<`, for integers, booleans, characters and strings
 
   ```cadence
   1 < 1  // is `false`
@@ -445,9 +445,17 @@ Comparison operators work with boolean and integer values.
   "z" < "a"    // is `false`
 
   "a" < "A"    // is `false`
+
+  "" < ""      // is `false`
+
+  "" < "a"     // is `true`
+
+  "az" < "b"   // is `true`
+
+  "xAB" < "Xab"  // is `false`
   ```
 
-- Less or equal than: `<=`, for integers, booleans and characters
+- Less or equal than: `<=`, for integers, booleans, characters and strings
 
   ```cadence
   1 <= 1  // is `true`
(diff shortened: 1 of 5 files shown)


dsainati1 (approved):

turbolent (approved):
Nice, thank you!

PR #2474, opened 2023-04-28
Introduce a `LocationFilter` field on `CoverageReport` type

m-Peter:
Work towards: https://github.com/onflow/developer-grants/issues/132
Context: https://github.com/onflow/flow-cli/pull/1013, https://github.com/onflow/cadence-tools/pull/110

## Description

As of now, there is an `ExcludeLocation` method which allows to exclude certain locations from the `CoverageReport`. However, regarding scripts/transactions, this is impossible to do.
With the proposed changes, our goal is to introduce a `LocationFilter` which can be utilized in order to inject custom logic on what types of locations should be inspected/collected from the `CoverageReport`.

______

### DIFF
--- a/runtime/coverage.go
+++ b/runtime/coverage.go
@@ -103,6 +103,8 @@ func NewLocationCoverage(lineHits map[int]int) *LocationCoverage {
 	}
 }
 
+type LocationFilter func(location Location) bool
+
 // CoverageReport collects coverage information per location.
 // It keeps track of inspected locations, and can also exclude
 // locations from coverage collection.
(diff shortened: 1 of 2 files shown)


m-Peter:
If this seems to be in the right direction, I'll complement it with tests :pray:

runtime/coverage.go
```diff
@@ -103,6 +103,8 @@ func NewLocationCoverage(lineHits map[int]int) *LocationCoverage {
 	}
 }
 
+type LocationInspectionHandler func(location Location) bool
```
turbolent, line 106: I wasn't sure what the purpose of the "handler" was, until I read the description of how it is used in the comment below https://github.com/onflow/cadence/pull/2474/files#diff-dd50d117359d58b9ec20ba8e56dc4b6b224a84ca819d943436617baec41d34d5R165. 

Maybe "filter" is a more specific name here? Maybe rename to just `LocationFilter`, and rename the `handler` field below to `LocationFilter` as well, so it is exported, like the other fields.

turbolent (commented):
Good idea!

runtime/coverage.go
m-Peter: Updated in https://github.com/onflow/cadence/pull/2474/commits/1a19dc8c17180fe902474cf87003ff64e5088989

PR #2478, opened 2023-05-02
Continue implementation of Cap Cons

turbolent:
Work towards #2091

## Description

- Improve type bound for type parameter of `AuthAccount.AccountCapabilities.issue`
- Implement `AuthAccount.AccountCapabilities.getController/getControllers/forEachController`
- Test `Capability.check` for ID capability value
- Fix borrow type of account reference value returned from account capability controller

______


turbolent:
@onflow/cadence PR is now ready for review, merged `master` into `feature/capability-controllers`, so this PR is now reviewable

dsainati1 (approved):

runtime/sema/authaccount.cdc
```diff
@@ -374,6 +374,6 @@ pub struct AuthAccount {
         pub fun forEachController(_ function: ((&AccountCapabilityController): Bool))
 
         /// Issue/create a new account capability.
-        pub fun issue<T: &AuthAccount>(): Capability<T>
+        pub fun issue<T: &AuthAccount{}>(): Capability<T>
```
SupunS, line 377: Why does the type parameter need to be a restricted type?
Can `AuthAccount` be actually restricted from any interface?

SupunS (approved):

runtime/sema/authaccount.cdc
turbolent: It looks a bit odd and is maybe counter-intuitive at first, but this is a type bound, so the type parameter must be a subtype of it. 

> Why does the type parameter need to be a restricted type?

Given that unrestricted types are sub-types of restricted types, the bound here _allows_ an unrestricted `&AuthAccount` type, but also a restricted `&AuthAccount{...}`, but it does not require it.

> Can `AuthAccount` be actually restricted from any interface?

Not at the moment, `AuthAccount` does not conform to any interfaces. 

I had made this change in light of https://forum.onflow.org/t/super-user-account/4088, the plan to somehow restrict access to `AuthAccount`, as a way to prepare for future additions. Given that we are likely going to define and use entitlements for `AuthAccount` or even refactor to a single `Account` type, this change here is probably not necessary. However, it just permits additional functionality (creation of restricted references) which is not possible yet, and does not reduce the functionality of the issue function, so we can just leave it for now and address it with the `Account` type refactor FLIP.

Issue #2484, opened 2023-05-03
Cadence excerpt on error formatting bug 


bluesign:
### Current Behavior

columns being negative after this causes problems, and cuts the error message.

https://github.com/onflow/cadence/blob/c6117d12ced0daa87e5097bc275c9c8e50744d0d/runtime/pretty/print.go#L337-L343

I think related to multiline error.

### Expected Behavior

Should handle multiline case and should not panic

### Steps To Reproduce

related issue: https://github.com/onflow/flow-cli/issues/1002

### Environment

```markdown
- Cadence version: v0.38.1
- Network: Emulator
```

turbolent:
Thank you for reporting this @bluesign! Do you have a reproducer for this?

turbolent:
@dsainati1 Could you maybe please have a look into this? You probably have the most knowledge about the multi-line support in the error pretty printer.

bluesign:
I can make a poc example later today. I saw when debugging find example in the linked issue.

bluesign:
This is the part causing problem:
startpos (287:29)
endpos (295:8)

when trying to mark:

```
287 |         var newNFT <- create NFT(
288 |             id: ExampleNFT.totalSupply,
289 |             name: name,
290 |             description: description,
291 |             thumbnail: thumbnail,
292 |             soulBound: soulBound,
293 |             traits: traits,
294 |             royalties: royalties
295 |         )
```

bluesign:
here is a minimal poc:

```cadence
pub contract X {

  pub resource R {

      init(s:String){
          panic("42")
      }

  }

  pub fun createR(): @R{
    return <- create R(
    s: "deniz"
    )

  }
}
```

dsainati1:
Thank you for the minimal repro! I will take a look at this when I get a chance

PR #2486, opened 2023-05-04
Don't truncate error message stack early in case of multiline errors

dsainati1:
Closes https://github.com/onflow/cadence/issues/2484

When the column of the `startPos` of an error was greater than the column of the `endPos`, this could cause the number of columns of `^` underline necessary for the error to be computed as negative. This fixes it to make the `^` length be based on the absolute value of the difference between these values, preventing the breakage.  

______


runtime/error_test.go
```diff
@@ -178,6 +178,66 @@ func TestRuntimeError(t *testing.T) {
...
+				"   |\n"+
+				"15 | 				destroy createR()\n"+
+				"   | 				        ^^^^^^^^^\n"+
+				"\n"+
+				"  --> 0100000000000000000000000000000000000000000000000000000000000000:9:21\n"+
+				"   |\n"+
+				" 9 | 				return <- create R(\n"+
+				"10 | 					s: \"argument\"\n"+
+				"11 | 				)\n"+
+				"   | 				^^^^^^^^^^^^^^^^^^\n"+
```
turbolent, line 231: It looks like the indicator is short of once character. Not quite sure, but is there maybe an "off-by-one" bug here?

turbolent (approved):

runtime/error_test.go
dsainati1: This isn't actually an off-by-one, it's an off by `len(name of resource)` error here that just looks like an off-by-one because the name here is `R`! Nice catch. Should be fixed now.

PR #2490, opened 2023-05-05
Add tag field to capability controllers

turbolent:
Work towards #2091 

## Description

The Capability Controllers FLIP was recently updated and added a `pub(set) var tag: String` field to the two capability controller types `StorageCapabilityController` and `AccountCapabilityController`:
[onflow/flow@`a1ed443` (#798)](https://github.com/onflow/flow/pull/798/commits/a1ed4433ce6e5c847f201e6be1567a2fb96f1995).

Store the tags in a separate account storage map instead of the controller itself, as the controller is loaded every time a capability is checked/borrowed. This was a great idea/recommended by @janezpodhostnik 🙏 

______


turbolent:
@onflow/cadence This is ready for review now

dsainati1 (approved):

runtime/interpreter/value_storagecapabilitycontroller.go
```diff
@@ -209,6 +214,9 @@ func (v *StorageCapabilityControllerValue) ChildStorables() []atree.Storable {
 func (v *StorageCapabilityControllerValue) GetMember(inter *Interpreter, _ LocationRange, name string) Value {
 
 	switch name {
+	case sema.StorageCapabilityControllerTypeTagFieldName:
+		return v.GetTag()
```
SupunS, line 218: I guess caching the tag value is not possible here because there can be multiple copies of the same controller?

SupunS (approved):

runtime/interpreter/value_storagecapabilitycontroller.go
turbolent: Actually, there is only ever one controller instance, all `get` calls that load controller always return the same value. So yes, actually we can cache the value here in the value 👌 I'll add that

turbolent: Added in ebafc50c4397a4c929eea84e3a2b395379f4aa9c

PR #2491, opened 2023-05-06
Implement Address.fromString

### LINKED ISSUE #2154: Add `fromString` conversion function to `Address` type
### Issue To Be Solved

Cadence now supports conversion of strings to all number types with [the latest mainnet spork](https://forum.onflow.org/t/cadence-updates-for-the-nov-2022-mainnet-spork/3748) using the built-in `fromString` method. 

It would be nice to have the same capability for creating addresses from strings.

This would allow for more flexibility in passing and manipulating variables, such as using `{String: String}` dictionaries to pass numbers and addresses as transaction inputs without having to use [conditional downcasting operators](https://developers.flow.com/cadence/language/operators#conditional-downcasting-operator-as) and related boilerplate code.

This would also allow users to query `PublicAccount` objects from strings, allowing for one to, for example, query a value stored on an account based on the address embedded in a run-time type identifier string.

### Suggested Solution

 An example of usage might look like:
```cadence
let input: String = "0xf8d6e0586b0a20c7"
let convertedAddress: Address? = Address.fromString(input) // returns a valid address

let badInput: String = "42"
let badConvertedAddress: Address? = Address.fromString(badInput) // returns nil
```

Given that this operation is quite similar to the `UInt64.fromString()` method (after all, addresses are [just unsigned 64-bit integers](https://developers.flow.com/cadence/language/values-and-types#addresses)), the implementation should be fairly straight-forward.

darkdrag00nv2:
Closes #2154 

## Description
Add `Address.fromString` method.

______

### DIFF
--- a/docs/language/values-and-types.mdx
+++ b/docs/language/values-and-types.mdx
@@ -415,16 +415,16 @@ let someAddress: Address = Address.fromBytes([67, 97, 100, 101, 110, 99, 101, 33
 let invalidAddress: Address = Address.fromBytes([12, 34, 56, 11, 22, 33, 44, 55, 66, 77, 88, 99, 111])
 
 // Declare an address with the string representation as "0x436164656E636521".
-let addressFromString: Address = Address.fromString("0x436164656E636521")
+let addressFromString: Address? = Address.fromString("0x436164656E636521")
 
-// Declare an address with the string representation without the "0x" prefix.
-let addressFromStringWithoutPrefix: Address = Address.fromString("436164656E636521")
+// Invalid: Provided value does not have the "0x" prefix. Returns Nil
+let addressFromStringWithoutPrefix: Address? = Address.fromString("436164656E636521")
 
-// Invalid: Provided value is a valid hex string. The function panics.
-let invalidAddressForInvalidHex: Address = Address.fromString("0xZZZ")
+// Invalid: Provided value is an invalid hex string. Return Nil.
+let invalidAddressForInvalidHex: Address? = Address.fromString("0xZZZ")
 
-// Invalid: Provided value is larger than 64 bits. The function panics.
-let invalidAddressForOverflow: Address = Address.fromString("0x436164656E63652146757265766572")
+// Invalid: Provided value is larger than 64 bits. Return Nil.
+let invalidAddressForOverflow: Address? = Address.fromString("0x436164656E63652146757265766572")
 ```
 
 ### Address Functions
--- a/docs/language/values-and-types.mdx
+++ b/docs/language/values-and-types.mdx
@@ -405,7 +405,7 @@ let aNumber = 0x436164656E636521
 // `aNumber` has type `Int`
 ```
 
-Address can also be created using a byte array.
+Address can also be created using a byte array or string.
 
 ```cadence
 // Declare an address with hex representation as 0x436164656E636521.
@@ -413,6 +413,18 @@ let someAddress: Address = Address.fromBytes([67, 97, 100, 101, 110, 99, 101, 33
 
 // Invalid: Provided value is not compatible with type `Address`. The function panics.
 let invalidAddress: Address = Address.fromBytes([12, 34, 56, 11, 22, 33, 44, 55, 66, 77, 88, 99, 111])
+
+// Declare an address with the string representation as "0x436164656E636521".
+let addressFromString: Address = Address.fromString("0x436164656E636521")
+
+// Declare an address with the string representation without the "0x" prefix.
+let addressFromStringWithoutPrefix: Address = Address.fromString("436164656E636521")
+
+// Invalid: Provided value is a valid hex string. The function panics.
+let invalidAddressForInvalidHex: Address = Address.fromString("0xZZZ")
+
+// Invalid: Provided value is larger than 64 bits. The function panics.
+let invalidAddressForOverflow: Address = Address.fromString("0x436164656E63652146757265766572")
 ```
 
 ### Address Functions
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -2688,13 +2688,22 @@ var ConverterDeclarations = []ValueConverterDeclaration{
 		nestedVariables: []struct {
 			Name  string
 			Value Value
-		}{{
-			Name: sema.AddressTypeFromBytesFunctionName,
-			Value: NewUnmeteredHostFunctionValue(
-				sema.AddressConversionFunctionType,
-				AddressFromBytes,
-			),
-		}},
+		}{
+			{
+				Name: sema.AddressTypeFromBytesFunctionName,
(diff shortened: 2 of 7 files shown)


docs/language/values-and-types.mdx
```diff
@@ -405,14 +405,26 @@ let aNumber = 0x436164656E636521
...
 let someAddress: Address = Address.fromBytes([67, 97, 100, 101, 110, 99, 101, 33])
 
 // Invalid: Provided value is not compatible with type `Address`. The function panics.
 let invalidAddress: Address = Address.fromBytes([12, 34, 56, 11, 22, 33, 44, 55, 66, 77, 88, 99, 111])
+
+// Declare an address with the string representation as "0x436164656E636521".
+let addressFromString: Address = Address.fromString("0x436164656E636521")
+
+// Declare an address with the string representation without the "0x" prefix.
+let addressFromStringWithoutPrefix: Address = Address.fromString("436164656E636521")
```
dsainati1, line 421: Hmm, I am not sure how I feel about this. This kind of "adaptive" API behavior feels very Javascript-y to me, for lack of a better way of phrasing it. In general I think it's better to have a simple, consistent behavior for these kinds of library functions, and not have them do anything other than exactly what they say on the tin. Since `toString` includes the `0x` prefix in the output, IMO `fromString` should require it, and if it's not present then treat the string as invalid input.

runtime/interpreter/interpreter.go
```diff
@@ -2688,13 +2688,22 @@ var ConverterDeclarations = []ValueConverterDeclaration{
...
+				Name: sema.AddressTypeFromBytesFunctionName,
+				Value: NewUnmeteredHostFunctionValue(
+					sema.AddressConversionFunctionType,
+					AddressFromBytes,
+				),
+			},
+			{
+				Name: sema.AddressTypeFromStringFunctionName,
+				Value: NewUnmeteredHostFunctionValue(
+					sema.AddressConversionFunctionType,
```
dsainati1, line 2702: Why does this have the same type as the function above?

docs/language/values-and-types.mdx
```diff
@@ -405,14 +405,26 @@ let aNumber = 0x436164656E636521
...
+
+// Invalid: Provided value is a valid hex string. The function panics.
+let invalidAddressForInvalidHex: Address = Address.fromString("0xZZZ")
```
dsainati1, line 424: Why is the behavior on invalid input to panic? This is pretty unfriendly to users. The specification in https://github.com/onflow/cadence/issues/2154 says the function should return `nil` on invalid input, which is a friendlier API. Can you change the behavior to have `fromString` return an optional instead of panicking on invalid input?

runtime/interpreter/value.go
```diff
@@ -18065,6 +18065,20 @@ func AddressFromBytes(invocation Invocation) Value {
...
 
+func AddressFromString(invocation Invocation) Value {
+	argument, ok := invocation.Arguments[0].(*StringValue)
+	if !ok {
+		panic(errors.NewUnreachableError())
+	}
+
+	addr, err := common.HexToAddress(argument.Str)
+	if err != nil {
+		panic(err)
```
dsainati1, line 18076: Should return `nil` instead.

dsainati1 (changes requested):

docs/language/values-and-types.mdx
turbolent: Agreed, the function should only accept the format that was generated by `toString`, i.e. require the `0x` prefix to indicate the hex format. This reduces potential confusion and safety issues, where e.g. a developer could intend an input like "1234" to mean "0x1234", whereas another could understand it to be in decimal format.

runtime/interpreter/interpreter.go
```diff
@@ -2688,13 +2688,22 @@ var ConverterDeclarations = []ValueConverterDeclaration{
...
-			Value: NewUnmeteredHostFunctionValue(
-				sema.AddressConversionFunctionType,
-				AddressFromBytes,
-			),
-		}},
+		}{
+			{
+				Name: sema.AddressTypeFromBytesFunctionName,
+				Value: NewUnmeteredHostFunctionValue(
+					sema.AddressConversionFunctionType,
```
turbolent, line 2695: While at it, rename `AddressConversionFunctionType` to `AddressTypeFromBytesFunctionType`

runtime/interpreter/value.go
```diff
@@ -18065,6 +18065,20 @@ func AddressFromBytes(invocation Invocation) Value {
...
+	}
+
+	return NewAddressValue(invocation.Interpreter, addr)
```
turbolent, line 18079: Once the function type has been corrected, return an optional here.

runtime/interpreter/interpreter.go
darkdrag00nv2: Done

runtime/interpreter/interpreter.go
darkdrag00nv2: Done, thanks for catching! I made the same mistake in the `Address.fromBytes` implementation. Fixed both of them.

runtime/interpreter/value.go

docs/language/values-and-types.mdx
darkdrag00nv2: Sure, done.

docs/language/values-and-types.mdx
darkdrag00nv2: Done, now we return Nil for those cases.

runtime/sema/type.go
```diff
@@ -3382,6 +3382,13 @@ var AddressTypeFromBytesFunctionType = &FunctionType{
 	ReturnTypeAnnotation: NewTypeAnnotation(TheAddressType),
 }
 
+const AddressTypeFromStringFunctionName = "fromString"
+const AddressTypeFromStringFunctionDocString = `
+Attempts to parse an Address from the input string. Panics on invalid input.
```
dsainati1, line 3387:
```suggestion
Attempts to parse an Address from the input string. Returns nil on invalid input.
```

dsainati1 (approved):

SupunS (approved):

turbolent (approved):
Nice!

runtime/sema/type.go
```diff
@@ -3382,6 +3382,22 @@ var AddressTypeFromBytesFunctionType = &FunctionType{
...
+var AddressTypeFromStringFunctionType = &FunctionType{
+	Parameters: []Parameter{
+		{
+			Label:          ArgumentLabelNotRequired,
+			Identifier:     "input",
+			TypeAnnotation: NewTypeAnnotation(StringType),
+		},
+	},
+	ReturnTypeAnnotation: NewTypeAnnotation(TheAddressType),
+}
```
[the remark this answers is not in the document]
turbolent: Ah, yes! 👍

PR #2492, opened 2023-05-07
Implement `check` functionality in AuthAccount

### LINKED ISSUE #564: Add `check` functionality to AuthAccount
### Issue To Be Solved

Currently we use `check` on capabilities to check if the type of the capability can actually be used to create a reference.
It would be nice to have the same thing for an `AuthAccount` storage path

### Suggest A Solution

`AuthAccount.check<&T>(StoragePath): Bool`

darkdrag00nv2:
Closes #564

## Description
Adds `AuthAccount.check(storagePath)` function to check if the type parameters can be used to create a reference for the object stored in the storage path.

TODO:
- [x] Tests
- [x] Documentation

______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -4034,6 +4034,18 @@ func (interpreter *Interpreter) authAccountCheckFunction(addressValue AddressVal
 				panic(errors.NewUnreachableError())
 			}
 
+			domain := path.Domain.Identifier()
+			identifier := path.Identifier
+
+			value := interpreter.ReadStored(address, domain, identifier)
+
+			if value == nil {
+				return FalseValue
+			}
+
+			// If there is value stored for the given path,
+			// check that it satisfies the type given as the type argument.
+
 			typeParameterPair := invocation.TypeParameterTypes.Oldest()
 			if typeParameterPair == nil {
 				panic(errors.NewUnreachableError())
@@ -4041,26 +4053,9 @@ func (interpreter *Interpreter) authAccountCheckFunction(addressValue AddressVal
 
 			ty := typeParameterPair.Value
 
-			referenceType, ok := ty.(*sema.ReferenceType)
-			if !ok {
-				panic(errors.NewUnreachableError())
-			}
+			valueStaticType := value.StaticType(interpreter)
 
-			reference := NewStorageReferenceValue(
-				interpreter,
-				referenceType.Authorized,
-				address,
-				path,
-				referenceType.Type,
-			)
-
-			// Attempt to dereference,
-			// which reads the stored value
-			// and performs a dynamic type check
-
-			return AsBoolValue(
-				reference.ReferencedValue(interpreter, invocation.LocationRange, false) != nil,
-			)
+			return AsBoolValue(interpreter.IsSubTypeOfSemaType(valueStaticType, ty))
 		},
 	)
 }
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -4037,7 +4037,9 @@ func (interpreter *Interpreter) authAccountCheckFunction(addressValue AddressVal
 			domain := path.Domain.Identifier()
 			identifier := path.Identifier
 
-			value := interpreter.ReadStored(address, domain, identifier)
+			storageMapKey := StringStorageMapKey(identifier)
+
+			value := interpreter.ReadStored(address, domain, storageMapKey)
 
 			if value == nil {
 				return FalseValue
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -4018,6 +4018,53 @@ func (interpreter *Interpreter) authAccountBorrowFunction(addressValue AddressVa
 	)
 }
 
+func (interpreter *Interpreter) authAccountCheckFunction(addressValue AddressValue) *HostFunctionValue {
+
+	// Converted addresses can be cached and don't have to be recomputed on each function invocation
+	address := addressValue.ToAddress()
+
+	return NewHostFunctionValue(
+		interpreter,
+		sema.AuthAccountTypeCheckFunctionType,
+		func(invocation Invocation) Value {
+			interpreter := invocation.Interpreter
+
+			path, ok := invocation.Arguments[0].(PathValue)
+			if !ok {
+				panic(errors.NewUnreachableError())
+			}
+
+			typeParameterPair := invocation.TypeParameterTypes.Oldest()
+			if typeParameterPair == nil {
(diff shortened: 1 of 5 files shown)


darkdrag00nv2:
This is ready for review now!

docs/language/accounts.mdx
```diff
@@ -220,6 +220,14 @@ pub struct AuthAccount {
...
     pub fun borrow<T: &Any>(from: StoragePath): T?
 
+    /// Returns true if the object in account storage under the given path satisfies the given type, 
+    /// i.e. could be borrowed using the given type.
+    ///
+    /// The given type must not necessarily be exactly the same as the type of the borrowed object.
+    ///
+    /// The path must be a storage path, i.e., only the domain `storage` is allowed.
+    pub fun check<T: &Any>(from: StoragePath): Bool
+
```
turbolent, line 230: @darkdrag00nv2 The Cadence documentation us currently being migrated to the https://github.com/onflow/docs repository. Could you please open a PR there, on top of https://github.com/onflow/docs/pull/58 (if it is still open at the time), then remove the docs change from this PR (another PR is removing the documentation)? Sorry for the inconvenience, I'm happy to assist

darkdrag00nv2: Sure, I'll move it.

darkdrag00nv2: Moved the docs changes to https://github.com/onflow/docs/pull/68 and removed from this PR.

dsainati1 (approved):

runtime/sema/authaccount.cdc
```diff
@@ -106,6 +106,14 @@ pub struct AuthAccount {
     /// The path must be a storage path, i.e., only the domain `storage` is allowed
     pub fun borrow<T: &Any>(from: StoragePath): T?
 
+    /// Returns true if the object in account storage under the given path satisfies the given type, 
+    /// i.e. could be borrowed using the given type.
+    ///
+    /// The given type must not necessarily be exactly the same as the type of the borrowed object.
+    ///
+    /// The path must be a storage path, i.e., only the domain `storage` is allowed.
+    pub fun check<T: &Any>(from: StoragePath): Bool
```
SupunS, line 115: The API here sounds a bit confusing to me. The reason is, references cannot be stored in a storage path, so checking whether the value stored belongs to a reference type doesn't sounds much sense. 

I think either the type parameter must be a non-reference type (i.e: `check<T: Any>` and the implementation should check whether the value belongs to concrete type `T`), OR we should maybe change the function name/description, to indicate that this API tries to check whether a reference can be borrowed from the given type `T` for the given path.

SupunS: > /// The given type must not necessarily be exactly the same as the type of the borrowed object.

My concern here is that it still is ambiguous: it says the type doesn't have to be the exact type, but doesn't say what the relationship between type `T` and the value stored at the path.

Just to add more: all other storage-related functions such as `load`, `copy`, and `save` functions (except `borrow`) take a non-reference type `T`. Similarly, I feel it's more-natural/less-ambiguous to have a non-reference type as the type-parameter. Such that, `check<T>(/some/path)` checks whether the value at the path is of type `T`. 
i.e: I feel this function should be the inverse check of `AuthAccount.typeAt()` (https://github.com/onflow/cadence/pull/1254)

docs/language/accounts.mdx
turbolent: Thank you, and sorry about the confusion there

runtime/sema/authaccount.cdc
turbolent: As far as I understood @joshuahannan's request in #564, the function is meant to determine if the path could be `borrow`ed with the given type. Maybe the main issue here is that the `check` name is potentially confusing.

SupunS: Yeah, either one of `canBorrow<T: &Any>()` or `check<T: Any>()` is OK to me. I think both of these should facilitate the requirement of https://github.com/onflow/cadence/issues/564.
Although I have a slight preference for `check` purely on naming-elegancy. But fine with either.

turbolent: @joshuahannan What are your thoughts?

turbolent (approved):
Nice!

We still might want to adjust the API (naming/type), but implementation looks good.

runtime/sema/authaccount.cdc
joshuahannan: I like `check` better also since it aligns with the `capability.check`. Also, will this work for stored capabilities, or will it only work for stored resources and structs?

like if I have a `@FlowToken.Vault` in a storage path, I can call `check` on the path and it'll return true, but if I have a valid `Capability<FlowToken.Vault>` in a storage path and I try to `check` it, will it also return true? That would be nice also

turbolent: @joshuahannan No, there is no such automatic/magic behaviour of handling stored capabilities specially.

@SupunS  `check<T>(path)` is basically then just sugar for `AuthAccount.type(at: path).isSubtype(of: Type<T>())`?

SupunS: yes, that's what I was thinking

darkdrag00nv2: Yeah, will change the signature to `check<T>(path)` for this PR. @joshuahannan I think a new issue can be filed for that.

darkdrag00nv2:
@turbolent @dsainati1 @SupunS Ready for review now. Have updated the signature as per discussion.

runtime/interpreter/interpreter.go
```diff
@@ -3726,6 +3726,48 @@ func (interpreter *Interpreter) authAccountBorrowFunction(addressValue AddressVa
...
+				panic(errors.NewUnreachableError())
+			}
+
+			domain := path.Domain.Identifier()
+			identifier := path.Identifier
+
+			value := interpreter.ReadStored(address, domain, identifier)
+
+			if value == nil {
+				return FalseValue
```
SupunS, line 3751: Can you also maybe add a test to cover this path?

SupunS (approved):
LGTM!

runtime/interpreter/interpreter.go
darkdrag00nv2: Done, added the test case.

darkdrag00nv2:
Force push after resolving merge conflicts.

runtime/sema/authaccount.cdc
darkdrag00nv2: Resolving this conversation.

turbolent (approved):
Nice!

turbolent:
@SupunS Could you please have a last look?

SupunS (approved):

PR #2494, opened 2023-05-09
Continue implementation of Cap Cons: Backward compatibility

turbolent:
Work towards #2091

## Description

- Consider ID capability when resolving path capability final target path; support borrow/check of path capability pointing to ID capability
- Add a temporary `migrateLink` function which replaces a link with a Cap Con

______


runtime/stdlib/account.go
```diff
@@ -3155,6 +3194,146 @@ func newAuthAccountCapabilitiesUnpublishFunction(
...
+					address,
+					borrowType,
+				)
+
+			default:
+				panic(errors.NewUnreachableError())
+			}
+
+			// Publish: overwrite link value with capability
+
```
turbolent, line 3306: @turbolent Explain why we have to publish also to private path

dsainati1 (approved):

runtime/sema/authaccount.cdc
```diff
@@ -338,6 +338,14 @@ pub struct AuthAccount {
         /// Returns the capability if one was published at the path.
         /// Returns nil if no capability was published at the path.
         pub fun unpublish(_ path: PublicPath): Capability?
+
+        /// **DEPRECATED**: This function only exists temporarily to aid in the migration of links.
+        /// This function will not be part of the final Capability Controller API.
+        ///
+        /// Migrates the link at the given path to a capability controller.
+        ///
+        /// Does not migrate intermediate links of the chain.
```
SupunS, line 347: Maybe add a brief description on what the return value is.

SupunS (approved):

runtime/stdlib/account.go
turbolent: Done in 2f9855661ea887ae4d85666a99cf2521483fe87a

runtime/sema/authaccount.cdc
turbolent: Added in fc6436f0a1f83bbffc7e44502f5102aa8f95ceb1

PR #2495, opened 2023-05-10
Cap Con implementation PR feedback

turbolent:
⚠️ Depends on #2494
Work towards #2091

## Description

- Improve storage capability controller unrecording: clear capability set from storage map if it becomes empty
- TODO: handle controller issuing/deletion during iteration

______

### DIFF
--- a/runtime/stdlib/account.go
+++ b/runtime/stdlib/account.go
@@ -2896,8 +2896,11 @@ func getPathCapabilityIDSet(
 	storageMap := inter.Storage().GetStorageMap(
 		address,
 		PathCapabilityStorageDomain,
-		true,
+		false,
 	)
+	if storageMap == nil {
+		return nil
+	}
 
 	readValue := storageMap.ReadValue(inter, storageMapKey)
 	if readValue == nil {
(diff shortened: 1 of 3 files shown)


runtime/stdlib/account.go
```diff
@@ -3747,6 +3853,18 @@ func newAuthAccountAccountCapabilitiesForEachControllerFunction(
 				panic(errors.NewUnreachableError())
 			}
 
+			// Prevent mutations (record/unrecord) to account capability controllers
+			// for this address during iteration
+
+			iterations := inter.SharedState.AccountCapabilityControllerIterations
+			iterations[address]++
```
dsainati1, line 3860: Hypothetically could we generalize this to being a map for all iterations over each different domain for each address? This would allow us to combine the storage iteration mutation and the capability controller iteration mutation into the same logic.

turbolent: Account Cap Cons do not have a path, but we can maybe use `common.PathDomainUnknown`

turbolent: Good idea! Simplified this in e5104616b035aaf4f0f9bf1e83252eca58c4a91a

SupunS (approved):

dsainati1 (approved):

PR #2496, opened 2023-05-12
Add `fromBigEndianBytes` conversion function to `Number` types

### LINKED ISSUE #2448: Add `fromBigEndianBytes` conversion function to `Number` types
## Issue to be solved

To convert a number to a byte array, `Number` types have a function `toBigEndianBytes`.

The inverse function to convert a byte array to a number is missing.

## Suggested Solution

Add `fromBigEndianBytes` functions to all `Number` types.

darkdrag00nv2:
Closes #2448 

## Description
Added `fromBigEndianBytes` to all the `Number` types - `Int*`, `UInt*`, `Fix*`, `Word*`.

We return `Nil` in case of invalid input such as overflow. Note that we also return `Nil` if the size of the byte array is greater than the size needed for the target type. E.g. `[0, 0]` does not lead to overflow for `UInt8` but we return `Nil` since we expect the byte array size to only be 1.

______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -2548,8 +2548,10 @@ var fromBigEndianBytesFunctionValues = func() map[string]fromBigEndianBytesFunct
 	declarations := []fromBigEndianBytesFunctionValue{
 		// signed int values
 		newFromBigEndianBytesFunction(sema.Int8Type, 1, func(i *Interpreter, b []byte) OptionalValue {
-			bytes := padWithZeroes(b, 1)
-			return NewSomeValueNonCopying(i, NewInt8Value(i, func() int8 { return int8(bytes[0]) }))
+			return NewSomeValueNonCopying(i, NewInt8Value(i, func() int8 {
+				bytes := padWithZeroes(b, 1)
+				return int8(bytes[0])
+			}))
 		}),
 		newFromBigEndianBytesFunction(sema.Int16Type, 2, func(i *Interpreter, b []byte) OptionalValue {
 			bytes := padWithZeroes(b, 2)
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -2502,8 +2502,22 @@ func padWithZeroes(b []byte, expectedLen int) []byte {
 		return b
 	}
 
-	res := make([]byte, expectedLen)
+	var res []byte
+	// use existing allocated slice if possible.
+	if cap(b) >= expectedLen {
+		res = b[:expectedLen]
+	} else {
+		res = make([]byte, expectedLen)
+	}
+
 	copy(res[expectedLen-l:], b)
+
+	// explicitly set to 0 for the first expectedLen - l bytes.
+	if cap(b) >= expectedLen {
+		for i := 0; i < expectedLen-l; i++ {
+			res[i] = 0
+		}
+	}
 	return res
 }
 
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -2530,6 +2530,7 @@ func newFromBigEndianBytesFunction(
 				return Nil
 			}
 
+			// overflow
 			if byteLength != 0 && len(bytes) > byteLength {
 				return Nil
 			}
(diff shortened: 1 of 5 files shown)


turbolent:
Just realized that #1917 fixed the behaviour for `toBigEndianBytes` on the Stable Cadence feature branch (`feature/stable-cadence`).
Maybe we should target this PR against Stable Cadence too @onflow/cadence ?

runtime/interpreter/interpreter.go
```diff
@@ -2488,6 +2489,176 @@ var fromStringFunctionValues = func() map[string]fromStringFunctionValue {
...
+		hostFunction: hostFunctionImpl,
+	}
+}
+
+var fromBigEndianBytesFunctionValues = func() map[string]fromBigEndianBytesFunctionValue {
+	declarations := []fromBigEndianBytesFunctionValue{
+		// signed int values
+		newFromBigEndianBytesFunction(sema.Int8Type, 1, func(i *Interpreter, b []byte) OptionalValue {
+			bytes := padWithZeroes(b, 1)
+			return NewSomeValueNonCopying(i, NewInt8Value(i, func() int8 { return int8(bytes[0]) }))
```
turbolent, line 2552: Perform the allocations inside of the metered constructor, so that memory metering occurs before allocation (here and for all other functions):

```suggestion
			return NewSomeValueNonCopying(i, NewInt8Value(i, func() int8 {
				bytes := padWithZeroes(b, 1)
				return int8(bytes[0])
			}))
```

turbolent, line 2552: It looks like all implementations always return `SomeValue`, maybe just automatically do that

runtime/interpreter/interpreter.go
```diff
@@ -2488,6 +2489,176 @@ var fromStringFunctionValues = func() map[string]fromStringFunctionValue {
...
+
+func padWithZeroes(b []byte, expectedLen int) []byte {
+	l := len(b)
+	if l > expectedLen {
+		panic(errors.NewUnreachableError())
+	} else if l == expectedLen {
+		return b
+	}
+
+	res := make([]byte, expectedLen)
```
turbolent, line 2505: It might be better to extend the capacity of the existing slice, instead of allocating a whole new slice

turbolent (commented):
Nice!

runtime/interpreter/interpreter.go
darkdrag00nv2: Done except for unbounded `Int` and `UInt` types where the constructor requires the memory usage beforehand.

darkdrag00nv2: Done

runtime/interpreter/interpreter.go
darkdrag00nv2: Done, now we reuse the slice if the capacity >= what we need. Else we have to allocate a new slice.

darkdrag00nv2:
> Maybe we should target this PR against Stable Cadence too @onflow/cadence ?

@turbolent sure, I can switch the branch to stable cadence if needed. In future, when the stable cadence branch is fully merged into master, these changes will also come in. Do let me know.

SupunS:
> Just realized that https://github.com/onflow/cadence/pull/1917 fixed the behaviour for toBigEndianBytes on the Stable Cadence feature branch (feature/stable-cadence).
Maybe we should target this PR against Stable Cadence too ?

Sorry, I missed this. Does the new API `fromBigEndianBytes` work with fixed sizes inputs or variable-sized inputs (both)?
If it seamlessly works for both, then I see no harm in merging this to master.

Otherwise, If we'll have to do any adjustments, then merging to Stable Cadence directly sounds like a good idea. It would save us the trouble of having to maintain two versions of this / update to match the Stable cadence. But like I said, if this is not the case, then I'm OK with merging to master.

darkdrag00nv2:
> Does the new API fromBigEndianBytes work with fixed-size inputs or variable-sized inputs (both)?
If it seamlessly works for both, then I see no harm in merging this to master.

@SupunS Yes, it works for both the cases. For variable-sized inputs, we add the necessary padding.

turbolent (approved):
Nice work!

turbolent:
@onflow/cadence could you please have a look?

SupunS (approved):

Issue #2389, opened 2023-03-16
Return overflowing value when multiplying for UInt256 and Int256 


quocle108:
### Issue to be solved

There are some cases in which we need to get the overflowing value instead of the maximum value when the overflow happens. For example this case [mulDiv(uint256 x, uint256 y, uint256 denominator)](https://github.com/PaulRBerg/prb-math/blob/main/src/Common.sol#L379).

In order to multiply a and b and get the result in 512 bits, we need to get the overflowing value when it happens.

darkdrag00nv2:
@turbolent If I understand correctly, we want to implement `Word128` and `Word256`? 

The title of this issue is a little confusing. We wouldn't want to change the behavior of `UInt256` and `Int256` I presume. They should still raise runtime errors on overflow. Please correct me if I am wrong.

darkdrag00nv2:
Thanks @turbolent 

I think we can implement `Word128` and `Word256` for starters. We can add more functions further if it is insufficient for @quocle108.

turbolent:
@darkdrag00nv2 Sounds good!

PR #2497, opened 2023-05-15
Introduce `Word128` integer type

darkdrag00nv2:
Work towards #2389 

## Description
This PR introduces the type `Word128` to Cadence.

______


runtime/interpreter/value.go
```diff
@@ -12939,6 +12939,558 @@ func (Word64Value) ChildStorables() []atree.Storable {
...
+		interpreter,
+		func() *big.Int {
+			res := new(big.Int)
+			if o.BigInt.Sign() < 0 {
+				panic(UnderflowError{LocationRange: locationRange})
+			}
+			if !o.BigInt.IsUint64() {
+				panic(OverflowError{LocationRange: locationRange})
+			}
+			return res.Lsh(v.BigInt, uint(o.BigInt.Uint64()))
```
darkdrag00nv2, line 13389: The `Lsh` and `Rsh` functions from the big integer module only supports uint64 for the shift size. So I am overflowing here for Word128. Not sure about the alternatives here. May be we can document that `a << b` is only supported when `b` fits a `uint64`.

runtime/sema/type.go
```diff
@@ -1606,6 +1612,19 @@ var (
 	Word64TypeMinInt = new(big.Int)
 	Word64TypeMaxInt = new(big.Int).SetUint64(math.MaxUint64)
 
+	// 2 << 128
```
turbolent, line 1615:
```suggestion
	// 1 << 128
```

runtime/sema/type_tags.go
```diff
@@ -219,6 +219,9 @@ const (
 	anyResourceAttachmentMask
 	anyStructAttachmentMask
 
+	// This is here because we didn't reserve type-tags for Word128 and Word256.
+	word128TypeMask
```
turbolent, line 223: @SupunS Type tags and masks are not stored right? Could we just move  this up?

runtime/tests/interpreter/builtinfunctions_test.go
```diff
@@ -474,6 +474,13 @@ func TestInterpretToBigEndianBytes(t *testing.T) {
 			"9223372036854775808":  {128, 0, 0, 0, 0, 0, 0, 0},
 			"18446744073709551615": {255, 255, 255, 255, 255, 255, 255, 255},
 		},
+		"Word128": {
+			"0":   {0},
+			"42":  {42},
+			"127": {127},
+			"128": {128},
+			"200": {200},
+		},
```
turbolent, line 483: Maybe also add some tests for larger values here, and while at it improve `UInt128` too

values.go
```diff
@@ -1229,6 +1229,70 @@ func (v Word64) String() string {
...
+	return Word128{
+		Value: big.NewInt(int64(i)),
+	}
+}
+
+func NewWord128FromBig(i *big.Int) (Word128, error) {
+	if i.Cmp(sema.Word128TypeMaxIntBig) > 0 || i.Sign() < 0 {
+		// When Sign() < 0, Mod will add sema.Word128TypeMaxIntPlusOneBig
+		// to ensure the range is [0, sema.Word128TypeMaxIntPlusOneBig)
+		i.Mod(i, sema.Word128TypeMaxIntPlusOneBig)
```
turbolent, line 1252: It might be better to just return an error here if the input is out of range, e.g. like in `NewUInt128FromBig`

values_test.go
```diff
@@ -572,6 +577,13 @@ func TestNumberValue_ToBigEndianBytes(t *testing.T) {
 			NewWord64(9223372036854775808):  {128, 0, 0, 0, 0, 0, 0, 0},
 			NewWord64(18446744073709551615): {255, 255, 255, 255, 255, 255, 255, 255},
 		},
+		"Word128": {
+			NewWord128(0):   {0},
+			NewWord128(42):  {42},
+			NewWord128(127): {127},
+			NewWord128(128): {128},
+			NewWord128(200): {200},
+		},
```
turbolent, line 586: Same as above, maybe add some test for larger values, and improve the tests for `UInt128` at the same time

turbolent (commented):
Nice!

runtime/sema/type_tags.go
darkdrag00nv2: FYI - The lower mask has reached its limit. So to move this up, we'll have to move something down into the upper mask section.

turbolent: Correct!

In most places we are not able to re-arrange things in such a way, because if there is existing data stored in the execution state, such a change would change the meaning of what is stored. 

If types/tags are not stored, but only exist at checking time in memory, then we could easily re-arrange entries in the lower and upper masks how we see fit.

SupunS: Yes, this is safe to re-arrange. Type tags are not stored.

The reserved positions are just to make it less cumbersome when adding new types. Please feel free to update this.

darkdrag00nv2: done, moved it to lower mask.

values.go
darkdrag00nv2: Done

runtime/tests/interpreter/builtinfunctions_test.go
darkdrag00nv2: Done

values_test.go
darkdrag00nv2: Done

runtime/interpreter/value.go
turbolent: Yes, probably best to document this

encoding/ccf/decode.go
```diff
@@ -797,6 +801,23 @@ func (d *Decoder) decodeWord64() (cadence.Value, error) {
...
+// language=CDDL
+// word128-value = bigint .ge 0
+func (d *Decoder) decodeWord128() (cadence.Value, error) {
+	// NewMeteredWord128FromBig checks if decoded big.Int is positive.
+	return cadence.NewMeteredWord128FromBig(
+		d.gauge,
+		func() *big.Int {
+			bigInt, err := d.dec.DecodeBigInt()
+			if err != nil {
+				panic(fmt.Errorf("failed to decode Word128: %s", err))
```
turbolent, line 814: Could you please add a test case for this line?

encoding/json/decode.go
```diff
@@ -565,6 +567,24 @@ func (d *Decoder) decodeWord64(valueJSON any) cadence.Word64 {
...
 }
 
+func (d *Decoder) decodeWord128(valueJSON any) cadence.Word128 {
+	value, err := cadence.NewMeteredWord128FromBig(
+		d.gauge,
+		func() *big.Int {
+			bigInt := d.decodeBigInt(valueJSON)
+			if bigInt == nil {
+				// TODO: propagate toString error from decodeBigInt
+				panic(errors.NewDefaultUserError("invalid Word128: %s", valueJSON))
```
turbolent, line 577: Is this a TODO for this PR? Also, please add a test case for this line, and the other uncovered lines below

runtime/sema/type_tags.go
```diff
@@ -650,6 +655,9 @@ func findSuperTypeFromUpperMask(joinedTypeTag TypeTag, types []Type) Type {
 	case invalidTypeMask:
 		return InvalidType
 
+	case interfaceTypeMask:
+		return getSuperTypeOfDerivedTypes(types)
+
```
SupunS, line 660: We can add `interfaceTypeMask` to the case below, which also calls the `getSuperTypeOfDerivedTypes` method. No need to add a new case.

dsainati1 (approved):

SupunS (approved):
Nice work!

runtime/interpreter/value.go
darkdrag00nv2: Sure, will do that in the documentation PR.

runtime/sema/type_tags.go
darkdrag00nv2: Done.

encoding/json/decode.go
darkdrag00nv2: > Is this a TODO for this PR? 

No, copy-paste from the implementation of `decodeUInt128` & `decodeInt128`.

Ack on the test case part, I'll add them soon.

encoding/ccf/decode.go
darkdrag00nv2: Done, added `TestDecodeWord128Invalid` in ccf_test.go

darkdrag00nv2:
Had to force push after resolving merge conflicts.

encoding/json/decode.go
darkdrag00nv2: @turbolent  Many of these uncovered lines are also not covered for other types. For e.g. `convert` function in `runtime/interpreter/interpreter.go` is uncovered. I think it would be nice to tackle that in a separate issue. Filed https://github.com/onflow/cadence/issues/2506 for the same, will be happy to tackle it post the grant milestones.

turbolent (approved):
Nice!

PR #2499, opened 2023-05-18
Add attachment keywords to keyword lists

dsainati1:
This properly adds the new attachments keywords to the keyword lists, which was missed when merging this to stable cadence. This also makes `remove`, `attach` and `to` soft keywords, so they can be used as names for fields and functions. Note that it will still fail to parse however when attempting to assign to a variable named `attach` or `remove`, as this is ambiguous. 

______

### DIFF
--- a/runtime/parser/statement_test.go
+++ b/runtime/parser/statement_test.go
@@ -2684,6 +2684,11 @@ func TestSoftKeywordsInStatement(t *testing.T) {
 	}
 
 	for _, keyword := range softKeywords {
+		// it's not worth the additional complexity to support assigning to `remove` or `attach`-named
+		// variables, so we just accept this as a parsing error
+		if keyword == keywordAttach || keyword == keywordRemove {
+			continue
+		}
 		testSoftKeyword(keyword)
 	}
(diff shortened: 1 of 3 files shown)


runtime/parser/statement_test.go
```diff
@@ -2684,6 +2684,11 @@ func TestSoftKeywordsInStatement(t *testing.T) {
 	}
 
 	for _, keyword := range softKeywords {
+		// it's not worth the additional complexity to support assigning to `remove` or `attach`-named
+		// variables, so we just accept this as a parsing error
+		if keyword == keywordAttach || keyword == keywordRemove {
+			continue
+		}
```
turbolent, line 2691: Maybe replace this with a different test above: Instead of checking the soft keyword in a statement, check it e.g. in a function declaration. That should never conflict, even for attach and remove statements

dsainati1: I want to leave this here just to document the behavior, but I will add an additional test for function decls as well.

turbolent (approved):
Nice!

PR #2505, opened 2023-05-25
Enrich test framework interface

m-Peter:
Work towards: https://github.com/onflow/developer-grants/issues/148
Companion PRs:
- https://github.com/onflow/cadence-tools/pull/113
- https://github.com/onflow/cadence-tools/pull/114
- https://github.com/onflow/cadence-tools/pull/122

## Description

Adds 3 new methods to the `TestFramework` interface, and their native implementations.

```cadence
import Test

pub let blockchain = Test.newEmulatorBlockchain()

// Retrieve all events from the blockchain
pub let events = blockchain.getEvents("")
// Retrieve events from the blockchain, filtered by type
pub let specificEvents = blockchain.getEvents("A.01cf0e2f2f715450.FooContract.NumberAdded")

// Retrieve the blockchain's service account. Can be used to sign transactions and send FLOW tokens to other test accounts
// among other things.
pub let serviceAccount = blockchain.serviceAccount()

// Retrieve all the log messages from the blockchain
blockchain.logs()
```

______


m-Peter:
I would like the team's input, on how to advance with this blocker: https://github.com/onflow/cadence/pull/2505/files#diff-48d2de22767ded1c341778e03aee3e1d3e1006ceeae566f4b0a3a7704c60758dR642-R646. Is there any workaround?

runtime/stdlib/contracts/test.cdc
```diff
@@ -258,6 +278,20 @@ pub contract Test {
...
+
+        /// Returns the service account of the blockchain. Can be used to sign
+        /// transactions with this account.
+        ///
+        pub fun serviceAccount(): Account
+
+        /// Returns all events emitted from the blockchain, optionally filtered
+        /// by event type.
+        ///
+        pub fun getEvents(_ eventType: String): [AnyStruct]
```
SupunS, line 294: I feel it would be better to have two separate functions for the two cases:
 1) Returning all events. Maybe `events()` with no params
 2) Returning filtered events. Maybe `eventsOfType(_ typeName: String)`

WDYT?

runtime/stdlib/test_emulatorbackend.go
```diff
@@ -494,6 +533,129 @@ func (t *testEmulatorBackendType) newUseConfigFunction(testFramework TestFramewo
...
+			arrayType := interpreter.NewVariableSizedStaticType(
+				inter,
+				interpreter.PrimitiveStaticTypeAnyStruct,
+			)
+
+			values := make([]interpreter.Value, len(events))
+			// for i, event := range events {
+			//  How do I convert a cadence.Event to interpreter.Value ?
+			// 	values[i] = event
+			// }
```
turbolent, line 646: Conversion should be possible using `runtime.ImportValue`

runtime/stdlib/test_emulatorbackend.go
```diff
@@ -494,6 +533,129 @@ func (t *testEmulatorBackendType) newUseConfigFunction(testFramework TestFramewo
...
+
```
SupunS, line 647: As for the conversion, we can return the `interpreter.Value` from the `testFramework.GetEvents()`. and have the conversion there.

This is also how the `RunScript` works currently: https://github.com/onflow/cadence-tools/blob/8ebd4b9835c4ad46bd28ece858292e367ece137a/test/emulator_backend.go#L167-L176. Probably don't need to return the errors like in scripts, returning just a `[]interpreter.Value` would be sufficient.

runtime/stdlib/test_emulatorbackend.go
m-Peter: Perfect, I found the relevant comments also: https://github.com/onflow/cadence/blob/master/runtime/interpreter/value.go#L102-L103. However, I think this should be done from the `EmulatorBackend` implementation, I ran into some cyclic import errors, while trying to use `runtime.ImportValue` inside the `stdlib` package (which already import `runtime`). Like the `RunScript` implementation https://github.com/onflow/cadence-tools/blob/8ebd4b9835c4ad46bd28ece858292e367ece137a/test/emulator_backend.go#L167-L176

runtime/stdlib/test_emulatorbackend.go
m-Peter: You are right. I think this is the only way to go about it, because I ran into some cyclic import errors, when trying to use `runtime.ImportValue` inside `stdlib` package (which already imports `runtime` package)

runtime/stdlib/contracts/test.cdc
m-Peter: Seems better indeed :+1: . I kept the single function in the `BlockchainBackend` interface, and added separate functions in the `Blockchain` struct. This way, developers will use the dedicated functions for the two cases, with one implementation under the hood. I did this to avoid adding boilerplate code. Let me know if that is fine, or we should split it even further down.

runtime/stdlib/test_emulatorbackend.go
m-Peter: I have made some changes in that direction, in this commit: https://github.com/onflow/cadence/pull/2505/commits/1a52bef821c72162cd4f3d3b216ce0a9312cd604. I will update accordingly the PR https://github.com/onflow/cadence-tools/pull/113. I am still not sure how to wrap a `[]interpreter.Value` inside a `interpreter.Value`, I will give it a shot though.

runtime/stdlib/test_emulatorbackend.go
m-Peter: Updated in https://github.com/onflow/cadence-tools/pull/113/commits/72eee50a3699d94abcc474141d6eec48668dfbcc

runtime/stdlib/test_emulatorbackend.go
m-Peter: Updated in https://github.com/onflow/cadence-tools/pull/113/commits/72eee50a3699d94abcc474141d6eec48668dfbcc

runtime/stdlib/contracts/test.cdc
SupunS: Sounds good

runtime/stdlib/contracts/test.cdc
```diff
@@ -95,6 +95,38 @@ pub contract Test {
...
+        /// Returns the service account of the blockchain. Can be used to sign
+        /// transactions with this account.
+        ///
+        pub fun serviceAccount(): Account {
+            return self.backend.serviceAccount()
+        }
+
+        /// Returns all events emitted from the blockchain.
+        ///
+        pub fun events(): [AnyStruct] {
```
SupunS, line 114: Given this API is guaranteed not to work because of https://github.com/onflow/cadence-tools/pull/113#discussion_r1205979891, maybe let's not expose it in the `Blockchain` yet (we can keep those in other places, like in the `BlockchainBackend`, etc.)?

runtime/stdlib/contracts/test.cdc
```diff
@@ -258,6 +290,24 @@ pub contract Test {
...
+
+        /// Returns the service account of the blockchain. Can be used to sign
+        /// transactions with this account.
+        ///
+        pub fun serviceAccount(): Account
+
+        /// Returns all events emitted from the blockchain, optionally filtered
+        /// by type name.
+        ///
+        pub fun events(_ typeName: String): [AnyStruct]
```
turbolent, line 306: Should this match the functions which got added above?

SupunS (commented):
Can you add some tests for the new functions where possible (e.g. `serviceAccount()`)

runtime/stdlib/contracts/test.cdc
SupunS: OR to save the trouble of having to do another release to get this in, maybe add a note saying this is still unimplemented, and at the implementation side (cadence-tools), we can return an empty array for now (better than panicing the tests).

runtime/stdlib/contracts/test.cdc
```diff
@@ -95,6 +95,38 @@ pub contract Test {
...
+        ///
+        pub fun events(): [AnyStruct] {
+            return self.backend.events("")
+        }
+
+        /// Returns all events emitted from the blockchain,
+        /// filtered by type name.
+        ///
+        pub fun eventsOfType(_ typeName: String): [AnyStruct] {
+            return self.backend.events(typeName)
```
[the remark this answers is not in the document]
m-Peter: That's a good idea :+1: I was worried that it will be impossible to pass a `Type` from a test script, because in integration tests, contracts are simply deployed to an address, and there is no import happening. Hence, it would not be possible to pass the concrete run-time type.

I found a workaround though:
```cadence
let type: Type = CompositeType("A.0ae53cb6e3f42a79.FlowToken.TokensDeposited")!
type == Type<FlowToken.TokensDeposited>() // => true
type.isSubtype(of: Type<FlowToken.TokensDeposited>()) // true
```

What do you think?

runtime/stdlib/contracts/test.cdc
m-Peter: I do not think that it's necessary. Relevant comment: https://github.com/onflow/cadence/pull/2505#discussion_r1206457859

runtime/stdlib/contracts/test.cdc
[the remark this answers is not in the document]
turbolent: That looks good to me!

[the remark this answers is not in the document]
m-Peter: I will change the API to accept a `Type` then, instead of a `String`

runtime/stdlib/contracts/test.cdc
SupunS: Actually, ^this can be part of the documentation, hence not a blocker for the PR.

runtime/stdlib/contracts/test.cdc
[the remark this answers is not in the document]
m-Peter: Updated here: https://github.com/onflow/cadence/pull/2505/files#diff-d3a08c1a13225bf0d1cee4f91a918d595e21c64c9fcfbcc1ac134d281dcf3a18R118-R123

For the test provider interface, I still use the type identifier, in order to match the relevant method signature from the emulator's blockchain:
- https://github.com/onflow/flow-emulator/blob/master/blockchain.go#L1077-L1078
- https://github.com/onflow/cadence-tools/pull/113/files#diff-250eb1fa6e3bb9b26ead394f26f14651772ad393772fc350e91608d608bba0adR497-R502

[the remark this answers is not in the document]
turbolent: Nice!

m-Peter:
For example: https://github.com/onflow/cadence-tools/blob/master/test/test_framework_test.go#L684-L700, test case for the `blockchain.createAccount()` method.

turbolent:
@m-Peter I see. Maybe we can add a mock/test implementation in a follow-up PR? It would be good to not have to rely on some downstream dependencies being the own tests (i.e. performing integration tests), and also have some unit tests up here. Doesn't block this PR, just some general thoughts

m-Peter:
@turbolent You are completely right, I think it will surely increase the developer's confidence for the added code. I will follow up with PR to tackle this :pray:

runtime/stdlib/contracts/test.cdc
```diff
@@ -95,6 +95,38 @@ pub contract Test {
...
+        ///
+        pub fun events(): [AnyStruct] {
+            return self.backend.events("")
+        }
+
+        /// Returns all events emitted from the blockchain,
+        /// filtered by type.
+        ///
+        pub fun eventsOfType(_ type: Type): [AnyStruct] {
+            return self.backend.events(type.identifier)
```
SupunS, line 122: Doesn't have to be in this PR (could be a follow-up PR), but maybe also change the `backend.events()` method to accept an optional `Type?` parameter, so we avoid converting from `string -> type -> string`. Implementation (cadence-tools) can decide to convert to string if needed, but probably don't have to do it here.

I suggested optional `Type?` so `nil` can be passed for retrieving all events.

turbolent, line 122:
```suggestion
            return self.backend.events(nil)
        }

        /// Returns all events emitted from the blockchain,
        /// filtered by type.
        ///
        pub fun eventsOfType(_ type: Type): [AnyStruct] {
            return self.backend.events(type)
```

runtime/stdlib/contracts/test.cdc
turbolent, line 306:
```suggestion
        pub fun events(_ type: Type?): [AnyStruct]
```

runtime/stdlib/test_emulatorbackend.go
```diff
@@ -494,6 +546,129 @@ func (t *testEmulatorBackendType) newUseConfigFunction(testFramework TestFramewo
...
+) *interpreter.HostFunctionValue {
+	return interpreter.NewUnmeteredHostFunctionValue(
+		t.eventsFunctionType,
+		func(invocation interpreter.Invocation) interpreter.Value {
+			eventType, ok := invocation.Arguments[0].(*interpreter.StringValue)
+			if !ok {
+				panic(errors.NewUnreachableError())
+			}
+
+			return testFramework.Events(invocation.Interpreter, eventType.Str)
```
turbolent, line 647: Need to assert `interpreter.OptionalValue` here, and maybe pass Go `nil` if `eventType` is `NilValue`, and `eventType.StaticType` otherwise.

runtime/stdlib/test-framework.go
```diff
@@ -57,6 +60,17 @@ type TestFramework interface {
...
 
 	StandardLibraryHandler() StandardLibraryHandler
+
+	Logs() []string
+
+	ServiceAccount() (*Account, error)
+
+	Events(
+		inter *interpreter.Interpreter,
+		typeName string,
```
turbolent, line 70:
```suggestion
		type interreter.StaticType,
```

turbolent (approved):
Nice!

runtime/stdlib/test_emulatorbackend.go
```diff
@@ -494,6 +546,135 @@ func (t *testEmulatorBackendType) newUseConfigFunction(testFramework TestFramewo
...
+		func(invocation interpreter.Invocation) interpreter.Value {
+			value, ok := invocation.Arguments[0].(interpreter.OptionalValue)
+			if !ok {
+				panic(errors.NewUnreachableError())
+			}
+
+			var eventType interpreter.StaticType = nil
+			_, isNilValue := value.(interpreter.NilValue)
+			if !isNilValue {
+				eventType = value.StaticType(invocation.Interpreter)
```
SupunS, line 650: I think we may have to cast the 'value' to a `TypeValue` and use the `TypeValue.Type` as the  `eventType`. Let me double check though.

turbolent: Yeah, the code should unwrap the optional (`value.(*interpreter.OptionalValue)`), then cast the inner value to `TypeValue`, and use its `Type` as `eventType`

PR #2509, opened 2023-05-27
Fix memory metering for loading stored values

### LINKED ISSUE #2507: Memory estimate to high when swapping an element in a Dictionary
### Current Behavior

Problem was discussed [here]( https://discord.com/channels/613813861610684416/1108479699732152503/1111755892644384819)

This cadence code 

```cadence
let oldToken <- self.ownedNFTs[id] <- token
```

will be very expensive if `self.ownedNFTs` is large, because this will be hit https://github.com/onflow/cadence/blob/v0.38.1/runtime/common/metering.go#L381  with the count being the number of items in that dictionary.

see debug stack:
Screenshot 2023-05-26 at 21 37 02

### Expected Behavior

It should not be that expensive.

### Steps To Reproduce

I don't have a minimal scenario unfortunately. I found this while debugging this transaction https://f.dnz.dev/9052fe617973a2b4a4f9c32bc4785b61308a951b33295532d3dbed2f305a2a94

### Environment

```markdown
- Cadence version: v0.38.1
- Network: Mainnet
```

turbolent:
Closes #2507 

## Description

Remove unnecessary memory metering for loading of stored values.

______


runtime/interpreter/value.go
```diff
@@ -15466,20 +15503,17 @@ func (v *CompositeValue) Transfer(
 	storable atree.Storable,
 ) Value {
 
-	baseUse, elementOverhead, dataUse, metaDataUse := common.NewCompositeMemoryUsages(v.dictionary.Count(), 0)
```
turbolent, line 15469: For composites and dictionaries, `NewAtreeMapPreAllocatedElementsMemoryUsage` is used when creating a copy of a value.

Does this include / is this an alternative to the memory metered when a new empty value is created?

Said differently: Is it OK to remove these lines, or do they have to be moved above `NewMapFromBatchData`, or do these memory kinds not apply for it and `MemoryKindAtreeMapPreAllocatedElement` is used instead?

runtime/common/metering.go
```diff
@@ -320,25 +320,28 @@ func atreeNodes(count uint64, elementSize uint) (leafNodeCount uint64, branchNod
 	return
 }
 
-func newAtreeMemoryUsage(count uint64, elementSize uint, array bool) (MemoryUsage, MemoryUsage) {
+func newAtreeArrayMemoryUsage(count uint64, elementSize uint) (MemoryUsage, MemoryUsage) {
```
turbolent, line 323: Let's have two self-describing functions instead of a boolean parameter

runtime/common/metering.go
```diff
@@ -370,28 +373,24 @@ func AdditionalAtreeMemoryUsage(originalCount uint64, elementSize uint, array bo
 	}
 }
 
-func NewArrayMemoryUsages(count uint64, elementSize uint) (MemoryUsage, MemoryUsage, MemoryUsage, MemoryUsage) {
```
turbolent, line 373: Separate the base memory usage out, so it is always metered, not just when atree memory metering is necessary

runtime/interpreter/value.go
```diff
@@ -1645,35 +1645,50 @@ func NewArrayValueWithIterator(
...
+}
+
+func newArrayValueFromAtreeArray(
+	gauge common.MemoryGauge,
+	staticType ArrayStaticType,
+	elementSize uint,
+	atreeArray *atree.Array,
+) *ArrayValue {
+
+	common.UseMemory(gauge, common.ArrayValueBaseMemoryUsage)
```
turbolent, line 1685: Always meter the base

runtime/interpreter/value.go
```diff
@@ -2527,19 +2542,17 @@ func (v *ArrayValue) Transfer(
 	remove bool,
 	storable atree.Storable,
 ) Value {
-	baseUsage, elementUsage, dataSlabs, metaDataSlabs := common.NewArrayMemoryUsages(v.array.Count(), v.elementSize)
```
turbolent, line 2530: Only meter if necessary, moved down

runtime/interpreter/value.go
```diff
@@ -2655,46 +2679,49 @@ func (v *ArrayValue) Transfer(
 func (v *ArrayValue) Clone(interpreter *Interpreter) Value {
 	config := interpreter.SharedState.Config
 
-	iterator, err := v.array.Iterator()
-	if err != nil {
-		panic(errors.NewExternalError(err))
-	}
-
-	baseUsage, elementUsage, dataSlabs, metaDataSlabs := common.NewArrayMemoryUsages(v.array.Count(), v.elementSize)
```
turbolent, line 2663: Use constructor function instead of manually metering and constructing

runtime/interpreter/value.go
```diff
@@ -16080,38 +16120,52 @@ func NewDictionaryValueWithAddress(
...
+}
+
+func newDictionaryValueFromAtreeMap(
+	gauge common.MemoryGauge,
+	staticType DictionaryStaticType,
+	elementSize uint,
+	atreeOrderedMap *atree.OrderedMap,
+) *DictionaryValue {
+
+	common.UseMemory(gauge, common.DictionaryValueBaseMemoryUsage)
```
turbolent, line 16162: Always meter base

runtime/interpreter/value.go
```diff
@@ -16938,23 +16992,18 @@ func (v *DictionaryValue) Transfer(
 	remove bool,
 	storable atree.Storable,
 ) Value {
-	baseUse, elementOverhead, dataUse, metaDataUse := common.NewDictionaryMemoryUsages(
```
turbolent, line 16941: Only meter if necessary. Is `AtreeMapPreAllocatedElementsMemoryUsage` enough though?

runtime/interpreter/storage.go
```diff
@@ -64,17 +64,33 @@ func ConvertStoredValue(gauge common.MemoryGauge, value atree.Value) (Value, err
...
+		switch staticType := typeInfo.(type) {
 		case DictionaryStaticType:
-			return newDictionaryValueFromConstructor(gauge, typeInfo, value.Count(), func() *atree.OrderedMap { return value }), nil
+			return newDictionaryValueFromAtreeMap(
+				gauge,
+				staticType,
+				DictionaryElementSize(staticType),
+				value,
+			), nil
+
```
bluesign, line 84: we already pass `staticType`, is there a case where count can be different then `DictionaryElementSize(staticType)` ?

turbolent: `newDictionaryValueFromAtreeMap` is the "lowest level" constructor, and requires an element size to be passed. 
It is true that here, it can be derived from the static type.

However, this allows [`newDictionaryValueFromConstructor`](https://github.com/onflow/cadence/pull/2509/files#diff-424575c293dc2bc3a28aab845829d7656fa83a7be4e026de8906935e23bbd52eR16132) to compute the element size once, use it for memory metering, and then pass it to `newDictionaryValueFromAtreeMap` so it can be reused.

dsainati1 (approved):
Reviewed for code quality; I am not 100% confident on the metering changes themselves.

turbolent:
Added it to the Impl Sync agenda

SupunS (approved):

runtime/interpreter/value.go
```diff
@@ -17735,13 +17792,18 @@ func (v *DictionaryValue) Clone(interpreter *Interpreter) Value {
...
-		isResourceKinded: v.isResourceKinded,
-		dictionary:       dictionary,
-		isDestroyed:      v.isDestroyed,
-	}
+	dictionary := newDictionaryValueFromAtreeMap(
+		interpreter,
+		v.Type,
+		v.elementSize,
+		orderedMap,
+	)
```
fxamacker, line 17800: Calling `newDictionaryValueFromAtreeMap` here in `DictionaryValue.Clone()` is different from `CompositeValue.Clone()` which [creates `CompositeValue` directly](https://github.com/onflow/cadence/blob/b83b7db3aa03c0d72b031170b04bc995832cab64/runtime/interpreter/value.go#L16318-16333).

`newDictionaryValueFromAtreeMap` meters `DictionaryValueBaseMemoryUsage`.

Should `DictionaryValue.Clone()` and/or `CompositeValue.Clone()` be updated to meter more consistently?

fxamacker (commented):
The approach makes sense!

I left a couple of comments regarding `NewDictionaryMemoryUsages` and `NewAtreeMapPreAllocatedElementsMemoryUsage` and a question about `DictionaryValue.Clone()`.

turbolent:
Rebased on Stable Cadence

runtime/interpreter/value.go
turbolent: @fxamacker I see. I tried to bring it back in [187ffc6](https://github.com/onflow/cadence/pull/2509/commits/187ffc6b45cb1c9f80dbeaf25758b0e16a163aac). Could you please have another look and see if that is correct?

runtime/interpreter/value.go
turbolent: @fxamacker I see. I tried to bring it back in [187ffc6](https://github.com/onflow/cadence/pull/2509/commits/187ffc6b45cb1c9f80dbeaf25758b0e16a163aac). Could you please have another look and see if that is correct?

runtime/interpreter/value.go
turbolent: @fxamacker Actually, Clone is never metered, it's never used by the interpreter / during execution, only for e.g. migrations. I removed the unnecessary metering in [351c3c6](https://github.com/onflow/cadence/pull/2509/commits/351c3c6d678b2a8adf33562df29eeebc8298a1f8). Could you please have another look if that makes sense?

turbolent:
@dsainati1 @SupunS @fxamacker Could you please have another look?

runtime/interpreter/value.go
fxamacker: The update looks good! :+1:

runtime/interpreter/value.go
fxamacker: The update looks good! :+1:

runtime/interpreter/value.go
fxamacker: Makes sense!  Thanks.

fxamacker (approved):
Nice!  :+1: I focused on metering related to atree.

PR #2511, opened 2023-05-29
Return a text message instead of NaN

kozlovb:
## Description

Return a text message instead of NaN for contracts that don't have non trivial runnable statments in relation to code coverage.

______

### DIFF
--- a/runtime/coverage.go
+++ b/runtime/coverage.go
@@ -64,6 +64,7 @@ func (c *LocationCoverage) Percentage() string {
 		// location.
 		coveredLines = c.Statements
 	}
+
 	percentage := 100 * float64(coveredLines) / float64(c.Statements)
 	return fmt.Sprintf("%0.1f%%", percentage)
 }
@@ -245,6 +246,10 @@ func (r *CoverageReport) Percentage() string {
 // String returns a human-friendly message for the covered
 // statements percentage.
 func (r *CoverageReport) String() string {
+	percentage := r.Percentage()
+	if percentage == "NaN%" {
+		return fmt.Sprintf("There are no statements to cover")
+	}
 	return fmt.Sprintf("Coverage: %v of statements", r.Percentage())
 }
 
--- a/runtime/coverage.go
+++ b/runtime/coverage.go
@@ -237,17 +237,20 @@ func (r *CoverageReport) IsLocationInspected(location Location) bool {
 func (r *CoverageReport) Percentage() string {
 	totalStatements := r.Statements()
 	totalCoveredLines := r.Hits()
+	var percentage float64 = 100
+	if totalStatements != 0 {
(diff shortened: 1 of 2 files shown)


runtime/coverage.go
```diff
@@ -245,6 +246,10 @@ func (r *CoverageReport) Percentage() string {
 // String returns a human-friendly message for the covered
 // statements percentage.
 func (r *CoverageReport) String() string {
+	percentage := r.Percentage()
+	if percentage == "NaN%" {
```
turbolent, line 250:
```suggestion
	if r.Statements() == 0 {
```

runtime/coverage.go
```diff
@@ -245,6 +246,10 @@ func (r *CoverageReport) Percentage() string {
 // String returns a human-friendly message for the covered
 // statements percentage.
 func (r *CoverageReport) String() string {
+	percentage := r.Percentage()
```
turbolent, line 249: `Percentage` itself should probably check for `Statements() == 0` itself, and return "100%" in that case.

turbolent (commented):
Thank you for opening a PR to improve this! 👍 

cc @m-Peter

runtime/coverage.go
kozlovb: @turbolent  Thank you for the comment. In a way 100% make sense - if there are no statements to cover then everything is covered. Although, if this method is used and combined with a JSON output like '{
  "coverage": {},
  "excluded_locations": [
    "I.Crypto",
    "I.Test",
    "s.7465737400000000000000000000000000000000000000000000000000000000"
  ]
}' or a similar summary this may be slightly misleading. And "NaN%" would still allow clients of Persentage() to treat it diffrently from 100%. @m-Peter told me that for example Percentage() would be used to generate a text summary sent to CI. But on the other hand pushing work on clients expecting a valid "string" number also isn't that good. I'll do as you proposed.

runtime/coverage.go
kozlovb: Thx, will apply.

turbolent (approved):
Nice! Thank you for improving this 👍

runtime/coverage.go
```diff
@@ -236,15 +237,22 @@ func (r *CoverageReport) IsLocationInspected(location Location) bool {
...
-		100*float64(totalCoveredLines)/float64(totalStatements),
+		percentage,
 	)
 }
 
 // String returns a human-friendly message for the covered
 // statements percentage.
 func (r *CoverageReport) String() string {
+	if r.Statements() == 0 {
+		return fmt.Sprintf("There are no statements to cover")
```
turbolent, line 254:
```suggestion
		return "There are no statements to cover"
```

kozlovb: good catch, indeed !

PR #2517, opened 2023-05-30
Improve export and encoding

turbolent:
## Description

- Add support for encoding attachments to JSON-CDC codec
- Handle `nil` values when encoding JSON-CDC. This occurs for example in the result of a cyclic reference
- Fix encoding of composite values which have attachments attached to them in JSON-CDC encoding 
- In general, test that script return values in test cases can also be encoded by JSON-CDC. 
  It would be nice to do that for CCF as well, once it is ready

cc @fxamacker 

______

### DIFF
--- a/encoding/json/encode.go
+++ b/encoding/json/encode.go
@@ -615,6 +615,8 @@ func prepareAttachment(v cadence.Attachment) jsonValue {
 }
 
 func prepareComposite(kind, id string, fieldTypes []cadence.Field, fields []cadence.Value) jsonValue {
+	// Ensure there are _at least _ as many field values as field types.
+	// There might be more field values in the case of attachments.
 	if len(fields) < len(fieldTypes) {
 		panic(fmt.Errorf(
 			"%s field count (%d) does not match declared type (%d)",
@@ -628,6 +630,9 @@ func prepareComposite(kind, id string, fieldTypes []cadence.Field, fields []cade
 
 	for i, value := range fields {
 		var name string
+		// Provide the field name, if the field type is available.
+		// In the case of attachments, they are provided as field values,
+		// but there is no corresponding field type.
 		if i < len(fieldTypes) {
 			fieldType := fieldTypes[i]
 			name = fieldType.Identifier
--- a/encoding/json/encode.go
+++ b/encoding/json/encode.go
@@ -253,6 +253,7 @@ const (
 	dictionaryTypeStr  = "Dictionary"
 	structTypeStr      = "Struct"
 	resourceTypeStr    = "Resource"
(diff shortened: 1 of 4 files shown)


encoding/json/encode.go
```diff
@@ -618,10 +627,14 @@ func prepareComposite(kind, id string, fieldTypes []cadence.Field, fields []cade
 	compositeFields := make([]jsonCompositeField, len(fields))
 
 	for i, value := range fields {
-		fieldType := fieldTypes[i]
+		var name string
+		if i < len(fieldTypes) {
+			fieldType := fieldTypes[i]
+			name = fieldType.Identifier
+		}
```
SupunS, line 634: We already check this at the top of the function, so wouldn't need to do it here again?

turbolent: I'll add comments to clarify

encoding/json/encode.go
```diff
@@ -605,8 +610,12 @@ func prepareEnum(v cadence.Enum) jsonValue {
 	return prepareComposite(enumTypeStr, v.EnumType.ID(), v.EnumType.Fields, v.Fields)
 }
 
+func prepareAttachment(v cadence.Attachment) jsonValue {
+	return prepareComposite(attachmentTypeStr, v.AttachmentType.ID(), v.AttachmentType.Fields, v.Fields)
+}
+
 func prepareComposite(kind, id string, fieldTypes []cadence.Field, fields []cadence.Value) jsonValue {
-	if len(fieldTypes) != len(fields) {
+	if len(fields) < len(fieldTypes) {
```
turbolent, line 618:
```suggestion
	// Ensure there are _at least _ as many field values as field types.
	// There might be more field values in the case of attachments.
	if len(fields) < len(fieldTypes) {
```

encoding/json/encode.go
```diff
@@ -618,10 +627,14 @@ func prepareComposite(kind, id string, fieldTypes []cadence.Field, fields []cade
 	compositeFields := make([]jsonCompositeField, len(fields))
 
 	for i, value := range fields {
-		fieldType := fieldTypes[i]
+		var name string
+		if i < len(fieldTypes) {
```
turbolent, line 631:
```suggestion
		// Provide the field name, if the field type is available.
		// In the case of attachments, they are provided as field values,
		// but there is no corresponding field type.
		if i < len(fieldTypes) {
```

encoding/json/encode.go
turbolent: Explained it in [a9b2f77](https://github.com/onflow/cadence/pull/2517/commits/a9b2f77ed8350778c742eb9a69cdf6118f028e83)

Issue #2482, opened 2023-05-02
Add a `Range<T: Integer>` type and allow in for-loop


turbolent:
- [x] Add a new built-in type `Range`:

  ```
  struct Range<T: Integer> {
     let start: T
     let endInclusive: T
  }
  ```

- [x] Add support for iterating over `Range` in for-in statements
```[tasklist]
### Tasks
- [ ] https://github.com/onflow/cadence/issues/2885
- [ ] https://github.com/onflow/cadence/issues/2884
- [ ] https://github.com/onflow/cadence/issues/2886
- [ ] https://github.com/onflow/cadence/issues/2870
- [ ] https://github.com/onflow/cadence/issues/2871
```

turbolent:
Before starting an implementation, we should discuss the design and publish a proper design document, potentially a FLIP.

bluesign:
Maybe it is good time to expose built in interfaces (comparable, enumerable etc) to Cadence. Also ‘step’ here can be useful too. ( reverse ranges especially)

turbolent:
It might be a good idea to include the concept of a step, but for the concept of a range, that seems a bit odd.

For example, in [Kotlin](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/-int-range/-init-.html), [Swift](https://developer.apple.com/documentation/swift/range), and [Rust](https://doc.rust-lang.org/std/ops/struct.Range.html), ranges do not have a step.

Instead, some languages have a separate concept for [progression](https://en.wikipedia.org/wiki/Arithmetic_progression), e.g. Kotlin has [`Progression`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/-int-progression/), and Swift has [`StrideThrough`](https://developer.apple.com/documentation/swift/stridethrough). It looks like Rust has the concept of a "wrapper iterator" that allows stepping, [`StepBy`](https://doc.rust-lang.org/std/iter/struct.StepBy.html).

Maybe we can start with a simple `Range`, and then follow up with a progression type.

bluesign:
Are we planning ranges to be Values? If so we also need to discuss about the API they will have.

turbolent:
@bluesign Yes, see the description

darkdrag00nv2:
Kotlin ranges only have two member functions:

1. `contains<T>(value: T)`: Check if T belongs in the Range
2. `isEmpty()`: Checks if the Range is empty

[...]

PR #2523, opened 2023-06-01
Introduce `Range<T: Integer>` type

darkdrag00nv2:
Work towards #2482

## Description
Adds a new `Range<T: Integer>` type to Cadence.

Example
```
let a = 1 .. 10

// TODO: functions on Range
```

______


runtime/interpreter/value_range.go
```diff
@@ -0,0 +1,260 @@
...
+	end IntegerValue,
+	rangeType InclusiveRangeStaticType,
+) *CompositeValue {
+	startComparable, startOk := start.(ComparableValue)
+	endInclusiveComparable, endInclusiveOk := end.(ComparableValue)
+	if !startOk || !endInclusiveOk {
+		panic(errors.NewUnreachableError())
+	}
+
+	step := GetValueForIntegerType(1, rangeType.ElementType)
```
turbolent, line 43: Given that the `GetValueForIntegerType` calls are all with fixed integers, and the set of integer types is known and small, maybe we can lazily pre-compute a lookup table?

runtime/interpreter/value_range.go
```diff
@@ -0,0 +1,260 @@
...
+			Value: step,
+		},
+	}
+
+	rangeSemaType := getInclusiveRangeSemaType(interpreter, rangeType)
+
+	rangeValue := NewCompositeValueWithStaticType(
+		interpreter,
+		locationRange,
+		sema.PublicKeyType.Location, // TODO: Figure out a location.
```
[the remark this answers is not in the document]
turbolent: For built-in types, the location is simply `nil`:
```suggestion
		nil
```

turbolent (commented):
Great work!

dsainati1 (commented):
Can you make a `feature/range-type` branch that includes all the `Range` changes, so we can put them in one place them merge them to `master` together?

darkdrag00nv2:
@dsainati1 So I don't have the access to do that. Can you please create one for me? I'll target this PR on that branch after that. Thanks!

runtime/interpreter/value_range.go
```diff
@@ -0,0 +1,260 @@
...
+	rangeType InclusiveRangeStaticType,
+	interpreter *Interpreter,
+	locationRange LocationRange,
+	needleValue Value,
+) BoolValue {
+	start := getFieldAsIntegerValue(rangeValue, interpreter, locationRange, sema.InclusiveRangeTypeStartFieldName)
+	endInclusive := getFieldAsIntegerValue(rangeValue, interpreter, locationRange, sema.InclusiveRangeTypeEndFieldName)
+	step := getFieldAsIntegerValue(rangeValue, interpreter, locationRange, sema.InclusiveRangeTypeStepFieldName)
+
+	needleInteger := convertAndAssertIntegerValue(needleValue)
```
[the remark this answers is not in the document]
darkdrag00nv2: Done.

runtime/format/range.go
```diff
@@ -0,0 +1,27 @@
...
+
+package format
+
+import (
+	"fmt"
+)
+
+func InclusiveRange(start, end string) string {
+	return fmt.Sprintf("InclusiveRange{ start: %s, end: %s }", start, end)
+}
```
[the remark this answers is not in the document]
darkdrag00nv2: oops, this formatter is no longer relevant once we moved to using `CompositeValue` to represent an `InclusiveRange`. I had written this when there was a separate `InclusiveRangeValue` for this.

Now `InclusiveRange` uses the formatting capabilities of `CompositeValue`.

runtime/interpreter/value_range.go
darkdrag00nv2: Nice idea. Done, added it as a lookup table in Interpreter. Do let me know if you think it should be stored somewhere else.

SupunS:
@darkdrag00nv2 created a new branch [feature/range-type](https://github.com/onflow/cadence/tree/feature/range-type) and changed the base to it.

runtime/format/range.go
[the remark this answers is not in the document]
turbolent: Yeah, please remove if it's dead code

runtime/interpreter/interpreter.go
```diff
@@ -242,14 +242,15 @@ type Storage interface {
...
-	interpreted  bool
+	Location            common.Location
+	statement           ast.Statement
+	Program             *Program
+	SharedState         *SharedState
+	Globals             GlobalVariables
+	activations         *VariableActivations
+	Transactions        []*HostFunctionValue
+	interpreted         bool
+	cachedIntegerValues map[StaticType]map[int64]IntegerValue
```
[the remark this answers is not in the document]
turbolent: This was in response to https://github.com/onflow/cadence/pull/2523#discussion_r1246912655. I should have maybe been clearer about the particular implementation of the lookup table.

The table should not be per interpreter, because there are many interpreters (one per location), but can be global. We can start using it here, but later also use if for other cases ("small integer optimization")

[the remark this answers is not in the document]
SupunS: If it's global, I think it would be better to have a limit on the range of integers that are memoized (e.g. precompute/memoize only if `n<10`, etc.). Otherwise (currently) it can store an infinite number of values which is a bit dangerous.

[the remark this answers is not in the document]
turbolent: Yes, it definitely should have a limit and it should be small. Currently, only 0 and 1 are used AFAICS.
I think Java uses range -128 to 127. So maybe just change the key type to `int8`?

runtime/interpreter/value_range.go
```diff
@@ -0,0 +1,208 @@
...
+	interpreter *Interpreter,
+	locationRange LocationRange,
+	start IntegerValue,
+	end IntegerValue,
+	step IntegerValue,
+	rangeType InclusiveRangeStaticType,
+) *CompositeValue {
+
+	// Validate that the step is non-zero.
+	if step.Equal(interpreter, locationRange, interpreter.GetValueForIntegerType(0, rangeType.ElementType)) {
```
SupunS, line 75: Can extract `interpreter.GetValueForIntegerType(0, rangeType.ElementType)` to a variable, and re-use it here and below.

runtime/interpreter/value_range.go
```diff
@@ -0,0 +1,208 @@
...
+	// Validate that the sequence is moving towards the end value.
+	// If start < end, step must be > 0
+	// If start > end, step must be < 0
+	// If start == end, step doesn't matter.
+	if (start.Less(interpreter, end, locationRange) && step.Less(interpreter, interpreter.GetValueForIntegerType(0, rangeType.ElementType), locationRange)) ||
+		(start.Greater(interpreter, end, locationRange) && step.Greater(interpreter, interpreter.GetValueForIntegerType(0, rangeType.ElementType), locationRange)) {
+
+		panic(InclusiveRangeConstructionError{
+			LocationRange: locationRange,
+			Message:       fmt.Sprintf("sequence is moving away from end: %s due to the value of step: %s and start: %s", end, step, start),
```
SupunS, line 91:
```suggestion
			Message:       fmt.Sprintf(
				"sequence is moving away from end: %s due to the value of step: %s and start: %s",
				 end,
				 step,
				 start,
			),
```

runtime/stdlib/range.go
```diff
@@ -0,0 +1,112 @@
...
+		locationRange := invocation.LocationRange
+
+		leftStaticType := start.StaticType(inter)
+		rightStaticType := endInclusive.StaticType(inter)
+		if leftStaticType != rightStaticType {
+			// Checker would only allow same type for both start & endInclusive.
+			panic(errors.NewUnreachableError())
+		}
+
+		rangeStaticType := interpreter.InclusiveRangeStaticType{ElementType: leftStaticType}
```
SupunS, line 99: instead of taking the statictype and then converting it sema-type inside `NewInclusiveRangeValueWithStep, maybe get the sema-type or the argument from the invocation directly (i.e: `invocation.ArgumentTypes`)

runtime/interpreter/value_range.go
```diff
@@ -0,0 +1,208 @@
...
+	interpreter *Interpreter,
+	locationRange LocationRange,
+	needleValue IntegerValue,
+) BoolValue {
+	start := getFieldAsIntegerValue(rangeValue, interpreter, locationRange, sema.InclusiveRangeTypeStartFieldName)
+	endInclusive := getFieldAsIntegerValue(rangeValue, interpreter, locationRange, sema.InclusiveRangeTypeEndFieldName)
+	step := getFieldAsIntegerValue(rangeValue, interpreter, locationRange, sema.InclusiveRangeTypeStepFieldName)
+
+	var result bool
+	result = start.Equal(interpreter, locationRange, needleValue) ||
```
SupunS, line 162:
```suggestion
	result := start.Equal(interpreter, locationRange, needleValue) ||
```

runtime/interpreter/value_range.go
```diff
@@ -0,0 +1,208 @@
...
+		endInclusive.Equal(interpreter, locationRange, needleValue)
+
+	if !result {
```
SupunS, line 165: maybe exit early here, by inverting the condtion
```go
if result {
    return TrueValue
}
```

runtime/interpreter/value_range.go
```diff
@@ -0,0 +1,208 @@
...
+			result = false
+		} else {
+			// needle is in between start and endInclusive.
+			// start + k * step should be equal to needle i.e. (needle - start) mod step == 0.
+			diff, ok := needleValue.Minus(interpreter, start, locationRange).(IntegerValue)
+			if !ok {
+				panic(errors.NewUnreachableError())
+			}
+
+			result = diff.Mod(interpreter, step, locationRange).Equal(interpreter, locationRange, interpreter.GetValueForIntegerType(0, rangeType.ElementType))
```
SupunS, line 180: Can break this long line into multiple lines:
```suggestion
			zeroValue := interpreter.GetValueForIntegerType(0, rangeType.ElementType)
			result = diff.Mod(interpreter, step, locationRange).Equal(interpreter, locationRange, zeroValue)
```

runtime/interpreter/value_range.go
```diff
@@ -0,0 +1,208 @@
...
+	startComparable, startOk := start.(ComparableValue)
+	endInclusiveComparable, endInclusiveOk := end.(ComparableValue)
+	if !startOk || !endInclusiveOk {
+		panic(errors.NewUnreachableError())
+	}
+
+	step := interpreter.GetValueForIntegerType(1, rangeType.ElementType)
+	if startComparable.Greater(interpreter, endInclusiveComparable, locationRange) {
+		elemSemaTy := interpreter.MustConvertStaticToSemaType(rangeType.ElementType)
+		if _, ok := sema.AllUnsignedIntegerTypesSet[elemSemaTy]; ok {
```
SupunS, line 46: instead of the map lookup, perhaps use the `TypeTag`.
e.g: `if elemSemaTy.Tag().BelongsTo(UnsignedIntegerTypeTag) {} `

That way, we also don't need to introduce a `AllUnsignedIntegerTypesSet`

runtime/interpreter/value_range.go
darkdrag00nv2: Done.

runtime/format/range.go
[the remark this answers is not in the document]
darkdrag00nv2: Done, removed.

runtime/convertTypes.go
```diff
@@ -62,6 +62,8 @@ func ExportMeteredType(
 			return exportInterfaceType(gauge, t, results)
 		case *sema.DictionaryType:
 			return exportDictionaryType(gauge, t, results)
+		case *sema.InclusiveRangeType:
+			return exportInclusiveRangeType(gauge, t, results)
```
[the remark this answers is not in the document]
darkdrag00nv2: I see, so I had totally missed the export part. I only implemented exporting the `InclusiveRangeType` but didn't declare a corresponding value in `values.go`. I'll declare that and then add these tests.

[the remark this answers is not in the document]
darkdrag00nv2: Done, added the support and tests.

runtime/convertTypes.go
```diff
@@ -628,6 +643,8 @@ func ImportType(memoryGauge common.MemoryGauge, t cadence.Type) interpreter.Stat
 			ImportType(memoryGauge, t.KeyType),
 			ImportType(memoryGauge, t.ElementType),
 		)
+	case *cadence.InclusiveRangeType:
+		return interpreter.NewInclusiveRangeStaticType(memoryGauge, ImportType(memoryGauge, t.ElementType))
```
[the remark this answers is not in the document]
darkdrag00nv2: Done, added the support and tests.

runtime/stdlib/range.go
```diff
@@ -0,0 +1,112 @@
...
+			panic(errors.NewUnreachableError())
+		}
+
+		inter := invocation.Interpreter
+		locationRange := invocation.LocationRange
+
+		leftStaticType := start.StaticType(inter)
+		rightStaticType := endInclusive.StaticType(inter)
+		if leftStaticType != rightStaticType {
+			// Checker would only allow same type for both start & endInclusive.
```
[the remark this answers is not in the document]
darkdrag00nv2: Added the check in the interpreter and a test case.

runtime/interpreter/value_range.go
darkdrag00nv2: Done.

darkdrag00nv2:
Incorporated half of the suggestions. A few of them are remaining.

runtime/sema/type.go
```diff
@@ -5541,6 +5820,14 @@ func checkSubTypeWithoutEquality(subType Type, superType Type) bool {
 		return IsSubType(typedSubType.KeyType, typedSuperType.KeyType) &&
 			IsSubType(typedSubType.ValueType, typedSuperType.ValueType)
 
+	// case *InclusiveRangeType:
+	// 	typedSubType, ok := subType.(*InclusiveRangeType)
+	// 	if !ok {
+	// 		return false
+	// 	}
+
+	// 	return IsSubType(typedSubType.MemberType, typedSuperType.MemberType)
```
dsainati1, line 5829: Remove commented code

runtime/interpreter/value_range.go
darkdrag00nv2: Done, using tags now and removed the set.

runtime/interpreter/decode.go
```diff
@@ -1428,6 +1428,9 @@ func (d TypeDecoder) DecodeStaticType() (StaticType, error) {
 	case CBORTagDictionaryStaticType:
 		return d.decodeDictionaryStaticType()
 
+	case CBORTagInclusiveRangeStaticType:
+		return d.decodeInclusiveRangeStaticType()
```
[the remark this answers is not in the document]
darkdrag00nv2: Done

runtime/sema/type.go
darkdrag00nv2: Done.

runtime/interpreter/interpreter.go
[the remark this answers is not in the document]
darkdrag00nv2: Done, made it global and key type to int8.

runtime/interpreter/decode.go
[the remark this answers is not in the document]
darkdrag00nv2: Done.

runtime/interpreter/statictype.go
```diff
@@ -230,6 +230,53 @@ func (t VariableSizedStaticType) Equal(other StaticType) bool {
...
+func (t InclusiveRangeStaticType) MeteredString(memoryGauge common.MemoryGauge) string {
+	common.UseMemory(memoryGauge, common.InclusiveRangeStaticTypeStringMemoryUsage)
+
+	elementStr := t.ElementType.MeteredString(memoryGauge)
+
+	return fmt.Sprintf("InclusiveRange<%s>", elementStr)
+}
+
+func (t InclusiveRangeStaticType) Equal(other StaticType) bool {
+	otherRangeType, ok := other.(InclusiveRangeStaticType)
```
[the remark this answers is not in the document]
darkdrag00nv2: Done, added in statictype_test.

runtime/stdlib/range.go
darkdrag00nv2: Done.

runtime/interpreter/value_range.go
```diff
@@ -0,0 +1,218 @@
...
+		panic(InclusiveRangeConstructionError{
+			LocationRange: locationRange,
+			Message: fmt.Sprintf(
+				"sequence is moving away from end: %s due to the value of step: %s and start: %s",
+				end,
+				step,
+				start,
+			),
+		})
+	}
```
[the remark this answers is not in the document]
darkdrag00nv2: Done, added.

runtime/sema/type_tags.go
```diff
@@ -690,6 +696,34 @@ func findSuperTypeFromUpperMask(joinedTypeTag TypeTag, types []Type) Type {
 	}
 }
 
+func commonSuperTypeOfRanges(types []Type) Type {
```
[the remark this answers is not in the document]
darkdrag00nv2: I removed this in order to model InclusiveRange around Capability which goes in a different branch. For Capability, we don't look at the inner type to find the super type. Did the same here.

darkdrag00nv2:
@SupunS @dsainati1 This is ready for another round of review. Thanks!

runtime/interpreter/interpreter.go
```diff
@@ -3535,6 +3606,36 @@ var runtimeTypeConstructors = []runtimeTypeConstructor{
...
+				if !ok {
+					panic(errors.NewUnreachableError())
+				}
+
+				ty := typeValue.Type
+				// InclusiveRanges must hold integers
+				// _, ok = ty.(Sig)
+				// if !ok {
+				// 	return Nil
+				// }
```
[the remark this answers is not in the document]
dsainati1: Regardless, there shouldn't be commented-out code in the final PR.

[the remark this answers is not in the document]
darkdrag00nv2: Agreed, sorry. I meant to ask about the code but forgot. The code is no longer commented.

turbolent (commented):
Looks good! Please add more tests that cover the added code which is reachable

runtime/interpreter/value_range.go
```diff
@@ -0,0 +1,226 @@
...
+		end.Equal(interpreter, locationRange, needleValue)
+
+	if result {
+		return TrueValue
+	}
+
+	greaterThanStart := needleValue.Greater(interpreter, start, locationRange)
+	greaterThanend := needleValue.Greater(interpreter, end, locationRange)
+
+	if greaterThanStart == greaterThanend {
```
SupunS, line 186: I was confused at first about what this condition is checking, and took me a while to realize. Maybe it would be more intuitive to have the two booleans checking `smallerThanStart` (or smallerOrEqualThanStart?) and `greaterThanEnd`, and have the condition as `if smallerThanStart || greaterThanEnd`.

runtime/sema/type_tags.go
```diff
@@ -670,7 +674,8 @@ func findSuperTypeFromUpperMask(joinedTypeTag TypeTag, types []Type) Type {
 		restrictedTypeMask,
 		transactionTypeMask,
 		interfaceTypeMask,
-		functionTypeMask:
+		functionTypeMask,
+		inclusiveRangeTypeMask:
```
SupunS, line 678: Do we want InclusiveRange to be covariant?

runtime/stdlib/range.go
```diff
@@ -0,0 +1,144 @@
...
+		locationRange := invocation.LocationRange
+
+		startStaticType := start.StaticType(inter)
+		endStaticType := end.StaticType(inter)
+		if startStaticType != endStaticType {
+			panic(interpreter.InclusiveRangeConstructionError{
+				LocationRange: locationRange,
+				Message: fmt.Sprintf("start and end are of different types. start: %s and end: %s",
+					startStaticType,
+					endStaticType),
```
SupunS, line 101:
```suggestion
				Message: fmt.Sprintf(
					"start and end are of different types. start: %s and end: %s",
					startStaticType,
					endStaticType,
				),
```

runtime/stdlib/range.go
```diff
@@ -0,0 +1,144 @@
...
+			})
+		}
+
+		rangeStaticType := interpreter.InclusiveRangeStaticType{ElementType: startStaticType}
```
SupunS, line 105: Use `NewInclusiveRangeStaticType` constructor

runtime/stdlib/range.go
```diff
@@ -0,0 +1,144 @@
...
+				panic(errors.NewUnreachableError())
+			}
+
+			stepStaticType := step.StaticType(inter)
+			if stepStaticType != startStaticType {
+				panic(interpreter.InclusiveRangeConstructionError{
+					LocationRange: locationRange,
+					Message: fmt.Sprintf("step must be of the same type as start and end. start/end: %s and step: %s",
+						startStaticType,
+						stepStaticType),
```
SupunS, line 120:
```suggestion
					Message: fmt.Sprintf(
						"step must be of the same type as start and end. start/end: %s and step: %s",
						startStaticType,
						stepStaticType,
					),
```

runtime/tests/checker/range_value_test.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+}
+
+func TestInclusiveRangeConstructionInvalid(t *testing.T) {
+	t.Parallel()
+
+	baseValueActivation := sema.NewVariableActivation(sema.BaseValueActivation)
+	baseValueActivation.DeclareValue(stdlib.InclusiveRangeConstructorFunction)
+
+	runInvalidCase := func(t *testing.T, label, code string, expectedErrorTypes []interface{}) {
```
SupunS, line 285:
```suggestion
	runInvalidCase := func(t *testing.T, label, code string, expectedErrorTypes []error) {
```

values.go
```diff
@@ -2051,6 +2051,91 @@ func (v Contract) GetFieldValues() []Value {
...
+	}
+}
+
+func NewMeteredInclusiveRange(
+	gauge common.MemoryGauge,
+	start, end, step Value,
+) (InclusiveRange, error) {
+	common.UseMemory(gauge, common.CadenceInclusiveRangeValueMemoryUsage)
+	return NewInclusiveRange(start, end, step), nil
+}
```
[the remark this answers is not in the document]
darkdrag00nv2: Yes, removed.

runtime/convertValues.go
```diff
@@ -613,6 +619,67 @@ func exportDictionaryValue(
...
+		inter,
+		startValue,
+		endValue,
+		stepValue,
+	)
+	if err != nil {
+		return cadence.InclusiveRange{}, err
+	}
+
+	t := exportInclusiveRangeType(inter, inclusiveRangeType, map[sema.TypeID]cadence.Type{}).(*cadence.InclusiveRangeType)
```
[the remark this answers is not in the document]
darkdrag00nv2: Done, we no longer cast here.

runtime/stdlib/range.go
darkdrag00nv2: Done.

runtime/interpreter/encoding_test.go
```diff
@@ -3988,6 +3988,62 @@ func TestEncodeDecodePathLinkValue(t *testing.T) {
 		)
 	})
 
+	t.Run("inclusiverange, int", func(t *testing.T) {
+
+		t.Parallel()
+
+		value := PathLinkValue{
```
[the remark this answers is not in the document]
darkdrag00nv2: cool, updated.

runtime/interpreter/value_range.go
```diff
@@ -0,0 +1,226 @@
...
+			Message:       "step value cannot be zero",
+		})
+	}
+
+	// Validate that the sequence is moving towards the end value.
+	// If start < end, step must be > 0
+	// If start > end, step must be < 0
+	// If start == end, step doesn't matter.
+	if (start.Less(interpreter, end, locationRange) && step.Less(interpreter, zeroValue, locationRange)) ||
+		(start.Greater(interpreter, end, locationRange) && step.Greater(interpreter, zeroValue, locationRange)) {
```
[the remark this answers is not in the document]
darkdrag00nv2: Done.

runtime/interpreter/value.go
```diff
@@ -15926,66 +15949,94 @@ func (v *CompositeValue) ConformsToStaticType(
...
+	} else if _, ok := staticType.(InclusiveRangeStaticType); ok {
+		inclusiveRangeType, ok := semaType.(*sema.InclusiveRangeType)
 		if !ok {
 			return false
 		}
 
-		fieldStaticType := value.StaticType(interpreter)
+		startValue := v.GetField(interpreter, locationRange, sema.InclusiveRangeTypeStartFieldName)
+		endValue := v.GetField(interpreter, locationRange, sema.InclusiveRangeTypeEndFieldName)
+		stepValue := v.GetField(interpreter, locationRange, sema.InclusiveRangeTypeStepFieldName)
```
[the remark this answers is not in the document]
darkdrag00nv2: Done, defined `sema.InclusiveRangeTypeFieldNames` and using it.

runtime/convertValues.go
```diff
@@ -382,7 +382,13 @@ func exportCompositeValue(
 
 	compositeType, ok := semaType.(*sema.CompositeType)
 	if !ok {
-		panic(errors.NewUnreachableError())
+		// InclusiveRange is stored as a CompositeValue but isn't a CompositeType.
+		inclusiveRangeType, ok := semaType.(*sema.InclusiveRangeType)
```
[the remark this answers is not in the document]
darkdrag00nv2: Done.

runtime/interpreter/value.go
```diff
@@ -15926,66 +15949,94 @@ func (v *CompositeValue) ConformsToStaticType(
...
-		member, ok := compositeType.Members.Get(fieldName)
+			if !value.ConformsToStaticType(
+				interpreter,
+				locationRange,
+				results,
+			) {
+				return false
+			}
+		}
+	} else if _, ok := staticType.(InclusiveRangeStaticType); ok {
```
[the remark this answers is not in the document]
darkdrag00nv2: done.

runtime/convertValues.go
```diff
@@ -1382,6 +1451,70 @@ func (i valueImporter) importDictionaryValue(
...
 
+func (i valueImporter) importInclusiveRangeValue(
+	v cadence.InclusiveRange,
+	expectedType sema.Type,
+) (
+	*interpreter.CompositeValue,
+	error,
+) {
+
+	inclusiveRangeType, ok := expectedType.(*sema.InclusiveRangeType)
```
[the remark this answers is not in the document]
darkdrag00nv2: Done, implemented and added test cases in `TestImportInclusiveRangeValue`.

runtime/interpreter/value_range.go
darkdrag00nv2: I don't think `if smallerThanStart || greaterThanEnd` will work since step can be negative also. So we need to check if needle is greater than one and smaller than another.

To improve the readability, I've extracted it to a separate function `isNeedleBetweenStartEndExclusive`. Hopefully it makes it a bit clearer.

runtime/sema/type_tags.go
darkdrag00nv2: What is covariant and what are the implications of it being one?

SupunS: So in this case, a covariant would mean, the `InclusiveRange<Int8>` value can be assigned to `InclusiveRange<Integer>`.

It would also mean that if there is an array with two (or more) values with different range types, say `var x = [InclusiveRange<Int8>(), InclusiveRange<Int16>()]`, the inferred type for `x` will be the covariant-super-type `InclusiveRange<Integer>`.

I feel we don't want to support this. But WDYT?
Also might be good to add some tests, to verify the behavior, if there aren't any already

encoding/json/decode.go
```diff
@@ -866,6 +872,28 @@ func (d *Decoder) decodeEnum(valueJSON any) cadence.Enum {
...
+	step := obj.GetValue(d, stepKey)
+
+	value := cadence.NewMeteredInclusiveRange(
+		d.gauge,
+		start,
+		end,
+		step,
+	)
+
+	// TODO: Ensure that start, end and step have the same type.
```
SupunS, line 889: maybe add some tests in `TestRuntimeScriptParameterTypeValidation`, for importing range values with mixed types.

runtime/convertValues.go
```diff
@@ -1382,6 +1451,90 @@ func (i valueImporter) importDictionaryValue(
...
 
+func (i valueImporter) importInclusiveRangeValue(
+	v cadence.InclusiveRange,
+	expectedType sema.Type,
+) (
+	*interpreter.CompositeValue,
+	error,
+) {
+	// start, end and step. The order matters.
+	members := make([]interpreter.IntegerValue, 3)
```
SupunS, line 1462: Can we just have three variables for `start`, `end` and `step` values (and for types)? creating multiple slices for every range value seems unnecessary.
e.g. even though a slice was created for `memberTypes`, only `memberTypes[0]` is used.

runtime/convertValues.go
```diff
@@ -1382,6 +1451,90 @@ func (i valueImporter) importDictionaryValue(
...
+
+	// import members.
+	for index, value := range []cadence.Value{v.Start, v.End, v.Step} {
+		importedValue, err := i.importValue(value, memberType)
+		if err != nil {
+			return nil, err
+		}
+		importedIntegerValue, ok := importedValue.(interpreter.IntegerValue)
+		if !ok {
+			return nil, err
```
SupunS, line 1489: here it is returning `nil` for both value and the error.

runtime/convertValues.go
```diff
@@ -1382,6 +1451,90 @@ func (i valueImporter) importDictionaryValue(
...
+	memberStaticTypes := make([]interpreter.StaticType, 3)
+
+	for i, member := range members {
+		memberStaticTypes[i] = member.StaticType(inter)
+		memberType, err := inter.ConvertStaticToSemaType(memberStaticTypes[i])
+		if err != nil {
+			return nil, err
+		}
+		memberTypes[i] = memberType
+	}
```
SupunS, line 1506: Maybe move this snippet also inside the `if inclusiveRangeType == nil {` condition (and merge it with the one below), so it's clear that this code is only needed when the expected range type is unknown

runtime/sema/type_tags.go
darkdrag00nv2: > I feel we don't want to support this. WDYT?

I am actually not sure of the pros and cons of supporting this. So it is fine for me if we don't support it.

cc: @turbolent if you have any thoughts.

encoding/json/decode.go
darkdrag00nv2: Done, added test cases and removed the TODO.

runtime/convertValues.go
darkdrag00nv2: Nice catch. Updated now.

runtime/convertValues.go
darkdrag00nv2: I've refactored the function a bit. Can you please take a look again and comment if this is still relevant.

The key point is that I need to ensure each individual field (start, end and step) has the same static type. I've added a comment in the code as to why that is necessary.

runtime/convertValues.go
darkdrag00nv2: Removed `memberTypes` and `memberStaticTypes` slices as they were not needed. Still keeping the `members` slice as it helps with the code reuse while importing members in the below snippet.

```go
	// import members.
	for index, value := range []cadence.Value{v.Start, v.End, v.Step} {
		importedValue, err := i.importValue(value, memberType)
		if err != nil {
			return nil, err
		}
		importedIntegerValue, ok := importedValue.(interpreter.IntegerValue)
		if !ok {
			return nil, errors.NewDefaultUserError(
				"cannot import inclusiverange: start, end and step must be integers",
			)
		}

		members[index] = importedIntegerValue
	}
```

runtime/sema/type_tags.go
darkdrag00nv2: > Also might be good to add some tests, to verify the behavior, if there aren't any already

I've added test cases in `TestRuntimeScriptParameterTypeValidation`.

encoding/ccf/decode_type.go
```diff
@@ -436,6 +440,46 @@ func (d *Decoder) decodeDictType(
...
+
+	// element 0: element type (inline-type or type-value)
+	elementType, err := decodeTypeFn(types)
+	if err != nil {
+		return nil, err
+	}
+
+	if elementType == nil {
+		return nil, errors.New("unexpected nil type as inclusiverange element type")
+	}
```
darkdrag00nv2, line 478: This seems like unnecessary. If I pass a nil elementType, it leads to an error in the above `decodeTypeFn` and we return in the `err != nil` check. It might still be OK to keep this check here as we have similar code in other decode methods such as for dictionary.

runtime/convertValues.go
```diff
@@ -613,6 +623,63 @@ func exportDictionaryValue(
...
+			inter,
+			locationRange,
+			seenReferences,
+		)
+	}
+
+	startValue, err := getNonComputedField(sema.InclusiveRangeTypeStartFieldName)
+	if err != nil {
+		return cadence.InclusiveRange{}, err
+	}
```
darkdrag00nv2, line 660: I am not sure how to create this scenario in the tests. We need the export of an integer value (`exportValueWithInterpreter` in above function) to fail to be able to test this failure path.

encoding/ccf/decode.go
```diff
@@ -1309,6 +1313,63 @@ func (d *Decoder) decodeEnum(typ *cadence.EnumType, types *cadenceTypeByCCFTypeI
...
+//	end: value,
+//	step: value,
+//
+// ]
+func (d *Decoder) decodeInclusiveRange(typ *cadence.InclusiveRangeType, types *cadenceTypeByCCFTypeID) (cadence.Value, error) {
+	// Decode array length.
+	n, err := d.dec.DecodeArrayHead()
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
darkdrag00nv2: Done.

runtime/interpreter/interpreter.go
```diff
@@ -3535,6 +3609,36 @@ var runtimeTypeConstructors = []runtimeTypeConstructor{
...
+				return NewSomeValueNonCopying(
+					invocation.Interpreter,
+					NewTypeValue(
+						invocation.Interpreter,
+						NewInclusiveRangeStaticType(
+							invocation.Interpreter,
+							ty,
+						),
+					),
+				)
```
SupunS, line 3638: Can you add a test to cover these lines? Creating a type-value in cadence should cover this path.

SupunS (approved):
Looks great!

SupunS:
@onflow/cadence can you please have a second round of review? 🙏

runtime/interpreter/storage_test.go
```diff
@@ -86,6 +86,62 @@ func TestCompositeStorage(t *testing.T) {
 	)
 }
 
+func TestInclusiveRangeStorage(t *testing.T) {
```
turbolent, line 89: 👌

runtime/interpreter/value_range.go
```diff
@@ -0,0 +1,269 @@
...
+		locationRange,
+		nil,
+		rangeSemaType.QualifiedString(),
+		common.CompositeKindStructure,
+		fields,
+		common.ZeroAddress,
+		rangeType,
+	)
+
+	rangeValue.Functions = map[string]FunctionValue{
```
turbolent, line 162: The functions must also be injected when the value is loaded from storage.

Could you please add a `runtime` test to ensure this works?
Basically, performs two transactions:
- Create range and save it to account storage
- Load range from storage and call `contains`

We'll need to see how we can inject it, given the value is/us stored as a `CompositeValue`

values.go
```diff
@@ -2051,6 +2051,91 @@ func (v Contract) GetFieldValues() []Value {
...
+			},
+			{
+				Identifier: sema.InclusiveRangeTypeEndFieldName,
+				Type:       v.InclusiveRangeType.ElementType,
+			},
+			{
+				Identifier: sema.InclusiveRangeTypeStepFieldName,
+				Type:       v.InclusiveRangeType.ElementType,
+			},
+		}
```
turbolent, line 2129:
```suggestion
		elementType := v.InclusiveRangeType.ElementType
		v.fields = []Field{
			{
				Identifier: sema.InclusiveRangeTypeStartFieldName,
				Type:       elementType,
			},
			{
				Identifier: sema.InclusiveRangeTypeEndFieldName,
				Type:       velementType,
			},
			{
				Identifier: sema.InclusiveRangeTypeStepFieldName,
				Type:       elementType,
			},
		}
```

types.go
```diff
@@ -1059,6 +1059,52 @@ func (t *DictionaryType) Equal(other Type) bool {
...
+	elementType Type,
+) *InclusiveRangeType {
+	common.UseMemory(gauge, common.CadenceInclusiveRangeTypeMemoryUsage)
+	return NewInclusiveRangeType(elementType)
+}
+
+func (*InclusiveRangeType) isType() {}
+
+func (t *InclusiveRangeType) ID() string {
+	if len(t.typeID) == 0 {
```
turbolent, line 1090:
```suggestion
	if t.typeID == "" {
```

runtime/tests/checker/range_value_test.go
```diff
@@ -0,0 +1,365 @@
...
+	"github.com/onflow/cadence/runtime/sema"
+	"github.com/onflow/cadence/runtime/stdlib"
+)
+
+type inclusiveRangeConstructionTest struct {
+	ty         sema.Type
+	s, e, step int64
+}
+
+func TestInclusiveRangeConstructionValid(t *testing.T) {
```
turbolent, line 37:
```suggestion
func TestCheckInclusiveRangeConstructionValid(t *testing.T) {
```

runtime/tests/checker/range_value_test.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+
+	// Run each test case with and without step.
+	for _, testCase := range validTestCases {
+		runValidCase(t, testCase, true)
+		runValidCase(t, testCase, false)
+	}
+}
+
+func TestInclusiveRangeConstructionInvalid(t *testing.T) {
```
turbolent, line 279:
```suggestion
func TestCheckInclusiveRangeConstructionInvalid(t *testing.T) {
```

runtime/stdlib/range.go
```diff
@@ -0,0 +1,148 @@
...
+		if !startOk || !endOk {
+			panic(errors.NewUnreachableError())
+		}
+
+		inter := invocation.Interpreter
+		locationRange := invocation.LocationRange
+
+		startStaticType := start.StaticType(inter)
+		endStaticType := end.StaticType(inter)
+		if startStaticType != endStaticType {
```
turbolent, line 96:
```suggestion
		if !startStaticType.Equal(endStaticType) {
```

runtime/stdlib/range.go
```diff
@@ -0,0 +1,148 @@
...
+				Identifier:     "step",
+				TypeAnnotation: typeAnnotation,
+			},
+		},
+		ReturnTypeAnnotation: sema.NewTypeAnnotation(
+			&sema.InclusiveRangeType{
+				MemberType: typeAnnotation.Type,
+			},
+		),
+		RequiredArgumentCount: sema.RequiredArgumentCount(2),
```
turbolent, line 75: On `master` this has changed:
```suggestion
		// `step` parameter is optional
		Arity: &sema.Arity{Min: 2, Max: 3},
```

runtime/sema/type.go
```diff
@@ -5021,6 +5022,286 @@ func (t *DictionaryType) Resolve(typeArguments *TypeParameterTypeOrderedMap) Typ
...
+	r.memberResolversOnce.Do(func() {
+		r.memberResolvers = withBuiltinMembers(r, map[string]MemberResolver{
+			InclusiveRangeTypeStartFieldName: {
+				Kind: common.DeclarationKindField,
+				Resolve: func(memoryGauge common.MemoryGauge, identifier string, _ ast.Range, _ func(error)) *Member {
+					return NewPublicConstantFieldMember(
+						memoryGauge,
+						r,
+						identifier,
+						r.ElementType(false),
```
turbolent, line 5225: Here and for the other fields:

```suggestion
						r.MemberType,
```

runtime/sema/type.go
```diff
@@ -5021,6 +5022,286 @@ func (t *DictionaryType) Resolve(typeArguments *TypeParameterTypeOrderedMap) Typ
...
+}
+
+func (r *InclusiveRangeType) Equal(other Type) bool {
+	otherRange, ok := other.(*InclusiveRangeType)
+	if !ok {
+		return false
+	}
+	if otherRange.MemberType == nil {
+		return r.MemberType == nil
+	}
```
turbolent, line 5081: When would the `MemberType` be `nil`?

runtime/sema/type.go
```diff
@@ -5021,6 +5022,286 @@ func (t *DictionaryType) Resolve(typeArguments *TypeParameterTypeOrderedMap) Typ
...
+
+func (*InclusiveRangeType) Tag() TypeTag {
+	return InclusiveRangeTypeTag
+}
+
+func (r *InclusiveRangeType) String() string {
+	memberString := ""
+	if r.MemberType != nil {
+		memberString = fmt.Sprintf("<%s>", r.MemberType.String())
+	}
```
turbolent, line 5053: When would the `MemberType` be `nil`?

turbolent (commented):
Great work and progress!

We'll still need to address some of the encoding and storage-related parts, but overall this is close to being done 👏

runtime/stdlib/range.go
darkdrag00nv2: It won't build unless we merge the latest master in `feature/range-type`. Can you do that please? I'll rebase on top of it and then update it here.

encoding/ccf/decode.go
```diff
@@ -1309,6 +1313,60 @@ func (d *Decoder) decodeEnum(typ *cadence.EnumType, types *cadenceTypeByCCFTypeI
...
+
+	if n != 3 {
+		return nil, fmt.Errorf(
+			"encoded inclusiverange-value has %d elements (expected %d elements)",
+			n,
+			3,
+		)
+	}
```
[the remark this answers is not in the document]
darkdrag00nv2: Done.

encoding/ccf/decode_type.go
```diff
@@ -436,6 +440,46 @@ func (d *Decoder) decodeDictType(
...
 }
 
+// decodeInclusiveRangeType decodes inclusiverange-type or inclusiverange-type-value as
+// language=CDDL
+// inclusiverange-type =
+//
+//	; cbor-tag-inclusiverange-type
+//	#6.145([
+//	    element-type: inline-type
+//	])
```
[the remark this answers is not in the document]
darkdrag00nv2: Done.

encoding/ccf/encode.go
```diff
@@ -1411,6 +1447,24 @@ func (e *Encoder) encodeDictTypeValue(typ *cadence.DictionaryType, visited ccfTy
...
 }
 
+// encodeInclusiveRangeTypeValue encodes cadence.InclusiveRangeType as
+// language=CDDL
+// inclusiverange-type-value =
+//
+//	; cbor-tag-inclusiverange-type-value
+//	#6.194([
+//	    element-type: type-value
+//	])
```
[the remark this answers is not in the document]
darkdrag00nv2: Done.

encoding/ccf/encode_type.go
```diff
@@ -296,6 +300,51 @@ func (e *Encoder) encodeDictTypeWithRawTag(
...
 }
 
+// encodeInclusiveRangeType encodes cadence.InclusiveRangeType as
+// language=CDDL
+// inclusiverange-type =
+//
+//	; cbor-tag-inclusiverange-type
+//	#6.145([
+//	    element-type: inline-type
+//	])
```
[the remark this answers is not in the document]
darkdrag00nv2: Done

runtime/sema/type.go
darkdrag00nv2: Done.

runtime/interpreter/interpreter.go
```diff
@@ -2337,6 +2342,75 @@ func (interpreter *Interpreter) WriteStored(
 	return accountStorage.WriteValue(interpreter, key, value)
 }
 
+// Get the provided int64 value in the required staticType.
+// Note: Assumes that the provided value fits within the constraints of the staticType.
+func GetValueForIntegerType(value int8, staticType StaticType) IntegerValue {
```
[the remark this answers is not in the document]
darkdrag00nv2: Done, added `TestGetValueForIntegerType`.

runtime/interpreter/interpreter.go
darkdrag00nv2: Added `TestInterpretInclusiveRangeType` and `TestCheckInclusiveRangeTypeConstructor`.

runtime/sema/type.go
darkdrag00nv2: I do not have a definitive answer to this but I'll explain why I added it.

The `InclusiveRangeType` must be declared in the `BaseTypeActivation` in the `init` function of `runtime/sema/type.go`. Since we don't know the inner type, we leave it as nil. `CapabilityType` is implemented the same way.

If I don't declare it in `BaseTypeActivation` then code such as below fails with "error: cannot find type in this scope: InclusiveRange".
```cadence
let r: InclusiveRange<Int> = InclusiveRange(s, e, step: step)
```

Please let me know if there is a better way.

runtime/sema/type.go
darkdrag00nv2: Same explanation as above.

runtime/stdlib/range.go
SupunS: Should be able to take a pull once https://github.com/onflow/cadence/pull/2747 has emerged.

darkdrag00nv2: Pulled and updated now.

runtime/interpreter/value_range.go
darkdrag00nv2: It works. Added a case in `TestInterpretAuthAccount_load`.

darkdrag00nv2:
@turbolent Gentle ping for review.

turbolent (commented):
Thank you for the nudge and for you patience a lot of other things are getting in the way and it's a lot to go through

runtime/interpreter/interpreter.go
```diff
@@ -3551,6 +3624,36 @@ var runtimeTypeConstructors = []runtimeTypeConstructor{
 			},
 		),
 	},
+	{
+		name: "InclusiveRangeType",
+		converter: NewUnmeteredHostFunctionValue(
+			sema.InclusiveRangeTypeFunctionType,
+			func(invocation Invocation) Value {
+				typeValue, ok := invocation.Arguments[0].(TypeValue)
```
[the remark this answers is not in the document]
darkdrag00nv2: Done.

runtime/interpreter/interpreter.go
```diff
@@ -261,10 +261,14 @@ var _ ast.ExpressionVisitor[Value] = &Interpreter{}
...
 var BaseActivation *VariableActivation
 
+var cachedIntegerValues map[StaticType]map[int8]IntegerValue
+
 func init() {
 	// No need to meter since this is only created once
 	BaseActivation = activations.NewActivation[*Variable](nil, nil)
 	defineBaseFunctions(BaseActivation)
+
+	cachedIntegerValues = make(map[StaticType]map[int8]IntegerValue)
```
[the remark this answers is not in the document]
darkdrag00nv2: Done, added an `integer.go` file and moved the logic in that.

values.go
```diff
@@ -2051,6 +2051,92 @@ func (v Contract) GetFieldValues() []Value {
...
+}
+
+func (v InclusiveRange) String() string {
+	if v.InclusiveRangeType == nil {
+		return ""
+	}
+
+	if v.fields == nil {
+		elementType := v.InclusiveRangeType.ElementType
+		v.fields = []Field{
```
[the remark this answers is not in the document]
darkdrag00nv2: Good catch! Made it a pointer type.

runtime/interpreter/interpreter.go
```diff
@@ -2353,6 +2357,75 @@ func (interpreter *Interpreter) WriteStored(
...
+	val := getValueForIntegerType(value, staticType)
+	if !typedOk {
+		cachedIntegerValues[staticType] = make(map[int8]IntegerValue)
+	}
+	cachedIntegerValues[staticType][value] = val
+	return val
+}
+
+func getValueForIntegerType(value int8, staticType StaticType) IntegerValue {
+	switch staticType {
```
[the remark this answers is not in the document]
darkdrag00nv2: Good point, added the comment.

runtime/sema/type.go
darkdrag00nv2: Good point, I've updated `InclusiveRangeType` to handle the case of MemberType as nil in 1f96574c0.

Looking into the `checkParameterizedTypeIsInstantiated` suggestion.

darkdrag00nv2: > We probably have to discuss this a bit more, but I think we can do this in checkTypeAnnotation, which e.g. already checks if an interface type is used as-is (i.e. calls checkInvalidInterfaceAsType). A new function "checkParameterizedTypeIsInstantiated" could check that the elements of TypeArguments() for all ParameterizedType is non-nil for all TypeParameters() that are not Optional.

Seems like we already have this check in place. I added a test case in e6eeaed9d to use InclusiveRange without providing a type argument and received a `InvalidTypeArgumentCountError`.

runtime/interpreter/interpreter.go
```diff
@@ -3551,6 +3624,36 @@ var runtimeTypeConstructors = []runtimeTypeConstructor{
...
+			func(invocation Invocation) Value {
+				typeValue, ok := invocation.Arguments[0].(TypeValue)
+				if !ok {
+					panic(errors.NewUnreachableError())
+				}
+
+				ty := typeValue.Type
+				// InclusiveRanges must hold integers
+				elemSemaTy := invocation.Interpreter.MustConvertStaticToSemaType(ty)
+				if !elemSemaTy.Tag().BelongsTo(sema.IntegerTypeTag) {
```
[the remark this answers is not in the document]
darkdrag00nv2: I didn't intend this to be an optimization. I did this because I thought this is the only way to do the check. I am doing something similar in `value_range.go` to detect if an integer is `unsigned` based on the suggestion from Supun here: https://github.com/onflow/cadence/pull/2523#discussion_r1256520701

Do you recommend using `sema.IsSubType` instead?

darkdrag00nv2:
@turbolent @SupunS 

Since we're using a global cache (`cachedIntegerValues`), parallel test execution leads to data race. I've fixed that by using a wrapper function `synchronizedGetValueForIntegerType` with a mutex in the `range_value_test.go`.  See https://github.com/onflow/cadence/pull/2523/commits/5f2ddd8165bb38254b50b7025a489563451e7069

Not sure if there is another way to do this.

SupunS:
@darkdrag00nv2 using a Mutex seems fine, just wondering if we need to do this for the original function itself, rather than only in tests. @turbolent Do you have any idea whether value creation could run in parallel at any point? e.g. Could decoders (i.e: atree decoder) run in parallel?

runtime/sema/type.go
turbolent: Ah, nice. Yeah, this handles the case of an incorrect, number of type arguments, but only if an instantiation is used (`ast.InstantiationType`). I don't think it handles the case where the generic type is used on its own yet (bare `ast.NominalType`, not wrapped in an `ast.InstantiationType`).

runtime/interpreter/interpreter.go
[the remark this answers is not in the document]
SupunS: We might be able to use `IsSameTypeKind` (probably not `IsSubType`, because `Void` is a subtype of every type).

In the `value_range.go` use-case, it checks if a type is an unsigned integer, but there is no equivalent sema-type (i.e: no `UnsignedIntegerType`), so using `IsSameTypeKind`/`IsSubType` in that case wasn't really possible. That was kind of a special case where the tag had to be used, or rather, using the tag was convenient.

However, here since we are checking whether the type is a subtype of `Integer` type, and given there is a `sema.IntegerType`, maybe it would be better to use `IsSameTypeKind` for consistency (all other places use IsSameTypeKind/IsSubType for sub-type checking).

runtime/sema/type.go
darkdrag00nv2: Makes sense. Added `checkParameterizedTypeIsInstantiated` in d97eef694.

runtime/sema/checker.go
```diff
@@ -2067,6 +2068,49 @@ func (checker *Checker) checkInvalidInterfaceAsType(ty Type, pos ast.HasPosition
 	}
 }
 
+func (checker *Checker) checkParameterizedTypeIsInstantiated(ty Type, pos ast.HasPosition) {
```
turbolent, line 2071: Nice! 👏

runtime/tests/checker/range_value_test.go
```diff
@@ -367,6 +367,13 @@ func TestCheckInclusiveRangeConstructionInvalid(t *testing.T) {
...
 		"without_inner_type_annotation",
 		"let r: InclusiveRange<> = InclusiveRange(1, 10)",
-		[]error{&sema.InvalidTypeArgumentCountError{}},
+		[]error{&sema.InvalidTypeArgumentCountError{}, &sema.MissingTypeArgumentError{}},
+	)
+
+	runInvalidCase(
+		t,
+		"without_inner_type_annotation",
+		"let r: InclusiveRange = InclusiveRange(1, 10)",
```
turbolent, line 376: Can you please also add a general / top-level checker test case for testing `Checker.checkParameterizedTypeIsInstantiated`? 
That way we don't miss testing this general feature if we ever make changes to the range types/API.

turbolent (commented):
Looking great, awesome work! I'll do a final review early next week. 

Could you please have a look if CI reports imported new code that is not yet covered by tests and add some tests for them.

runtime/interpreter/interpreter.go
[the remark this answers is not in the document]
darkdrag00nv2: Thanks Supun and Bastian. Updated to use `IsSameTypeKind` in 7904b9544.

runtime/tests/checker/range_value_test.go
darkdrag00nv2: Added in 8732e3dee.

darkdrag00nv2:
> Could you please have a look if CI reports imported new code that is not yet covered by tests and add some tests for them.

@turbolent CI isn't reporting code coverage for some reason. https://github.com/onflow/cadence/pull/2523#issuecomment-1595811408

turbolent:
It looks like all outstanding feedback is addressed now (from what I can see), except for maybe improving the integer value caching (but that can be addresses in a follow up PR), and ensuring the new code is covered by tests. I'll look into why code coverage is broken and a workaround

turbolent:
I don't quite understand why `feature/range-type` has no coverage info.

However, I was able to still get the necessary information, the coverage for the diff between this PR and the target branch locally:

- Install the [gocover](https://github.com/Azure/gocover) tool: 
  ```sh
  go install github.com/Azure/gocover@latest
  ```
- Get full coverage:
  ```sh
  go test ./... -coverprofile=coverage.out
  ```
- Generate report for diff:
  ```sh
  gocover diff --repository-path=`pwd`  --cover-profile=`pwd`/coverage.out --compare-branch=feature/range-type
  ```
- Open coverage.html

SupunS:
+1 for improving the caching in a separate PR. Given this one targets a feature branch, we can always do improvements iteratively in follow-up PRs.

darkdrag00nv2:
May be I'm doing something wrong but the coverage results seem to funky in a few cases.

image

This is for the InclusiveRange constructor function which is definitely wrong since we have tests in `runtime/tests/interpreter/range_value_test.go`.

I generated it via the following command:

```
go test ./... -coverprofile=coverage.out

gocover diff --repository-path=`pwd`  --cover-profile=`pwd`/coverage.out --compare-branch=upstream/feature/range-type
```

darkdrag00nv2:
@turbolent Does your report also have these lines shown as uncovered?

At a high level, I didn't see any major uncovered test cases but hard to tell due to the inaccuracy of the report. I guess we could merge this one and hopefully that will upload a coverage report which we could use to add missing test cases in future PRs. Thoughts?

turbolent:
@darkdrag00nv2 Yeah, that coverage looks indeed wrong. Having a look into how to get Codecov data

turbolent:
Updated the feature branch, which successfully uploaded the coverage data: https://app.codecov.io/github/onflow/cadence/commit/edf547b1c124430b62b91946c5e453bb7fc65180

turbolent:
The small integer cache was still not thread-safe and caused data races to be reported (https://github.com/onflow/cadence/actions/runs/6398301138/job/17368137493?pr=2523#step:6:58), so I improved it in [1af52af](https://github.com/onflow/cadence/pull/2523/commits/1af52af7eb1468c0f40bcdc2a5f00e11af5aa24a)

runtime/interpreter/value.go
```diff
@@ -17045,67 +17076,102 @@ func (v *CompositeValue) ConformsToStaticType(
...
-
+	staticType := v.StaticType(interpreter)
 	semaType := interpreter.MustConvertStaticToSemaType(staticType)
 
-	compositeType, ok := semaType.(*sema.CompositeType)
-	if !ok ||
-		v.Kind != compositeType.Kind ||
-		v.TypeID() != compositeType.ID() {
+	switch staticType.(type) {
+	case CompositeStaticType:
```
turbolent, line 17083: Maybe separate the blocks of the two cases out into separate functions

turbolent (commented):
Coverage is updated now and should be helpful.

runtime/interpreter/value_range.go
turbolent: The added testcase isn't quite testing that yet, see https://github.com/onflow/cadence/pull/2523/files#r1344848593. 

Add a runtime test with two transactions. It will likely fail, because the function isn't linked in after decoding from storage (or I can't quite see how it is at the moment)

runtime/tests/interpreter/account_test.go
```diff
@@ -501,6 +503,56 @@ func TestInterpretAuthAccount_load(t *testing.T) {
...
+				fun save() {
+					let ir = InclusiveRange(10, 20)
+					account.save(ir, to: /storage/ir)
+				}
+
+				fun loadIRAndCallContains(): InclusiveRange<Int>? {
+					let ir = account.load<InclusiveRange<Int>>(from: /storage/ir)
+					let containsFifteen = ir!.contains(15)
+					return ir
+				}
```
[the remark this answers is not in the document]
darkdrag00nv2: I am facing a strange issue while adding a new runtime test to save and load inclusive range in two different transactions.

https://github.com/onflow/cadence/pull/2523/commits/dd77fb755cb101d46462a4d1b655013aa10daf3c -- The save of InclusiveRange in the first transaction fails with error

```
Execution failed:
        	            	error: failed to encode map value: cannot store non-storable value
        	            	--> 0100000000000000000000000000000000000000000000000000000000000000
```

Its strange because the tests I added in https://github.com/onflow/cadence/commit/a4d4c96d888de388dd735d96ae74c16148cb1c46 works. So not sure what the difference is between these two cases.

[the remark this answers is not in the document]
darkdrag00nv2: Per discussion on Discord, we'll add support for storing InclusiveRange in account storage in a future PR. For now, I've added a test case which indicates that this is an unsupported feature.

runtime/runtime_test.go
```diff
@@ -2365,6 +2365,62 @@ func TestRuntimeCompositeFunctionInvocationFromImportingProgram(t *testing.T) {
...
+		},
+		Context{
+			Interface: runtimeInterface,
+			Location:  nextTransactionLocation(),
+		},
+	)
+	RequireError(t, err)
+
+	var subErr *atree.ExternalError
+	require.ErrorAs(t, err, &subErr)
```
turbolent, line 2421: Assert that the reported error is for the non-storable value:

```suggestion
	var nonStorableValueErr interpreter.NonStorableValueError
	require.ErrorAs(t, err, &nonStorableValueErr)
```

This error is reported quite "late", i.e. at run-time, so might surprise developers.

Maybe consider making also the type non-storable (for now), by changing `InclusiveRangeType.IsStorable()` to return `false`.

turbolent (approved):
Great work! 👏👏 

I think we should get this merged in the current state, and then then potentially remaining work (e.g. making ranges storable) can be addressed in a follow-up PR.

Could you please have a final look @SupunS @dsainati1 ?

runtime/runtime_test.go
darkdrag00nv2: Updated to make the type non-storable. Now we get a checker `TypeMismatchError`, so I am asserting that - db22c2cc6

dsainati1 (approved):

runtime/program_params_validation_test.go
```diff
@@ -285,6 +285,99 @@ func TestRuntimeScriptParameterTypeValidation(t *testing.T) {
...
+
+		assert.NoError(t, err)
+	})
+
+	// Since InclusiveRange isn't covariant.
+	t.Run("Invalid InclusiveRange<Integer>", func(t *testing.T) {
+		t.Parallel()
+
+		script := `
+			pub fun main(arg: InclusiveRange<Integer>) {
```
SupunS, line 331: Sorry I forgot some of the details, is it possible to create an `InclusiveRange<Integer>` in cadence?
i.e: Is following allowed?
```cadence
let a: Integer = 0
let b: Integer = 10
var range: InclusiveRange<Integer> = InclusiveRange(a, b)
```

runtime/sema/type.go
```diff
@@ -3411,6 +3411,7 @@ func init() {
 			HashAlgorithmType,
 			StorageCapabilityControllerType,
 			AccountCapabilityControllerType,
+			&InclusiveRangeType{},
```
SupunS, line 3414: I guess this is not needed here, because `InclusiveRange` is a dynamically created type (just like arrays, dictionaries, etc.)

SupunS (approved):
Should be good to merge in the current state 👍 

@darkdrag00nv2 Can you maybe create separate issues for the remaining work / tech-debt, so they won't fall through the cracks

runtime/program_params_validation_test.go
darkdrag00nv2: Good question.

Yes, the code works but I think it should not work. 

I disallowed it during import by checking it in `InclusiveRangeStaticTypeConformsToStaticType` (Called by `ConformsToStaticType` for `CompositeValue`)

I think I can do the same in the InclusiveRange constructor as well. WDYT?

runtime/interpreter/value_range.go
darkdrag00nv2: Sorry, just forgot to put the final comment in this thread. We can't do this since InclusiveRange cannot be stored in account storage.

Filed https://github.com/onflow/cadence/issues/2870 for this.

runtime/sema/type.go
darkdrag00nv2: It is needed right. If I don't declare it, you can't refer to it in the code.

```
let a : InclusiveRange<Int> = InclusiveRange(1, 2)
```

It is different from arrays, dictionary because you don't refer to it in the code.

```
let a : Dictionary<Int, String> = ...
```

darkdrag00nv2:
@SupunS 

> Can you maybe create separate issues for the remaining work / tech-debt, so they won't fall through the cracks

Filed https://github.com/onflow/cadence/issues/2871 and https://github.com/onflow/cadence/issues/2870.

runtime/sema/type.go
SupunS: ah ok. I think then the `checkParameterizedTypeIsInstantiated` must be recursive.
i.e: It should also look for the types nested within other types.

For e.g: An array of `InclusiveRange` (without the type parameter), can be defined without any errors.
```cadence
let r: [InclusiveRange] = []    // no errors
```

SupunS: Maybe fix that in a follow-up PR - would make it easy to review (this PR is already big)

runtime/program_params_validation_test.go
SupunS: It seems not just the constructor, the checker also allows covariant subtyping:
e.g:
```cadence
let a: InclusiveRange<Int> = InclusiveRange(0, 10)
let b: InclusiveRange<Integer> = a
```
This would need to be fixed in the `IsSubType()` method, which would fix it universally (won't need to special case in multiple places)

SupunS: While testing, I also ran into the following cases that currently report errors, but shouldn't:

1) The inclusive range constructor always takes `Int` as the type parameter's type.
```cadence
let r = InclusiveRange<Int>(0, 10)    // Works

let r = InclusiveRange<Int8>(0, 10)    // Checker error (should also work)
```

2) Also probably related to above, Type inferring also doesn't infer the type parameter from LHS.
```cadence
let b: InclusiveRange<Int8> = InclusiveRange(0, 10)   // Type-mismatch error
```

SupunS (approved):

darkdrag00nv2:
@SupunS Lets merge? I'll send follow-up PRs post that on the fixes and future features such as `for-in` loop support.

SupunS:
@darkdrag00nv2 Can you please create issues for any remaining open discussions (e.g: https://github.com/onflow/cadence/pull/2523#discussion_r1358641806, https://github.com/onflow/cadence/pull/2523#discussion_r1358686069) and resolve those? Then we can merge the PR.

PR cannot be merged until all conversations are resolved (there is a GitHub PR restriction that is in place for this repo)

darkdrag00nv2:
@SupunS Thanks, Filed all the issues.

PR #2531, opened 2023-06-05
enhance ToGoValue() in cadence.Value

bjartek:
## Description

Started on suggesting new ToGo() values 
______

### DIFF
--- a/values.go
+++ b/values.go
@@ -1632,7 +1632,7 @@ func (v Dictionary) WithType(dictionaryType *DictionaryType) Dictionary {
 }
 
 func (v Dictionary) ToGoValue() any {
-	ret := map[string]any{}
+	ret := map[any]any{}
 
 	for _, p := range v.Pairs {
 		value := p.Value.ToGoValue()
--- a/values.go
+++ b/values.go
@@ -1565,12 +1565,12 @@ func (v Array) WithType(arrayType ArrayType) Array {
 }
 
 func (v Array) ToGoValue() any {
-	ret := make([]any, len(v.Values))
+	ret := []any{}
 
-	for i, e := range v.Values {
+	for _, e := range v.Values {
 		value := e.ToGoValue()
 		if value != nil {
-			ret[i] = value
+			ret = append(ret, value)
 		}
 	}
 
@@ -1735,10 +1735,11 @@ func (v Struct) WithType(typ *StructType) Struct {
 }
 
 func (v Struct) ToGoValue() any {
-	ret := make([]any, len(v.Fields))
+	ret := map[string]any{}
 
 	for i, field := range v.Fields {
-		ret[i] = field.ToGoValue()
+		key := v.StructType.Fields[i].Identifier
+		ret[key] = field.ToGoValue()
 	}
 
 	return ret
@@ -1818,10 +1819,11 @@ func (v Resource) WithType(typ *ResourceType) Resource {
 }
 
 func (v Resource) ToGoValue() any {
-	ret := make([]any, len(v.Fields))
+	ret := map[string]any{}
 
 	for i, field := range v.Fields {
-		ret[i] = field.ToGoValue()
+		key := v.ResourceType.Fields[i].Identifier
+		ret[key] = field.ToGoValue()
 	}
 
 	return ret
@@ -1880,10 +1882,11 @@ func (v Attachment) WithType(typ *AttachmentType) Attachment {
 }
 
 func (v Attachment) ToGoValue() any {
-	ret := make([]any, len(v.Fields))
+	ret := map[string]any{}
 
 	for i, field := range v.Fields {
-		ret[i] = field.ToGoValue()
+		key := v.AttachmentType.Fields[i].Identifier
+		ret[key] = field.ToGoValue()
 	}
 
 	return ret
@@ -1942,14 +1945,16 @@ func (v Event) WithType(typ *EventType) Event {
 }
 
 func (v Event) ToGoValue() any {
-	ret := make([]any, len(v.Fields))
+	ret := map[string]any{}
 
 	for i, field := range v.Fields {
-		ret[i] = field.ToGoValue()
+		key := v.EventType.Fields[i].Identifier
+		ret[key] = field.ToGoValue()
 	}
 
 	return ret
 }
+
 func (v Event) String() string {
 	return formatComposite(v.EventType.ID(), v.EventType.Fields, v.Fields)
 }
@@ -2003,10 +2008,11 @@ func (v Contract) WithType(typ *ContractType) Contract {
 }
 
 func (v Contract) ToGoValue() any {
-	ret := make([]any, len(v.Fields))
+	ret := map[string]any{}
 
 	for i, field := range v.Fields {
-		ret[i] = field.ToGoValue()
+		key := v.ContractType.Fields[i].Identifier
+		ret[key] = field.ToGoValue()
 	}
 
 	return ret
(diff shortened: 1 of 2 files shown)


values.go
```diff
@@ -214,7 +215,11 @@ func (v String) MeteredType(common.MemoryGauge) Type {
 }
 
 func (v String) ToGoValue() any {
-	return string(v)
+	value := string(v)
+	if value == "" {
+		return nil
```
turbolent, line 220: Why return `nil` here for an empty string?

values.go
```diff
@@ -333,7 +338,7 @@ func (Address) MeteredType(common.MemoryGauge) Type {
 }
 
 func (v Address) ToGoValue() any {
-	return [AddressLength]byte(v)
+	return v.String()
```
turbolent, line 341: Why change to a string here? A Go byte array seems like a better mapping

values.go
```diff
@@ -1419,7 +1424,11 @@ func (v Fix64) MeteredType(common.MemoryGauge) Type {
 }
 
 func (v Fix64) ToGoValue() any {
-	return int64(v)
+	value, err := strconv.ParseFloat(v.String(), 64)
+	if err != nil {
+		panic(err)
+	}
+	return value
```
turbolent, line 1431: Fixed point numbers are not floating point numbers. This potentially a lossy conversion, as some fixed-point values cannot be represented as floating point numbers.

values.go
```diff
@@ -1552,12 +1565,18 @@ func (v Array) WithType(arrayType ArrayType) Array {
...
+	for _, e := range v.Values {
+		value := e.ToGoValue()
+		if value != nil {
+			ret = append(ret, value)
+		}
 	}
 
+	if len(ret) == 0 {
+		return nil
+	}
```
turbolent, line 1579: Maybe move this to the top of the function, before unnecessarily allocating an empty slice

values.go
```diff
@@ -1617,12 +1636,18 @@ func (v Dictionary) WithType(dictionaryType *DictionaryType) Dictionary {
 }
 
 func (v Dictionary) ToGoValue() any {
-	ret := map[any]any{}
+	ret := map[string]any{}
```
turbolent, line 1639: Why change the key type to always being a string? What if the key type of the dictionary is e.g. `Int`?

values_test.go
```diff
@@ -59,154 +60,196 @@ func newValueTestCases() map[string]valueTestCase {
...
 		"Character": {
 			value:        Character("✌️"),
 			string:       "\"\\u{270c}\\u{fe0f}\"",
+			goValue:      "✌️",
 			expectedType: CharacterType{},
 		},
-		"Array": {
+		"StringArray": {
 			value: NewArray([]Value{
-				NewInt(10),
```
turbolent, line 201: Instead of changing the existing value, add an additional one

values.go
bjartek: I guess this is just a leftover from my blindly porting over the overflow code. The reasoning there was to make the output as terse as possible.

I can change this all the places here because I think that in cadence we should not make this distinction.

values.go
bjartek: ok

values.go
bjartek: Another thing left over from overflow. Will fix.

values.go
bjartek: Interesting, what situations would that occur in?

values.go
```diff
@@ -1493,7 +1502,11 @@ func (v UFix64) MeteredType(common.MemoryGauge) Type {
 }
 
 func (v UFix64) ToGoValue() any {
-	return uint64(v)
+	value, err := strconv.ParseFloat(v.String(), 64)
```
bjartek, line 1505: But here it is OK to use ParseFloat or can we not do that as per the comment in Fix64?

values.go
bjartek: If we want to skip nil values in the list we cant do this. Do we want to do that or not? I personally prefer it, but if you do not want it I can change it around.

values_test.go
bjartek: The old test has an array that has both and int and a string, I can always add it back i just though it was very strange.

bjartek:
Sorry for the delay here, I added some comments and fixed some things!

bjartek:
Status?

turbolent:
@SupunS Could you please own this?

SupunS:
Yeah, can do. I'll have a look later today.

In the meantime, were there any particular use-case(s) we wanted to enhance, or is this just a general improvement?

SupunS:
Discussion around the use-cases: https://discord.com/channels/613813861610684416/1108479699732152503/1146911629481619606

turbolent:
Given there are many different use cases and needs, and there is no "correct" Go type to return in all cases.
Developers might expect a certain Go type to be returned, but the return type is `any`, so it makes the API vry brittle.

It might be a good idea to "remove" this method from `Value`, and instead:

- Have developers define/use their own
- Provide multiple implementations of the method as stand-alone functions. For example, we already have well-defined conversion functions, like `ByteArrayValueToByteSlice` ("Cadence byte array `[UInt8]` to Go byte slice `[]byte`").
  We could extract/refactor the existing methods in a similar way, and have this PR add additional conversion functions.

How does that sound @bjartek?

turbolent:
`ToGoValue` got removed in #3291. As mentioned above, maybe the removed functions and the new variants of the functions in this PR can be converted to co-existing global functions that convert to concrete return types (instead of `any` with `ToGoValue`).

bjartek:
I dont have much time for this atm :( a lot to do.

turbolent:
> I dont have much time for this atm :( a lot to do.

No worries! We can pick this up after shipping Cadence 1.0, thank you for these contributions and your patience with getting them in

bjartek:
When will the removal of ToGoValue() hit a network? I think i use that in overflow/underflow.

turbolent:
> When will the removal of ToGoValue() hit a network? 

It's not affecting networks, only code that imports Cadence / the `cadence` package. So basically, as soon as that code updates to the latest version.

turbolent:
@bjartek We only have FLIPs for the language, not the implementation. The fact that we expose some functionality from this implementation of the language is just a side-effect, there are no guarantees about what is exposed from it. 

In any case, the functionality got removed from the package, but it can be easily provided by third-party code (e.g. another open-source Go module, or directly as part of an application), none of the underlying functionality needed for it is gone.

bjartek:
Can I ask why it is removed? Ref Hyrums law. The cost og this change could be quite high.

turbolent:
See my comment above: https://github.com/onflow/cadence/pull/2531#issuecomment-2075423041. I had also tried to explain it here: https://github.com/onflow/flow-go/pull/5820#discussion_r1586467347

The goal of https://github.com/onflow/cadence/pull/3290 was to remove access by index. `Value.ToGoValue` for composites (structs, resources, etc.) still returned field values as a slice, which basically left a "loop-hole", where users could keep making wrong assumptions (order of fields). Changing the result of `ToGoValue` for composites, e.g. to return a `map`, would have silently (!) broken all users of the function, because they likely expected (used a cast to) a slice, e.g. `Struct.ToGoValue.([]cadence.Value)`.

bjartek:
Ok. So when will this version of cadence be used? Because this is more work for everybody and they need to know that they have to do the work.

turbolent:
@bjartek I though I had already answered the same question above: https://github.com/onflow/cadence/pull/2531#issuecomment-2088785582.

This change only affects developers using the Go package, and the changes are in the release notes.

bjartek:
So i assume this can just be closed then?

PR #2547, opened 2023-06-07
Fix test: Pass interpreter when iterating over arrays

turbolent:
## Description

Since #2544, the interpreter must be passed to the `ArrayValue.Iterate` function. Previously it only needed a memory gauge.

Pass the interpreter everywhere the `Iterate` function is called, `nil` is not a valid value.

Thanks @SupunS for finding this issue.

______


SupunS (approved):
Thanks for fixing this! 🙏

PR #2551, opened 2023-06-08
Bring back type IDs in JSON encoding of funtion types and restricted types

turbolent:
## Description

The JSON-CDC encoding of restricted types and function types used to have type IDs. 

The type IDs got replaced with structured equivalents. For example, function types used to _only_ have a type ID, but now include parameter types, return type, etc.

#2401 removed the redundant type IDs.

However, before the removal, the JSON decoder _required_ these type IDs. The older decoder is still used by developers in older Go SDK and Flow CLI releases.

Bring back the type IDs for backward compatibility. 

______

### DIFF
--- a/encoding/json/encode.go
+++ b/encoding/json/encode.go
@@ -178,6 +178,7 @@ type jsonReferenceType struct {
 
 type jsonRestrictedType struct {
 	Kind         string      `json:"kind"`
+	TypeID       string      `json:"typeID"`
 	Type         jsonValue   `json:"type"`
 	Restrictions []jsonValue `json:"restrictions"`
 }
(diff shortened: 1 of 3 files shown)


dsainati1 (approved):

SupunS (approved):

turbolent:
Verified the fixes by switching the Emulator to use this PR, and then running scripts that return function types and restricted types:

```sh
$ flow scripts execute <(echo 'pub fun main(): Type { return Type<((): Void)>() }')

Result: Type<(():Void)>()
```

```sh
flow scripts execute <(echo 'pub struct S {}; pub fun main(): Type { return Type<S{}>() }')

Result: Type<s.468a847f9bdbfee1049c59934b561d32cba5f3045bf3bf7549653cf77e04e6d5.S{}>()
```

These scripts currently fail with:
```
Command Error: client: failed to construct cadence.Value entity from protobuf value: convert: failed to decode JSON-Cadence value: missing property: typeID
```

PR #2553, opened 2023-06-08
Adding contract init as entry points

bthaile:
Closes: https://github.com/onflow/cadence-tools/issues/129

## Description

Change to Sema which is used by the Cadence Checker.
The Change includes Contract file types `init` method as a "entry point".
This will treat contracts files the same as transactions and scripts.

As I understand it "Entry points" are user inputs. The issue is with the Language Server and how it adds methods for getting entry points and parsing entry points. It has a method for getting contract initialization parameters but not for parsing. By adding contracts into the "Entry Points" then the language server can treat all file types the same.

Also, added tests for contracts init parameters

______

### DIFF
--- a/runtime/sema/entrypoint.go
+++ b/runtime/sema/entrypoint.go
@@ -70,5 +70,12 @@ func (checker *Checker) EntryPointParameters() []Parameter {
 		return functionType.Parameters
 	}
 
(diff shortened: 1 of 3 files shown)


turbolent (approved):
Nice!

PR #2554, opened 2023-06-09
fetch event fields by name

rrrkren:
## Description

Added two methods to make it easy fetching event fields by their names. 
- `GetFieldByName`
- `GetFieldsMappedByName`
______

### DIFF
--- a/values.go
+++ b/values.go
@@ -1929,6 +1929,23 @@ func (v Event) String() string {
 	return formatComposite(v.EventType.ID(), v.EventType.Fields, v.Fields)
 }
 
+func (v Event) GetFieldByName(fieldName string) Value {
+	for i, field := range v.EventType.Fields {
+		if field.Identifier == fieldName {
+			return v.Fields[i]
+		}
+	}
+	return nil
+}
+
+func (v Event) GetFieldsMappedByName() map[string]Value {
+	fields := make(map[string]Value, len(v.EventType.Fields))
+	for i, field := range v.EventType.Fields {
+		fields[field.Identifier] = v.Fields[i]
+	}
+	return fields
+}
+
 // Contract
 
 type Contract struct {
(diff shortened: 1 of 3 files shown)


values.go
```diff
@@ -1929,6 +1929,23 @@ func (v Event) String() string {
...
+	return nil
+}
+
+func (v Event) GetFieldsMappedByName() map[string]Value {
+	fields := make(map[string]Value, len(v.EventType.Fields))
+	for i, field := range v.EventType.Fields {
+		fields[field.Identifier] = v.Fields[i]
+	}
+	return fields
+}
```
turbolent, line 1947: Nice! 

While at it, could you please also add those functions for the other composite types, i.e. `Struct`, `Resource`, `Contract`, and `Enum`? 

The current implementation could maybe be factored out into helper functions (e.g. `func GetFieldByName(values []Value, fields []Field) Value` and `func GetFieldsMappedByName(values []Value, fields []Field) map[string]Value`) and used/called in all those methods.

turbolent (approved):
Nice! Thank you for adding these! 👍

values.go
rrrkren: what do you think about something like this?
```go

type HasFields interface {
	GetFields() []Field
	GetFieldValues() []Value
}

func GetFieldByName(v HasFields, fieldName string) Value {
	for i, field := range v.GetFields() {
		if field.Identifier == fieldName {
			return v.GetFieldValues()[i]
		}
	}
	return nil
}

func GetFieldsMappedByName(v HasFields) map[string]Value {
	fieldValues := v.GetFieldValues()
	fields := v.GetFields()
	fieldsMap := make(map[string]Value, len(fields))
	for i, field := range fields {
		fieldsMap[field.Identifier] = fieldValues[i]
	}
	return fieldsMap
}
```

then implement GetFields and GetFieldValues for those composite types?

turbolent: That also works 👍 There's a small performance hit, but it's neligable.

turbolent (approved):
Nice! Thank you

turbolent:
Tests had failed because `TestRuntimeMalformedArgumentPassing` is racy

Issue #2449, opened 2023-04-17
CCF codec: allow default CBOR limits to be overridden


fxamacker:
### Issue to be solved

Currently, CBOR limits in CCF codec are hardcoded based on GRPC size limit and other factors.

In the future, it can be useful for each new CCF-based protocol to specify their own CBOR limits.

A GRPC limit of 20 MB can support (at most) a 20,000,000 element array (for an unrealistic message with zero-overhead and 1 byte elements).  

In practice, it would take many thousands of non-malicious CCF messages (like average-sized events) to reach a 20 MB GRPC limit, so it doesn't make sense to allow more than 20,000,000 elements for each array within a single CCF message.

The main tradeoff for decoder limits:
- too high will allow memory exhaustion attacks, etc. to succeed.
- too low will create the possibility of being unable to decode a non-malicious message that exceeds  limits.
- NOTE: Encoders usually don't enforce limits because it's much simpler and more efficient for apps to enforce it.

### Suggested Solution

For CCF Codec, add API to specify different decoder limits without breaking existing API.

Update https://github.com/fxamacker/ccf_draft/issues/64

fxamacker:
This is closed by #2559.

Issue #2283, opened 2023-01-31
Event encoding with CCF


j1010001:
### Issue to be solved

Events payloads are encoded with Cadence JSON, for example when transmitting data to ANs as a result of AN GetTransactionResult call. Cadence JSON encoding format is not designed for optimized performance and minimal payload size. This is causing issues with large events like reward payouts.

### Suggested Solution

Switch event encoding to CCF (https://github.com/onflow/cadence/issues/2157). This requires changes in the ENs as well as transparent handling of the new event format on ANs.

PR #2559, opened 2023-06-10
Add support for CCF encoding/decoding modes and options

### LINKED ISSUE #2558: Add support for CCF codec options and modes
### Issue to be solved

Some client programs are accessing fields by index rather than name.  This created a dependency on the sort order of fields which surfaced when we switched from JSON-CDC to CCF for encoding events because CCF explicitly sorts as specified in CCF specification for deterministic encodings.

### Suggested Solution

One solution is for CCF codec to allow different CCF-based protocols to use different CCF settings such as sort order.

Provide presets for CCF events encoding and decoding:
- at startup, create events encoding mode that can be used concurrently.
- at startup, create events decoding mode that can be used concurrently.
- allow custom options to be specified and allow new modes (safe for concurrent use) to be created from custom options.

For now, make the default CCF encoder and decoder settings the same as required by events.

`ccf.Encode()` uses default encoding options and is safe for concurrent use.
`ccf.EventsEncMode.Encode()` uses preset options for events encoding and is safe for concurrent use.

And etc.  See PR for details.

fxamacker:
Closes #2558 #2449 
Updates #2283 #2157 

## Description

Add CCF options which can be used to specify encoding and decoding settings.

Add CCF modes to have immutable CCF options and be safe for concurrent use.

Create CCF event encoding and decoding mode with CCF presets at startup. Calling `ccf.EventsEncMode.Encode()` or `ccf.EventsDecMode.Decode()` is safe for concurrent use and uses appropriate settings for events.

For now, the default CCF settings match the presets for events. So calling `ccf.Encode()` is like calling `ccf.EventsEncMode.Encode()`. However, it is safer to use a separate explicit mode for each CCF-based protocol.

### TODO:

- [x] Add more tests.  This PR can be reviewed while I add more tests.
- [ ] Fuzz test.
______

### DIFF
--- a/encoding/ccf/encode.go
+++ b/encoding/ccf/encode.go
@@ -32,11 +32,11 @@ import (
 	cadenceErrors "github.com/onflow/cadence/runtime/errors"
 )
 
-// CBOREncMode
+// defaultCBOREncMode
 //
 // See https://github.com/fxamacker/cbor:
 // "For best performance, reuse EncMode and DecMode after creating them."
-var CBOREncMode = func() cbor.EncMode {
+var defaultCBOREncMode = func() cbor.EncMode {
 	options := cbor.CoreDetEncOptions()
 	options.BigIntConvert = cbor.BigIntConvertNone
 	encMode, err := options.EncMode()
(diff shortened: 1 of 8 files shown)


encoding/ccf/encode.go
```diff
@@ -943,6 +1037,18 @@ func (e *Encoder) encodeComposite(
 		return err
 	}
 
+	if e.em.sortCompositeFields == SortNone {
```
turbolent, line 1040: Here, and for the similar code below: `EnforceSortMode` is an enum. It currently only has two cases, so checking for the first and returning, then implicitly continuing for the second case works. Maybe make the handling of each case explicit by using a switch case, and panic with `NewUnreachableError` in the default case.

turbolent (commented):
Great work! The changes look good 👍

turbolent (approved):
Great work! 👏

turbolent:
@onflow/cadence could you please also have a look before we deploy?

SupunS (commented):
LGTM!

PR #2564, opened 2023-06-12
Add tests for entitlement inheritance between interfaces

dsainati1:
With interface inheritance and entitlements now in the same branch, we now need tests for their interaction. I think however that the implementations just work together out of the box.
______

### DIFF
--- a/runtime/tests/checker/entitlements_test.go
+++ b/runtime/tests/checker/entitlements_test.go
@@ -2607,7 +2607,7 @@ func TestCheckEntitlementInheritance(t *testing.T) {
 		require.IsType(t, &sema.ConformanceError{}, errs[0])
 	})
 
-	t.Run("default function entitlemnts", func(t *testing.T) {
+	t.Run("default function entitlements", func(t *testing.T) {
 		t.Parallel()
 		_, err := ParseAndCheck(t, `
 			entitlement E
@@ -2632,7 +2632,7 @@ func TestCheckEntitlementInheritance(t *testing.T) {
 		assert.NoError(t, err)
 	})
 
-	t.Run("attachment default function entitlemnts", func(t *testing.T) {
+	t.Run("attachment default function entitlements", func(t *testing.T) {
 		t.Parallel()
 		_, err := ParseAndCheck(t, `
 			entitlement E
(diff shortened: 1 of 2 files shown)


runtime/tests/checker/interface_test.go
```diff
@@ -2998,6 +2998,30 @@ func TestCheckInterfaceInheritance(t *testing.T) {
...
+
+		_, err := ParseAndCheck(t, `
+			entitlement X
+
+            struct interface Foo {
+                access(X) fun hello()
+            }
+
+            struct interface Bar: Foo {
+                pub fun hello(): String
```
SupunS, line 3013: Could be mismatching also because of the different return types. So maybe keep the rest of the function signature identical.
```suggestion
                pub fun hello()
```

runtime/tests/checker/entitlements_test.go
```diff
@@ -1975,6 +1975,21 @@ func TestCheckEntitlementInheritance(t *testing.T) {
...
+				access(E) fun foo() 
+			}
+			struct interface S: I {
+				access(E) fun foo() 
+			}
+		`)
+
+		assert.NoError(t, err)
+	})
+
```
SupunS, line 1992: I remember we once discussed about entitlements coming from default functions: I think it was something like taking an `auth{M}` reference to a composite `T` where `T` conforms to interface `I`, and `I` have default function(s) with entitlement mapping `M`, while `T` doesn't have any entitlements directly specified on it. 

Maybe also add a test for the same, but this time for entitlements coming from inherited default functions. Think it should also work out of the box.

runtime/tests/checker/entitlements_test.go
```diff
@@ -2550,6 +2606,92 @@ func TestCheckEntitlementInheritance(t *testing.T) {
...
+				E -> F
+			}
+			entitlement mapping N {
+				G -> E
+			}
+			struct interface I {
+				access(M) fun foo(): auth(M) &Int {
+					return &1 as auth(M) &Int
+				}
+			}
```
SupunS, line 2651: just one last test: also make this to inherit from another interface.
i.e:
```cadence
struct interface I1 {
    access(M) fun foo(): auth(M) &Int {
        return &1 as auth(M) &Int
    }
}

struct interface I2: I1 {}

struct S {}

access(N) attachment A for S: I2 {}
```

SupunS (approved):

turbolent (approved):
Nice!

PR #2567, opened 2023-06-12
Convert generic container types on upcast

dsainati1:
As outlined in https://github.com/onflow/flips/pull/54, entitlements should be "erased" at runtime, and should exactly match the static entitlements present on any static reference types. This is currently implemented for references by converting all reference types on transfer, but should also work for any types that may contain references. 

This PR changes any types that may potentially contain references (optionals, arrays, dictionaries and capabilities) to convert on transfer as well. 

______


runtime/entitlements_test.go
```diff
@@ -501,3 +501,322 @@ func TestAccountEntitlementNamingConflict(t *testing.T) {
...
+				let capX = signer.getCapability<auth(Test.X) &Test.R>(/public/foo)
+				let capY = signer.getCapability<auth(Test.Y) &Test.R>(/public/bar)
+
+				let dict: {Type: Capability} = {}
+				dict[capX.getType()] = capX
+				dict[capY.getType()] = capY
+
+				let newCapX = dict[capX.getType()]!
+				let ref = newCapX.borrow<auth(Test.X) &Test.R>()!
+				let downCast = ref as! auth(Test.X) &Test.R
```
dsainati1, line 762: This succeeds, which is the required use case for generic Capabilities that we identified during the discussion on https://github.com/onflow/flips/pull/95

runtime/interpreter/interpreter.go
```diff
@@ -1776,17 +1776,100 @@ func (interpreter *Interpreter) ConvertAndBox(
...
+func (interpreter *Interpreter) convertStaticType(
+	valueStaticType StaticType,
+	targetSemaType sema.Type,
+) StaticType {
+	switch valueStaticType := valueStaticType.(type) {
+	case ReferenceStaticType:
+		if targetReferenceType, isReferenceType := targetSemaType.(*sema.ReferenceType); isReferenceType {
+			return NewReferenceStaticType(
+				interpreter,
+				ConvertSemaAccesstoStaticAuthorization(interpreter, targetReferenceType.Authorization),
```
turbolent, line 1788: Is my understanding here correct that `convertStaticType` "converts" a static type to another static type, given a semantic type that must "match" the input static type, and this, the authorization of the reference type, is the only case where the target type's info is used?

runtime/interpreter/interpreter.go
```diff
@@ -1905,10 +1988,88 @@ func (interpreter *Interpreter) convert(value Value, valueType, targetType sema.
 			return ConvertAddress(interpreter, value, locationRange)
 		}
 
+	case *sema.ConstantSizedType, *sema.VariableSizedType:
+		if arrayValue, isArray := value.(*ArrayValue); isArray && !valueType.Equal(unwrappedTargetType) {
+
+			arrayStaticType := interpreter.convertStaticType(arrayValue.StaticType(interpreter), unwrappedTargetType).(ArrayStaticType)
+			targetElementType := interpreter.MustConvertStaticToSemaType(arrayStaticType.ElementType())
+
+			values := make([]Value, 0, arrayValue.Count())
```
turbolent, line 1997: This allocation needs to be metered. Ideally we might want to avoid it (see below).

runtime/interpreter/interpreter.go
```diff
@@ -1905,10 +1988,88 @@ func (interpreter *Interpreter) convert(value Value, valueType, targetType sema.
...
+				values = append(values, interpreter.convert(v, valueType, targetElementType, locationRange))
+				return true
+			})
+
+			return NewArrayValue(
+				interpreter,
+				locationRange,
+				arrayStaticType,
+				arrayValue.GetOwner(),
+				values...,
```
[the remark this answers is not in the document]
turbolent: Instead of allocating a temporary slice, `NewArrayValueWithIterator` can be used to construct the new array value  from an existing "iterator". Same below for dictionary values.

runtime/interpreter/interpreter.go
```diff
@@ -1905,10 +1988,88 @@ func (interpreter *Interpreter) convert(value Value, valueType, targetType sema.
...
+				)
+			case *IDCapabilityValue:
+				valueBorrowType := capability.BorrowType.(ReferenceStaticType)
+				borrowType := interpreter.convertStaticType(valueBorrowType, targetBorrowType)
+				return NewIDCapabilityValue(
+					interpreter,
+					capability.ID,
+					capability.Address,
+					borrowType,
+				)
```
turbolent, line 2064: Maybe panic with an unreachable error in the default case (unsupported capability value)

runtime/tests/interpreter/entitlements_test.go
```diff
@@ -577,6 +577,320 @@ func TestInterpretEntitledReferenceCasting(t *testing.T) {
...
+				let upCap = capX as Capability
+				return upCap as? Capability<auth(X) &Int> == nil
+			}
+			`,
+			sema.Config{})
+
+		value, err := inter.Invoke("test")
+		require.NoError(t, err)
+
+		AssertValuesEqual(
```
turbolent, line 638: Maybe return the resulting capability value and assert the value, instead of just that it is non-nil

turbolent (commented):
Looks good!

runtime/interpreter/interpreter.go
turbolent: Could you please add a comment to the function and explain this behaviour a bit more?
That the only different is in using the target reference authorization is very subtle / difficult to notice.

runtime/interpreter/value.go
```diff
@@ -16191,6 +16191,68 @@ func newDictionaryValueFromOrderedMap(
 	}
 }
 
+func newDictionaryValueWithIterator(
```
turbolent, line 16194: 👍

turbolent (commented):
Nice!

turbolent (approved):
Great work!

@SupunS could you please also have a look?

runtime/interpreter/interpreter.go
```diff
@@ -1905,10 +1992,119 @@ func (interpreter *Interpreter) convert(value Value, valueType, targetType sema.
...
+					}
+
+					value := MustConvertStoredValue(interpreter, element)
+					valueType := interpreter.MustConvertStaticToSemaType(value.StaticType(interpreter))
+					return interpreter.convert(value, valueType, targetElementType, locationRange)
+				},
+			)
+		}
+
+	case *sema.DictionaryType:
```
SupunS, line 2029: So for arrays and dictionaries, this function didn't create new value instances previously, but now it does.

Given this can be expensive, is it possible to ensure that this conversion happens only for reference types (containers with references inside), but not for other types?

SupunS: Also, have we considered the alternative solution: making arrays and dictionaries non-covariant?
Might need a FLIP though

runtime/interpreter/interpreter.go
```diff
@@ -1776,17 +1776,104 @@ func (interpreter *Interpreter) ConvertAndBox(
...
+	valueStaticType StaticType,
+	targetSemaType sema.Type,
+) StaticType {
+	switch valueStaticType := valueStaticType.(type) {
+	case ReferenceStaticType:
+		if targetReferenceType, isReferenceType := targetSemaType.(*sema.ReferenceType); isReferenceType {
+			return NewReferenceStaticType(
+				interpreter,
+				ConvertSemaAccesstoStaticAuthorization(interpreter, targetReferenceType.Authorization),
+				valueStaticType.ReferencedType,
```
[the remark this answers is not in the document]
dsainati1: No, this is consistent with the behavior for other types like structs and resources when upcast to interfaces, the underlying value type at runtime does not change. The only thing that should change to match the sema type is the authorization, so that the runtime and static authorizations are always the same.

runtime/interpreter/interpreter.go
dsainati1: I think we should be able to do this only when the converted static type doesn't match the original, as that would indicate a reference type within has changed. Implemented this in 9bae858d79ee52359fd88c7ce037584f78f532bd

SupunS (approved):

PR #2571, opened 2023-06-13
Introduce the `Test.assertEqual` function

m-Peter:
Closes https://github.com/onflow/cadence-tools/issues/93

## Description

This function has built-in support for showing the difference between the expected and the actual value. For example:

```bash
Test results: "tests/test_foo_contract.cdc"
- FAIL: testGetIntegerTrait
		Execution failed:
			error: assertion failed: not equal: expected: "Hugee", actual: "Huge"
			  --> 7465737400000000000000000000000000000000000000000000000000000000:27:8
			
- PASS: testAddSpecialNumber
Coverage: 95.7% of statements
```

______

### DIFF
--- a/runtime/stdlib/test_contract.go
+++ b/runtime/stdlib/test_contract.go
@@ -106,6 +106,75 @@ var testTypeAssertFunction = interpreter.NewUnmeteredHostFunctionValue(
 	},
 )
 
+// 'Test.assertEqual' function
+
+const testTypeAssertEqualFunctionDocString = `
+Fails the test-case if the given values are not equal, and
+reports a message which explains how the two values differ.
+`
+
+const testTypeAssertEqualFunctionName = "assertEqual"
+
+var testTypeAssertEqualFunctionType = &sema.FunctionType{
+	Parameters: []sema.Parameter{
+		{
+			Label:      sema.ArgumentLabelNotRequired,
+			Identifier: "expected",
+			TypeAnnotation: sema.NewTypeAnnotation(
+				sema.AnyStructType,
+			),
+		},
+		{
+			Label:      sema.ArgumentLabelNotRequired,
+			Identifier: "actual",
+			TypeAnnotation: sema.NewTypeAnnotation(
+				sema.AnyStructType,
+			),
+		},
+	},
+	RequiredArgumentCount: sema.RequiredArgumentCount(2),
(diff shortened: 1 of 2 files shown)


runtime/stdlib/test_contract.go
```diff
@@ -106,6 +106,75 @@ var testTypeAssertFunction = interpreter.NewUnmeteredHostFunctionValue(
...
+			Identifier: "expected",
+			TypeAnnotation: sema.NewTypeAnnotation(
+				sema.AnyStructType,
+			),
+		},
+		{
+			Label:      sema.ArgumentLabelNotRequired,
+			Identifier: "actual",
+			TypeAnnotation: sema.NewTypeAnnotation(
+				sema.AnyStructType,
```
kozlovb, line 131: I wonder if equality of two types could be enforced easily.

By that I mean that if types are not equal then the very first test execution is not going to pass anyway. So, its not something that important. On the other hand having comparisons of ints to string fail at the compile time could've been nice too. 

Also, may be there is already a check at the compile time that I've missed.

runtime/stdlib/test_test.go
```diff
@@ -651,6 +651,291 @@ func TestTestEqualMatcher(t *testing.T) {
...
+	t.Parallel()
+
+	t.Run("success", func(t *testing.T) {
+		t.Parallel()
+
+		script := `
+		    import Test
+
+		    pub fun test() {
+		        Test.assertEqual("this string", "this string")
```
kozlovb, line 665: In relation to my previous comment - Would something like Test.assertEqual(10, "this string") fail at compile time ?

If not it would be interesting to see a test like this in order to see if expected.Equal survives.

m-Peter (commented):
All good!

runtime/stdlib/test_contract.go
kozlovb: Right, cadence is interpreted language, my bad.

Great, that you've covered that in test. "not equal: expected: true, actual: 1" as discussed in PMs may be error message can specify that types differ.

kozlovb: Although, commit is merged and I think this point is "matter of taste"/minor. So, fine for me either way.

m-Peter: Great, thanks for pointing it out though :raised_hands:

PR #2573, opened 2023-06-14
Change member access semantics

### LINKED ISSUE #2534: Change member-access semantics
### Issue to be solved

This is to track the implementation of https://github.com/onflow/flips/pull/89

### Suggested Solution

_No response_

SupunS:
Closes #2534

## Description

This is the prototype implementation for: https://github.com/onflow/flips/pull/89

Changes member-access/index-access to return a reference, if the parent is also a reference.

______


runtime/sema/check_variable_declaration.go
```diff
@@ -280,6 +280,16 @@ func (checker *Checker) referencedVariables(expr ast.Expression) (variables []*V
...
+		case *ast.IndexExpression:
+			// If it is a reference expression, then find the "root variable".
+			// As nested resources cannot be tracked, at least track the "root" if possible.
+			// For example, for an expression `a[b][c]`, the "root variable" is `a`.
+			variableRefExpr = rootVariableOfExpression(refExpr.TargetExpression)
+		case *ast.MemberExpression:
+			// If it is a reference expression, then find the "root variable".
+			// As nested resources cannot be tracked, at least track the "root" if possible.
+			// For example, for an expression `a.b.c`, the "root variable" is `a`.
+			variableRefExpr = rootVariableOfExpression(refExpr.Expression)
```
SupunS, line 292: This is an improvement to the static reference invalidation check.

Since now member-access/index-access can also produce a reference, they are also tracked now.

A nice side effect of this is, now attachment invalidation is also happening statically (see the new test case added in `runtime/tests/checker/reference_test.go` below)

runtime/attachments_test.go
```diff
@@ -170,10 +170,18 @@ func TestAccountAttachmentExportFailure(t *testing.T) {
 		import Test from 0x1
 		pub fun main(): &Test.A? { 
 			let r <- Test.makeRWithA()
-			let a = r[Test.A]
+			var a = r[Test.A]
+
+			// just to trick the checker
+			a = returnSameRef(a)
```
dsainati1, line 176: What is the need for this now?

runtime/interpreter/interpreter_expression.go
```diff
@@ -222,6 +248,53 @@ func (interpreter *Interpreter) memberExpressionGetterSetter(memberExpression *a
...
+
+		if optionalValue, ok := value.(*SomeValue); ok {
+			value = optionalValue.value
+		}
+		innerValue := interpreter.getReferenceValue(value, semaType)
+		return NewSomeValueNonCopying(interpreter, innerValue)
+	}
+
+	interpreter.maybeTrackReferencedResourceKindedValue(value)
+	return NewEphemeralReferenceValue(interpreter, false, value, semaType)
```
dsainati1, line 295: I know this feature is theoretically not dependent on entitlements, but it might be worthwhile to base this PR on `feature/entitlements` to save yourself the merge conflict later. Here you would need to propagate the parent entitlements through any potential entitlement mapping.

runtime/tests/checker/member_test.go
```diff
@@ -418,3 +418,246 @@ func TestCheckMemberNotDeclaredSecondaryError(t *testing.T) {
...
+
+		require.NoError(t, err)
+	})
+
+	t.Run("array reference, element", func(t *testing.T) {
+		_, err := ParseAndCheck(t, `
+            fun test() {
+                let array: [[Int]] = [[1, 2]]
+                let arrayRef = &array as &[[Int]]
+                var x: &[Int] = arrayRef[0]
```
dsainati1, line 552: Is this the behavior we want? I can see arguments both for and against, but it is not immediately clear from the FLIP (imo) that things should work this way, since I'm not sure it's necessarily the case that array/dictionary elements are really "fields".

SupunS: Yup, this is in accordance with the objective of the FLIP. The FLIP actually explicitly states about the array/dictionary behavior.
Maybe I can add more examples there, if that helps eliminate any potential confusion.

runtime/interpreter/interpreter_expression.go
SupunS: Good idea, actually entitlements are needed for the follow-up FLIPs anyway.

Rebased to a new feature branch `feature/mutability-restrictions` which is a fork from `feature/entitlements`

runtime/attachments_test.go
turbolent: Could you please add this explanation as a comment to the code?

runtime/interpreter/interpreter_expression.go
```diff
@@ -212,6 +221,23 @@ func (interpreter *Interpreter) memberExpressionGetterSetter(memberExpression *a
 				}
 			}
 
+			// Return a reference, if the member is accessed via a reference.
+			//
+			// However, for attachments, `self` is always a reference.
+			// But we do not want to return a reference for `self.something`.
+			// Otherwise, things like `destroy self.something` would become invalid.
+			// Hence, special case `self`, and return a reference only if the member is not accessed via self.
```
turbolent, line 229: Do we only need to handle the case where there's a member expression, and the LHS is `self`, e.g. `self.foo`? Do we also need to handle nested cases, e.g. `self.foo.bar`?

runtime/interpreter/interpreter_expression.go
```diff
@@ -212,6 +221,23 @@ func (interpreter *Interpreter) memberExpressionGetterSetter(memberExpression *a
 				}
 			}
 
+			// Return a reference, if the member is accessed via a reference.
```
turbolent, line 224: In the current implementation, the interpreter detects if a reference needs to be returned based on the value, on each evaluation. Would it be possible to determine this statically, i.e. move the logic here to the checker and store the result in the elaboration?

runtime/tests/checker/reference_test.go
```diff
@@ -830,17 +827,15 @@ func TestCheckInvalidReferenceResourceLoss(t *testing.T) {
       fun test() {
           let rs <- [<-create R()]
           let ref = &rs as &[R]
-          ref[0]
+          ref[0]  // This result in a reference, so no resource loss
```
turbolent, line 830: Do we keep a test for what was originally tested, indexing returning a resources and it not being dropped implicitly?

runtime/interpreter/interpreter_expression.go
SupunS: Good idea. Added in 5c8ea343e51e568e3fd756f233c97436556c82af

runtime/interpreter/interpreter_expression.go
SupunS: Because the expression visitor is recursive, handling only one level is sufficient. i.e. in `self.foo.bar`, the result of `.bar` depends on the result of `self.foo`. So as long as evaluating `self.foo` returns the correct result, then `.bar` doesn't need to be special-cased.

Added a test: eca92dfc94ed739876e321a3b80464398c8a45db

runtime/tests/checker/reference_test.go
SupunS: Not really sure. All the tests in this file are for reference testing. There is a separate set of tests for resources in the `checker/reference_test.go` file. There are quite a lot of tests there, so hard find whether the exact test is already available. So added a new one anyway https://github.com/onflow/cadence/pull/2573/commits/498c63a860a8d24c8f5765689416d33705499573.

runtime/tests/checker/member_test.go
SupunS: This came up in a follow PR as well, and posted a sample code there on why we would need this: https://github.com/onflow/cadence/pull/2587#discussion_r1237231577.

Like it was discussed there, we will return a reference for array/dictionary elements, but it will be an unauthorized one.

runtime/sema/check_swap.go
```diff
@@ -25,38 +25,28 @@ import (
...
-	leftType := checker.VisitExpression(swap.Left, nil)
-	rightType := checker.VisitExpression(swap.Right, nil)
+	// First visit the two expressions as if they were the target of the assignment.
+	leftTargetType := checker.checkSwapStatementExpression(swap.Left, common.OperandSideLeft)
+	rightTargetType := checker.checkSwapStatementExpression(swap.Right, common.OperandSideRight)
+
+	// Then re-visit the same expressions, this time treat them as the value-expr of the assignment.
+	// The 'expected type' of the two expression would be the types obtained from the previous visit, swapped.
+	leftValueType := checker.VisitExpression(swap.Left, rightTargetType)
+	rightValueType := checker.VisitExpression(swap.Right, leftTargetType)
```
SupunS, line 35: This change is more like a clean-up: But also needed to handle when the swap-statement has references.

runtime/interpreter/interpreter_expression.go
```diff
@@ -222,6 +238,43 @@ func (interpreter *Interpreter) memberExpressionGetterSetter(memberExpression *a
...
+// e.g.1: Given type T, this method returns &T.
+// e.g.2: Given T?, this returns (&T)?
+func (interpreter *Interpreter) getReferenceValue(value Value, semaType sema.Type) Value {
+	switch value.(type) {
+	case NilValue:
+		// Reference to a nil, should return a nil.
+		return value
+	case ReferenceValue:
+		// If the value is already a reference then return the same reference.
+		return value
```
dsainati1, line 253:
```suggestion
	case NilValue, ReferenceValue:
		// Reference to a nil, should return a nil.
		// If the value is already a reference then return the same reference.
		return value
```

runtime/sema/check_member_expression.go
```diff
@@ -83,30 +83,78 @@ func (checker *Checker) VisitMemberExpression(expression *ast.MemberExpression)
...
+
+func isReferenceType(typ Type) bool {
+	unwrappedType := UnwrapOptionalType(typ)
+	_, isReference := unwrappedType.(*ReferenceType)
+	return isReference
+}
+
+func isContainerType(typ Type) bool {
+	switch typ := typ.(type) {
+	case *CompositeType,
```
dsainati1, line 126: Shouldn't we return references when the member type is an interface or restricted type as well?

runtime/tests/checker/member_test.go
```diff
@@ -418,3 +418,299 @@ func TestCheckMemberNotDeclaredSecondaryError(t *testing.T) {
 		assert.Equal(t, "unknown member", memberErr.SecondaryError())
 	})
 }
+
+func TestCheckMemberAccess(t *testing.T) {
```
dsainati1, line 422: This has test cases for each of {Dictionary, Array, Composite} container types, but only tests access for Array and Dictionary member types. We should have tests for Composite and Restricted type members as well.

runtime/sema/check_member_expression.go
SupunS: That's a good flag! Realized we would need to do it for all built-in types as well. Moved this logic to the type definitions itself: 9a64a7fddffd31976ee9abc25caaaf650083fd68

runtime/tests/checker/member_test.go
SupunS: added in 9a64a7fddffd31976ee9abc25caaaf650083fd68

runtime/interpreter/interpreter_expression.go
```diff
@@ -169,6 +172,12 @@ func (interpreter *Interpreter) memberExpressionGetterSetter(memberExpression *a
 
 	isNestedResourceMove := interpreter.Program.Elaboration.IsNestedResourceMoveExpression(memberExpression)
 
+	memberInfo, ok := interpreter.Program.Elaboration.MemberExpressionMemberInfo(memberExpression)
```
turbolent, line 175: Just related: Maybe we should improve the naming of `MemberInfo` to e.g. `MemberAccessInfo` or just `MemberAccess`, as it is the info for a particular access, not the member in general. 

I was confused for a moment why `ReturnReference` was added there, thinking it would apply for any member usage (but it isn't)

runtime/interpreter/interpreter_expression.go
```diff
@@ -212,6 +221,13 @@ func (interpreter *Interpreter) memberExpressionGetterSetter(memberExpression *a
 				}
 			}
 
+			// Return a reference, if the member is accessed via a reference.
+			// This is pre-computed at the checker.
+			if memberInfo.ReturnReference {
```
turbolent, line 226: Nice!

runtime/interpreter/interpreter_expression.go
```diff
@@ -222,6 +238,41 @@ func (interpreter *Interpreter) memberExpressionGetterSetter(memberExpression *a
 	}
 }
 
+// getReferenceValue Returns a reference to a given value.
+// Reference to an optional should return an optional reference.
+// This has to be done recursively for nested optionals.
+// e.g.1: Given type T, this method returns &T.
+// e.g.2: Given T?, this returns (&T)?
+func (interpreter *Interpreter) getReferenceValue(value Value, semaType sema.Type) Value {
```
turbolent, line 246: Does this match the checker implementation?

runtime/missingmember_test.go
```diff
@@ -3542,18 +3542,18 @@ access(all) contract AuctionDutch {
...
-			self.status=status
-			self.currentPrice=currentPrice
-			self.totalItems=totalItems
-			self.acceptedBids=acceptedBids
-			self.startTime=startTime
-			self.currentTime= 42.0 // Clock.time()
-			self.tickStatus=tickStatus
-			self.metadata=metadata
+		access(all) let tickStatus: &{UFix64:TickStatus}
+		access(all) let metadata: &{String:String}
```
turbolent, line 3546: Why doe the field types have to be changed to references?

runtime/resourcedictionary_test.go
```diff
@@ -252,11 +252,12 @@ func TestRuntimeResourceDictionaryValues(t *testing.T) {
      transaction {
 
          prepare(signer: AuthAccount) {
-             let c = signer.borrow<&Test.C>(from: /storage/c)!
```
turbolent, line 255: Here and below: Why do we have to load+save now and can't keep borrowing? Maybe keep the borrow and adjust the test cases to allow the operations on the reference, like it used to. 

For example, `forceInsert` was already introduced earlier in the same manner.

runtime/sema/type.go
```diff
@@ -135,6 +135,10 @@ type Type interface {
 	// IsComparable returns true if values of the type can be compared
 	IsComparable() bool
 
+	// IsMemberAccessible returns true if value of the type can have members values.
+	// Examples are composite types, restricted types, arrays, dictionaries, etc.
+	IsMemberAccessible() bool
```
turbolent, line 140: Do we need this notion and `MemberAccessible`? Isn't it possible to access members of all types?

runtime/sema/check_variable_declaration.go
turbolent: Great! 👌

runtime/storage_test.go
```diff
@@ -2526,7 +2526,7 @@ func TestRuntimeReferenceOwnerAccess(t *testing.T) {
                   // At this point the resource is in storage
                   account.link<&[TestContract.TestResource]>(/public/test, target: /storage/test)
                   let ref2 = account.getCapability<&[TestContract.TestResource]>(/public/test).borrow()!
-                  let ref3 = &ref2[0] as &TestContract.TestResource
```
turbolent, line 2529: Love how this simplifies the code 👌

runtime/tests/checker/reference_test.go
```diff
@@ -790,10 +790,8 @@ func TestCheckReferenceIndexingIfReferencedIndexable(t *testing.T) {
           fun test() {
               let rs <- [<-create R()]
               let ref = &rs as &[R]
-              var other <- create R()
-              ref[0] <-> other
```
turbolent, line 794: Do we still have tests for the non-reference / owned case that swaps? Same below

runtime/tests/checker/reference_test.go
turbolent, line 830: Do we still have a case for the non-reference / owned case?

runtime/tests/checker/swap_test.go
```diff
@@ -122,9 +123,10 @@ func TestCheckInvalidTypesSwap2(t *testing.T) {
       }
     `)
 
-	errs := RequireCheckerErrors(t, err, 1)
+	errs := RequireCheckerErrors(t, err, 2)
 
 	assert.IsType(t, &sema.TypeMismatchError{}, errs[0])
+	assert.IsType(t, &sema.TypeMismatchError{}, errs[0])
```
turbolent, line 129:
```suggestion
	assert.IsType(t, &sema.TypeMismatchError{}, errs[1])
```

runtime/tests/checker/swap_test.go
```diff
@@ -105,9 +105,10 @@ func TestCheckInvalidTypesSwap(t *testing.T) {
...
```
turbolent, line 111:
```suggestion
	assert.IsType(t, &sema.TypeMismatchError{}, errs[1])
```

turbolent (commented):
Great work!

runtime/interpreter/interpreter_expression.go
SupunS: Yes; checker's counter-part of this: https://github.com/onflow/cadence/blob/c16e5682fa679469cd52d9331d1f117d1275eed9/runtime/sema/check_member_expression.go#L95-L108

runtime/missingmember_test.go
SupunS: These `tickStatus` and `metadata` fields are populated from fields of a reference. So either we need to change the type of these two fields, or have to change the entire test to work with owned values. The former is the minimal change so went with that.

runtime/sema/type.go
SupunS: This notion is to indicate that a type can be used to access its members using either index-express or member-expression. e.g; `foo.bar` or `foo[bar]`. Only a subset of types has this ability. e.g:
- Composites
- Interfaces
- Arrays (Variable/constant sized)
- Dictionaries
- Restricted types
- Optionals of any of the above.
- 
This is used to determine if the field should be returning a reference or not (primitive fields doesn't need a reference tp be returned)

We could move this to a switch-case (originally that's how I implemented) but it becomes harder to maintain in the long run.

Maybe a better name would avoid any confusions?

runtime/resourcedictionary_test.go
SupunS: ah this is no longer needed. Reverted in https://github.com/onflow/cadence/pull/2573/commits/b72a3b1f04e100888d52f2a59c8a6b47cc95b312

runtime/interpreter/interpreter_expression.go
SupunS: Renamed in 84e7e0fd759fed08fdbd04e0b0be0b8d722b8eed

runtime/tests/checker/reference_test.go
SupunS: Yes, they are in `swap_test.go` file.

runtime/tests/checker/reference_test.go
SupunS: Deja-vu: https://github.com/onflow/cadence/pull/2573#discussion_r1234448193 😄

runtime/sema/type.go
turbolent: I see. "Member accessible" in itself just means accessing a member, e.g. `foo.bar`; `foo[bar]` is an indexing access. Maybe rename the type and function to indicate that (maybe `MemberAccessibleOrIndexable`?)

It's still a bit unclear why we need it though, given we already have `ValueIndexableType`/`TypeIndexableType` for indexable types, and `Type.GetMembers` (every type is member-accessible).

SupunS: Yeah, can rename it to `MemberAccessibleOrIndexable`. My first choice was to use `IsContainerType`, but we already use that (and have a `ContainerType`) to represent the notion of "a type that has nested types".

The problem with `ValueIndexableType` is that simple types also implement this, but not all simple types are indexable. Even some indexable types (e.g.: `String`) shouldn't be treated/returned as references.

SupunS: Renamed it to `ContainFieldOrElements` in e47f4517f08f5cb6ea8bf24f8ec8dc88fbc081ed.
Probably not the most appropriate, but hopefully it'll reduce the chance of confusion with "members".

I didn't go with the "ReferenceReturnable" or anything specifying references, because for a type, reference-returnable sounds like "can take a reference to this type", which is not really what we are intending.
The function which calls this method (shouldReturnReference), is the one that actually checks whether a reference should be returned or not, and checking whether the type `ContainFieldOrElements` is only a part of that check.

SupunS: As discussed, we can of-course change this in the future.

turbolent (approved):
Great work!

PR #2577, opened 2023-06-15
Add the `CoverageReport.MarshalLCOV` method to generate LCOV format

m-Peter:
## Description

Add a new method to the `CoverageReport` type, to generate the appropriate LCOV format. Currently only line coverage is supported. Function/branch coverage are to be added next up.
The generated LCOV output, can be tested here: https://lcov-viewer.netlify.app/ . For example:
```
TN:
SF:S.IntegerTraits
DA:9,1
DA:13,10
DA:14,1
DA:15,9
DA:16,1
DA:17,8
DA:18,1
DA:19,7
DA:20,1
DA:21,6
DA:22,1
DA:25,5
DA:26,4
DA:29,1
LF:14
LH:14
end_of_record
```

______


runtime/coverage.go
```diff
@@ -483,3 +484,45 @@ func (r *CoverageReport) UnmarshalJSON(data []byte) error {
...
+// MarshalLCOV serializes each common.Location/*LocationCoverage
+// key/value pair on the *CoverageReport.Coverage map, to the
+// LCOV format. Currently supports only line coverage, function
+// and branch coverage are not yet available.
+func (r *CoverageReport) MarshalLCOV() ([]byte, error) {
+	buf := new(bytes.Buffer)
+	for location, coverage := range r.Coverage { // nolint:maprange
+		_, err := buf.WriteString(
+			fmt.Sprintf("TN:\nSF:%s\n", location.ID()),
+		)
```
turbolent, line 497: Here and below: Directly write to the buffer using `fmt.Fprintf`, which avoids allocating temporary strings:

```suggestion
		_, err := fmt.Fprintf(buf, "TN:\nSF:%s\n", location.ID()),
```

runtime/coverage.go
```diff
@@ -483,3 +484,45 @@ func (r *CoverageReport) UnmarshalJSON(data []byte) error {
 
 	return nil
 }
+
+// MarshalLCOV serializes each common.Location/*LocationCoverage
+// key/value pair on the *CoverageReport.Coverage map, to the
+// LCOV format. Currently supports only line coverage, function
+// and branch coverage are not yet available.
```
turbolent, line 491: Could you please add a link to a description of the file format here?

I found it in lcov's geninfo man page, https://github.com/linux-test-project/lcov/blob/master/man/geninfo.1#L948, not sure if there's a better description somewhere else.

turbolent (approved):
Very nice! It's going to be great to make use of the existing tooling around this, e.g. visualizing coverage in source files 👍

runtime/coverage.go
m-Peter: Looks quite official, so I am going with this :+1:
Added in https://github.com/onflow/cadence/pull/2577/commits/117d51eb2a1e623d9a61613a5a4b2be55a6b1e88

runtime/coverage.go
```diff
@@ -483,3 +484,42 @@ func (r *CoverageReport) UnmarshalJSON(data []byte) error {
...
+// Description for the LCOV file format, can be found here
+// https://github.com/linux-test-project/lcov/blob/master/man/geninfo.1#L948.
+func (r *CoverageReport) MarshalLCOV() ([]byte, error) {
+	buf := new(bytes.Buffer)
+	for location, coverage := range r.Coverage { // nolint:maprange
+		_, err := fmt.Fprintf(buf, "TN:\nSF:%s\n", location.ID())
+		if err != nil {
+			return nil, err
+		}
+		lines := make([]int, 0)
```
kozlovb, line 501: I think you can allocate directly lines of size len(coverage.LineHits) to avoid memory reallocations later.

runtime/coverage.go
```diff
@@ -483,3 +484,42 @@ func (r *CoverageReport) UnmarshalJSON(data []byte) error {
...
+
+// MarshalLCOV serializes each common.Location/*LocationCoverage
+// key/value pair on the *CoverageReport.Coverage map, to the
+// LCOV format. Currently supports only line coverage, function
+// and branch coverage are not yet available.
+// Description for the LCOV file format, can be found here
+// https://github.com/linux-test-project/lcov/blob/master/man/geninfo.1#L948.
+func (r *CoverageReport) MarshalLCOV() ([]byte, error) {
+	buf := new(bytes.Buffer)
+	for location, coverage := range r.Coverage { // nolint:maprange
```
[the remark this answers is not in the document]
m-Peter: Ah yes, just like with line hits. The LCOV file generation will be non-deterministic, even without actual changes. I will fix this, good catch :100:

runtime/coverage.go
```diff
@@ -483,3 +484,42 @@ func (r *CoverageReport) UnmarshalJSON(data []byte) error {
...
+		_, err := fmt.Fprintf(buf, "TN:\nSF:%s\n", location.ID())
+		if err != nil {
+			return nil, err
+		}
+		lines := make([]int, 0)
+		for line := range coverage.LineHits { // nolint:maprange
+			lines = append(lines, line)
+		}
+		sort.Ints(lines)
```
kozlovb, line 505: I wonder if you need an array of lines. We know in advance that lines range from 0 to max line of the file. 
I am not sure how code coverage works but if each line is present in the map i.e. even with 0 hits then you may find the max key, so the max line and just iterate from 0 to max line while retrieving values from the map as you ve done.

runtime/coverage.go
m-Peter: Good catch, updated in https://github.com/onflow/cadence/pull/2577/commits/f5f4602fd9dfde924aac7eb1151abcf8979d3356

runtime/coverage.go
m-Peter: Resolved offline :pray: The bottom line is that we need to only keep track of the executable lines of a file.

PR #2581, opened 2023-06-16
rename restricted type to intersection type

dsainati1:
Work towards #2639

Part of the implementation of https://github.com/onflow/flips/pull/85

This renames `RestrictedType` to `IntersectionType`, and removes references to "restricted" and "restriction" in the codebase.

______


runtime/interpreter/decode.go
```diff
@@ -1807,15 +1807,15 @@ func (d TypeDecoder) decodeDictionaryStaticType() (StaticType, error) {
 	return NewDictionaryStaticType(d.memoryGauge, keyType, valueType), nil
 }
 
-func (d TypeDecoder) decodeRestrictedStaticType() (StaticType, error) {
-	const expectedLength = encodedRestrictedStaticTypeLength
+func (d TypeDecoder) decodeIntersectionStaticType() (StaticType, error) {
```
turbolent, line 1810: I assume this and the other internal decoding functions stay backward compatible, i.e. they'll be able to decode data stored in the old format?

turbolent (approved):
👍

AFAICS, this only renames, but does not change any logic?

runtime/interpreter/decode.go
dsainati1: For the ccf and interpreter decoders, yes I believe they remain backwards compatible, although it might be worth asking @fxamacker to double check this.

The JSON decoder will change because the old fields were named things like `"restrictions"` or the tag was "RestrictedType", which have been changed to have more appropriate names.

dsainati1:
> AFAICS, this only renames, but does not change any logic?

It does not change any language behavior; however I think the external API does change a little since anywhere where the name "restricted type" appeared in the output (e.g. parser json output or json encoding).

runtime/interpreter/decode.go
fxamacker: > For the ccf and interpreter decoders, yes I believe they remain backwards compatible

Yes, CCF and also internal value encodings are backwards compatible.  :+1:

PR #2588, opened 2023-06-20
Add identity entitlement mapping

SupunS:
Work towards https://github.com/onflow/flips/pull/94

⚠️ Depends on https://github.com/onflow/cadence/pull/2586

## Description

______


runtime/tests/checker/entitlements_test.go
```diff
@@ -5132,3 +5132,32 @@ func TestCheckBuiltinEntitlements(t *testing.T) {
...
+	_, err := ParseAndCheck(t, `
+        struct S {
+            access(Identity) fun foo(): auth(Identity) &AnyStruct {
+                let a: AnyStruct = "hello"
+                return &a as auth(Identity) &AnyStruct
+            }
+        }
+
+        fun main() {
+            let s = S()
```
dsainati1, line 5149: We probably want some more tests here. In particular:

* Test that an unauthorized `&S` maps to an unauthorized `&AnyStruct`
* The identity map preserves conjunctions/disjunctions (i.e. `A, B` yields `A, B` while `C | D` yields `C | D`
* Test the behavior of an identity-mapped member when accessed on an owned value.

runtime/sema/access.go
```diff
@@ -339,6 +339,11 @@ func (e EntitlementMapAccess) entitlementImage(entitlement *EntitlementType) (ou
 // defined by the map in `e`, producing a new entitlement set of the image of the
 // arguments.
 func (e EntitlementMapAccess) Image(inputs Access, astRange func() ast.Range) (Access, error) {
+
+	if e.Type == IdentityMappingType {
+		return inputs, nil
+	}
```
dsainati1, line 345: Unfortunately I am not sure it's quite this simple.  While this works for any mapped member accessed on a reference type, when an entitlement-mapped member is accessed on an owned value, we don't use the `Image` function to compute the set of output entitlements; we just use the entire codomain of the map. However, if the map is the `Identity` map, the codomain is theoretically infinite, so we cannot do that. We need some sensible behavior in this scenario.

One possible solution would be to just produce a reference that is entitled to all the entitlements supported on the output reference type. So, for example given some definition like:

```
struct X {
   access(A) fun foo()
   access(B | C) var s: String
}
struct Y {
   access(Identity) fun getX(): auth(Identity) &X { ... }
}
```
if `getX()` is called on an owned `Y` value, the resulting reference should be authorized to all the entitlements supported by `X`, i.e. `A`, `B`, and `C`.

turbolent (dismissed):
Nice! Love how simple it was to add this 👌

runtime/sema/access.go
SupunS: Yeah, you beat me to it 😄. I was writing a comment saying we might need some more work here, but not sure what else is needed. Currently, this has the minimal set of changes needed to unblock the follow-up work on mutability.

runtime/tests/checker/entitlements_test.go
SupunS: added 7a4ff32be9dc566cba30726f1ffddf2588ccc3e8

runtime/tests/interpreter/entitlements_test.go
```diff
@@ -2972,3 +2972,111 @@ func TestInterpretBuiltinEntitlements(t *testing.T) {
...
+                    let a: AnyStruct = "hello"
+                    return &a as auth(Identity) &AnyStruct
+                }
+            }
+
+            fun main() {
+                let s = S()
+
+                // OK: Must return an unauthorized ref
+                let resultRef1: &AnyStruct = s.foo()
```
dsainati1, line 2995: It doesn't make a lot of sense to me that accessing `foo` on an owned value would give you fewer permissions than accessing it on an entitled reference.

SupunS: but unfortunately, the possible entitlement set is infinite, and currently we don't have a way to represent such an entitlement.

turbolent: Discussed this with Supun, and agree that this is the expected behaviour given the current feature definition.

Maybe in the future we could introduce a notion of "all entitlements", basically something authorized to have the same access as owner-access has. Not sure, but this is maybe what Dete mentioned with "owner" access before.

turbolent (approved):
Nice work!

As discussed, maybe address Daniel's feedback and return an entitled reference, but also feel free to to that in a follow-up PR.

runtime/sema/access.go
SupunS: While adding this, realized it conflicts with the refactoring I did in https://github.com/onflow/cadence/pull/2587.

So I will add it in a separate PR, once both this and 2587 are merged.

runtime/tests/interpreter/entitlements_test.go
SupunS: Created issue https://github.com/onflow/cadence/issues/2635 to track this.

PR #2589, opened 2023-06-20
Allow emit statements in conditions

### LINKED ISSUE #2069: Events emitted from interfaces
### Issue To Be Solved

As we discussed in the language design meeting, I made a gist of what is on my mind. 

https://gist.github.com/bluesign/2a654b7ce0cb889ff65447567e0ce069

Here the system works ( as of today ) with the trampoline function to emit the event. A language change allowing the event to be emitted directly from the interface can make it more straightforward. 

I think with the addition of interface default implementations and interfaces conforming to other interfaces, we may end up with a really nice and less bloated contract code for FTs and NFTs.

@bjartek @turbolent @dsainati1 @joshuahannan @dete @janezpodhostnik

turbolent:
Closes #2069

## Description

Implementation of https://github.com/onflow/flips/pull/111
______


turbolent:
Ready for review, but going to add some more tests

runtime/sema/check_conditions.go
```diff
@@ -40,15 +43,157 @@ func (checker *Checker) visitConditions(conditions []*ast.Condition) {
...
+		checker.VisitEmitStatement((*ast.EmitStatement)(condition))
+
+	default:
+		panic(errors.NewUnreachableError())
 	}
 
 	return nil
 }
+
+func (checker *Checker) rewritePostConditions(postConditions ast.Conditions) PostConditionsRewrite {
```
turbolent, line 69: Moved from https://github.com/onflow/cadence/pull/2589/files#diff-74d0954852bf407e9259a7c9bb55e8af093f080f00f5e35689c250886c29860aL1961 and refactored into sub-functions

runtime/sema/check_conditions.go
```diff
@@ -40,15 +43,157 @@ func (checker *Checker) visitConditions(conditions []*ast.Condition) {
...
+			messageExtraction.ExtractedExpressions...,
+		)
+	}
+
+	newPostCondition = &newPostTestCondition
+
+	return
+}
+
+func (checker *Checker) rewriteEmitPostCondition(
```
turbolent, line 170: New

runtime/interpreter/interpreter.go
```diff
@@ -823,40 +823,50 @@ func (interpreter *Interpreter) resultValue(returnValue Value, returnType sema.T
...
+	switch condition := condition.(type) {
+	case *ast.TestCondition:
+		// Evaluate the condition as a statement, so we get position information in case of an error
+		statement := ast.NewExpressionStatement(interpreter, condition.Test)
 
-	statement := ast.NewExpressionStatement(interpreter, condition.Test)
+		result, ok := interpreter.evalStatement(statement).(ExpressionResult)
 
-	result, ok := interpreter.evalStatement(statement).(ExpressionResult)
+		value, valueOk := result.Value.(BoolValue)
```
SupunS, line 841: If the above casting failed, `result` would be `nil` here

runtime/sema/check_conditions.go
```diff
@@ -40,15 +43,157 @@ func (checker *Checker) visitConditions(conditions []*ast.Condition) {
 	}
 }
 
-func (checker *Checker) checkCondition(condition *ast.Condition) Type {
+func (checker *Checker) checkCondition(condition ast.Condition) Type {
```
SupunS, line 46: Not related to this pr: The return value seems to be always `nil` and never used. Maybe clean up this function to a void return, while at it.

turbolent: Removed in 43756fb470f7774748d87866ce4057470c786bae

runtime/interpreter/interpreter.go
turbolent: Just kept the code as-is, but I get your point.

Both type assertions must always succeed (the checker ensures it), so we can make the failures unreachable cases.

Refactored in 835bd6d53e66c9dde66a6aef11d68d0fbfe79522

SupunS (approved):

dsainati1 (approved):

turbolent:
Added more tests in the last commits, but also fixed a parsing bug in [0638f55](https://github.com/onflow/cadence/pull/2589/commits/0638f5568f52361bca6bbbaad0e32512f32e0592), where condition parse errors would not be propagated properly. 

@onflow/cadence Could you please have another look, especially at this fix?

dsainati1 (approved):

turbolent:
@SupunS could you please have another look?

turbolent:
Rebased on Stable Cadence, as the parsing fix looks like a breaking change

turbolent:
Even though the FLIP says so, event emission was not considered a pure operation yet. I changed the event constructor to be always pure in [379df98](https://github.com/onflow/cadence/pull/2589/commits/379df9881d076c1cc89bcb09f1b0d2345384c8ea). cc @dsainati1

SupunS (approved):

SupunS:
Curious about what was causing the parser fix to break existing code? Were there any invalid codes that were treated as correct?

turbolent:
@SupunS yes, invalid trailing tokens were just ignored, but now the errors for them are propagated and reported properly

PR #2596, opened 2023-06-22
Remove Restricted Type from Cadence

dsainati1:
Work towards #2639

Implementation of https://github.com/onflow/flips/pull/85

This removes the `Type` field from the `RestrictedType` (renamed in a previous PR to `IntersectionType`). This also updates all tests to be compatible with the new behavior. 

______

 

runtime/interpreter/decode.go
```diff
@@ -1808,38 +1808,6 @@ func (d TypeDecoder) decodeDictionaryStaticType() (StaticType, error) {
 }
 
 func (d TypeDecoder) decodeIntersectionStaticType() (StaticType, error) {
-	const expectedLength = encodedIntersectionStaticTypeLength
```
turbolent, line 1811: The decoding function needs to stay backward-compatible, so old stored data can be migrated

turbolent: Is this what is proposed in the entitlements migration (https://github.com/onflow/flips/pull/95)?

It might make sense to decouple the decoding from the migration, and not perform this conversion/migration during decoding. This could be done by keeping a "restricted type" field in the interface set type, which is not used anywhere else but when decoding a restricted type inside of the state migration. We can then remove it later in a follow-up PR once the state was migrated.

dsainati1: Implemented in 3fcf2e9 and 685e59d

@fxamacker could you double check that these are backwards compatible 🙏

runtime/interpreter/decode.go
```diff
@@ -1856,13 +1856,27 @@ func (d TypeDecoder) decodeIntersectionStaticType() (StaticType, error) {
...
+	t, err := d.decoder.NextType()
 	if err != nil {
-		return nil, errors.NewUnexpectedError(
-			"invalid intersection static type key type encoding: %w",
-			err,
-		)
+		return nil, err
+	}
+
+	if t != cbor.NilType {
```
turbolent, line 1866: Maybe invert the then and else blocks

runtime/interpreter/decode.go
turbolent: 685e59d looks good!

3fcf2e9, backward-compatibility in CCF, is actually not needed (same for e.g. JSON) – it's an external format, we only need backward-compatibility for the internal encoding (execution state).

fxamacker (changes requested):
I mainly focused on non-test code related to CCF and internal value encoding.

encoding/ccf/ccf_test.go
```diff
@@ -6882,7 +6880,6 @@ func TestEncodeValueOfIntersectionType(t *testing.T) {
 		)
 
 		countSumIntersectionType := cadence.NewIntersectionType(
-			statsType,
 			[]cadence.Type{
 				hasCountInterfaceType,
 				hasSumInterfaceType,
```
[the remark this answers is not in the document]
dsainati1: If we don't actually need to keep CCF backwards compatible as @turbolent said above, do we still want to test these behaviors?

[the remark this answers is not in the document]
fxamacker: Good point.  We don't need to keep this test (with legacy type) as-is, if CCF backwards compatibility is not needed.

encoding/ccf/decode_type.go
```diff
@@ -560,7 +560,7 @@ func (d *Decoder) decodeIntersectionType(
 	}
 
 	// element 0: type
-	typ, err := decodeTypeFn(types)
+	legacyRestrictedType, err := d.decodeNullableTypeValue(types)
```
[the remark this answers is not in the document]
fxamacker: Thanks for fixing this! :+1:

runtime/sema/type.go
```diff
@@ -6151,120 +6145,27 @@ func checkSubTypeWithoutEquality(subType Type, superType Type) bool {
...
 
 	case *IntersectionType:
-
-		intersectionSuperType := typedSuperType.Type
-		switch intersectionSuperType {
+		// `Any` is never a subtype of a intersection type
+		switch subType {
 		case AnyResourceType, AnyStructType, AnyType:
+			return false
+		}
```
SupunS, line 6152: might not need this first switch-case.

runtime/sema/type_tags.go
```diff
@@ -935,8 +935,11 @@ func commonSuperTypeOfComposites(types []Type) Type {
 	}
 
 	if hasCommonInterface {
+		if len(commonInterfacesList) == 0 {
+			panic(errors.NewUnreachableError())
+		}
```
SupunS, line 940: this is rather unnecessary, because `hasCommonInterface` is set based on the length of `commonInterfaces`/`commonInterfacesList`

runtime/sema/type_tags.go
```diff
@@ -935,8 +935,11 @@ func commonSuperTypeOfComposites(types []Type) Type {
...
+
 		return &IntersectionType{
-			Type:  superType,
```
SupunS, line 939: I think the initialization of `superType` can now be moved below, to the end of this function

runtime/storage_test.go
```diff
@@ -466,7 +466,7 @@ func TestRuntimePublicCapabilityBorrowTypeConfusion(t *testing.T) {
 
         // Create a public capability to the stored Vault that only exposes
         // the balance field through the Balance interface
-        self.account.link<&DapperUtilityCoin.Vault{FungibleToken.Balance}>(
+        self.account.link<&DapperUtilityCoin.Vault>(
```
SupunS, line 469: Here and below, I think it might be better to keep the interface set here
```suggestion
        self.account.link<&{FungibleToken.Balance}>(
```

runtime/ft_test.go
```diff
@@ -371,15 +371,15 @@ access(all) contract FlowToken: FungibleToken {
         // Create a public capability to the stored Vault that only exposes
         // the 'deposit' method through the 'Receiver' interface
         //
-        adminAccount.link<&FlowToken.Vault{FungibleToken.Receiver}>(
+        adminAccount.link<&FlowToken.Vault>(
```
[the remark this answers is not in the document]
dsainati1: This doesn't matter anymore with entitlements; you can no longer "expose the whole `Vault`" because the reference here is still unauthorized. This is the basis for the planned migration to remove restricted types; any types of the form `T{I}` are just going to be replaced with `T` because it's no longer unsafe to give out a reference to the whole type as long as you use proper entitlements.

runtime/sema/check_casting_expression.go
```diff
@@ -204,192 +204,53 @@ func FailableCastCanSucceed(subType, superType Type) bool {
...
-					return true
-				default:
-					return typedSubType.Type.Equal(typedSuperType.Type)
-				}
-			}
+			// A intersection type `{Us}`
+			// is a subtype of a intersection type `{Vs}`:
+			// if the run-time type conforms to `Vs`
+			// `Us` and `Vs` do *not* have to be subsets.
+			return true
```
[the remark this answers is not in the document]
dsainati1: Yeah, this is already checked before this point.

runtime/storage_test.go
dsainati1: See previous comment, this is no longer the suggested best practice post-entitlements.

runtime/ft_test.go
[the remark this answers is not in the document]
SupunS: ah ok, makes sense!

PR #2609, opened 2023-06-26
Add utility function `ccf.HasMsgPrefix`

### LINKED ISSUE #2608: Provide function to detect if data has CCF message prefix
It would be useful to check if data begins with CCF message prefix (first few bytes) before attempting to decode it.

One possible solution is to add `ccf.HasMsgPrefix(msg []byte) bool` which would return true if `msg` has matching prefix without incurring more expensive checks for well-formedness or validity.

Thanks @peterargue for inquiring about CCF magic prefix on Friday.

fxamacker:
Closes #2608 

## Description

`HasMsgPrefix` returns `true` if the msg prefix (first few bytes) matches currently implemented top-level CCF messages.

For simplicity and performance, `HasMsgPrefix` does not check if msg is actually a CCF message, or well-formed, or valid.

cc: @peterargue 
______

### DIFF
--- a/encoding/ccf/decode.go
+++ b/encoding/ccf/decode.go
@@ -32,12 +32,6 @@ import (
 	cadenceErrors "github.com/onflow/cadence/runtime/errors"
 )
 
-const (
-	// Top level CCF messages are CBOR tag (tag number: CBORTagTypeDefAndValue | CBORTagTypeAndValue, content: array of 2 elements).
-	// Minimum size is 5 bytes: top level tag number (2 bytes) + min size for array of 2 elements (3 bytes).
-	minTopLevelMsgSize = 5
-)
-
 // HasMsgPrefix returns true if the msg prefix (first few bytes)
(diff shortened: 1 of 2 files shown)


dsainati1 (approved):

encoding/ccf/decode.go
```diff
@@ -32,6 +32,41 @@ import (
 	cadenceErrors "github.com/onflow/cadence/runtime/errors"
 )
 
+const (
+	// Top level CCF messages are CBOR tag (tag number: CBORTagTypeDefAndValue | CBORTagTypeAndValue, content: array of 2 elements).
+	// Minimum size is 5 bytes: top level tag number (2 bytes) + min size for array of 2 elements (3 bytes).
+	minTopLevelMsgSize = 5
+)
```
turbolent, line 39: Given that it is not exported and only used in `HasMsgPrefix`, maybe move it into the function

turbolent (approved):
Nice!

PR #2622, opened 2023-06-30
Delegate storage address to Atree

fxamacker:
Updates https://github.com/onflow/atree/issues/292

## Description

Currently, storage address is derived from `StorageID()`.

In the future, some Atree values can be inlined so they don't have `StorageID` and address needs be derived differently.

This PR delegates storage address to its internal Atree values, by calling `Address()`.

______

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -2583,11 +2583,10 @@ func (v *ArrayValue) Transfer(
 	}
 
 	currentStorageID := v.StorageID()
-	currentAddress := currentStorageID.Address
 
 	array := v.array
 
-	needsStoreTo := address != currentAddress
+	needsStoreTo := v.NeedsStoreTo(address)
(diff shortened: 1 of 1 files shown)


turbolent (approved):

turbolent:
Is there maybe a way we can also prevent future uses of `StorageID().Address`, i.e. force the use of `StorageAddress()`?

fxamacker:
> Is there maybe a way we can also prevent future uses of StorageID().Address, i.e. force the use of StorageAddress()?

@turbolent Good point! :+1:  It would be great to unexport fields in `StorageID`.  Since that would be a breaking change in atree API, I will open a separate PR targeting feature branch.

turbolent:
@fxamacker Sounds good! It's OK if there are breaking changes in the API, we can adjust Cadence accordingly, only breaking changes to the storage format (i.e. need for storage migration) are a concern

Issue #2605, opened 2023-06-23
Implement additional Array & Variable-size Array functions


darkdrag00nv2:
### Issue to be solved

Additional functions can be implemented on the Array & variable-size Array types.

For `[T]` or `[T; N]`, these include:
1. `reverse()`: Reverses the array by mutating it. Available on both Array and Variable-size Arrays
2.  `map<U: AnyStruct>(_ f: (T: U)) : [U; N]` and `map<T: AnyStruct, U: AnyStruct>(_ f: (T: U)) : [U]`. Will create a new Array without mutating
3. `filter(_ pred: (T: Bool)) : [U]`. Only available on variable-size Arrays. Will create a new value without mutating

1 and 2 are taken from https://github.com/green-goo-dao/flow-utils/blob/main/cadence/contracts/ArrayUtils.cdc.

We could also add `zip`, `flatten` and others but I think they are not needed right now. We can wait for requests for those before implementing them.

turbolent:
Good idea!

For functions that take a function which is passed the elements, e.g. `map`, `transform` and `filter`:
For resource-kinded inputs, the callbacks can't be the resources themselves, but could be references, or the functions are not defined for such inputs (already done for other functions). 

For `transform` in particular cannot exist for resource-kinded inputs: The result would overwrite the existing resource element, implicitly destroying it, which is not allowed.

For `filter`, the type parameter for `U` should be defined. Again, for resource-kinded inputs, the function should probably return references, or not be defined.

darkdrag00nv2:
Thanks for the insights.

> For filter, the type parameter for U should be defined.

It was a typo. I meant to write `T` instead of `U`. Basically `filter` on `[T]` will also return `[T]`.

darkdrag00nv2:
@turbolent I started looking into `reverse`.

In the issue, I've proposed that we mutate the array itself. To do that we probably need support for reverse in `atree` itself.

Alternatively, we can have the function return a new array with entries reversed (except for resources). Similar to what the function in https://github.com/green-goo-dao/flow-utils/blob/main/cadence/contracts/ArrayUtils.cdc does.

Thoughts?

turbolent:
Starting with either variant works (mutating in-place or returning a new value) 

It might be a good idea to name the functions accordingly, e.g. Swift has `reverse` for the in-place operation, and `reversed` for the one returning a new value (likewise, e.g. `sort` and `sorted`).

[...]

PR #2626, opened 2023-07-01
Introduce `reverse` in Variable/Fixed size Array

darkdrag00nv2:
Work towards #2605 

## Description
Introduce `reverse` function for in-place reversal of a Variable/Fixed size Array value.

______

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -2192,6 +2192,18 @@ func (v *ArrayValue) FirstIndex(interpreter *Interpreter, locationRange Location
 	return NilOptionalValue
 }
 
+func (v *ArrayValue) Reverse(interpreter *Interpreter, locationRange LocationRange) VoidValue {
+	count := v.Count()
+
+	for i := 0; i < count/2; i++ {
+		tempValue := v.Get(interpreter, locationRange, i)
+		v.Set(interpreter, locationRange, i, v.Get(interpreter, locationRange, count-i-1))
+		v.Set(interpreter, locationRange, count-i-1, tempValue)
+	}
+
+	return VoidValue{}
+}
+
 func (v *ArrayValue) Contains(
 	interpreter *Interpreter,
 	locationRange LocationRange,
@@ -2367,6 +2379,18 @@ func (v *ArrayValue) GetMember(interpreter *Interpreter, locationRange LocationR
 					invocation.Interpreter,
 					invocation.LocationRange,
 					invocation.Arguments[0],
+				)
+			},
+		)
+
+	case "reverse":
+		return NewHostFunctionValue(
+			interpreter,
+			sema.ArrayReverseFunctionType,
+			func(invocation Invocation) Value {
+				return v.Reverse(
+					invocation.Interpreter,
+					invocation.LocationRange,
 				)
 			},
 		)
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -1714,6 +1714,10 @@ Returns the index of the first element matching the given object in the array, n
 Available if the array element type is not resource-kinded and equatable.
 `
 
+const arrayTypeReverseFunctionDocString = `
+Reverses the elements of the array.
+`
+
 const arrayTypeContainsFunctionDocString = `
 Returns true if the given object is in the array
 `
@@ -1862,6 +1866,18 @@ func getArrayMembers(arrayType ArrayType) map[string]MemberResolver {
 					identifier,
 					ArrayFirstIndexFunctionType(elementType),
 					arrayTypeFirstIndexFunctionDocString,
+				)
+			},
+		},
+		"reverse": {
+			Kind: common.DeclarationKindFunction,
+			Resolve: func(memoryGauge common.MemoryGauge, identifier string, targetRange ast.Range, report func(error)) *Member {
+				return NewPublicFunctionMember(
+					memoryGauge,
+					arrayType,
+					identifier,
+					ArrayReverseFunctionType,
+					arrayTypeReverseFunctionDocString,
 				)
 			},
 		},
@@ -2116,6 +2132,12 @@ func ArrayFirstIndexFunctionType(elementType Type) *FunctionType {
 		),
 	}
 }
+
+var ArrayReverseFunctionType *FunctionType = &FunctionType{
+	Parameters:           []Parameter{},
+	ReturnTypeAnnotation: NewTypeAnnotation(VoidType),
+}
+
 func ArrayContainsFunctionType(elementType Type) *FunctionType {
 	return &FunctionType{
 		Parameters: []Parameter{
--- a/runtime/tests/interpreter/interpreter_test.go
+++ b/runtime/tests/interpreter/interpreter_test.go
@@ -10401,7 +10401,7 @@ func TestInterpretArrayReverse(t *testing.T) {
 
 	inter := parseCheckAndInterpret(t, `
       let xs = [1, 2, 3, 100, 200]
-	  let ys = [100, 467, 297, 23]
+      let ys = [100, 467, 297, 23]
 
       fun reversexs() {
           return xs.reverse()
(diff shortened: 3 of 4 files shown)


runtime/interpreter/value.go
```diff
@@ -2192,6 +2192,18 @@ func (v *ArrayValue) FirstIndex(interpreter *Interpreter, locationRange Location
 	return NilOptionalValue
 }
 
+func (v *ArrayValue) Reverse(interpreter *Interpreter, locationRange LocationRange) VoidValue {
+	count := v.Count()
+
+	for i := 0; i < count/2; i++ {
+		tempValue := v.Get(interpreter, locationRange, i)
+		v.Set(interpreter, locationRange, i, v.Get(interpreter, locationRange, count-i-1))
+		v.Set(interpreter, locationRange, count-i-1, tempValue)
```
darkdrag00nv2, line 2201: Wondering if we need some special handling in case something goes wrong in between the swap. 

Say the first `Set` is successful while the second `Set` panics. Not sure why that would happen but that could lead to both index `i` and `count - i - 1` having the same value.

turbolent: If something goes wrong, i.e. panics, execution aborts and all state changes are thrown away, so there is no need to revert anything manually.

darkdrag00nv2: Nice, thanks for sharing.

turbolent, line 2201: Maybe make naming a bit more clear:

```suggestion
	for leftIndex := 0; leftIndex < count/2; leftIndex++ {
		rightIndex := count - leftIndex - 1

		leftValue := v.Get(interpreter, locationRange, leftIndex)
		rightValue := v.Get(interpreter, locationRange, rightIndex)
		
		v.Set(interpreter, locationRange, leftIndex, rightValue)
		v.Set(interpreter, locationRange, rightIndex, leftValue)
```

runtime/interpreter/value.go
```diff
@@ -2371,6 +2383,18 @@ func (v *ArrayValue) GetMember(interpreter *Interpreter, locationRange LocationR
 			},
 		)
 
+	case "reverse":
```
turbolent, line 2386: Define a `const ArrayTypeReverseFunctionName = "reverse"` in `sema`, and then use it there and here

runtime/sema/type.go
```diff
@@ -2116,6 +2132,12 @@ func ArrayFirstIndexFunctionType(elementType Type) *FunctionType {
 		),
 	}
 }
+
+var ArrayReverseFunctionType *FunctionType = &FunctionType{
+	Parameters:           []Parameter{},
```
turbolent, line 2137: No need to allocate a slice here:

```suggestion
```

runtime/interpreter/value.go
```diff
@@ -2192,6 +2192,18 @@ func (v *ArrayValue) FirstIndex(interpreter *Interpreter, locationRange Location
...
+func (v *ArrayValue) Reverse(interpreter *Interpreter, locationRange LocationRange) VoidValue {
+	count := v.Count()
+
+	for i := 0; i < count/2; i++ {
+		tempValue := v.Get(interpreter, locationRange, i)
+		v.Set(interpreter, locationRange, i, v.Get(interpreter, locationRange, count-i-1))
+		v.Set(interpreter, locationRange, count-i-1, tempValue)
+	}
+
+	return VoidValue{}
```
turbolent, line 2204:
```suggestion
	return Void
```

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -10395,6 +10395,67 @@ func TestInterpretArrayFirstIndexDoesNotExist(t *testing.T) {
 	)
 }
 
+func TestInterpretArrayReverse(t *testing.T) {
```
turbolent, line 10398: Could you please also add test cases where the elements are structs and resources?

turbolent (commented):
Nice!

runtime/interpreter/value.go
darkdrag00nv2: Done.

runtime/tests/interpreter/interpreter_test.go
darkdrag00nv2: Hmm, thanks for pointing it out.

When I added the test case for struct, it doesn't work.

```cadence
fun reverseStructArray(): [Int] {
  let sa = [TestStruct(1), TestStruct(2), TestStruct(3)]
  sa.reverse()
  
  let res: [Int] = [];
  for s in sa {
    res.append(s.test)
  }

  return res
}
```

I get `error: member test is used before it has been initialized`. So it seems like using `Get` and `Set` messes up the struct members.

darkdrag00nv2: By adding transfer call, I was able to get it to work for struct.

```go
leftValue := v.Get(interpreter, locationRange, leftIndex)
rightValue := v.Get(interpreter, locationRange, rightIndex)

leftValue = leftValue.Transfer(
	interpreter,
	locationRange,
	v.array.Address(),
	true,
	nil,
)
rightValue = rightValue.Transfer(
	interpreter,
	locationRange,
	v.array.Address(),
	true,
	nil,
)

v.Set(interpreter, locationRange, leftIndex, rightValue)
v.Set(interpreter, locationRange, rightIndex, leftValue)
```

It doesn't work for resource because after the first `Set` call, we end up with two resources at the `leftIndex` which returns the error `// error: two parents are captured for the slab`.

I'll look further into how we swap resources atomically. Another option might be to just not support reverse for resource typed arrays. Most of the array functions do that already.

SupunS: So the problem is, structs and resources behave differently when "Transferred". Structs are copied, whereas references are just 'moved'. Here what you would want to do is, instead of `v.Get`, do a `v.Remove`, which will "remove" the value from the array. Then instead of setting (i.e: `v.Set`), do a `v.Insert`, because by removing earlier, we literally remove the value from the array, hence the values are shifted by one index.

Now one thing to note is that, because the remaining values are shifted by one index by removing, if you try to remove the first element (leftValue) first, and then try to remove the last element (rightElement), it will mess-up the indexing for the second/right value removal. (could get an array-out-of-bound error / or could end up removing the wrong value) because after the first removal, the size of the array is (n-1).
So, you'll have to swap the removals, to first remove the right index first, and then remove the left index.
i.e:
```go
// Remove the right index (from the rear end) first, because removing the left index (from the front of the array) first could mess up the indexes of the rest of the elements.
rightValue := v.Remove(interpreter, locationRange, rightIndex)
leftValue := v.Remove(interpreter, locationRange, leftIndex)

v.Insert(interpreter, locationRange, leftIndex, rightValue)
v.Insert(interpreter, locationRange, rightIndex, leftValue)
```

SupunS: Also no need for an additional "Transfer" because `Remove` and `Insert` already do a transform underneath.

bluesign: This feels too expensive to me; transfers here are not necessary, Why move stuff to stack and then move again to strange with totally new storageID and slab tree?   

We are 100% sure that storage address will not change in this operation. Why not add `swap` to `atree.Array` ? @fxamacker can confirm, but I think this way, all array is read and written again.

darkdrag00nv2: I thought so as well and asked Bastian about adding support for `reverse` in atree on https://github.com/onflow/cadence/issues/2605#issuecomment-1615164488. Didn't ask about `swap` but that will work nicely as well.

We can also think about adding it using `Remove` and `Insert` and later optimizing it using support in atree.

darkdrag00nv2: @SupunS Makes sense, opened up https://github.com/onflow/cadence/pull/2654 for copy-based reversal.

darkdrag00nv2:
Have decided to update the function name to `reverseInPlace` based on the suggestion on https://github.com/onflow/cadence/pull/2654.

runtime/tests/interpreter/interpreter_test.go
fxamacker: > We are 100% sure that storage address will not change in this operation. Why not add swap to atree.Array ? 

Good point :+1: .  I opened https://github.com/onflow/atree/issues/326 to add `Array.Swap`.

PR #2632, opened 2023-07-05
Improve cyclic import handling

### LINKED ISSUE #2631: Can create cyclic imports by importing the self address
### Issue to be solved

When a contract imports an address that is the same address as the contract is deployed to, it creates a cycle right after the deployment.

e.g:
Assume the below contract is deployed to `0x01`.
```cadence
import 0x01

access(all) contract Foo {}
```

Since there are no contracts existing in `0x01`, there is no cycle before the deployment. 
But as soon as `Foo` is delpoyed, then importing the same address as itself, would be equivalent to importing `Foo` itself.

### Suggested Solution

Reject these cyclic imports at the time of deployment.

SupunS:
Closes #2631

## Description

When a contract imports an address that is the same address as the contract is deployed to, it creates a cycle right after the deployment.

e.g:
Assume the below contract is deployed to `0x01`.

```cadence
import 0x01

access(all) contract Foo {}
```

Since there are no contracts existing in `0x01`, there is no cycle before the deployment. 
But as soon as `Foo` is delpoyed, then importing the same address as itself, would be equivalent to importing `Foo` itself.

This PR reject such deployments.

______

### DIFF
--- a/runtime/sema/check_import_declaration.go
+++ b/runtime/sema/check_import_declaration.go
@@ -45,7 +45,7 @@ func (checker *Checker) VisitImportDeclaration(declaration *ast.ImportDeclaratio
 	})
 }
 
-func (checker *Checker) declareImportDeclaration(declaration *ast.ImportDeclaration) Type {
+func (checker *Checker) declareImportDeclaration(declaration *ast.ImportDeclaration) {
 	locationRange := ast.NewRange(
 		checker.memoryGauge,
 		declaration.LocationPos,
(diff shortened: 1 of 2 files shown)


runtime/sema/check_import_declaration.go
```diff
@@ -45,27 +45,52 @@ func (checker *Checker) VisitImportDeclaration(declaration *ast.ImportDeclaratio
...
+	// Then this could create a cycle, as soon the current contract is deployed.
+	// e.g:
+	//    import 0x01
+	//    access(all) contract Foo {}
+	//
+
+	if len(identifiers) == 0 {
+		if currentProgramLocation, ok := checker.Location.(common.AddressLocation); ok {
+			if importLocation, ok := declaration.Location.(common.AddressLocation); ok {
+				if currentProgramLocation.Address == importLocation.Address {
```
turbolent, line 75: Is there a particular reason why we do this for address locations specifically? Should we generalize this to any location kind?

SupunS: Not sure if we need this for other location types: i.e. AFAIU this can only occur for deployed contracts (hence address locations)

turbolent (approved):

PR #2637, opened 2023-07-06
Return a reference with all entitlements, when access through owned values

### LINKED ISSUE #2536: Improve entitlement mappings
### Issue to be solved

This is to track the implementation of https://github.com/onflow/flips/pull/94

### Suggested Solution

_No response_

SupunS:
Work towards #2638
Work towards onflow/flips#94
Closes #2536

## Description

Adds the improvement discussed in https://github.com/onflow/cadence/pull/2588#discussion_r1237203195

______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -5251,7 +5251,7 @@ func (interpreter *Interpreter) getMemberWithAuthMapping(self Value, locationRan
 		return nil
 	}
 	// once we have obtained the member, if it was declared with entitlement-mapped access, we must compute the output of the map based
-	// on the runtime authorizations of the acccessing reference or composite
+	// on the runtime authorizations of the accessing reference or composite
 	memberAccess := interpreter.getAccessOfMember(self, identifier)
 	return interpreter.mapMemberValueAuthorization(self, memberAccess, result)
 }
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -5215,10 +5215,17 @@ func (interpreter *Interpreter) getAccessOfMember(self Value, identifier string)
 	return &member.Resolve(interpreter, identifier, ast.EmptyRange, func(err error) {}).Access
 }
 
-func (interpreter *Interpreter) mapMemberValueAuthorization(self Value, memberAccess *sema.Access, resultValue Value) Value {
+func (interpreter *Interpreter) mapMemberValueAuthorization(
+	self Value,
+	memberAccess *sema.Access,
+	resultValue Value,
+	resultingType sema.Type,
+) Value {
+
 	if memberAccess == nil {
 		return resultValue
 	}
+
 	if mappedAccess, isMappedAccess := (*memberAccess).(sema.EntitlementMapAccess); isMappedAccess {
 		var auth Authorization
 		switch selfValue := self.(type) {
@@ -5230,7 +5237,16 @@ func (interpreter *Interpreter) mapMemberValueAuthorization(self Value, memberAc
 			}
 			auth = ConvertSemaAccesstoStaticAuthorization(interpreter, imageAccess)
 		default:
-			auth = ConvertSemaAccesstoStaticAuthorization(interpreter, mappedAccess.Codomain())
+			var access sema.Access
+			if mappedAccess.Type == sema.IdentityMappingType {
+				access = sema.AllSupportedEntitlements(resultingType)
+			}
+
+			if access == nil {
+				access = mappedAccess.Codomain()
+			}
+
+			auth = ConvertSemaAccesstoStaticAuthorization(interpreter, access)
 		}
 
 		switch refValue := resultValue.(type) {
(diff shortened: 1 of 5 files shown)


runtime/sema/check_member_expression.go
```diff
@@ -373,33 +373,12 @@ func (checker *Checker) visitMember(expression *ast.MemberExpression) (accessedT
 
 // isReadableMember returns true if the given member can be read from
 // in the current location of the checker, along with the authorzation with which the result can be used
-func (checker *Checker) isReadableMember(accessedType Type, member *Member, accessRange func() ast.Range) (bool, Access) {
-	var mapAccess func(EntitlementMapAccess, Type) (bool, Access)
-	mapAccess = func(mappedAccess EntitlementMapAccess, accessedType Type) (bool, Access) {
```
SupunS, line 378: This function didn't need to be a closure, so extracted it to a top-level function

turbolent (approved):
Nice!

turbolent:
It would be good to get a review from @dsainati1 here before merging

runtime/sema/check_member_expression.go
```diff
@@ -443,12 +422,69 @@ func (checker *Checker) isReadableMember(accessedType Type, member *Member, acce
...
+}
+
+func AllSupportedEntitlements(typ Type) Access {
+	switch typ := typ.(type) {
+	case *ReferenceType:
+		return AllSupportedEntitlements(typ.Type)
+	case *OptionalType:
+		return AllSupportedEntitlements(typ.Type)
+	case *FunctionType:
+		return AllSupportedEntitlements(typ.ReturnTypeAnnotation.Type)
```
dsainati1, line 473: This is used for accessor functions right? That use case makes sense, but we should add a test to make sure that a function-typed field with an identity mapping should not get the entitlements to return type. I.e. some field defined with `access(Identity) let fn: ((): T)` should not be entitled to `T` when it's accessed through an owned value.

runtime/interpreter/interpreter.go
```diff
@@ -5230,7 +5237,16 @@ func (interpreter *Interpreter) mapMemberValueAuthorization(self Value, memberAc
 			}
 			auth = ConvertSemaAccesstoStaticAuthorization(interpreter, imageAccess)
 		default:
-			auth = ConvertSemaAccesstoStaticAuthorization(interpreter, mappedAccess.Codomain())
+			var access sema.Access
+			if mappedAccess.Type == sema.IdentityMappingType {
```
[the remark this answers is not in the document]
SupunS: Can you add this requirement to the feature-request https://github.com/onflow/cadence/issues/2643.

(I guess there's nothing to add/change as for now / this PR?)

runtime/tests/checker/entitlements_test.go
```diff
@@ -5454,4 +5454,65 @@ func TestCheckIdentityMapping(t *testing.T) {
...
+                   self.s = "hello"
+               }
+
+               access(C) fun foo() {}
+            }
+
+            struct Y {
+
+                // Reference
+                access(Identity) var x1: auth(Identity) &X
```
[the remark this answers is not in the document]
dsainati1: Oh I just realized why this doesn't make sense; if the value is owned we won't produce a reference. Nevermind this 😄

runtime/sema/check_member_expression.go
SupunS: Didn't quite understand the question. For an owned value, `fn` won't return a reference, so we are not actually computing any entitlements for reference, the field is readable for anyone since the value is owned, correct?

On the other hand, it is not allowed to use Mappings on function-typed fields. Added a test e2ccbb0edf51eee1f7af6a8f520fa31b4875252d.

runtime/tests/checker/entitlements_test.go
```diff
@@ -5515,4 +5515,44 @@ func TestCheckIdentityMapping(t *testing.T) {
...
+            }
+
+            fun main() {
+                let y = Y()
+                let v = y.fn()
+            }
+        `)
+
+		errors := RequireCheckerErrors(t, err, 1)
+		invalidMapping := &sema.InvalidMappedEntitlementMemberError{}
```
dsainati1, line 5555: Oh hmm, if functions are just disallowed here maybe use a reference to a function type so we can actually test the mapping behavior.

runtime/sema/check_member_expression.go
dsainati1: Sorry, I should have been clearer. In https://github.com/onflow/cadence/pull/2637#discussion_r1258458218 I suggested adding a field with a `&((): T)` type to exercise this case.

SupunS: I see what you mean.

Weird enough, in stable cadence (not just this feature branch), accessing a field of type `&((): T)`, returns a `((): T)`. Don't know if this was changed intentionally, or was a regression (to me it seems like a regression, and probably needs fixing). So the use-case you mentioned, is currently not possible. But if we change that other behavior, then yeah, we might run into what you said.

I will add a fix to that hypothetical scenario for now, so it should seamlessly work, once the other issue is fixed.

Good catch nevertheless!

SupunS: added in 842a9cd482a53e47943756ca6817c45774c531ea

runtime/tests/checker/entitlements_test.go
SupunS: added in 842a9cd482a53e47943756ca6817c45774c531ea

dsainati1 (approved):
Looks good to me! Thanks for investigating all the issues here.

Issue #2643, opened 2023-07-06
Entitlement mapping composition


SupunS:
### Issue to be solved

As discussed in the FLIP https://github.com/onflow/flips/pull/94, add support for including/re-using entitlement mappings in other mappings by means of 'composition'.

e.g:
```cadence
entitlement mapping M1 {
    Removable -> Mutable
}

entitlement mapping M2 {
    Insertable -> Mutable
}

entitlement mapping M3 {
    include M1
    include M2
}
```

Issue #2642, opened 2023-07-06
Cadence 1.0 (formerly known as "Stable Cadence")


turbolent:
```[tasklist]
### 🚨 Testnet Upgrade blockers
- [x] Release new Emulator with https://github.com/onflow/flow-go/pull/6422
```

```[tasklist]
### Features
```

```[tasklist]
### Tech Debts / Chore
- [ ] https://github.com/onflow/cadence/issues/3335
- [ ] https://github.com/onflow/cadence-lang.org/issues/166
- [ ] https://github.com/onflow/cadence/issues/3268
- [ ] https://github.com/onflow/cadence/issues/2503
- [ ] https://github.com/onflow/cadence/issues/2719
- [ ] Remove uses of and definitions for AuthAccount/PublicAccount types after migrations
- [ ] Account export (https://www.notion.so/dapperlabs/Cadence-Implementation-Sync-Notes-Agenda-640389e72df942668d89deadd8d291de?pvs=4#c854ad5cc91b4a88922557449f5be3b9)
- [ ] https://github.com/onflow/cadence/issues/2766
- [ ] https://github.com/onflow/cadence/issues/2726
- [ ] Improve documentation: https://github.com/onflow/cadence-lang.org/issues/26
- [ ] https://www.notion.so/dapperlabs/Cadence-Implementation-Sync-Notes-Agenda-640389e72df942668d89deadd8d291de?pvs=4#72e9a6394ad44c3883f5d515772f95e3
- [ ] Improve documentation: Full pass of language reference, follow up on e.g. tutorial
- [ ] https://github.com/onflow/cadence/issues/3108
- [ ] https://github.com/onflow/cadence/issues/3113
- [ ] https://github.com/onflow/cadence/issues/3114
- [ ] https://github.com/dapperlabs/cadence-internal/issues/187
- [ ] https://github.com/onflow/cadence/issues/3206
```

```[tasklist]
### Maybe? (on-hold)
- [ ] https://github.com/onflow/cadence/issues/2170
- [ ] https://github.com/onflow/cadence/issues/2953
- [ ] https://github.com/dapperlabs/cadence-internal/pull/116
```

turbolent:
@onflow/cadence I wonder if we maybe want to clear out all the done items in the list, to only leave what is left, so it is easier to see?

SupunS:
I moved some on-hold items to a separate section `Maybe?(on-hold)`, to make it more clear on the "must do" items.

[...]

PR #2648, opened 2023-07-07
Account type

### LINKED ISSUE #2641: Refactor Account Types
```[tasklist]
### Tasks
- [x] FLIP: https://github.com/onflow/flips/pull/92
- [ ] Implementation: https://github.com/onflow/cadence/pull/2648
```

turbolent:
Closes #2641

## Description

- [x] Improve code generator:
  - [x] Add support for entitlement mappings 
  - [x] Add support for entitlement mapping entries
  - [x] Add support for non-primitive access 
  - [x] Improve generation of nested types (avoid name clashes in generated code) 
- [x] Checker:
  - [x] Refactor `AuthAccount` and `PublicAccount` into `Account` type
  - [x] Replace `AuthAccount` and `PublicAccount` uses with variations of `&Account` type
  - [x] Replace uses of `AuthAccount` storage functions to `Account.Storage` functions
- [x] Interpreter, stdlib, runtime:
  - [x] Refactor account storage functions into `Account.Storage`
  - [x] Refactor `AuthAccount` value and related nested value implementations (e.g. `AuthAccount.Capabilities`) to `Account`
  - [x] Remove `PublicAccount` value and related nested value implementations (e.g. `PublicAccount.Capabilities`)
  - [x] Adjust runtime environment
  - [x] Refactor `AuthAccount` constructor to `Account` constructor
  - [x] Adjust conversion of addresses to account values in contract invocation
  - [x] Refactor account-related types in `cadence` package
- [x] Types and conversion for `StorageCapabilityController` and `AccountCapabilityController`
- [x] Types and conversion for all entitlements
  - `Mutate`
  - `Insert`
  - `Remove`
  - `Identity`
  - `Storage`
  - `SaveValue`
  - `LoadValue`
  - `BorrowValue`
  - `Contracts`
  - `AddContract`
  - `UpdateContract`
  - `RemoveContract`
  - `Keys`
  - `AddKey`
  - `RevokeKey`
  - `Inbox`
  - `PublishInboxCapability`
  - `UnpublishInboxCapability`
  - `ClaimInboxCapability`
  - `Capabilities`
  - `StorageCapabilities`
  - `AccountCapabilities`
  - `PublishCapability`
  - `UnpublishCapability`
  - `GetStorageCapabilityController`
  - `IssueStorageCapabilityController`
  - `GetAccountCapabilityController`
  - `IssueAccountCapabilityController`
- [x] CCF
- [x] JSON
- [x] Test resource `owner` value
- [x] Test `getAuthAccount` with more entitlements

______


encoding/ccf/simpletype.go
```diff
@@ -0,0 +1,516 @@
...
+	SimpleType_Count
+)
+
+// NOTE: cadence.FunctionType isn't included in simpleTypeIDByType
+// because this function is used by both inline-type and type-value.
+// cadence.FunctionType needs to be handled differently when this
+// function is used by inline-type and type-value.
+func simpleTypeIDByType(typ cadence.Type) (SimpleType, bool) {
+
+	switch typ {
```
turbolent, line 149: Maybe refactor into Go bi-map

runtime/account_test.go
```diff
@@ -377,7 +375,7 @@ func TestRuntimeAuthAccountKeys(t *testing.T) {
 		assert.Nil(t, testEnv.storage.returnedKey)
 	})
 
-	t.Run("get key count", func(t *testing.T) {
+	t.Run("get key count afte revocation", func(t *testing.T) {
```
turbolent, line 378:
```suggestion
	t.Run("get key count after revocation", func(t *testing.T) {
```

runtime/convertTypes.go
```diff
@@ -49,139 +49,212 @@ func ExportMeteredType(
...
-			return cadence.TheAddressType
-		case *sema.ReferenceType:
-			return exportReferenceType(gauge, t, results)
-		case *sema.IntersectionType:
-			return exportIntersectionType(gauge, t, results)
-		case *sema.CapabilityType:
-			return exportCapabilityType(gauge, t, results)
-		}
-
 		switch t {
```
turbolent, line 52: Maybe replace with a mapping

runtime/interpreter/statictype.go
```diff
@@ -95,19 +82,14 @@ func (t CompositeStaticType) String() string {
 	if t.Location == nil {
 		return t.QualifiedIdentifier
 	}
-	return string(t.TypeID)
```
turbolent, line 98: Changed to what is done for `InterfaceStaticType`

runtime/sema/check_event_declaration.go
```diff
@@ -61,6 +61,9 @@ func IsValidEventParameterType(t Type, results map[*Member]bool) bool {
 	case *AddressType:
 		return true
 
+	case *ReferenceType:
```
SupunS, line 64: How do references get serialized?

runtime/sema/errors.go
```diff
@@ -3036,7 +3036,7 @@ func (*InvalidAccessError) IsUserError() {}
 
 func (e *InvalidAccessError) Error() string {
 	return fmt.Sprintf(
-		"cannot access `%s`: %s has %s access",
+		"cannot access `%s`: %s requires `%s` authorization",
```
SupunS, line 3039: should the field `RestrictingAccess` be also changed accordingly? e.g. `RestrictingAccess` -> `RequiredAccess`

runtime/sema/gen/main.go
```diff
@@ -1612,9 +1791,7 @@ func gen(inPath string, outFile *os.File, registerTypes bool) {
 		_ = ast.AcceptDeclaration[struct{}](declaration, &gen)
 	}
 
-	if registerTypes {
-		gen.generateTypeInit(program)
-	}
```
SupunS, line 1617: Does this no longer create conflicts/duplicate type declarations?

SupunS (commented):
Looks good so far! I checked most of the implementation code. What's remaining is the tests. Will check them in the next iteration.

runtime/sema/errors.go
turbolent: This was a temporary improvement. @dsainati1 has now improved this and other error messages for entitlements in #2730

runtime/sema/check_event_declaration.go
turbolent: Correct, it was possible to export the account types, but now the account "value" is a reference, so we should maybe keep support for allowing the use in an event.

I'll add some more tests around this.

runtime/sema/gen/main.go
turbolent: As far as I could see, previously the init code was always generated when the tool was run https://github.com/onflow/cadence/pull/2648/files/8c56d2d3f8960d7be25dc1224aacceb63a8c03a3#diff-9b1bcb947820865585b33db840df55c910a9e4edda99592b8983958d14998112L1658-L1661, but was never generated for the test cases https://github.com/onflow/cadence/pull/2648/files/8c56d2d3f8960d7be25dc1224aacceb63a8c03a3#diff-5758071b202796a3a1b0e6809ead97950514807dc7950b646160a23d3803498cL53-L56. 

I felt like it's better to always test all code that gets generated. When I was working on the code generation part for the init function, a run of the tests passed, but then a run of the tool crashed. It turned out that I had made some mistakes, but those weren't encountered during the tests.

runtime/sema/account.cdc
```diff
@@ -0,0 +1,521 @@
...
+    access(AccountMapping)
+    let keys: Account.Keys
+
+    /// The inbox allows bootstrapping (sending and receiving) capabilities.
+    access(AccountMapping)
+    let inbox: Account.Inbox
+
+    /// The capabilities of the account.
+    access(AccountMapping)
+    let capabilities: Account.Capabilities
```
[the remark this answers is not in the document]
dsainati1: > It would be nice if there was a way for a field to restrict access to a field using an entitlement

I like the fact that entitlement-mapped fields reflect the reality of on-chain data in some ways; everything is publicly readable, no matter what level of access you have, so I think it makes sense to not be able to use maps to "pretend" like a field can be made unreadable when that is not the reality.

runtime/parser/declaration_test.go
```diff
@@ -5010,7 +5010,7 @@ func TestParseTransactionDeclaration(t *testing.T) {
 
 		    var x: Int
 
-		    prepare(signer: AuthAccount) {
+		    prepare(signer: &Account) {
```
turbolent, line 5013: Create an tech debt issue to adjust the remaining parser tests that use the old account types

runtime/sema/account.cdc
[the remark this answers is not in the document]
turbolent: Took some notes for potential improvements to the entitlements feature and documentation.

runtime/sema/account.cdc
```diff
@@ -0,0 +1,521 @@
...
+entitlement GetStorageCapabilityController
+entitlement IssueStorageCapabilityController
+
+entitlement GetAccountCapabilityController
+entitlement IssueAccountCapabilityController
+
+/* Entitlement mappings */
+
+entitlement mapping AccountMapping {
+    // TODO: include Identity
```
dsainati1, line 448: Will update https://github.com/onflow/cadence/issues/2643 to include this

runtime/sema/gen/main.go
```diff
@@ -588,9 +589,20 @@ func (g *generator) VisitEntitlementDeclaration(decl *ast.EntitlementDeclaration
...
 }
 
-func (*generator) VisitEntitlementMappingDeclaration(_ *ast.EntitlementMappingDeclaration) struct{} {
-	// TODO
-	panic("entitlement declarations are not supported")
+func (g *generator) VisitEntitlementMappingDeclaration(decl *ast.EntitlementMappingDeclaration) (_ struct{}) {
+
+	entitlementMappingName := decl.Identifier.Identifier
+	typeVarName := typeVarName(entitlementMappingName)
+	typeVarDecl := entitlementMapTypeLiteral(entitlementMappingName, decl.Associations)
```
dsainati1, line 596: Once https://github.com/onflow/cadence/issues/2643 is implemented, this should be updated to include support for includes

runtime/sema/gen/main.go
turbolent: Go ignores directories named `testdata`, see https://pkg.go.dev/cmd/go/internal/test#pkg-variables, the files in the directory are not checked or compiled, so there shouldn't be any problems, like conflicts, because even the `.golden.go` files are basically just "plain-text files"

runtime/sema/check_event_declaration.go
turbolent: Added tests in cbbd45c70a471796b95002c33ea45da13adf670a

encoding/json/decode.go
```diff
@@ -1170,6 +1171,41 @@ func (d *Decoder) decodeIntersectionType(
...
+			continue
+		}
+
+		typeMap[string(semaType.ID())] = cadenceType
+	}
+
+	return typeMap
+}()
+
+func encodeAsSimpleType(primitiveType cadence.PrimitiveType) bool {
```
dsainati1, line 1205: The name of this function sounds like it's actually performing an encoding; can we rename it to maybe `canEncodeAsSimpleType`?

runtime/interpreter/primitivestatictype.go
```diff
@@ -280,28 +353,70 @@ func (t PrimitiveStaticType) elementSize() uint {
...
-		PrimitiveStaticTypeStorageCapabilityController,
-		PrimitiveStaticTypeAccountCapabilityController,
+		PrimitiveStaticTypeAuthAccountInbox,
 		PrimitiveStaticTypeAuthAccountStorageCapabilities,
 		PrimitiveStaticTypeAuthAccountAccountCapabilities,
 		PrimitiveStaticTypeAuthAccountCapabilities,
-		PrimitiveStaticTypePublicAccountCapabilities:
+		PrimitiveStaticTypePublicAccountCapabilities,
+		PrimitiveStaticTypeAccountKey:
+		// These types are deprecated, and only exist for migration purposes
```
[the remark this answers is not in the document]
turbolent: Added to the task list in #2642

runtime/parser/declaration_test.go
turbolent: Added to the task list in #2642

encoding/json/decode.go
turbolent: Good idea, improved in ff83742c9a5fbe509e03e33bd262bee3fdb5b99d

runtime/imported_values_memory_metering_test.go
```diff
@@ -396,7 +396,7 @@ func TestImportedValueMemoryMetering(t *testing.T) {
 
 		executeScript(t, script, meter, structValue)
 		assert.Equal(t, uint64(1), meter[common.MemoryKindCompositeValueBase])
-		assert.Equal(t, uint64(71), meter[common.MemoryKindRawString])
+		assert.Equal(t, uint64(284), meter[common.MemoryKindRawString])
```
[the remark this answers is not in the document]
turbolent: Hmm, this is a result of removing the type ID caching. I might have to revert the changes and fix it on master instead

runtime/interpreter/primitivestatictype.go
```diff
@@ -191,21 +196,79 @@ const (
...
+	_
+	_
+	PrimitiveStaticTypeMutate
+	PrimitiveStaticTypeInsert
+	PrimitiveStaticTypeRemove
+	PrimitiveStaticTypeIdentity
+	_
+	_
+	_
+	PrimitiveStaticTypeStorage
```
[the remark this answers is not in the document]
turbolent: Hmm, though there are no values that belong to entitlements, they are still types. Likewise, we do not currently encode kind information in the other type names (e.g. struct, resource, enum, etc.).

Beside keeping the definitions consistent, it also prevents us from accidentally introducing `*Type` constant when an `*Entilement` definition exists. That was my original reason for e.g. renaming the entitlement type constants in sema.

I don't feel strongly about it, we can change it back.

runtime/imported_values_memory_metering_test.go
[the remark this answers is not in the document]
turbolent: In #2751 I improved composite and interface static types on master in a different way, by adding type ID caching to interface static types, instead of removing it from composite static types.

I'll revert the changes here once that PR is merged.

encoding/ccf/simpletype.go
turbolent: Added as a TODO in #2642

runtime/convertTypes.go
turbolent: Added as a TODO in #2642

runtime/interpreter/statictype.go
turbolent: In https://github.com/onflow/cadence/pull/2751 I improved composite and interface static types on master in a different way, by adding type ID caching to interface static types, instead of removing it from composite static types.

I'll revert the changes here once that PR is merged.

runtime/sema/account.cdc
turbolent: Will add this in a follow-up PR, as it requires additional work in the type code generator

runtime/sema/gen/main.go
turbolent: Will add this in a follow-up PR, as it requires additional work in the type code generator

turbolent:
With #2751 and #2755 merged into Stable Cadence, I now reverted the removal of the composite static type ID caching changes in this PR.

turbolent:
@SupunS @dsainati1 Could you please have another look (see comment above)

dsainati1 (approved):
🙏

PR #2654, opened 2023-07-11
Introduce `reversed` in Fixed/Variable sized Array type

darkdrag00nv2:
Work towards https://github.com/onflow/cadence/issues/2605

## Description

Introduce `reversed` function for creating a copy of a Variable/Fixed size Array value with its entries reversed.

This function would be unavailable to resource arrays since resources cannot be copied.

Will send docs PR post the merge.

______


dsainati1:
Nit: I'm not sold on the naming here. How about having this reverse-and-copy function be called `reverse`, and then change the mutating one here https://github.com/onflow/cadence/pull/2626 to be `reverseInPlace` or something similar?

darkdrag00nv2:
@dsainati1 I was trying to follow the Swift convention but your suggestion does sound clearer. Will update this and the other PR accordingly.

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -10395,6 +10395,181 @@ func TestInterpretArrayFirstIndexDoesNotExist(t *testing.T) {
...
+		return res
+	  }
+	  fun originalsa(): [Int] {		
+		let res: [Int] = [];
+		for s in sa {
+		  res.append(s.test)
+		}
+	  
+		return res
+	  }
```
SupunS, line 10454: There seems to be a mix of tabs and whitespaces in the cadence code. Maybe just reformat to use either tabs or whitespaces, but not both

runtime/interpreter/value.go
```diff
@@ -2870,6 +2884,37 @@ func (v *ArrayValue) Slice(
...
+func (v *ArrayValue) Reverse(
+	interpreter *Interpreter,
+	locationRange LocationRange,
+) Value {
+	count := v.Count()
+	idx := count - 1
+
+	return NewArrayValueWithIterator(
+		interpreter,
+		NewVariableSizedStaticType(interpreter, v.Type.ElementType()),
```
SupunS, line 2896: The type here can also be `ConstantSizedStaticType`. Instead of constructing a new type, can simply pass `v.Type`.

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -10395,6 +10395,181 @@ func TestInterpretArrayFirstIndexDoesNotExist(t *testing.T) {
 	)
 }
 
+func TestInterpretArrayReverse(t *testing.T) {
+	t.Parallel()
+
+	inter := parseCheckAndInterpret(t, `
+      let xs = [1, 2, 3, 100, 200]
+      let ys = [100, 467, 297, 23]
```
SupunS, line 10403: Can you add some tests for constant-sized arrays as well?

runtime/tests/interpreter/interpreter_test.go
darkdrag00nv2: Done.

runtime/tests/interpreter/interpreter_test.go
darkdrag00nv2: Done, this uncovered a bug as well. The return type of function was always set to `VariableSized`. I fixed it to return whatever the type of the original array is.

runtime/sema/type.go
```diff
@@ -2193,6 +2225,13 @@ func ArraySliceFunctionType(elementType Type) *FunctionType {
 	}
 }
 
+func ArrayReverseFunctionType(arrayType Type) *FunctionType {
```
SupunS, line 2228:
```suggestion
func ArrayReverseFunctionType(arrayType sema.ArrayType) *FunctionType {
```

runtime/interpreter/value.go
```diff
@@ -2900,6 +2914,38 @@ func (v *ArrayValue) Slice(
 	)
 }
 
+func (v *ArrayValue) Reverse(
+	interpreter *Interpreter,
+	locationRange LocationRange,
+) Value {
+	count := v.Count()
+	idx := count - 1
```
SupunS, line 2922: Using the full noun `index` is more intuitive 
```suggestion
	index := count - 1
```

runtime/tests/checker/arrays_dictionaries_test.go
```diff
@@ -1078,6 +1078,55 @@ func TestCheckInvalidResourceFirstIndex(t *testing.T) {
...
+		fun test(): @[X] {
+			let xs <- [<-create X()]
+			return <-xs.reverse()
+		}
+    `)
+
+	errs := RequireCheckerErrors(t, err, 2)
+
+	assert.IsType(t, &sema.InvalidResourceArrayMemberError{}, errs[0])
+	assert.IsType(t, &sema.ResourceLossError{}, errs[1])
```
SupunS, line 1127: maybe destroy the resources to avoid the second error, since we are only interested in checking for the invalid-member error.

SupunS (approved):
LGTM! Just some minor suggestions

runtime/tests/checker/arrays_dictionaries_test.go
darkdrag00nv2: done

dsainati1 (approved):

turbolent (commented):
Nice!

Issue #2602, opened 2023-06-23
Declare events in interface


turbolent:
### Issue to be solved

Follow up from #2069: Remove nested type requirements, and change the semantics of declaring an event in an interface to declaring a concrete type

PR #2662, opened 2023-07-18
Allow declaration of events in interfaces

dsainati1:
Closes https://github.com/onflow/cadence/issues/2602

Allows contract interfaces to declare events as their own types, rather than implicitly using such declarations to create type requirements. Currently, these declarations are only in scope within the body of the interface, and thus can only be emitted in default implementations or in pre/post conditions in these interfaces, but subject to the discussion in https://github.com/onflow/flips/pull/118 this may change. 

______


turbolent:
@dsainati1 Is this PR marked as a draft because the FLIP isn't approved yet, or is there still something missing code-wise?

turbolent (approved):
Great work! Good idea to first change the behaviour of event declarations, then remove type requirements in a separate PR 👌

runtime/sema/check_interface_declaration.go
```diff
@@ -357,62 +372,93 @@ func (checker *Checker) declareInterfaceMembers(declaration *ast.InterfaceDeclar
...
+					Identifier:            identifier,
+					Access:                checker.accessFromAstAccess(nestedCompositeDeclaration.DeclarationAccess()),
+					ContainerType:         interfaceType,
+					TypeAnnotation:        NewTypeAnnotation(nestedCompositeDeclarationVariable.Type),
+					DeclarationKind:       nestedCompositeDeclarationVariable.DeclarationKind,
+					VariableKind:          ast.VariableKindConstant,
+					ArgumentLabels:        nestedCompositeDeclarationVariable.ArgumentLabels,
+					IgnoreInSerialization: true,
+					DocString:             nestedCompositeDeclaration.DeclarationDocString(),
+				})
```
SupunS, line 447: Looks like this function doesn't have to be a closure function. If that's the case, maybe extract out to top level

SupunS (commented):
Do events get inherited for interfaces (or do we want them to be inherited)?
Could you also add some tests to verify?

dsainati1:
> Do events get inherited for interfaces?

No. There's a test for that in the PR already I think as well.

SupunS (approved):

PR #2664, opened 2023-07-18
Make indexing assignment entitled

### LINKED ISSUE #2647: Make indexing assignment entitled
Make index assignments (`arrayRef[0] = something` or `dictRef[index] = something`) entitled

SupunS:
Closes #2647
Work towards https://github.com/onflow/flips/pull/86

## Description

Make index assignments on references (`arrayRef[0] = something` or `dictRef[index] = something`) to require entitlements.

______


dsainati1 (commented):
I don't think it's safe to only require `Insertable` access for indexed assignment, unfortunately. In the example in the test: 

```
 let array: [String] = ["foo", "bar"]
fun test() {
    var arrayRef = &array as auth(Insertable) &[String]
    arrayRef[0] = "baz"
}
```

This both inserts `"baz"` and also removes `"foo"`. Similarly, when doing a swap, you are both inserting one element and removing another. As such I think this operation needs to have `Mutable`-only access. Ideally we could say that it has `(Mutable) | (Insertable, Removable)`, but we don't support this level of complexity. I think we can just permit it only for `Mutable` now, and later if we add the ability for entitlements to inherit or extend each other we can define `Mutable` to be the intersection of `Insertable` and `Removable` to fix this oddity.

SupunS:
ah that's a good flag! I think internally we can make the `(Mutable) | (Insertable, Removable)` work.

dsainati1:
Oh nice! If we can special case this internally that would be great. We should just make sure to make the error message here not produce the "raw" entitlement set so as not to mislead people about what is possible in user-land.

dsainati1 (approved):
Nice!

runtime/sema/check_member_expression.go
```diff
@@ -112,17 +112,21 @@ func (checker *Checker) getReferenceType(typ Type, substituteAuthorization bool,
 }
 
 func shouldReturnReference(parentType, memberType Type) bool {
-	if memberType == nil || !isReferenceType(parentType) {
+	if memberType == nil {
+		return false
```
turbolent, line 116: Could you please add a test case covering this case?

runtime/sema/errors.go
```diff
@@ -2929,6 +2929,46 @@ func (e *InvalidAssignmentAccessError) SecondaryError() string {
 	)
 }
 
+// UnauthorizedReferenceAssignmentError
+
+type UnauthorizedReferenceAssignmentError struct {
+	RequiredAccess [2]Access
```
turbolent, line 2935: Why are exactly two accesses needed here?

turbolent (commented):
Nice!

runtime/sema/errors.go
SupunS: Currently, to assign to an array/dictionary element, one would need to have either an `auth(Mutate)` reference or an `auth(Insert, Remove)` reference. So the required two accesses are that.

dsainati1 (changes requested):
I forgot one thing when reviewing this: because dictionaries and arrays now support entitled operations, `ArrayType` and `DictionaryType` need to be updated to conform to `EntitlementSupportingType` in `type.go`, and the implementation of `SupportedEntitlements` for these types should include `Mutable`, `Insertable` and `Removable`.

These can be tested by adding a test to `TestCheckEntitlementConditions` to test that when returning a resource array or dictionary type, the created reference type of `result` in a function postcondition contains these entitlements properly.

runtime/sema/check_member_expression.go
SupunS: don't think `memberType` can be nil actually. Can remove this.

SupunS: Member type cannot be nil actually. Removed the nil check: c142c39b8556ec9812b186bbd30a3fbd7f60e633

SupunS:
@dsainati1 Added in ae64ba2bc4621b3442437a7e790b7ff609197f36

dsainati1 (approved):

turbolent (approved):
Nice!

PR #2665, opened 2023-07-19
Fix positioning of default function conflict errors

dsainati1:
Fixes error positioning for two errors that could try to report errors positioned in upstream files. Cadence doesn't allow this currently, so the error would be reported at the same row/column location in the current file, which was nonsensical. Instead, we position the errors at the definition of the conflicting composite type. 

______


SupunS (approved):
Nice!
So the issue was mostly in `DefaultFunctionConflictError` error, because it reported a location in a different file? We might later want to scan all other errors to see if there are any similar cases.

dsainati1:
Yeah that and also `MultipleInterfaceDefaultImplementationsError`.

turbolent (commented):
Nice!

Issue #1283, opened 2021-11-30
Remove type requirements


turbolent:
Scope:
Remove feature completely from the codebase.
Feature: Interface can require a concrete type to specify a nested type.
Code removal, updating test cases and docs.

Not in scope:
1) storage migration for FT/NFT contracts.
2) rewrite of FT/NFT token standards

Note: This is only need if the community agrees we need a token contract rewrite.

- [x] #2069 
- [x] #2081 

```[tasklist]
### Tasks
- [ ] https://github.com/onflow/cadence/pull/2670
- [ ] https://github.com/onflow/cadence/pull/2662
```

j1010001:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Flow-Protocol-5a2ad022cee40e25b0885f82/issues/onflow/cadence/1283?planning-poker) with ZenHub @dsainati1 @robert-e-davidson3 @SupunS @turbolent

j1010001:
Needs to be broken down.

dsainati1:
As discussed in the last [LDM](https://github.com/onflow/cadence/blob/master/meetings/2022-11-02.md#removalreduction-of-nested-type-requirements), in order to do this we need to provide support for allowing interfaces to have events; as the new NFT/FT standards are going to feature interfaces with events. 

There are currently a few competing ideas for how this might be done:
    1) Keep type requirements for events, and limit them only to events
    2) Allow events to be emitted from interfaces directly
    3) Allow composites to inherit events from interfaces 

All of these have pros and cons worth considering. 

There is also a separate but related question of whether or how we want to enforce that events are emitted in certain cases. IMO this is not a prerequisite for solving this issue, but it is a related problem and our potential solution to that may inform our solution to this.

PR #2670, opened 2023-07-20
Sainati/remove type requirements

dsainati1:
Closes https://github.com/onflow/cadence/issues/1283

This removes support for type requirements from Cadence, as proposed in https://github.com/onflow/flips/pull/118

______


turbolent:
@dsainati1 Is this PR marked as a draft because the FLIP isn't approved yet, or is there still something missing code-wise?

dsainati1:
> Is this PR marked as a draft because the FLIP isn't approved yet

Yes, if the design is approved as-is, this implementation is complete, so it's ready to be reviewed.

runtime/contract_update_validation_test.go
```diff
@@ -1750,21 +1750,6 @@ func TestRuntimeContractUpdateValidation(t *testing.T) {
 		assertContractRemovalError(t, err, "Test")
 	})
 
-	t.Run("Remove contract interface with enum", func(t *testing.T) {
```
turbolent, line 1753: Maybe keep the test and adjust it accordingly

runtime/interpreter/interpreter.go
```diff
@@ -2253,7 +2234,8 @@ func (interpreter *Interpreter) declareInterface(
 			if nestedCompositeDeclaration.Kind() == common.CompositeKindEvent {
 				interpreter.declareNonEnumCompositeValue(nestedCompositeDeclaration, lexicalScope)
 			} else {
-				interpreter.declareTypeRequirement(nestedCompositeDeclaration, lexicalScope)
+				// this should be statically prevented in the checker
```
turbolent, line 2237: Is this a TODO, or does it mean that the PR already changes the checker to reject this case? (Maybe change tense to avoid confusion)

runtime/sema/check_composite_declaration.go
```diff
@@ -155,15 +155,12 @@ func (checker *Checker) visitAttachmentDeclaration(declaration *ast.AttachmentDe
 }
 
 // visitCompositeDeclaration checks a previously declared composite declaration.
-// Checking behaviour depends on `kind`, i.e. if the composite declaration declares
```
turbolent, line 158: It's great we can finally get rid of this 👌

runtime/sema/check_composite_declaration.go
```diff
@@ -196,72 +193,48 @@ func (checker *Checker) visitCompositeLikeDeclaration(declaration ast.CompositeL
...
 
-	if kind == ContainerKindComposite {
-		checker.enterValueScope()
-		defer checker.leaveValueScope(declaration.EndPosition, false)
-	}
+	checker.enterValueScope()
+	defer checker.leaveValueScope(declaration.EndPosition, false)
 
-	checker.declareCompositeLikeNestedTypes(declaration, kind, true)
+	checker.declareCompositeLikeNestedTypes(declaration, ContainerKindComposite, true)
```
turbolent, line 199: The container kind parameter for `declareCompositeLikeNestedTypes` seems to always be `ContainerKindComposite` now. I tried removing it in https://github.com/onflow/cadence/commit/f8e84d85d89998accbe31412ed41408a1bdeff01 and tests are still passing. Feel free to cherry pick it in, or we can open the clean up as a follow-up PR (if it is correct).

runtime/sema/check_composite_declaration.go
```diff
@@ -468,10 +422,15 @@ func (checker *Checker) declareNestedDeclarations(
 
 			firstNestedCompositeDeclaration := nestedCompositeDeclarations[0]
 
-			reportInvalidNesting(
-				firstNestedCompositeDeclaration.DeclarationKind(),
-				firstNestedCompositeDeclaration.Identifier,
-			)
+			// composite-in-interface nesting errors are already reported elsewhere
```
turbolent, line 425: Here and in the similar cases below: Where is it already reported? Could this maybe be unified in one place (here?), so the filtering code here is not necessary? (of course not if this makes the code worse)

runtime/sema/check_interface_declaration.go
```diff
@@ -156,21 +156,26 @@ func (checker *Checker) VisitInterfaceDeclaration(declaration *ast.InterfaceDecl
 	}
 
 	for _, nestedComposite := range declaration.Members.Composites() {
-		// Non-event composite declarations nested in interface declarations are type requirements,
-		// i.e. they should be checked like interfaces
-
+		// only event types may be declared in interfaces
 		if nestedComposite.Kind() != common.CompositeKindEvent {
-			checker.visitCompositeLikeDeclaration(nestedComposite, kind)
+			checker.report(&InvalidNestedDeclarationError{
```
turbolent, line 161: Looks like the error reporting for invalid nested declarations is reported here. Maybe do nothing here, remove the filtering above (see comment above), and add a comment here instead noting that this case is a NO-OP, and an error is reported for all invalid nestings elsewhere.

turbolent (commented):
Great work! 👏 The removal really cleans up the code a lot 👌

runtime/contract_update_validation_test.go
dsainati1: There's nothing to adjust I think, this test is testing that removing a type requirement is not a valid update, which isn't a relevant thing to test anymore.

turbolent: Ah, I see, makes sense!

runtime/common/declarationkind.go
```diff
@@ -237,3 +237,13 @@ func (s DeclarationKindSet) With(kind DeclarationKind) DeclarationKindSet {
 func (s DeclarationKindSet) Has(kind DeclarationKind) bool {
 	return s&(1<<kind) != 0
 }
+
+func (k DeclarationKind) IsInterfaceDeclaration() bool {
```
turbolent, line 241: Maybe move this up, before `DeclarationKindSet` and methods for it are defined

runtime/tests/checker/interface_test.go
```diff
@@ -1525,317 +1525,132 @@ func TestCheckInterfaceSelfUse(t *testing.T) {
...
-	)
+		errs := RequireCheckerErrors(t, err, 1)
 
-	require.NoError(t, err)
+		assert.IsType(t, &sema.InvalidNestedDeclarationError{}, errs[0])
+	})
 }
 
-func TestCheckContractInterfaceFungibleToken(t *testing.T) {
+// TODO: re-enable this test with the v2 fungible token contract
```
SupunS, line 1633: Might be good to create an issue for enabling these tests, and add it to the stable-cadence epic, so that we don't forget.

runtime/tests/checker/interface_test.go
```diff
@@ -3717,244 +3180,6 @@ func TestCheckInterfaceTypeDefinitionInheritance(t *testing.T) {
 
 	t.Parallel()
 
-	t.Run("type requirement", func(t *testing.T) {
```
SupunS, line 3720: We might want to add some tests for the behavior of interface inheritance with events. Previously they worked the same way as type requirements, but now that they behave differently, existing tests may not cover their new behavior.

SupunS (approved):

runtime/tests/checker/interface_test.go
dsainati1: https://github.com/onflow/cadence/issues/2719

runtime/tests/checker/interface_test.go
SupunS: we could add a test to ensure they are not inherited (if not already exist)

dsainati1: I believe the tests in https://github.com/onflow/cadence/pull/2662 already assert that they are not inherited.

SupunS: nice!

PR #2674, opened 2023-07-21
Fix interface inheritance for nested interfaces

dsainati1:
Interfaces heirarchies nested in interfaces were not being constructed properly, causing the tests added here to fail before the fix. 
______

### DIFF
--- a/runtime/sema/checker.go
+++ b/runtime/sema/checker.go
@@ -379,10 +379,6 @@ func (checker *Checker) CheckProgram(program *ast.Program) {
 	for _, declaration := range program.InterfaceDeclarations() {
 		interfaceType := checker.Elaboration.InterfaceDeclarationType(declaration)
 
-		// Resolve conformances
-		interfaceType.ExplicitInterfaceConformances =
-			checker.explicitInterfaceConformances(declaration, interfaceType)
-
 		VisitThisAndNested(interfaceType, registerInElaboration)
 	}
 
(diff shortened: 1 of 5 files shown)


runtime/sema/checker.go
```diff
@@ -379,10 +379,6 @@ func (checker *Checker) CheckProgram(program *ast.Program) {
 	for _, declaration := range program.InterfaceDeclarations() {
 		interfaceType := checker.Elaboration.InterfaceDeclarationType(declaration)
 
-		// Resolve conformances
-		interfaceType.ExplicitInterfaceConformances =
-			checker.explicitInterfaceConformances(declaration, interfaceType)
-
```
SupunS, line 385: I think we need it here. i.e: done in two steps, to handle the case where `MiddleInterface` is defined before the `TopInterface`.
e.g.
```cadence
access(all) resource ConcreteResource: MiddleInterface {}

access(all) resource interface MiddleInterface: TopInterface {}

access(all) resource interface TopInterface {}
```

I remember adding a test for that exact case (but for top level typ-defs). Wonder if that still works.

dsainati1: Ah you're right, I forgot to run the full test suite here, and that test fails now...

Hmm okay this might be a bit more complex than I thought. Will take a closer look.

dsainati1: Ok we do need it here, but after the initial pass over the top-level composites, because contracts can declare interfaces and those need to be processed before we check the conformances. Then we also need it when declaring members for contracts and interfaces to catch all nested interface declarations.

SupunS (approved):
Thanks for fixing!

turbolent (commented):
Nice!

Issue #2672, opened 2023-07-20
Storage iteration doesn't skip broken values


SupunS:
### Current Behavior

$subject. Thus, when trying to load the value from storage during iteration, it panics, and the transaction/script aborts.

### Expected Behavior

Should skip the broken values.

### Steps To Reproduce

Refer https://discord.com/channels/613813861610684416/1108479699732152503/1131634801280045157 for more details.

### Environment

```markdown
- Cadence version: v0.39.14
- Network: mainnet/testnet
```

SupunS:
Maybe related: https://github.com/dapperlabs/cadence-internal/issues/138

PR #2676, opened 2023-07-21
Fix storage iteration and error misclassification

SupunS:
Closes https://github.com/onflow/cadence/issues/2672
Closes https://github.com/dapperlabs/cadence-internal/issues/138

## Description

There were two issues:
 1) Broken values skipping logic didn't skip some broken values
    - The root cause was that, for links, the existing check was only checking the linked type.
    - It wasn't following the linked path, and checking the actual value in the storage.
    - Fix: a93d153392e3926b5483a78a5c361387e66a6cf8
 2) The reported error was an internal error
    - Some errors were returned from FVM as wrapped errors.
    - The error recovery mechanism wasn't unwrapping them, so they get treated as internal errors.
    - Fix: ae4c52160bd1f0d7b04adf133247f4caae02670a

______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -519,7 +519,7 @@ func (interpreter *Interpreter) InvokeTransaction(index int, arguments ...Value)
 func (interpreter *Interpreter) RecoverErrors(onError func(error)) {
 	if r := recover(); r != nil {
 		// Recover all errors, because interpreter can be directly invoked by FVM.
-		err := AsCadenceError(r)
+		err := asCadenceError(r)
 
 		// if the error is not yet an interpreter error, wrap it
 		if _, ok := err.(Error); !ok {
@@ -549,7 +549,7 @@ func (interpreter *Interpreter) RecoverErrors(onError func(error)) {
 	}
 }
 
-func AsCadenceError(r any) error {
+func asCadenceError(r any) error {
 	err, isError := r.(error)
 	if !isError {
 		return errors.NewUnexpectedError("%s", r)
@@ -3838,7 +3838,7 @@ func (interpreter *Interpreter) checkValue(
 	}()
 
 	// Here, the value at the path could be either:
-	//	1) The actual stored value
+	//	1) The actual stored value (storage path)
 	//	2) A link to the value at the storage (private/public paths)
 	//
 	// Therefore, try to find the final path, and try loading the value.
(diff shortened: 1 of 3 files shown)


runtime/interpreter/interpreter.go
```diff
@@ -4481,50 +4530,63 @@ func (interpreter *Interpreter) pathCapabilityCheckFunction(
 				panic(errors.NewUnreachableError())
 			}
 
-			target, authorized, err :=
```
SupunS, line 4484: the logic here was moved as-is to the `Interpreter.checkValueAtPath` function, so it can be reused in the `Interpreter.checkValue()` function.

runtime/interpreter/interpreter.go
```diff
@@ -3817,8 +3837,37 @@ func (interpreter *Interpreter) checkTypeLoading(staticType StaticType) (typeErr
 		}
 	}()
 
-	// Here it is only interested in whether the type can be properly loaded.
-	_, typeError = interpreter.ConvertStaticToSemaType(staticType)
```
SupunS, line 3821: **Fix 1**: Here, it is not sufficient to just load the types. But rather, need to follow links, similar to `pathCapabilityCheckFunction ` function, and check the type of the actual value stored.

runtime/interpreter/interpreter.go
```diff
@@ -561,6 +549,31 @@ func (interpreter *Interpreter) RecoverErrors(onError func(error)) {
...
+
+	for {
+		switch typedError := err.(type) {
+		case Error,
+			errors.ExternalError,
+			errors.InternalError,
+			errors.UserError:
+			return typedError
+		case xerrors.Wrapper:
+			err = typedError.Unwrap()
```
SupunS, line 568: **Fix 2**: Need to unwrap any unknown error, and look for the error chain.

dsainati1 (approved):

runtime/interpreter/interpreter.go
```diff
@@ -3817,8 +3837,37 @@ func (interpreter *Interpreter) checkTypeLoading(staticType StaticType) (typeErr
...
+		return valueError
+	}
+
+	// This is guaranteed to be a reference type, because `borrowType` is always a reference.
+	referenceType, ok := semaType.(*sema.ReferenceType)
+	if !ok {
+		panic(errors.NewUnreachableError())
+	}
+
+	_, valueError = interpreter.checkValueAtPath(address, path, referenceType, EmptyLocationRange)
```
turbolent, line 3870: Maybe pass in the location range to `Interpreter.checkValue` and pass it along here (if available at call-site)

runtime/interpreter/interpreter.go
```diff
@@ -561,6 +549,31 @@ func (interpreter *Interpreter) RecoverErrors(onError func(error)) {
 	}
 }
 
+func asCadenceError(r any) error {
```
turbolent, line 552: Good idea to refactor this into a function 👍

turbolent (approved):
👍

runtime/interpreter/interpreter.go
SupunS: added e5694e7d6d5549f4653e906c78e32d68128eba18

turbolent (approved):

PR #2683, opened 2023-07-27
Wrap host env errors with external errors

### LINKED ISSUE #2682: Computation limit exceeding errors are treated as internal errors
### Current Behavior

If the user exceeds the computation limit, it should be a user error. But currently it is treated as an internal error (and prints the go-stacktrace). e.g: https://github.com/onflow/cadence/issues/2679

### Expected Behavior

Such errors should be treated as user errors

### Steps To Reproduce

Run a transaction/code that exceeds the computation limit.

### Environment

```markdown
- Cadence version: latest/master
- Network: TN/MN
```

SupunS:
Closes #2682

## Description

Wrap the errors **returned** from host-env using the external error. 
This is already done for the errors panicked from host-env.

______


turbolent (commented):
The convention for runtime interface functions is that:
- Errors returned from the function are user errors
- Panics are internal errors (e.g. implementation bug)

For example, when a requested block does not exist, the implementation of `GetBlockAtHeight` can return an error indicating that.

Why do we wrap paniced errors (internal errors) as external errors? To differentiate that they did not originate in Cadence itself. 

Why would we wrap returned errors, i.e. classify user errors, as external errors?

Both questions probably stem from my lack of understanding what external errors are used/needed for.

SupunS:
> The convention for runtime interface functions is that:
> 
>  - Errors returned from the function are user errors
>  - Panics are internal errors (e.g. implementation bug)

So the problem is this convention is not always followed (and is not guaranteed to be followed either). There can be situations where certain user errors are also panicked, while internal errors are returned. This is a really grey area, sometimes a 'user error' to the host-env, etc. could actually be wrong information being sent by Cadence such that it's actually an internal error for the user.

> Why do we wrap paniced errors (internal errors) as external errors? To differentiate that they did not originate in Cadence itself?
>
> Why would we wrap returned errors, i.e. classify user errors, as external errors?

Cadence doesn't know if a returned/thrown error is actually an internal error or a user error (unless it was originated from Cadence). So there's this "external error" category where it just says it's an error that came from outside and Cadence can't determine which one it is.
Also, the wrapping doesn't wrap crashers (i.e: go-runtime errors as NPE, cast errors, etc.), it only wraps errors that were explicitly thrown/returned by the implementation.

SupunS:
The real problem I think is, in cadence, we use panics to propagate errors (rather than returning). So any un-categorized error creates a problem at the top of the call tree where all errors are handled.
IMO the proper fix is to refactor the entire code base to use error-returning for all errors, and panic only the internal errors (which is not the case now). But unfortunately, that's going to be a massive refactoring.

turbolent:
@SupunS Thank you for the great explanation, that makes sense!

turbolent (approved):

PR #2687, opened 2023-08-02
Fix error on reference creation with invalid type

### LINKED ISSUE #2685: Incorrect error when deploying contract with bad imports
### Current Behavior

Discord user **a_coimbra** was having problems deploying a contract on testnet because an import address was incorrect.

https://discord.com/channels/613813861610684416/621847426201944074/1134143072943485019

However the error itself is incorrect:

```
[Error Code: 1101] error caused by: 1 error occurred:
    * transaction execute failed: [Error Code: 1101] cadence runtime error: %!v(PANIC=Error method: runtime error: invalid memory address or nil pointer dereference)
```

### Expected Behavior

the error should state: `... cannot find declaration ...`

### Steps To Reproduce

I wasn't able to reproduce by just repeating the scenario in an FVM unit test

### Environment

```markdown
- Cadence version: v0.39.14
- Network: testnet
```

SupunS:
Closes #2685

## Description

If the RHS type of the reference expression was an invalid type, then the reported error has a `nil` expected type.

It is OK to skip reporting this error, if the type is invalid.

______

### DIFF
--- a/runtime/sema/check_reference_expression.go
+++ b/runtime/sema/check_reference_expression.go
@@ -92,7 +92,9 @@ func (checker *Checker) VisitReferenceExpression(referenceExpression *ast.Refere
 		// If the reference type was a non-optional type,
 		// check that the referenced expression does not have an optional type
 
-		if _, ok := actualType.(*OptionalType); ok != isOpt {
+		// Do not report an error if the `expectedLeftType` is unknown
+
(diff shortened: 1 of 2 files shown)


SupunS:
The `BackwardCompatibilityCheck` is reporting a diff, because this now reports one less error. But technically it is not a breaking change because those programs always have other errors (invalid type due to `NotDeclaredError`).

turbolent (approved):
Nice!

PR #2693, opened 2023-08-03
Conversion to entitled type

dsainati1:
Part of the proposed migration in https://github.com/onflow/flips/pull/95

Part of https://github.com/onflow/cadence/issues/2709

This implements and tests a function that converts types to entitled versions of themselves according to the rules described in the linked migration proposal. In particular:

`ConvertToEntitledType` converts its input to an entitled type according to the following rules:
* `ConvertToEntitledType(&T) ---> auth(Entitlements(T)) &T`
* `ConvertToEntitledType(Capability<T>) ---> Capability<ConvertToEntitledType(T)>`
* `ConvertToEntitledType(T?) ---> ConvertToEntitledType(T)?`
* `ConvertToEntitledType(T) ---> T`
where `Entitlements(T)` is defined as the result of `T.SupportedEntitlements()`

Note that these are the only types converted by this function; an existing composite type will not have its fields or function types migrated. This is because in the migration, composite values are not going to be migrated; they will simply possess the required entitlements already by virtue of being owned. This does mean, however, that existing composite values with capability-typed fields will break. This is unfortunately unavoidable, as capability-typed fields cannot have their types changed anyways according to the contract updatability rules. 

______


dsainati1:
Obseleted by https://github.com/onflow/cadence/pull/2951

Issue #2076, opened 2022-10-18
Provide detailed information about the error during smart contract execution in test


satyamakgec:
There is no way to handle the errors in the test, as `Test.Error` gives a full stack trace, but there is no way to know the error string programmatically because of that, it is hard to test bad cases within the contract.

Maybe better to return the `Error` as a struct which contains the error string, stack trace, position etc.

PR #2696, opened 2023-08-04
Cadence testing framework improvements

m-Peter:
Closes https://github.com/onflow/cadence/issues/2076
**Companion PR:** https://github.com/onflow/cadence-tools/pull/181

## Description

Adding the following improvements:
- Allow **rolling back** at a specific block height
- Add helper function to **assert error message** on `ScriptResult`/`TransactionResult`

Sample usage:

```cadence
import Test

pub let blockchain = Test.newEmulatorBlockchain()
pub let account = blockchain.createAccount()

// Retrieve current block height
pub let height = getCurrentBlockHeight(blockchain: blockchain)

// Reset (rollback) the blockchain to a given block height
blockchain.reset(height)

pub let script = Test.readFile("../scripts/get_integer_traits.cdc")
pub let result = blockchain.executeScript(script, [])

// Assert that a script/transaction errored and the error message contains a specified sub-string
Test.assertError(result, errorMessage: "exceeding limit")
```

______

### DIFF
--- a/runtime/stdlib/contracts/test.cdc
+++ b/runtime/stdlib/contracts/test.cdc
@@ -122,10 +122,10 @@ pub contract Test {
             return self.backend.events(type)
         }
 
-        /// Resets the state of the blockchain.
+        /// Resets the state of the blockchain at the given height.
         ///
-        pub fun reset() {
-            self.backend.reset()
+        pub fun reset(to height: UInt64) {
+            self.backend.reset(to: height)
         }
     }
 
@@ -305,9 +305,9 @@ pub contract Test {
         ///
         pub fun events(_ type: Type?): [AnyStruct]
 
-        /// Resets the state of the blockchain.
+        /// Resets the state of the blockchain at the given height.
         ///
-        pub fun reset()
+        pub fun reset(to height: UInt64)
     }
 
     /// Returns a new matcher that negates the test of the given matcher.
--- a/runtime/stdlib/contracts/test.cdc
+++ b/runtime/stdlib/contracts/test.cdc
@@ -171,6 +171,10 @@ pub contract Test {
         /// The resulted status of an executed operation.
         ///
         pub let status: ResultStatus
+
+        /// The optionally resulted error of an executed operation.
+        ///
+        pub let error: Error?
     }
 
     /// The result of a transaction execution.
@@ -344,6 +348,30 @@ pub contract Test {
         return Matcher(test: fun (value: AnyStruct): Bool {
             return value == nil
         })
+    }
+
+    /// Asserts that the result of an executed operation, such as
+    /// scripts & transactions, has errored and the error message
+    /// contains the given sub-string.
+    ///
+    pub fun assertError(_ result: {Result}, errorMessage: String) {
+        pre {
+            result.status == ResultStatus.failed: "no error was found"
+        }
+
+        var found = false
+        let msg = result.error!.message
+        let msgLength = msg.length - errorMessage.length + 1
+        var i = 0
+        while i < msgLength {
+            if msg.slice(from: i, upTo: i + errorMessage.length) == errorMessage {
+                found = true
+                break
+            }
+            i = i + 1
+        }
+
+        assert(found, message: "the error message did not contain the given sub-string")
     }
 
 }
--- a/runtime/stdlib/contracts/test.cdc
+++ b/runtime/stdlib/contracts/test.cdc
@@ -122,7 +122,7 @@ pub contract Test {
             return self.backend.events(type)
         }
 
-        /// Resets the state of the blockchain at the given height.
+        /// Resets the state of the blockchain to the given height.
         ///
         pub fun reset(to height: UInt64) {
             self.backend.reset(to: height)
@@ -168,11 +168,11 @@ pub contract Test {
     /// operations, such as transactions and scripts.
     ///
     pub struct interface Result {
-        /// The resulted status of an executed operation.
+        /// The result status of an executed operation.
         ///
         pub let status: ResultStatus
 
-        /// The optionally resulted error of an executed operation.
+        /// The optional error of an executed operation.
         ///
         pub let error: Error?
     }
@@ -309,7 +309,7 @@ pub contract Test {
         ///
         pub fun events(_ type: Type?): [AnyStruct]
 
-        /// Resets the state of the blockchain at the given height.
+        /// Resets the state of the blockchain to the given height.
         ///
         pub fun reset(to height: UInt64)
     }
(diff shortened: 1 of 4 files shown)


runtime/stdlib/contracts/test.cdc
```diff
@@ -12,6 +12,103 @@ pub contract Test {
             self.backend = backend
         }
 
+        /// Returns the Flow token balance for the given account.
+        ///
+        pub fun getFlowBalance(_ account: Test.Account): UFix64 {
+            let script = "pub fun main(address: Address): UFix64 { return getAccount(address).balance }"
```
m-Peter, line 18: Instead of this line:
```cadence
let script = "pub fun main(address: Address): UFix64 { return getAccount(address).balance }"
```
I would have like to be able to use the following:
```cadence
let script = Test.readFile("scripts/get_flow_token_balance.cdc")
```
However, the `Test.readFile` method is implemented natively, and I can't seem to be able to use it from within the `Test` contract.

runtime/stdlib/contracts/test.cdc
```diff
@@ -12,6 +12,103 @@ pub contract Test {
...
+            }
+            return scriptResult.returnValue! as! UInt64
+        }
+
+        /// Mints the given amount of Flow tokens to a specified test account.
+        /// The transaction is authorized and signed by the service account.
+        /// Returns the result of the transaction.
+        ///
+        pub fun mintFlow(_ receiver: Test.Account, amount: UFix64): TransactionResult {
+            let lines: [String] = [
```
m-Peter, line 44: I would have liked to be able to write this in a multi-line `String`, but that isn't possible yet.
So I had to improvise, in order to make this somewhat more readable :innocent:

turbolent:
In general, these additions are great ideas! The FLOW token helpers I'm not so sure about (`mintFlow` and `burnFlow`). 

Though they are obviously convenient, the Cadence repo is probably not the best place to add them. We're trying to keep Cadence chain-agnostic / independent of Flow. Unfortunately there have been some places where we were not strict (e.g. naming of built-in events), but it would be nice to at least avoid adding new code tying Cadence to Flow, i.e. here the assumption that there is a FLOW token. 

Maybe these helper functions could be added upstream, e.g. in `cadence-tools/test`? That way we would not force embedders of Cadence to provide the FLOW token-specific functionality.

runtime/stdlib/contracts/test.cdc
```diff
@@ -122,10 +219,10 @@ pub contract Test {
             return self.backend.events(type)
         }
 
-        /// Resets the state of the blockchain.
+        /// Resets the state of the blockchain at the given height.
         ///
-        pub fun reset() {
-            self.backend.reset()
+        pub fun reset(_ height: UInt64) {
```
turbolent, line 224: Embrace the use of argument labels, they help writers and readers of the code. For example here, consider a short but useful argument label, e.g. `to` to indicate the number is a block height, or a more verbose / descriptive `toHeight`
```suggestion
        pub fun reset(to height: UInt64) {
```

Same below for the other `reset` function

runtime/stdlib/contracts/test.cdc
```diff
@@ -346,4 +447,27 @@ pub contract Test {
...
+    /// Asserts that the result of an executed operation, such as
+    /// scripts & transactions, has errored and the error message
+    /// contains the given sub-string.
+    ///
+    pub fun assertError(_ result: {Result}, substr: String) {
+        pre {
+            result.status == ResultStatus.failed: "no error was found"
+        }
+
+        var found: Bool = false
```
turbolent, line 459:
```suggestion
        var found = false
```

runtime/stdlib/contracts/test.cdc
```diff
@@ -346,4 +447,27 @@ pub contract Test {
         })
     }
 
+    /// Asserts that the result of an executed operation, such as
+    /// scripts & transactions, has errored and the error message
+    /// contains the given sub-string.
+    ///
+    pub fun assertError(_ result: {Result}, substr: String) {
```
satyamakgec, line 454: maybe rename the variable `substr` to `errorMessage`

satyamakgec (approved):
LGTM

runtime/stdlib/contracts/test.cdc
m-Peter: Nice, I wasn't aware of this little gem :100: 
Updated in https://github.com/onflow/cadence/pull/2696/commits/aee08f0d524c5b237ccb052618abfa9d3e507ab7

runtime/stdlib/contracts/test.cdc
m-Peter: Fixed in https://github.com/onflow/cadence/pull/2696/commits/600ed65a245e0d67a65f677c60c707fa047dfff1

runtime/stdlib/contracts/test.cdc
m-Peter: Good idea :ok_hand: 
Fixed in https://github.com/onflow/cadence/pull/2696/commits/600ed65a245e0d67a65f677c60c707fa047dfff1

m-Peter:
> In general, these additions are great ideas! The FLOW token helpers I'm not so sure about (mintFlow and burnFlow).
> 
> Though they are obviously convenient, the Cadence repo is probably not the best place to add them. We're trying to keep Cadence chain-agnostic / independent of Flow. Unfortunately there have been some places where we were not strict (e.g. naming of built-in events), but it would be nice to at least avoid adding new code tying Cadence to Flow, i.e. here the assumption that there is a FLOW token.
> 
> Maybe these helper functions could be added upstream, e.g. in cadence-tools/test? That way we would not force embedders of Cadence to provide the FLOW token-specific functionality.

@turbolent Thanks for the feedback :pray: You've got a good point. I wanted to add the helper functions as part of the `Test.Blockchain` type, but I wasn't satisfied in the first place, because I had to embed the scripts/transactions in Cadence strings. I have moved all 4 helper functions to the companion PR on the `cadence-tools/test` repository. The new API looks like this:

```cadence
import Test
import BlockchainHelpers

pub let blockchain = Test.newEmulatorBlockchain()
pub let account = blockchain.createAccount()

// Retrieve current block height
let height = getCurrentBlockHeight(blockchain: blockchain)

// Reset (rollback) the blockchain to a given block height
blockchain.reset(height)

// Retrieve the Flow token balance of a test account
let balance = getFlowBalance(for: account, blockchain: blockchain)

// Mint a given amount of Flow tokens to a specified test account
mintFlow(to: account, amount: 1500.0, blockchain: blockchain)

// Burn a given amount of Flow tokens from a specified test account
burnFlow(from: account, amount: 500.0, blockchain: blockchain)
```

We have to pass around the `blockchain` instance, to all these 4 helper functions, as they are no longer defined on the `Test.Blockchain` type.

turbolent (approved):
Nice!

SupunS (approved):

PR #2704, opened 2023-08-09
Hex encode user input in error to avoid invalid UTF-8

### LINKED ISSUE #2703: Error message can include invalid UTF-8 from user input
### Current Behavior

Error message includes actualPrefix as a string.  However, user input can contain invalid UTF-8.

Archive node can encounter CBOR text containing invalid UTF-8.  For example:

https://flowscan.org/transaction/f6c8e65646a3b140902aa7559ae2e740bbe92fbef65f414a441c141340a5756f

### Expected Behavior

Errors should not have invalid UTF-8 text.  It should be hex encoded to make sure errors contain valid UTF-8.

### Steps To Reproduce

See https://flowscan.org/transaction/f6c8e65646a3b140902aa7559ae2e740bbe92fbef65f414a441c141340a5756f

### Environment

```markdown
- Cadence version:
- Network:
```

fxamacker:
Closes #2703 

Currently, error message can include invalid UTF-8 when it shows content of actualPrefix (user input).

This PR hex encodes actualPrefix in the error message.

https://flowscan.org/transaction/f6c8e65646a3b140902aa7559ae2e740bbe92fbef65f414a441c141340a5756f

______

### DIFF
--- a/encoding/json/decode.go
+++ b/encoding/json/decode.go
@@ -301,7 +301,7 @@ func (d *Decoder) decodeAddress(valueJSON any) cadence.Address {
 	actualPrefix := v[:prefixLength]
 	if actualPrefix != addressPrefix {
 		panic(errors.NewDefaultUserError(
-			"invalid address prefix: expected `%s`, got `%s`",
+			"invalid address prefix: (shown as hex) expected %x, got %x", // hex encoding user input (actualPrefix) avoids invalid UTF-8.
 			addressPrefix,
 			actualPrefix,
 		))
--- a/encoding/json/encoding_test.go
+++ b/encoding/json/encoding_test.go
@@ -197,6 +197,16 @@ func TestEncodeAddress(t *testing.T) {
 	)
 }
 
+func TestDecodeInvalidAddress(t *testing.T) {
+
+	t.Parallel()
+
+	msg := `{"type":"Address","value":"000000000102030405"}`
+
+	_, err := json.Decode(nil, []byte(msg))
+	require.ErrorContains(t, err, "invalid address prefix: (shown as hex) expected 3078, got 3030")
+}
+
 func TestEncodeInt(t *testing.T) {
 
 	t.Parallel()


sideninja:
Thank you

encoding/json/decode.go
```diff
@@ -301,7 +301,7 @@ func (d *Decoder) decodeAddress(valueJSON any) cadence.Address {
 	actualPrefix := v[:prefixLength]
 	if actualPrefix != addressPrefix {
 		panic(errors.NewDefaultUserError(
-			"invalid address prefix: expected `%s`, got `%s`",
+			"invalid address prefix: expected `%s`, got 0x%x", // hex encoding user input (actualPrefix) avoids invalid UTF-8.
```
turbolent, line 304: Could you please add a test case? (to prevent regression)

turbolent (commented):
👍

encoding/json/encoding_test.go
```diff
@@ -197,6 +197,16 @@ func TestEncodeAddress(t *testing.T) {
...
 }
 
+func TestDecodeInvalidAddress(t *testing.T) {
+
+	t.Parallel()
+
+	msg := `{"type":"Address","value":"000000000102030405"}`
+
+	_, err := json.Decode(nil, []byte(msg))
+	require.ErrorContains(t, err, "invalid address prefix: expected `0x`, got 0x3030")
```
turbolent, line 207: I see what we're trying to fix, but the new error message has also quite the potential for confusion:
If someone looks just at the error message, it looks like it is wrong – `0x3030` has the prefix `0x`.
Of course, it is the hex encoding of the characters (`'0'` is `0x30`), but the hex-encoding (`0x3030`) in the error message looks as if it was the raw input. 

Maybe indicate that this is the hex-encoding of the value, not the value itself.

turbolent (approved):

encoding/json/encoding_test.go
fxamacker: Thanks, I agree.  I improved it by removing "0x" and encoding both values to hex like this:

`"invalid address prefix: (shown as hex) expected %x, got %x"`

PR #2717, opened 2023-08-15
Give `CompositeTypeHandler` higher precedence over `getUserCompositeType`

m-Peter:
Closes https://github.com/onflow/cadence-tools/issues/172

## Description

When running the following script, in the Flow Emulator:
```cadence
pub fun main(): Type? {
    return CompositeType("flow.AccountContractAdded")
}
```

The following error is returned:
```bash
error: [Error Code: 1054] location (flow) is not a valid location: expecting an AddressLocation, but other location types are passed
 --> b2b8424d1ef6de8f49bd006c8b7cbdba4a98c3be6e5834fb4ed1b281655293fe:2:4
  |
2 |     return CompositeType("flow.AccountContractAdded")
```

After the proposed change, the script runs successfully:
```bash
Result: Type<flow.AccountContractAdded>()
```

**Note:** This seemed to work in the test suite of the `cadence-tools/test` repository, however, it didn't work when I tried it in an actual Cadence test file.
______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -4797,16 +4797,16 @@ func (interpreter *Interpreter) GetCompositeType(
 			return compositeType, nil
 		}
 	} else {
-		compositeType = interpreter.getUserCompositeType(location, typeID)
-		if compositeType != nil {
-			return compositeType, nil
+		config := interpreter.SharedState.Config
+		compositeTypeHandler := config.CompositeTypeHandler
+		if compositeTypeHandler != nil {
+			compositeType = compositeTypeHandler(location, typeID)
+			if compositeType != nil {
+				return compositeType, nil
+			}
 		}
-	}
 
-	config := interpreter.SharedState.Config
-	compositeTypeHandler := config.CompositeTypeHandler
-	if compositeTypeHandler != nil {
-		compositeType = compositeTypeHandler(location, typeID)
+		compositeType = interpreter.getUserCompositeType(location, typeID)
 		if compositeType != nil {
 			return compositeType, nil
 		}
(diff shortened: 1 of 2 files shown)


runtime/interpreter/interpreter.go
```diff
@@ -4797,16 +4797,16 @@ func (interpreter *Interpreter) GetCompositeType(
 			return compositeType, nil
 		}
 	} else {
-		compositeType = interpreter.getUserCompositeType(location, typeID)
```
SupunS, line 4800: I wonder if this is the real issue. You had mentioned this works correctly in `cadence-tools/test` but not working in emulator/cli? Could that be because of the way certain APIs are implemented by those? e.g. The implementation of `importResolver` (and/or some other interface methods) could be different in the two repos?

This fix also works, but trying to figure out if there's a different/bigger issue that might get buried by this fix, and might pop again later in a different form.

m-Peter: That's a good point @SupunS :+1: 

I have opened a PR (https://github.com/onflow/cadence-tools/pull/195/files), which adds the relevant assertions for loading of standard library flow events. As you can see in the changed files, the `importResolver` implementation there is a NO-OP:

```go
importResolver := func(location common.Location) (string, error) {
	return "", nil
}

runner := NewTestRunner().WithImportResolver(importResolver)
```
This is probably the reason that it works only in the testing suite, and not in the actual testing framework (released build).

Since the `CompositeTypeHandler` was introduced rather recently (https://github.com/onflow/cadence/pull/2533), I am guessing that the misbehavior arose from that PR. It's rather strange that it does not work for either Emulator or the Cadence Testing Framework.

SupunS: > This is probably the reason that it works only in the testing suite, and not in the actual testing framework (released build).

yeah, that seems a possible root-cause 👍

SupunS (approved):
LGTM!

turbolent (approved):
Makes sense to ask the handler first, if available, and then check for a user-defined type otherwise 👍

Issue #2471, opened 2023-04-26
Improve interface default functions


SupunS:
### Issue to be solved

Originated from discussion: https://github.com/onflow/cadence/pull/2112#discussion_r1178115861

Currently, it is invalid for one interface to provide a default implementation, and another to declare the function definition only.

```cadence
struct interface A {
    // Have default impl
    pub fun hello() {
        var a = 1
    }
}
            
struct interface B {
    // Declaration only
    pub fun hello()
}

struct C: A, B {}  // Error: missing default implementation for 'B.foo'
```

This is too restrictive, and also avoids use-cases where certain interfaces only want to define pre/post conditions, while some other interface provides the default implementation.

### Suggested Solution

The above example should not give errors.

Furthermore, interface `B` should be able to define a pre/post condition. e.g:

```cadence
struct interface A {
    // Have default impl
    pub fun hello() {
        var a = 1
    }
}
            
struct interface B {
    // Pre-condition only
    pub fun hello() {
       pre { someCondition() }
    }
}

struct C: A, B {}
```

bluesign:
This is restrictive I agree, but also isn't it preventitive ? Chances are my contract ( in this case C ) will break is obviously higher in this case vs B not having hello method.

turbolent:
Is this related to discussion we had in the PR that added interface default functions? https://github.com/onflow/cadence/pull/1076#discussion_r675861413

SupunS:
Yeah, it is the same.

The concern now is, with interface inheritance, that particular case can become a useful pattern, where interfaces in the chain might define just conditions (which is part of the restrictions that you would get by conforming to that interface). But with the current restrictions, it makes it impossible to add such conditions if some other interface provides a default implementation already or vise-versa.
e.g:
```cadence
pub resource interface Receiver {
    pub fun deposit(_ something: @AnyResource)
}

pub resource interface Vault: Receiver {
    pub fun withdraw(_ amount: Int): @Vault {
        // some default implementation
    }
}

pub resource interface ThrottledVault: Vault {
    pub fun withdraw(_ amount: Int): @Vault {    // will report an error
        pre { amount < 50 }
    }
}

^ might not be the best example, but just want to show the idea.

Also, with stable-cadence, since conditions can now be view-only, I think some of the original concerns had at that time are no longer applicable? e.g: order of conditions isn't going tobe a problem
```

turbolent:
### Current behaviour of conflicting default functions

[...]

PR #2725, opened 2023-08-16
Allow default functions to co-exist with empty function declarations

SupunS:
Implementation of https://github.com/onflow/flips/pull/134

Closes https://github.com/onflow/cadence/issues/2471

## Description

Allow default functions to co-exist with empty function declarations.

______


runtime/error_test.go
```diff
@@ -483,133 +483,6 @@ func TestRuntimeError(t *testing.T) {
 	})
 }
 
-func TestRuntimeDefaultFunctionConflictPrintingError(t *testing.T) {
```
SupunS, line 486: This error is no longer produced. Below test case tests the related other errors.

turbolent:
Could you please keep the current behaviour (report an error) based on a feature flag? 

That way, we could already merge the feature into the Stable Cadence feature branch, and then assume the feature will end up in the final release and just enable the feature flag in the previous build. 

That way we're not blocked on the FLIP, and there's nothing to revert in case it gets rejected.

runtime/error_test.go
turbolent: The test seems to have tested the error printing. Maybe there was a problem in the error type before? In that case the test can be removed. Maybe the test was also exercising something in the error pretty printer? (the long body in test function is suspicious)

turbolent (approved):
Nice!

Only one concern regarding the removed test case

runtime/error_test.go
SupunS: The error pretty printing for related other errors is tested in the test immediately below this one (e.g: `TestRuntimeMultipleInterfaceDefaultImplementationsError`). This one tests the same for `DefaultFunctionConflictError` specifically, which no longer exists. So I think this is safe to remove.

SupunS: cc: @dsainati1

dsainati1: Yeah, if the error no longer exists it makes sense to me to remove this test.

dsainati1 (approved):

PR #2729, opened 2023-08-17
Add a utility method to the `Test.Blockchain` struct for moving the blockchain time

m-Peter:
**Companion PR:** https://github.com/onflow/cadence-tools/pull/196

## Description

Add the `TestFramework.MoveTime(int64)`, so that test providers can allow the manipulation of blockchain's time.

Example usage:
```cadence
import Test

pub fun test() {
    let blockchain = Test.newEmulatorBlockchain()
    // timeDelta is the representation of 35 days,
    // in the form of seconds.
    let timeDelta = UFix64(35 * 24 * 60 * 60)
    blockchain.moveTime(by: timeDelta)
}
```
______

### DIFF
--- a/runtime/stdlib/contracts/test.cdc
+++ b/runtime/stdlib/contracts/test.cdc
@@ -127,6 +127,13 @@ pub contract Test {
         pub fun reset(to height: UInt64) {
             self.backend.reset(to: height)
         }
+
+        /// Moves the time of the blockchain by the given delta,
+        /// which should be passed in the form of seconds.
+        ///
+        pub fun moveTime(by delta: UFix64) {
+            self.backend.moveTime(by: delta)
+        }
     }
 
     pub struct Matcher {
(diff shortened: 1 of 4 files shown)


runtime/stdlib/contracts/test.cdc
```diff
@@ -312,6 +319,11 @@ pub contract Test {
         /// Resets the state of the blockchain to the given height.
         ///
         pub fun reset(to height: UInt64)
+
+        /// Moves the time of the blockchain by the given delta,
+        /// which should be passed in the form of seconds.
+        ///
+        pub fun moveTime(by delta: UFix64)
```
SupunS, line 326: Seems in the implementation, we are trimming off the decimal points. So maybe it would make sense to just use `UInt64` here?

turbolent: This is probably for consistency with [`Block.timestamp`](https://github.com/onflow/cadence/blob/e287ab0316547b0a7bd6a527eb9b67115d6dc834/runtime/sema/block.cdc#L26C13-L26C30)

turbolent (approved):
Nice!

runtime/stdlib/contracts/test.cdc
SupunS: oh, that makes sense 👍

SupunS (approved):

runtime/stdlib/contracts/test.cdc
m-Peter: Yes, exactly, just like Bastian mentioned.

PR #2733, opened 2023-08-17
Add test for inner container mutation while iterating outer

### LINKED ISSUE #2663: Add tests for child container being mutated while parent container is iterated
It would be good to have this for https://github.com/onflow/atree/issues/292.

### Issue to be solved

Currently, child containers are allowed to be mutated while their parent container is iterated.  However, there does not appear to be any tests for this.

### Suggested Solution

Add tests for child container being mutated while parent container is iterated.

~~Alternatively, determine if it is feasible to not allow child containers being mutated while parent container is iterated.~~

For more details, see Discord [thread](https://discord.com/channels/613813861610684416/1108479699732152503/1130905641851035700).

SupunS:
Closes #2663

## Description

So the current behavior is since during the iteration, what is available for each iteration is a copy of the element (i.e. copy of the child/inner container). So any mutation on that copy wouldn't affect the outer container.

______


turbolent:
Thank you for picking this up @SupunS! 👍 

Could you please also add test cases where mutations are not on copies, but on the original data?

SupunS:
> Could you please also add test cases where mutations are not on copies, but on the original data?

Don't think that is possible. The only case where data is not copied is for resources, but then again iterating resources is not allowed.

SupunS:
Just realized it can actually mutate the original data, if the update happens via the outer container, rather than directly mutating the loop variable. i.e. mutating using `outer[a][b] = something`  rather than doing `inner[0] = something` inside the loop.

Added more tests: b775cdaf1c7be2e14de40a9d36fe29aecdbffaf9

dsainati1 (approved):

turbolent (approved):
Thank you for adding the additional tests 👍

fxamacker:
Thank you @SupunS for tackling this!  Really appreciate it!  :+1:

PR #2735, opened 2023-08-18
Expose `Test.Blockchain` as a singleton through `Test.emulatorBlockchain` public field

m-Peter:
Closes https://github.com/onflow/cadence-tools/issues/139

## Description

We also add a deprecation hint for `Test.newEmulatorBlockchain()`.

______

### DIFF
--- a/runtime/stdlib/test_contract.go
+++ b/runtime/stdlib/test_contract.go
@@ -439,6 +439,50 @@ func (t *TestContractType) newNewEmulatorBlockchainFunction(
 	)
 }
 
+// 'Test.emulatorBlockchain' public field
+
+const testTypeEmulatorBlockchainFieldDocString = `
+Returns a blockchain singleton which is backed by an emulator instance.
+`
+
+const testTypeEmulatorBlockchainFieldName = "emulatorBlockchain"
+
+func (t *TestContractType) newEmulatorBlockchain(
+	inter *interpreter.Interpreter,
+	testFramework TestFramework,
+	parent *interpreter.CompositeValue,
+) interpreter.Value {
+	locationRange := interpreter.EmptyLocationRange
+
+	// Create an `EmulatorBackend`
+	emulatorBackend := t.emulatorBackendType.newEmulatorBackend(
+		inter,
+		testFramework,
+		locationRange,
+	)
+
+	// Create a 'Blockchain' struct value, that wraps the emulator backend,
+	// by calling the constructor of 'Blockchain'.
+	blockchainConstructor := getNestedTypeConstructorValue(
+		parent,
+		testBlockchainTypeName,
+	)
+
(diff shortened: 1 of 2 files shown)


turbolent:
With the singleton, do tests need to make sure that state is reset at the start of each test case?

m-Peter:
@turbolent The current problem is that `Test.newEmulatorBlockchain` creates just a new `Test.Blockchain` struct, which still encapsulates the same emulator instance. This comes as a surprise to developers.

To counter that, my suggestion is to expose only a singleton for `Test.Blockchain` and it is up to developers to decide how to setup their test cases.

We have `beforeEach/afterEach` hooks, which can reset the blockchain state at any given block height. This will allow them to still maintain the operations on the `setup` helper, which usually involves deploying contracts to certain accounts.

Or developers can choose to write their tests sequentially, building up on previous emulator state.

Any thoughts?

turbolent:
@m-Peter 

> The current problem is that `Test.newEmulatorBlockchain` creates just a new `Test.Blockchain` struct, which still encapsulates the same emulator instance. This comes as a surprise to developers.

Definitely see the issue here and agree we should fix it. Refactoring to a singleton to make the current implementation more obvious to developers is a good start. 

Given that stateless tests are a best practice, 
just wondering if we could instead "fix"/adjust the implementation of the current API (`Test.newEmulatorBlockchain`) and deliver on its promise of spinning up a new Emulator-backed blockchain instance. 

For example, Cadence tests are already often written in a style where each test case spins up their own Emulator blockchain, e.g. [in the FT repo](https://github.com/onflow/flow-ft/blob/b723f1eee3f96d4385faf1d17649cded222c1b3e/lib/go/test/test.go#L26).

We can definitely look into the `newEmulatorBlockchain` implementation together. Do you have a pointer to where it's implemented? (in `cadence-tools/test` right?)

m-Peter:
We can go down that road as easily, I can change the `Test.newEmulatorBlockchain` to return a new Emulator-backed blockchain each time. The relevant piece of code is located here: https://github.com/onflow/cadence-tools/blob/master/test/test_runner.go#L511-L523. All we need to do is to pass a closure that can instantiate a fresh backend each time, and not just a concrete backend. What do you think? Should I drop this PR and go for this alternative?

turbolent:
@m-Peter If it's not too much effort and you are OK with it, yes, I would recommend dropping this PR and going with the alternative, it seems less error prone for developers.

SupunS:
+1 for fixing the `Test.newEmulatorBlockchain` to return a new Emulator-backed every time, rather than the singleton. I think that is the desired behavior.

m-Peter:
Fair point :raised_hands: I am closing this PR and will follow-up with a new one, implementing the agreed functionality.

PR #2737, opened 2023-08-18
Introduce `map` in Fixed/Variable sized Array types

darkdrag00nv2:
Work towards https://github.com/onflow/cadence/issues/2605

## Description

Introduce `map` function for creating a copy of an Array value with its entries transformed by a transformation function.

This function would be unavailable to resource arrays. It can be enabled for resources as a reference in the future.

Will send docs PR post the merge.
______

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -2482,6 +2482,35 @@ func (v *ArrayValue) GetMember(interpreter *Interpreter, locationRange LocationR
 				)
 			},
 		)
+
+	case sema.ArrayTypeMapFunctionName:
+		return NewHostFunctionValue(
+			interpreter,
+			sema.ArrayMapFunctionType(
+				interpreter,
+				v.SemaType(interpreter),
+			),
+			func(invocation Invocation) Value {
+				interpreter := invocation.Interpreter
+
+				funcArgument, ok := invocation.Arguments[0].(FunctionValue)
+				if !ok {
+					panic(errors.NewUnreachableError())
+				}
+
+				transformFunctionType, ok := invocation.ArgumentTypes[0].(*sema.FunctionType)
+				if !ok {
+					panic(errors.NewUnreachableError())
+				}
+
+				return v.Map(
+					interpreter,
+					invocation.LocationRange,
+					funcArgument,
+					transformFunctionType,
+				)
+			},
+		)
 	}
 
 	return nil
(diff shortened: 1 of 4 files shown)


runtime/interpreter/value.go
```diff
@@ -3058,6 +3087,91 @@ func (v *ArrayValue) Filter(
...
+		invocation := NewInvocation(
+			interpreter,
+			nil,
+			nil,
+			[]Value{arrayElement},
+			elementTypeSlice,
+			nil,
+			locationRange,
+		)
+		return invocation
```
turbolent, line 3108:
```suggestion
		return NewInvocation(
			interpreter,
			nil,
			nil,
			[]Value{arrayElement},
			elementTypeSlice,
			nil,
			locationRange,
		)
```

runtime/interpreter/value.go
```diff
@@ -3058,6 +3087,91 @@ func (v *ArrayValue) Filter(
...
+	if !ok {
+		panic(errors.NewUnreachableError())
+	}
+
+	var returnArrayStaticType ArrayStaticType
+	switch v.Type.(type) {
+	case VariableSizedStaticType:
+		returnArrayStaticType = NewVariableSizedStaticType(
+			interpreter,
+			procedureStaticType.ReturnType(interpreter),
```
turbolent, line 3121: Maybe refactor this repeated expression in both cases into a variable declaration before the switch

runtime/interpreter/value.go
```diff
@@ -3058,6 +3087,91 @@ func (v *ArrayValue) Filter(
...
+			}
+
+			// Also handles the end of array case since iterator.Next() returns nil for that.
+			if atreeValue == nil {
+				return nil
+			}
+
+			value = MustConvertStoredValue(interpreter, atreeValue)
+			if value == nil {
+				return nil
```
turbolent, line 3159: When/why would `MustConvertStoredValue` return `nil`? Maybe remove this if-statement

turbolent (commented):
Nice!

runtime/interpreter/value.go
darkdrag00nv2: Done.

runtime/interpreter/value.go
darkdrag00nv2: Yeah, makes sense. I've updated it to the pattern used in similar functions elsewhere.

runtime/interpreter/value.go
```diff
@@ -3058,6 +3087,89 @@ func (v *ArrayValue) Filter(
...
+				panic(errors.NewExternalError(err))
+			}
+
+			if atreeValue != nil {
+				value = MustConvertStoredValue(interpreter, atreeValue)
+			}
+
+			if value == nil {
+				return nil
+			}
```
turbolent, line 3158: Sorry, I should have been more clear here: 
- ` iterator.Next` may return nothing, so exit from the array construction callback if there's nothing left to insert into the array
- `MustConvertStoredValue` will never return `nil`, so no need to handle it

```suggestion
			atreeValue, err := iterator.Next()
			if err != nil {
				panic(errors.NewExternalError(err))
			}

			if atreeValue == nil {
				return nil
			}

			value := MustConvertStoredValue(interpreter, atreeValue)
```

turbolent (approved):

runtime/interpreter/value.go
darkdrag00nv2: Yeah, understood. The reason why I wrote it this way is because that's how it is written in other similar functions such as `forEachKey` in dictionary. But your proposed style makes sense and probably clearer.

runtime/tests/checker/arrays_dictionaries_test.go
```diff
@@ -1225,6 +1225,104 @@ func TestCheckResourceArrayFilterInvalid(t *testing.T) {
...
+
+	_, err := ParseAndCheck(t, `
+		fun test() {
+			let x = [1, 2, 3]
+			let trueForEven =
+				fun (_ x: Int): Bool {
+					return x % 2 == 0
+				}
+
+			let y = x.map(trueForEven)
```
SupunS, line 1240: For all the checker tests, maybe declare the `y` variable with explicit type, so the test will also check whether the return type of the map function in each case is correct. Currently, it is not being tested/asserted.

```suggestion
			let y: [Int] = x.map(trueForEven)
```

SupunS (approved):

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -11085,6 +11085,475 @@ func TestInterpretArrayFilter(t *testing.T) {
 	})
 }
 
+func TestInterpretArrayMap(t *testing.T) {
+
```
turbolent, line 11089:
```suggestion
	t.Parallel()

```

darkdrag00nv2: Added.

darkdrag00nv2:
@turbolent @SupunS Addressed feedback. Should be ready to merge!

SupunS (approved):

PR #2743, opened 2023-08-24
Composition of entitlement mappings

dsainati1:
Closes https://github.com/onflow/cadence/issues/2643

This implements the proposed `include` syntax for entitlement mappings, allowing them to "inherit" relations from other maps for less code duplication. In general, `include A` for some entitlement map `A` is equivalent to just copy-pasting all of `A`s relations into the new map. 

Inclusions are all "resolved" when the map's definition is type checked, so that when they are "used" by a field or function access, all inherited relations are already present in the map, i.e. the hierarchy is flattened. 

______

### DIFF
--- a/runtime/ast/entitlement_declaration.go
+++ b/runtime/ast/entitlement_declaration.go
@@ -165,8 +165,6 @@ func (d *EntitlementMapRelation) Doc() prettier.Doc {
 	)
 }
 
-func (*NominalType) isEntitlementMapElement() {}
-
 // EntitlementMappingDeclaration
 type EntitlementMappingDeclaration struct {
 	Access     Access
--- a/runtime/ast/entitlement_declaration.go
+++ b/runtime/ast/entitlement_declaration.go
@@ -125,19 +125,26 @@ func (d *EntitlementDeclaration) String() string {
 	return Prettier(d)
 }
 
-type EntitlementMapElement struct {
+type EntitlementMapElement interface {
+	isEntitlementMapElement()
+	Doc() prettier.Doc
+}
+
+type EntitlementMapRelation struct {
 	Input  *NominalType
 	Output *NominalType
 }
 
-func NewEntitlementMapElement(
+var _ EntitlementMapElement = &EntitlementMapRelation{}
+
+func NewEntitlementMapRelation(
 	gauge common.MemoryGauge,
 	input *NominalType,
 	output *NominalType,
-) *EntitlementMapElement {
+) *EntitlementMapRelation {
 	common.UseMemory(gauge, common.EntitlementMappingElementMemoryUsage)
 
-	return &EntitlementMapElement{
+	return &EntitlementMapRelation{
 		Input:  input,
 		Output: output,
 	}
@@ -145,7 +152,9 @@ func NewEntitlementMapElement(
 
 var arrowKeywordSpaceDoc = prettier.Text(" -> ")
 
-func (d EntitlementMapElement) Doc() prettier.Doc {
+func (*EntitlementMapRelation) isEntitlementMapElement() {}
+
+func (d *EntitlementMapRelation) Doc() prettier.Doc {
 	var doc prettier.Concat
 
 	return append(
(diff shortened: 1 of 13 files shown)


runtime/ast/entitlement_declaration.go
```diff
@@ -162,6 +162,7 @@ type EntitlementMappingDeclaration struct {
 	DocString    string
 	Identifier   Identifier
 	Associations []*EntitlementMapElement
+	Inclusions   []*NominalType
```
turbolent, line 165: It might be good to merge `Associations` and `Inclusions`, so that the order of the declarations is preserved. This would be similar to `CompositeDeclaration.Members`, which keeps e.g. fields, functions, etc. in order `Members.desclarations`, but provides getters for the individual kinds of declarations.

This enables the AST to be formatted back into how it was parsed.

runtime/sema/check_interface_declaration.go
```diff
@@ -557,6 +557,69 @@ func (checker *Checker) declareEntitlementMappingType(declaration *ast.Entitleme
...
+// Uses the sync primitive stored in `resolveInclusions` to ensure each map type's includes are computed only once.
+// This assumes that any includes coming from imported files are necessarily already completely resolved, since that imported file
+// must necessarily already have been fully checked. Additionally, because import cycles are not allowed in Cadence, we only
+// need to check for map-include cycles within the currently-checked file
+func (checker *Checker) resolveEntitlementMappingInclusions(
+	mapType *EntitlementMapType,
+	declaration *ast.EntitlementMappingDeclaration,
+	visitedMaps map[*EntitlementMapType]struct{},
+) {
+	mapType.resolveInclusions.Do(func() {
```
turbolent, line 570: Maybe make this a method on `EntitlementMapType`, it is a bit odd that `Checker` reaches into `EntitlementMapType` and updates its internal state (e.g. uses `resolveInclusions`)

runtime/sema/errors.go
```diff
@@ -4360,6 +4360,96 @@ func (e *InvalidMappedAuthorizationOutsideOfFieldError) EndPosition(common.Memor
...
+	)
+}
+
+func (e *InvalidEntitlementMappingInclusionError) StartPosition() ast.Position {
+	return e.StartPos
+}
+
+func (e *InvalidEntitlementMappingInclusionError) EndPosition(common.MemoryGauge) ast.Position {
+	return e.EndPos
+}
```
turbolent, line 4391: Embedding `ast.Range` should already provide these methods

```suggestion
```

runtime/sema/errors.go
```diff
@@ -4360,6 +4360,96 @@ func (e *InvalidMappedAuthorizationOutsideOfFieldError) EndPosition(common.Memor
...
+	)
+}
+
+func (e *DuplicateEntitlementMappingInclusionError) StartPosition() ast.Position {
+	return e.StartPos
+}
+
+func (e *DuplicateEntitlementMappingInclusionError) EndPosition(common.MemoryGauge) ast.Position {
+	return e.EndPos
+}
```
turbolent, line 4421: Same as above

```suggestion
```

runtime/sema/errors.go
```diff
@@ -4360,6 +4360,96 @@ func (e *InvalidMappedAuthorizationOutsideOfFieldError) EndPosition(common.Memor
...
+
+func (e *CyclicEntitlementMappingError) StartPosition() ast.Position {
+	return e.StartPos
+}
+
+func (e *CyclicEntitlementMappingError) EndPosition(common.MemoryGauge) ast.Position {
+	return e.EndPos
+}
+
```
turbolent, line 4452: Same as above

```suggestion
```

turbolent (commented):
Great work!

Could you please also add some interpreter tests?

turbolent (approved):
Great work!

runtime/ast/entitlement_declaration.go
```diff
@@ -156,12 +165,14 @@ func (d EntitlementMapElement) Doc() prettier.Doc {
 	)
 }
 
+func (*NominalType) isEntitlementMapElement() {}
```
SupunS, line 168: Maybe move this to the same location where `NominalType` defined.
Also add:
```go
var _ EntitlementMapElement = &NominalType{}
```

runtime/entitlements_test.go
```diff
@@ -1107,3 +1107,169 @@ func TestRuntimeCapabilityEntitlements(t *testing.T) {
...
+        access(all) contract FurtherUpstream {
+            access(all) entitlement X
+			access(all) entitlement Y
+			access(all) entitlement Z
+
+			access(all) entitlement mapping M {
+				X -> Y 
+				Y -> Z
+			}
+        }
```
SupunS, line 1128: just a nit: string literals here and below seem to have a mix of tabs and whitespaces. Would be nice to have only one of them.

SupunS (approved):
Nice!

PR #2744, opened 2023-08-25
Add the `NewEmulatorBackend` method on `TestFramework` interface

m-Peter:
Closes https://github.com/onflow/cadence-tools/issues/139

## Description

This method is used to create a new emulator backend, each time a developer calls `Test.newEmulatorBlockchain()`.
Up to this point, one and the same emulator backend was used, which meant that there was no way to reset state.

______


runtime/stdlib/test_test.go
```diff
@@ -2084,25 +2091,27 @@ func TestBlockchain(t *testing.T) {
...
 
-		    pub struct Foo {}
+                Test.expect(events, Test.beEmpty())
+            }
 		`
 
 		eventsInvoked := false
 
-		testFramework := &mockedTestFramework{
+		emulatorBackend := &mockedTestFramework{
```
turbolent, line 2114: Here and below: Maybe also move this into `newEmulatorBackend`, so a new `&mockedTestFramework` is created on each call of the `newEmulatorBackend` function

turbolent (approved):
Nice! Just one small suggestion

runtime/stdlib/test_test.go
m-Peter: Ah yes, good suggestion :100: 
Now it's the same instance again :sweat_smile:

runtime/stdlib/test_test.go
```diff
@@ -2048,18 +2048,20 @@ func TestBlockchain(t *testing.T) {
...
+                let events = blockchain.events()
+
+                Test.expect(events, Test.beEmpty())
+            }
 		`
 
 		eventsInvoked := false
 
-		testFramework := &mockedTestFramework{
+		emulatorBackend := &mockedTestFramework{
```
turbolent, line 2064: Also inline this variable

runtime/stdlib/test-framework.go
```diff
@@ -73,6 +73,8 @@ type TestFramework interface {
 	Reset(uint64)
 
 	MoveTime(int64)
+
+	NewEmulatorBackend() TestFramework
```
SupunS, line 77: `TestFramework` returning another `TestFramework` seems a bit strange/can be confusing.

Instead, could we split the current `TestFramework` into two:

```go
type TestFramework interface {

    NewEmulatorBackend() Blockchain

    ReadFile(string) (string, error)
}

type Blockchain interface {

    RunScript(...) *ScriptResult

    CreateAccount() (*Account, error)

    AddTransaction(...) error

    ExecuteNextTransaction() *TransactionResult

    CommitBlock() error

    DeployContract( ... )

    ...
}
```

m-Peter: Yeah, good point, this was feeling little bit counter-intuitive in the tests as well :sweat_smile: 
I will update it :+1:

m-Peter: I have been thinking about moving all these into the `cadence-tools/test` repository, if it's possible. I haven't checked if all the methods involved are actually publicly exposed.
What are your thoughts on this? I like the fact that Cadence exposes interfaces to be implemented by various providers, but in this case, the reality is that there's only one provider, and this somewhat slows down development. Certainly out of scope of this PR, just thought of bringing it up.

SupunS: The problem is that the standard libraries and the `Test` contract implementation is in the `cadence` repo. So moving this interface to `cadence-tools` would create a circular dependency between the two repos. That was the main reason to have this interface inside this repo.

m-Peter: I have extracted the `Blockchain` interface out of the `TestFramework` interface here: https://github.com/onflow/cadence/pull/2744/commits/6862627915b88a3c64d8de99dc34ffdd0c788616

SupunS (approved):
Nice!

PR #2751, opened 2023-08-25
Improve composite and interface static types

turbolent:
## Description

In #2648 I refactored composite static types and _removed_ support for a caching the type ID.

Instead, it is better to also add the type ID caching to interface static types.

While at it, clean up the uses of composite and interface static types and use the constructor functions instead of Go composite literals, to ensure the type ID is always available and correct.

Also, use `NewTypeIDFromQualifiedName` everywhere.

______


turbolent:
@SupunS Could you please have a look? It would unblock me to make the appropriate changes in https://github.com/onflow/cadence/pull/2648

SupunS (approved):
Nice!

PR #2754, opened 2023-08-28
sort entitlement sets alphabetically when generating strings

dsainati1:
Closes https://github.com/onflow/cadence/issues/2749

Sort entitlement sets when producing strings from them. This guarantees that equal but differently-ordered entitled references will hash to the same value. 
______

- [ ] Targeted PR against `master` branch
- [X] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [X] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [X] Updated relevant documentation 
- [X] Re-reviewed `Files changed` in the Github PR explorer
- [X] Added appropriate labels

PR #2750, opened 2023-08-25
Avoid unnecessary static to sema type conversions for `Type` 

turbolent:
Closes #2748 

## Description

`interpreter.StaticType` currently does not expose a mechanism to get the type's ID.

`interpreter.TypeValue` needs to get a type ID in two places:
- For the `identifier` field
- For the `HashInput` function, e.g. used when the type value is used as a dictionary key

`interpreter.TypeValue` calculated the type ID by converting the static type to a sema type. This causes the whole program to be loaded, which is inefficient or even unavailable (e.g. during a state migration).

Introduce `func StaticType.ID() TypeID` to allow getting the type ID for a static type without overhead, and remove the unnecessary conversions from static to sema type.

cc @janezpodhostnik 

______

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #2756, opened 2023-08-31
Improve and fix static types and their ID and string functions

turbolent:
Depends on #2648 

## Description

So far, both `sema.Type` and `cadence.Type` had `ID` functions, which returned the type ID. With #2750, also `interpreter.StaticType` got an `ID` function. 

Most of these functions re-implemented the ID generation, which is hard to maintain and error prone (inconsistencies). In some cases the ID generation was mixed with the string (human-readable) representation, and the string formatting was re-implemented as both metered and unmetered variants.  

Reduce the potential for inconsistencies by implementing the ID generation in a central place, as `sema.Format*TypeID` functions, and use them in the `ID` functions of all three Go types representing types.

#2754 had previously already started to sort entitlement sets. 
We also need to sort type IDs in intersection type IDs.

Avoid code duplication in string formatting functions, by delegating unmetered functions (`String`) to metered ones (`MeteredString`). Also fix memory metering along the way (in some cases memory was allocated before it was metered).

Ensure that the type ID generation for entitlement sets in entitlement set authorizations and for intersection types sorts the type IDs. Keep the user-defined order in string formatting functions (e.g. `String` and `QualifiedString`). Avoid allocating multiple ordered maps when generating the sorted type IDs.

Add tests for the ID and string functions of authorizations, reference types, and intersection types, on all three levels (`sema`, `interpreter`, and `cadence`).

Also:
- Refactor `interpreter.StaticType`s to be pointer types, as they might get nested and could potentially lead to lots of data being copied. This wasn't really necessary, but noticed a bit late, but maybe still a good improvement.
- Remove any type ID caching, as it might result in lots of memory being allocated
- Improve the tests for entitlements, e.g. use `assert` where it makes sense, fix argument order (expected, actual), avoid repeated casts, etc. 

______

### DIFF
--- a/runtime/sema/access.go
+++ b/runtime/sema/access.go
@@ -19,7 +19,7 @@
 package sema
 
 import (
-	"fmt"
+	"sort"
 	"strings"
 	"sync"
 
@@ -31,17 +31,14 @@ import (
 
 type Access interface {
 	isAccess()
+	ID() TypeID
+	String() string
+	QualifiedString() string
+	Equal(other Access) bool
 	// IsLessPermissiveThan returns whether receiver access is less permissive than argument access
 	IsLessPermissiveThan(Access) bool
 	// PermitsAccess returns whether receiver access permits argument access
 	PermitsAccess(Access) bool
-	Equal(other Access) bool
-	string(func(ty Type) string) string
-	Description() string
-	// string representation of this access when it is used as an access modifier
-	AccessKeyword() string
-	// string representation of this access when it is used as an auth modifier
-	AuthKeyword() string
 }
 
 type EntitlementSetKind uint8
@@ -50,6 +47,8 @@ const (
 	Conjunction EntitlementSetKind = iota
 	Disjunction
 )
+
+// EntitlementSetAccess
 
 type EntitlementSetAccess struct {
 	Entitlements *EntitlementOrderedSet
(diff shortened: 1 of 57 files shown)


dsainati1 (approved):

runtime/sema/access.go
```diff
@@ -74,48 +74,93 @@ func NewEntitlementSetAccess(
...
-	return fmt.Sprintf("auth(%s)", a.AccessKeyword())
+func FormatEntitlementSetTypeID[T ~string](entitlementTypeIDs []T, kind EntitlementSetKind) T {
+	var builder strings.Builder
+	var separator string
+
+	switch kind {
+	case Conjunction:
+		separator = ","
+	case Disjunction:
+		separator = "|"
```
SupunS, line 98: Should maybe keep a space after the separator?

runtime/sema/type.go
```diff
@@ -6860,22 +6869,23 @@ func (t *IntersectionType) Tag() TypeTag {
...
-		result.WriteString(intersectionString)
+		result.WriteString(string(interfaceString))
 	}
 	result.WriteByte('}')
 	return result.String()
 }
 
-func FormatIntersectionTypeID(intersectionStrings []string) string {
-	return formatIntersectionType("", intersectionStrings)
+func FormatIntersectionTypeID[T ~string](interfaceTypeIDs []T) T {
```
SupunS, line 6886: TIL: `~string` 👌

types.go
SupunS, line 1: If I remember correctly, the type ID caching was also used as an optimization for CCF. Might be worth checking the impact of this on the external encoding.
cc: @fxamacker

SupunS (approved):

runtime/sema/access.go
turbolent: We do not need and should not keep any spaces in type IDs, given that space usage matters (encoding in execution state).

types.go
turbolent: Yeah, there's a time/space tradeoff here, let's discuss it

PR #2760, opened 2023-09-01
PoC: WebAssembly API

turbolent:
## Description

This PoC explores the [idea of adding a WebAssembly API to Cadence](https://forum.onflow.org/t/idea-wasm-execution-engine-in-cadence/5164), i.e. exposing a WebAssembly VM to Cadence programs.

The PRs adds:
- A `WebAssembly` contract with an API inspired by the [WebAssembly JavaScript Interface](https://www.w3.org/TR/wasm-js-api-1/). The beginnings of the API support module instantiation, without imports, and access to function exports with `i32` and `i64` types.
- An implementation of the WebAssembly API based on [wasmtime](https://wasmtime.dev/). The VM is not yet configured for metering and limits have not been set.
- A test case which demonstrates loading a simple `add` `(i32, i32) -> i32` function, calling it with Cadence `UInt32` values, and getting back the result as a Cadence `UInt32`.

______

### DIFF
--- a/runtime/stdlib/webassembly.cdc
+++ b/runtime/stdlib/webassembly.cdc
@@ -1,0 +1,23 @@
+access(all)
+contract WebAssembly {
+
+    /// Compile WebAssembly binary code into a Module
+    /// and instantiate it with the given imports.
+    access(all)
+    view fun compileAndInstantiate(bytes: [UInt8]): &WebAssembly.InstantiatedSource
+
+    access(all)
+    struct InstantiatedSource {
+
+        /// The instance.
+        access(all)
+        let instance: &WebAssembly.Instance
+    }
+
+    struct Instance {
+
+        /// Get the exported value.
+        access(all)
+        view fun getExport<T: AnyStruct>(name: String): T
+    }
+}
--- a/runtime/webassembly_test.go
+++ b/runtime/webassembly_test.go
@@ -1,0 +1,257 @@
+/*
+ * Cadence - The resource-oriented smart contract programming language
+ *
+ * Copyright Dapper Labs, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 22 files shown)


runtime/webassembly_test.go
```diff
@@ -0,0 +1,257 @@
...
+				ty := functionType.Parameters[i].TypeAnnotation.Type
+
+				var convertedArgument any
+
+				switch ty {
+				case sema.Int32Type:
+					convertedArgument = int32(argument.(interpreter.Int32Value))
+
+				case sema.Int64Type:
+					convertedArgument = argument.(interpreter.Int64Value)
```
turbolent, line 161:
```suggestion
					convertedArgument = int64(argument.(interpreter.Int64Value))
```

runtime/webassembly_test.go
```diff
@@ -0,0 +1,257 @@
...
+	}, nil
+}
+
+func TestRuntimeWebAssembly(t *testing.T) {
+
+	t.Parallel()
+
+	runtime := newTestInterpreterRuntime()
+
+	addProgram := []byte{
```
turbolent, line 201:
```suggestion
	// A simple program which exports a function `add` with type `(i32, i32) -> i32`,
	// which sums the arguments and returns the result
	addProgram := []byte{
```

PratikDhanave:
hi @turbolent thank you for adding a WebAssembly API to Cadence.
I want to know that next task would exposing wasmtime table ?

darkdrag00nv2:
@turbolent Would it be possible to merge this into a feature branch? It'll allow me to open future PRs targeted on that branch.

runtime/stdlib/webassembly.cdc
```diff
@@ -0,0 +1,23 @@
+access(all)
+contract WebAssembly {
+
+    /// Compile WebAssembly binary code into a Module
+    /// and instantiate it with the given imports.
```
darkdrag00nv2, line 5: Do we (plan to) support imports here? I thought there won't be any imports. At least for the first cut.

turbolent: Ah, good catch. I had started writing out the API here, and had initially included support for e.g. imports, but then revised the API to only what is supported by the PoC, but forgot to update the comments. 
```suggestion
    /// and instantiate it. Imports are not yet supported.
```

Issue #2752, opened 2023-08-27
Implement `split` and `join` for String


darkdrag00nv2:
### Issue to be solved

A few more functions can be implemented for Strings. 

### Suggested Solution

This issue proposes adding two functions:

1. `split(_ s: String, _ delimiter: String): [String]`: Returns a string array after splitting the provided string based on the provided delimiter. Delimiter can be made optional and `,` can be used as default.
2. `join(_ strs: [String], _ separator: String): String`: Returns a string value after concatenating elements of the provided array of string with the given separator. Separator can be made optional and `,` can be used as default.

Taken from https://github.com/green-goo-dao/flow-utils/blob/main/cadence/contracts/StringUtils.cdc.

Will file issues for more functions in the future.

Epic: https://github.com/onflow/cadence/issues/1972

turbolent:
Good ideas!

- `join` seems easy to implement
- `split` is actually not as straight-forward to implement, given the semantics of strings and characters in Cadence:
  In Cadence, string equality is semantic and not byte-wise. That means that functions that search, like `split`, but also e.g. a `replace` function, need to consider this. I explained this in more detail in a previous PR that attempted to add a `replace` function: https://github.com/onflow/cadence/pull/1353#issuecomment-1010533088

darkdrag00nv2:
Nice, thank you for sharing. 

You mentioned on the `replace` PR that unicode normalization might be used in these cases. I agree with that too. Seems like we could leverage the unicode/norm package from the Go standard library: https://go.dev/blog/normalization. Along with the conversion to normalized form, it also provides an iterator (`norm.Iter`) over them which would be useful to map the characters back to the original string.

I'll give it a shot.

turbolent:
Yeah, for `String` and `Character` equality, we already use NFC via `norm.NFC`.

darkdrag00nv2:
@turbolent @SupunS 
Just a note, seems like in Dart - the `replace` function doesn't take the equivalence into account.

You can try out the following on https://dartpad.dev/.
```dart
print('Caf\u00E9'.replaceAll(RegExp(r'\u00E9'), 'X'));  // CafX
print('Caf\u0065\u0301'.replaceAll(RegExp(r'\u0065\u0301'), 'X'));  // CafX
print('Caf\u0065\u0301'.replaceAll(RegExp(r'\u00E9'), 'X'));  // Café
```

The same with Kotlin: https://pl.kotl.in/_F5jCJfXN
```kotlin
[...]

Issue #2466, opened 2023-04-24
Add purity annotations to newly added functions


SupunS:
### Issue to be solved

There are some native function types added to the `master` branch, after the purity annotation change was added in `stable-cadence`. e.g: 
- [x] [Address.fromBytes()](https://github.com/onflow/cadence/pull/2452)

When the stable-cadence branch is synced/merged to `master`, ensure these newly added functions have the proper purity annotation.

### Suggested Solution

Scan through all the newly added functions, and added annotations as required.
```[tasklist]
### Tasks
- [x] `String.join` added in #2762
- [x] `String.split` added in #2791
```

j1010001:
Do one last pass before the 1.0 release.

PR #2762, opened 2023-09-03
Introduce `String.join` function

darkdrag00nv2:
Work towards #2752

## Description

Adds `join(_ strs: [String], _ separator: String): String` function to Cadence. The function returns a string value after concatenating elements of the provided array of string with the given separator.
______

### DIFF
--- a/runtime/interpreter/value_string.go
+++ b/runtime/interpreter/value_string.go
@@ -84,22 +84,14 @@ func stringFunctionFromCharacters(invocation Invocation) Value {
 
 	inter := invocation.Interpreter
 
-	common.UseMemory(inter,
-		common.MemoryUsage{
-			Kind:   common.MemoryKindStringValue,
-			Amount: 1,
-		},
-	)
+	// NewStringMemoryUsage already accounts for empty string.
+	common.UseMemory(inter, common.NewStringMemoryUsage(0))
 	var builder strings.Builder
 
 	argument.Iterate(inter, func(element Value) (resume bool) {
 		character := element.(CharacterValue)
-		common.UseMemory(inter,
-			common.MemoryUsage{
-				Kind:   common.MemoryKindStringValue,
-				Amount: uint64(len(character)),
-			},
-		)
+		// -1 to offset the double counting for empty string.
+		common.UseMemory(inter, common.NewStringMemoryUsage(len(character)-1))
 		builder.WriteString(string(character))
 
 		return true
@@ -135,24 +127,16 @@ func stringFunctionJoin(invocation Invocation) Value {
 		separator = StringTypeJoinDefaultSeparator
 	}
 
-	common.UseMemory(inter,
-		common.MemoryUsage{
-			Kind:   common.MemoryKindStringValue,
-			Amount: 1,
-		},
-	)
+	// NewStringMemoryUsage already accounts for empty string.
+	common.UseMemory(inter, common.NewStringMemoryUsage(0))
 	var builder strings.Builder
 	first := true
 
 	argument.Iterate(inter, func(element Value) (resume bool) {
 		// Add separator
 		if !first {
-			common.UseMemory(inter,
-				common.MemoryUsage{
-					Kind:   common.MemoryKindStringValue,
-					Amount: uint64(len(separator.Str)),
-				},
-			)
+			// -1 to offset the double counting for empty string.
+			common.UseMemory(inter, common.NewStringMemoryUsage(len(separator.Str)))
 			builder.WriteString(separator.Str)
 		}
 		first = false
@@ -162,12 +146,8 @@ func stringFunctionJoin(invocation Invocation) Value {
 			panic(errors.NewUnreachableError())
 		}
 
-		common.UseMemory(inter,
-			common.MemoryUsage{
-				Kind:   common.MemoryKindStringValue,
-				Amount: uint64(len(str.Str)),
-			},
-		)
+		// -1 to offset the double counting for empty string.
+		common.UseMemory(inter, common.NewStringMemoryUsage(len(str.Str)))
 		builder.WriteString(str.Str)
 
 		return true
--- a/runtime/interpreter/value_string.go
+++ b/runtime/interpreter/value_string.go
@@ -108,6 +108,8 @@ func stringFunctionFromCharacters(invocation Invocation) Value {
 	return NewUnmeteredStringValue(builder.String())
 }
 
+var StringTypeJoinDefaultSeparator = NewUnmeteredStringValue(",")
+
 func stringFunctionJoin(invocation Invocation) Value {
 	argument, ok := invocation.Arguments[0].(*ArrayValue)
 	if !ok {
@@ -116,6 +118,13 @@ func stringFunctionJoin(invocation Invocation) Value {
 
 	inter := invocation.Interpreter
 
+	switch argument.Count() {
+	case 0:
+		return EmptyString
+	case 1:
+		return argument.Get(inter, invocation.LocationRange, 0)
+	}
+
 	var separator *StringValue
 	if len(invocation.Arguments) > 1 {
 		separator, ok = invocation.Arguments[1].(*StringValue)
@@ -123,15 +132,7 @@ func stringFunctionJoin(invocation Invocation) Value {
 			panic(errors.NewUnreachableError())
 		}
 	} else {
-		separator = NewStringValue(
-			inter,
-			common.MemoryUsage{
-				Kind:   common.MemoryKindStringValue,
-				Amount: 1,
-			},
-			func() string {
-				return ","
-			})
+		separator = StringTypeJoinDefaultSeparator
 	}
 
 	common.UseMemory(inter,
--- a/runtime/interpreter/value_string.go
+++ b/runtime/interpreter/value_string.go
@@ -106,8 +106,6 @@ func stringFunctionFromCharacters(invocation Invocation) Value {
 	return NewUnmeteredStringValue(builder.String())
 }
 
-var StringTypeJoinDefaultSeparator = NewUnmeteredStringValue(",")
-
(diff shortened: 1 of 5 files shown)


SupunS (approved):
Nice!

runtime/interpreter/value_string.go
```diff
@@ -108,6 +108,73 @@ func stringFunctionFromCharacters(invocation Invocation) Value {
...
+	} else {
+		separator = NewStringValue(
+			inter,
+			common.MemoryUsage{
+				Kind:   common.MemoryKindStringValue,
+				Amount: 1,
+			},
+			func() string {
+				return ","
+			})
```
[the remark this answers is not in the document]
darkdrag00nv2: Moved outside and removed metering.

runtime/interpreter/value_string.go
```diff
@@ -108,6 +108,73 @@ func stringFunctionFromCharacters(invocation Invocation) Value {
...
+	common.UseMemory(inter,
+		common.MemoryUsage{
+			Kind:   common.MemoryKindStringValue,
+			Amount: 1,
+		},
+	)
+	var builder strings.Builder
+	first := true
+
+	argument.Iterate(inter, func(element Value) (resume bool) {
```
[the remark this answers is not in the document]
darkdrag00nv2: Good suggestion, done.

SupunS (approved):

runtime/interpreter/value_string.go
```diff
@@ -108,6 +108,74 @@ func stringFunctionFromCharacters(invocation Invocation) Value {
...
+		// Add separator
+		if !first {
+			common.UseMemory(inter,
+				common.MemoryUsage{
+					Kind:   common.MemoryKindStringValue,
+					Amount: uint64(len(separator.Str)),
+				},
```
turbolent, line 154: Use `NewStringMemoryUsage` instead. Also please use it in `stringFunctionFromCharacters` and `TypeValue.GetMember` while at it.

runtime/interpreter/value_string.go
```diff
@@ -108,6 +108,74 @@ func stringFunctionFromCharacters(invocation Invocation) Value {
...
+		str, ok := element.(*StringValue)
+		if !ok {
+			panic(errors.NewUnreachableError())
+		}
+
+		common.UseMemory(inter,
+			common.MemoryUsage{
+				Kind:   common.MemoryKindStringValue,
+				Amount: uint64(len(str.Str)),
+			},
```
turbolent, line 169: Same as above, use `NewStringMemoryUsage`

runtime/sema/string_type.go
```diff
@@ -37,6 +37,11 @@ const StringTypeFromCharactersFunctionDocString = `
 Returns a string from the given array of characters
 `
 
+const StringTypeJoinFunctionName = "join"
+const StringTypeJoinFunctionDocString = `
+Returns a string after joining the array of strings with the given separator.
```
turbolent, line 42: Also document the default separator if none is given

runtime/sema/string_type.go
```diff
@@ -298,3 +310,23 @@ var StringTypeFromCharactersFunctionType = &FunctionType{
...
+	Parameters: []Parameter{
+		{
+			Label:      ArgumentLabelNotRequired,
+			Identifier: "strs",
+			TypeAnnotation: NewTypeAnnotation(&VariableSizedType{
+				Type: StringType,
+			}),
+		},
+		{
+			Label:          ArgumentLabelNotRequired,
```
turbolent, line 324: Require an argument label, to make the purpose clear:

```suggestion
```

turbolent (commented):
Nice!

turbolent:
@SupunS Contributors can't add labels themselves, please add one for the release notes

runtime/interpreter/value_string.go
darkdrag00nv2: Done, do note that in some places in `stringFunctionJoin` and `stringFunctionFromCharacters`, I had to do a `-1` to account for double counting of empty string.

runtime/interpreter/value_string.go
```diff
@@ -108,6 +100,62 @@ func stringFunctionFromCharacters(invocation Invocation) Value {
...
+	// NewStringMemoryUsage already accounts for empty string.
+	common.UseMemory(inter, common.NewStringMemoryUsage(0))
+	var builder strings.Builder
+	first := true
+
+	argument.Iterate(inter, func(element Value) (resume bool) {
+		// Add separator
+		if !first {
+			// -1 to offset the double counting for empty string.
+			common.UseMemory(inter, common.NewStringMemoryUsage(len(separator.Str)))
```
SupunS, line 139: comment mentions a `-1`, but the implementation doesn't actually decrement.
Same for line#150

runtime/interpreter/value_string.go
```diff
@@ -108,6 +100,62 @@ func stringFunctionFromCharacters(invocation Invocation) Value {
 	return NewUnmeteredStringValue(builder.String())
 }
 
+var StringTypeJoinDefaultSeparator = NewUnmeteredStringValue(",")
+
+func stringFunctionJoin(invocation Invocation) Value {
+	argument, ok := invocation.Arguments[0].(*ArrayValue)
```
SupunS, line 106: maybe rename this variable `argument` to a more intuitive one like `stringArray` or  even just `array`

runtime/interpreter/value_string.go
```diff
@@ -84,22 +84,14 @@ func stringFunctionFromCharacters(invocation Invocation) Value {
...
 	argument.Iterate(inter, func(element Value) (resume bool) {
 		character := element.(CharacterValue)
-		common.UseMemory(inter,
-			common.MemoryUsage{
-				Kind:   common.MemoryKindStringValue,
-				Amount: uint64(len(character)),
-			},
-		)
+		// -1 to offset the double counting for empty string.
+		common.UseMemory(inter, common.NewStringMemoryUsage(len(character)-1))
```
SupunS, line 94: It looks a bit odd to decrement here (though I understand the reason). Maybe that's probably why the original code did it without the constructor. Maybe keep it in the original way, but adding a comment explaining why it is done this way might be better?
cc: @turbolent

darkdrag00nv2: Personally, I prefer how it was written before. Processing why the extra `-1` is used could take a few seconds and can be avoided in this case. I've updated the code and added the clarification comment.

runtime/interpreter/value_string.go
darkdrag00nv2: Done.

runtime/interpreter/value_string.go
darkdrag00nv2: Reverted back to manual creation and added a comment. The `-1` seems odd.

runtime/interpreter/value_string.go
turbolent: Yeah, whatever is clearest is best 👍

SupunS (approved):

runtime/interpreter/value_string.go
```diff
@@ -108,6 +106,74 @@ func stringFunctionFromCharacters(invocation Invocation) Value {
 	return NewUnmeteredStringValue(builder.String())
 }
 
+var StringTypeJoinDefaultSeparator = NewUnmeteredStringValue(",")
```
turbolent, line 109: It's probably better to have the separator default to an empty string
```suggestion
var StringTypeJoinDefaultSeparator = EmptyString
```

runtime/sema/string_type.go
```diff
@@ -37,6 +37,12 @@ const StringTypeFromCharactersFunctionDocString = `
 Returns a string from the given array of characters
 `
 
+const StringTypeJoinFunctionName = "join"
+const StringTypeJoinFunctionDocString = `
+Returns a string after joining the array of strings with the optional separator if provided.
+Uses ',' as the default separator when not provided.
```
turbolent, line 43:
```suggestion
Uses an empty string as the default separator when not provided.
```

turbolent (commented):
Looks good!

Regarding the default: Every language's standard library seems to have a different default, so maybe choosing the empty string as the "lowest common denominator" is best, as it avoids surprises based on wrong expectations.

Maybe we should just require it, i.e. not have any default? 

For example:
- Swift: `""` (empty string)
- JavaScript: `","`
- Kotlin: `", "` (additional space)
- Python, Rust, Go: no default

darkdrag00nv2:
I think it is probably best to just make it required. Made it required in b92d5e1a5.

If we want to have a default, then I would prefer having the same default in `split` and `join`. Empty string won't work in `split`, so I don't prefer it.

runtime/interpreter/value_string.go
darkdrag00nv2: Not applicable.

runtime/sema/string_type.go
darkdrag00nv2: Not applicable.

dsainati1 (approved):
If I understand correctly, this function doesn't mutate the input array, so making a note to myself to add the appropriate `view` annotation as per https://github.com/onflow/cadence/issues/2466

darkdrag00nv2:
@SupunS @turbolent Should we merge this?

runtime/sema/string_type.go
```diff
@@ -298,3 +310,21 @@ var StringTypeFromCharactersFunctionType = &FunctionType{
...
+				Type: StringType,
+			}),
+		},
+		{
+			Identifier:     "separator",
+			TypeAnnotation: NewTypeAnnotation(StringType),
+		},
+	},
+	ReturnTypeAnnotation: NewTypeAnnotation(StringType),
+	Arity:                &Arity{Min: 2, Max: 2},
```
turbolent, line 329:
```suggestion
```

turbolent (approved):
Nice!

Issue #2759, opened 2023-09-01
Better error message in stable-cadence if using restricted types


bjartek:
### Issue to be solved

Using restricted types was valid before stable cadence, so an example here is

```cadence
 transaction(foo: AnyStruct{MetadataViews.File}) {
}
```

When upgrading to stable-cadence this is not allowed and we get the error message 

```
error: expected comma or end of parameter list, got '{'
  |
3 |         transaction(foo: AnyStruct{MetadataViews.File}) {
```

### Suggested Solution

Make the error message better and suggestion a fix.

PR #2764, opened 2023-09-05
Better error messages for use of old restricted types

dsainati1:
Closes https://github.com/onflow/cadence/issues/2759

Restricted types have been removed, but parsing old code that uses them results in a confusing error. Add special handling for this case for now. 

______

### DIFF
--- a/runtime/parser/type.go
+++ b/runtime/parser/type.go
@@ -481,7 +481,7 @@ func defineIntersectionOrDictionaryType() {
 
 	// While restricted types have been removed from Cadence, during the first few months of the
 	// migration period, leave a special error in place to help developers
-
+	// TODO: remove this after Stable Cadence migration period is finished
 	setTypeMetaLeftDenotation(
 		lexer.TokenBraceOpen,
 		func(p *parser, rightBindingPower int, left ast.Type) (result ast.Type, err error, done bool) {
--- a/runtime/parser/type.go
+++ b/runtime/parser/type.go
@@ -503,8 +503,7 @@ func defineIntersectionOrDictionaryType() {
 				return left, nil, true
 			}
 
-			// It was determined that a restricted type is parsed, so error
-			return nil, p.syntaxError("restricted types have been removed; replace with an intersection type"), true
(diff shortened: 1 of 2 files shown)


runtime/parser/type.go
```diff
@@ -467,6 +478,35 @@ func defineIntersectionOrDictionaryType() {
...
+			// The buffered tokens are replayed to allow them to be re-parsed.
+			if p.current.Is(lexer.TokenSpace) {
+				p.current = current
+				p.tokens.Revert(cursor)
+
+				return left, nil, true
+			}
+
+			// It was determined that a restricted type is parsed, so error
+			return nil, p.syntaxError("restricted types have been removed; replace with an intersection type"), true
```
turbolent, line 507: Depending on the program, there are two possible resolutions:
- Restricted type is `AnyStruct`/`AnyResource` (e.g. `AnyStruct{I1, ..., In}`): replace with equivalent intersection type (`{I1, ..., In}`)
- Restricted type is concrete type (e.g. `T{I1, ..., In}`): remove restrictions / leave concrete type

```suggestion
			return nil, p.syntaxError("restricted types have been removed; consider using an intersection type or removing the restrictions"), true
```

turbolent (commented):
Nice!

It might make sense to put this behind a feature flag and only enable it in the preview builds (e.g. "`StableCadenceMigrationAssistanceEnabled`")

dsainati1:
Currently it doesn't seem possible to selectively enable or disable parsing rules based on config options (i.e. `setTypeMetaLeftDenotation` and related functions accept no `parser` argument). Did you have a different way in mind of doing this?

runtime/parser/type.go
```diff
@@ -467,6 +478,34 @@ func defineIntersectionOrDictionaryType() {
 			}
 		},
 	)
+
+	// While restricted types have been removed from Cadence, during the first few months of the
+	// migration period, leave a special error in place to help developers
+	// TODO: remove this after Stable Cadence migration period is finished
```
SupunS, line 484: While at it, maybe also create an issue to revert this, so we don't forget later

SupunS (approved):

turbolent (approved):

PR #2769, opened 2023-09-07
Add `tryUpdate` method to `Account.Contracts`

### LINKED ISSUE #2700: Add contract update method that doesn't panic
### Issue to be solved

Currently, there is no way to attempt to deploy or update a contract without a failure resulting in a `panic` as in [`Contracts.update__experimental()`](https://developers.flow.com/cadence/language/accounts). This approach makes sense for the most part, but I found it can create difficulties when attempting to deploy/update multiple contracts as is the case in my exploration of [delegated contract updates](https://github.com/sisyphusSmiling/contract-updater) via onchain resources.

### Suggested Solution

As discussed in [this linked Discord thread](https://discord.com/channels/613813861610684416/1108479699732152503/1137138165140967628), we might add some method(s) `tryAdd()/tryUpdate()` that adds/updates a contract, returning some structured data reflecting the deployment state. An example might be 

```cadence
pub enum ErrorType: UInt8 {
  pub case CONTRACT_NOT_FOUND
  pub case MULTIPLE_CONTRACTS_DECLARED
  pub case MISMATCHED_NAME
  pub case UNDEFINED
}

pub struct DeploymentError {
  pub let errorType: ErrorType
  pub let errorMessage: String
}

pub struct DeploymentResult {
  pub let success: Bool
  pub let errorMessage: DeploymentError?
}

pub fun tryUpdate(name: String, code: [UInt8]): DeploymentResult
```

This is submitted not to be prescriptive, but as an example of what the structured data value might look like. @bjartek might have some thoughts here as well.

SupunS:
Closes #2700

## Description

Adds a `tryUpdate` method that gracefully handles deployment errors.

```cadence
struct Account {
    ...
    struct Contracts {
        ...
        access(Contracts | UpdateContract)
        fun tryUpdate(name: String, code: [UInt8]): DeploymentResult
    }
}

access(all) struct DeploymentResult {
    access(all) let deployedContract: DeployedContract?
}
```

If the update was unsuccessful, `DeploymentResult.deployedContract` will be `nil`.

This `tryUpdate` method handles any user error (syntax errors, semantic errors, etc.) gracefully. Any internal error would still cause the script/transaction to terminate.
______

### DIFF
--- a/runtime/contract_test.go
+++ b/runtime/contract_test.go
@@ -26,8 +26,11 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/onflow/cadence/runtime/errors"
 	"github.com/onflow/cadence/runtime/interpreter"
+	"github.com/onflow/cadence/runtime/sema"
 	"github.com/onflow/cadence/runtime/stdlib"
+	"github.com/onflow/cadence/runtime/tests/checker"
 	. "github.com/onflow/cadence/runtime/tests/utils"
 
 	"github.com/onflow/cadence"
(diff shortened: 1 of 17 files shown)


runtime/stdlib/account.go
```diff
@@ -1369,250 +1376,303 @@ func newAccountContractsChangeFunction(
 		gauge,
 		functionType,
 		func(invocation interpreter.Invocation) interpreter.Value {
+			return changeAccountContracts(invocation, handler, addressValue, isUpdate)
```
SupunS, line 1379: The code starting from here was extracted out to a function, so that it can be used in both `update` function as well as `tryUpdate` function.

It was moved as-is, and no change was made to the logic.

-- Refactor starts here  --

runtime/stdlib/account.go
```diff
@@ -1369,250 +1376,303 @@ func newAccountContractsChangeFunction(
...
+		locationRange,
+	)
+
+	return interpreter.NewDeployedContractValue(
+		inter,
+		addressValue,
+		nameValue,
+		newCodeValue,
+	)
+}
```
SupunS, line 1634: -- Refactor ends here --

bluesign:
should we unset `TemporarilyRecordCode` (set original code) on failure here ?

SupunS:
@bluesign That's a good point! 
The code that is being set in `TemporarilyRecordCode` is only used for error pretty-printing/debugging. Since errors are not reported for the deploying contract during the transaction, it has no impact for now. There is also a plan to remove codes and programs from the errors, so this will go away soon.

runtime/sema/account.cdc
```diff
@@ -205,6 +205,26 @@ struct Account {
...
+        ///
+        /// Fails if no contract/contract interface with the given name exists in the account,
+        /// if the given code does not declare exactly one contract or contract interface,
+        /// or if the given name does not match the name of the contract/contract interface declaration in the code.
+        ///
+        /// Returns the deployment result.
+        /// Result would contain the deployed contract for the updated contract, if the update was successfull.
+        /// Otherwise, the deployed contract would be nil.
+        access(Contracts | UpdateContract)
+        fun tryUpdate(name: String, code: [UInt8]): DeploymentResult
```
dsainati1, line 226: Why not just have `DeployedContract?` be the return type? What is the reason for a separate container struct here?

SupunS: The plan is to also include some info about the error, so developers can decide what to do about it. Would eventually be something like:
```cadence
access(all) struct DeploymentResult {
    access(all) let deployedContract: DeployedContract?
    access(all) let deploymentError: DeploymentError?
}
```
(would have been better if it was something like `DeployedContract|DeploymentError` but we don't have union types)

But as discussed in https://github.com/onflow/cadence/issues/2700#issuecomment-1669992213, there is some work needed on the error/error-messages to make it happen, so it is being excluded for now.

dsainati1 (approved):

turbolent (dismissed):
Nice!

runtime/tests/checker/account_test.go
```diff
@@ -1083,6 +1083,33 @@ func TestCheckAccountContractsUpdate(t *testing.T) {
         `)
 		require.NoError(t, err)
 	})
+
```
turbolent, line 1086: Can you please also add (a) test case(s) for `DeploymentResult` itself, i.e. access its field and "assert" the returned type matches?

runtime/contract_test.go
```diff
@@ -1014,3 +1017,289 @@ func TestRuntimeContractInterfaceConditionEventEmission(t *testing.T) {
...
+		t.Parallel()
+
+		rt := newTestInterpreterRuntime()
+
+		deployTx := DeploymentTransaction("Foo", []byte(`access(all) contract Foo {}`))
+
+		updateTx := []byte(`
+			transaction {
+				prepare(signer: auth(UpdateContract) &Account) {
+					signer.contracts.tryUpdate(
```
turbolent, line 1063: Also test usage of the result of the function, i.e. the `DeploymentResult` and its field `.deployedContract`, and in turn its fields (e.g. assert its code matches input code)

turbolent (commented):
Looks good! 

Good idea of just generating a `CompositeType`.

Just one concern (optional wrapping) and some more tests would be good to cover all cases.

turbolent (approved):
👌

PR #2771, opened 2023-09-08
Add view annotations to newly added array functions

dsainati1:
Continue https://github.com/onflow/cadence/issues/2466

These functions can be marked as `view`. 

Note that this marks `filter` as `view` as well by restricting it's inputs to `view` functions. This makes sense to me personally considering these are intended to be boolean predicates, which should not have side effects. 
______

### DIFF
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -2492,6 +2492,7 @@ func ArrayFilterFunctionType(memoryGauge common.MemoryGauge, elementType Type) *
 			},
 		},
 		ReturnTypeAnnotation: NewTypeAnnotation(BoolType),
+		Purity:               FunctionPurityView,
 	}
 
 	return &FunctionType{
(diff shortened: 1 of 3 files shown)


runtime/sema/type.go
```diff
@@ -2503,6 +2505,7 @@ func ArrayFilterFunctionType(memoryGauge common.MemoryGauge, elementType Type) *
 			},
 		},
 		ReturnTypeAnnotation: NewTypeAnnotation(NewVariableSizedType(memoryGauge, elementType)),
+		Purity:               FunctionPurityView,
```
turbolent, line 2508: This also force the predicate to be view. Not sure if we want to require it, given it isn't necessary? Not a big deal, just wanted to flag it. In the future we could maybe have two variants of the function (until we have a means of abstracting over this).

turbolent (approved):
Nice!

runtime/sema/type.go
dsainati1: Yeah, as mentioned in the PR summary, I think this is the right call in this case because predicates are generally thought of as non-mutating operations. But I don't feel too strongly about it. We could relax it later if people complain.

SupunS (approved):

runtime/sema/type.go
turbolent: Sorry, I missed that! Sounds good 👍

PR #2777, opened 2023-09-11
Cache normalized string

turbolent:
Work towards #2752 

## Description

String functions, like checking for equality, already operate on the normalized form of the string.
However, the results are currently discarded. 

As strings are immutable, we can cache the normalized form and re-use it.

As the hash value is changed, this will need a migration. Given that we can perform breaking changes in the Stable Cadence release, we should use this opportunity. 

______

### DIFF
--- a/runtime/convertValues.go
+++ b/runtime/convertValues.go
@@ -88,11 +88,12 @@ func exportValueWithInterpreter(
 	case interpreter.BoolValue:
 		return cadence.NewMeteredBool(inter, bool(v)), nil
 	case *interpreter.StringValue:
+		str := v.Str(inter)
 		return cadence.NewMeteredString(
 			inter,
-			common.NewCadenceStringMemoryUsage(len(v.Str)),
+			common.NewCadenceStringMemoryUsage(len(str)),
 			func() string {
-				return v.Str
+				return str
 			},
 		)
 	case interpreter.CharacterValue:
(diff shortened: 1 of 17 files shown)


dsainati1 (approved):

darkdrag00nv2 (approved):
Nice! Thanks

SupunS (approved):

turbolent:
There's a bit of a problem with the lazy approach:

Cadence currently stores Strings inline or as a separate slab, if too large. This is possible because strings are immutable. However, in this PR we're lazily normalizing the string, effectively "mutating" it (replacing it, with a potentially smaller or larger form).

In the `ByteSize` and `Encode` we should ideally use the normalized form, instead of the unnormalized form. If we don't, the calculated byte size might be too long or too short. The normalization allocates, which we need to meter. However, there's no metering available in either of the two functions.

Discussing this in https://discord.com/channels/613813861610684416/1108479699732152503/1151271319548211210

Ideas:
- Should/can we pass in a metering gauge to `ByteSize` and `Encode`? This would need to be done on the atree side unfortunately, and the "threading through" of the metering gauge would be quite tedious.
- Alternatively, we could switch from lazily normalizing, potentially during encoding, to eagerly normalizing during creation time.

I opened an implementation of the later as #2781

turbolent:
Closing in favor of #2781

PR #2781, opened 2023-09-12
Eagerly normalize String and Character values at construction time

turbolent:
Work towards #2752
Alternative to #2777

## Description

Instead of lazily normalizing, if needed, always eagerly normalize at construction time.

______

### DIFF
--- a/runtime/interpreter/decode.go
+++ b/runtime/interpreter/decode.go
@@ -362,7 +362,9 @@ func (d StorableDecoder) decodeCharacter() (CharacterValue, error) {
 		)
 	}
 
-	// NOTE: already metered by decodeCharacter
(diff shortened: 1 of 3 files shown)


bluesign:
I like this version more tbh. Eventually this guarantees what is stored will be the same all the time.

dsainati1 (approved):

SupunS (approved):

turbolent:
I also like this version better: Though it has a performance overhead, it's predictable and its implementation is much simpler compared to  #2777 (which was also incomplete and still would need more work in Cadence and atree to meter the lazy allocations)

fxamacker:
Nice!

Issue #2726, opened 2023-08-16
Replace `Test.TestAccount` with `Account`


turbolent:
### Issue to be solved

#2648 introduces the built-in type `Account`. The `Test` contract define a `Account` type, but shadowing is not allowed. Therefore, the type was renamed to `TestAccount`.

`TestAccount` is just a holder for the address of an account which was previously created using `createAccount`.

It might be nice to replace `TestAccount` with `Account` (as `&Account`).

### Suggested Solution

Replace `Test.TestAccount` with `Account`

turbolent:
cc @m-Peter

PR #2783, opened 2023-09-14
Refine API of the testing framework

m-Peter:
## Description

- Remove the `Test.newEmulatorBlockchain()` method, since `EmulatorBackend` is now a singleton
- Remove the `Test.useConfiguration()` method, since the mappings will be provided by `flow.json` config file
- Introduce the `Test.getAccount(_ address: Address): Test.Account` method, to retrieve an account from the `EmulatorBackend`
- Change `Test.deployContract` to require only `name` / `path` & `arguments`

These changes will allow testing providers to unify unit & integration tests.

An example of unit testing:

```cadence
import Test
import FooContract from "../contracts/FooContract.cdc"

pub fun setup() {
    let err = Test.deployContract(
        name: "FooContract",
        path: "../contracts/FooContract.cdc",
        arguments: []
    )

    Test.expect(err, Test.beNil())
}

pub fun testGetIntegerTrait() {
    // Arrange
    let testInputs: {Int: String} = {
        -1: "Negative",
        0: "Zero",
        9: "Small"
    }

    for input in testInputs.keys {
        // Act
        let result = FooContract.getIntegerTrait(input)

        // Assert
        Test.assertEqual(result, testInputs[input]!)
    }
}

pub fun testAddSpecialNumber() {
    // Act
    FooContract.addSpecialNumber(78557, "Sierpinski")

    // Assert
    Test.assertEqual("Sierpinski", FooContract.getIntegerTrait(78557))
}
```

An example of integration testing:

```cadence
import Test

pub let account = Test.getAccount(0x0000000000000011)

pub fun setup() {
    let err = Test.deployContract(
        name: "FooContract",
        path: "../contracts/FooContract.cdc",
        arguments: []
    )

    Test.expect(err, Test.beNil())
}

pub fun testGetIntegerTrait() {
    let script = Test.readFile("../scripts/get_integer_traits.cdc")
    let result = Test.executeScript(script, [])

    Test.expect(result, Test.beSucceeded())
    Test.assert(result.returnValue! as! Bool)
}

pub fun testAddSpecialNumber() {
    let code = Test.readFile("../transactions/add_special_number.cdc")
    let tx = Test.Transaction(
        code: code,
        authorizers: [account.address],
        signers: [account],
        arguments: [78557, "Sierpinski"]
    )

    let result = Test.executeTransaction(tx)
    Test.expect(result, Test.beSucceeded())
}
```

All contracts are to be deployed on a blockchain.
In unit tests, developers can import the deployed contracts and directly access fields and call methods.
In integration tests, developers can still import the deployed contracts, to make use of the defined nested types, and interact with the contract by re-using their scripts & transactions.

To make this possible, developers will need to make use of `flow.json` config file, e.g:
```json
"FooContract": {
      "source": "contracts/FooContract.cdc",
      "aliases": {
        "testing": "0x0000000000000007"
      }
    }
```

The account where the contracts should be deployed, is set above.
______

### DIFF
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -4262,7 +4262,11 @@ func (t *CompositeType) GetMembers() map[string]MemberResolver {
 }
 
 func (t *CompositeType) initializeMemberResolvers() {
-	t.memberResolversOnce.Do(func() {
+	t.memberResolversOnce.Do(t.initializerMemberResolversFunc())
+}
+
+func (t *CompositeType) initializerMemberResolversFunc() func() {
+	return func() {
 		memberResolvers := MembersMapAsResolvers(t.Members)
 
 		// Check conformances.
@@ -4281,7 +4285,13 @@ func (t *CompositeType) initializeMemberResolvers() {
 			})
 
 		t.memberResolvers = withBuiltinMembers(t, memberResolvers)
-	})
+	}
+}
+
+func (t *CompositeType) ResolveMembers() {
+	if t.Members.Len() != len(t.GetMembers()) {
+		t.initializerMemberResolversFunc()()
+	}
 }
 
 func (t *CompositeType) FieldPosition(name string, declaration ast.CompositeLikeDeclaration) ast.Position {
--- a/runtime/stdlib/contracts/test.cdc
+++ b/runtime/stdlib/contracts/test.cdc
@@ -27,6 +27,12 @@ pub contract Test {
             return self.backend.createAccount()
         }
 
+        /// Returns the account for the given address.
+        ///
+        pub fun getAccount(_ address: Address): Account {
+            return self.backend.getAccount(address)
+        }
+
         /// Add a transaction to the current block.
         ///
         pub fun addTransaction(_ tx: Transaction) {
@@ -77,14 +83,12 @@ pub contract Test {
         ///
         pub fun deployContract(
             name: String,
-            code: String,
-            account: Account,
+            path: String,
             arguments: [AnyStruct]
         ): Error? {
             return self.backend.deployContract(
                 name: name,
-                code: code,
-                account: account,
+                path: path,
                 arguments: arguments
             )
         }
@@ -297,6 +301,10 @@ pub contract Test {
         ///
         pub fun createAccount(): Account
 
+        /// Returns the account for the given address.
+        ///
+        pub fun getAccount(_ address: Address): Account
+
         /// Add a transaction to the current block.
         ///
         pub fun addTransaction(_ tx: Transaction)
@@ -315,8 +323,7 @@ pub contract Test {
         ///
         pub fun deployContract(
             name: String,
-            code: String,
-            account: Account,
+            path: String,
             arguments: [AnyStruct]
         ): Error?
 
--- a/runtime/stdlib/contracts/test.cdc
+++ b/runtime/stdlib/contracts/test.cdc
@@ -249,17 +249,6 @@ pub contract Test {
         }
     }
 
-    /// Configuration to be used by the blockchain.
-    /// Can be used to set the address mappings.
-    ///
-    pub struct Configuration {
-        pub let addresses: {String: Address}
-
-        init(addresses: {String: Address}) {
-            self.addresses = addresses
-        }
-    }
-
     /// Transaction that can be submitted and executed on the blockchain.
     ///
     pub struct Transaction {
--- a/runtime/stdlib/contracts/test.cdc
+++ b/runtime/stdlib/contracts/test.cdc
@@ -1,10 +1,10 @@
 /// Test contract is the standard library that provides testing functionality in Cadence.
 ///
-pub contract Test {
+access(all) contract Test {
 
     /// backend emulates a real network.
     ///
-    pub let backend: AnyStruct{BlockchainBackend}
+    access(self) let backend: AnyStruct{BlockchainBackend}
 
     init(backend: AnyStruct{BlockchainBackend}) {
         self.backend = backend
@@ -13,7 +13,8 @@ pub contract Test {
     /// Executes a script and returns the script return value and the status.
     /// `returnValue` field of the result will be `nil` if the script failed.
     ///
-    pub fun executeScript(_ script: String, _ arguments: [AnyStruct]): ScriptResult {
+    access(all)
+    fun executeScript(_ script: String, _ arguments: [AnyStruct]): ScriptResult {
         return self.backend.executeScript(script, arguments)
     }
 
@@ -21,39 +22,45 @@ pub contract Test {
     /// The transaction is paid by the service account.
     /// The returned account can be used to sign and authorize transactions.
     ///
-    pub fun createAccount(): Account {
+    access(all)
+    fun createAccount(): Account {
         return self.backend.createAccount()
     }
 
     /// Returns the account for the given address.
     ///
-    pub fun getAccount(_ address: Address): Account {
+    access(all)
+    fun getAccount(_ address: Address): Account {
         return self.backend.getAccount(address)
     }
 
     /// Add a transaction to the current block.
     ///
-    pub fun addTransaction(_ tx: Transaction) {
+    access(all)
+    fun addTransaction(_ tx: Transaction) {
         self.backend.addTransaction(tx)
     }
 
     /// Executes the next transaction in the block, if any.
     /// Returns the result of the transaction, or nil if no transaction was scheduled.
     ///
-    pub fun executeNextTransaction(): TransactionResult? {
+    access(all)
+    fun executeNextTransaction(): TransactionResult? {
         return self.backend.executeNextTransaction()
     }
 
     /// Commit the current block.
     /// Committing will fail if there are un-executed transactions in the block.
     ///
-    pub fun commitBlock() {
+    access(all)
+    fun commitBlock() {
         self.backend.commitBlock()
     }
 
     /// Executes a given transaction and commit the current block.
     ///
-    pub fun executeTransaction(_ tx: Transaction): TransactionResult {
+    access(all)
+    fun executeTransaction(_ tx: Transaction): TransactionResult {
         self.addTransaction(tx)
         let txResult = self.executeNextTransaction()!
         self.commitBlock()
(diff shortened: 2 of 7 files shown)


SupunS:
Great work @m-Peter! 

I haven't looked into the code in detail yet, but I have one suggestion on the API: Can we completely hide the "blockchain" from the users?
For e.g:
```cadence
import Test 

 // Every function call to `Test.xyz` will use the singleton blockchain instance 

pub let account = Test.getAccount(0x0000000000000011)

pub fun setup() {
    let err = Test.deployContract(     // e.g: Will deploy to the singleton blockchain instance 
        name: "FooContract",
        path: "../contracts/FooContract.cdc",
        arguments: []
    )

    Test.expect(err, Test.beNil())
}

pub fun testGetIntegerTrait() {
    let script = Test.readFile("../scripts/get_integer_traits.cdc")
    let result = Test.executeScript(script, [])
    ...
}

pub fun testAddSpecialNumber() {
    ...
    let tx = Test.Transaction(...)
    let result = Test.executeTransaction(tx)
    ...
}
```

m-Peter:
@SupunS I can try to see how much effort that would require :+1: . In fact, I first went for this approach, since you had already mentioned this API, but fell into some blockers. The `Test` contract has both a `.cdc` source file, which is first checked, and then enriched with natively implemented functions. I wanted to do something like:

```cadence
pub contract Test {
    pub let backend: AnyStruct{BlockchainBackend}

    init(backend: AnyStruct{BlockchainBackend}) {
        self.backend = backend
    }

    pub fun addTransaction(_ tx: Transaction) {
        self.backend.addTransaction(tx)
    }
}
```
But that resulted in checking errors, so I went for the alternative :innocent: 

To be honest, to me it feels more intuitive to have the `Test` contract expose functions for assertions and other miscellaneous tasks, while the `Test.blockchain` singleton, exposes all the methods for interacting with the blockchain.

```cadence
Test.deployContract(...)
Test.getAccount(...)
Test.executeScript(...)
Test.executeTransaction(...)
```

vs

```cadence
blockchain.deployContract(...)
blockchain.getAccount(...)
blockchain.executeScript(...)
blockchain.executeTransaction(...)
```

Not a very strong opinion though :sweat_smile:

SupunS:
Yeah, but then again, since there will be one blockchain anyway, having to work with a `blockchain` field/variable is kind of redundant. Removing that would bring us even closer to unifying unit tests vs integration tests (and simplifying things with that).

Also, with this, it might be possible to replace `(Test/blockchain).getAccount()` with the `getAccount()` builtin function eventually, and solve https://github.com/onflow/cadence/issues/2726 in the process. (of course, we don't have to do it now, just pointing out that this will make it easier in future)

m-Peter:
@SupunS I have removed entirely the `Test.Blockchain` struct :raised_hands: .
During the process, I think I have come across a likely bug in `initializeMemberResolvers`.
Can you double check this commit: https://github.com/onflow/cadence/pull/2783/commits/9d26500a1acb9844bdc4bad2f664d751980b3f4d ? That caching seems to not work well for contracts that are partly written in Cadence, and partly with natively implemented functions/fields.

SupunS:
Nice! I'll have a look 👍 

Thank you for making all these changes 🙌

m-Peter:
I have also removed the `Test.useConfiguration()` method in https://github.com/onflow/cadence/pull/2783/commits/09498dd8ccfe6ab22ffc8d722776e9addb060151, since moving forward we'll be using the `flow.json` config file.

runtime/sema/type.go
```diff
@@ -4281,7 +4285,13 @@ func (t *CompositeType) initializeMemberResolvers() {
...
 
 		t.memberResolvers = withBuiltinMembers(t, memberResolvers)
-	})
+	}
+}
+
+func (t *CompositeType) ResolveMembers() {
+	if t.Members.Len() != len(t.GetMembers()) {
+		t.initializerMemberResolversFunc()()
+	}
```
turbolent, line 4294: Why is this needed? Can the code calling this new `ResolveMembers` function just call `GetMember`, which will initialize members once/lazily?

turbolent (commented):
Nice! Just wondering about the `CompositeType` member resolution/initialization changes

runtime/sema/type.go
m-Peter: Hey @turbolent, thanks for looking into this :pray: 
I am puzzled myself, because this shouldn't be needed. I have just added this commit to showcase a strange behavior.
I have changed the `Test` contract, to have an `init` function, which expects an argument, e.g:
```cadence
pub contract Test {

    /// backend emulates a real network.
    ///
    pub let backend: AnyStruct{BlockchainBackend}

    init(backend: AnyStruct{BlockchainBackend}) {
        self.backend = backend
    }
    ...
}
```
Which resulted in tests failing with: 
```shell
Checking failed:
error: value of type `Test` has no member `assertEqual`
```
The reason behind this, appears to be a mismatch between the values returned by `compositeType.GetMembers()` and `compositeType.Members`, e.g.:
```go
compositeType.GetMembers():  [createAccount executeNextTransaction eventsOfType ScriptResult executeScript deployContract loadSnapshot Error account executeTransactions executeTransaction serviceAccount events assertError Matcher Transaction getType logs reset createSnapshot Account getAccount backend not isInstance beSucceeded beFailed beNil ResultStatus TransactionResult moveTime commitBlock addTransaction]
compositeType.Members:  [isInstance getType account backend executeScript createAccount getAccount addTransaction executeNextTransaction commitBlock executeTransaction executeTransactions deployContract logs serviceAccount events eventsOfType reset moveTime createSnapshot loadSnapshot not beSucceeded beFailed beNil assertError Matcher ResultStatus TransactionResult ScriptResult Error Account Transaction assert assertEqual fail readFile expect newMatcher equal beEmpty haveElementCount contain beGreaterThan beLessThan expectFailure]
```

The `Test` contract is partially written in a Cadence file, and then enriched with functions that are implemented natively. All this functions appear in the `compositeType.Members` field, but they are not included in `compositeType.GetMembers()`. That's why I added and manually call `compositeType.ResolveMembers()`.

I am guessing this has something to do with the addition of the `init()` function, I am not sure what's exactly happening though.

runtime/stdlib/contracts/test.cdc
```diff
@@ -2,160 +2,154 @@
 ///
 pub contract Test {
 
-    /// Blockchain emulates a real network.
+    /// backend emulates a real network.
     ///
-    pub struct Blockchain {
+    pub let backend: AnyStruct{BlockchainBackend}
```
SupunS, line 7: Can this be `access(self)`?
```suggestion
    access(self) let backend: AnyStruct{BlockchainBackend}
```

runtime/stdlib/contracts/test.cdc
```diff
@@ -2,160 +2,154 @@
...
+        self.backend = backend
+    }
 
-        init(backend: AnyStruct{BlockchainBackend}) {
-            self.backend = backend
-        }
+    /// Executes a script and returns the script return value and the status.
+    /// `returnValue` field of the result will be `nil` if the script failed.
+    ///
+    pub fun executeScript(_ script: String, _ arguments: [AnyStruct]): ScriptResult {
```
SupunS, line 16: While at it, maybe also replace `pub` with `access(all)` for all these methods, so won't run into conflicts when we sync this with stable-cadence branch

runtime/stdlib/test_test.go
```diff
@@ -16,7 +16,8 @@
  * limitations under the License.
  */
 
-package stdlib
+// This is in order to avoid cyclic import errors with runtime package
+package stdlib_test
```
SupunS, line 20: 👍

runtime/sema/type.go
SupunS: Having a look on this now

SupunS: ah yeah, So what happens is:
- Test contract `.cdc` file is parsed and checked
- `Test` type gets initialized, and their members are also initialized and cached.
- However, at this point only the members written in cadence are available.
- Then the natively written members get "injected" to the `Test` type, but the member cache never gets updated.

Makes sense to re-instantiate the cache explicitly after injecting member 👍 
The only downside is, we won't be able to use these natively defined functions inside the `Test` contract `.cdc` file itself.

I think the longer-term approach is to add support for the `native` modifier in the checker as well (I think currently it is only supported in the parser). And then define those native functions also in Cadence, but with the `native` modifier.
e.g:
```cadence
struct Test {
    ...
    native assertEqual(...) {} 
}
```
The checker can declare these functions as members (only the signature/type), but the implementation will be provided natively. That way we won't need to manually inject the members to the `Test` type like we do now.

Anyway, we are still not here yet, and the current solution sounds good for now.

runtime/stdlib/contracts/test.cdc
m-Peter: Updated in https://github.com/onflow/cadence/pull/2783/commits/72bbfe1e5d4a5ed438762cfe05a6c7b1c1f731ef

runtime/stdlib/contracts/test.cdc
m-Peter: Updated in https://github.com/onflow/cadence/pull/2783/commits/72bbfe1e5d4a5ed438762cfe05a6c7b1c1f731ef

m-Peter:
I have also renamed the `TestFramework.NewEmulatorBackend` method to `TestFramework.EmulatorBackend`, since there will be only one.

runtime/sema/type.go
m-Peter: That's a great explanation @SupunS :clap:

SupunS (approved):
Looks good!

runtime/sema/type.go
turbolent: I see. Yeah, it would be great to revert this workaround and properly declare all functions upfront in the Cadence code.

As discussed yesterday, adding support for `native` functions to the checker would be great, and should be very simple: if the function is declared as `native`, require the body to be empty and ignore it.

@SupunS Could you please open issues for adding this functionality and a tech-debt issue for reverting these changes here?

turbolent: Actually, I quickly implemented checking native functions in #2821. 

It would still be nice to have an issue that switches the test contract to make use of this feature, and reverts the workaround introduced here.

SupunS: Nice! Opened https://github.com/onflow/cadence/issues/2822.
I merged this PR for now, so the downstream PRs are not blocked by the refactor.

PR #2788, opened 2023-09-14
Fix capability controller deletion

turbolent:
## Description

Properly handle deletion of capability controllers.

After deletion, the previous implementation replaced functions replaced the functions implementing the controller operations (e.g. delete, retarget, etc.) by replacing the function values.

However, those function values might have been bound.

Properly handle deletion, by keeping a `deleted` flag, keeping function values, and checking the deletion flag in member callbacks (`GetMember` and `SetMember`) and inside of the function values. 

In addition, some of the host function function types were incorrect.

Reported by @oebeling 🙏 

______

### DIFF
--- a/runtime/interpreter/value_accountcapabilitycontroller.go
+++ b/runtime/interpreter/value_accountcapabilitycontroller.go
@@ -210,7 +210,20 @@ func (v *AccountCapabilityControllerValue) ChildStorables() []atree.Storable {
 	}
 }
 
-func (v *AccountCapabilityControllerValue) GetMember(inter *Interpreter, _ LocationRange, name string) Value {
+type deletionCheckedFunctionValue struct {
+	FunctionValue
+}
+
+func (v *AccountCapabilityControllerValue) GetMember(inter *Interpreter, _ LocationRange, name string) (result Value) {
+	defer func() {
+		switch typedResult := result.(type) {
+		case deletionCheckedFunctionValue:
+			result = typedResult.FunctionValue
+		case FunctionValue:
+			panic(errors.NewUnexpectedError("functions need to check deletion. Use newHostFunctionValue"))
+		}
+	}()
+
 	// NOTE: check if controller is already deleted
 	v.checkDeleted()
 
@@ -233,7 +246,8 @@ func (v *AccountCapabilityControllerValue) GetMember(inter *Interpreter, _ Locat
 		}
 		return v.deleteFunction
 
-		// NOTE: when adding new functions, ensure checkDeleted is called!
+		// NOTE: when adding new functions, ensure checkDeleted is called,
+		// by e.g. using AccountCapabilityControllerValue.newHostFunction
 	}
 
 	return nil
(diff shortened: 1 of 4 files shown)


runtime/interpreter/value_accountcapabilitycontroller.go
```diff
@@ -278,29 +284,32 @@ func (v *AccountCapabilityControllerValue) ReferenceValue(
...
-	}
+func (v *AccountCapabilityControllerValue) newDeleteFunction(
+	inter *Interpreter,
+) *HostFunctionValue {
+	return NewHostFunctionValue(
+		inter,
+		sema.AccountCapabilityControllerTypeDeleteFunctionType,
+		func(invocation Invocation) Value {
+			// NOTE: check if controller is already deleted
+			v.checkDeleted()
```
dsainati1, line 303: It seems very fragile to require that `v.checkDeleted()` is called in every new function added to a CapabilityController like this. Is it possible instead to wrap these functions in a wrapper that checks whether the controller is deleted in `GetMember`, instead of requiring it to be handled here in each function? I am concerned about how easy it is to forget to do this by accident.

turbolent: All I can think of is a helper like e.g. `func (StorageCapabilityControllerValue) newHostFunctionValue(...)` which calls `NewHostFunctionValue` with a function that calls `checkDeleted` at the beginning. But then we have to make sure we call `newHostFunctionValue` instead of `NewHostFunctionValue`, so didn't really gain much. It would also be harder to see a difference between a normal host function construction vs one that uses the helper – currently the calls to `checkDeleted` are explicit, but at least it's easier to see when it is missing, when comparing to other function bodies.

Did you have a particular idea how we could make this less error prone? I'd love to, but don't really have any ideas that would make it much better.

dsainati1: What if we made `newHostFunctionValue` return a tuple of `Value` and some special other type (`DeletionChecked`?) that is only created by the helper. We could then have all of these other functions like `newDeleteFunction` and `newTargetFunction` return this same tuple, and then enforce in `GetMember` and `SetMember` that these functions do return this `DeletionChecked` that serves as "proof" that the special `newHostFunctionValue` was actually called. This would enforce at the type system level that this function is called, but on the other hand perhaps it is getting slightly overcomplicated?

turbolent: If we still need to ensure a helper is used (`newHostFunctionValue`), then we can maybe just check the deletion there, instead of requiring the implementation (the function passed to `newHostFunctionValue`) to perform the deletion check?

turbolent: Added a run-time based check in 4957cb11689d80c2459085ea58f28d897a536057. How does that look?

dsainati1: Nice! That makes me feel a lot better about this.

dsainati1 (approved):

SupunS (approved):

PR #2789, opened 2023-09-15
Remove support for custom destructors

dsainati1:

Reference implementation for https://github.com/onflow/flips/pull/131, and as such will not be merged unless this FLIP is approved. This does not (yet) include support for default events

______


runtime/interpreter/interpreter_expression.go
```diff
@@ -1207,8 +1207,6 @@ func (interpreter *Interpreter) VisitCreateExpression(expression *ast.CreateExpr
 func (interpreter *Interpreter) VisitDestroyExpression(expression *ast.DestroyExpression) Value {
 	value := interpreter.evalExpression(expression.Expression)
 
-	interpreter.invalidateResource(value)
```
dsainati1, line 1210: This gets moved to body of the value's `Destroy`, since nested resources are no longer explicitly destroyed with a `destroy` expression.

runtime/interpreter/value.go
```diff
@@ -16444,46 +16447,12 @@ func (v *CompositeValue) Destroy(interpreter *Interpreter, locationRange Locatio
...
-			v.forEachAttachment(interpreter, locationRange, func(attachment *CompositeValue) {
-				// an attachment's destructor may make reference to `base`, so we must set the base value
-				// for the attachment before invoking its destructor. For other functions, this happens
-				// automatically when the attachment is accessed with the access expression `v[A]`, which
-				// is a necessary pre-requisite for calling any members of the attachment. However, in
-				// the case of a destructor, this is called implicitly, and thus must have its `base`
-				// set manually
-				attachment.setBaseValue(interpreter, v, attachmentBaseAuthorization(interpreter, attachment))
-				attachment.Destroy(interpreter, locationRange)
-			})
```
dsainati1, line 16457: Attachments are considered as part of `ForEachField`, since they are represented as fields with special name prefixes. As such, we no longer need to explicitly delete them here.

runtime/nft_test.go
```diff
@@ -700,15 +696,6 @@ access(all) contract TopShot: NonFungibleToken {
                 return nil
             }
         }
-
-        // If a transaction destroys the Collection object,
-        // All the NFTs contained within are also destroyed
-        // Kind of like when Damien Lillard destroys the hopes and
-        // dreams of the entire city of Houston
```
dsainati1, line 707: lol

dsainati1 (commented):
Note that at the moment this doesn't have tests to enforce that the appropriate destructors run; we were previously doing this by setting flags in the body of the destructors, obviously this is no longer possible. However, once we implement support for default events, we can assert that the appropriate events are emitted when we expect in order to test this again.

dsainati1:
@turbolent @SupunS now that the FLIP associated with this is ready to be approved, this PR is ready for review

runtime/parser/declaration.go
```diff
@@ -1884,10 +1884,7 @@ func parseSpecialFunctionDeclaration(
 		declarationKind = common.DeclarationKindInitializer
 
 	case KeywordDestroy:
-		if purity == ast.FunctionPurityView {
-			return nil, NewSyntaxError(*purityPos, "invalid view annotation on destructor")
-		}
-		declarationKind = common.DeclarationKindDestructor
+		p.report(NewSyntaxError(identifier.Pos, "custom destructor definitions are no longer permitted"))
```
turbolent, line 1887: Maybe consider introducing a dedicated error with a suggested fix (removal of the destructor). This could be also done in a follow-up PR

runtime/resourcedictionary_test.go
```diff
@@ -278,14 +269,14 @@ func TestRuntimeResourceDictionaryValues(t *testing.T) {
...
 		[]string{
-			"3",
-			`"destroying R"`,
 			"3",
 			"4",
 		},
 		loggedMessages,
 	)
 
+	// DestructorTODO: add test for destruction event of R
```
turbolent, line 278: Are these TODOs for this PR? It would be good to keep asserting the recursive destruction is performed in the proper order

runtime/runtime_test.go
```diff
@@ -7338,111 +7313,6 @@ func TestRuntimeUserPanicToError(t *testing.T) {
 	require.Equal(t, retErr, err)
 }
 
-func TestRuntimeDestructorReentrancyPrevention(t *testing.T) {
```
turbolent, line 7341: 👌

runtime/runtime_test.go
```diff
@@ -7676,435 +7541,6 @@ func TestRuntimeInvalidatedResourceUse(t *testing.T) {
 
 }
 
-func TestRuntimeInvalidatedResourceUse2(t *testing.T) {
```
turbolent, line 7679: 👌

runtime/tests/checker/attachments_test.go
```diff
@@ -540,35 +540,12 @@ func TestCheckAttachmentWithMembers(t *testing.T) {
...
-				destroy() {
-					destroy self.x
-				}
 			}`,
 		)
 
 		require.NoError(t, err)
 	})
 
-	t.Run("resource field no destroy", func(t *testing.T) {
```
turbolent, line 552: I guess this is removed because the test above now is equivalent and tests the same?

runtime/tests/interpreter/resources_test.go
```diff
@@ -2153,196 +2133,6 @@ func TestInterpretResourceDestroyedInPreCondition(t *testing.T) {
 	require.True(t, didError)
 }
 
-func TestInterpretInvalidReentrantResourceDestruction(t *testing.T) {
```
turbolent, line 2156: 👌

turbolent (commented):
🧹 🗑️ 👌 Nice!

Looks good, just a couple questions regarding the TODOs

runtime/resourcedictionary_test.go
dsainati1: These are addressed in https://github.com/onflow/cadence/pull/2817, once support for default events is finished

turbolent: Nice! Maybe target a separate feature branch for this PR, so we don't forget about these TODOs

runtime/interpreter/interpreter_expression.go
SupunS: It might be sufficient to keep it here. In `interpreter.invalidateResource`, it only maintains a resource-to-variable mapping. Nested resources are not variables (either they are fields or values at an index), so the nested resources are anyway not tracked by this, if I'm not mistaken. i.e: It only tracks the outer-most resource.

Moving this to `Destroy` of each value might introduce an overhead.

runtime/interpreter/value.go
```diff
@@ -16577,46 +16580,13 @@ func (v *CompositeValue) Destroy(interpreter *Interpreter, locationRange Locatio
 		storageID,
 		locationRange,
 		func() {
-			// if this type has attachments, destroy all of them before invoking the destructor
-			v.forEachAttachment(interpreter, locationRange, func(attachment *CompositeValue) {
```
SupunS, line 16581: Do we not need to destroy attachments anymore?

dsainati1: Attachments are handled in the later `forEachField` iteration we do below to handle any nested resources, since they are no longer explicitly destroyed during the invocation of the destructor. Attachments are represented as fields, so iterating over all fields will include attachments.

runtime/interpreter/interpreter_expression.go
dsainati1: Reverted this change.

turbolent (approved):

SupunS (approved):

Issue #2790, opened 2023-09-15
Implement Custom Destructor Removal


j1010001:
Implementation of https://github.com/onflow/flips/pull/95 
Includes Default event emission
```[tasklist]
### Tasks
- [ ] https://github.com/onflow/flips/pull/131
- [ ] https://github.com/onflow/cadence/pull/2789
- [ ] https://github.com/onflow/cadence/pull/2799
- [ ] https://github.com/onflow/cadence/pull/2812
- [ ] https://github.com/onflow/cadence/pull/2817
```

PR #2799, opened 2023-09-20
Support for parsing default destroy events

dsainati1:

Reference implementation for https://github.com/onflow/flips/pull/131, and as such will not be merged unless this FLIP is approved. This adds parsing support for default events

______


runtime/tests/utils/utils.go
```diff
@@ -50,13 +50,12 @@ const ImportedLocation = common.StringLocation("imported")
...
-		// this prevents infinite recursion from circular references
-		deep.MaxDepth = 100
-
-		diff := deep.Equal(expected, actual)
+	// the maximum levels of a struct to recurse into
+	// this prevents infinite recursion from circular references
+	deep.MaxDepth = 100
+	diff := deep.Equal(expected, actual)
 
+	if !assert.Nil(t, diff) {
```
dsainati1, line 58: It's unclear to me why exactly these changes exercised this, but it's possible for `assert.Equal` to consider unequal two objects that `deep.Equal` says are equivalent. This results in nonsensical errors where a test fails, but the "diff" between the two objects is empty.

runtime/parser/declaration.go
```diff
@@ -861,7 +861,8 @@ func parseEventDeclaration(
 	// Skip the identifier
 	p.next()
 
-	parameterList, err := parseParameterList(p)
+	// if this is a `ResourceDestroyed` event (i.e., a default event declaration), parse default arguments
+	parameterList, err := parseParameterList(p, identifier.Identifier == ast.ResourceDestructionDefaultEventName)
```
turbolent, line 865: Maybe factor the expression out and give it a descriptive name to make the code a bit more self-explaining

```suggestion
	parseDefaultArguments := identifier.Identifier == ast.ResourceDestructionDefaultEventName
	parameterList, err := parseParameterList(p, parseDefaultArguments)
```

runtime/parser/function.go
```diff
@@ -167,11 +167,34 @@ func parseParameter(p *parser) (*ast.Parameter, error) {
...
+
+		// Skip the =
+		p.nextSemanticToken()
+
+		defaultArgument, err = parseExpression(p, lowestBindingPower)
+		if err != nil {
+			return nil, err
+		}
+
+	}
```
turbolent, line 190: Maybe add an else-case here, which still checks for an equal token and reports a helpful error, e.g. "unexpected default argument". This would improve the DX a bit over the current, generic error `"expected comma or end of parameter list, got '='"`

runtime/tests/utils/utils.go
turbolent: That's odd. Does the PR itself exercise this if the change here is not made? If not, do you have a way to reproduce it in some other way?

turbolent (commented):
Nice!

runtime/tests/utils/utils.go
turbolent: It turns out that `deep.Equal` does not report all differences (IDK why). I opened #2819 to replace it.

runtime/parser/declaration_test.go
```diff
@@ -2557,6 +2557,150 @@ func TestParseEvent(t *testing.T) {
...
+						[]ast.Declaration{
+							&ast.SpecialFunctionDeclaration{
+								FunctionDeclaration: &ast.FunctionDeclaration{
+									ParameterList: &ast.ParameterList{
+										Parameters: []*ast.Parameter{
+											{
+												TypeAnnotation: &ast.TypeAnnotation{
+													Type: &ast.NominalType{
+														Identifier: ast.Identifier{
+															Identifier: "Int",
```
turbolent, line 2580: The report from `assert.Equal` is actually correct, and `deep.Equal` is not properly reporting the difference here, and for the position info

SupunS: I though I fixed this somewhere 🤔 
Let me double check

runtime/ast/composite.go
```diff
@@ -280,6 +282,10 @@ func (d *CompositeDeclaration) ConformanceList() []*NominalType {
 	return d.Conformances
 }
 
+func (d *CompositeDeclaration) IsResourceDestructionDefaultEvent() bool {
+	return d.CompositeKind == common.CompositeKindEvent && d.Identifier.Identifier == ResourceDestructionDefaultEventName
```
SupunS, line 286:
```suggestion
	return d.CompositeKind == common.CompositeKindEvent &&
		d.Identifier.Identifier == ResourceDestructionDefaultEventName
```

SupunS (approved):

turbolent:
After #2820 is merged, the PR should fail, as it has the fix for the test assertions (#2819)

runtime/ast/composite.go
```diff
@@ -280,6 +282,11 @@ func (d *CompositeDeclaration) ConformanceList() []*NominalType {
 	return d.Conformances
 }
 
+func (d *CompositeDeclaration) IsResourceDestructionDefaultEvent() bool {
```
turbolent, line 285: This should maybe be moved to `sema`

runtime/ast/parameter.go
```diff
@@ -25,25 +25,28 @@ import (
 )
 
 type Parameter struct {
-	TypeAnnotation *TypeAnnotation
-	Label          string
-	Identifier     Identifier
-	StartPos       Position `json:"-"`
+	TypeAnnotation  *TypeAnnotation
+	DefaultArgument Expression
```
turbolent, line 29: This also needs to get included in `ParameterList.Doc()`. Maybe while at it, move the loop body, the Doc construction for `Parameter` out of the loop, and into a separate function `Parameter.Doc()`

runtime/parser/declaration_test.go
```diff
@@ -2557,6 +2557,150 @@ func TestParseEvent(t *testing.T) {
...
+	})
+
+	t.Run("non-default event with default arg", func(t *testing.T) {
+
+		t.Parallel()
+
+		_, errs := testParseDeclarations(" access(all) event Foo ( a : Int = 3)")
+
+		utils.AssertEqualWithDiff(t, []error{
+			&SyntaxError{
```
turbolent, line 2697: Can you please also add a test for parsing a function expression and function declaration which have parameter lists with default arguments, even though that is unexpected?

runtime/ast/composite.go
turbolent: Just from a syntax vs semantics point-of-view. Syntactically there is nothing special about this declaration, but semantic analysis then deems it special, based on its name.

Feel free to leave here, not a big deal.

turbolent (approved):
Nice! 👌 

Sorry for the delay, didn't realize this was still open. Please feel free to re-request reviews when I'm not responsive 🙏

runtime/parser/declaration_test.go
turbolent: @dsainati1 Did a test already exist?

dsainati1: I added them here in 15543eb9ab6665c8bcb9dd08d3d88bcb29c00897

turbolent: Ah, I missed that, thanks!

turbolent (approved):

PR #2806, opened 2023-09-22
Extend type code generator

turbolent:
Depends on #2805 

## Description

- Add support for constructors/initializers, also in the nested case (requires defining member in container type)
- Refactor tests so they are generated into separated packages, and import them, so they are actually type checked, and can be tested with the existing parsing/checking testing infrastructure
- Do not automatically add entitlements to the base type activation. This isn't done for any other type declaration, and pollutes the base type activation when the types are not generated into sema.
  We might want to solve this in a general case for all type declaration in a follow-up PR.
- Properly refer to built-in types in the `sema` package. For example, if `Int` is mentioned in the `cdc` file, generate `sema.IntType` when the type code output is not generated into `sema`.

______


runtime/sema/gen/golden_test.go
```diff
@@ -0,0 +1,95 @@
...
+          let x = Foo(bar: 1)
+        `,
+		checker.ParseAndCheckOptions{
+			Config: &sema.Config{
+				BaseValueActivation: baseValueActivation,
+			},
+		},
+	)
+	require.NoError(t, err)
+}
```
turbolent, line 71: The new package structure allows us to test the generated type declarations with the type checker

runtime/sema/gen/golden_test.go
```diff
@@ -0,0 +1,95 @@
...
+	_ "github.com/onflow/cadence/runtime/sema/gen/testdata/equatable"
+	_ "github.com/onflow/cadence/runtime/sema/gen/testdata/exportable"
+	_ "github.com/onflow/cadence/runtime/sema/gen/testdata/fields"
+	_ "github.com/onflow/cadence/runtime/sema/gen/testdata/functions"
+	_ "github.com/onflow/cadence/runtime/sema/gen/testdata/importable"
+	_ "github.com/onflow/cadence/runtime/sema/gen/testdata/member_accessible"
+	_ "github.com/onflow/cadence/runtime/sema/gen/testdata/nested"
+	_ "github.com/onflow/cadence/runtime/sema/gen/testdata/simple_resource"
+	_ "github.com/onflow/cadence/runtime/sema/gen/testdata/simple_struct"
+	_ "github.com/onflow/cadence/runtime/sema/gen/testdata/storable"
```
turbolent, line 44: Import the "generated" code from all test cases, so it is type checked.

This ensures the type code generator does not generate code that does not actually work.

runtime/sema/gen/main.go
```diff
@@ -323,8 +326,106 @@ func (g *generator) declarationDocString(decl ast.Declaration) dst.Expr {
 	return renderDocString(docString)
 }
 
-func (*generator) VisitSpecialFunctionDeclaration(_ *ast.SpecialFunctionDeclaration) struct{} {
-	panic("special function declarations are not supported")
+func (g *generator) VisitSpecialFunctionDeclaration(decl *ast.SpecialFunctionDeclaration) (_ struct{}) {
```
turbolent, line 329: Main contribution: add support for initializers

runtime/sema/gen/main.go
```diff
@@ -527,65 +630,93 @@ func (g *generator) VisitCompositeDeclaration(decl *ast.CompositeDeclaration) (_
...
-			members := membersExpr(typeDecl.fullTypeName, tyVarName, memberDeclarations)
+			members := membersExpr(
+				fullTypeName,
+				tyVarName,
+				memberDeclarations,
+			)
 
 			const membersVariableIdentifier = "members"
 
+			stmts := []dst.Stmt{
```
turbolent, line 644: Moved up from below

runtime/sema/gen/main.go
```diff
@@ -1020,7 +1171,14 @@ func functionTypeExpr(
 	if t.ReturnTypeAnnotation != nil {
 		returnTypeExpr = typeExpr(t.ReturnTypeAnnotation.Type, typeParams)
 	} else {
-		returnTypeExpr = typeVarIdent("Void")
```
turbolent, line 1023: Needed, to properly package-qualified `sema.VoidType`, if needed

runtime/stdlib/builtin.go
```diff
@@ -18,10 +18,6 @@
 
 package stdlib
 
-import "github.com/onflow/cadence/runtime/sema"
-
-var UInt8Type = sema.UInt8Type
```
turbolent, line 23: Such declarations are no longer needed

dsainati1 (approved):

SupunS (approved):

runtime/sema/account.gen.go
```diff
@@ -2099,57 +2099,30 @@ var CapabilitiesMappingType = &EntitlementMapType{
 
 func init() {
 	BuiltinEntitlementMappings[AccountMappingType.Identifier] = AccountMappingType
-	addToBaseActivation(AccountMappingType)
```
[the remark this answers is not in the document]
turbolent: No, this isn't related to adding support for initializers/constructors.

The generator produced this and the statement above literally: always register the entitlement (mapping) type in `BuiltinEntitlements` / `BuiltinEntitlementMappings`, and register the type in the base activation. 

However, there are a couple problems with this: This approach assumes the code is generated into the `sema` package. First, this is not always the case: The code generator may be used outside of `sema`, e.g. to generate type code for the standard library (`stdlib` package). There, `addToBaseActivation` is neither available (it is a private/unexported function), nor should the type be defined in the **global** base activation.

This became apparent when I refactored the test cases to not generated into a package named `stdlib` and made the test code importable, but instead generated it into a package named with the test name (see packages in `testdata`)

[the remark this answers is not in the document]
SupunS: ah ok, thanks for the details 👍

PR #2808, opened 2023-09-22
Replace panic with nil return in `newInjectedCompositeFieldsHandler`

m-Peter:
## Description

There is no reason to panic when the location of a `common.CompositeKindContract` is not a `common.AddressLocation`. We can simply return `nil`, as we already do for `stdlib.CryptoCheckerLocation`.
______


turbolent (approved):
Thanks!

SupunS (approved):

PR #2812, opened 2023-09-22
Type checking for default events

dsainati1:

Reference implementation for https://github.com/onflow/flips/pull/131, and as such will not be merged unless this FLIP is approved. This adds checking support for default events
______


runtime/sema/check_composite_declaration.go
```diff
@@ -1988,6 +2021,96 @@ func (checker *Checker) enumMembersAndOrigins(
...
+		}
+	}
+
+	for index, param := range eventType.ConstructorParameters {
+		paramType := param.TypeAnnotation.Type
+		paramExpr := constructorFunctionParameters[index]
+		paramDefaultArgument := paramExpr.DefaultArgument
+
+		// make `self` and `base` available when checking default arguments so the fields of the composite are available
+		checker.declareSelfValue(containerType, containerDeclaration.DeclarationDocString())
```
SupunS, line 2087: where (as in, in which scope) does this `self` variable get declared?

runtime/sema/check_composite_declaration.go
```diff
@@ -1988,6 +2021,96 @@ func (checker *Checker) enumMembersAndOrigins(
...
+			break
+		case *ast.IdentifierExpression:
+			// these are guaranteed to exist at time of destruction, so we allow them
+			if arg.Identifier.Identifier == SelfIdentifier || arg.Identifier.Identifier == BaseIdentifier {
+				break
+			}
+			// if it's an attachment, then it's also okay
+			if checker.typeActivations.Find(arg.Identifier.Identifier) != nil {
+				break
+			}
```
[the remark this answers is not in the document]
dsainati1: In theory this is true, but when combined with the restriction that default event arguments must be primitive types, using `self`, `base` or an attachment name is functionally impossible.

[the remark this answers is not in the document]
SupunS: Makes sense. Maybe add a test, just to verify proper/expected errors are reported, etc.

runtime/sema/check_composite_declaration.go
SupunS: >  I don't actually think there was any harm in declaring them like this

In the old way (i.e. defining them in the overall composite scope), could it conflict with field names of the container? (or are they defined in a separate scope? can't really remember). Maybe also add a test.

SupunS (approved):

runtime/sema/check_composite_declaration.go
```diff
@@ -1963,6 +1996,99 @@ func (checker *Checker) enumMembersAndOrigins(
...
+			*ast.FixedPointExpression,
+			*ast.PathExpression:
+			break
+		case *ast.IdentifierExpression:
+			// these are guaranteed to exist at time of destruction, so we allow them
+			if arg.Identifier.Identifier == SelfIdentifier || arg.Identifier.Identifier == BaseIdentifier {
+				break
+			}
+			// if it's an attachment, then it's also okay
+			if checker.typeActivations.Find(arg.Identifier.Identifier) != nil {
```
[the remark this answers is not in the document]
dsainati1: Not technically, but there is no possible syntax form that accepts types in value positions other than attachment access. I'll add some tests for this.

runtime/sema/check_composite_declaration.go
```diff
@@ -819,6 +843,15 @@ func (checker *Checker) declareCompositeLikeMembersAndValue(
 			nestedCompositeDeclarationVariable :=
 				checker.valueActivations.Find(identifier.Identifier)
 
+			if identifier.Identifier == ast.ResourceDestructionDefaultEventName {
+				// Find the default event's type declaration
+				defaultEventType :=
+					checker.typeActivations.Find(identifier.Identifier)
+				defaultEventComposite := defaultEventType.Type.(*CompositeType)
+				compositeType.DefaultDestroyEvent = defaultEventComposite
+				return
```
turbolent, line 852: Should this really return from the whole function?

runtime/sema/check_composite_declaration.go
```diff
@@ -1963,6 +1996,99 @@ func (checker *Checker) enumMembersAndOrigins(
...
+	containerDeclaration ast.Declaration,
+) {
+
+	// an event definition always has one "constructor" function in its declaration list
+	members := eventDeclaration.DeclarationMembers()
+	functions := members.SpecialFunctions()
+	constructorFunctionParameters := functions[0].FunctionDeclaration.ParameterList.Parameters
+
+	var checkParamExpressionKind func(ast.Expression)
+	checkParamExpressionKind = func(arg ast.Expression) {
```
turbolent, line 2012: Maybe refactor this function out into receiver function on the Checker

runtime/sema/check_composite_declaration.go
```diff
@@ -1963,6 +1996,99 @@ func (checker *Checker) enumMembersAndOrigins(
...
+			checkParamExpressionKind(arg.Expression)
+		case *ast.IndexExpression:
+			checkParamExpressionKind(arg.TargetExpression)
+			checkParamExpressionKind(arg.IndexingExpression)
+
+			// indexing expressions on arrays can fail, and must be disallowed, but
+			// indexing expressions on dicts, or composites (for attachments) will return `nil` and thus never fail
+			targetExprType := checker.Elaboration.IndexExpressionTypes(arg).IndexedType
+			switch targetExprType.(type) {
+			case ArrayType:
```
turbolent, line 2043: Maybe flip the logic here and disallow by default, and allow only known cases, like mentioned in the comment

runtime/sema/check_interface_declaration.go
```diff
@@ -432,7 +442,16 @@ func (checker *Checker) declareInterfaceMembersAndValue(declaration *ast.Interfa
...
 		for _, nestedCompositeDeclaration := range declaration.Members.Composites() {
 			if nestedCompositeDeclaration.Kind() == common.CompositeKindEvent {
-				checker.declareNestedEvent(nestedCompositeDeclaration, eventMembers, interfaceType)
+				if nestedCompositeDeclaration.IsResourceDestructionDefaultEvent() {
+					// Find the value declaration
+					nestedEvent :=
+						checker.typeActivations.Find(nestedCompositeDeclaration.Identifier.Identifier)
+					defaultEventComposite := nestedEvent.Type.(*CompositeType)
+					interfaceType.DefaultDestroyEvent = defaultEventComposite
+					// interfaceType.DefaultDestroyEvent =
```
turbolent, line 451: Leftover code?

runtime/sema/check_composite_declaration.go
dsainati1: No, removed

runtime/sema/check_composite_declaration.go
```diff
@@ -1963,6 +1995,111 @@ func (checker *Checker) enumMembersAndOrigins(
...
+	case *ast.StringExpression,
+		*ast.BoolExpression,
+		*ast.NilExpression,
+		*ast.IntegerExpression,
+		*ast.FixedPointExpression,
+		*ast.PathExpression:
+		break
+	case *ast.IdentifierExpression:
+		// these are guaranteed to exist at time of destruction, so we allow them
+		if arg.Identifier.Identifier == SelfIdentifier || arg.Identifier.Identifier == BaseIdentifier {
```
turbolent, line 2011: Maybe refactor the repeated `arg.Identifier.Identifier` expression throughout the case into a variable to make the code easier to read

runtime/sema/check_composite_declaration.go
```diff
@@ -1963,6 +1995,111 @@ func (checker *Checker) enumMembersAndOrigins(
...
+func (checker *Checker) checkDefaultDestroyEvent(
+	eventType *CompositeType,
+	eventDeclaration ast.CompositeLikeDeclaration,
+	containerType ContainerType,
+	containerDeclaration ast.Declaration,
+) {
+
+	// an event definition always has one "constructor" function in its declaration list
+	members := eventDeclaration.DeclarationMembers()
+	functions := members.SpecialFunctions()
```
turbolent, line 2064:
```suggestion
	functions := members.Initializers()
```

runtime/sema/check_composite_declaration.go
```diff
@@ -1963,6 +1995,111 @@ func (checker *Checker) enumMembersAndOrigins(
...
+			!IsSubType(unwrappedParamType, TheAddressType) &&
+			!IsSubType(unwrappedParamType, PathType) &&
+			!IsSubType(unwrappedParamType, BoolType) {
+			checker.report(&DefaultDestroyInvalidParameterError{
+				ParamType: paramType,
+				Range:     ast.NewRangeFromPositioned(checker.memoryGauge, paramExpr),
+			})
+		}
+
+		checker.checkDefaultDestroyParamExpressionKind(paramDefaultArgument)
```
turbolent, line 2099: Maybe refactor the loop body into a separate function

runtime/sema/check_composite_declaration.go
```diff
@@ -819,6 +843,14 @@ func (checker *Checker) declareCompositeLikeMembersAndValue(
 			nestedCompositeDeclarationVariable :=
 				checker.valueActivations.Find(identifier.Identifier)
 
+			if identifier.Identifier == ast.ResourceDestructionDefaultEventName {
```
turbolent, line 846: `identifier.Identifier == ast.ResourceDestructionDefaultEventName` is repeated in a few places, maybe create a function for it, like e.g. `IsResourceDestructionDefaultEventIdentifier`

turbolent (commented):
Nice!

runtime/sema/check_composite_declaration.go
```diff
@@ -2052,6 +2053,44 @@ func (checker *Checker) checkDefaultDestroyParamExpressionKind(
...
 
+func (checker *Checker) checkDefaultDestroyEventParam(
+	param Parameter,
+	index int,
+	constructorFunctionParameters []*ast.Parameter,
+	containerType ContainerType,
+	containerDeclaration ast.Declaration,
+) {
+	paramType := param.TypeAnnotation.Type
+	paramExpr := constructorFunctionParameters[index]
```
turbolent, line 2064: Instead of passing all constructor function parameters and the index, pass the `ast.Parameter` from the call-site directly. Maybe also rename `paramExpr`, as it is not an expression.

runtime/sema/check_composite_declaration.go
```diff
@@ -283,6 +290,10 @@ func (checker *Checker) visitCompositeLikeDeclaration(declaration ast.CompositeL
 	}
 
 	for _, nestedComposite := range members.Composites() {
+		if compositeType.DefaultDestroyEvent != nil {
+			// we enforce elsewhere that each composite can have only one default destroy event
+			checker.checkDefaultDestroyEvent(compositeType.DefaultDestroyEvent, nestedComposite, compositeType, declaration)
+		}
```
turbolent, line 296: Why does this have to be done for every nested composite?

runtime/sema/check_composite_declaration.go
```diff
@@ -819,6 +843,14 @@ func (checker *Checker) declareCompositeLikeMembersAndValue(
 			nestedCompositeDeclarationVariable :=
 				checker.valueActivations.Find(identifier.Identifier)
 
+			if ast.IsResourceDestructionDefaultEvent(identifier.Identifier) {
+				// Find the default event's type declaration
+				defaultEventType :=
+					checker.typeActivations.Find(identifier.Identifier)
+				defaultEventComposite := defaultEventType.Type.(*CompositeType)
+				compositeType.DefaultDestroyEvent = defaultEventComposite
```
turbolent, line 851: Maybe check here if `compositeType.DefaultDestroyEvent` is overwritten?

dsainati1: If it is, we will get a `RedeclarationError` (we have a test for this), so it doesn't really need to be checked here.

runtime/sema/check_composite_declaration.go
dsainati1: Good catch, this is missing a clause in the condition. Fixed.

runtime/sema/check_composite_declaration.go
turbolent: Ah, great! Maybe leave a comment above with this explanation

turbolent (approved):
Great work! 👍

Issue #2813, opened 2023-09-24
Add `replace` and `replaceAll` functions to String


darkdrag00nv2:
### Issue to be solved

Add `replace` and `replaceAll` instance functions to Cadence. 

1. `replace(_ of: String, _ with: String): String`: Returns a new String after replacing first occurrence of `of` with `with`.
2. `replaceAll(_ of: String, _ with: String): String`: Returns a new String after replacing all occurrences of `of` with `with`. 

Epic: https://github.com/onflow/cadence/issues/1972

turbolent:
`replaceAll` was added in #2814. We can add e.g. a `replaceFirst` when there is demand.

PR #2814, opened 2023-09-24
Introduce `String.replaceAll` instance function

darkdrag00nv2:
Work towards #2813

## Description

Adds `str.replaceAll(_ of: String, _ with: String): String` function to Cadence.
______

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -1358,6 +1358,25 @@ func (v *StringValue) GetMember(interpreter *Interpreter, locationRange Location
 				return v.Split(invocation.Interpreter, invocation.LocationRange, separator.Str)
 			},
 		)
+
+	case sema.StringTypeReplaceAllFunctionName:
+		return NewHostFunctionValue(
+			interpreter,
+			sema.StringTypeReplaceAllFunctionType,
+			func(invocation Invocation) Value {
+				of, ok := invocation.Arguments[0].(*StringValue)
+				if !ok {
+					panic(errors.NewUnreachableError())
+				}
+
+				with, ok := invocation.Arguments[1].(*StringValue)
+				if !ok {
+					panic(errors.NewUnreachableError())
+				}
+
+				return v.ReplaceAll(invocation.Interpreter, invocation.LocationRange, of.Str, with.Str)
+			},
+		)
 	}
 
 	return nil
@@ -1437,6 +1456,23 @@ func (v *StringValue) Split(inter *Interpreter, locationRange LocationRange, sep
 					return str
 				},
 			)
+		},
+	)
+}
+
+func (v *StringValue) ReplaceAll(inter *Interpreter, locationRange LocationRange, of string, with string) *StringValue {
+	// Over-estimate the resulting string length.
+	// In the worst case, `of` can be empty in which case, `with` will be added at every index.
+	// e.g. `of` = "", `v` = "ABC", `with` = "1": result = "1A1B1C1".
+	lengthOverEstimate := (2*len(v.Str) + 1) * len(with)
+
+	memoryUsage := common.NewStringMemoryUsage(lengthOverEstimate)
+
+	return NewStringValue(
+		inter,
+		memoryUsage,
+		func() string {
+			return strings.ReplaceAll(v.Str, of, with)
 		},
 	)
 }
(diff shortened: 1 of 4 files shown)


runtime/sema/string_type.go
```diff
@@ -152,6 +158,13 @@ It does not modify the original string.
 If either of the parameters are out of the bounds of the string, or the indices are invalid (` + "`from > upTo`" + `), then the function will fail
 `
 
+const StringTypeReplaceAllFunctionName = "replaceAll"
+const StringTypeReplaceAllFunctionDocString = `
+Returns a new string after replacing all the occurrences of parameter ` + "`of` with the parameter `with`" + `.
+
+If with is empty, it matches at the beginning of the string and after each UTF-8 sequence, yielding up to k+1 replacements for a k-rune string.
```
darkdrag00nv2, line 165: I am not 100% on whether we should use the term `rune` in the public documentation. Open to suggestions on a better term.

SupunS: We could say "... yielding up to k+1 replacements for a string of length `k`"

runtime/interpreter/value.go
```diff
@@ -1342,6 +1342,25 @@ func (v *StringValue) GetMember(interpreter *Interpreter, locationRange Location
...
+			func(invocation Invocation) Value {
+				of, ok := invocation.Arguments[0].(*StringValue)
+				if !ok {
+					panic(errors.NewUnreachableError())
+				}
+
+				with, ok := invocation.Arguments[1].(*StringValue)
+				if !ok {
+					panic(errors.NewUnreachableError())
+				}
```
SupunS, line 1359: Can these two parameters be of type `CharacterValue` (other than `StringValue`)?
Maybe also add a test(s) for passing `Character` typed values for `of` and `with` parameters?
```cadence
let a: Character = "x"
let b: Character = "y"

"something".replaceAll(of: a, with: b)
```

runtime/sema/string_type.go
darkdrag00nv2: Thanks, makes more sense. I removed the `up to` part since it'll be exactly `k + 1` for an empty `with` param.

darkdrag00nv2:
Force push to resolve merge conflict.

runtime/interpreter/value.go
darkdrag00nv2: ok, I can see that Kotlin `replace` allows either Char or String. May be we can allow that too.

SupunS: Actually, what I meant / my question was, whether we can get a `Character` value at runtime somehow.
If not that's totally fine. Just having a test-case to verify would be good.

With respect to: whether we need to support `Character`s in `replaceAll`, I think that could be a separate discussion/feature. And might be good it add it (if we want to add) in a separate PR. The smaller the PR is the better :)

darkdrag00nv2: Ah I see. No it is disallowed. Added a checker test case in 34067af33.

runtime/sema/string_type.go
```diff
@@ -163,6 +169,13 @@ It does not modify the original string.
 If either of the parameters are out of the bounds of the string, or the indices are invalid (` + "`from > upTo`" + `), then the function will fail
 `
 
+const StringTypeReplaceAllFunctionName = "replaceAll"
+const StringTypeReplaceAllFunctionDocString = `
+Returns a new string after replacing all the occurrences of parameter ` + "`of` with the parameter `with`" + `.
+
+If with is empty, it matches at the beginning of the string and after each UTF-8 sequence, yielding k+1 replacements for a string of length k.
```
SupunS, line 176:
```suggestion
If ` + "`with`" + ` is empty, it matches at the beginning of the string and after each UTF-8 sequence, yielding k+1 replacements for a string of length k.
```

SupunS (approved):

runtime/interpreter/value.go
SupunS: nice!

darkdrag00nv2:
@turbolent @dsainati1 Could you please take a look? Thanks!

dsainati1 (approved):

darkdrag00nv2:
@SupunS Should we merge this? Seems a relatively straightforward change.

turbolent:
Sorry I missed this! The PR looks good, but the function needs to have it's computation metered, as it performs a non-constant operation. The same also applies to the other stdlib PRs that got opened so far.

This might require "inlining" the replaceAll call to the Go stdlib, and calling `ReportComputation` with a new (?) `common.ComputationKind`

I opened #2879 to track this, see the details there.

PR #2815, opened 2023-09-24
Fix cadence-parser's esm package

wfalcon0x:
Closes #???

## Description

cadence-parser's build configuration is using tsc to build a dual package, however, the created esm module cannot be imported from a nodejs es module, as the `"type": "module"` is missing from the es module specific package.json.

This PR contains a way to solve this issue, although many other ways exist based on the choice of tooling. The solution included is an attempt to fix it with the smallest amount of change to the current toolings and configuration.

______

### DIFF
--- a/npm-packages/cadence-parser/package.json
+++ b/npm-packages/cadence-parser/package.json
@@ -20,8 +20,8 @@
   "scripts": {
     "build": "npm run build:types && npm run build:esm && npm run build:cjs && GOARCH=wasm GOOS=js go build -o ./dist/cadence-parser.wasm ../../runtime/cmd/parse",
     "build:types": "tsc --emitDeclarationOnly --module system --outDir dist/types",
-    "build:esm": "tsc --module es6 --target es6 --outDir dist/esm --declaration false",
-    "build:cjs": "tsc --module commonjs --target es6 --outDir dist/cjs --declaration false",
+    "build:esm": "tsc -p tsconfig.json --declaration false",
+    "build:cjs": "tsc -p tsconfig-cjs.json --declaration false && tsconfig-to-dual-package",
     "test": "jest"
   },
   "license": "Apache-2.0",
@@ -31,6 +31,7 @@
     "jest": "^28.1.2",
     "node-fetch": "^2.6.1",
     "ts-jest": "^28.0.5",
+    "tsconfig-to-dual-package": "^1.2.0",
     "typescript": "^4.7.4"
   },
   "files": [
--- a/npm-packages/cadence-parser/src/index.ts
+++ b/npm-packages/cadence-parser/src/index.ts
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-import { go } from './go'
+import { go } from './go.js'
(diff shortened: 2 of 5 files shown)


turbolent:
Thank you! Is there an easy way to verify that this setup allows usage of the package in various environments (e.g. browser, Node.JS, etc.)?

nialexsan:
@wfalcon0x could you please share code where you have issues with importing this package?
with our current setup it should be possible to use the package with `import` and with `require` from nodejs apps
https://github.com/onflow/cadence/blob/54fe090a65e377e5b64148b94d8fc3351d893dab/npm-packages/cadence-parser/package.json#L10-L19

wfalcon0x:
> @wfalcon0x could you please share code where you have issues with importing this package? with our current setup it should be possible to use the package with `import` and with `require` from nodejs apps
> 
> https://github.com/onflow/cadence/blob/54fe090a65e377e5b64148b94d8fc3351d893dab/npm-packages/cadence-parser/package.json#L10-L19

@nialexsan Running the following code from nodejs fails with a syntax error.
```
import {CadenceParser} from "@onflow/cadence-parser"
```
The error is:
```
import {CadenceParser} from "@onflow/cadence-parser"
        ^^^^^^^^^^^^^
SyntaxError: Named export 'CadenceParser' not found. The requested module '@onflow/cadence-parser' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@onflow/cadence-parser';
const {CadenceParser} = pkg;

    at ModuleJob._instantiate (node:internal/modules/esm/module_job:127:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:193:5)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:337:24)
    at async loadESM (node:internal/process/esm_loader:88:5)
    at async handleMainPromise (node:internal/modules/run_main:61:12)
```
The solution is to add a package.json into the /dist/esm folder with the following content:
```
{
  "type": "module"
}
```

wfalcon0x:
> Thank you! Is there an easy way to verify that this setup allows usage of the package in various environments (e.g. browser, Node.JS, etc.)?

I tested it as follows: I first published the changes to my local npm repository, and then verified its functionality in the following scenarios:

- In Node.js using CommonJS with the "require" syntax.
- In Node.js using ES6 modules with the "import" syntax.
- In a web browser, I created a React application using "create-react-app" and imported the library there.

Please note, the added dev dependency (tsconfig-to-dual-package@1.2.0) requires { node: '>=18.3.0 || >=16.17.0' }

turbolent:
@nialexsan Thank you for having a look! Could you please have another look at what wfalcon0x posted?

turbolent:
@wfalcon0x Thank you again for your contribution and your patience! Could you please see if #2851 improved things for you? If yes, we can maybe close this PR. Alex provides some details on how this PR might not be the way we want to go here: https://github.com/onflow/cadence/pull/2851#issuecomment-1749784391. Hope that makes sense (I don't know much about it), but please report back if you're still running into issues. Thanks!

PR #2817, opened 2023-09-26
Interpreting for default events

dsainati1:

Reference implementation for https://github.com/onflow/flips/pull/131, and as such will not be merged unless this FLIP is approved. This adds runtime support for default events
______


dsainati1:
@turbolent @SupunS this is the last PR necessary for removal of destructors

turbolent (commented):
Nice!

runtime/interpreter/value.go
```diff
@@ -16379,7 +16379,8 @@ type CompositeField struct {
 	Name  string
 }
 
-const attachmentNamePrefix = "$"
+const unrepresentableNamePrefix = "$"
+const resourceDefaultDestroyEventPrefix = "ResourceDestroyed" + unrepresentableNamePrefix
```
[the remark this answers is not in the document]
dsainati1: The default destroy event is encoded like a function on resources, since it behaves differently than other type definitions that can appear inside e.g. interfaces. This is used to generate a guaranteed fresh name for that function to be used when looking up the event constructor in the resource's function map.

runtime/interpreter/interpreter.go
```diff
@@ -978,6 +980,47 @@ func (interpreter *Interpreter) declareAttachmentValue(
 	return interpreter.declareCompositeValue(declaration, lexicalScope)
 }
 
+// evaluates all the implicit default arguments to the default destroy event
```
turbolent, line 983:
```suggestion
// evaluateDefaultDestroyEvent evaluates all the implicit default arguments to the default destroy event.
```

runtime/interpreter/interpreter.go
```diff
@@ -978,6 +980,47 @@ func (interpreter *Interpreter) declareAttachmentValue(
...
+	var self MemberAccessibleValue = containerComposite
+	if containerComposite.Kind == common.CompositeKindAttachment {
+		var base *EphemeralReferenceValue
+		base, self = attachmentBaseAndSelfValues(interpreter, containerComposite)
+		interpreter.declareVariable(sema.BaseIdentifier, base)
+	}
+	interpreter.declareVariable(sema.SelfIdentifier, self)
+
+	for _, parameter := range parameters {
+		// lazily evaluate the default argument expressions
```
turbolent, line 1011: In what sense is the evaluation "lazy" here? It looks like all arguments are eagerly evaluated

runtime/interpreter/interpreter.go
```diff
@@ -978,6 +980,47 @@ func (interpreter *Interpreter) declareAttachmentValue(
...
+		// note that we must evaluate them in the interpreter context that existed when the event
+		// was defined (i.e. the contract defining the resource) rather than the interpreter context
+		// that exists when the resource is destroyed. We accomplish this by using the original interpreter of the
+		// composite declaration, rather than the interpreter of the destroy expression
```
turbolent, line 1015: Where is the "right" interpreter (even definition instead of destroy expression) used?

Maybe just use (or perform the same logic as) `Interpreter.invokeInterpretedFunction`, i.e. push the function's (event constructors) activation above.

runtime/interpreter/interpreter.go
```diff
@@ -978,6 +980,47 @@ func (interpreter *Interpreter) declareAttachmentValue(
...
+//
+// the handling of default arguments makes a number of assumptions to simplify the implementation;
+// namely that a) all default arguments are lazily evaluated at the site of the invocation,
+// b) that either all the parameters or none of the parameters of a function have default arguments,
+// and c) functions cannot currently be explicitly invoked if they have default arguments
+//
+// if we plan to generalize this further, we will need to relax those assumptions
+func (interpreter *Interpreter) evaluateDefaultDestroyEvent(
+	containerComposite *CompositeValue,
+	compositeDecl *ast.CompositeDeclaration,
```
turbolent, line 993: Is `compositeDecl` here the composite declaration of the `ResourceDestroyed` event? Maybe improve the naming to make that more obvious.

Maybe also leave a note in the function docstring that `containerComposite` is the resource that is being destroyed, not the composite value representing the event.

runtime/interpreter/interpreter.go
```diff
@@ -978,6 +980,47 @@ func (interpreter *Interpreter) declareAttachmentValue(
...
+	compositeType *sema.CompositeType,
```
turbolent, line 994: It is not quite clear what `compositeType` is here – is it the composite type of the `ResourceDestroyed` event?

runtime/interpreter/value.go
```diff
@@ -16585,6 +16589,22 @@ func (v *CompositeValue) IsDestroyed() bool {
 	return v.isDestroyed
 }
 
+func resourceDefaultDestroyEventName(t sema.ContainerType) string {
+	return resourceDefaultDestroyEventPrefix + string(t.ID())
+}
+
+func (v *CompositeValue) defaultDestroyEventConstructors() (constructors []FunctionValue) {
```
turbolent, line 16596: I was a bit confused at first why the constructor for the resource destroyed event was not just simply stored by the constant name (`ResourceDestroyed`), but instead is suffixed with the type.

Could you please document that this is necessary to support programs where the resource conforms to interfaces, and the interfaces likewise define resource destroyed events, maybe even with a small example? 

Also, do we have tests for the case of multiple events having to be emitted for the destruction of a single resource?

runtime/interpreter/value.go
```diff
@@ -16613,6 +16633,39 @@ func (v *CompositeValue) Destroy(interpreter *Interpreter, locationRange Locatio
...
+	// so that we can leverage existing atree encoding and decoding. However, we need to make sure functions are initialized
+	// if the composite was recently loaded from storage
+	if v.Functions == nil {
+		v.Functions = interpreter.SharedState.typeCodes.CompositeCodes[v.TypeID()].CompositeFunctions
+	}
+	for _, constructor := range v.defaultDestroyEventConstructors() {
+
+		// pass the container value to the creation of the default event as an implicit argument, so that
+		// its fields are accessible in the body of the event constructor
+		mockInvocation := NewInvocation(
```
turbolent, line 16651: Can probably just be called "invocation" (or even "eventConstructorInvocation"), there's nothing "mock" about it here I think

runtime/interpreter/value.go
```diff
@@ -16613,6 +16633,39 @@ func (v *CompositeValue) Destroy(interpreter *Interpreter, locationRange Locatio
...
+			[]sema.Type{},
+			nil,
+			locationRange,
+		)
+
+		event := constructor.invoke(mockInvocation).(*CompositeValue)
+		eventType := interpreter.MustSemaTypeOfValue(event).(*sema.CompositeType)
+
+		// emit the event once destruction is complete
+		defer interpreter.emitEvent(event, eventType, locationRange)
```
turbolent, line 16666: 👌

turbolent (commented):
Nice, great work!

Thank you again for the sync, it clarified thins a lot 🙏

runtime/interpreter/interpreter.go
dsainati1: They are "lazy" with respect to the event's declaration: if i declare a default event `ResourceDestroyed(foo: Int = self.x)`, `self.x` is evaluated in the context that exists when the event is destroyed, not the context when it is declared. This function is only called after the destroy triggers the event emission, so with respect to this function it's "eager".

runtime/interpreter/interpreter.go
dsainati1: Removed this argument, it was unnecessary

runtime/interpreter/value.go
dsainati1: Yes! E.g. https://github.com/onflow/cadence/pull/2817/files#diff-ed4abd754a588f51cf2a7e972e435b1c81c5b4153b25fd927f18ab36d6c429fdR2249

runtime/interpreter/interpreter.go
turbolent: Ah, I see! Could you please add a (variant of) this explanation as a comment?

runtime/interpreter/interpreter.go
```diff
@@ -978,6 +980,51 @@ func (interpreter *Interpreter) declareAttachmentValue(
...
+) (arguments []Value) {
+	parameters := eventDecl.DeclarationMembers().Initializers()[0].FunctionDeclaration.ParameterList.Parameters
+
+	defer func() {
+		// Only unwind the call stack if there was no error
+		if r := recover(); r != nil {
+			panic(r)
+		}
+		interpreter.SharedState.callStack.Pop()
+	}()
```
turbolent, line 1003: I guess pushing the invocation onto the call-stack is mostly useful for nicer error reporting?

Eventually we should have equivalents for `invokeInterpretedFunction` and `invokeInterpretedFunctionActivated` for host functions. Currently, invocations of host functions don't push the invocation onto the call-stack, so code like this (and the push at the call-site of `evaluateDefaultDestroyEvent`) are needed

runtime/interpreter/interpreter.go
```diff
@@ -1107,6 +1161,23 @@ func (interpreter *Interpreter) declareNonEnumCompositeValue(
...
+					return nil
+				}
+
+				// event interfaces do not exist
+				compositeDecl := declaration.(*ast.CompositeDeclaration)
+				if compositeDecl.IsResourceDestructionDefaultEvent() {
+					// we implicitly pass the containing composite value as an argument to this invocation
+					containerComposite := invocation.Arguments[0].(*CompositeValue)
+					interpreter.activations.PushNewWithParent(inter.activations.CurrentOrNew())
+					interpreter.SharedState.callStack.Push(invocation)
```
turbolent, line 1174: Maybe move this into `evaluateDefaultDestroyEvent` if possible, to keep the push and pop in the same place

runtime/interpreter/interpreter.go
```diff
@@ -1107,6 +1161,23 @@ func (interpreter *Interpreter) declareNonEnumCompositeValue(
...
+					invocation.Arguments = interpreter.evaluateDefaultDestroyEvent(
```
turbolent, line 1175: I had to re-read this piece a couple times to understand what is going on: There are two interpreters:
1. The one used when the event is declared, the "outer" `interpreter`
2. The one of the invocation, when the event is emitted, the "inner" `inter`

The declaration `interpreter` is used for evaluating the default arguments in the last line (https://github.com/onflow/cadence/pull/2817/files#diff-a36f730b18ca31316909bd74125003c1fbe785a13f926b4a0631e7c0cc6a51e4R1175). 

Is it correct to push the invocation `inter`'s current activation before though?

I'm worried that that is effectively dynamic instead of lexical scope. I think this works, because there is no way to actually refer to the `inter`'s activation's variables, because all the default argument expressions can only be constants or refer to `self`, and `base` in an attachment, right? The default arguments cannot refer to e.g. global variables?

Maybe add a test case that ensures the pushed activation is correct, e.g.

```cadence
let x = 1

resource R {
    event ResourceDestroyed(x: x)
}

fun test() {
    let x = 2
    let r <- create R()
    // should emit `R.ResourceDestroyed(x: 1)`, not `R.ResourceDestroyed(x: 2)`
    destroy r
}
```

Also, maybe rename `inter` to `invocationInter` or even shadow `interpreter`, to make it clearer what interpreter is what. It matters a lot here because of the scoping / activations.

dsainati1: This test is not possible; `x` is not a valid argument to a default destroy event, since in general we cannot guarantee that an external variable access will not fail (e.g. an invalidated resource reference).

dsainati1: I've added a variant of it with a field on `R` though.

dsainati1: I renamed the variables to make it clearer what is going on though

turbolent: Could you please still add the example above as a test case? It would be good to ensure that the activation code is correct. 

In the test, allow the example above to type check, i.e. ignore the expected type checking error that is produced for `= x`. You can do this by using `parseCheckAndInterpretWithOptions` and 
```go
ParseCheckAndInterpretOptions{
	HandleCheckerError: func(err error) {
		errs := checker.RequireCheckerErrors(t, err, 1)
		// ...
	},
}
```

dsainati1: ah okay, yeah I didn't know you could do that! Will add the test.

dsainati1: Great test! Turns out neither `invocationInterpreter` nor `declarationInterpreter` were the right scoping for this; instead I captured the activations that exist at the time the event's host function is created, and manually put them on top of the stack for the duration of default argument interpreting. This makes the behavior correct I think. I've added a comment to this effect as well.

runtime/interpreter/interpreter.go
```diff
@@ -1093,24 +1142,44 @@ func (interpreter *Interpreter) declareNonEnumCompositeValue(
...
+				compositeDecl := declaration.(*ast.CompositeDeclaration)
+				if compositeDecl.IsResourceDestructionDefaultEvent() {
+					// we implicitly pass the containing composite value as an argument to this invocation
+					containerComposite := invocation.Arguments[0].(*CompositeValue)
+					invocation.Arguments = declarationInterpreter.evaluateDefaultDestroyEvent(
+						containerComposite,
+						compositeDecl,
+						// to properly lexically scope the evaluation of default arguments, we capture the
+						// activations existing at the time when the event was defined and use them here
+						declarationActivation,
```
turbolent, line 1175: 👌

runtime/interpreter/interpreter.go
```diff
@@ -978,6 +980,46 @@ func (interpreter *Interpreter) declareAttachmentValue(
...
+	declarationInterpreter.activations.Push(declarationActivation)
+	defer declarationInterpreter.activations.Pop()
+
+	var self MemberAccessibleValue = containingResourceComposite
+	if containingResourceComposite.Kind == common.CompositeKindAttachment {
+		var base *EphemeralReferenceValue
+		base, self = attachmentBaseAndSelfValues(declarationInterpreter, containingResourceComposite)
+		declarationInterpreter.declareVariable(sema.BaseIdentifier, base)
+	}
+	declarationInterpreter.declareVariable(sema.SelfIdentifier, self)
```
turbolent, line 1007: I'm not sure, but it might be necessary to push new activation _on top_ of the declaration activation, i.e. have a separate activation for the scope of just the event constructor arguments. The activation that is pushed onto the activation stack is currently the declaration activation, and then it is mutated below and `self` (and `base` for attachments) are declared in it – that should happen in a fresh activation.

I think, but am not sure and don't know if it's an issue, that the current implementation basically "leaks" `self` (and `base` for attachments) to the code following the `ResourceDestroyed` event declaration. I can't think of a test case of the top of my head which could test it though.

Maybe also double check this is done for the checker part of this feature.

turbolent (commented):
Looks good! Sorry, just one last things re: interpreter activations (and maybe also check checker activations)

turbolent (approved):
Great work!

turbolent:
@SupunS Can you please also have a look?

SupunS (approved):
Looks good!

runtime/tests/interpreter/attachments_test.go
```diff
@@ -1240,71 +1240,154 @@ func TestInterpretAttachmentDestructor(t *testing.T) {
...
-		_, err := inter.Invoke("test")
+		_, err = inter.Invoke("test")
 		require.NoError(t, err)
 
-		// DestructorTODO: replace with test for destruction event for R being the last emitted
+		require.Len(t, events, 4)
+		// the only part of this order that is important is that `R` is last
+		require.Equal(t, events[0].QualifiedIdentifier, "B.ResourceDestroyed")
+		require.Equal(t, events[1].QualifiedIdentifier, "C.ResourceDestroyed")
+		require.Equal(t, events[2].QualifiedIdentifier, "A.ResourceDestroyed")
```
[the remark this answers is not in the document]
dsainati1: Yes, we just don't specify the order in the spec, but it is deterministic.

PR #2819, opened 2023-09-26
Use kr/pretty instead of go-test/deep to prevent empty diffs, remove gocap

turbolent:
## Description

#2799 ran into an odd case: `assert.Equal` was reporting a difference, while `deep.Equal` reported no difference.

It turns out that `deep.Equal` is wrong, and `assert.Equal` is right (I do not fully understand why `deep` does not report diffs as expected.).

While `assert.Equal` is right, it does not return diffs.

Switch to and rely on `pretty.Diff`, which correctly performs and reports diffs.

This also removes the global `MaxDepth` settings.

______


dsainati1 (approved):

SupunS (approved):

turbolent:
https://github.com/cugu/gocap is no longer maintained, so I removed it. TBH, it was not as useful as hoped. 

After updating to the latest version (v0.2.1), pretty much all packages required `unsafe` and `reflect`, and we pretty much just have to accept that – the Go community does not care as much about unsafe code as e.g. the Rust community, where usage of Rust's `unsafe` is frowned upon and often attempted to be avoided.

SupunS (approved):

PR #2821, opened 2023-09-27
Check native function declarations

turbolent:
## Description

Previously, the parser gained support for parsing the `native` modifier for functions, and the checker rejected the modifier, gated behind a configuration option.

Complete support for checking native functions, by requiring that if native functions are enabled, the function body is empty, and ignore checking it. 

______

### DIFF
--- a/runtime/sema/errors.go
+++ b/runtime/sema/errors.go
@@ -878,6 +878,23 @@ func (e *InvalidNativeModifierError) Error() string {
 	return "invalid native modifier for declaration"
 }
 
+// NativeFunctionWithImplementationError
+
+type NativeFunctionWithImplementationError struct {
+	ast.Range
+}
+
+var _ SemanticError = &NativeFunctionWithImplementationError{}
+var _ errors.UserError = &NativeFunctionWithImplementationError{}
+
+func (*NativeFunctionWithImplementationError) isSemanticError() {}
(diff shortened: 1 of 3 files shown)


runtime/sema/errors.go
```diff
@@ -878,6 +878,23 @@ func (e *InvalidNativeModifierError) Error() string {
...
+
+var _ SemanticError = &NativeFunctionWithImplementationError{}
+var _ errors.UserError = &NativeFunctionWithImplementationError{}
+
+func (*NativeFunctionWithImplementationError) isSemanticError() {}
+
+func (*NativeFunctionWithImplementationError) IsUserError() {}
+
+func (e *NativeFunctionWithImplementationError) Error() string {
+	return "native function may not have an implementation"
```
SupunS, line 895:
```suggestion
	return "native function must not have an implementation"
```

runtime/tests/checker/function_test.go
SupunS, line 1: Can you please also add a test to see if the composite functions get defined inside the composite type as members?

turbolent: Good idea! Added in d174f7be8598456f5c126c11ccb742b193208391

SupunS (approved):
Nice!

PR #2838, opened 2023-10-04
Skip adding malformed entitlement relations to maps

dsainati1:
When a non-entitlement type was used in an entitlement mapping relation, we would properly report an error here, but add the "relation" to the map anyways. This could cause issues later in the checker when attempting to check code that used the map. Instead, just skip adding the malformed relation. 

Thanks to @oebeling for the report

______

### DIFF
--- a/runtime/tests/checker/entitlements_test.go
+++ b/runtime/tests/checker/entitlements_test.go
@@ -7334,15 +7334,15 @@ func TestCheckEntitlementMissingInMap(t *testing.T) {
             access(M) var foo: auth(M) &Int
             init() {
                 self.foo = &3 as auth(X) &Int
-                var selfRef = &self as auth(X) &S;
-                selfRef.foo;
+                var selfRef = &self as auth(X) &S
+                selfRef.foo
             }
         }
     `)
 
 		errors := RequireCheckerErrors(t, err, 2)
-		require.IsType(t, errors[0], &sema.NotDeclaredError{})
-		require.IsType(t, errors[1], &sema.InvalidNonEntitlementTypeInMapError{})
+		require.IsType(t, &sema.NotDeclaredError{}, errors[0])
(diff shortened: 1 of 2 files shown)


runtime/tests/checker/entitlements_test.go
```diff
@@ -7312,3 +7315,57 @@ func TestCheckEntitlementOptionalChaining(t *testing.T) {
...
+                self.foo = &3 as auth(X) &Int
+                var selfRef = &self as auth(X) &S;
+                selfRef.foo;
+            }
+        }
+    `)
+
+		errors := RequireCheckerErrors(t, err, 2)
+		require.IsType(t, errors[0], &sema.NotDeclaredError{})
+		require.IsType(t, errors[1], &sema.InvalidNonEntitlementTypeInMapError{})
```
turbolent, line 7345: Here and below: Argument order is expected, actual, so need to swap arguments

runtime/tests/checker/entitlements_test.go
```diff
@@ -7312,3 +7315,57 @@ func TestCheckEntitlementOptionalChaining(t *testing.T) {
...
+        access(all) entitlement X
+        access(all) entitlement mapping M {
+            X -> X
+            NonExistingEntitlement -> X
+        }
+        access(all) struct S {
+            access(M) var foo: auth(M) &Int
+            init() {
+                self.foo = &3 as auth(X) &Int
+                var selfRef = &self as auth(X) &S;
```
turbolent, line 7337: Here and below: maybe remove unnecessary semicolons.

Can the code here simplified to just what is needed, or is is already a minimal test?

dsainati1: Yes, the use of the entitlement and the member access were necessary to reproduce the crash; the reported errors here were already reported, we just continued checking after that point and then crashed.

turbolent (approved):
Nice!

Issue #2839, opened 2023-10-04
Possible to cause a runtime error due to invalid entitlements that should be caught at checker time


dsainati1:
### Current Behavior

```
entitlement X
entitlement Y
entitlement mapping M {
    X -> Insert
    Y -> Remove
}
struct S {
    access(M) var member: auth(M) &[Int]?
    init() {
        self.member = nil;
    }
    access(all) fun grantRemovePrivileges(param: auth(Insert) &[Int]): Void{
        var selfRef = &self as auth(X) &S;
        selfRef.member = param;
    }
}
fun main(): Void {
    var arr: [Int] = [123];
    var arrRef = &arr as auth(Insert) &[Int];
    let s = S()
    s.grantRemovePrivileges(param: arrRef);
    s.member?.removeLast() // Caught by checkMemberAccess type check
}
```

### Expected Behavior

Should fail in the type checker

### Steps To Reproduce

Create this as a test on stable cadence branch

### Environment

```markdown
- Cadence version: Stable Cadence
- Network:
```

PR #2840, opened 2023-10-04
Require full codomain of map when assigning to mapped field

dsainati1:
Closes https://github.com/onflow/cadence/issues/2839

It was possible to circumvent the checker when assigning to a mapped entitlement field, by assigning a field to a reference and then accessing it on an owned value, tricking the checker into "escalating" the entitlements on the inner value. This was caught at runtime, but should be caught by the checker instead. The checker should require a fully-entitled input when assigning to a mapped field. 

Thanks @oebeling for the report
______


bluesign:
I was just gonna report something similar. Let me check if this fixes that.

PR #2847, opened 2023-10-05
Properly set base on loaded attachments during attachment iteration

dsainati1:
The `base` value was not being properly set on attachment values loaded during iteration. 

Thanks @oebeling for the report

______

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -16513,7 +16513,7 @@ func (v *CompositeValue) Destroy(interpreter *Interpreter, locationRange Locatio
 				// is a necessary pre-requisite for calling any members of the attachment. However, in
 				// the case of a destructor, this is called implicitly, and thus must have its `base`
 				// set manually
-				attachment.setBaseValue(interpreter, v, attachmentBaseAuthorization(interpreter, attachment))
+				attachment.setBaseValue(interpreter, v)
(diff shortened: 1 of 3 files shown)


runtime/interpreter/value.go
```diff
@@ -17762,6 +17763,7 @@ func (v *CompositeValue) forEachAttachment(interpreter *Interpreter, _ LocationR
 			// attachments is added that takes a `fun (&Attachment): Void` callback, the `f` provided here
 			// should convert the provided attachment value into a reference before passing it to the user
 			// callback
+			attachment.setBaseValue(interpreter, v)
```
turbolent, line 17766: This seems to be the fix, setting the base value on the attachment before invoking the iteration callback.

Are the other changes in the PR also fixes, i.e. removing the last parameter of `setBaseValue` function and adjusting the call-sites, or is that just a simplification?

turbolent (approved):
👍

runtime/interpreter/value.go
dsainati1: Yeah, while doing this I noticed that every call to `setBaseValue` computed the `baseAuthorization` the same way, so I removed the argument and moved that computation into the body of the method.

turbolent: Nice!

turbolent (approved):

PR #2851, opened 2023-10-05
mjs build for cadence-parser

nialexsan:
alternative approach for #2815

## Description

cadence-parser's build configuration is using tsc to build a dual package, however, the created esm module cannot be imported from a nodejs es module, as the "type": "module" is missing from the es module specific package.json.

This PR contains a way to solve this issue, although many other ways exist based on the choice of tooling. The solution included is an attempt to fix it with the smallest amount of change to the current toolings and configuration.

______

### DIFF
--- a/npm-packages/cadence-parser/package-lock.json
+++ b/npm-packages/cadence-parser/package-lock.json
@@ -1,0 +1,3822 @@
+{
+  "name": "@onflow/cadence-parser",
+  "version": "0.41.1",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "@onflow/cadence-parser",
+      "version": "0.41.1",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "get-random-values": "^2.0.0"
+      },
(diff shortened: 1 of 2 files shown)


jribbink (approved):
beautiful

turbolent:
Thank you for looking into this and opening an alternative @nialexsan! How does it differ from #2815? What is wrong with approach in the existing PR and how is the approach in this PR better? (Not arguing in any way, just trying to understand)

Could you please also follow and fill out the checklist in the template? Thanks!

nialexsan:
@turbolent this PR implements a generic approach for bundling and it uses a well known tool to do that
a few of the team's concerns regarding #2815 were 
- that it uses [tsconfig-to-dual-package](https://www.npmjs.com/package/tsconfig-to-dual-package) package with only 600 weekly downloads
- having two different `tsconfig.json`s is not a widely acceptable practice, and feels rather hacky

turbolent:
@nialexsan Thank you for the explanation! Agreed that it is preferable to have a single config 👍

turbolent (approved):

PR #2859, opened 2023-10-09
Allow specifying a mapping with source paths for locations in coverage reports

m-Peter:
## Description

Having `AddressLocation` or `StringLocation` in coverage reports, is not usable when trying to use these reports with standard tools, such as `LCOV`. The file path is what is needed to exploit the existing tools dealing with the `LCOV` format.

The `flow.json` config file, already defines the source file for all contracts involved in a dApp, e.g.:

```json
{
  "contracts": {
    "ApprovalVoting": {
      "source": "contracts/ApprovalVoting.cdc",
      "aliases": {
        "testing": "0000000000000007"
      }
    },
    "ArrayUtils": {
      "source": "contracts/ArrayUtils.cdc",
      "aliases": {
        "testing": "0000000000000007"
      }
    },
    "FooContract": {
      "source": "contracts/FooContract.cdc",
      "aliases": {
        "testing": "0000000000000007"
      }
    },
    "StringUtils": {
      "source": "contracts/StringUtils.cdc",
      "aliases": {
        "testing": "0000000000000007"
      }
    }
  }
}
```
______

### DIFF
--- a/runtime/coverage.go
+++ b/runtime/coverage.go
@@ -120,6 +120,9 @@ type CoverageReport struct {
 	// This filter can be used to inject custom logic on
 	// each location/program inspection.
 	LocationFilter LocationFilter `json:"-"`
+	// Contains a mapping with human-friendly names for
+	// locations. This could be the filepath of a location.
+	LocationMappings map[string]string `json:"-"`
 }
 
 // WithLocationFilter sets the LocationFilter for the current
@@ -128,6 +131,14 @@ func (r *CoverageReport) WithLocationFilter(
 	locationFilter LocationFilter,
 ) {
 	r.LocationFilter = locationFilter
+}
+
+// WithLocationMappings sets the LocationMappings for the current
+// CoverageReport.
+func (r *CoverageReport) WithLocationMappings(
+	locationMappings map[string]string,
+) {
+	r.LocationMappings = locationMappings
 }
 
 // ExcludeLocation adds the given location to the map of excluded
@@ -423,7 +434,8 @@ type lcAlias struct {
 func (r *CoverageReport) MarshalJSON() ([]byte, error) {
 	coverage := make(map[string]lcAlias, len(r.Coverage))
 	for location, locationCoverage := range r.Coverage { // nolint:maprange
-		coverage[location.ID()] = lcAlias{
+		locationSource := r.locationSource(location)
+		coverage[locationSource] = lcAlias{
 			LineHits:    locationCoverage.LineHits,
 			MissedLines: locationCoverage.MissedLines(),
 			Statements:  locationCoverage.Statements,
@@ -505,7 +517,8 @@ func (r *CoverageReport) MarshalLCOV() ([]byte, error) {
 	buf := new(bytes.Buffer)
 	for _, location := range locations {
 		coverage := r.Coverage[location]
-		_, err := fmt.Fprintf(buf, "TN:\nSF:%s\n", location.ID())
+		locationSource := r.locationSource(location)
+		_, err := fmt.Fprintf(buf, "TN:\nSF:%s\n", locationSource)
 		if err != nil {
 			return nil, err
 		}
@@ -538,4 +551,26 @@ func (r *CoverageReport) MarshalLCOV() ([]byte, error) {
 	}
 
 	return buf.Bytes(), nil
+}
+
+// Given a common.Location, returns its mapped source, if any.
+// Defaults to the location's ID().
+func (r *CoverageReport) locationSource(location common.Location) string {
+	var locationIdentifier string
+
+	switch loc := location.(type) {
+	case common.AddressLocation:
+		locationIdentifier = loc.Name
+	case common.StringLocation:
+		locationIdentifier = loc.String()
+	default:
+		locationIdentifier = loc.ID()
+	}
+
+	locationSource, ok := r.LocationMappings[locationIdentifier]
+	if !ok {
+		locationSource = location.ID()
+	}
+
+	return locationSource
 }
--- a/runtime/coverage.go
+++ b/runtime/coverage.go
@@ -119,10 +119,10 @@ type CoverageReport struct {
 	ExcludedLocations map[common.Location]struct{} `json:"-"`
 	// This filter can be used to inject custom logic on
 	// each location/program inspection.
-	LocationFilter LocationFilter `json:"-"`
-	// Contains a mapping with human-friendly names for
-	// locations. This could be the filepath of a location.
-	LocationMappings map[string]string `json:"-"`
+	locationFilter LocationFilter
+	// Contains a mapping with source paths for each
+	// location.
+	locationMappings map[string]string
 }
 
 // WithLocationFilter sets the LocationFilter for the current
@@ -130,7 +130,7 @@ type CoverageReport struct {
 func (r *CoverageReport) WithLocationFilter(
 	locationFilter LocationFilter,
 ) {
-	r.LocationFilter = locationFilter
+	r.locationFilter = locationFilter
 }
 
 // WithLocationMappings sets the LocationMappings for the current
@@ -138,7 +138,7 @@ func (r *CoverageReport) WithLocationFilter(
 func (r *CoverageReport) WithLocationMappings(
 	locationMappings map[string]string,
 ) {
-	r.LocationMappings = locationMappings
+	r.locationMappings = locationMappings
 }
 
 // ExcludeLocation adds the given location to the map of excluded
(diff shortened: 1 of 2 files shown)


runtime/coverage.go
```diff
@@ -505,7 +517,8 @@ func (r *CoverageReport) MarshalLCOV() ([]byte, error) {
 	buf := new(bytes.Buffer)
 	for _, location := range locations {
 		coverage := r.Coverage[location]
-		_, err := fmt.Fprintf(buf, "TN:\nSF:%s\n", location.ID())
+		locationSource := r.locationSource(location)
```
m-Peter, line 520: I think the `LCOV` format will consider any report problematic, if `SF` does not contain a legal filename.
I am doing some `sed` manipulation, in order to replace any `AddressLocation`, see: https://github.com/m-Peter/flow-code-coverage/blob/main/normalize_coverage_report.sh
Would it make sense to drop any location for which there is not location source file?

turbolent: When would a location be not a legal file name? AFAICS, the IDs of location kinds themselves are legal, e.g. `A.<hex-address>` for address locations, and the qualified identifier suffix is always a Cadence identifier (`[A-Za-z0-9_]+`), plus dots, which is also legal in filenames

runtime/coverage.go
```diff
@@ -539,3 +552,25 @@ func (r *CoverageReport) MarshalLCOV() ([]byte, error) {
 
 	return buf.Bytes(), nil
 }
+
+// Given a common.Location, returns its mapped source, if any.
+// Defaults to the location's ID().
+func (r *CoverageReport) locationSource(location common.Location) string {
```
turbolent, line 558: Maybe always use the ID as the "location source", instead of stripping parts of it (i.e. the kind), which could lead to clashes (e.g. `import Crypto` vs `import "Crypto"`)

turbolent (commented):
Great idea! Only have one concern regarding location clashes

runtime/coverage.go
m-Peter: They might be legal, but they cannot be utilized by 3rd party tools that consume code coverage reports in LCOV format. For example, Codecov cannot translate an address location to its source file. And as such, it cannot provide visualizations. I think all of these tools need filepaths, in order to open them and display their source code.

runtime/coverage.go
```diff
@@ -120,6 +120,9 @@ type CoverageReport struct {
 	// This filter can be used to inject custom logic on
 	// each location/program inspection.
 	LocationFilter LocationFilter `json:"-"`
+	// Contains a mapping with human-friendly names for
+	// locations. This could be the filepath of a location.
```
SupunS, line 124: I guess this mapping holds the "source path" rather than a "human-friendly name", against the location?
Maybe better to put it that way in the description?

runtime/coverage.go
SupunS, line 558: maybe `sourcePathForLocation` is more intuitive?

runtime/coverage.go
```diff
@@ -120,6 +120,9 @@ type CoverageReport struct {
...
+	LocationMappings map[string]string `json:"-"`
```
SupunS, line 125: Just some general Qs on the `CoverageReport` struct:
- Do we need to export these fields?
- I think this coverage report struct is not marshaled as JSON as-is, correct? If so, we could remove these JSON marshaling tags. i.e: `json:"-"`

SupunS (commented):
👍

runtime/coverage.go
m-Peter: As far as I understand, we cannot use the `ID`, because the `flow.json` config file contains only contract names, e.g.:
```json
{
  "contracts": {
    "ApprovalVoting": {
      "source": "contracts/ApprovalVoting.cdc",
      "aliases": {
        "testing": "0000000000000007"
      }
    }
  }
}
```
So, if we are dealing with address, string or identifier location, the `location.ID()` would return:
- `A.0000000000000006.ApprovalVoting`
- `S.ApprovalVoting`
- `I.ApprovalVoting`

but the `locationMappings` field, only contains the `"ApprovalVoting"` key. That's why to strip parts of it, and get only the contract name from a location.

runtime/coverage.go
m-Peter: Updated in https://github.com/onflow/cadence/pull/2859/commits/4871d50d108d715a582746143ff09d72e78c525b

runtime/coverage.go
m-Peter: Good point :+1: 
Updated in https://github.com/onflow/cadence/pull/2859/commits/4871d50d108d715a582746143ff09d72e78c525b

runtime/coverage.go
m-Peter: Good catch :+1: 
Updated in https://github.com/onflow/cadence/pull/2859/commits/4871d50d108d715a582746143ff09d72e78c525b

m-Peter: I have left the necessary fields because the coverage report struct is indeed marshaled as JSON in https://github.com/onflow/cadence/blob/master/runtime/coverage.go#L433-L441

runtime/coverage.go
m-Peter: I have added a test case for identifier locations, which wasn't handled previously: https://github.com/onflow/cadence/commit/4871d50d108d715a582746143ff09d72e78c525b

runtime/coverage.go
SupunS: > I have not touched the necessary fields because the coverage report struct is marshaled as JSON in

But aren't all the fields of `CoverageReport` have the "ignore-tag" (i.e: `json:"-"`), and thus all the fields are skipped by the marshaller?

m-Peter: Ah, you are right :clap: 
For some reason I was using a redundant type alias for `CoverageReport`, which was also marshaled.
Fixed in https://github.com/onflow/cadence/pull/2859/commits/6b9d9f2fcd558ec316794b345bb8d9a9aadd0525

SupunS: nice! 👍

runtime/coverage.go
SupunS: I guess a different way to phrase the same question is, if the flow.json is something like below:
```json
{
  "contracts": {
    "ApprovalVoting": {
      "source": "contracts/ApprovalVoting.cdc",
      "aliases": {
        "testing": "0000000000000007"
      }
    }
  }
}
```

then should all of the following import formats work (are they all refer to the same `ApprovalVoting` in flow.json), with respect to the coverage ?
- `import ApprovalVoting`
- `import "ApprovalVoting"`
- `import ApprovalVoting from 0000000000000007`

m-Peter: Yes, they work indeed, and I have added test cases for all three location types:
- [IdentifierLocation](https://github.com/onflow/cadence/pull/2859/files#diff-e3df797dc0e65b26af99a2311291d53b5202848e6ac9b83cdfab3009d0723c3cR707-R716) - `import ApprovalVoting`
- [StringLocation](https://github.com/onflow/cadence/pull/2859/files#diff-e3df797dc0e65b26af99a2311291d53b5202848e6ac9b83cdfab3009d0723c3cR679-R688) - `import "ApprovalVoting"`
- [AddressLocation](https://github.com/onflow/cadence/pull/2859/files#diff-e3df797dc0e65b26af99a2311291d53b5202848e6ac9b83cdfab3009d0723c3cR641-R660) - `import ApprovalVoting from 0000000000000007`

Let me know if I understood the question correctly :sweat_smile:

SupunS: hmm, not sure if `import ApprovalVoting` should actually work (does importing like that actually work in the emulator? i.e: just simple importing, without code coverage or anything). Identifier imports are for built-in contracts. e.g: `import Crypto` or `import Test`.

SupunS: If the emulator already supports identifier imports to be resolved from `flow.json` (like above), then this looks good to me. But might be good to double-check.

m-Peter: You are indeed right :+1: 
The emulator does not support importing custom identifier locations. But identifier imports for built-in contracts, do produce code coverage (https://github.com/onflow/cadence-tools/blob/master/test/test_runner.go#L454-L456), that's why these locations are excluded in the `cadence-tools/test` repository.
I can remove the case for handling identifier locations, but it doesn't really have any effect at this point.

SupunS (approved):
👏

runtime/coverage.go
SupunS: ah ok, makes sense 👍 
Thanks for the clarification

turbolent (approved):

turbolent:
Nice!

PR #2866, opened 2023-10-12
Restore old parenthesized type parsing behavior

dsainati1:
ad0cbfab973fa2203bd8e1afbaf8c2322d060cf5 introduced a regression by adding support for parenthesized types, but the type checker makes a number of assumptions that don't account for the possibility that types are arbitrarily nested like this change allowed them to be. In particular, reference field access assumes that references are never double-nested (it's not possible to have a reference to a reference), because prior to this point it was never possible to express such a type. This restores the old behavior, whereby only function types can be wrapped in parens. 

This also makes a minor breaking change by banning `& &T` doubly nested reference types separated by a space. 

Thanks to @oebeling for the report

______

### DIFF
--- a/runtime/parser/type_test.go
+++ b/runtime/parser/type_test.go
@@ -481,7 +481,7 @@ func TestParseReferenceType(t *testing.T) {
 		utils.AssertEqualWithDiff(t,
 			[]error{
 				&SyntaxError{
-					Message: "unexpected non-function parenthesized type",
+					Message: "expected 'fun', got '&'",
 					Pos:     ast.Position{Offset: 2, Line: 1, Column: 2},
 				},
 			},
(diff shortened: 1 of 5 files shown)


SupunS (approved):

runtime/parser/type_test.go
```diff
@@ -472,6 +472,70 @@ func TestParseReferenceType(t *testing.T) {
...
+
+	t.Run("double nested reference", func(t *testing.T) {
+
+		t.Parallel()
+
+		_, errs := testParseType("&(&S)")
+		utils.AssertEqualWithDiff(t,
+			[]error{
+				&SyntaxError{
+					Message: "expected 'fun', got '&'",
```
turbolent, line 484: This error message seems surprising. Why is `fun` in particular expected here?

turbolent (commented):
I don't really think that that commit was a regression, but rather just a generalization.

Maybe instead of reverting the commit, it might be better to reject references to references specifically. Also, if it is done in the type checker, it might be possible to provide better error messages.

PR #2876, opened 2023-10-17
Add support iterating references to iterables

### LINKED ISSUE #2784: Support using references in loops
### Issue to be solved

Currently, for-each loop only supports arrays, and doesn't support array references.
i.e:
```cadence
var arrayRef: &[String] = ...

for element in arrayRef {    // Error

}
```

However, with the mutability changes, given fields can now return references, it would be useful to be able to use array references in loops as well.

### Suggested Solution

Allow using array references in for loops.
```cadence
for v in arrayRef {}
```
For an array of type `[T]`, the type of the loop-variable `v` would be:
  - The concrete type `T`, if the array element type `T` is a primitive.
  - A reference type `&T`, if the array element is a container type.

This is to be consistent with the member access semantics: https://github.com/onflow/flips/blob/120b8aa473e5040db2b99034d2118fb78d053cdf/cadence/20230517-member-access-semnatics.md

SupunS:
Closes #2784

## Description

Allow using array references in for loops.
```cadence
for v in arrayRef {}
```

For an array of type `[T]`, the type of the loop-variable `v` would be:

- The concrete type `T`, if the array element type `T` is a primitive.
- A reference type `&T`, if the array element is a container type.

This is to be consistent with the member access semantics: https://github.com/onflow/flips/blob/120b8aa473e5040db2b99034d2118fb78d053cdf/cadence/20230517-member-access-semnatics.md

______

### DIFF
--- a/runtime/interpreter/interpreter_statement.go
+++ b/runtime/interpreter/interpreter_statement.go
@@ -338,7 +338,7 @@ func (interpreter *Interpreter) VisitForStatement(statement *ast.ForStatement) S
 		panic(errors.NewUnreachableError())
 	}
 
-	iterator := iterable.Iterator(interpreter)
+	iterator := iterable.Iterator(interpreter, locationRange)
 
 	var index IntValue
 	if statement.Index != nil {
--- a/runtime/interpreter/interpreter_statement.go
+++ b/runtime/interpreter/interpreter_statement.go
@@ -313,7 +313,7 @@ func (interpreter *Interpreter) VisitWhileStatement(statement *ast.WhileStatemen
 
 var intOne = NewUnmeteredIntValueFromInt64(1)
 
-func (interpreter *Interpreter) VisitForStatement(statement *ast.ForStatement) StatementResult {
+func (interpreter *Interpreter) VisitForStatement(statement *ast.ForStatement) (result StatementResult) {
 
 	interpreter.activations.PushNewWithCurrent()
 	defer interpreter.activations.Pop()
@@ -339,28 +339,35 @@ func (interpreter *Interpreter) VisitForStatement(statement *ast.ForStatement) S
 	}
 
 	forStmtTypes := interpreter.Program.Elaboration.ForStatementType(statement)
-	iterator := iterable.Iterator(interpreter, forStmtTypes.ValueVariableType, locationRange)
 
 	var index IntValue
 	if statement.Index != nil {
 		index = NewIntValueFromInt64(interpreter, 0)
 	}
 
-	for {
-		value := iterator.Next(interpreter)
-		if value == nil {
-			return nil
-		}
-
+	executeBody := func(value Value) (resume bool) {
 		statementResult, done := interpreter.visitForStatementBody(statement, index, value)
 		if done {
-			return statementResult
+			result = statementResult
 		}
+
+		resume = !done
 
 		if statement.Index != nil {
 			index = index.Plus(interpreter, intOne, locationRange).(IntValue)
 		}
+
+		return
 	}
+
+	iterable.ForEach(
+		interpreter,
+		forStmtTypes.ValueVariableType,
+		executeBody,
+		locationRange,
+	)
+
+	return
 }
 
 func (interpreter *Interpreter) visitForStatementBody(
--- a/runtime/interpreter/interpreter_statement.go
+++ b/runtime/interpreter/interpreter_statement.go
@@ -338,7 +338,8 @@ func (interpreter *Interpreter) VisitForStatement(statement *ast.ForStatement) S
 		panic(errors.NewUnreachableError())
 	}
 
-	iterator := iterable.Iterator(interpreter, locationRange)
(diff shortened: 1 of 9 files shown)


runtime/tests/checker/for_test.go
```diff
@@ -273,3 +274,122 @@ func TestCheckInvalidForShadowing(t *testing.T) {
 
 	assert.IsType(t, &sema.RedeclarationError{}, errs[0])
 }
+
+func TestCheckReferencesInForLoop(t *testing.T) {
+
```
dsainati1, line 279: I see earlier that the references to the elements are created with unauthorized access, but can we add a couple tests for that behavior too when the outer array reference is entitled?

runtime/sema/check_for.go
```diff
@@ -123,3 +99,68 @@ func (checker *Checker) VisitForStatement(statement *ast.ForStatement) (_ struct
 
 	return
 }
+
+func (checker *Checker) loopVariableType(valueType Type, hasPosition ast.HasPosition) Type {
```
turbolent, line 103: How does this differ from the type checking result of indexing? Maybe the existing indexing checking can be re-used here for the loop variable checking?

runtime/sema/check_for.go
```diff
@@ -123,3 +99,68 @@ func (checker *Checker) VisitForStatement(statement *ast.ForStatement) (_ struct
...
+}
+
+func (checker *Checker) iterableElementType(valueType Type, hasPosition ast.HasPosition) Type {
+	if arrayType, ok := valueType.(ArrayType); ok {
+		return arrayType.ElementType(false)
+	}
+
+	if valueType == StringType {
+		return CharacterType
+	}
```
turbolent, line 155: Maybe this can be improved/generalized to just check for `ValueIndexableType`? That way we implementing the interface for a type automatically enables both indexing and looping support?

turbolent (commented):
Nice!

runtime/sema/check_for.go
SupunS: Same as above, `ValueIndexableType` includes non-iterable types (e.g: Dictionary), so it's not usable here.
Once/If we add the support for dictionaries in for loops, we can revisit and refactor this.

runtime/tests/checker/for_test.go
SupunS: Added: c659d79c67e50a58a2b1f981488c2be00858d77f

SupunS:
I just realized this need to track resource references.

runtime/interpreter/interpreter_statement.go
```diff
@@ -338,28 +338,36 @@ func (interpreter *Interpreter) VisitForStatement(statement *ast.ForStatement) S
...
+		resume = !done
+
 		if statement.Index != nil {
 			index = index.Plus(interpreter, intOne, locationRange).(IntValue)
 		}
+
+		return
 	}
+
+	iterable.ForEach(
```
SupunS, line 363: Switched to `ForEach` as opposed to directly using the iterator here, since we need to track resource references and prevent container mutation while iterating (https://github.com/dapperlabs/cadence-internal/pull/119). 
 
`Array.Iterate(inter, f)` already implements this, so switching to foreach (which delegates to Array.Iterate) allows re-using the same.

runtime/interpreter/value.go
```diff
@@ -1586,6 +1592,25 @@ func (v *StringValue) Iterator(_ *Interpreter) ValueIterator {
...
 
+func (v *StringValue) ForEach(
+	interpreter *Interpreter,
+	_ sema.Type,
+	function func(value Value) (resume bool),
+	_ LocationRange,
+) {
+	iterator := v.Iterator(interpreter)
+	for {
+		value := iterator.Next(interpreter)
```
dsainati1, line 1603: Does `StringValue` not have an iterate function like array does?

runtime/interpreter/value.go
```diff
@@ -20196,6 +20231,62 @@ func (*StorageReferenceValue) DeepRemove(_ *Interpreter) {
...
+
+	updatedFunction := func(value Value) (resume bool) {
+		if isResultReference {
+			value = interpreter.getReferenceValue(value, elementType)
+		}
+
+		return function(value)
+	}
+
+	referencedElementType := elementType
```
dsainati1, line 20277: can we just change the name of this in the parameters instead of assigning a new variable?

runtime/interpreter/value.go
SupunS: Unfortunately no, only `ArrayValue` and `DictionaryValue` currently have iterate functions

runtime/interpreter/value.go
SupunS: `referencedElementType` gets re-assigned again below, inside the condition (`if`-statement). So these two actually mean two different things at this point.
i.e: `elementType` is the element type of the reference. `referencedElementType` is the element type of the referenced value.

dsainati1 (approved):

turbolent (approved):
Nice 👌

PR #2877, opened 2023-10-17
Entitlement mapping escape fixes

dsainati1:
@oebeling reported a couple of cases where entitlement mappings could "escape" (a la generics) their defining contexts. This cleans up those issues with a few related changes. 

______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -2115,6 +2115,14 @@ func (interpreter *Interpreter) convert(value Value, valueType, targetType sema.
 			// transferring a reference at runtime does not change its entitlements; this is so that an upcast reference
 			// can later be downcast back to its original entitlement set
 
+			// check defensively that we never create a runtime mapped entitlement value
+			if _, isMappedAuth := unwrappedTargetType.Authorization.(*sema.EntitlementMapAccess); isMappedAuth {
+				panic(UnexpectedMappedEntitlementError{
+					Type:          unwrappedTargetType,
+					LocationRange: locationRange,
+				})
+			}
+
 			switch ref := value.(type) {
 			case *EphemeralReferenceValue:
 				return NewEphemeralReferenceValue(
--- a/runtime/sema/check_function.go
+++ b/runtime/sema/check_function.go
@@ -187,20 +187,23 @@ func (checker *Checker) checkFunction(
 			functionActivation.InitializationInfo = initializationInfo
 
 			if functionBlock != nil {
-				oldMappedAccess := checker.entitlementMappingInScope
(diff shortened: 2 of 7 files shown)


runtime/interpreter/interpreter.go
```diff
@@ -2115,6 +2115,14 @@ func (interpreter *Interpreter) convert(value Value, valueType, targetType sema.
 			// transferring a reference at runtime does not change its entitlements; this is so that an upcast reference
 			// can later be downcast back to its original entitlement set
 
+			// check defensively that we never create a runtime mapped entitlement value
+			if _, isMappedAuth := unwrappedTargetType.Authorization.(*sema.EntitlementMapAccess); isMappedAuth {
```
dsainati1, line 2119: We should never create an entitlement value at runtime whose authorization is a map, so we assert that the target reference type here is never an entitlement map.

runtime/sema/check_function.go
```diff
@@ -187,6 +187,7 @@ func (checker *Checker) checkFunction(
 			functionActivation.InitializationInfo = initializationInfo
 
 			if functionBlock != nil {
+				oldMappedAccess := checker.entitlementMappingInScope
```
turbolent, line 190: Maybe move the code starting here and ending with the new reset (L203) into an immediately invoked function, and move the reset line (203) below this line, as a `defer` statement. That makes the scope more obvious and keeps setting and resetting closer together.

runtime/sema/checker.go
```diff
@@ -1256,20 +1256,22 @@ func (checker *Checker) functionType(
 	parameterList *ast.ParameterList,
 	returnTypeAnnotation *ast.TypeAnnotation,
 ) *FunctionType {
+
+	oldMappedAccess := checker.entitlementMappingInScope
```
turbolent, line 1260: Same as above

turbolent (commented):
Nice!

SupunS (approved):

runtime/interpreter/interpreter.go
```diff
@@ -2115,6 +2115,14 @@ func (interpreter *Interpreter) convert(value Value, valueType, targetType sema.
...
+				panic(UnexpectedMappedEntitlementError{
+					Type:          unwrappedTargetType,
+					LocationRange: locationRange,
+				})
```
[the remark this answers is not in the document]
dsainati1: This is an internal error that we raise if we accidentally would create a runtime entitlement-mapped value. This shouldn't ever be reachable though since we should prevent this statically, so as far as I can tell this should not be exercise-able by tests.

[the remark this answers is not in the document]
SupunS: makes sense 👍

Issue #2879, opened 2023-10-18
Meter computation in new stdlib functions


turbolent:
### Issue to be solved

Some of the new stdlib functions perform non-constant operations, which require computation metering:

- Master
  - [x] #2762
  - [x] #2737
  - [x] #2678
  - [x] #2654
- Master, not merged yet
  - [ ] #2626
- Stable Cadence
  - [x] #2814
  - [x] #2791

### Suggested Solution

Implementations should call `ReportComputation`, potentially with a new (?) `common.ComputationKind`.

Metering might also require "inlining" existing calls to the Go stdlib, so that each iteration of an operation can be metered up-front.

turbolent:
@SupunS can you please have a look at this/next with @darkdrag00nv2 this week? The PRs that got merged into `master` need to get metered before the spork

PR #2880, opened 2023-10-19
Add computation metering to the new stdlib functions

SupunS:
Work towards #2879

## Description

This PR only adds computation metering for the newly added functions:
- https://github.com/onflow/cadence/pull/2762
- https://github.com/onflow/cadence/pull/2737
- https://github.com/onflow/cadence/pull/2678
- https://github.com/onflow/cadence/pull/2654

The remaining functions need to be metered in the stable-cadence branch since it can be a breaking change.
______

### DIFF
--- a/runtime/common/computationkind.go
+++ b/runtime/common/computationkind.go
@@ -57,7 +57,7 @@ const (
 	ComputationKindCreateArrayValue
 	ComputationKindTransferArrayValue
 	ComputationKindDestroyArrayValue
-	_
+	ComputationKindIterateArrayValue
 	_
 	_
 	_
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -2986,9 +2986,6 @@ func (v *ArrayValue) Reverse(
 	count := v.Count()
 	index := count - 1
 
-	// Meter computation for iterating the array.
-	interpreter.ReportComputation(common.ComputationKindIterateArrayValue, uint(v.Count()))
-
 	return NewArrayValueWithIterator(
 		interpreter,
 		v.Type,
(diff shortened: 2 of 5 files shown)


runtime/common/computationkind.go
```diff
@@ -57,7 +57,7 @@ const (
 	ComputationKindCreateArrayValue
 	ComputationKindTransferArrayValue
 	ComputationKindDestroyArrayValue
-	_
+	ComputationKindIterateArrayValue
```
SupunS, line 60: I added a new computation kind here. We could also use `ComputationKindLoop` because this is literally looping through the array. But introduced a new one in case we want to assign different weights to them.

runtime/interpreter/value.go
```diff
@@ -2986,6 +2986,9 @@ func (v *ArrayValue) Reverse(
 	count := v.Count()
 	index := count - 1
 
+	// Meter computation for iterating the array.
+	interpreter.ReportComputation(common.ComputationKindIterateArrayValue, uint(v.Count()))
```
turbolent, line 2990: Metering up-front works, but might be overly restrictive: If the array is large, the up-front metering cost might already exceed the limit.

It might be good to do the metering like for loops, i.e., meter before each iteration step.

turbolent (commented):
Nice! Thanks for adding this so quickly

runtime/interpreter/value.go
SupunS: It doesn't make much difference really, since all elements of the array are loaded at the time of creation, by the atree.
Anyway updated a07342f so if the atree behavior is changed to not load everything, then it'll reflect here.

turbolent (approved):
Nice!

PR #2878, opened 2023-10-18
Allow injection of functions into composite values, refactor PublicKey based on it

turbolent:
Closes #2874 

## Description

The host might want composite values of certain types to have native functions. Add a function `SetCompositeValueFunctionsHandler` to the environment which allows that.

Extend the `CompositeValue.GetMember` implementation, from only linking in interpreted functions, to also querying a handler set in the interpreter configuration.

As the built-in `PublicKey` type has native functions (e.g. `verify`) and is already implemented as a `CompositeValue`, refactor its existing implementation from injecting the native functions at construction time, to using the new injection mechanism added in this PR.

This essentially allows making the `PublicKey` type storable (in a follow-up PR), because the native functions get linked in even after a public key value got loaded from storage. `PublicKey` values are not yet storable exactly because of this missing function injection feature.

______

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #2881, opened 2023-10-19
Allow built-in composite types to be storable

turbolent:
⚠️ Depends on #2878

## Description

- [Allow built-in composite **types** to be storable](https://github.com/onflow/cadence/commit/6681cdaeb6b1e559fc5cd1aae48b20602bcab8c8)
- Allow built-in composite **values** to be storable.

    As discussed in https://discord.com/channels/613813861610684416/1108479699732152503/1164334273684324402, allowing `CompositeValue.Storable` to determine if the value's type is storable, by getting the `CompositeType` for the value, is tricky, as the function is called by atree, and it does not provide access to an `Interpreter` easily.

   [Allow getting root interpreter from interpreter storage](https://github.com/onflow/cadence/commit/c1b61a42f90e944e1dcff9d9de287ebe044b947a), which is ugly and works. If you have any better ideas, please let me know – I don't like this, but couldn't find another easy workable solution. 

    For now this is tech-debt: Eventually in Stable Cadence (which cannot be used, we need this feature on master), we'll have fewer `CompositeValue`s with built-in `CompositeType`s that are non-storable (e.g. currently `AuthAccount`), and we might find a better alternative to this approach later. At least for now this unblocks the work on the EVM integration in flow-go.

    Finally, [allow built-in composite values to be storable](https://github.com/onflow/cadence/commit/02f91f009512966380460ebb2962032db1ef1744), by checking for built-in types, if the composite value's composite type is storable (some are not, e.g. `AuthAccount`). Only check built-in types, because non-built-in composite values/types are always considered storable.

______

### DIFF
--- a/runtime/cmd/decode-state-values/main.go
+++ b/runtime/cmd/decode-state-values/main.go
@@ -228,6 +228,16 @@ func (i interpreterStorage) CheckHealth() error {
 	panic("unexpected CheckHealth call")
 }
 
+func (i interpreterStorage) RootInterpreter() *interpreter.Interpreter {
+	panic("unexpected RootInterpreter call")
+
+}
+
+func (i interpreterStorage) SetRootInterpreter(_ *interpreter.Interpreter) {
+	panic("unexpected SetRootInterpreter call")
+
+}
+
 // load
 
 func load() {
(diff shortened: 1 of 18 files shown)


turbolent:
As mentioned above, this adds some tech debt. After discussing implementation approaches for the Flex contract with @ramtinms and @janezpodhostnik, we realized that the Flex contract does not necessarily need to be a built-in, so this PR is no longer needed

PR #2882, opened 2023-10-19
Integrate with atree inlining and data deduplication

### LINKED ISSUE #2809: Integrate new version of Atree with register inlining and data deduplication
### Issue to be solved

Atree register inlining and encoding size optimizations introduced in https://github.com/onflow/atree/pull/342, etc. required breaking changes to Atree's API, so existing Cadence code needs to be updated to use it.

### Suggested Solution

TODO:

- @fxamacker:  Update existing code in Cadence to use Atree's new API after https://github.com/onflow/atree/pull/342 and related PRs are merged.

- Cadence Team:  Add more tests to Cadence to cover all new edge cases

Adding tests for this requires expertise in Cadence internals.

fxamacker:
Closes #2809 
Updates https://github.com/onflow/atree/issues/292

This commit updates Cadence to use newer version of onflow/atree with register inlining and data deduplication features introduced by https://github.com/onflow/atree/pull/342 and https://github.com/onflow/atree/pull/345.

This PR requires https://github.com/onflow/atree/pull/352.

### Changes

- Updated StorableDecoder callbacks

- Updated StringAtreeValue to implement atree.ComparableStorable

- Updated SomeStorable to implement atree.ContainerStorable

- Updated these structs to implement TypeInfo interface:
  * compositeTypeInfo
  * EmptyTypeInfo
  * VariableSizedStaticType
  * ConstantSizedStaticType
  * DictionaryStaticType

- Updated decodeStorable() to decode atree inlined array, map, and compact map

- Updated to use ReadOnly iterators when possible

- Updated to use NonReadOnly iterators

### TODO

Cadence team :pray: will need to make additional changes:
- [ ] identify and implement more tests (helps prevent data loss)
- [x] fix metering tests broken by integration (I fixed other tests)
- [ ] resolve some TODOs added as code comments
- [x] meeting on Oct 20, 2023 to discuss integration, tests, and code review of this PR etc.  (thanks for attending and asking great questions! 👍)
- [ ] run Cadence smoke tests
- [x] @turbolent add `*Interpreter` and `LocationRange` to some functions in order to use non-readonly `atree.OrderedMap` iterators
- [x] @turbolent check if any storable can contain `atree.Array` and `atree.OrderedMap` and needs to implement `atree.ContainerStorable` interface
______


runtime/interpreter/encode.go
```diff
@@ -1601,6 +1622,21 @@ var _ atree.TypeInfo = compositeTypeInfo{}
...
+
+func (c compositeTypeInfo) Identifier() string {
+	// TODO: maybe improve this to be more efficient, etc.
+	return "composite(" + c.location.ID() + "/" + c.qualifiedIdentifier + ")"
+}
+
+func (c compositeTypeInfo) Copy() atree.TypeInfo {
+	// TODO: is creating a new compositeTypeInfo necessary?
+	// atree needs immutable TypeInfo or a copy of mutable TypeInfo because TypeInfo can be shared.
+	return NewCompositeTypeInfo(nil, c.location, c.qualifiedIdentifier, c.kind)
```
turbolent, line 1637: Can this function just return `c` as-is, given it's a value type? The other `Copy` functions below do just that

runtime/interpreter/encode.go
```diff
@@ -1601,6 +1622,21 @@ var _ atree.TypeInfo = compositeTypeInfo{}
 
 const encodedCompositeTypeInfoLength = 3
 
+func (c compositeTypeInfo) IsComposite() bool {
+	return true
+}
+
+func (c compositeTypeInfo) Identifier() string {
+	// TODO: maybe improve this to be more efficient, etc.
+	return "composite(" + c.location.ID() + "/" + c.qualifiedIdentifier + ")"
```
turbolent, line 1631: Can this maybe just be `c.Location.TypeID(gauge, c.qualifiedIdentifier)`?

runtime/interpreter/encode.go
```diff
@@ -699,6 +701,25 @@ func (s SomeStorable) Encode(e *atree.Encoder) error {
 	return s.Storable.Encode(e)
 }
 
+// EncodeAsElement encodes SomeStorable as
+//
+//	cbor.Tag{
+//			Number: CBORTagSomeValue,
+//			Content: Value(v.Value),
+//	}
+func (s SomeStorable) EncodeAsElement(e *atree.Encoder, inlinedExtraData atree.InlinedExtraData) error {
```
turbolent, line 710: Why is this duplicate of `SomeStorable.Encode` needed?

runtime/interpreter/encode.go
```diff
@@ -1645,4 +1681,16 @@ func (e EmptyTypeInfo) Encode(encoder *cbor.StreamEncoder) error {
 	return encoder.EncodeNil()
 }
 
+func (e EmptyTypeInfo) IsComposite() bool {
+	return false
+}
+
+func (e EmptyTypeInfo) Identifier() string {
+	return "null"
```
turbolent, line 1689: Can this maybe just be an empty string `""` to avoid the overhead of storing the literal `"null`"?

runtime/interpreter/encoding_test.go
```diff
@@ -126,6 +127,24 @@ func testEncodeDecode(t *testing.T, test encodeDecodeTest) {
 	}
 }
 
+// encodeStorable wraps storable.Encode()
+func encodeStorable(storable atree.Storable, encMode cbor.EncMode) ([]byte, error) {
+	var buf bytes.Buffer
```
turbolent, line 132: Would it make sense to pool these buffers, like in atree?

runtime/interpreter/encoding_test.go
```diff
@@ -389,6 +408,8 @@ func TestEncodeDecodeUint64AtreeValue(t *testing.T) {
 
 func TestEncodeDecodeArray(t *testing.T) {
 
+	t.Skip("skipping ArrayValue encoding and decoding test because it involves implementation details in atree repo")
```
turbolent, line 411: Should this test and `TestEncodeDecodeComposite` be adjusted and unskipped before merging the PR / is this a TODO?

runtime/interpreter/statictype.go
```diff
@@ -255,6 +256,21 @@ func NewVariableSizedStaticType(
...
+
+func (t VariableSizedStaticType) Copy() atree.TypeInfo {
+	// TODO: is creating a new VariableSizedStaticType necessary?
+	// atree needs immutable TypeInfo or a copy of mutable TypeInfo because TypeInfo can be shared.
+	return t
+}
+
+func (t VariableSizedStaticType) Identifier() string {
+	// TODO: is ID guaranteed to be unique?
+	return string(t.ID())
```
turbolent, line 271: What is the uniqueness requirement of `Identifier()`? All types that have the same "contents" have the same ID. Is that sufficient? Or does "unique" mean here that every type (here, `VariableSizedStaticType`) gets a unique ID, even for other types that are "equal"?

runtime/interpreter/value.go
```diff
@@ -17868,18 +17874,25 @@ func (v *DictionaryValue) Accept(interpreter *Interpreter, visitor Visitor) {
 }
 
 func (v *DictionaryValue) Iterate(interpreter *Interpreter, f func(key, value Value) (resume bool)) {
+
+	valueComparator := newValueComparator(interpreter, EmptyLocationRange)
+	hashInputProvider := newHashInputProvider(interpreter, EmptyLocationRange)
```
turbolent, line 17879: `DictionaryValue.Iterate` should ideally be passed the proper location range

runtime/interpreter/value.go
```diff
@@ -17907,7 +17920,10 @@ func (i DictionaryIterator) NextKey(gauge common.MemoryGauge) Value {
 }
 
 func (v *DictionaryValue) Iterator() DictionaryIterator {
-	mapIterator, err := v.dictionary.Iterator()
+	// TODO: Is it safe to use ReadOnly Iterator here
+	// because DictionaryIterator is only used to iterate keys?
+	// Maybe rename DictionaryIterator to DictionaryKeyIterator.
+	mapIterator, err := v.dictionary.ReadOnlyIterator()
```
turbolent, line 17926: We need to check where `DictionaryValue.Iterator` is used. The key might still be used to perform a mutating operation. It might be better to be conservative here and assume mutation, i.e. reverting this  switch to a read-only iterator and leaving the TODO, we can optimize later

runtime/interpreter/value.go
```diff
@@ -18043,7 +18059,7 @@ func (v *DictionaryValue) ForEachKey(
 	}
 
 	iterate := func() {
-		err := v.dictionary.IterateKeys(
+		err := v.dictionary.IterateReadOnlyKeys(
```
turbolent, line 18062: Are we sure that the `procedure` passed to `DictionaryValue.ForEachKey` does not mutate? I'd be conservative and switch to a read-only iterator after we ensured it. It might also be good to indicate that in the name then

runtime/interpreter/value.go
```diff
@@ -19517,7 +19536,16 @@ type SomeStorable struct {
 	Storable atree.Storable
 }
 
-var _ atree.Storable = SomeStorable{}
+var _ atree.ContainerStorable = SomeStorable{}
+
+func (s SomeStorable) HasPointer() bool {
```
turbolent, line 19541: What is the purpose of `Storable.HasPointer`?

runtime/interpreter/stringatreevalue.go
```diff
@@ -28,6 +28,7 @@ type StringAtreeValue string
 
 var _ atree.Value = StringAtreeValue("")
 var _ atree.Storable = StringAtreeValue("")
+var _ atree.ComparableStorable = StringAtreeValue("")
```
turbolent, line 31: Is this also needed for `Uint64AtreeValue`?

turbolent (commented):
Great work!

runtime/interpreter/encoding_test.go
fxamacker: > Would it make sense to pool these buffers, like in atree?

It is probably not necessary here because this function is used only in tests.

runtime/interpreter/encoding_test.go
fxamacker: Although we can adjust these two tests, I am not sure if we should.

`TestEncodeDecodeArray` encodes `atree.Array.Storable()`.  Encode function can fail if array is inlined because we can't encode inlined array by itself.  So we can either:
- create a large atree.Array under the hood (larger than max size limit), so `Storable()` returns a `SlabID`  like before, or 
- encode slab containing array and then test encoded bytes.  

Either way, these two approaches depend on implementation details in atree (any encoding changes to atree array/map would break these tests).  These two tests are different from other encoding tests because others test encoding defined in Cadence.

Thoughts?

runtime/interpreter/encode.go
fxamacker: > Why is this duplicate of SomeStorable.Encode needed?

`SomeStorable.EncodeAsElement` and `SomeStorable.Encode` have a different API and are used differently.

```Go
Encode(*Encoder) error

EncodeAsElement(*Encoder, InlinedExtraData) error
```

`EncodeAsElement` is function in `atree.ContainerStorable` interface.  `ContainerStorable` is implemented by storable containing other storables, such as `atree.ArrayDataSlab` and `atree.MapDataSlab`.  Those storables and their child storables may need to be encoded differently when they are an element of parent array/map.  

`SomeStorable` implements `atree.ContainerStorable` as well because it contains another storable and that storable can be `atree.ArrayDataSlab` that needs to be encoded differently.

On the other hand, I don't know of a use case for `SomeStorable.Encode()`.  If `SomeStorable.Encode()` returns `UnreachableError`, no cadence tests fail as result of it. I left it as is for now, and Cadence team can decide which approach makes more sense.

runtime/interpreter/statictype.go
```diff
@@ -255,6 +256,21 @@ func NewVariableSizedStaticType(
...
 }
 
+func (t VariableSizedStaticType) IsComposite() bool {
+	return false
+}
+
+func (t VariableSizedStaticType) Copy() atree.TypeInfo {
+	// TODO: is creating a new VariableSizedStaticType necessary?
+	// atree needs immutable TypeInfo or a copy of mutable TypeInfo because TypeInfo can be shared.
+	return t
```
SupunS, line 266: This might only do a shallow copy (of the element type `t.Type`) since certain types (e.g: RestrictedStaticType) have pointer receivers.

runtime/interpreter/statictype.go
```diff
@@ -314,6 +330,21 @@ func NewConstantSizedStaticType(
...
 }
 
+func (t ConstantSizedStaticType) IsComposite() bool {
+	return false
+}
+
+func (t ConstantSizedStaticType) Copy() atree.TypeInfo {
+	// TODO: is creating a new ConstantSizedStaticType necessary?
+	// atree needs immutable TypeInfo or a copy of mutable TypeInfo because TypeInfo can be shared.
+	return t
```
SupunS, line 340: Same as above: this might only do a shallow-copy since certain types (e.g: `RestrictedStaticType`) have pointer receivers.

runtime/interpreter/stringatreevalue.go
fxamacker: Unless I'm mistaken, it probably isn't needed because:
- `atree.ComparableStorable` is only needed for `atree.OrderedMap` keys in Cadence `CompositeValue`
- `StringAtreeValue` seems to be the only key type used by `CompositeValue`

runtime/interpreter/statictype.go
fxamacker: > All types that have the same "contents" have the same ID. Is that sufficient? Or does "unique" mean here that every type (here, VariableSizedStaticType) gets a unique ID, even for other types that are "equal"?

Different types must have different identifiers.  For example, `VariableSizedStaticType` and `ConstantSizedStaticType` must have different identifiers, even though "contents" are the same. This is used by atree for deduplication during encoding.

runtime/interpreter/statictype.go
SupunS: Does the `TypeInfo` returned by this method get mutated by atree?

runtime/interpreter/value.go
fxamacker: Hmm, would we need to modify API to add `LocationRange`?

It is currently:
```Go
func (v *DictionaryValue) Iterate(interpreter *Interpreter, f func(key, value Value) (resume bool))
```

runtime/interpreter/value.go
fxamacker: Make sense.  However, it seems we may need to modify `Iterator()` API to add these new parameters for non-readonly iterator:

```Go
valueComparator := newValueComparator(interpreter, EmptyLocationRange)
hashInputProvider := newHashInputProvider(interpreter, EmptyLocationRange)
```

Current `Iterator` API:
```Go
func (v *DictionaryValue) Iterator() DictionaryIterator {
```

runtime/interpreter/value.go
fxamacker: > Are we sure that the procedure passed to DictionaryValue.ForEachKey does not mutate?

Hmm, do you mean the procedure might mutate map value by iterated key?  

Using a non-readonly vs readonly iterator probably wouldn't make a difference here because mutation through iterator is done by callback in returned map value.  Since procedure only processes map key, mutating map value is not possible by callback.  Actually, atree only has readonly iterators for map keys for the same reason.

Thoughts?

runtime/interpreter/value.go
fxamacker: `ContainerStorable.HasPointer` is used to set a flag bit during encoding if encoded slab refers to other slabs.

`SomeStorable.Storable` might contain reference to another slab.

runtime/interpreter/statictype.go
fxamacker: > Does the TypeInfo returned by this method get mutated by atree?

No, atree doesn't mutate `TypeInfo`.  

`Copy` is used to prevent mutation of shared `TypeInfo` in case it is mutated by Cadence.  However, if `TypeInfo` is immutable, there is no reason to create a new object.    

@SupunS Can TypeInfo be mutated by Cadence?

runtime/interpreter/encode.go
turbolent: It's not ideal. Theoretically we need to meter all allocations. We've seen this issue of atree calling out to Cadence without providing e.g. access to a necessary memory gauge or interpreter a few times before, so I think we should maybe discuss how we can to address this in general (probably as a follow-up and not part of this PR)

runtime/interpreter/encode.go
turbolent: Thank you for the great explanation. Could you please add it to the function as a docstring on the atree side? See https://github.com/onflow/atree/pull/352/files#r1366193018

runtime/interpreter/encoding_test.go
turbolent: Ah, right, I missed that this was in tests 👍

runtime/interpreter/value.go
turbolent: yes, `DictionaryValue.Iterate` signature should get a new parameter `locationRange LocationRange`, similar to many of the other functions of `Value`s. It might still not be available at each use-site, but we should at least try to provide the right location where available.

runtime/interpreter/value.go
turbolent: Yeah, the signature of `DictionaryValue.Iterator` should be changed to include the `*Interpreter` and the `LocationRange`.

runtime/interpreter/value.go
turbolent: Thanks for the explanation! 

Could you please add a comment/docstring to the function on the atree side? See https://github.com/onflow/atree/pull/352#discussion_r1366192016

runtime/interpreter/stringatreevalue.go
turbolent: `StringAtreeValue` is also used for "non-`interpreter.Value`" storage, i.e. for the atree ordered maps used for account storage, and `Uint64AtreeValue` is only used for those.

For example, saving a value to `/storage/foo` actually is implemented as storing a value inside of the account storage's ordered map for the `storage` domain, under key `StringAtreeValue("foo")`

runtime/interpreter/statictype.go
SupunS: Ah ok, thanks for the explanation.
TypeInfo is not mutated by cadence. So I guess then this should be sufficient 👍

runtime/interpreter/encode.go
j1010001: If we can do that as a followup let's please do that.

runtime/cmd/decode-state-values/main.go
```diff
@@ -136,6 +141,11 @@ type slabStorage struct{}
 
 var _ atree.SlabStorage = &slabStorage{}
 
+func (s *slabStorage) RetrieveIfLoaded(atree.SlabID) atree.Slab {
+	// RetrieveIfLoaded() is used for loaded resource tracking.  So it isn't needed here.
+	panic(runtimeErr.NewUnreachableError())
```
ramtinms, line 146: Is this a fatal error that we have to stop the execution node, or it might mean that user tx has to fail and move on?

runtime/interpreter/value.go
turbolent: Done in 5e51834ec43d9a2033115015ca3bc151595e05bf

runtime/interpreter/value.go
turbolent: @fxamacker As you mentioned above, but I didn't understand at the time, `DictionaryIterator` is only used for key iteration at the moment. Refactored the naming and removed the TODO in a93c048de48505a6e09a004f3487042233d6d172

runtime/cmd/decode-state-values/main.go
turbolent: `UnreachableError` is an implementation error. Still, we recover from implementation errors gracefully and abort transactions ("treat them like user errors"), so this won't bring down nodes

runtime/interpreter/encode.go
turbolent: Improved this in da81afe818f6f126d3b37e11b003bf8af8db9626, using the un-metered approach for now

runtime/interpreter/stringatreevalue.go
fxamacker: This was discussed in today's meeting and it isn't for `Uint64AtreeValue` needed as mentioned above.

fxamacker:
I found an edge case on Sunday related to `cadence.SomeStorable` (if it contains inlined array/map).

Several alternative ways to fix it were explored/discussed today and we (Bastian and I) decided to proceed with the initial fix I implemented and mentioned this morning.

Rather than opening new PR as mentioned in today's async, I'll just push a commit to this PR after adding more tests, etc.

runtime/interpreter/decode.go
```diff
@@ -865,6 +868,59 @@ func (d StorableDecoder) decodeSome() (SomeStorable, error) {
...
+
+	nestedLevels, err := d.decoder.DecodeUint64()
+	if err != nil {
+		return SomeStorable{}, errors.NewUnexpectedError(
+			"invalid nested levels for some value with nested levels encoding: %w",
+			err,
+		)
+	}
+
+	if nestedLevels == 1 {
```
turbolent, line 895:
```suggestion
	if nestedLevels <= 1 {
```

runtime/interpreter/decode.go
```diff
@@ -865,6 +868,59 @@ func (d StorableDecoder) decodeSome() (SomeStorable, error) {
...
+			"invalid innermost value for some value with nested levels encoding: %w",
+			err,
+		)
+	}
+
+	storable := SomeStorable{
+		gauge:    d.memoryGauge,
+		Storable: innermostStorable,
+	}
+	for i := 1; i < int(nestedLevels); i++ {
```
turbolent, line 914:
```suggestion
	for i := uint64(1); i < nestedLevels; i++ {
```

runtime/interpreter/encoding_test.go
```diff
@@ -2895,6 +3288,49 @@ func TestEncodeDecodeSomeValue(t *testing.T) {
...
+					// nested levels: 2
+					0x02,
+					// value
+					0xd8, atree.CBORTagSlabID,
+					// storage ID
+					0x50, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x42, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x1,
+				},
+			},
+		)
+	})
```
turbolent, line 3333: Can you please also add a test for the case where there are multiple "parts" of nest `Some`s, e.g. `Some(Some([Some(Some(Some(1)))]))` (2 Some, Array, 3 Some), to test that `getInnermostStorable` does not traverse returns the array and not the integer.

runtime/interpreter/value.go
```diff
@@ -19443,26 +19443,64 @@ func (v *SomeValue) Storable(
...
+	// No need to call maybeLargeImmutableStorable() here for SomeStorable because:
+	// - encoded SomeStorable size = someStorableEncodedPrefixSize + innermost storable size
+	// - innermost storable size < maxInlineSize - someStorableEncodedPrefixSize
+	return SomeStorable{
+		Storable: v.valueStorable,
+	}, nil
+}
+
+// getInnermostValue returns innermost value of SomeValue and nested levels.
+// For example, SomeValue{true} has innermost value true, and nested levels 1.
```
turbolent, line 19491:
```suggestion
// For example, `SomeValue{true}` has innermost value `true` and nested levels 1,
// and `SomeValue{SomeValue{1}}` has innermost value `1`and nested levels 2.
```

runtime/interpreter/value.go
```diff
@@ -19443,26 +19443,64 @@ func (v *SomeValue) Storable(
...
+func (v *SomeValue) getInnermostValue() (atree.Value, uint64) {
```
turbolent, line 19492: Maybe rename the function to indicate that this function only gets the innermost value _as long as there are `SomeValue`s_ / traverses `SomeValue`s, and for `Some(Some([Some(Some(Some(1)))])) (2 Some, Array, 3 Some), the array, and not the integer `1`

runtime/interpreter/value.go
```diff
@@ -19586,8 +19624,32 @@ func (s SomeStorable) HasPointer() bool {
...
+
 func (s SomeStorable) ByteSize() uint32 {
-	return cborTagSize + s.Storable.ByteSize()
+	innermostStorable, nestedLevels := s.getInnermostStorable()
+	return getSomeStorableEncodedPrefixSize(nestedLevels) + innermostStorable.ByteSize()
+}
+
+// getInnermostStorable returns innermost storable of SomeStorable and nested levels.
+// For example, SomeStorable{true} has innermost storable true, and nested levels 1.
+func (s SomeStorable) getInnermostStorable() (atree.Storable, uint64) {
```
turbolent, line 19641: Same as above: Maybe add another example and clarify the function naming, it could be misinterpreted.

turbolent (commented):
Reviewed [aa087cf0b697711822031aedc8ec000dabf580b4](https://github.com/onflow/cadence/commit/aa087cf0b697711822031aedc8ec000dabf580b4) specifically, LGTM, nice!

turbolent:
With f01949a7b7fa4c08a4f156f061152ec5e600c7ff, it becomes viable to run the smoke tests with validation enabled:

```sh
go test -v  -count=100 -run '^TestRandom' ./runtime/tests/interpreter/... -runSmokeTests=true -validateAtree=true
```

This seems to produce "slab was not reachable from leaves" errors now. Here are a few of them: https://gist.github.com/turbolent/5b6b75b5af4c6e012fb055267748ea01

fxamacker:
> With [f01949a](https://github.com/onflow/cadence/commit/f01949a7b7fa4c08a4f156f061152ec5e600c7ff), it becomes viable to run the smoke tests with validation enabled:
> 
> ```shell
> go test -v  -count=100 -run '^TestRandom' ./runtime/tests/interpreter/... -runSmokeTests=true -validateAtree=true
> ```
> 
> This seems to produce "slab was not reachable from leaves" errors now. Here are a few of them: https://gist.github.com/turbolent/5b6b75b5af4c6e012fb055267748ea01

Great work Bastian! :+1: The validation code in atree called by Cadence smoke tests was updated this summer for atree inlining so I'll check to see if this might be a validation bug. A brief look yesterday into atree didn't identify anything obvious yet. I'll resume looking into this on Monday.

fxamacker:
@turbolent 

It looks like the issue reported by Cadence smoke test is a false alarm.  It was caused by failed storage health check in the middle of `PopIterate` removing container element when removed element is stored in another slab (such as child composite).

Since atree `PopIterate` is designed to iterate and remove all elements, elements (slice) under the hood isn't reset one by one during iteration.  Instead, elements are reset after iteration is complete. 

```Go
func (a *ArrayDataSlab) PopIterate(_ SlabStorage, fn ArrayPopIterationFunc) error {

	// Iterate and reset elements backwards
	for i := len(a.elements) - 1; i >= 0; i-- {
		fn(a.elements[i])
	}

	// Reset data slab
	a.elements = nil
	a.header.count = 0
	a.header.size = a.getPrefixSize()

	return nil
}
```

However, during `PopIterate`, storage health check can fail if removed element is in another slab and that slab is removed from storage in Cadence `DeepRemove`.  This is because removed element slab is still referenced in parent and parent isn't reset until `PopIterate` is finished.

Two different ways to resolve this:
* On way (if feasible) is to change Cadence to skip storage health check in `DeepRemove` at nested child and only perform it at root.
* Another way is to change atree `PopIterate`to reset element slice one by one.  However, this isn't optimal because `PopIterate` is designed to pop and remove all elements.

Thoughts?

fxamacker:
@turbolent 

I re-reviewed non-test code in this PR.  Thanks so much for adding extra tests, making Cadence smoke tests viable with validation enabled, running smoke tests, and non-cursory PR reviews!  :raised_hands:

turbolent:
After running the smoke tests for a bit longer after merging #2920, I ran into another validation error: https://gist.github.com/turbolent/6e470ec342eb3e80ecff697a039b93cb. Maybe this error is still for the same problem?

fxamacker:
Hey @turbolent Thanks for reporting this! :raised_hands:

I've been looking into many slabs to identify the problem and have new info.

> After running the smoke tests for a bit longer after merging #2920, I ran into another validation error: https://gist.github.com/turbolent/6e470ec342eb3e80ecff697a039b93cb. Maybe this error is still for the same problem?

Yes, so far, this new smoke test error looks like same issue already resolved by #2920.

However, trying to create small reproducer for it led to discovery of unrelated issue (not caused by this PR or atree inlining).  If confirmed to be a real issue instead of a testing mistake, I'll open PR.

fxamacker:
Hey @turbolent I have another update.

>> After running the smoke tests for a bit longer after merging #2920, I ran into another validation error: https://gist.github.com/turbolent/6e470ec342eb3e80ecff697a039b93cb. Maybe this error is still for the same problem?
>
>Yes, so far, this new smoke test error looks like same issue already resolved by #2920.
>
>However, trying to create small reproducer for it led to discovery of unrelated issue (not caused by this PR or atree inlining).  If confirmed to be a real issue instead of a testing mistake, I'll open PR.

Thanks again for answering a lot of questions about Cadence usage of Atree yesterday when I thought I found a different problem!

Both the Cadence smoke test error you reported and a different issue I found turned out to be false alarms. :tada: 

I added these to the PR:

- https://github.com/onflow/cadence/pull/2882/commits/b98a84ec7cdb098c47dc5d9425c924d2c56491b3 adds reproducer and test case for smoke test false alarm by reducing many slabs into a smaller subset.

- https://github.com/onflow/cadence/pull/2882/commits/1ecf35ce2d16f344d653e41deb6534409dd0a0e4 improves validation by adding a new parameter `atRoot` to `Transfer()`.  So Atree storage is validated only if both `remove` and `atRoot` parameters are true in `ArrayValue.Transfer()`, `DictionaryValue.Transfer()`, and `CompositeValue.Transfer()`.

PTAL :pray:

runtime/interpreter/interpreter.go
```diff
@@ -4048,6 +4050,7 @@ func (interpreter *Interpreter) authAccountReadFunction(addressValue AddressValu
 				false,
 				nil,
 				nil,
+				false,
```
turbolent, line 4053: At root here:

```suggestion
				true,
```

runtime/interpreter/value.go
```diff
@@ -1906,6 +1923,7 @@ func (v *ArrayValue) Concat(interpreter *Interpreter, locationRange LocationRang
 				false,
 				nil,
 				nil,
+				false,
```
turbolent, line 1926:
```suggestion
				true,
```

runtime/interpreter/value.go
```diff
@@ -2978,6 +3021,7 @@ func (v *ArrayValue) Slice(
...
```
turbolent, line 3024:
```suggestion
				true,
```

runtime/interpreter/value.go
```diff
@@ -3010,6 +3054,7 @@ func (v *ArrayValue) Reverse(
...
```
turbolent, line 3057:
```suggestion
				true,
```

runtime/interpreter/value.go
```diff
@@ -3083,6 +3128,7 @@ func (v *ArrayValue) Filter(
...
```
turbolent, line 3131:
```suggestion
				true,
```

runtime/interpreter/value.go
```diff
@@ -3162,6 +3208,7 @@ func (v *ArrayValue) Map(
...
```
turbolent, line 3211:
```suggestion
				true,
```

runtime/interpreter/value.go
```diff
@@ -18274,13 +18402,17 @@ func (v *DictionaryValue) GetMember(
 						false,
 						nil,
 						nil,
+						false,
```
turbolent, line 18405:
```suggestion
						true,
```

runtime/interpreter/value.go
```diff
@@ -18308,6 +18440,7 @@ func (v *DictionaryValue) GetMember(
...
```
turbolent, line 18443:
```suggestion
						true,
```

runtime/stdlib/account.go
```diff
@@ -1099,6 +1100,7 @@ func newAuthAccountInboxUnpublishFunction(
 				true,
 				nil,
 				nil,
+				false,
```
turbolent, line 1103:
```suggestion
				true,
```

runtime/stdlib/account.go
```diff
@@ -1185,6 +1187,7 @@ func newAuthAccountInboxClaimFunction(
...
```
turbolent, line 1190:
```suggestion
				true,
```

runtime/stdlib/account.go
```diff
@@ -3236,6 +3240,7 @@ func newAuthAccountCapabilitiesUnpublishFunction(
...
```
turbolent, line 3243:
```suggestion
				true,
```

turbolent (commented):
Thank you for looking into the issue @fxamacker and great work figuring out what the problem is! 

Adding `atRoot` to `Transfer` makes sense 👍 

I had a look through 1ecf35ce2d16f344d653e41deb6534409dd0a0e4 and it looks like there several cases where `atRoot` should be true. Not sure if those cases were intentionally false. I think I caught all of them, but maybe double check too

fxamacker:
@turbolent Thanks for reviewing commit https://github.com/onflow/cadence/commit/1ecf35ce2d16f344d653e41deb6534409dd0a0e4 on Friday!

Today is Canadian holiday and Friday was US holiday, maybe we can sync about this before more holidays come up.  :smile: 

>I had a look through https://github.com/onflow/cadence/commit/1ecf35ce2d16f344d653e41deb6534409dd0a0e4 and it looks like there several cases where atRoot should be true. Not sure if those cases were intentionally false. I think I caught all of them, but maybe double check too

Unless I missed something, I think we can skip incorporating all of the separate suggestions related to the last feedback.  However, my reasons vary for some suggestions and I would like to get your thoughts about some new ideas/questions.  :pray: 

I will summarize here instead of replying to each suggestion individually because they are related.

There are two main scenarios in your comments:
- Transferred value from Iterators, `Get`, or `ReadStored`, with `remove` parameter as false (in interpreter.go, value.go)
- Transferred value from `ReadStored` with `remove` parameter as true (in account.go)

### First Scenario

`atRoot` is false because:
- Transferred value is from Iterators, `Get()` or `ReadStored()`.  So the transferred value isn't a root value since it is an element read from parent container.
- `atRoot` parameter only takes effect if `remove` parameter is true in `Transfer()`.  In other words, in `Transfer()`, storage check is performed if both `atRoot` and `remove` parameters are true.  Since `remove` is false in this scenario, `atRoot` is effectively no-op.  This behavior is exactly same as before this commit (without `atRoot`).

https://github.com/onflow/cadence/blob/1ecf35ce2d16f344d653e41deb6534409dd0a0e4/runtime/interpreter/value.go#L2769-L2783

### Second Scenario

This is more involved because:
- Transferred value is read from storage map (so value isn't root because it has parent)
- `remove` is true for `Transfer()`, so it is possible to remove storable for this value from storage while storage map still manages it officially
- after `Transfer()`, we set `nil` in its place in storage map

https://github.com/onflow/cadence/blob/1ecf35ce2d16f344d653e41deb6534409dd0a0e4/runtime/stdlib/account.go#L1072-L1111

### Alternative Suggestion for Second Scenario

Would it make sense to do these instead?
- remove data from storage map (not storage)
- transfer removed data to stack address and remove data in owner address in storage

So `atRoot` in the second scenario really depends on the flow of data outlined above.  For example, if transferred value is removed from storage map first, `atRoot` is `true` for transferred value; othewise, I think `atRoot` should be `false` because the transferred data has parent (not root). 

Thoughts?

turbolent:
TODO for this PR:
- Rename the `atRoot` parameter for the `Transfer` and `DeepRemove` functions and clarify it in a comment @fxamacker 
- Double-check and potentially switch read-only iterators in `Transfer` functions to write-iterator, as the potential nested removal (`remove == true`) is a mutation @fxamacker 
- Adjust argument for `DeepRemove` calls to align with semantics of `atRoot` parameter of `Transfer` @turbolent 

TODO for another PR, ideally before:
- Simplify `Transfer` function @turbolent 
  - Split: Copy and Move
  - Can remove flag be inferred from resource-kindedness?
  - Do not remove while constructing physical copy. Instead, create copy, then DeepRemove if needed

fxamacker:
Hey @turbolent, thanks for a great meeting this morning to go over this PR again and also cover other issues in `DeepRemove()` and `Transfer()`.

>TODO for this PR:

Thanks for posting these TODOs from our meeting!  :+1: 

>  - Rename the atRoot parameter for the Transfer and DeepRemove functions and clarify it in a comment @fxamacker

Done in commit: [Rename atRoot param for Transfer() and DeepRemove()](https://github.com/onflow/cadence/pull/2882/commits/7ddc21df165331d1ff7b63b62c09601b90614ac3)

>  - Double-check and potentially switch read-only iterators in Transfer functions to write-iterator, as the potential nested removal (remove == true) is a mutation @fxamacker

Done in commit: [Use Atree non-readonly iterators in Transfer()](https://github.com/onflow/cadence/pull/2882/commits/d35b7d6017ae1cd450da506e4a7b27479c3ce3bb) 

>  - Adjust argument for DeepRemove calls to align with semantics of atRoot parameter of Transfer @turbolent

Done in commit: [Use hasNoParentContainer for inner value DeepRemove()](https://github.com/onflow/cadence/pull/2882/commits/d08a2a81ad0491bd9394ca10817f6f696ce63187)

I hope you don't mind, :pray:  I was already there and wanted to have this PR wrapped up.

> TODO for another PR, ideally before:
> - Simplify `Transfer` function @turbolent 
> - Can we maybe split the function into one for value-kinded (e.g. `Copy`) and one for resource-kinded transfer (e.g. `Move`)?
> - Can the `remove` flag be removed and inferred? Maybe from resource-kindedness?
> - When constructing physical a new value ("physical copy" in the target address), do not remove while iterating. 
>    Instead, create the copy, then `DeepRemove` (if needed)

Thanks for tackling those existing items outside this PR!  :+1:

runtime/interpreter/encode.go
```diff
@@ -681,13 +681,23 @@ func (v UFix64Value) Encode(e *atree.Encoder) error {
...
 }
 
-// Encode encodes SomeStorable as
+var _ atree.ContainerStorable = &SomeStorable{}
+
+func (s SomeStorable) Encode(e *atree.Encoder) error {
+	nonSomeStorable, nestedLevels := s.nonSomeStorable()
+	if nestedLevels == 1 {
+		return s.encode(e)
+	}
```
ramtinms, line 690: can this also be handled as part of encodeMultipleNestedLevels (instead of returning error there) ?

ramtinms (approved):
Looks good to me

runtime/interpreter/value.go
turbolent: > [...] because mutation through iterator is done by callback in returned map value

Not quite sure what that means. Do you maybe have an example?

> mutating map value is not possible by callback

The  callback is free to do anything, including removing keys, inserting new keys, mutating values in the dictionary, etc.

I guess your point is still that mutable vs read-only iterator doesn't matter?

runtime/interpreter/value.go
```diff
@@ -18274,13 +18420,17 @@ func (v *DictionaryValue) GetMember(
...
 			},
 		)
 
 	case "values":
 
-		iterator, err := v.dictionary.Iterator()
+		valueComparator := newValueComparator(interpreter, locationRange)
+		hashInputProvider := newHashInputProvider(interpreter, locationRange)
+
+		iterator, err := v.dictionary.Iterator(valueComparator, hashInputProvider)
```
turbolent, line 18433: We have to use a mutable iterator here, because the values that are returned by the iterator might get mutated, so need to have the parent notification callback set up, correct?

turbolent (commented):
Double checked the switches of iterators to read-only ones, LGTM in that regard 👍

turbolent:
@fxamacker Do you happen to remember what I meant with 
> When constructing physical a new value ("physical copy" in the target address), do not remove while iterating.
Instead, create the copy, then DeepRemove (if needed)

I vaguely remember we discussed how there's a potential problem with that approach in the inlining version

runtime/interpreter/value.go
fxamacker: Correct.  :+1:

fxamacker:
> @fxamacker Do you happen to remember what I meant with
>>    When constructing physical a new value ("physical copy" in the target address), do not remove while iterating.
>>    Instead, create the copy, then DeepRemove (if needed)
>
> I vaguely remember we discussed how there's a potential problem with that approach in the inlining version

@turbolent Yeah, I remember that too.  For context:
- we added parameter to `Transfer()`, etc. to support this PR, which increased complexity of `Transfer()`
- we reviewed `Transfer()` together and found old code in `Transfer() could use refactor (see below) but that refactor is outside scope of this PR

Part of the discussion to simplify `Transfer()`:
- it was complicated already because both "copy" and "move" functionality is combined in one function, 
- we had to add new parameter `hasNoParentContainer` to both `Transfer()` and `DeepRemove()` to make atree storage health check work (uncovered by cadence smoke test) 
  - 1ecf35ce2d16f344d653e41deb6534409dd0a0e4
  - 7ddc21df165331d1ff7b63b62c09601b90614ac3
  - d08a2a81ad0491bd9394ca10817f6f696ce63187
- since atree storage health fail during removing while iterating nested structures in `Transfer()`, we want to simplify by separating copy and remove in `Transfer`.

turbolent (approved):
Amazing work @fxamacker! 👏 👏 👏

runtime/interpreter/encode.go
fxamacker: Sorry for delay in reply, I missed this comment.  :pray: 

Maybe not in this case because encoding format is differently for `nestedLevel == 1` (using `encode()`)  vs `nestedLevel > 1` (using `encodeMultipleNestedLevels()`):
- if nested level == 1, only innermost unwrapped storable is encoded
- if nested level > 1, then 2-element array is encoded: first element is nested levels, second element is innermost unwrapped storable.

PR #2890, opened 2023-10-20
Removing an attachment is an impure operation

dsainati1:
This was not properly marked as impure in the checker. 

Thanks to @oebeling for the report. 

______

### DIFF
--- a/runtime/sema/check_remove_statement.go
+++ b/runtime/sema/check_remove_statement.go
@@ -31,6 +31,8 @@ func (checker *Checker) VisitRemoveStatement(statement *ast.RemoveStatement) (_ 
 		})
 	}
 
+	checker.ObserveImpureOperation(statement)
+
 	nominalType := checker.convertNominalType(statement.Attachment)
 	base := checker.VisitExpression(statement.Value, nil)
 	checker.checkUnusedExpressionResourceLoss(base, statement.Value)
--- a/runtime/sema/check_remove_statement.go
+++ b/runtime/sema/check_remove_statement.go
@@ -31,8 +31,6 @@ func (checker *Checker) VisitRemoveStatement(statement *ast.RemoveStatement) (_ 
 		})
 	}
 
-	checker.ObserveImpureOperation(statement)
-
 	nominalType := checker.convertNominalType(statement.Attachment)
 	base := checker.VisitExpression(statement.Value, nil)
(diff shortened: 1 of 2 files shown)


runtime/sema/check_remove_statement.go
```diff
@@ -31,6 +31,8 @@ func (checker *Checker) VisitRemoveStatement(statement *ast.RemoveStatement) (_
 		})
 	}
 
+	checker.ObserveImpureOperation(statement)
```
turbolent, line 34: Maybe it's worth verifying checking of all statements and expressions to see if there are similar cases in the code base.

turbolent (approved):
👍

runtime/tests/checker/attachments_test.go
```diff
@@ -4578,3 +4578,94 @@ func TestCheckAttachmentRemoveLossTracking(t *testing.T) {
...
+			`
+			struct S {}
+			attachment Test for S {}
+			view fun foo(s: S) {
+				remove Test from s
+			}`,
+		)
+
+		errs := RequireCheckerErrors(t, err, 1)
+		assert.IsType(t, &sema.PurityError{}, errs[0])
```
SupunS, line 4669: Why does this need report a purity error?

dsainati1: I suppose it doesn't, since this `s` is a copy. Will update.

SupunS (approved):

runtime/tests/checker/attachments_test.go
```diff
@@ -4665,6 +4665,65 @@ func TestCheckAttachmentPurity(t *testing.T) {
...
+
+		t.Parallel()
+
+		_, err := ParseAndCheck(t,
+			`
+			struct S {}
+			attachment Test for S {}
+			var s: S = S()
+			view fun foo() {
+				remove Test from s
```
turbolent, line 4681: Are the semantics for `remove` on a value-type (e.g. struct) that the variable is re-assigned? Is `s` after the removal a new copy of `s` without the attachment (and the original value is unmodified, but "lost")?

dsainati1: The variable is mutated by the `remove`, so a globally scoped or field-scoped struct will be modified.

dsainati1:
cc @turbolent does this explanation make sense?

turbolent (approved):
Yep! Nice 👌

PR #2892, opened 2023-10-22
Allow `InclusiveRange<T>` in a for-in loop

darkdrag00nv2:
Work towards https://github.com/onflow/cadence/issues/2482
Work towards https://github.com/onflow/flips/pull/96

## Description

Add support to use an `InclusiveRange<T>` in a `for-in` loop
______

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -17813,7 +17813,7 @@ func NewInclusiveRangeIterator(
 	v *CompositeValue,
 	typ InclusiveRangeStaticType,
 ) *InclusiveRangeIterator {
-	startValue := v.GetField(interpreter, EmptyLocationRange, sema.InclusiveRangeTypeStartFieldName).(IntegerValue)
+	startValue := getFieldAsIntegerValue(interpreter, EmptyLocationRange, sema.InclusiveRangeTypeStartFieldName)
 
 	zeroValue := GetSmallIntegerValue(0, typ.ElementType)
 	endValue := getFieldAsIntegerValue(interpreter, v, EmptyLocationRange, sema.InclusiveRangeTypeEndFieldName)
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -246,7 +246,7 @@ type LinkValue interface {
 // IterableValue is a value which can be iterated over, e.g. with a for-loop
 type IterableValue interface {
 	Value
-	Iterator(interpreter *Interpreter) ValueIterator
+	Iterator(interpreter *Interpreter, locationRange LocationRange) ValueIterator
 }
 
 // ValueIterator is an iterator which returns values.
@@ -1516,7 +1516,7 @@ func (v *StringValue) ConformsToStaticType(
 	return true
 }
 
-func (v *StringValue) Iterator(_ *Interpreter) ValueIterator {
+func (v *StringValue) Iterator(_ *Interpreter, _ LocationRange) ValueIterator {
 	return StringValueIterator{
 		graphemes: uniseg.NewGraphemes(v.Str),
 	}
@@ -1550,7 +1550,7 @@ type ArrayValueIterator struct {
 	atreeIterator *atree.ArrayIterator
 }
 
-func (v *ArrayValue) Iterator(_ *Interpreter) ValueIterator {
+func (v *ArrayValue) Iterator(_ *Interpreter, _ LocationRange) ValueIterator {
 	arrayIterator, err := v.array.Iterator()
 	if err != nil {
 		panic(errors.NewExternalError(err))
@@ -17783,12 +17783,12 @@ func (v *CompositeValue) RemoveTypeKey(
 	return v.RemoveMember(interpreter, locationRange, attachmentMemberName(attachmentType))
 }
 
-func (v *CompositeValue) Iterator(interpreter *Interpreter) ValueIterator {
+func (v *CompositeValue) Iterator(interpreter *Interpreter, locationRange LocationRange) ValueIterator {
 	staticType := v.StaticType(interpreter)
 
 	switch typ := staticType.(type) {
 	case InclusiveRangeStaticType:
-		return NewInclusiveRangeIterator(interpreter, v, typ)
+		return NewInclusiveRangeIterator(interpreter, locationRange, v, typ)
 
 	default:
 		// Must be caught in the checker.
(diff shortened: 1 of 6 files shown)


runtime/tests/interpreter/for_test.go
```diff
@@ -255,3 +259,108 @@ func TestInterpretForString(t *testing.T) {
 		value,
 	)
 }
+
+type inclusiveRangeForInLoopTest struct {
+	s, e, step        int8
```
SupunS, line 264:
```suggestion
	start, end, step        int8
```

runtime/tests/interpreter/for_test.go
```diff
@@ -255,3 +259,108 @@ func TestInterpretForString(t *testing.T) {
...
+		})
+	}
+
+	for _, typ := range sema.AllIntegerTypes {
+		for _, testCase := range testCases {
+			if testCase.onlySignedTest {
+				continue
+			}
+			runTestCase(t, typ, testCase)
+		}
```
SupunS, line 355: Maybe have two slices `signedTestCases` and `unsignedTestCases` and loop through the two slices separately for the two scenarios. Then checking for `onlySignedTest` is not needed.

runtime/tests/interpreter/for_test.go
```diff
@@ -255,3 +259,108 @@ func TestInterpretForString(t *testing.T) {
...
+				`
+					fun test(): Int {
+						let s : %[1]s = %[2]d
+						let e : %[1]s = %[3]d
+						let step : %[1]s = %[4]d
+						let r: InclusiveRange<%[1]s> = InclusiveRange(s, e, step: step)
+
+						var count = 0
+						for c in r {
+							count = count + 1
```
SupunS, line 314: inside the loop, also use the variable `c` and check if the values assigned to `c` are the correct ones.
e.g: append each `c` to an array and return it from the `test()` function instead of the `count`

runtime/interpreter/value.go
```diff
@@ -17782,6 +17783,68 @@ func (v *CompositeValue) RemoveTypeKey(
...
+	case InclusiveRangeStaticType:
+		elementType = typ.ElementType
+	default:
+		panic(errors.NewUnreachableError())
+	}
+
+	zeroValue := GetSmallIntegerValue(0, elementType)
+
+	// Ensure that valueToReturn is within the bounds.
+	if step.Less(interpreter, zeroValue, EmptyLocationRange) {
```
SupunS, line 17828: This check doesn't need to be done every time the `next` is called.
Maybe introduce a constructor for the `InclusiveRangeIterator`, compute this inside the constructor and memoize it.

Initialization of the `startValue` can also be moved there instead of having it in the `Iterator()` method. So all initializations are in one place.

runtime/interpreter/value_range.go
```diff
@@ -219,7 +219,7 @@ func rangeContains(
 	return AsBoolValue(result)
 }
 
-func getFieldAsIntegerValue(
+func GetFieldAsIntegerValue(
```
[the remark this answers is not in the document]
darkdrag00nv2: Yes, doesn't have to be exported.

runtime/tests/interpreter/for_test.go
darkdrag00nv2: Done, good suggestion. Thanks

runtime/tests/interpreter/for_test.go
darkdrag00nv2: Updated.

runtime/interpreter/value.go
darkdrag00nv2: Thanks, good suggestion. Done in d21296ea9.

runtime/interpreter/value.go
```diff
@@ -17782,6 +17783,68 @@ func (v *CompositeValue) RemoveTypeKey(
...
+	rangeValue   *CompositeValue
+	next         IntegerValue
+	stepNegative bool
+}
+
+var _ ValueIterator = &InclusiveRangeIterator{}
+
+func NewInclusiveRangeIterator(
+	v *CompositeValue,
+	interpreter *Interpreter,
```
SupunS, line 17809: Just a nit-pcik: Could you please swap these two params, and have the interpreter as the first param. That's the convention we have followed elsewhere.

```suggestion
	interpreter *Interpreter,
	v *CompositeValue,

```

runtime/interpreter/value.go
```diff
@@ -17782,6 +17783,68 @@ func (v *CompositeValue) RemoveTypeKey(
...
+	typ InclusiveRangeStaticType,
+) *InclusiveRangeIterator {
+	startValue := v.GetField(interpreter, EmptyLocationRange, sema.InclusiveRangeTypeStartFieldName).(IntegerValue)
+
+	zeroValue := GetSmallIntegerValue(0, typ.ElementType)
+	stepValue := v.GetField(interpreter, EmptyLocationRange, sema.InclusiveRangeTypeStepFieldName).(IntegerValue)
```
SupunS, line 17815: Could use `getFieldAsIntegerValue` instead of `v.GetField()` and then casting.

runtime/interpreter/value.go
```diff
@@ -17782,6 +17783,68 @@ func (v *CompositeValue) RemoveTypeKey(
...
+		next:         startValue,
+		stepNegative: bool(stepNegative),
+	}
+}
+
+func (i *InclusiveRangeIterator) Next(interpreter *Interpreter) Value {
+	valueToReturn := i.next
+
+	end := getFieldAsIntegerValue(i.rangeValue, interpreter, EmptyLocationRange, sema.InclusiveRangeTypeEndFieldName)
+	step := getFieldAsIntegerValue(i.rangeValue, interpreter, EmptyLocationRange, sema.InclusiveRangeTypeStepFieldName)
```
SupunS, line 17829: `step` was already computed once in the constructor. Could store it and re-used here

runtime/interpreter/value.go
```diff
@@ -17782,6 +17783,68 @@ func (v *CompositeValue) RemoveTypeKey(
...
+		rangeValue:   v,
+		next:         startValue,
+		stepNegative: bool(stepNegative),
+	}
+}
+
+func (i *InclusiveRangeIterator) Next(interpreter *Interpreter) Value {
+	valueToReturn := i.next
+
+	end := getFieldAsIntegerValue(i.rangeValue, interpreter, EmptyLocationRange, sema.InclusiveRangeTypeEndFieldName)
```
SupunS, line 17828: `end` also needs to be computed only once. Can move this to the constructor as well.

runtime/interpreter/value.go
darkdrag00nv2: cool, updated `getFieldAsIntegerValue` to follow this convention as well.

SupunS (approved):

runtime/interpreter/value.go
```diff
@@ -17782,6 +17783,73 @@ func (v *CompositeValue) RemoveTypeKey(
...
+}
+
+var _ ValueIterator = &InclusiveRangeIterator{}
+
+func NewInclusiveRangeIterator(
+	interpreter *Interpreter,
+	v *CompositeValue,
+	typ InclusiveRangeStaticType,
+) *InclusiveRangeIterator {
+	startValue := v.GetField(interpreter, EmptyLocationRange, sema.InclusiveRangeTypeStartFieldName).(IntegerValue)
```
turbolent, line 17816: Pass the proper location range here and in the rest of the function. This requires adding a new parameter for the location range, and passing it along from the call-site

runtime/interpreter/value.go
```diff
@@ -17782,6 +17783,73 @@ func (v *CompositeValue) RemoveTypeKey(
...
+	return &InclusiveRangeIterator{
+		rangeValue:   v,
+		next:         startValue,
+		stepNegative: bool(stepNegative),
+		step:         stepValue,
+		end:          endValue,
+	}
+}
+
+func (i *InclusiveRangeIterator) Next(interpreter *Interpreter) Value {
```
turbolent, line 17833: Same as above: pass the proper location range and use it inside of the function

turbolent (commented):
Nice!

runtime/interpreter/value.go

runtime/interpreter/value.go

runtime/interpreter/value.go
```diff
@@ -17834,14 +17837,14 @@ func (i *InclusiveRangeIterator) Next(interpreter *Interpreter) Value {
...
+	if i.stepNegative && bool(valueToReturn.Less(interpreter, i.end, i.locationRange)) {
 		return nil
-	} else if !i.stepNegative && bool(valueToReturn.Greater(interpreter, i.end, EmptyLocationRange)) {
+	} else if !i.stepNegative && bool(valueToReturn.Greater(interpreter, i.end, i.locationRange)) {
 		return nil
 	}
 
 	// Update the next value.
-	nextValueToReturn, ok := valueToReturn.Plus(interpreter, i.step, EmptyLocationRange).(IntegerValue)
+	nextValueToReturn, ok := valueToReturn.Plus(interpreter, i.step, i.locationRange).(IntegerValue)
```
turbolent: For the use-case here, for-looping using the iterator it actually doesn't matter, but the place of the iterator use (call to `Next`) and the place of the iterator construction (`Iterator` function) might be different.

Could you please add a location range parameter to `Next` instead and use that parameter here and in the two cases above, then remove the `locationRange` field from `InclusiveRangeIterator` again? (It looks like it wasn't populated anyway)

darkdrag00nv2: I believe you're looking at an older commit. Can you please check again? b512835679ec724dadded298f08ce3fc159d3d78 is the latest commit.

turbolent: Ah, right. Sorry, my bad!

turbolent (approved):
Nice! 👍

Issue #2869, opened 2023-10-13
New revertibleRandom function (FLIP 120) 


tarakby:
### Issue to be solved

The reasons are discussed in https://github.com/onflow/flips/pull/120. 

### Suggested Solution

- [ ] add a new function `revertibleRandom` with the same interface as the existing `unsafeRandom`.
- [ ] update the doc
 
Note that `fun unsafeRandom(): UInt64`  should not be removed to avoid immediate breaking change. 
The remaining FLIP120 implementation tasks are described in https://github.com/onflow/cadence/issues/2712

tarakby:
FYI, I'll make a PR to implement this issue.

tarakby:
This PR hasn't been merged yet https://github.com/onflow/docs/pull/423 and it's part of the issue definition. Though I don't know if we can merge it while the function isn't on Mainnet yet.

PR #2896, opened 2023-10-24
add new `revertibleRandom` function

tarakby:
Closes https://github.com/onflow/cadence/issues/2869

Please also review the doc changes https://github.com/onflow/docs/pull/423 
______

- [x] Updated relevant documentation (https://github.com/onflow/docs/pull/423)

### DIFF
--- a/runtime/common/computationkind.go
+++ b/runtime/common/computationkind.go
@@ -133,7 +133,7 @@ const (
 	//
 	ComputationKindSTDLIBPanic
 	ComputationKindSTDLIBAssert
-	ComputationKindSTDLIBUnsafeRandom
+	ComputationKindSTDLIBRevertibleRandom
 	_
 	_
 	_
--- a/runtime/common/computationkind_string.go
+++ b/runtime/common/computationkind_string.go
@@ -24,7 +24,7 @@ func _() {
 	_ = x[ComputationKindEncodeValue-1080]
 	_ = x[ComputationKindSTDLIBPanic-1100]
 	_ = x[ComputationKindSTDLIBAssert-1101]
-	_ = x[ComputationKindSTDLIBUnsafeRandom-1102]
+	_ = x[ComputationKindSTDLIBRevertibleRandom-1102]
 	_ = x[ComputationKindSTDLIBRLPDecodeString-1108]
 	_ = x[ComputationKindSTDLIBRLPDecodeList-1109]
 }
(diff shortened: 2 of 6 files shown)


runtime/common/computationkind.go
```diff
@@ -133,7 +133,7 @@ const (
 	//
 	ComputationKindSTDLIBPanic
 	ComputationKindSTDLIBAssert
-	ComputationKindSTDLIBUnsafeRandom
+	ComputationKindSTDLIBRevertibleRandom
```
tarakby, line 136: Not sure if I should keep `ComputationKindSTDLIBUnsafeRandom` here, knowing that `unsafeRandom` will be deprecated.

runtime/common/computationkind_string.go
```diff
@@ -24,7 +24,7 @@ func _() {
 	_ = x[ComputationKindEncodeValue-1080]
 	_ = x[ComputationKindSTDLIBPanic-1100]
 	_ = x[ComputationKindSTDLIBAssert-1101]
-	_ = x[ComputationKindSTDLIBUnsafeRandom-1102]
+	_ = x[ComputationKindSTDLIBRevertibleRandom-1102]
```
tarakby, line 27: file updated through `go generate`

runtime/stdlib/builtin.go
```diff
@@ -20,7 +20,7 @@ package stdlib
 
 type StandardLibraryHandler interface {
 	Logger
-	UnsafeRandomGenerator
+	RandomGenerator
```
tarakby, line 23: renamed the generator name

dsainati1 (approved):

runtime/common/computationkind.go
turbolent: I think that's OK, on the FVM side we can probably just use the weights for the old kind for the new one, right @onflow/flow-execution-and-verification ?

runtime/stdlib/random.go
```diff
@@ -26,26 +26,56 @@ import (
...
+					return binary.LittleEndian.Uint64(buffer[:])
+				},
+			)
+		},
+	)
+}
+
+// `unsafeRandom` related constants and functions will be deleted
+// when the function is deprecated
+const unsafeRandomFunctionDocString = revertibleRandomFunctionDocString
```
turbolent, line 74: Maybe add copy of the docstring for `revertibleRandom` here, but also prefix it with a deprecation warning. That way the language server will show developers a deprecation warning. Finally, in https://github.com/onflow/cadence-tools/blob/9f82dd687af84e8b6d2ec21c52d80a2b42146009/lint/deprecated_member_analyzer.go#L30 we can then suggest a known replacement

turbolent (approved):
Nice!

runtime/stdlib/random.go
```diff
@@ -26,26 +26,64 @@ import (
...
+				},
+			)
+		},
+	)
+}
+
+// `unsafeRandom` related constants and functions will be deleted
+// when the function is deprecated
+const unsafeRandomFunctionDocString = `
+Warning: this function will be deprecated, use revertibleRandom function instead.
```
turbolent, line 75: Use "deprecated" instead, as it is detected by the tooling (https://github.com/onflow/cadence-tools/blob/9f82dd687af84e8b6d2ec21c52d80a2b42146009/lint/deprecated_member_analyzer.go#L58) and annotates uses of the function with a warning:

```suggestion
Deprecated: Use revertibleRandom instead.
```

PR #2915, opened 2023-10-31
Make checkParameterizedTypeIsInstantiated recursive

### LINKED ISSUE #2884: Make checkParameterizedTypeIsInstantiated recursive
### Issue to be solved

https://github.com/onflow/cadence/pull/2523#discussion_r1361259479

```
I think then the checkParameterizedTypeIsInstantiated must be recursive.
i.e: It should also look for the types nested within other types.

For e.g: An array of InclusiveRange (without the type parameter), can be defined without any errors.

let r: [InclusiveRange] = []    // no errors
```

### Suggested Solution

_No response_

darkdrag00nv2:
Closes #2884

## Description

This PR makes the checker function `checkParameterizedTypeIsInstantiated` recursive so that a non-instantiated type parameter is caught even if it is inside a container.

See linked issue for more.

______

### DIFF
--- a/runtime/sema/checker.go
+++ b/runtime/sema/checker.go
@@ -2101,9 +2101,6 @@ func (checker *Checker) checkParameterizedTypeIsInstantiated(ty Type, pos ast.Ha
 		if t.EnumRawType != nil {
 			checker.checkParameterizedTypeIsInstantiated(t.EnumRawType, pos)
 		}
-		if t.containerType != nil {
-			checker.checkParameterizedTypeIsInstantiated(t.containerType, pos)
-		}
 		if t.baseType != nil {
 			checker.checkParameterizedTypeIsInstantiated(t.baseType, pos)
 		}
@@ -2117,10 +2114,6 @@ func (checker *Checker) checkParameterizedTypeIsInstantiated(ty Type, pos ast.Ha
 		}
 
 	case *InterfaceType:
-		if t.containerType != nil {
-			checker.checkParameterizedTypeIsInstantiated(t.containerType, pos)
-		}
-
 		for _, param := range t.InitializerParameters {
 			checker.checkParameterizedTypeIsInstantiated(param.TypeAnnotation.Type, pos)
 		}
--- a/runtime/sema/checker.go
+++ b/runtime/sema/checker.go
@@ -2069,37 +2069,74 @@ func (checker *Checker) checkInvalidInterfaceAsType(ty Type, pos ast.HasPosition
 }
 
 func (checker *Checker) checkParameterizedTypeIsInstantiated(ty Type, pos ast.HasPosition) {
-	parameterizedType, ok := ty.(ParameterizedType)
-	if !ok {
-		return
-	}
+	switch t := ty.(type) {
+	case *OptionalType:
+		checker.checkParameterizedTypeIsInstantiated(t.Type, pos)
 
-	typeArgs := parameterizedType.TypeArguments()
-	typeParameters := parameterizedType.TypeParameters()
+	case ArrayType:
+		checker.checkParameterizedTypeIsInstantiated(t.ElementType(false), pos)
 
-	typeArgumentCount := len(typeArgs)
-	typeParameterCount := len(typeParameters)
+	case *DictionaryType:
+		checker.checkParameterizedTypeIsInstantiated(t.KeyType, pos)
+		checker.checkParameterizedTypeIsInstantiated(t.ValueType, pos)
 
-	if typeArgumentCount != typeParameterCount {
-		checker.report(
-			&InvalidTypeArgumentCountError{
-				TypeParameterCount: typeParameterCount,
-				TypeArgumentCount:  typeArgumentCount,
-				Range: ast.NewRange(
-					checker.memoryGauge,
-					pos.StartPosition(),
-					pos.EndPosition(checker.memoryGauge),
-				),
-			},
-		)
-	}
+	case *ReferenceType:
+		checker.checkParameterizedTypeIsInstantiated(t.Type, pos)
 
-	// Ensure that each non-optional typeparameter is non-nil.
-	for index, typeParam := range typeParameters {
-		if !typeParam.Optional && typeArgs[index] == nil {
+	case *FunctionType:
+		for _, tyParam := range t.TypeParameters {
+			checker.checkParameterizedTypeIsInstantiated(tyParam.TypeBound, pos)
+		}
+
+		for _, param := range t.Parameters {
+			checker.checkParameterizedTypeIsInstantiated(param.TypeAnnotation.Type, pos)
+		}
+
+		checker.checkParameterizedTypeIsInstantiated(t.ReturnTypeAnnotation.Type, pos)
+
+	case *RestrictedType:
+		checker.checkParameterizedTypeIsInstantiated(t.Type, pos)
+
+	case *CompositeType:
+		if t.EnumRawType != nil {
+			checker.checkParameterizedTypeIsInstantiated(t.EnumRawType, pos)
+		}
+		if t.containerType != nil {
+			checker.checkParameterizedTypeIsInstantiated(t.containerType, pos)
+		}
+		if t.baseType != nil {
+			checker.checkParameterizedTypeIsInstantiated(t.baseType, pos)
+		}
+
+		for _, typ := range t.ImplicitTypeRequirementConformances {
+			checker.checkParameterizedTypeIsInstantiated(typ, pos)
+		}
+
+		for _, typ := range t.ExplicitInterfaceConformances {
+			checker.checkParameterizedTypeIsInstantiated(typ, pos)
+		}
+
+	case *InterfaceType:
+		if t.containerType != nil {
+			checker.checkParameterizedTypeIsInstantiated(t.containerType, pos)
+		}
+
+		for _, param := range t.InitializerParameters {
+			checker.checkParameterizedTypeIsInstantiated(param.TypeAnnotation.Type, pos)
+		}
+
+	case ParameterizedType:
+		typeArgs := t.TypeArguments()
+		typeParameters := t.TypeParameters()
+
+		typeArgumentCount := len(typeArgs)
+		typeParameterCount := len(typeParameters)
+
+		if typeArgumentCount != typeParameterCount {
 			checker.report(
-				&MissingTypeArgumentError{
-					TypeArgumentName: typeParam.Name,
+				&InvalidTypeArgumentCountError{
+					TypeParameterCount: typeParameterCount,
+					TypeArgumentCount:  typeArgumentCount,
 					Range: ast.NewRange(
 						checker.memoryGauge,
 						pos.StartPosition(),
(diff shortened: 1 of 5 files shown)


runtime/sema/checker.go
```diff
@@ -2069,37 +2069,74 @@ func (checker *Checker) checkInvalidInterfaceAsType(ty Type, pos ast.HasPosition
 }
 
 func (checker *Checker) checkParameterizedTypeIsInstantiated(ty Type, pos ast.HasPosition) {
-	parameterizedType, ok := ty.(ParameterizedType)
-	if !ok {
-		return
-	}
+	switch t := ty.(type) {
```
turbolent, line 2072: It might make sense to make this a function on `sema.Type` instead, so the implementation is ensured to be exhaustive for all types now and in the future.

runtime/sema/checker.go
```diff
@@ -2069,37 +2069,74 @@ func (checker *Checker) checkInvalidInterfaceAsType(ty Type, pos ast.HasPosition
...
+	case *RestrictedType:
+		checker.checkParameterizedTypeIsInstantiated(t.Type, pos)
+
+	case *CompositeType:
+		if t.EnumRawType != nil {
+			checker.checkParameterizedTypeIsInstantiated(t.EnumRawType, pos)
+		}
+		if t.containerType != nil {
+			checker.checkParameterizedTypeIsInstantiated(t.containerType, pos)
+		}
```
turbolent, line 2106: `containerType` here and below is the type inside of which this composite type is declared, e.g. in a nested declaration `contract C { struct S {} }`, `C` is the container type of `S`.

The current implementation recursed "upwards", so this will lead to an infinite recursion. Can you please add tests for nested types, for composite types and for interface types below?

runtime/tests/checker/typeargument_test.go
```diff
@@ -178,6 +155,79 @@ func TestCheckTypeArguments(t *testing.T) {
 	})
 }
 
+type checkParameterizedTypeIsInstantiatedTestCase struct {
+	name string
+	code string
+}
```
turbolent, line 161: Move into `TestCheckParameterizedTypeIsInstantiated`, it does not have to be a global declaration.

runtime/tests/checker/typeargument_test.go
```diff
@@ -178,6 +155,79 @@ func TestCheckTypeArguments(t *testing.T) {
...
+		},
+		{
+			name: "DictionaryType with InclusiveRange",
+			code: "let r: {Int: InclusiveRange} = {}",
+		},
+		{
+			name: "Struct with InclusiveRange",
+			code: `
+				pub struct Foo {
+					pub let a: InclusiveRange
```
turbolent, line 216: No need for access modifiers in checker tests by default. Also, this makes the code ready for the Cadence 1.0 branch.

```suggestion
				struct Foo {
					let a: InclusiveRange
```

runtime/tests/checker/typeargument_test.go
```diff
@@ -178,6 +155,79 @@ func TestCheckTypeArguments(t *testing.T) {
...
+				options,
+			)
+
+			errs := RequireCheckerErrors(t, err, 1)
+
+			assert.IsType(t, &sema.MissingTypeArgumentError{}, errs[0])
+		})
+	}
+
+	testCases := []checkParameterizedTypeIsInstantiatedTestCase{
```
turbolent, line 191: Please add more test cases to cover all paths of `checkParameterizedTypeIsInstantiated`, e.g. composite type declarations, enum type declarations, etc.

turbolent (commented):
👍

runtime/tests/checker/typeargument_test.go
darkdrag00nv2: moved.

runtime/sema/checker.go
darkdrag00nv2: I didn't get you actually. Can you please share a reference if you have?

runtime/tests/checker/typeargument_test.go
darkdrag00nv2: Added more test cases. Can you please take a look and suggest if there is anything else I should add? Thanks!

runtime/sema/checker.go
turbolent: Basically:
- Add a function `CheckInstantiated(pos ast.HasPosition, report func(err error))` to `sema.Type`
- Refactor each `case` in the switch into an implementation of `CheckInstantiated` for that `Type`, e.g. for `OptionalType` (pseudo-code, not tested):

    ```cadence
    func (t *OptionalType) CheckInstantiated(pos ast.HasPosition, report func(err error)) {
    	t.Type.CheckInstantiated(pos, report)
    }
    ```

- Inside of the function, use the `report` parameter/function instead of `checker.report`
- Change `checkParameterizedTypeIsInstantiated` to simply `ty.CheckInstantiated(pos, checker.report)`

For many types, e.g. `Void`, `String`, etc. the implementation is simply a NO-OP.

runtime/sema/checker.go
darkdrag00nv2: Done.

runtime/sema/checker.go
darkdrag00nv2: Done, thanks for the detailed explanation!

darkdrag00nv2:
Somehow CI is complaining about data race in tests. Need to check that but the PR can be looked at in the meantime.

runtime/tests/checker/typeargument_test.go
```diff
@@ -178,6 +155,681 @@ func TestCheckTypeArguments(t *testing.T) {
...
+
+	t.Parallel()
+
+	baseValueActivation := sema.NewVariableActivation(sema.BaseValueActivation)
+	baseValueActivation.DeclareValue(stdlib.InclusiveRangeConstructorFunction)
+	options := ParseAndCheckOptions{
+		Config: &sema.Config{
+			BaseValueActivation: baseValueActivation,
+		},
+	}
```
turbolent, line 168: Try making this a function: 

```go
test := func(t *testing.T, code) error {
	baseValueActivation := sema.NewVariableActivation(sema.BaseValueActivation)
	baseValueActivation.DeclareValue(stdlib.InclusiveRangeConstructorFunction)
	options := ParseAndCheckOptions{
		Config: &sema.Config{
			BaseValueActivation: baseValueActivation,
		},
	}

	_ , err := ParseAndCheckWithOptions(t, code, options)
	return err
}
```

That will likely fix the reported races

runtime/sema/simple_type.go
```diff
@@ -168,3 +168,6 @@ func (t *SimpleType) CompositeKind() common.CompositeKind {
 		return common.CompositeKindStructure
 	}
 }
+
+func (t *SimpleType) CheckInstantiated(pos ast.HasPosition, memoryGauge common.MemoryGauge, report func(err error)) {
+}
```
turbolent, line 173: Here and in the rest of the PR:

Remove unused receiver variable and replace unused parameters with `_`. Also, add an explicit `// NO-OP` comment to indicate the empty body is intentional

```suggestion
func (*SimpleType) CheckInstantiated(_ ast.HasPosition, _ common.MemoryGauge, _ func(error)) {
	// NO-OP
}
```

```suggestion
func (t *SimpleType) CheckInstantiated(pos ast.HasPosition, memoryGauge common.MemoryGauge, report func(err error)) {
}
```

darkdrag00nv2: Done.

runtime/tests/checker/typeargument_test.go
darkdrag00nv2: Thanks for the suggestion! Done.

runtime/sema/type.go
```diff
@@ -291,6 +293,49 @@ func MustInstantiate(t ParameterizedType, typeArguments ...Type) Type {
...
+				TypeParameterCount: typeParameterCount,
+				TypeArgumentCount:  typeArgumentCount,
+				Range: ast.NewRange(
+					memoryGauge,
+					pos.StartPosition(),
+					pos.EndPosition(memoryGauge),
+				),
+			},
+		)
+	}
```
SupunS, line 320: codecov reports this section is not covered by tests. Can you add a test(s) that could cover this part as well. e.g: initialized, but with invalid arg count?

SupunS (approved):
Looks good!

runtime/sema/type.go
darkdrag00nv2: This was redundant as the check also happens in the `convertInstantiationType` function. So I removed it from here.

runtime/tests/checker/typeargument_test.go
```diff
@@ -178,6 +155,665 @@ func TestCheckTypeArguments(t *testing.T) {
...
+		t.Parallel()
+
+		err := test(t,
+			"let inclusiveRange: InclusiveRange<Int, UInt> = InclusiveRange(1, 10)",
+		)
+
+		errs := RequireCheckerErrors(t, err, 2)
+
+		assert.IsType(t, &sema.InvalidTypeArgumentCountError{}, errs[0])
+		assert.IsType(t, &sema.MissingTypeArgumentError{}, errs[1])
```
darkdrag00nv2, line 197: I added this test case where the count of type arguments > type parameters. Ideally, we should only get one checker error i.e. the `InvalidTypeArgumentCountError`.

We get this extra error because we skip the instantiation of the parameterized type in case of invalid count which leads to the second error - https://github.com/onflow/cadence/blob/b48fd6b857c697ffdbbf16374b0fce683480b443/runtime/sema/checker.go#L2317-L2320

We report a missing type argument since we skipped instantiate call.

runtime/sema/type.go
```diff
@@ -3369,6 +3423,18 @@ func (t *FunctionType) initializeMemberResolvers() {
 	})
 }
 
+func (t *FunctionType) CheckInstantiated(pos ast.HasPosition, memoryGauge common.MemoryGauge, report func(err error)) {
```
turbolent, line 3426: Could you please add a test case for a function type with type parameters?

Such a function cannot be defined in a Cadence program, so you'll need to define it in Go and inject it

turbolent (approved):
Great work! 👏 

Once the last test case is added, this is good to go 👍

runtime/tests/checker/typeargument_test.go
SupunS: Thank you for adding the test. I think this is OK. There can be situations where one semantic error could lead to multiple other errors.

runtime/sema/type.go
darkdrag00nv2: Can you please guide me on how to write such a test? This check is done on function definition. So if we cannot define it in Cadence program, then not sure how to add the test case.

I tried the following:

```
	t.Run("Struct with function with type parameters", func(t *testing.T) {

		t.Parallel()

		_, err := parseAndCheckWithTestValue(t,
			`
				struct Bar {
					init() {
						let f = test
					}
				}
			`,
			stdlib.InclusiveRangeConstructorFunctionType,
		)
		require.NoError(t, err)
	})
```

The function being injected is the `InclusiveRange<T>` constructor function.

turbolent: Added tests in 89c1a6008dc212f18a1a178c30d88c9f51294aee

turbolent:
@dsainati1 Could you please also have a look?

runtime/tests/checker/typeargument_test.go
```diff
@@ -945,7 +1081,7 @@ func TestCheckTypeArgumentSubtyping(t *testing.T) {
 				},
 			},
 		)
-		require.NotNil(t, checker)
+		require.NoError(t, err)
```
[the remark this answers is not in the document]
SupunS: since this test already asserts the checker errors below, we don't really need to check the types of the global variables.

PR #2916, opened 2023-10-31
Refactor resource-reference tracking

SupunS:
## Description

Refactor the resource reference tracking and move the tracking to the `NewEphemeralReferenceValue`.

Storage references don't need any changes as they are just a reference to a storage location, and not for a value.

This should also solves the problem addressed in https://github.com/onflow/cadence/pull/2914

______


runtime/interpreter/interpreter_expression.go
```diff
@@ -1436,9 +1428,6 @@ func (interpreter *Interpreter) VisitAttachExpression(attachExpression *ast.Atta
 		panic(errors.NewUnreachableError())
 	}
 
-	// Because `self` in attachments is a reference, we need to track the attachment if it's a resource
-	interpreter.trackReferencedResourceKindedValue(attachment.StorageID(), attachment)
```
SupunS, line 1440: This is the only place where the reference is tracked outside of creating a reference.
Removing this didn't cause any tests to fail. So I assume this is redundant. 
But lemme know if you can think of a way to exploit this

dsainati1: This is the one I removed in https://github.com/onflow/cadence/pull/2914 because it was redundant, so I'm pretty sure it's safe to delete here.

runtime/interpreter/value.go
```diff
@@ -2867,12 +2830,7 @@ func (v *ArrayValue) Transfer(
...
-		if config.InvalidatedResourceValidationEnabled {
-			v.array = nil
-		} else {
-			v.array = array
-			res = v
-		}
+		v.array = nil
 	}
 
 	if res == nil {
```
turbolent, line 2836: `res` is now always `nil`, remove the surrounding if-block

runtime/interpreter/value.go
```diff
@@ -17459,12 +17401,7 @@ func (v *CompositeValue) Transfer(
...
-		if config.InvalidatedResourceValidationEnabled {
-			v.dictionary = nil
-		} else {
-			v.dictionary = dictionary
-			res = v
-		}
+		v.dictionary = nil
 	}
 
 	if res == nil {
```
turbolent, line 17407: `res` is now always `nil`, remove the surrounding if-block

runtime/interpreter/value.go
```diff
@@ -19196,12 +19097,7 @@ func (v *DictionaryValue) Transfer(
...
```
turbolent, line 19103: Same as above

runtime/interpreter/value.go
```diff
@@ -19811,15 +19682,7 @@ func (v *SomeValue) Transfer(
...
-		} else {
-			v.value = innerValue
-			v.valueStorable = nil
-			res = v
-		}
-
+		v.value = nil
 	}
 
 	if res == nil {
```
turbolent, line 19688: Same as above

runtime/interpreter/interpreter_expression.go
```diff
@@ -1036,10 +1035,7 @@ func (interpreter *Interpreter) visitInvocationExpressionWithImplicitArgument(in
 	// Bound functions
 	if boundFunction, ok := function.(BoundFunctionValue); ok && boundFunction.Self != nil {
 		self := *boundFunction.Self
-		if resource, ok := self.(ReferenceTrackedResourceKindedValue); ok {
-			storageID := resource.StorageID()
-			interpreter.trackReferencedResourceKindedValue(storageID, resource)
-		}
+		interpreter.maybeTrackReferencedResourceKindedValue(self)
```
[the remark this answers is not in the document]
turbolent: Good explanation! Please add it as a comment to the code

dsainati1 (approved):

Issue #2918, opened 2023-10-31
Implementation of new behavior for attachments with entitlements, FLIP 217


turbolent:
Implement new behavior for attachments with entitlements, [FLIP 217](https://github.com/onflow/flips/issues/217) (https://github.com/onflow/flips/pull/213)

PR #2919, opened 2023-10-31
Function types carry information about their access modifier

dsainati1:
Necessary for the implementation of https://github.com/onflow/cadence/issues/2918

______


dsainati1:
In order to determine the entitlements that the runtime values for `self` and `base` will need inside of a function invocation, we need to know what access that function was defined with. `FunctionValue` carries a `FunctionType`, which after this change would carry an `Access`, so we would be able to figure out what authorization to create the `self` and `base` references with.

turbolent:
For example, argument labels are not declared on the function type either, but are part of the variable and member, and checked at call-sites. Is the access of variables and member not sufficient here?

dsainati1:
This is not necessary for type checking, no. This is for the interpreter. If I have some attachment definition:

```
attachment A for R {
    access(E) fun foo() {}
}
```

When we interpret an expression `a.foo()` for some reference to `A`, we need to know what the appropriate entitlements are that we need to provide to the `self` and `base` references that we create inside of the body of `foo`, and we need information about `foo`'s access modifier in order to do that.

turbolent:
@dsainati1 Let's maybe sync on this tomorrow, it might be easier to clarify this in a call

PR #2921, opened 2023-11-01
New Behavior for Attachments and Entitlements

dsainati1:
Closes https://github.com/onflow/cadence/issues/2918

Reference implementation of https://github.com/onflow/flips/pull/213

______


runtime/sema/check_member_expression.go
```diff
@@ -376,7 +376,9 @@ func (checker *Checker) visitMember(expression *ast.MemberExpression, isAssignme
 // in the current location of the checker, along with the authorzation with which the result can be used
 func (checker *Checker) isReadableMember(accessedType Type, member *Member, resultingType Type, accessRange func() ast.Range) (bool, Access) {
 	if checker.Config.AccessCheckMode.IsReadableAccess(member.Access) ||
-		checker.containerTypes[member.ContainerType] {
+		// only allow references unrestricted access to members in their own container that are not entitled
+		// this prevents rights escalation attacks on entitlements
+		(member.Access.isPrimitiveAccess() && checker.containerTypes[member.ContainerType]) {
```
dsainati1, line 381: This change is needed so that the following fails:

```
entitlement E
resource R {
	access(E) fun foo() {}
}
attachment Test for R {
	access(E) fun bar() {}
	fun foo(t: &Test) {
		t.bar()
	}
}
```

This test was added to the attachment test suite

runtime/interpreter/value.go
```diff
@@ -16772,7 +16774,8 @@ func (v *CompositeValue) GetFunction(interpreter *Interpreter, locationRange Loc
 	var base *EphemeralReferenceValue
 	var self MemberAccessibleValue = v
 	if v.Kind == common.CompositeKindAttachment {
-		base, self = attachmentBaseAndSelfValues(interpreter, v)
+		functionAccess := function.FunctionType().Access
```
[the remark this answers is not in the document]
turbolent: Thank you for adding context by putting up this draft PR! I also had another read through the FLIP yesterday, but I'm still a bit confused about it, e.g. why is this a dynamic behavior, and not a static one? I'm sure we can clarify it in our sync, it's likely just a lack of understanding from my side

runtime/sema/check_member_expression.go
turbolent: `t` has type `&Test`, which is unauthorized/unentitled. Why does the call `t.bar()` need to be checked specially?

dsainati1: In the original logic, the `checker.containerTypes[member.ContainerType]` would be true, since the `t` has type `&Test`, and the container during the checking of that member access is `Test`. Without this change, we would thus permit the `t.bar()` call because the container is the defining attachment, even though `bar` requires `E`, which `t` doesn't have.

runtime/interpreter/value.go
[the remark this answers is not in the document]
turbolent: I had another look at this. It seems like we need the static access of the accessed member here, correct?

Replacing the access lookup from the function value's function type with a lookup from the composite type's member seems to pretty much work the same. The helper `Interpeter.getAccessOfMember` can be used for that.

Basically, I replaced the code here with

```go
		functionAccess := sema.EffectiveCompositeMemberAccess(
			interpreter.getAccessOfMember(v, name),
			// TODO: temporary, need to use checker config access check mode. needed for tests
			sema.AccessCheckModeNotSpecifiedUnrestricted,
		)
```

I also had to:
- Expose `semaChecker.effectiveCompositeMemberAccess` as `sema.EffectiveCompositeMemberAccess`
- Handle test cases, which mostly specify no access, so it needs to get translated to the effective access (see TODO above)
- `ConvertSemaAccessToStaticAuthorization` only handles `sema.UnauthorizedAccess` (i.e. `ast.AccessAll`; I always get confused by that naming ...), and not the other kinds of primitive access (self, contract, account). I'm always returning `UnauthorizedAccess`, but not sure that's correct

Also, I changed `Interpreter.getAccessOfMember` to not return a pointer, that seems unnecessary.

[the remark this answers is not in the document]
dsainati1: Nice! I had considered this approach, but got tripped up with how to handle "unbound" members; i.e. in the case of something like:

```
let f = attachmentRef.foo
// .. some more code
f()
```

I wasn't sure how or whether it was possible to retrieve the member that `f` originally was; it seems like in the interpreter all that remains in the `f` variable is just the function value, which only knows the function's type, and carries no information about the member it was originally accessed from. However if you have a way to handle this case then I think your approach is better!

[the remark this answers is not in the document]
dsainati1: updated to this in cb4f675cd0543ebd4c0560e6dd4bf4dff90fa597

runtime/interpreter/interpreter_expression.go
```diff
@@ -1180,8 +1180,11 @@ func (interpreter *Interpreter) VisitCastingExpression(expression *ast.CastingEx
 
 	switch expression.Operation {
 	case ast.OperationFailableCast, ast.OperationForceCast:
-		valueStaticType := value.StaticType(interpreter)
-		valueSemaType := interpreter.MustConvertStaticToSemaType(valueStaticType)
+		// if the value itself has a mapped entitlement type in its authorization
+		// (e.g. if it is a reference to `self` or `base`  in an attachment function with mapped access)
+		// substitution must also be performed on its entitlements
+		valueSemaType := interpreter.substituteMappedEntitlements(interpreter.MustSemaTypeOfValue(value))
+		valueStaticType := ConvertSemaToStaticType(interpreter, valueSemaType)
```
SupunS, line 1187: Why the `substituteMappedEntitlements` is specifically needed only during casting?
Is this something that needs to/could be done generically in all places (for e.g: inside `IsSubTypeOfSemaType`)?

runtime/sema/checker.go
```diff
@@ -2376,12 +2381,12 @@ func (checker *Checker) effectiveInterfaceMemberAccess(access Access) Access {
...
 }
 
-func (checker *Checker) effectiveCompositeMemberAccess(access Access) Access {
+func EffectiveCompositeMemberAccess(access Access, checkMode AccessCheckMode) Access {
 	if !access.Equal(PrimitiveAccess(ast.AccessNotSpecified)) {
 		return access
 	}
 
-	switch checker.Config.AccessCheckMode {
+	switch checkMode {
```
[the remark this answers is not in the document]
dsainati1: This is leftover from a previous attempt to make this function available in the interpreter, but is no longer needed. Removed.

runtime/interpreter/interpreter_expression.go
dsainati1: There are specific semantics for when an entitlement mapping appears in a casting expression, as this is the only place in the language where mappings can traverse the "boundary" between static and runtime behavior:

```
access(mapping M) fun foo(): auth(mapping M) &Int {
    return &self.intField as auth(mapping M) &Int
}
```

for example, needs to "instantiate" the `auth(mapping M) &Int` that appears in the cast to the concrete entitlements that it actually should have at runtime based on the entitlements used to access the method. This is what `substituteMappedEntitlements` does. Applying this more generally is not possible since there are places where we don't want entitlement mappings to be replaced with concrete ones (e.g. generally during subtyping or determining interface conformance).

turbolent: Great explanation! Could you please extend the comment in the code with a variant of it?

runtime/interpreter/interpreter.go
```diff
@@ -1001,7 +999,10 @@ func (interpreter *Interpreter) evaluateDefaultDestroyEvent(
 	var self MemberAccessibleValue = containingResourceComposite
 	if containingResourceComposite.Kind == common.CompositeKindAttachment {
 		var base *EphemeralReferenceValue
-		base, self = attachmentBaseAndSelfValues(declarationInterpreter, containingResourceComposite)
+		// in evaluation of destroy events, base and self are fully entitled, as the value must be owned
+		supportedEntitlements := interpreter.MustSemaTypeOfValue(containingResourceComposite).(*sema.CompositeType).SupportedEntitlements()
```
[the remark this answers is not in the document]
turbolent: Also, should this maybe just be `sema.EntitlementSupportingType` instead, like below?

runtime/sema/access.go
```diff
@@ -32,6 +32,7 @@ import (
 
 type Access interface {
 	isAccess()
+	IsPrimitiveAccess() bool
```
turbolent, line 35: Could call-sites simply check if a given `var access Access` is a primitive access via `_, ok := access.(PrimitiveAccess)` instead? It looks like implementations of the function are simply true or false based on the type being `PrimitiveAccess` or not

runtime/sema/check_composite_declaration.go
```diff
@@ -146,11 +132,11 @@ func (checker *Checker) visitAttachmentDeclaration(declaration *ast.AttachmentDe
 
 	checker.visitCompositeLikeDeclaration(declaration)
 	attachmentType := checker.Elaboration.CompositeDeclarationType(declaration)
-	checker.checkAttachmentMembersAccess(attachmentType)
 	checker.checkAttachmentBaseType(
 		attachmentType,
 		declaration.BaseType,
 	)
+	checker.checkAttachmentMembersAccess(attachmentType)
```
turbolent, line 139: Is the change of order significant?

runtime/sema/type.go
```diff
@@ -6019,6 +6024,9 @@ func (t *ReferenceType) String() string {
 	if t.Authorization != UnauthorizedAccess {
 		authorization = t.Authorization.String()
 	}
+	if _, isMapping := t.Authorization.(*EntitlementMapAccess); isMapping {
+		authorization = "mapping " + authorization
```
turbolent, line 6028: Maybe this should be part of the authorization `String`/`QualifiedString` functions?

runtime/sema/check_composite_declaration.go
dsainati1: No, reverted the swap

runtime/sema/access.go
dsainati1: They could, but this makes it easier to pass to a function, or to otherwise use without having to declare a new variable. It's just personally preference, if you feel strongly I can remove this.

runtime/sema/type.go
dsainati1: Those are used to compute the TypeIDs for `EntitlementMapAccess`es, and we want those not to include the "mapping" prefix since they are used for map lookups/indexing.

turbolent (approved):
Great work! 👏

PR #2922, opened 2023-11-01
Fix esm build to make it work with node

wfalcon0x:
Closes #???

## Description

Previous PR  [https://github.com/onflow/cadence/pull/2815](https://github.com/onflow/cadence/pull/2815) has been provided a solution to build cadence-parser in a way that supports running it nodejs.

An alternative solution has been merged [https://github.com/onflow/cadence/pull/2851](https://github.com/onflow/cadence/pull/2851), which unfortunately was still not working with nodejs and failed with the following error:
```
const result = global[_CadenceParser.functionName("parse")](code);      
TypeError: global[_CadenceParser.functionName(...)] is not a function
```

This PR provides with a similar approach to [https://github.com/onflow/cadence/pull/2851], but making sure the solution works both in browser and nodejs.

______

### DIFF
--- a/npm-packages/cadence-parser/package.json
+++ b/npm-packages/cadence-parser/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@onflow/cadence-parser",
-  "version": "0.42.1",
+  "version": "0.42.2",
   "description": "The Cadence parser",
   "homepage": "https://github.com/onflow/cadence",
   "repository": {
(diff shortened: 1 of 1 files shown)


jribbink:
Hey @wfalcon0x!  Thanks for the PR!

From your error it looks like the package is actually importing correctly, and instead, this is a runtime error.  Specifically, it looks like you may not have loaded the WASM module into the parser before trying to use the `CadenceParser.parse` function.

Can you try this code?  (Make sure to convert it to ESM if you prefer to use that instead)

```js
const {CadenceParser} = require("@onflow/cadence-parser")
const fs = require("fs")

const file = fs.readFileSync("./node_modules/@onflow/cadence-parser/dist/cadence-parser.wasm", "binary")
CadenceParser.load(Buffer.from(file, "binary")).then(() => {
  const parser = new CadenceParser()
  console.log(parser.parse("pub fun main() {}"))
})
```

wfalcon0x:
Hey @jribbink ,

I'm using cadence parser from Node.js ES module (ESM), and my code looks the following:

```
import {CadenceParser} from "@onflow/cadence-parser"
import * as fs from "fs"

const wasm = fs.readFileSync('./node_modules/@onflow/cadence-parser/dist/cadence-parser.wasm')
const parser = await CadenceParser.create(wasm)
...
```

There are additional details about the original issue in my previous PR: [https://github.com/onflow/cadence/pull/2815](https://github.com/onflow/cadence/pull/2815). 
Unfortunately the solution that has been picked to be merged instead, doesn't work with Node.js ES module.

I have tested this and my original PR with both in browser and with Node.js.

I have also published the solution created in this PR here:
https://www.npmjs.com/package/@wfalcon/cadence-parser/v/0.0.17

It works perfectly with the following code from a Node.js ES module:

```
import {CadenceParser} from "@wfalcon/cadence-parser"
import * as fs from "fs"

const wasm = fs.readFileSync('./node_modules/@wfalcon/cadence-parser/dist/cadence-parser.wasm')
const parser = await CadenceParser.create(wasm)
...
```

Please note, package.json "type" value need to be set to "module", which tells Node.js to interpret .js files within that package as using ES module syntax.

PR #2923, opened 2023-11-01
Add account type migration

### LINKED ISSUE #2861: Migration - Account type refactor
- Mapping
    - `AuthAccount` → `auth(Storage, …) &Account`
    - `PublicAccount` → `&Account`
- `Type` values
    - `Type<AuthAccount>()` stores `AuthAccount` as type ID
    - Might be used keys → “rehash” by removing and reinserting

SupunS:
Closes #2861

## Description

______


migrations/account_type/migration.go
```diff
@@ -0,0 +1,237 @@
...
+		}
+
+	case *interpreter.CapabilityStaticType:
+		convertedBorrowType := m.maybeConvertAccountType(staticType.BorrowType)
+		if convertedBorrowType != nil {
+			return interpreter.NewCapabilityStaticType(nil, convertedBorrowType)
+		}
+
+	case *interpreter.IntersectionStaticType:
+		// Nothing to do. Inner types can only be interfaces.
```
turbolent, line 134: Maybe still migrate `LegacyType`, the previous restricted type

migrations/account_type/migration.go
```diff
@@ -0,0 +1,237 @@
...
+	accountStorage.ForEachValue(
+		m.interpreter,
+		common.AllPathDomains,
+		m.migrateValue,
+		reporter,
+	)
+}
+
+func (m *AccountTypeMigration) migrateValue(value interpreter.Value) (newValue interpreter.Value, updatedInPlace bool) {
+	return migrations.MigrateNestedValue(m.interpreter, value, m.migrateTypeValue)
```
turbolent, line 84: The migration also needs to handle other values in which static types occur, e.g. dictionaries, arrays, and capabilities.

Links (`PathLinkValue`) should probably also be migrated? 🤔 I assume the migration will run before the Cap Cons migration?

turbolent (commented):
Nice!

migrations/migrate_value.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/onflow/cadence/runtime/interpreter"
+)
+
+var emptyLocationRange = interpreter.EmptyLocationRange
+
+func MigrateNestedValue(
+	inter *interpreter.Interpreter,
+	value interpreter.Value,
+	migrate func(value interpreter.Value) (newValue interpreter.Value, updatedInPlace bool),
+) (newValue interpreter.Value, updatedInPlace bool) {
```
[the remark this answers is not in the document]
SupunS: It's used for reporting (i.e. if a nested value was updated, then report the path for the outermost value) but was removed in the refactor PR: https://github.com/onflow/cadence/pull/2942. #2942 does the same when reporting but, without this flag.

migrations/migrate_value.go
```diff
@@ -0,0 +1,149 @@
...
+			existingValue := composite.GetField(inter, interpreter.EmptyLocationRange, fieldName)
+
+			migratedValue, valueUpdated := MigrateNestedValue(inter, existingValue, migrate)
+			if migratedValue == nil {
+				continue
+			}
+
+			composite.SetMember(inter, emptyLocationRange, fieldName, migratedValue)
+
+			updatedInPlace = updatedInPlace || valueUpdated
```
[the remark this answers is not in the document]
SupunS: Ah that's a bug. FIxed it in [fd80537](https://github.com/onflow/cadence/pull/2923/commits/fd80537019373f81ec736770a877143004fcbfd7). Anyway this flag is removed in the refactor PR.

migrations/account_type/migration.go
SupunS: The account type is not storable, so an array/dictionary with an inner type (or nested) as an account type cannot exist in the storage in theory, right?

SupunS: > Links (PathLinkValue) should probably also be migrated? 🤔 I assume the migration will run before the Cap Cons migration?

In stable-cadence branch, I think we don't have `PathLinkValue`. Also, the Capability values are the ID-based ones (not the old path-based ones). So to avoid having to bring back those values, which may have been already done in the CapCpn migration, It might be good/easier to run the other migrations after the CapCon migration, WDYT?

turbolent: > The account type is not storable, so an array/dictionary with an inner type (or nested) as an account type cannot exist in the storage in theory, right?

I guess we can skip those containers, just worried we'll miss something.

> It might be good/easier to run the other migrations after the CapCon migration.

We can try running this migration after the Cap Cons migration, but in that case, the capability controllers (borrow types) might need to be migrated?

turbolent (approved):

migrations/account_type/migration.go
SupunS: I'll add this in a follow-up PR

Issue #2871, opened 2023-10-14
Improve the `smallIntegerValueCache` added during InclusiveRange implementation


darkdrag00nv2:
### Issue to be solved

The cache is thread safe and we should measure if there is an advantage in having the cache vs the downside of having a lock.

Comment from @turbolent on https://github.com/onflow/cadence/pull/2523

> Given that the cache is global, different interpreter runs, in tests or not, might concurrently hit the cache, so it needs to be safe. So far all other "state" is interpreter-instance specific and not global.

> My gut feeling is that the overhead of making this thread-safe and lock contention might be worse than the benefits of caching/re-using existing values. We might want to check and optimize this if needed, Maybe best to do that in a follow-up PR, this one is already quite large.

darkdrag00nv2:
Based on the benchmark results from https://github.com/onflow/cadence/pull/2936, we are okay with using the cache.

PR #2936, opened 2023-11-09
Introduce benchmark for the `smallIntegerValueCache`

darkdrag00nv2:
Work towards #2871

## Description

Introduces benchmark for the `smallIntegerValueCache`.

Results on my machine

##### Without Cache
```
BenchmarkIntegerCreationWithoutCache-16    	      91	  15024692 ns/op	12304062 B/op	  971404 allocs/op
```

Command: `go test -benchmem -run=^$ -bench ^BenchmarkIntegerCreationWithoutCache$ github.com/onflow/cadence/runtime/interpreter`

##### With Cache
```
BenchmarkSmallIntegerValueCache-16    	      86	  13857062 ns/op	 7401608 B/op	  767457 allocs/op
```

Command: `go test -benchmem -run=^$ -bench ^BenchmarkSmallIntegerValueCache$ github.com/onflow/cadence/runtime/interpreter`
______

### DIFF
--- a/runtime/interpreter/integer_test.go
+++ b/runtime/interpreter/integer_test.go
@@ -44,7 +44,7 @@ func runBench(b *testing.B, getValue func(value int8, staticType StaticType) Int
 	bench := func() {
 		defer wg.Done()
 
-		for i := int8(0); i <= 127; i++ {
+		for i := 0; i <= math.MaxInt8; i++ {
 			for _, integerType := range sema.AllIntegerTypes {
 				switch integerType {
 				case sema.IntegerType, sema.SignedIntegerType:
--- a/runtime/interpreter/integer_test.go
+++ b/runtime/interpreter/integer_test.go
@@ -30,7 +30,7 @@ import (
 const numGoRoutines = 100
 
 func runBench(b *testing.B, getValue func(value int8, staticType StaticType) IntegerValue) {
-	semaToStaticType := make(map[sema.Type]StaticType)
+	staticTypes := make([]StaticType, 0)
 	for _, integerType := range sema.AllIntegerTypes {
 		switch integerType {
 		case sema.IntegerType, sema.SignedIntegerType:
@@ -38,7 +38,7 @@ func runBench(b *testing.B, getValue func(value int8, staticType StaticType) Int
 		}
 
 		integerStaticType := ConvertSemaToStaticType(nil, integerType)
-		semaToStaticType[integerType] = integerStaticType
+		staticTypes = append(staticTypes, integerStaticType)
 	}
 
 	var wg sync.WaitGroup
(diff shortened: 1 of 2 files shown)


runtime/interpreter/integer_test.go
```diff
@@ -0,0 +1,82 @@
...
+		semaToStaticType[integerType] = integerStaticType
+	}
+
+	var wg sync.WaitGroup
+
+	bench := func() {
+		defer wg.Done()
+
+		for i := int8(0); i <= 127; i++ {
+			for _, integerType := range sema.AllIntegerTypes {
```
SupunS, line 48: This could simply iterate through `semaToStaticType`.
`semaToStaticType` can be a slice in that case.

SupunS (commented):
Can you please also post the results for the two benchmarks?

runtime/interpreter/integer_test.go
```diff
@@ -0,0 +1,82 @@
...
+				switch integerType {
+				case sema.IntegerType, sema.SignedIntegerType:
+					continue
+				}
+
+				value := getValue(i, semaToStaticType[integerType])
+
+				// Force a NOOP function call to ensure value isn't optimized away
+				value.isValue()
```
turbolent, line 57: https://github.com/golang/go/issues/27400 indicates that `runtime.KeepAlive` is the best option

runtime/interpreter/integer_test.go
```diff
@@ -0,0 +1,82 @@
...
+		integerStaticType := ConvertSemaToStaticType(nil, integerType)
+		semaToStaticType[integerType] = integerStaticType
+	}
+
+	var wg sync.WaitGroup
+
+	bench := func() {
+		defer wg.Done()
+
+		for i := int8(0); i <= 127; i++ {
```
turbolent, line 47: That causes an infinite loop (`i <= 127` is always true, as the increment will overflow and cause a wrap-around):

```suggestion
		for i := 0; i <= math.MaxInt8; i++ {
```

turbolent (commented):
Thank you for adding the benchmarks!

Something seems off in the cached version, either in the benchmark or the reporting for it:

```
BenchmarkSmallIntegerValueCache-10         456	   2576811 ns/op	      46 B/op	       0 allocs/op
BenchmarkIntegerCreationWithoutCache-10    368	   3275893 ns/op	 4089900 B/op	  204002 allocs/op
```

How come the cached version has 0 allocations? I'd expect it to allocate less, but at least _something_

turbolent:
Using the value, e.g. by calling Sprintf, the results look a bit better and show the cache saves quite some allocations and isn't slower 👍

runtime/interpreter/integer_test.go
darkdrag00nv2: Done

darkdrag00nv2:
@SupunS Added the results on my machine as well in the description. The cache does seem to help.

@turbolent  Had to use `Sprintf`. `runtime.KeepAlive` was also leading to 0 allocs as you pointed our earlier.

bluesign:
Very nice to have cache for small integers. 👏

> How come the cached version has 0 allocations? I'd expect it to allocate less, but at least something

It is optimized to use register there probably.

darkdrag00nv2:
@turbolent @SupunS Shall we merge this? The results look fine and I think it'll be nice to check-in the benchmark.

SupunS (approved):
Looks good!
My only concern is the `sprintf()` call is affecting the benchmark numbers quite a bit (i.e: sprintf is more expensive than the number creation). But seems there is no other way to avoid optimizing away.

turbolent:
I don't think we need to merge this in actually, the benchmarks were just a means to see if the caching approach is worthwhile. We usually have benchmarks checked in for things we do not want to regress performance on, as well as like Supun pointed out, the benchmarks are tricky to get right / are not ideal in the current form.

It was still very valuable to have these benchmarks to give us data to make decisions 🙏

PR #2942, opened 2023-11-10
Generalize the migrations and make common codes re-usable

SupunS:
## Description

Any new migration can implement the `Migration` interface (below), and only needs to implement the conversion logic for the value it migrates/converts. Any new migration doesn't need to worry about the iteration of storage and traversing containers typed values to reach an inner value that needs to be converted.

```go
type Migration interface {
	Name() string
	Migrate(value interpreter.Value) (newValue interpreter.Value)
}
```

______


dsainati1 (approved):
Nice! This is much simpler.

migrations/account_storage.go
```diff
@@ -68,23 +71,17 @@ func (i *AccountStorage) ForEachValue(
...
-				storageMap.SetValue(
-					inter,
-					storageKey,
-					newValue,
-				)
-			}
-
-			if reporter != nil {
-				reporter.Report(i.address, domain, key)
-			}
```
turbolent, line 87: Any reason for removing the reporting here?

turbolent (commented):
Nice!

migrations/account_storage.go
SupunS: It is moved to `StorageMigration` at: https://github.com/onflow/cadence/blob/3382e8db54825f25d589342215384dda03c263ad/migrations/migration.go#L209-L211

so the migration does the reporting, rather than the storage iterator

turbolent (approved):

Issue #2886, opened 2023-10-19
Fix incorrect type inference for InclusiveRange


darkdrag00nv2:
### Issue to be solved

https://github.com/onflow/cadence/pull/2523/files#r1361281514

> While testing, I also ran into the following cases that currently report errors, but shouldn't:
>
> The inclusive range constructor always takes Int as the type parameter's type.
> ```cadence
> let r = InclusiveRange<Int>(0, 10)    // Works
> ```
>
> ```cadence
> let r = InclusiveRange<Int8>(0, 10)    // Checker error (should also work)
> ```
>
> Also probably related to above, Type inferring also doesn't infer the type parameter from LHS.
> ```cadence
> let r: InclusiveRange<Int8> = InclusiveRange(0, 10)   // Type-mismatch error
> ```

darkdrag00nv2:
@SupunS 

Did a bit of investigation on this. I can see that we don't support type inference of function arguments as suggested by this comment: https://github.com/onflow/cadence/blob/323b887c1202eaa99119d44308f8f70758b20426/runtime/sema/check_invocation_expression.go#L577

So seems like both the cases mentioned are not bugs but more of an existing limitation.

I am happy to take a shot at adding this support but would it be possible to untie this with the InclusiveRange work? I would like to wrap up the grant work sooner if possible and then take this up unless we consider this to be a blocker for the InclusiveRange feature. Thoughts?

cc: @turbolent

turbolent:
I didn't realize we still had some gaps there. Aren't we already using this elsewhere, e.g. for the storage `save` function? Has this maybe improved on the `feature/stable-cadence` branch?

The work on the standard library identified this gap in the checker, but the work / grant for the standard library is is not dependent on resolving these gaps, no worries. 

We might still need to address these gaps before we can ship the feature, but that is separate from your grant.

SupunS:
Makes sense. If this is already an existing limitation, and fixing it requires changing the way type-parameters are type-checked in general, then yeah, we can/should do it separately 👍 

Thanks for looking into this!

darkdrag00nv2:
> I didn't realize we still had some gaps there. Aren't we already using this elsewhere, e.g. for the storage save function? Has this maybe improved on the feature/stable-cadence branch?

[...]

Issue #2712, opened 2023-08-10
Update `revertibleRandom` interface (FLIP 120)


tarakby:
### Issue to be solved

The reasons are discussed in https://github.com/onflow/flips/pull/120. 
This issue should be implemented after the first task described in https://github.com/onflow/cadence/issues/2869.

### Suggested Solution

- [x] update  `revertibleRandom` interface. The new function signature should be `fun revertibleRandom<T: FixedSizeUnsignedInteger>([modulo: T]): T`
- [x] `FixedSizeUnsignedInteger` is the list `UInt8`, `UInt16`, `UInt32`, `UInt64`, `UInt128`, `UInt256`, `Word8`, `Word16`, `Word32`, `Word64`
- [ ] `modulo` is an optional parameter, if provided the returned value `r` satisfies `0<= r < modulo`. The function panics if `modulo == 0` - algorithm implemented should be sample rejection. 
- [ ] Update the documentation
 
Note that `fun unsafeRandom(): UInt64`  should not be removed to avoid immediate breaking change. 
Implementation can start by merging https://github.com/onflow/cadence/pull/2706 to use the new `Interface` method.

tarakby:
FYI, I'll provide the sample rejection method to return a random less than `modulo`, but that can be done as the last implementation step .

tarakby:
@turbolent would it be simpler for the Cadence team if I separate the first implementation item "add new function `revertibleRandom`" in a separate issue, and leave the other items for later? 

The reason is that that the first item is what we need to have a safe randomness supported by Cadence. The other items are only functionality improvements. 

`revertibleRandom` would basically be an exact copy of the current function `unsafeRandom`.

AlexHentschel:
Following up on a slack conversation with Tarak here:

If I understood correctly, the `modulo` parameter in the final `revertibleRandom` is optional.  Furthermore, we have a flavour of    `revertibleRandom` generating an `UInt64`. Hence, one possible expression of `revertibleRandom` is 
```
fun revertibleRandom(): UInt64
```
which is essentially the same as the current `unsafeRandom`. I totally support Tarak's proposal to implement `fun revertibleRandom(): UInt64` first as this would be a really impactful interim deliverable!

tarakby:
I went ahead and split the item tasks in this issue. The first task is now described in https://github.com/onflow/cadence/issues/2869 and I'll try to implement it myself.

turbolent:
[...]

PR #2943, opened 2023-11-14
Update interface of `revertibleRandom`

darkdrag00nv2:
Work towards #2712

## Description

1. Introduced a new type `FixedSizeUnsignedInteger`
2. Update the interface of the `revertibleRandom` function to `fun revertibleRandom<T: FixedSizeUnsignedInteger>([modulo: T]): T`

Note that updating the interface is a **breaking change** as the existing usage of the function such as below fails.

```
let rand1 = revertibleRandom()  // error: cannot infer type parameter: `T`
```

#### Noteworthy Points
1. Updating the interface is a breaking change as pointed out in the description and also the issue https://github.com/onflow/cadence/issues/2712#issuecomment-1785810009
2. Since we don't support type inference of function arguments, we will run into the same issues as the `InclusiveRange` constructor function: https://github.com/onflow/cadence/issues/2886

Basically:

```
let r = revertibleRandom<Int8>(0) // checker error, type mismatch - expected: Int8, Found - Int
let r = revertibleRandom<Int8>(Int8(0)) // success
```
3. The modulo operation hasn't been implemented in this PR. It'll be done in a future PR based on the discussion on the issue.
______


darkdrag00nv2:
Opened up for some early feedback. I haven't yet done the modulo part.

A couple more things to point out:
1. Updating the interface is a breaking change as pointed out in the description and also the issue https://github.com/onflow/cadence/issues/2712#issuecomment-1785810009
2. Since we don't support type inference of function arguments, we will run into the same issues as the `InclusiveRange` constructor function: https://github.com/onflow/cadence/issues/2886

Basically:

```
let r = revertibleRandom<Int8>(0) // checker error, type mismatch
let r = revertibleRandom<Int8>(Int8(0)) // success
```

runtime/stdlib/random.go
```diff
@@ -34,37 +35,164 @@ NOTE: The use of this function is unsafe if not used correctly.
...
+					func() uint64 {
+						return binary.LittleEndian.Uint64(getRandomBytes(generator, 8))
+					},
+				)
+			case interpreter.PrimitiveStaticTypeWord128:
+				return interpreter.NewWord128ValueFromBigInt(
+					inter,
+					func() *big.Int {
+						buffer := getRandomBytes(generator, 16)
+						return interpreter.LittleEndianBytesToUnsignedBigInt(buffer)
```
SupunS, line 180: Are the bytes returned from `getRandomBytes`/ `generator.ReadRandom()` always little-endian (or is it a safe assumption)?
The API doesn't specify this. cc: @tarakby

SupunS (commented):
Changes look good so far!

Given we are introducing a new type, we should probably add some more high-level Cadence tests (you've already added low-level Go tests, which is great!), for usages of `FixedSizeUnsignedIntegerType` in various places, such as:
- As a type annotation for variable
- Constructing a value (valid/invalid)
- `main` function arguments/return type (for import/export)
- Storing to the storage and reading from it.
- Casting, sub-type assignment, etc.

SupunS:
Given this is a breaking change, can you please rebase this to the `feature/stable-cadence` branch?

runtime/stdlib/random.go
tarakby: `ReadRandom` only generates bytes, it doesn't generate numbers. This means that there is not endianness notion at the level of `ReadRandom` and its output (that's why the doc doesn't talk about it).

Endianness only makes sense when we want to convert the bytes into a number (at `revertibleRandom` implementation level, since it currently returns a 64-bits number). For that purpose, I [chose little endianness](https://github.com/onflow/cadence/blob/25e5405ecf9a1a2c580155012975b1764281fc24/runtime/stdlib/random.go#L65) but we could also choose big. That choice changes the output numbers, but does not impact the random number distribution.

SupunS: oh I see. Thanks for the explanation!

@darkdrag00nv2 Maybe we could avoid the conversion here (in `LittleEndianBytesToUnsignedBigInt`) and directly use `BigEndianBytesToUnsignedBigInt`?

SupunS: Might also make sense to consider big-endian for all the byte-to-number conversion in that case, for consistency?

runtime/sema/type_test.go
```diff
@@ -892,9 +892,18 @@ func TestCommonSuperType(t *testing.T) {
...
 				},
 				expectedSuperType: IntegerType,
 			},
+			{
+				name: "fixed size unsigned integers",
+				types: []Type{
+					UInt8Type,
+					UInt128Type,
+				},
+				expectedSuperType: FixedSizeUnsignedIntegerType,
```
turbolent, line 905: 👌

runtime/stdlib/random.go
turbolent: Agreed! The `binary.(Little|Big)Endian.UintN` functions are useful to convert the bytes into an integer, but it doesn't matter which we use. Maybe consistently use big, and remove `LittleEndianBytesToUnsignedBigInt`

turbolent (commented):
Great work! 

Looks good, only thing left is the removal of the unnecessary endianness conversion

runtime/stdlib/random.go
darkdrag00nv2: Done

tarakby:
@turbolent, @SupunS, @darkdrag00nv2  do you think we should add the optional parameter `modulo` as part of this PR?
The optional parameter could be added but the implementation of `revertibleRandom` would ignore it for now. This would allow me to open a follow-up PR and use the parameter to generate an output less than `modulo`. 

Another option is to leave this PR as it is, open a new PR where the optional parameter is added, and I'll continue to commit on that new PR. 
Feel free to choose either way, but it would be good to continue the work on this topic now that it's started.

turbolent:
@tarakby either way works. The PR has the `modulo` parameter in the definition, but the implementation still ignores it: https://github.com/onflow/cadence/pull/2943/files#diff-a1bff80973b499df2d90ec34d5ef987dae71f400bc6a8fab5c6fc5856661fcf7R94. We can merge it as-is and address it in a follow-up PR

turbolent:
There's a data race in the tests: https://github.com/onflow/cadence/actions/runs/7121193421/job/19389931602?pr=2943, looking into it

tarakby:
@turbolent, Nice! I didn't notice the parameter is already there. I'll create an issue for myself to work on the remaining part.

runtime/stdlib/random.go
```diff
@@ -34,37 +35,164 @@ NOTE: The use of this function is unsafe if not used correctly.
...
-				},
-			)
+			inter := invocation.Interpreter
+
+			// TODO: Check if invocation has an argument and implement modulo operation.
+
+			returnIntegerType := invocation.TypeParameterTypes.Oldest().Value
+			returnIntegerStaticType := interpreter.ConvertSemaToStaticType(inter, returnIntegerType)
+
+			switch returnIntegerStaticType {
```
turbolent, line 99: Avoid the unnecessary conversion here and simply switch on the sema type instead

runtime/tests/interpreter/equality_test.go
```diff
@@ -225,6 +225,70 @@ func TestInterpretEqualityOnNumericSuperTypes(t *testing.T) {
...
+			interpreter.PrimitiveStaticTypeWord64,
+			interpreter.PrimitiveStaticTypeWord128,
+			interpreter.PrimitiveStaticTypeWord256,
+		}
+
+		for _, subtype := range subtypes {
+			rhsType := interpreter.PrimitiveStaticTypeUInt8
+			if subtype == rhsType {
+				rhsType = interpreter.PrimitiveStaticTypeWord128
+			}
```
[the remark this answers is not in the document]
darkdrag00nv2: Not strictly required but just to keep the static type of `x` and `y` different. We are doing the same in the `Integer subtypes` test just above.

```
  let x: FixedSizeUnsignedInteger = 5 as %s
  let y: FixedSizeUnsignedInteger = 2 as %s
```

runtime/stdlib/random.go
darkdrag00nv2: Done. https://github.com/onflow/cadence/pull/2943/commits/1def84b8a5b61e20a54be4094264da9792bd26e8

darkdrag00nv2:
I believe this should labelled with `Language Breaking Change` as the interface is not backwards compatible.

SupunS:
Not sure why this got closed. This should now be opened against the master (https://github.com/onflow/cadence/pull/2971)

dsainati1:
Same, it seems as though it was closed automatically when `feature/stable-cadence` was merged into `master`. I can't re-open it for some reason though.

PR #2945, opened 2023-11-16
Introduce `dereference` on the Reference type

### LINKED ISSUE #2803: Need a way get a dereferenced copy from a reference
### Issue to be solved

Sometimes it is useful to be able to get a dereferenced-copy from a reference.

For e.g. assume a large dictionary is in storage, and one would need to get a copy of just one entry of that dictionary. In the latest released version, this is possible by:
- First, borrowing a reference to the dictionary (say `dictionaryRef`)
- Accessing the entry using index-access `dictionaryRef[key]`, which would return a copy of the value at the `key`.

However, with stable cadence, given the index syntax `dictionaryRef[key]` also returns another reference, the above approach no longer works. Instead, they would need to 
- Take a "copy" of the entire dictionary first (say `dictionaryCopy`)
- Then, get a copy of the value using the index-access `dictionaryCopy[key]`.
The downside of this is now they would need to copy the entire outer dictionary which could be a very expensive operation.

### Suggested Solution

It would be nice to have a way to get a dereferenced copy from a reference (say a built-in method `dereference()`), so that they can get a copy from the reference itself.
i.e:
```cadence
let elementRef = dictionaryRef[key]

let elementCopy = dereference(elementRef)
```

Some suggestions for the name of the built-in method:
- `dereference()`
- `dereferenceCopy()`
- Could also use the c-like dereference operator- asterisk (`*`), but I personally prefer one of the above, because they are verbose.

darkdrag00nv2:
Closes #2803

## Description

______

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -20311,6 +20311,17 @@ func (v *EphemeralReferenceValue) GetMember(
 ) Value {
 	self := v.MustReferencedValue(interpreter, locationRange)
 
+	switch name {
+	case sema.ReferenceTypeDereferenceFunctionName:
+		return NewHostFunctionValue(
+			interpreter,
+			sema.ReferenceDereferenceFunctionType(v.BorrowedType),
+			func(invocation Invocation) Value {
+				return v.Value
+			},
+		)
+	}
+
 	return interpreter.getMember(self, locationRange, name)
 }
 
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -5957,8 +5957,10 @@ func (t *DictionaryType) SupportedEntitlements() *EntitlementOrderedSet {
 
 // ReferenceType represents the reference to a value
 type ReferenceType struct {
-	Type          Type
-	Authorization Access
+	Type                Type
+	Authorization       Access
+	memberResolvers     map[string]MemberResolver
+	memberResolversOnce sync.Once
 }
 
 var _ Type = &ReferenceType{}
@@ -6113,10 +6115,6 @@ func (t *ReferenceType) Map(gauge common.MemoryGauge, typeParamMap map[*TypePara
 	return f(NewReferenceType(gauge, t.Authorization, mappedType))
 }
 
-func (t *ReferenceType) GetMembers() map[string]MemberResolver {
-	return t.Type.GetMembers()
-}
-
 func (t *ReferenceType) isValueIndexableType() bool {
 	referencedType, ok := t.Type.(ValueIndexableType)
 	if !ok {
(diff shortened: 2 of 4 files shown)


runtime/interpreter/value.go
```diff
@@ -20311,6 +20311,17 @@ func (v *EphemeralReferenceValue) GetMember(
...
 	self := v.MustReferencedValue(interpreter, locationRange)
 
+	switch name {
+	case sema.ReferenceTypeDereferenceFunctionName:
+		return NewHostFunctionValue(
+			interpreter,
+			sema.ReferenceDereferenceFunctionType(v.BorrowedType),
+			func(invocation Invocation) Value {
+				return v.Value
+			},
```
darkdrag00nv2, line 20321: The implementation of `dereference` looks straightforward. Is this correct or am I missing something?

bluesign: shouldn't this create new Value ?

turbolent:
@darkdrag00nv2 yeah, dereferencing needs to be implemented for both `EphemeralReferenceValue` and `StorageReferenceValue`.

Dereferencing should be supported at the type level for references in general (`&T`), for all primitive types (e.g. `Int` -> `&Int`), or containers of primitive types (e.g. `[Int]` -> `&[Int]`). At the value level there are two different kinds of values that both represent values of reference types. Basically, `let v: &T` might either hold be a `EphemeralReferenceValue` or a `StorageReferenceValue` at run-time.

runtime/interpreter/value.go
turbolent: Yeah, this looks right. The copy should be done automatically by the interpreter. Maybe test this via, e.g. 

```cadence
let ints = [1, 2, 3]
let ref: &[Int] = &ints
let ints2 = *ref // not sure what the syntax will be
ints.append(4)
assert(ints == [1, 2, 3, 4])
assert(ints2 == [1, 2, 3])
ints2.append(5)
assert(ints == [1, 2, 3, 4])
assert(ints2 == [1, 2, 3, 5])
```

bluesign: I was thinking more like: 

```cadence
let x: &[Int] = &[1]
x.dereference().append(2)
``

**Comment by darkdrag00nv2** [comment]
@turbolent You've mentioned primitive types and containers of primitive types.

Could we not define it for all subtypes of `AnyStruct`? Basically ignoring all resource types. Do you see any complications/extra-work to support them, say for Struct?

File: runtime/interpreter/value.go (hunk 1 above)
Comment (turbolent) [reply runtime/interpreter/value.go:20321]: Yeah, this needs a `Transfer` to the stack, i.e. 
```go
.Transfer(
	interpreter,
	locationRange,
	atree.Address{},
	false,
	nil,
	nil,
)
```

turbolent: Great point @bluesign to test by operating on the returned value directly, my example above, `let ints2 = ref.dereference()` was adding a copy 👍

bluesign:
> Could we not define it for all subtypes of `AnyStruct`? Basically ignoring all resource types. Do you see any complications/extra-work to support them, say for Struct?

Only edge case I see is attachments ( Contracts are covered by transfer ), attachments would allow detaching and attaching to new struct. ( which we don't allow currently )

SupunS:
Not sure why this got closed. This should now be opened against the master (https://github.com/onflow/cadence/pull/2971)

dsainati1:
Same, it seems as though it was closed automatically when `feature/stable-cadence` was merged into `master`. I can't re-open it for some reason though.

PR #2946, opened 2023-11-16
Include interface conformances when computing the supported entitlements of an interface 

dsainati1:
Address TODO in code

______

### DIFF
--- a/runtime/tests/checker/entitlements_test.go
+++ b/runtime/tests/checker/entitlements_test.go
@@ -5561,6 +5561,33 @@ func TestCheckEntitlementConditions(t *testing.T) {
 
 		assert.NoError(t, err)
 	})
+
+	t.Run("result value inherited interface entitlement resource", func(t *testing.T) {
+		t.Parallel()
+		_, err := ParseAndCheck(t, `
+		entitlement X
+		entitlement Y
+		resource interface I {
+			access(X) view fun foo(): Bool {
+				return true
+			}
+		}
+		resource interface J: I {
+			access(Y) view fun bar(): Bool {
+				return true
+			}
+		}
+		fun bar(r: @{J}): @{J} {
+			post {
+				result.foo(): ""
+				result.bar(): ""
+			}
+			return <-r
+		}
+		`)
+
+		assert.NoError(t, err)
+	})
 }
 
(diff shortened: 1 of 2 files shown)


runtime/tests/checker/entitlements_test.go
```diff
@@ -5561,6 +5561,33 @@ func TestCheckEntitlementConditions(t *testing.T) {
...
+			access(X) view fun foo(): Bool {
+				return true
+			}
+		}
+		resource interface J: I {
+			access(Y) view fun bar(): Bool {
+				return true
+			}
+		}
+		fun bar(r: @{J}): @{J} {
```
turbolent, line 5580: Could you please also add a test case for using a reference (non-owned case)? 

Is there also a test for the case where there is just a single function (e.g. `foo`), e.g.

```cadence
resource interface I {
	access(X) view fun foo(): Bool
}

resource interface J: I {
	access(Y) view fun foo(): Bool
}
```

dsainati1: > Could you please also add a test case for using a reference (non-owned case)?

No, this behavior does not occur for references; this only happens when `result` is created as a reference to a returned resource value. 

The other case I have added as a test.

turbolent: Ah, I meant just in general, ignoring post-conditions/`result`: The changes to `InterfaceType.SupportedEntitlements` are not specific to conditions I assume, but affect entitlement checking in general, right?

It would be great to also have tests that ensure that entitlement work in interface inheritance in general. Maybe these already exist, in which case there's nothing to do, just wanted to ensure they exist.

For example, it would be great to check that the example above is rejected (inheriting interface cannot escalate entitlements)

dsainati1: Oh, yes these tests all already exist. This PR only changes `SupportedEntitlements`, which is only used atm in the computation of the `result` variable in conditions. It will also be used as part of the entitlements migration, but that will be tested in the PR that actually adds that code.

turbolent (commented):
@SupunS Could you please have a look, and see if this change is correct, given you have the most knowledge about the new interface inheritance feature implementation

SupunS (approved):
👍

Issue #2709, opened 2023-08-09
Migration - Entitlements


j1010001:

j1010001:
Remaining decision to make:
How do we deal with contracts that are not updated in time for migration during network upgrade ?
The way this migration is implemented depends on the decision - we could do some kind of lazy-migration at runtime.

dsainati1:
https://github.com/onflow/cadence/pull/2705 and https://github.com/onflow/cadence/pull/2705 implement most of this, the remaining work here is just to hook up this functionality to the new migration framework, and pull in contracts as necessary to typecheck the new values.

turbolent:
@dsainati1 Do you think we can close this know as done, or is there anything left to be done here?

dsainati1:
Yeah! I’ll close it.

PR #2951, opened 2023-11-21
Entitlements Migration

dsainati1:
Adds the entitlements migration as detailed in https://github.com/onflow/flips/pull/95. 

Part of https://github.com/onflow/cadence/issues/2709

______


dsainati1:
This is currently in a draft state, I need to add more tests, especially for the cases where contracts are being loaded from multiple different locations and addresses, and for the case where one or more contracts has not yet been updated and thus fails to typecheck.

migrations/entitlements/migration.go
```diff
@@ -0,0 +1,90 @@
...
+	return "EntitlementsMigration"
+}
+
+// Converts its input to an entitled type according to the following rules:
+// * `ConvertToEntitledType(&T) ---> auth(Entitlements(T)) &T`
+// * `ConvertToEntitledType(Capability<T>) ---> Capability<ConvertToEntitledType(T)>`
+// * `ConvertToEntitledType(T?) ---> ConvertToEntitledType(T)?
+// * `ConvertToEntitledType(T) ---> T`
+// where Entitlements(I) is defined as the result of T.SupportedEntitlements()
+func ConvertToEntitledType(t sema.Type) sema.Type {
```
dsainati1, line 47: This defines the main logic for converting a "pre-entitlements" reference type into a "post-entitlements" one

runtime/interpreter/interpreter.go
```diff
@@ -5474,3 +5468,116 @@ func (interpreter *Interpreter) withResourceDestruction(
 
 	f()
 }
+
+// Converts the input value into a version compatible with the new entitlements feature,
+// with the same members/operations accessible on any references as would have been accessible in the past.
+// Modifies the input `v` in place.
+// This is used for migrations, but must be located in this package because some of the fields we wish to modify are
+// privately scoped to this package
+func (interpreter *Interpreter) ConvertValueToEntitlements(
```
dsainati1, line 5477: This defines the main logic of the migration: converting a value from before entitlements into a post-entitlements value

SupunS: Can this be moved to `migrations/entitlements/migration.go`?

dsainati1: Unfortunately it uses some fields that are privately scoped to the interpreter package

turbolent:
Took feedback notes during the review session, but had forgotten to post them:

- The type migration needs to be recursive to handle nested cases, e.g. `Capability<&[T{I}]>`
- Refactor the mutation-based migration to return new values instead
- Use the existing value migration code (`migrations` package). Might need to get extended for types?
- Follow up: Add a "FT-like" example test case: Deploy old "FT-like" contract, create some values, run migration. We could maybe do this with the Emulator

runtime/interpreter/interpreter.go
dsainati1: This can be moved now with later changes!

dsainati1:
Of the feedback identified in the review session, the only one left is:
* Add a "FT-like" example test case: Deploy old "FT-like" contract, create some values, run migration. We could maybe do this with the Emulator

We can do this once the emulator is updated to the newest Stable Cadence release

migrations/entitlements/migration.go
```diff
@@ -0,0 +1,240 @@
...
+			return true
+		})
+
+		newDict := interpreter.NewDictionaryValue(
+			inter,
+			interpreter.EmptyLocationRange,
+			dictionaryStaticType,
+			values...,
+		)
+		return newDict
```
SupunS, line 212: not sure, maybe transferring the value and changing the static-type could be sufficient as well (could avoid having to manually iterate and copy).
e.g:
```go
transferredDictionary := v.Transfer(
	interpreter,
	EmptyLocationRange,
	atree.Address{},
	true,
	nil,
	nil,
).(*DictionaryValue)

transferredDictionary.Type = dictionaryStaticType
transferredDictionary.semaType = entitledDictionaryType
return transferredDictionary
```

dsainati1: Tried this and ran into the same crash :/

migrations/entitlements/migration.go
```diff
@@ -0,0 +1,245 @@
...
+		case *interpreter.VariableSizedStaticType:
+			convertLegacyStaticType(t.Type)
+		case *interpreter.ConstantSizedStaticType:
+			convertLegacyStaticType(t.Type)
+		case *interpreter.DictionaryStaticType:
+			convertLegacyStaticType(t.KeyType)
+			convertLegacyStaticType(t.ValueType)
+		case *interpreter.OptionalStaticType:
+			convertLegacyStaticType(t.Type)
+		}
```
SupunS, line 147: Maybe also convert `LegacyType` of `IntersectionStaticType`

SupunS (approved):

migrations/entitlements/migration.go
dsainati1: This case is impossible; the instersection only has interfaces in it, and the legacy type of the `IntersectionStaticType` is only ever a composite

migrations/entitlements/migration.go
```diff
@@ -0,0 +1,245 @@
...
+		return interpreter.NewEphemeralReferenceValue(
+			inter,
+			staticAuthorization,
+			convertedValue,
+			entitledReferenceType.Type,
+			interpreter.EmptyLocationRange,
+		)
+
+	case *interpreter.StorageReferenceValue:
```
turbolent, line 176: Same as above: These two cases, ephemeral reference values and storage reference values, should be unreachable

migrations/entitlements/migration.go
turbolent: What is the crash?

turbolent (commented):
Looks good! Just some last suggestions and questions

migrations/entitlements/migration.go
```diff
@@ -0,0 +1,245 @@
...
+			inter,
+			referenceValue.Authorization,
+			interpreter.ConvertSemaToStaticType(inter, referenceValue.BorrowedType),
+		)
+	case *interpreter.StorageReferenceValue:
+		staticType = interpreter.NewReferenceStaticType(
+			inter,
+			referenceValue.Authorization,
+			interpreter.ConvertSemaToStaticType(inter, referenceValue.BorrowedType),
+		)
```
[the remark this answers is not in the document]
dsainati1: These won't be exercised in "real" migrations, but they are here for testing purposes; it's a lot easier to construct an ephemeral reference value to test the creation of entitlements.

migrations/entitlements/migration.go
dsainati1: These are here for the same reason as above.

migrations/entitlements/migration.go
dsainati1: Previously this was crashing in the atree code, but it seems like with the changes to the migration framework since then this no longer crashes. However this does still fail tests; the `MigrateDictOfValues` test specifically loads the dictionary post-migration and the static type of the dict is not properly converted using this code, whereas creating a new dictionary succeeds.

migrations/entitlements/migration.go
turbolent: Makes sense! Could you please leave a comment in the code?

turbolent:
@SupunS Do we have to use the same "trick" that we used in the type value migration, which forces the re-hashing / removal + reinsertion, in this migration?

SupunS:
ah, that's a good flag. It is currently handled in the migration framework: https://github.com/onflow/cadence/blob/f5f26f6b21df4b040f90bbf354bc7eb4ab74aa36/migrations/migration.go#L243-L246

But might need to make a small adjustment to return a 'legacy reference type' for the reference type: https://github.com/onflow/cadence/blob/f5f26f6b21df4b040f90bbf354bc7eb4ab74aa36/migrations/migration.go#L353-L357

@dsainati1 A way to test this is to have a type-value as a dictionary key, and the inner type of that type-value should get migrated by the entitlement migrations.

dsainati1:
> A way to test this is to have a type-value as a dictionary key, and the inner type of that type-value should get migrated by the entitlement migrations.

Added this test, which passes with no code changes.

migrations/entitlements/migration.go
turbolent: Interesting. I guess Cadence (atree?) does not currently support changes to the container's static type. 

For now, recreating the container works and is probably sufficient, but we might run into problems during the migration of large real-world containers, and might need to go back and address this overhead.

turbolent:
Let's double check https://discord.com/channels/613813861610684416/1108479699732152503/1194054315644571750,
i.e. wait for https://discord.com/channels/613813861610684416/1194054315644571750/1194066844479393952

(sorry for being overly cautious with the migration code)

SupunS:
I think we can merge this one. Any changes required can/will be added in https://github.com/onflow/cadence/pull/3013

turbolent (approved):
Great work! Sorry it took so long to get in

migrations/migration.go
```diff
@@ -273,38 +268,31 @@ func (m *StorageMigration) migrateNestedValue(
...
-		for _, migration := range valueMigrations {
-			converted, err := m.migrate(migration, addressPath, value)
-
-			if err != nil {
-				if reporter != nil {
-					reporter.Error(addressPath, migration.Name(), err)
-				}
-				continue
+	for _, migration := range valueMigrations {
+		converted, err := migration.Migrate(addressPath, value, m.interpreter)
```
SupunS, line 274: `m.migrate` must be called here to gracefully handle runtime panics.

```suggestion
		converted, err := m.migrate(migration, addressPath, value)
```

migrations/entitlements/migration_test.go
```diff
@@ -0,0 +1,2109 @@
...
+	`)
+
+	saveValues := []byte(`
+		import C from 0x1	
+
+		transaction {
+			prepare(signer: auth(Storage, Capabilities) &Account) {
+				let r1 <- C.makeR()
+				let r2 <- C.makeR()
+				let rType = ReferenceType(entitlements: [], type: r1.getType())!
```
SupunS, line 1960: I'm not sure whether this test exercises the issue we are trying to cover. The values should be stored using the 'older' version of cadence, but here it stores using Cadence 1.0 (`ReferenceType(entitlements: [], type: r1.getType())!` is the new format of reference with the entitlements), right?

dsainati1: Correct yeah. I don't think there's a way to create an old reference type within the new version of Cadence, so this exact case I think will need to be tested via the emulator.

SupunS: Makes sense. I have added a test in the FVM for this particular case, but would need this PR to be merged.

PR #2958, opened 2023-11-30
Improve resource-reference tracking

SupunS:
## Description

Currently, in reference tracking, the mapping is kept for storageID against the referenced resource. When the resource is moved, it goes and updates the inner atree value of all such tracked resources matching the storageID. This can lead to the revival of invalidated resources.

This PR changes the bookkeeping to track the 'reference' itself (i.e: storageID against the reference) instead of the referenced value.

______


runtime/interpreter/interpreter_expression.go
```diff
@@ -303,7 +303,19 @@ func (interpreter *Interpreter) VisitIdentifierExpression(expression *ast.Identi
...
+
+	resourceKindedValue, ok := result.(ResourceKindedValue)
+	if ok && resourceKindedValue.isInvalidatedResource(interpreter) {
+		panic(InvalidatedResourceError{
+			LocationRange: LocationRange{
+				Location:    interpreter.Location,
+				HasPosition: expression,
+			},
+		})
+	}
```
SupunS, line 316: Perform this check at the end of each expression evaluation. i.e: only in one place.
It is more foolproof this way than having it in multiple places which is easy to miss.

runtime/interpreter/value.go
```diff
@@ -1785,10 +1782,6 @@ func (v *ArrayValue) Destroy(interpreter *Interpreter, locationRange LocationRan
 
 	config := interpreter.SharedState.Config
 
-	if config.InvalidatedResourceValidationEnabled {
-		v.checkInvalidatedResourceUse(interpreter, locationRange)
-	}
```
SupunS, line 1790: Like mentioned above, since we do the check after evaluating each expression, these are now redundant and can be removed

runtime/interpreter/interpreter.go
```diff
@@ -246,7 +246,7 @@ type Storage interface {
 	CheckHealth() error
 }
 
-type ReferencedResourceKindedValues map[atree.StorageID]map[ReferenceTrackedResourceKindedValue]struct{}
+type ReferencedResourceKindedValues map[atree.StorageID]map[*EphemeralReferenceValue]struct{}
```
dsainati1, line 249: Do we not need to track invalidation of `StorageReferenceValue`s?

SupunS: No, storage references are always pointing to the same storage location. Storage references don't keep the value inside the reference (like ephemeral references do), rather always looked up from the storage, every time the value is used. So if the value is changed in the storage, it will automatically reflect when the storage look-up happens.

runtime/interpreter/value.go
```diff
@@ -19952,13 +19715,15 @@ func NewUnmeteredEphemeralReferenceValue(
...
-	gauge common.MemoryGauge,
+	interpreter *Interpreter,
 	authorized bool,
 	value Value,
 	borrowedType sema.Type,
 ) *EphemeralReferenceValue {
-	common.UseMemory(gauge, common.EphemeralReferenceValueMemoryUsage)
-	return NewUnmeteredEphemeralReferenceValue(authorized, value, borrowedType)
+	common.UseMemory(interpreter, common.EphemeralReferenceValueMemoryUsage)
+	ref := NewUnmeteredEphemeralReferenceValue(authorized, value, borrowedType)
```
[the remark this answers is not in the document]
SupunS: tat's a good flag. The `NewUnmeteredEphemeralReferenceValue` is currently mostly (only) used for tests, but think I will just move the tracking to the `NewUnmeteredEphemeralReferenceValue` so even in the future if this was used in a non-test use-case it would be safe.

runtime/interpreter/value_function.go
```diff
@@ -341,9 +342,22 @@ func NewBoundFunctionValue(
 
 	common.UseMemory(interpreter, common.BoundFunctionValueMemoryUsage)
 
+	// Since 'self' work as an implicit reference, create an explicit one and hold it.
```
turbolent, line 345: 👍

runtime/interpreter/interpreter_expression.go
```diff
@@ -1251,14 +1251,15 @@ func (interpreter *Interpreter) VisitReferenceExpression(referenceExpression *as
...
+		// This path shouldn't be reachable. This is only a defensive step
+		// to ensure references are properly created/tracked.
 		if someValue, ok := result.(*SomeValue); ok {
 			locationRange := LocationRange{
 				Location:    interpreter.Location,
 				HasPosition: referenceExpression.Expression,
 			}
 			innerValue := someValue.InnerValue(interpreter, locationRange)
-			interpreter.maybeTrackReferencedResourceKindedValue(innerValue)
+			return NewEphemeralReferenceValue(interpreter, typ.Authorized, innerValue, typ.Type)
```
turbolent, line 1262: Is this a bugfix? Previously, this case was falling through, still returning a reference to a Some (https://github.com/onflow/cadence/pull/2958/files#diff-484f7e68f1a1b9f96531632abacc84c131416a281cb29a502ac3b98c041ddb78L1265). Now a reference to the inner value is returned.

runtime/tests/interpreter/resources_test.go
```diff
@@ -2192,7 +2192,7 @@ func TestInterpreterResourcePostCondition(t *testing.T) {
       struct interface Receiver {
           pub fun deposit(from: @S) {
               post {
-                  from != nil: ""
+                  from != nil: ""  // This be an error. Resource is destroyed at this point
```
turbolent, line 2195:
```suggestion
                  from != nil: ""  // This is an error. Resource is destroyed at this point
```

turbolent (commented):
Great work! 👏 🧱

runtime/interpreter/interpreter_expression.go
```diff
@@ -960,15 +972,6 @@ func (interpreter *Interpreter) visitInvocationExpressionWithImplicitArgument(in
...
 
-	// Bound functions
-	if boundFunction, ok := function.(BoundFunctionValue); ok && boundFunction.Self != nil {
-		self := *boundFunction.Self
-		if resource, ok := self.(ReferenceTrackedResourceKindedValue); ok {
-			storageID := resource.StorageID()
-			interpreter.trackReferencedResourceKindedValue(storageID, resource)
-		}
-	}
-
```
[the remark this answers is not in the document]
SupunS: yes, correct.

runtime/interpreter/interpreter_expression.go
SupunS: Not really a bug fix. This path isn't reachable at runtime, was there only as a defensive check for proper resource tracking. This is still the case: By returning a reference (to the inner value), it makes sure the inner value's reference is tracked (in the previous revision as well, it was the inner value that was tracked). So this keeps the same old behaviour.

turbolent (approved):
👏

dsainati1 (approved):

runtime/interpreter/interpreter.go
SupunS: Do you mean bound functions for storage references? In that case - yes.
(added a test https://github.com/onflow/cadence/pull/2958/commits/bfdd84a72a72cea758cf310be50069e3de562e60 to verify)

PR #2959, opened 2023-12-01
Disallow InclusiveRange<T> if T is a non-leaf integer

### LINKED ISSUE #2885: Disallow covariant subtyping in checker
### Issue to be solved

Came out of: https://github.com/onflow/cadence/pull/2523/files#r1361276644

> Sorry I forgot some of the details, is it possible to create an `InclusiveRange<Integer>` in cadence?
> i.e: Is following allowed?
> ```
> let a: Integer = 0
> let b: Integer = 10
> var range: InclusiveRange<Integer> = InclusiveRange(a, b)
>```

> It seems not just the constructor, the checker also allows covariant subtyping:
> e.g:
>
> ```go
> let a: InclusiveRange<Int> = InclusiveRange(0, 10)
> let b: InclusiveRange<Integer> = a
> ```
> This would need to be fixed in the IsSubType() method, which would fix it universally (won't need to special case in multiple places)

darkdrag00nv2:
Closes #2885

## Description

Disallow InclusiveRange<T> if T is a non-leaf integer. At the moment, the two non-leaf integer types are - `Integer` and `SignedInteger`.

______

### DIFF
--- a/runtime/sema/errors.go
+++ b/runtime/sema/errors.go
@@ -3735,6 +3735,29 @@ func (e *MissingTypeArgumentError) Error() string {
 	return fmt.Sprintf("non-optional type argument %s missing", e.TypeArgumentName)
 }
 
+// InvalidTypeArgumentError
+
+type InvalidTypeArgumentError struct {
+	TypeArgumentName string
+	Details          string
+	ast.Range
+}
+
+var _ SemanticError = &InvalidTypeArgumentError{}
+var _ errors.UserError = &InvalidTypeArgumentError{}
+
+func (e *InvalidTypeArgumentError) isSemanticError() {}
+
+func (*InvalidTypeArgumentError) IsUserError() {}
+
+func (e *InvalidTypeArgumentError) Error() string {
+	return fmt.Sprintf("type argument %s invalid", e.TypeArgumentName)
+}
+
+func (e *InvalidTypeArgumentError) SecondaryError() string {
+	return e.Details
+}
+
 // TypeParameterTypeInferenceError
 
 type TypeParameterTypeInferenceError struct {
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -277,7 +277,7 @@ type LocatedType interface {
 type ParameterizedType interface {
 	Type
 	TypeParameters() []*TypeParameter
-	Instantiate(typeArguments []Type, report func(err error)) Type
+	Instantiate(typeArguments []Type, report func(err error), astRange *ast.Range) Type
 	BaseType() Type
 	TypeArguments() []Type
 }
@@ -288,6 +288,7 @@ func MustInstantiate(t ParameterizedType, typeArguments ...Type) Type {
 		func(err error) {
 			panic(errors.NewUnexpectedErrorFromCause(err))
 		},
+		nil,
 	)
 }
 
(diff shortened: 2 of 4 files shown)


runtime/sema/errors.go
```diff
@@ -3735,6 +3735,29 @@ func (e *MissingTypeArgumentError) Error() string {
...
+type InvalidTypeArgumentError struct {
+	TypeArgumentName string
+	Details          string
+	ast.Range
+}
+
+var _ SemanticError = &InvalidTypeArgumentError{}
+var _ errors.UserError = &InvalidTypeArgumentError{}
+
+func (e *InvalidTypeArgumentError) isSemanticError() {}
```
turbolent, line 3749:
```suggestion
func (InvalidTypeArgumentError) isSemanticError() {}
```

runtime/sema/type.go
```diff
@@ -277,7 +277,7 @@ type LocatedType interface {
 type ParameterizedType interface {
 	Type
 	TypeParameters() []*TypeParameter
-	Instantiate(typeArguments []Type, report func(err error)) Type
+	Instantiate(typeArguments []Type, report func(err error), astRange *ast.Range) Type
```
turbolent, line 280: Currently, the full range of all type arguments is constructed and passed to the instantiation function.

We  should have the ranges of the individual type arguments already available, so maybe just pass in the AST type arguments to the function, and then the instantiation function can report errors for individual type arguments, instead just for the full range of all arguments.

turbolent (commented):
Nice! 👌

runtime/tests/checker/range_value_test.go
```diff
@@ -377,3 +377,48 @@ func TestCheckInclusiveRangeConstructionInvalid(t *testing.T) {
...
+	}
+
+	test := func(t *testing.T, ty sema.Type) {
+		t.Run(fmt.Sprintf("InclusiveRange<%s>", ty), func(t *testing.T) {
+			t.Parallel()
+
+			_, err := ParseAndCheckWithOptions(t, fmt.Sprintf(`
+				let a: %[1]s = 0
+				let b: %[1]s = 10
+				var range: InclusiveRange<%[1]s> = InclusiveRange<%[1]s>(a, b)
```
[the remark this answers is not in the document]
darkdrag00nv2: Yes, I missed that point. Without further changes, the expression `var range = InclusiveRange<Integer>(a, b)` didn't return any checker error since the RHS is a function invocation. The name `InclusiveRange` is used in both the type as well as the constructor function.

I had to implement the same check in the constructor function as well. This required me to implement support for custom type parameter validation in `FunctionType`. I did that as part of 7b65121e7.

runtime/sema/type.go
darkdrag00nv2: Done. Added that in the `Instantiate` function as well as in the constructor function.

Note that the type arguments are optional in the constructor function. So, I fallback to the invocation range in that case. 733cf4706

runtime/tests/checker/range_value_test.go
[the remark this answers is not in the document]
turbolent: Very nice @darkdrag00nv2!

runtime/sema/check_invocation_expression.go
```diff
@@ -459,6 +459,11 @@ func (checker *Checker) checkInvocation(
 		}
 	}
 
+	invocationRange := ast.NewRangeFromPositioned(
+		checker.memoryGauge,
+		invocationExpression,
+	)
```
turbolent, line 465: The range is now always allocated, but only used if the function type has a `ArgumentExpressionsCheck` or `TypePrametersCheck`. Maybe make this a lazy/memoized, so it's only allocated if needed, once:


```suggestion
	// Compute the invocation range, once, if needed
	getInvocationRange := func() func() ast.Range {
		var invcocationRange ast.Range
		return func() ast.Range {
			if invocationRange == ast.EmptyRange {
				invocationRange = ast.NewRangeFromPositioned(
					checker.memoryGauge,
					invocationExpression,
				)
			}
			return invocationRange
		}
	}()
```

turbolent (approved):
Great work!

turbolent (approved):

SupunS (approved):
👌

PR #2962, opened 2023-12-04
Concretize the definition of Primitive type

### LINKED ISSUE #2929: Define what a "Primitive Type" is in Cadence
### Issue to be solved

Some logic (like in https://github.com/onflow/cadence/pull/2928) relies on a soft definition of certain types (e.g. string, int, bool) in Cadence being "primitive", as opposed to a complex type like an array, dictionary or composite. We should concretize this by specifying exactly which types are primitive vs concrete. 

### Suggested Solution

_No response_

darkdrag00nv2:
Closes #2929

## Description

This PR concretizes the definition of Primitive type by introducing a `IsPrimitiveType` function on `sema.Type`.

The following types have been marked as primitive:
- `Number` types 
- `Character`
- `Bool`
- `String`
- `Address`
- `Path`
- `Void`
- `Optional<T>` - if `T` is primitive
______


runtime/sema/gen/main.go
```diff
@@ -1583,12 +1584,14 @@ func simpleTypeLiteral(ty *typeDecl) dst.Expr {
 	//}
 
 	isResource := ty.compositeKind == common.CompositeKindResource
+	_, isPrimitive := sema.GeneratedPrimitiveSimpleTypes[ty.typeName]
```
turbolent, line 1587: Similar to how other "kind"s are indicated through conformances in declarations of built-in types which produce Go code, like if a type is storable via conforming to the `Storable` type (also, `Equatable`, `Comparable`, etc), we might consider introducing a `Primitive` type in the code generator (which later then also could be introduced "properly" in the type checker.

runtime/sema/type.go
```diff
@@ -669,6 +674,10 @@ func (t *OptionalType) IsResourceType() bool {
 	return t.Type.IsResourceType()
 }
 
+func (t *OptionalType) IsPrimitiveType() bool {
+	return false
```
turbolent, line 678: We might want to also consider optional types which wrap a primitive type to in turn be primitive, too

turbolent (commented):
Nice!

runtime/sema/type.go
darkdrag00nv2: Done - https://github.com/onflow/cadence/pull/2962/commits/61e79f1c1b35bfc2e029b43f597e5916d6ce738a.

runtime/sema/gen/main.go
darkdrag00nv2: Done in https://github.com/onflow/cadence/pull/2962/commits/7b3805cd5b839f47c96336c13251b3d6b10cdd7b.

PR #2964, opened 2023-12-05
Sync stable-cadence branch with master

SupunS:
## Description

$subject

______


runtime/runtime_test.go
```diff
@@ -7979,8 +7994,141 @@ func TestRuntimeInvalidatedResourceUse(t *testing.T) {
...
+		},
+	)
+	RequireError(t, err)
+
+	assertRuntimeErrorIsUserError(t, err)
+
+	var parserError parser.Error
+	require.ErrorAs(t, err, &parserError)
 
+	assert.IsType(t, &parser.CustomDestructorError{}, parserError.Errors[0])
```
SupunS, line 8131: This and the below ones are no longer valid, since destructors are no longer supported and those edge cases were only possible to be reached in the destructor. So updated the test to assert the new static error (`CustomDestructorError`).

We could technically move these tests to checker tests, but in future, if/when custom destructors are allowed, these would become valid tests again, so keeping it here would make life easier + won't miss out on adding these back again.

runtime/sema/check_interface_declaration.go
```diff
@@ -33,6 +33,12 @@ import (
 // through `declareInterfaceMembers`.
 func (checker *Checker) VisitInterfaceDeclaration(declaration *ast.InterfaceDeclaration) (_ struct{}) {
 
+	wasInInterface := checker.inInterface
+	checker.inInterface = true
+	defer func() {
+		checker.inInterface = wasInInterface
+	}()
```
SupunS, line 40: This was removed in https://github.com/dapperlabs/cadence-internal/pull/179, but had to add this back again, since entitlements were relying on it.
We should double-check if an issue similar to the original one also exists with entitlements. cc: @dsainati1

runtime/tests/interpreter/resources_test.go
```diff
@@ -2137,158 +2137,6 @@ func TestInterpretResourceDestroyedInPreCondition(t *testing.T) {
 	require.ErrorAs(t, err, &interpreter.InvalidatedResourceError{})
 }
 
-func TestInterpretInvalidReentrantResourceDestruction(t *testing.T) {
```
SupunS: Removed this test since this is no longer valid, after destructor removal

runtime/tests/interpreter/resources_test.go
```diff
@@ -2658,205 +2506,6 @@ func TestInterpretDefaultDestroyEventArgumentScoping(t *testing.T) {
 	require.Equal(t, interpreter.NewIntValueFromInt64(nil, 1), events[0].GetField(inter, interpreter.EmptyLocationRange, "x"))
 }
 
-func TestInterpretInnerResourceMove(t *testing.T) {
```
SupunS: Same here: Removed this test since this is no longer valid, after destructor removal

runtime/tests/interpreter/resources_test.go
```diff
@@ -2658,205 +2506,6 @@ func TestInterpretDefaultDestroyEventArgumentScoping(t *testing.T) {
...
-            }`,
-		)
-
-		_, err := inter.Invoke("main")
-		RequireError(t, err)
-		require.ErrorAs(t, err, &interpreter.UseBeforeInitializationError{})
-	})
-}
-
-func TestInterpretSetMemberResourceLoss(t *testing.T) {
```
SupunS: Removed this test since this is no longer valid, after destructor removal

runtime/tests/interpreter/resources_test.go
```diff
@@ -2658,205 +2506,6 @@ func TestInterpretDefaultDestroyEventArgumentScoping(t *testing.T) {
...
-       destroy rl
-    }`,
-	)
-
-	_, err := inter.Invoke("main")
-	RequireError(t, err)
-	require.ErrorAs(t, err, &interpreter.DestroyedResourceError{})
-}
-
-func TestInterpretValueTransferResourceLoss(t *testing.T) {
```
SupunS: Moved to runtime tests, since the custom destructors are not allowed, destruction of a resource can only be asserted from the emitted event.

runtime/tests/interpreter/resources_test.go
```diff
@@ -2934,43 +2583,14 @@ func TestInterpretVariableDeclarationEvaluationOrder(t *testing.T) {
 	)
 }
 
-func TestInterpretIfLetElseBranchConfusion(t *testing.T) {
```
SupunS: This gives a checker error now, so moved to checker tests

runtime/tests/interpreter/resources_test.go
```diff
@@ -3039,41 +2657,3 @@ func TestInterpretMovedResourceInSecondValue(t *testing.T) {
 	assert.Equal(t, 15, errorStartPos.Line)
 	assert.Equal(t, 53, errorStartPos.Column)
 }
-
-func TestInterpretOptionalBindingElseBranch(t *testing.T) {
```
SupunS: This gives a checker error now, so moved to checker tests

runtime/tests/checker/resources_test.go
```diff
@@ -9526,61 +9529,6 @@ func TestCheckInvalidationInCondition(t *testing.T) {
 				))
 				require.NoError(t, err)
 			})
-
-			t.Run("in nested type requirement, definite invalidation", func(t *testing.T) {
```
SupunS: No longer valid after the type requirement removal

turbolent (approved):
Perfect! Thank you for adding the comments with the reasoning for (re)moving the test cases 🙏

dsainati1 (approved):

Issue #2873, opened 2023-10-16
[Stable Cadence] References to references are not properly checked


turbolent:
### Current Behavior

It is currently possible to create nested references.

In Stable Cadence, this can be used to escalate privileges, by creating a reference that has more entitlements than the referenced reference.

Thank you to @oebeling for reporting this issue.

### Expected Behavior

References to references, and references to optional references should be either:
- Rejected in general
- Properly checked, by ensuring that the entitlements of the resulting reference is a subset of the available entitlements of the referenced reference.

i.e., the following programs should be rejected:

```cadence
let ref: &[String] = &["foo"]
let outerRef = &ref as auth(Mutate) &(&[String])
```

```cadence
let ref: &[String] = &["foo"]
let optRef: &[String]? = ref
let outerRef = &optRef as auth(Mutate) &(&[String])?
```

### Steps To Reproduce

See above

### Environment

```markdown
- Cadence version: Stable Cadence
```

PR #2965, opened 2023-12-05
throw error on the creation of a nested reference

dsainati1:
Part of https://github.com/onflow/cadence/issues/2873

Throw a runtime error when attempting to create an `EphemeralReferenceValue` whose underlying value is also an `EphemeralReferenceValue`.

______

### DIFF
--- a/runtime/contract_function_executor.go
+++ b/runtime/contract_function_executor.go
@@ -262,7 +262,7 @@ func (executor *interpreterContractFunctionExecutor) convertArgument(
 				authorization,
 				accountValue,
 				sema.AccountType,
-				interpreter.EmptyLocationRange,
+				locationRange,
 			)
 
 			return accountReferenceValue, nil
--- a/runtime/contract_function_executor.go
+++ b/runtime/contract_function_executor.go
@@ -262,6 +262,7 @@ func (executor *interpreterContractFunctionExecutor) convertArgument(
 				authorization,
 				accountValue,
 				sema.AccountType,
+				interpreter.EmptyLocationRange,
 			)
 
 			return accountReferenceValue, nil
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -752,6 +752,7 @@ func (interpreter *Interpreter) visitFunctionBody(
 	body func() StatementResult,
 	postConditions ast.Conditions,
 	returnType sema.Type,
+	declarationLocationRange LocationRange,
 ) Value {
 
 	// block scope: each function block gets an activation record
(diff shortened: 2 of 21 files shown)


runtime/contract_function_executor.go
```diff
@@ -262,6 +262,7 @@ func (executor *interpreterContractFunctionExecutor) convertArgument(
 				authorization,
 				accountValue,
 				sema.AccountType,
+				interpreter.EmptyLocationRange,
```
turbolent, line 265:
```suggestion
				locationRange,
```

runtime/interpreter/interpreter.go
```diff
@@ -832,6 +832,8 @@ func (interpreter *Interpreter) resultValue(returnValue Value, returnType sema.T
 				resultAuth(returnType),
 				returnValue.value,
 				optionalType.Type,
+				// result must be a resource to get here, so it will never be a reference
+				EmptyLocationRange,
```
turbolent, line 836: Here and below: Pass the proper location range to the function (`resultValue`) and use it

runtime/interpreter/interpreter.go
```diff
@@ -840,7 +842,8 @@ func (interpreter *Interpreter) resultValue(returnValue Value, returnType sema.T
 		}
 	}
 
-	return NewEphemeralReferenceValue(interpreter, resultAuth(returnType), returnValue, returnType)
+	// result must be a resource to get here, so it will never be a reference
+	return NewEphemeralReferenceValue(interpreter, resultAuth(returnType), returnValue, returnType, EmptyLocationRange)
```
turbolent, line 846: Here and for similar calls: Maybe chop down the long argument list to multiple lines

runtime/interpreter/value.go
```diff
@@ -20203,7 +20206,15 @@ func NewUnmeteredEphemeralReferenceValue(
 	authorization Authorization,
 	value Value,
 	borrowedType sema.Type,
+	locationRange LocationRange,
 ) *EphemeralReferenceValue {
+	if reference, isReference := value.(*EphemeralReferenceValue); isReference {
```
turbolent, line 20211: 👍

runtime/interpreter/value_accountcapabilitycontroller.go
```diff
@@ -316,6 +316,8 @@ func (v *AccountCapabilityControllerValue) ReferenceValue(
 		authorization,
 		account,
 		resultBorrowType.Type,
+		// okay to pass an empty range here because the account value is never a reference, so this can't fail
+		EmptyLocationRange,
```
turbolent, line 320: Here and below (e.g. `account.go`): Maybe pass the proper location range here.

Though that statement is true, because the location range is currently only used to report an error when the referenced value is a reference, that might not stay true in the future, the position information should be valid on its own.

runtime/stdlib/account.go
```diff
@@ -247,6 +247,8 @@ func NewAccountReferenceValue(
 		authorization,
 		account,
 		sema.AccountType,
+		// okay to pass an empty range here because the account value is never a reference, so this can't fail
```
turbolent, line 250: Same as above: Maybe pass the proper location range here

turbolent (commented):
Nice!

SupunS (approved):

Issue #2749, opened 2023-08-25
Order entitlements in entitlement sets in runtime types


dsainati1:
### Issue to be solved

As mentioned in https://github.com/onflow/cadence/issues/2681, the order of entitlements in an entitled runtime reference type should be ordered alphabetically

PR #2967, opened 2023-12-06
Test type order insignificance

### LINKED ISSUE #2681: Dictionaries key'd by type are order-dependent
### Current Behavior

If I have a dictionary key'd by `Type`, the order of their interface conformance matters:

```cadence
import FungibleToken from 0x1

pub fun main(): String {
  let t1 = Type<&{FungibleToken.Provider, FungibleToken.Balance}>()
  let t2 = Type<&{FungibleToken.Balance, FungibleToken.Provider}>()

  let dict: {Type: Bool} = {}
  dict[t1] = true

  assert(dict[t2] != nil, message: "key not found!")
  return "Done"
}
```

However, if I compare the two types, they are equal:

```cadence
import FungibleToken from 0x1

pub fun main(): String {
  let t1 = Type<&{FungibleToken.Provider, FungibleToken.Balance}>()
  let t2 = Type<&{FungibleToken.Balance, FungibleToken.Provider}>()

  assert(t1 == t2, message: "types are not equal")

  return "Done"
}
```

### Expected Behavior

I expect that two types which are equal should match to the same key in a dictionary

If I have a type `&{A, B}` and it is equivalent to `&{B, A}`, then my expectation is that a dictionary key'd on types should match regardless of their order

### Steps To Reproduce

Run this script on the flow emulator:

```cadence
import FungibleToken from 0xee82856bf20e2aa6

pub fun main(): String {
  let t1 = Type<&{FungibleToken.Provider, FungibleToken.Balance}>()
  let t2 = Type<&{FungibleToken.Balance, FungibleToken.Provider}>()

  // this will pass
  assert(t1 == t2, message: "types are not equal")

  let dict: {Type: Bool} = {}
  dict[t1] = true

  // this will fail
  assert(dict[t2] != nil, message: "key not found!")

  return "done"
}
```

### Environment

```markdown
- Cadence version: v1.3.2 (flow cli version)
- Network: Emulator
```

turbolent:
Work towards #2681

## Description

Test that the order of types is insignificant, even across multiple transactions (encoding/decoding), for intersection types and authorizations (entitlements). In both cases the types are sets.

This was previously fixed in #2750 and #2749.

______


runtime/storage_test.go
```diff
@@ -5541,3 +5541,180 @@ func TestRuntimeAccountIterationMutation(t *testing.T) {
...
+          }
+        `)
+
+		tx2 := []byte(`
+          import Test from 0x1
+
+          transaction {
+              prepare(account: auth(Storage) &Account) {
+
+                  let t1 = Type<&{Test.A, Test.B}>()
```
turbolent, line 5620: - Add a migration that migrates dictionary keys that are `Type` values, if the static type contains a restricted type (force re-hash and removal of old key / insert of new key) 
- Maybe we can force Go code to not use `==` through the Go code of embedding an incomparable, but zero size field 
    ```go
    type S struct {
        _ [0]func()
        /* ... */
    }
    ```

turbolent:
@SupunS Could you please have a look here?

SupunS (approved):
Sorry, I thought I already reviewed/approved this.

LGTM!

PR #2974, opened 2023-12-08
Add state migration for removing all values from private domain

turbolent:
⚠️ Depends on #2912 

## Description

After all links got migrated to capability controllers, remove all values from the `private` domain.

- Refactor the existing shared state migration code: Generalize the ability to migrate a path of an account, and make value conversion a particular path migrator that is commonly used
- Introduce a new path migrator which removes all values in the private domain

______


dsainati1 (approved):
So we are just deleting every capability that was linked via the private domain? Should we tell developers to prepare for this at all?

turbolent:
@dsainati1 In 1.0, the `private` domain is completely gone, and the concept of a private link is replaced by a capability controller. This migration is just removing the dead capability values which previously all got migrated to capability controllers in #2912

migrations/capcons/migration_test.go
```diff
@@ -1586,3 +1592,111 @@ func TestLinkMigration(t *testing.T) {
...
+		Interface: runtimeInterface,
+	})
+	require.NoError(t, err)
+
+	storeTestPathLinks(t, pathLinks, storage, inter)
+
+	storeTestAccountLinks(accountLinks, storage, inter)
+
+	err = storage.Commit(inter, false)
+	require.NoError(t, err)
```
SupunS, line 1662: Maybe also assert the size of the private domain storage before the migration

SupunS (approved):

migrations/capcons/migration_test.go
turbolent: Good idea! Added in 25fdcb41d3c5779b679431d8cb0fc14446ee419e

PR #2980, opened 2023-12-15
Use IsPrimitiveType to check default destructor params

### LINKED ISSUE #2929: Define what a "Primitive Type" is in Cadence

dsainati1:
Closes #2929 

Code reuse

______


turbolent (approved):
Nice!

PR #2982, opened 2023-12-16
Introduce dereference operator for references

### LINKED ISSUE #2803: Need a way get a dereferenced copy from a reference

darkdrag00nv2:
Closes #2803

## Description

______

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -19691,6 +19691,27 @@ type ReferenceValue interface {
 	ReferencedValue(interpreter *Interpreter, locationRange LocationRange, errorOnFailedDereference bool) *Value
 }
 
+func ReferenceTypeDereferenceFunctionValue(
+	inter *Interpreter,
+	borrowedType sema.Type,
+	value Value,
+) *HostFunctionValue {
+	return NewHostFunctionValue(
+		inter,
+		sema.ReferenceDereferenceFunctionType(borrowedType),
+		func(invocation Invocation) Value {
+			return value.Transfer(
+				invocation.Interpreter,
+				invocation.LocationRange,
+				atree.Address{},
+				false,
+				nil,
+				nil,
+			)
+		},
+	)
+}
+
 // StorageReferenceValue
 type StorageReferenceValue struct {
 	BorrowedType         sema.Type
@@ -19857,20 +19878,7 @@ func (v *StorageReferenceValue) GetMember(
 
 	switch name {
 	case sema.ReferenceTypeDereferenceFunctionName:
-		return NewHostFunctionValue(
-			interpreter,
-			sema.ReferenceDereferenceFunctionType(v.BorrowedType),
-			func(invocation Invocation) Value {
-				return self.Transfer(
-					invocation.Interpreter,
-					invocation.LocationRange,
-					atree.Address{},
-					false,
-					nil,
-					nil,
-				)
-			},
-		)
+		return ReferenceTypeDereferenceFunctionValue(interpreter, v.BorrowedType, self)
 	}
 
 	return interpreter.getMember(self, locationRange, name)
@@ -20245,20 +20253,7 @@ func (v *EphemeralReferenceValue) GetMember(
 ) Value {
 	switch name {
 	case sema.ReferenceTypeDereferenceFunctionName:
-		return NewHostFunctionValue(
-			interpreter,
-			sema.ReferenceDereferenceFunctionType(v.BorrowedType),
-			func(invocation Invocation) Value {
-				return v.Value.Transfer(
-					invocation.Interpreter,
-					invocation.LocationRange,
-					atree.Address{},
-					false,
-					nil,
-					nil,
-				)
-			},
-		)
+		return ReferenceTypeDereferenceFunctionValue(interpreter, v.BorrowedType, v.Value)
 	}
 
 	return interpreter.getMember(v.Value, locationRange, name)
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -19691,24 +19691,19 @@ type ReferenceValue interface {
 	ReferencedValue(interpreter *Interpreter, locationRange LocationRange, errorOnFailedDereference bool) *Value
 }
 
-func ReferenceTypeDereferenceFunctionValue(
+func DereferenceValue(
 	inter *Interpreter,
-	borrowedType sema.Type,
-	value Value,
-) *HostFunctionValue {
-	return NewHostFunctionValue(
+	locationRange LocationRange,
+	referenceValue ReferenceValue,
+) Value {
+	referencedValue := referenceValue.ReferencedValue(inter, locationRange, true)
+	return (*referencedValue).Transfer(
 		inter,
-		sema.ReferenceDereferenceFunctionType(borrowedType),
-		func(invocation Invocation) Value {
-			return value.Transfer(
-				invocation.Interpreter,
-				invocation.LocationRange,
-				atree.Address{},
-				false,
-				nil,
-				nil,
-			)
-		},
+		locationRange,
+		atree.Address{},
+		false,
+		nil,
+		nil,
 	)
 }
 
@@ -19875,11 +19870,6 @@ func (v *StorageReferenceValue) GetMember(
 	name string,
 ) Value {
 	self := v.mustReferencedValue(interpreter, locationRange)
-
-	switch name {
-	case sema.ReferenceTypeDereferenceFunctionName:
-		return ReferenceTypeDereferenceFunctionValue(interpreter, v.BorrowedType, self)
-	}
 
 	return interpreter.getMember(self, locationRange, name)
 }
@@ -20251,11 +20241,6 @@ func (v *EphemeralReferenceValue) GetMember(
 	locationRange LocationRange,
 	name string,
 ) Value {
-	switch name {
-	case sema.ReferenceTypeDereferenceFunctionName:
-		return ReferenceTypeDereferenceFunctionValue(interpreter, v.BorrowedType, v.Value)
-	}
-
 	return interpreter.getMember(v.Value, locationRange, name)
 }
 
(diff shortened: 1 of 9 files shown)


darkdrag00nv2:
Need to cover a few more types in the interpreter tests.

runtime/sema/type.go
```diff
@@ -6309,6 +6307,78 @@ func (t *ReferenceType) Resolve(typeArguments *TypeParameterTypeOrderedMap) Type
...
 
+func (t *ReferenceType) GetMembers() map[string]MemberResolver {
+	t.initializeMemberResolvers()
+	return t.memberResolvers
+}
+
+const ReferenceTypeDereferenceFunctionName = "dereference"
+
+const referenceTypeDereferenceFunctionDocString = `
+	Returns a copy of the reference value after dereferencing.
```
dsainati1, line 6318:
```suggestion
	Returns a copy of the referenced value after dereferencing.
```

dsainati1: IMO we should also add something to doc string about how this only works for primitive values

runtime/sema/type.go
```diff
@@ -6448,6 +6518,33 @@ func (t *AddressType) initializeMemberResolvers() {
 	})
 }
 
+func IsPrimitiveOrContainerOfPrimitive(ty Type) bool {
+	if ty.IsPrimitiveType() {
+		return true
+	}
+
+	// TODO: Do we also want to count Dictionary?
```
dsainati1, line 6526: Yes! Dictionaries should also be considered containers the same way arrays are

runtime/sema/type.go
```diff
@@ -6448,6 +6518,33 @@ func (t *AddressType) initializeMemberResolvers() {
...
+	switch ty.(type) {
```
dsainati1, line 6527:
```suggestion
	switch typedTy := ty.(type) {
```

dsainati1: This way you don't have to manually cast inside the `case`s below, or panic on an unreachable failure.

runtime/tests/interpreter/reference_test.go
```diff
@@ -1815,3 +1816,1073 @@ func TestInterpretReferenceToReference(t *testing.T) {
 		require.ErrorAs(t, err, &interpreter.NestedReferenceError{})
 	})
 }
+
+func TestInterpretReferenceDereference(t *testing.T) {
```
dsainati1, line 1820: Can you add a test for optional references using optional chaining? I.e. something of type `&Int?` being dereferenced using `ref?.dereference()`?

runtime/sema/type.go
darkdrag00nv2: So the main question is, do we want `Dictionary<PrimitiveKey, Non-PrimitiveValue>` or `Dictionary<Non-Primitive key, Primitive value` to also support `dereference`?

I guess we only want `Dictionary<primitive key, primitive value>` to support dereference but wanted to confirm.

dsainati1: Dictionary keys are already restricted to be subtypes of "Hashable Structs" (see `IsHashableStructType`), which similar to the Primitive type designation (and might actually be able to be refactored to use the new `IsPrimitiveType` method). Point is though, that I don't think you need to worry about the key types at all when checking whether a dictionary reference is dereference-able, just check the values.

runtime/sema/type.go
turbolent: Or even shadow: `ty := ty.(type)`

runtime/sema/errors.go
```diff
@@ -2755,6 +2755,29 @@ func (e *InvalidResourceOptionalMemberError) Error() string {
...
+var _ SemanticError = &InvalidMemberError{}
+var _ errors.UserError = &InvalidMemberError{}
+
+func (*InvalidMemberError) isSemanticError() {}
+
+func (*InvalidMemberError) IsUserError() {}
+
+func (e *InvalidMemberError) Error() string {
+	return fmt.Sprintf(
+		"%s `%s` is not available for the type",
```
turbolent, line 2775: Maybe add a reason here, i.e. add a `Reason string` field to the error type, and maybe include it if non-empty with `": %s"`

runtime/interpreter/value.go
```diff
@@ -19855,6 +19855,24 @@ func (v *StorageReferenceValue) GetMember(
...
+				return self.Transfer(
+					invocation.Interpreter,
+					invocation.LocationRange,
+					atree.Address{},
+					false,
+					nil,
+					nil,
+				)
+			},
+		)
```
turbolent, line 19873: Maybe refactor this into a function and use it here and in `EphemeralReferenceValue`, so it is ensured that the same code is used in both values

turbolent (commented):
Nice!

turbolent:
CI is reporting data races: https://github.com/onflow/cadence/actions/runs/7240009930/job/19722511487?pr=2982

runtime/sema/type.go
darkdrag00nv2: thanks, done.

runtime/sema/type.go
darkdrag00nv2: Done

runtime/sema/errors.go
darkdrag00nv2: Done

runtime/interpreter/value.go

runtime/tests/interpreter/reference_test.go
darkdrag00nv2: Done in 43d998901

runtime/sema/type.go
```diff
@@ -6309,6 +6307,84 @@ func (t *ReferenceType) Resolve(typeArguments *TypeParameterTypeOrderedMap) Type
...
+							ReferenceDereferenceFunctionType(t.Type),
+							referenceTypeDereferenceFunctionDocString,
+						)
+					},
+				},
+			},
+		}
+
+		for _, member := range members {
+			resolvers[member.name] = member.resolver
```
turbolent, line 6374: If there is already a member resolve, it will be overwritten here

turbolent (commented):
Looks good, great work!

In general I'm still not sure if this should be a magic function on references, or rather something else, like an operator.

For the functionality to be a magic function, the PR should also handle the case where a composite declares a member named the same as the magic function, as it seems like the current implementation shadows / makes it inaccessible. We might want to forbid declaring members which have that magic name. An operator would not have that problem.

turbolent:
We might want to use an operator here instead of / in addition to the method, e.g. the unary prefix operator `*`, like in Rust, Go, C, C++, etc.

turbolent:
@darkdrag00nv2 Not sure if you started yet, but if not: I'm half-way done and could finish and push it up here

darkdrag00nv2:
@turbolent Haven't started yet. You can push it. Thank you!

turbolent:
Refactored the `dereference` function to the prefix `*` operator.

@onflow/cadence could you please have a look?

dsainati1 (commented):
Now that `dereference` isn't a member function, it would be nice for `*` to have simple/clean behavior with respect to optional references. I.e. if we have a `&Int?` type, it would be nice for the `*` operator on that type to produce a `Int?`. Previously we could do this with `?.` chaining, but with the change to an operator we lose this functionality. Can we add this (and tests for it)?

turbolent:
@dsainati1 Would this simply be `*(optRef!)` ?

dsainati1:
Yeah, it would be nice for the `*` operator to implicitly "lift" itself to operate on optional values as well as normal ones.

turbolent:
@dsainati1 Sounds good! Let's get that added in a follow-up PR

runtime/parser/expression_test.go
SupunS, line 1: Can you maybe add a test for an expression like `a**b` to ensure it parses correctly / presidency is correct?
i.e: should result in `(a*(*b))`

runtime/sema/type.go
```diff
@@ -6448,6 +6448,25 @@ func (t *AddressType) initializeMemberResolvers() {
...
+		return IsPrimitiveOrContainerOfPrimitive(ty.Type)
+
+	case *ConstantSizedType:
+		return IsPrimitiveOrContainerOfPrimitive(ty.Type)
+
+	case *DictionaryType:
+		return IsPrimitiveOrContainerOfPrimitive(ty.ValueType)
+	}
+
+	return false
```
SupunS, line 6467: Just a nit-pick: calling for `IsPrimitiveType` for arrays and dictionaries is guranteed to return false.

```suggestion
	switch ty := ty.(type) {
	case *VariableSizedType:
		return IsPrimitiveOrContainerOfPrimitive(ty.Type)

	case *ConstantSizedType:
		return IsPrimitiveOrContainerOfPrimitive(ty.Type)

	case *DictionaryType:
		return IsPrimitiveOrContainerOfPrimitive(ty.ValueType)
		
	default:
	  	return ty.IsPrimitiveType()
	}
```

runtime/parser/expression_test.go
turbolent: Good idea, added in 01ea3f3860478c32eea8b09d3b0f60eafc3de1a0

SupunS (approved):
LGTM!

dsainati1 (approved):

PR #2985, opened 2023-12-20
Report bespoke error for restricted types parsed as type arguments

dsainati1:
See discussion in https://discord.com/channels/613813861610684416/1108479699732152503/1186788225365327952

## Description

Report a better error when legacy restricted types are encountered as type arguments
______

### DIFF
--- a/runtime/parser/errors.go
+++ b/runtime/parser/errors.go
@@ -318,3 +318,28 @@ func (e *CustomDestructorError) Error() string {
 func (e *CustomDestructorError) SecondaryError() string {
 	return "remove the destructor definition"
 }
+
+// RestrictedTypeError
+
+type RestrictedTypeError struct {
+	Pos ast.Position
+}
+
+var _ ParseError = &CustomDestructorError{}
+var _ errors.UserError = &CustomDestructorError{}
+
+func (*RestrictedTypeError) isParseError() {}
+
+func (*RestrictedTypeError) IsUserError() {}
+
+func (e *RestrictedTypeError) StartPosition() ast.Position {
+	return e.Pos
+}
+
+func (e *RestrictedTypeError) EndPosition(_ common.MemoryGauge) ast.Position {
+	return e.Pos
+}
+
+func (e *RestrictedTypeError) Error() string {
+	return "restricted types have been removed; replace with the concrete type or an equivalent intersection type"
+}
--- a/runtime/parser/errors.go
+++ b/runtime/parser/errors.go
@@ -322,7 +322,7 @@ func (e *CustomDestructorError) SecondaryError() string {
 // RestrictedTypeError
 
 type RestrictedTypeError struct {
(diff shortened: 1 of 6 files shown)


SupunS (approved):

runtime/parser/expression.go
```diff
@@ -601,7 +601,7 @@ func defineLessThanOrTypeArgumentsExpression() {
 				return nil
 			}()
 
-			// `err` is nil means the expression is an invocation
+			// `err` is nil means the expression is an invocation,
```
turbolent, line 604: Typo?
```suggestion
			// `err` is nil means the expression is an invocation
```

runtime/parser/expression.go
```diff
@@ -640,6 +640,12 @@ func defineLessThanOrTypeArgumentsExpression() {
 
 			} else {
 
+				// if the error specifically occurred during parsing restricted types,
+				// this is still an invocation and we should report that error instead
+				if strings.Contains(err.Error(), "restricted type") {
```
turbolent, line 645: Using the error message to detect this case is brittle: if the error message is changed, the code here will still compile, but will fail at run-time.

Instead of returning a generic `SyntaxError` for the restricted type, introduce and report a new error type, and check for it here.

runtime/parser/parser_test.go
```diff
@@ -506,7 +506,7 @@ func TestParseBuffering(t *testing.T) {
 
           fun isneg(x: SignedFixedPoint): Bool {   /* I kinda forget what this is all about */
               return x                             /* but we probably need to figure it out */
-                     <                             /* ************/((TODO?{/*))************ *//
+                     <                             /* ************/((TODO?/*))************ *//
```
turbolent, line 509: Why was this change necessary?

turbolent (commented):
Nice! Thank you for figuring out what's wrong and fixing it

runtime/parser/parser_test.go
dsainati1: It's not strictly necessary, but the presence of the `{` here makes the parser think that the reason this failed is because it was trying to parse a restricted type. The test immediately after this keeps the `{` and changes its error message to be the restricted type parsing failure message, so here I chose to remove the `{` to test both behaviors.

runtime/parser/errors.go
```diff
@@ -318,3 +318,28 @@ func (e *CustomDestructorError) Error() string {
 func (e *CustomDestructorError) SecondaryError() string {
 	return "remove the destructor definition"
 }
+
+// RestrictedTypeError
+
+type RestrictedTypeError struct {
+	Pos ast.Position
```
turbolent, line 325: This works, but providing a range would be nicer

turbolent (approved):
Looks good! Maybe consider replacing the start position with a whole range, given it is available

PR #2986, opened 2023-12-20
Implement modulo argument of `revertibleRandom`

tarakby:
Closes https://github.com/onflow/cadence/issues/2712

## Description

Specifically the item:
- [x] modulo is an optional parameter, if provided the returned value r satisfies 0<= r < modulo. The function panics if modulo == 0 - algorithm implemented should be sample rejection.

Docs is updated in a separate PR against the doc repo: https://github.com/onflow/cadence-lang.org/pull/37

______


runtime/runtime_test.go
```diff
@@ -4460,139 +4462,251 @@ func TestRuntimeBlock(t *testing.T) {
...
+
+	// sanity statistical test to make sure the random numbers less than modulo
+	// are uniform in [0,modulo-1].
+	// The test requires the original random source (here `crypto/rand`)
+	// to be uniform.
+	// The test uses the same small values for all types: one is a power of 2
+	// and the other is not.
+	t.Run("basic uniformity with modulo", func(t *testing.T) {
+		// skipped in CI because the test is slow
+		t.Skip()
```
tarakby, line 4682: Not sure about this `Skip`. This is a statistical test that runs the script about 80k times to test uniformity and validate that the modulo computation isn't biased. It takes about 20s when run locally, which is significantly slower than the rest of the tests.

It should be still run when the code changes, at least locally.

runtime/stdlib/random.go
```diff
@@ -91,54 +89,69 @@ func NewRevertibleRandomFunction(generator RandomGenerator) StandardLibraryValue
...
+				returnIntegerType == sema.UInt32Type || returnIntegerType == sema.UInt64Type ||
+				returnIntegerType == sema.Word8Type || returnIntegerType == sema.Word16Type ||
+				returnIntegerType == sema.Word32Type || returnIntegerType == sema.Word64Type {
+				randomUint64 = getUint64RandomNumber(generator, returnIntegerType, moduloValue)
+			}
+			var randomBig *big.Int
+			if returnIntegerType == sema.UInt128Type || returnIntegerType == sema.UInt256Type ||
+				returnIntegerType == sema.Word128Type || returnIntegerType == sema.Word256Type {
+				randomBig = getBigRandomNumber(generator, returnIntegerType, moduloValue)
+			}
```
turbolent, line 111: Refactor to a `switch` and add a default case which panics with an unreachable error

runtime/stdlib/random.go
```diff
@@ -195,3 +206,207 @@ func NewRevertibleRandomFunction(generator RandomGenerator) StandardLibraryValue
...
+func getUint64RandomNumber(
+	generator RandomGenerator,
+	ty sema.Type,
+	moduloArg interpreter.Value,
+) uint64 {
+	// map for a quick type to byte-size lookup
+	typeToBytes := map[sema.Type]int{
+		sema.UInt8Type: 1, sema.UInt16Type: 2, sema.UInt32Type: 4, sema.UInt64Type: 8,
+		sema.Word8Type: 1, sema.Word16Type: 2, sema.Word32Type: 4, sema.Word64Type: 8,
+	}
```
turbolent, line 224: refactor this out of the function, so the map is not re-allocated each time the function is called.
Same below in `getBigRandomNumber`

turbolent (commented):
Great work!

runtime/stdlib/random.go
```diff
@@ -195,3 +205,204 @@ func NewRevertibleRandomFunction(generator RandomGenerator) StandardLibraryValue
 		},
 	)
 }
+
+// map for a quick type to byte-size lookup
+var typeToBytes = map[sema.Type]int{
```
turbolent, line 210: In the future it would be nice to replace this with a "bit-width" field on `sema.NumericType`

runtime/runtime_test.go
turbolent, line 4682:
```suggestion
		if testing.Short() {
			t.Skip()
		}
```

turbolent (approved):
Nice!

turbolent:
@AlexHentschel Could you please review the implementation?

runtime/runtime_test.go
tarakby: nice!

runtime/stdlib/random.go
tarakby: https://github.com/onflow/cadence/pull/2986/commits/e18191c347180fce435f8bfa70604c4c429df477

runtime/stdlib/random.go
```diff
@@ -195,3 +205,207 @@ func NewRevertibleRandomFunction(generator RandomGenerator) StandardLibraryValue
...
+		mask = (mask << 1) | 1
+	}
+	byteSize := (bitSize + 7) >> 3
+
+	// Generate a number less or equal than `max`.
+	// use the reject-sample method to avoid the modulo bias.
+	// the function isn't constant-time in this case and may take longer than computing
+	// a modular reduction.
+	// However, sampling exactly the size of `max` in bits makes the loop return fast:
+	// loop returns after (k) iterations with a probability of at most 1-(1/2)^k.
```
AlexHentschel, line 305: I have problems following your argument of 
```
loop returns after (k) iterations with a probability of at most 1-(1/2)^k.
```
There is probably some detail that I am overlooking. So I'll try to explain my thoughts and please correct me, where I am mistaken:
* To the best of my understanding, we are interested in the probability bound of the _worst case_. Specifically, the _lower bound_. Hence, I would expect a statement like
   ``` 
   The probability of the loop running for k iterations is _at most_ (1/2)^k
   ``` 
   or
   ``` 
   The probability of the loop taking k iterations or less is _at least_ 1 - (1/2)^k
   ```

runtime/stdlib/random.go
```diff
@@ -195,3 +205,207 @@ func NewRevertibleRandomFunction(generator RandomGenerator) StandardLibraryValue
...
+	//
+	// (a different approach would be to pull 128 bits more bits than the size of `max`
+	// from the random generator and use big number reduction by `modulo`)
+	random := modulo
+	for random > max {
+		// only generate `byteSize` random bytes
+		getRandomBytes(buffer[bufferSize-byteSize:], generator)
+		// big endianness must be used in this case
+		random = binary.BigEndian.Uint64(buffer[:])
+		// adjust to the size of max in bits
```
AlexHentschel, line 315:
```suggestion
		// truncate to bitSize number of bits
```

runtime/stdlib/random.go
```diff
@@ -195,3 +205,207 @@ func NewRevertibleRandomFunction(generator RandomGenerator) StandardLibraryValue
...
+func getBigRandomNumber(
+	generator RandomGenerator,
+	ty sema.Type,
+	moduloArg interpreter.Value,
+) *big.Int {
+
+	// buffer to get random bytes from the generator
+	// 32 is the size of the largest type supported
+	const bufferSize = 32
+	var buffer [bufferSize]byte
```
AlexHentschel, line 332: You instantiate a buffer of the max length, then slice it to the desired size and only work with the sliced buffer:
* https://github.com/onflow/cadence/blob/507c3ff22dcfa4e1f8350f01c1b2c981ca22de1b/runtime/stdlib/random.go#L341-L343
* https://github.com/onflow/cadence/blob/507c3ff22dcfa4e1f8350f01c1b2c981ca22de1b/runtime/stdlib/random.go#L404-L406

I would suggest to initialize a buffer with the correct size to avoid unnecessary large memory allocations and the additional slicing operations.

runtime/stdlib/random.go
```diff
@@ -195,3 +205,207 @@ func NewRevertibleRandomFunction(generator RandomGenerator) StandardLibraryValue
...
+	if moduloArg == nil {
+		numericType, ok := ty.(*sema.NumericType)
+		if !ok {
+			// checker should prevent this
+			panic(errors.NewUnreachableError())
+		}
+		bytes := numericType.ByteSize()
+		getRandomBytes(buffer[:bytes], generator)
+		// SetBytes considers big endianness (although little endian could be used too)
+		return new(big.Int).SetBytes(buffer[:bytes])
```
AlexHentschel, line 343: following my comment above:
```suggestion
		buffer := make([]byte, bytes)
		getRandomBytes(buffer, generator)
		// SetBytes considers big endianness (although little endian could be used too)
		return new(big.Int).SetBytes(buffer)
```

runtime/stdlib/random.go
```diff
@@ -195,3 +205,207 @@ func NewRevertibleRandomFunction(generator RandomGenerator) StandardLibraryValue
...
+	// loop returns after (k) iterations with a probability of at most 1-(1/2)^k.
+	//
+	// (a different approach would be to pull 128 bits more bits than the size of `max`
+	// from the random generator and use big number reduction by `modulo`)
+	random := new(big.Int).Set(modulo)
+	for random.Cmp(max) > 0 {
+		// only generate `byteSize` random bytes
+		getRandomBytes(buffer[:byteSize], generator)
+		// big endianness is used for consistency (but little can be used too)
+		random.SetBytes(buffer[:byteSize])
```
AlexHentschel, line 406: same here:
```suggestion
		buffer := make([]byte, byteSize)
		getRandomBytes(buffer, generator)
		// big endianness is used for consistency (but little can be used too)
		random.SetBytes(buffer)
```

runtime/stdlib/random.go
```diff
@@ -195,3 +205,207 @@ func NewRevertibleRandomFunction(generator RandomGenerator) StandardLibraryValue
...
+	// from the random generator and use big number reduction by `modulo`)
+	random := modulo
+	for random > max {
+		// only generate `byteSize` random bytes
+		getRandomBytes(buffer[bufferSize-byteSize:], generator)
+		// big endianness must be used in this case
+		random = binary.BigEndian.Uint64(buffer[:])
+		// adjust to the size of max in bits
+		random &= mask
+	}
```
AlexHentschel, line 317: for code readability, I would move the `random > max` check into the loop. When reading the code, I was confused why you initially assign https://github.com/onflow/cadence/blob/507c3ff22dcfa4e1f8350f01c1b2c981ca22de1b/runtime/stdlib/random.go#L309 and then I realized it is a round-about way of always going through the loop once. I think it would be better to make this explicit:

```suggestion
for {
  getRandomBytes(buffer[bufferSize-byteSize:], generator) // only generate `byteSize` random bytes
  random = binary.BigEndian.Uint64(buffer[:]) // big endianness must be used in this case
  random &= mask // truncate to bitSize number of bits
  if random <= max {
    break
  }
}
```

turbolent (approved):
Nice!

runtime/stdlib/random.go
```diff
@@ -195,3 +205,207 @@ func NewRevertibleRandomFunction(generator RandomGenerator) StandardLibraryValue
...
+	// Generate a number less or equal than `max`
+	// use the reject-sample method to avoid the modulo bias.
+	// the function isn't constant-time in this case and may take longer than computing
+	// a modular reduction.
+	// However, sampling exactly the size of `max` in bits makes the loop return fast:
+	// loop returns after (k) iterations with a probability of at most 1-(1/2)^k.
+	//
+	// (a different approach would be to pull 128 bits more bits than the size of `max`
+	// from the random generator and use big number reduction by `modulo`)
+	random := new(big.Int).Set(modulo)
```
AlexHentschel, line 401: This `Set` call likely instantiates a new slice (I think every set call for `big.Int` does) and to me it is confusing why we do it. 

So same suggestion as my previous comment to change the for loop to a posterior break. 

```golang
for {
   ... 
   if random.Cmp(max) <= 0 {
      break
   }
}
```

AlexHentschel (approved):
nice work. Your algorithm looks solid to me, but I struggled a bit reviewing it. Tried to add suggestions for improving readability; should all be very quick

runtime/stdlib/random.go
tarakby: - for the small integer types, `8` bytes are needed for the call to `binary.BigEndian.Uint64`. The function only accepts slices of 8 bytes. I also believe allocating a slice of less than 8 bytes on a 64-bits machine would result in allocating at least 8 bytes anyway in memory. 
- I agree about allocating only the needed size for the larger types, updated

runtime/stdlib/random.go
tarakby: I agree my sentence isn't clear, I'll update

turbolent:
One last think I noticed regarding the discussions above: All allocations need memory meterings

runtime/runtime_test.go
tarakby: @turbolent once I implement the suggestion, I noticed that CI hits a 10mn timeout. Can that timeout be increased?

tarakby:
> One last thing I noticed regarding the discussions above: All allocations need memory meterings

Is that something I need to add in the code itself? If yes, can you point me to an example?

runtime/runtime_test.go
tarakby: I'll revert https://github.com/onflow/cadence/pull/2986/commits/e0f07e3901eb8ce8bf227bee414004bccd95216c and https://github.com/onflow/cadence/pull/2986/commits/42271b5b35d71dcfe5bbe080952aa908485c9bdc once we overcome the CI constraint

turbolent: Can the test be reduced to take less than 10 min? e.g. 1-2 min?

turbolent:
I improved the tests in 0f775bc3d0dce849532269b1d52c879bcc35c3f5: Each test case (which potentially might run in concurrently) should use its own `Runtime`, and all sub-tests are now parallelized

runtime/runtime_test.go
tarakby: It actually takes less than a minute on my laptop, I'm surprised CI is more than 10x slower. I'll see what I can do, otherwise we may need to disable the test on CI.

tarakby: Btw I saw you added extra parallelizations in https://github.com/onflow/cadence/pull/2986/commits/2aaa16e5fd3ce8546100c22d99a3573c60083003. To clarify:
- I think sub tests are already parallelized per type here https://github.com/onflow/cadence/blob/2aaa16e5fd3ce8546100c22d99a3573c60083003/runtime/runtime_test.go#L4539
- only [one test](https://github.com/onflow/cadence/blob/2aaa16e5fd3ce8546100c22d99a3573c60083003/runtime/runtime_test.go#L4708) is causing slowness, the others are quick. This is why I disabled that test in CI in my initial PR submission, before you suggested to check `testing.Short()`.

turbolent: I'm also surprised each sub-test for each type takes much longer on CI than on my machine (seconds vs minutes).
If tests are disabled on CI, they will unlikely be run on developer machines, and when code that is tested by it is changed, CI will not run it.

Instead it is better to run as many tests on CI as possible, even if they take a a bit longer. If a developer wants to skip those long tests on their machine, they can do so using `go test -short`.

We can up the default timeout of 10min if needed, but it seems odd why the tests take so long in the first place.

Regarding the parallelization: I also assumed that it was "sufficient" to just parallelize the innermost tests (the one for each type, which, as you pointed out above, was already parallelized).
However, when the outer test scheduling the tests for each type was not itself parallelized, the test framework waited until the first `testTypes` was complete, before scheduling the second `testTypes`.
We want as much parallelism as possible, so also need to indicate that the two `testTypes` can be run at the same time.

It makes a difference on machines with many cores, but not on CI. We might want to also consider upping the CI runners like we do on flow-go

turbolent (approved):
Looks good now, great work!

Should be ready to get merged once the memory metering has been added.

turbolent:
@dsainati1 @SupunS Could you also please have a look?

tarakby:
@Thanks @turbolent for the comments and commits. 
I am still making changes to optimize the long test and enable more type testing (other than the metering). Please do not review till I see how all these changes work on CI. I will ping you once the PR is ready to be merged.

dsainati1 (approved):

runtime/runtime_test.go
tarakby: I agree with running the heavy checks in CI, and keeping the light version for `go test -short`. However there is a limitation for "heavy" if all CI tests have to wrap up under the timeout. Btw I noticed that succeeding CI tests in other PRs currently run in 9-10mn, which is already very close to the timeout.

I've refactored the statistical test in  https://github.com/onflow/cadence/pull/2986/commits/395fb0a22d9b1c0694aa09a94ac6fe7554cdcfc9 so that the 85000 randoms needed from scripts are returned in batches of 5000, that way we minimize the number of times a script is run, unfortunately we still hit the timeout (and I'm not even enabling all types). 
I see two options to move forward: 
- increase the 10mn timeout
- Or keep the statistical test in the code base but disable it in CI till we up the CI runners.

turbolent:
@tarakby 1f0042f looks good 👌

runtime/runtime_test.go
turbolent: I'll try to improve the test further or disable it on CI

turbolent:
@tarakby can you please have a look at the new commits?
- [325e3d4](https://github.com/onflow/cadence/pull/2986/commits/325e3d450451543f733facb48c0860436c23b63b) reverts https://github.com/onflow/cadence/commit/395fb0a22d9b1c0694aa09a94ac6fe7554cdcfc9, the batching isn't necessary in the optimized test
- [83ff4f9](https://github.com/onflow/cadence/pull/2986/commits/83ff4f9f79e56bbc86a3e69ff403bba76380e503) improves the tests a bit, and also removes the statistical tests from the runtime tests (they are brought back at the end)
- [8e0bfad](https://github.com/onflow/cadence/pull/2986/commits/8e0bfadfca0694461868f025906f646f0c351e3d) just refactors out the revertible random function out of the host function wrapper, so it can be called with "normal" arguments instead of having to construct an interpreter invocation
- [8131b27](https://github.com/onflow/cadence/pull/2986/commits/8131b27795b8f0887f710898e518e38682ad1705) re-adds the statistical tests. Instead of running a script, it directly calls the revertible random function, which is much more efficient

tarakby:
Nice, thanks @turbolent 👌🏼 
I am fine with https://github.com/onflow/cadence/pull/2986/commits/8131b27795b8f0887f710898e518e38682ad1705. Although nothing really enforces that `revertibleRandom` calls the tested stdlib function under the hood, It's a good compromise to keep the statistical test in CI. 

I added a minor improvement in https://github.com/onflow/cadence/pull/2986/commits/41117f81fc453496c5a52fb2b12a786b2c40c56c. The PR is ready for merging.

turbolent:
@tarakby Sounds good! 41117f81fc453496c5a52fb2b12a786b2c40c56c looks good 👌 

Agreed, the PR should be ready for review now. @dsainati1 @SupunS could you please have a final look?

dsainati1 (approved):

SupunS (approved):

PR #2987, opened 2023-12-21
Use old hash/ID generation to remove keys

SupunS:
Depends on https://github.com/onflow/cadence/pull/2981

## Description

Solves https://github.com/onflow/cadence/pull/2981#issuecomment-1865328400, (i.e: uses the 'old' hashing mechanism) by overriding the `ID` method (using a wrapper) to use the old hashing mechanism, just before the removal. So the removal would use the old hash, and hence the removal would succeed.

______

### DIFF
--- a/migrations/legacy_intersection_type.go
+++ b/migrations/legacy_intersection_type.go
@@ -1,0 +1,56 @@
+/*
+ * Cadence - The resource-oriented smart contract programming language
+ *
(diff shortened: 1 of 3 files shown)


turbolent (approved):
Great idea to simply reuse what was in the test to make this work! :clap: :ok_hand:

PR #2996, opened 2024-01-08
Statically prevent referenced-typed subexpressions in default arguments to destroy events

dsainati1:
It should not be possible to have a reference-typed sub expression (with the exception of `self` and `base`) appear in a destroy event default argument, as these references may be invalidated. If they are invalidated, accessing them will result in a runtime error, thus potentially creating an indestructible resource. 

This PR adds a check enforcing that the (optional-unwrapped) `ResultingType` of any `MemberExpression` or `IndexExpression` is not a `ReferenceType`. 

Thanks to @oebeling for the report. 

______

### DIFF
--- a/runtime/sema/check_composite_declaration.go
+++ b/runtime/sema/check_composite_declaration.go
@@ -1983,9 +1983,22 @@ func (checker *Checker) checkDefaultDestroyParamExpressionKind(
 		}
 		checker.report(&DefaultDestroyInvalidArgumentError{
 			Range: ast.NewRangeFromPositioned(checker.memoryGauge, arg),
+			Kind:  InvalidIdentifier,
 		})
 
 	case *ast.MemberExpression:
+
+		memberExpressionInfo, ok := checker.Elaboration.MemberExpressionMemberAccessInfo(arg)
+		if !ok {
+			panic(errors.NewUnreachableError())
+		}
+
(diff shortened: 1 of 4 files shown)


turbolent (approved):

runtime/sema/check_composite_declaration.go
```diff
@@ -1983,25 +1993,37 @@ func (checker *Checker) checkDefaultDestroyParamExpressionKind(
...
 		})
 
 	case *ast.MemberExpression:
 
+		memberExpressionInfo, ok := checker.Elaboration.MemberExpressionMemberAccessInfo(arg)
+		if !ok {
+			panic(errors.NewUnreachableError())
+		}
+
+		rejectReferenceTypedSubExpressions(memberExpressionInfo.ResultingType)
```
SupunS, line 2006: In an expression like `self.ref.i` where `i` is of `Int` type, isn't the `ResultingType` also `Int`?

dsainati1: Yes, the `ResultingType` of `self.ref.i` is `Int`. However, we will recursively check the `self.ref` subexpression, and that has a `ResultingType` of `&InnerResource`, and will trigger the error.

SupunS: oh nice! Didn't realize it checks the expression recursively 👍

SupunS (approved):

PR #3000, opened 2024-01-09
Prevent use of non-attachment type names in default arguments

dsainati1:
We previously (mistakenly) allowed non-attachment type names to be used in default destroy arguments, which could cause issues with Contract-kinded composites being used as variables, since these declarations create both a type and a value at the same time.  

______


turbolent (approved):
Nice!

SupunS (approved):

PR #3008, opened 2024-01-10
Add support for dereferencing optional references

### LINKED ISSUE #3005: Lift `*` to optional references
### Issue to be solved

As mentioned in https://github.com/onflow/cadence/pull/2982, it would be nice for the newly introduced `*` dereference operator to implicitly lift to optional references. I.e., assuming the `*` operator has type `fun<T: Primitive>(_ ref: &T): T`, we should implicitly overload the operator to also have type `fun<T: Primitive>(_ ref: &T?): T?`.

### Suggested Solution

_No response_

turbolent:
Closes #3005 

______


dsainati1:
Does this only unwrap optionals one layer? I.e. will `*` fail on a `&Int??`

turbolent:
@dsainati1 
> Does this only unwrap optionals one layer? I.e. will `*` fail on a `&Int??`

Yes, only one layer of optionality is unwrapped, just like the optional chaining version did. This is also similar to how e.g. `!` behaves. `*` on `&Int??` won't fail, but result in `Int?`.

dsainati1:
Yeah, that makes sense. 

> To be honest, this is already quite some syntactic sugar, maybe we shouldn't even have this?

IMO this is useful for the same reason that optional chaining is; it makes a common case more convenient to use. You're right that this is doable another way, but it would probably be more frustrating to do that way. The same is true for optional chaining!

I'm happy with this only unwrapping one layer of the optional.

dsainati1 (approved):

turbolent:
> IMO this is useful for the same reason that optional chaining is; it makes a common case more convenient to use.

Good point, agreed!

Issue #3006, opened 2024-01-10
Investigate key-hashing for string/entitlement migrations


SupunS:
### Issue to be solved

Type value migration needed the intersection-types to have the old/legacy hashing function, when 'removing' the dictionary key during the migration.

Similar to type-values, string and reference values also now have an updated hashing function, however, the key-removal seems to work seamlessly, without the need for the old/legacy hashing. Investigate this further to see why, and to see if there are any potential issues with that.

PR #3013, opened 2024-01-11
Use legacy hashing function for key removal

SupunS:
Closes https://github.com/onflow/cadence/issues/3006

## Description

Use the legacy hashing function for existing key removal during value migrations.

______


migrations/legacy_intersection_type.go
```diff
@@ -43,6 +43,11 @@ func (t *LegacyIntersectionType) ID() common.TypeID {
 	}
 
 	var result strings.Builder
+
+	if t.LegacyType != nil {
+		result.WriteString(string(t.LegacyType.ID()))
+	}
```
SupunS, line 49: This was a bug in intersection-type hashing. The old restricted-type also includes the 'restricting' type in hashing.

runtime/interpreter/value.go
```diff
@@ -1017,14 +1017,16 @@ type StringValue struct {
 	// that are based on grapheme clusters
 	graphemes *uniseg.Graphemes
 	Str       string
+	RawStr    string
```
turbolent, line 1020: Small nit: Maybe call this "non-normalized"

migrations/migration.go
```diff
@@ -327,19 +327,22 @@ func (m *StorageMigration) migrate(
...
+			return interpreter.NewUnmeteredTypeValue(legacyType)
+		}
 
-	legacyType := legacyType(typeValue.Type)
-	if legacyType == nil {
-		return key
+	case *interpreter.StringValue:
+		return &LegacyStringValue{
+			key,
+		}
```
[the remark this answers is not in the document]
turbolent: Is this still a problem? The linked commit doesn't exist anymore

turbolent (commented):
Thank you for investigating and fixing this!

migrations/migration.go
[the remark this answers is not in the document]
SupunS: Ah sorry, I may have forced push the changes so the commit id is now different. Updated the broken link.

Migration works correctly now, with this change. However, still couldn't reproduce it here.

turbolent:
Could we maybe have a code review session for this tomorrow and go over this together?

dsainati1 (approved):

turbolent:
We'll also need a legacy value for `Character` (and associated `TestRehash` test)

migrations/entitlements/migration_test.go
```diff
@@ -1867,244 +1867,3 @@ func TestMigrateDictOfValues(t *testing.T) {
 		ref.Authorization,
 	)
 }
-
-func TestMigrateDictOfWithTypeValueKey(t *testing.T) {
```
SupunS, line 1871: This test is no longer valid/working, since the migration uses the "legacy" hashing functions to remove the reference-typed key, whereas this test stores the "new format" of the reference type; thus the removal fails.
I've added a proper test at the FVM, which stores a reference type in the "old format", and then uses the legacy-hashing function for key-removal.

migrations/entitlements/migration.go
```diff
@@ -254,10 +261,14 @@ func ConvertValueToEntitlements(
...
 		}
 
-		convertedType, _ := ConvertToEntitledType(
+		convertedType, converted := ConvertToEntitledType(
 			inter.MustConvertStaticToSemaType(v.Type),
 		)
 
+		if !converted {
+			return nil
+		}
```
SupunS, line 270: If the type is not converted, it returns a `nil`. So exit early here, and avoid wrapping `nil` with a type value.

migrations/entitlements/migration_test.go
turbolent: Is there any way to also keep a variant of this test here on the Cadence side?

migrations/entitlements/migration.go
```diff
@@ -131,6 +131,13 @@ func ConvertValueToEntitlements(
 			referenceValue.Authorization,
 			interpreter.ConvertSemaToStaticType(inter, referenceValue.BorrowedType),
 		)
+	case interpreter.LinkValue: //nolint:staticcheck
+		// Link values are not supposed to reach here.
+		// But it could, if the type used in the link is not migrated,
+		// then the link values would be left un-migrated.
```
turbolent, line 160: Is this a possible result of the migration?

migrations/legacy_reference_type.go
```diff
@@ -33,9 +33,13 @@ type LegacyReferenceType struct {
 var _ interpreter.StaticType = &LegacyReferenceType{}
 
 func (t *LegacyReferenceType) ID() common.TypeID {
-	isAuthorized := t.Authorization == interpreter.UnauthorizedAccess
```
turbolent: Good catch!

turbolent (approved):

migrations/entitlements/migration.go
SupunS: yeah, for e.g: if type-loading fails during link-migration, then the link values currently remains in the storage as-is.

migrations/entitlements/migration_test.go
turbolent: Obviously I don't want to cause extra work here,  but could we maybe just add a legacy value definition in addition to the existing one, and stub everything out, but only leave the encoding part, similar to how we do this for link values in the cap cons migration?

SupunS: actually, turned out to be easier: only needed to 'override' the `Encode`method of the existing `LegacyReferenceType`: https://github.com/onflow/cadence/pull/3013/commits/efd96205743b1df5bf4ed6cf6bff56162f03d736

turbolent: Awesome!

turbolent (approved):
Great work!

PR #3014, opened 2024-01-11
Add type assertion on values passed to `Test.assertEqual`

m-Peter:
## Description

The `Test.assertEqual` method did not handle the case where the given inputs were equal in value, but had different types, e.g.:
```cadence
import Test

access(all)
fun test() {
    let expected: Int = 15
    let actual: Int? = 15
    Test.assertEqual(expected, actual)
}
```
The assertion message from above would be:

```bash
assertion failed: not equal: expected: 15, actual: 15
```
Which is quite confusing to developers.

Before performing value equality, we perform a type equality, so the assertion message now becomes:

```bash
assertion failed: not equal types: expected: Int, actual: Int?
```
______

### DIFF
--- a/runtime/stdlib/test_test.go
+++ b/runtime/stdlib/test_test.go
@@ -726,7 +726,7 @@ func TestAssertEqual(t *testing.T) {
             fun test() {
                 Test.assertEqual("this string", "this string")
             }
-        `
+		`
 
 		inter, err := newTestContractInterpreter(t, script)
 		require.NoError(t, err)
@@ -745,7 +745,7 @@ func TestAssertEqual(t *testing.T) {
             fun test() {
                 Test.assertEqual(15, 21)
             }
-        `
+		`
 
 		inter, err := newTestContractInterpreter(t, script)
 		require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


runtime/stdlib/test_test.go
```diff
@@ -726,7 +726,7 @@ func TestAssertEqual(t *testing.T) {
             fun test() {
                 Test.assertEqual("this string", "this string")
             }
-        `
+		`
```
SupunS, line 729: Could you maybe undo this [accidental] change? Having whitespaces is preferred over tabs, given rest of the code snippet also uses whitespaces for indentation (in code blocks)

SupunS (commented):
Nice! 👌

runtime/stdlib/test_test.go
m-Peter: Ah yeah, fixed in https://github.com/onflow/cadence/pull/3014/commits/43b146a6435cf1a78e67e1d001bcf2d96a43c4bf

SupunS (approved):

turbolent (commented):
Good idea, nice!

PR #3015, opened 2024-01-11
Support milliseconds for time delta in `Test.moveTime`

m-Peter:
## Description

Although the `timeDelta` in `Test.moveTime` accepts a `Fix64` value, under the hood it would truncate the value to `int64`.
We change this to deal with `float64`, so that we can support millisecond granularity.
Relevant issue: https://github.com/onflow/cadence-tools/issues/264#issuecomment-1881355919

______

### DIFF
--- a/runtime/stdlib/test_emulatorbackend.go
+++ b/runtime/stdlib/test_emulatorbackend.go
@@ -726,7 +726,7 @@ func (t *testEmulatorBackendType) newMoveTimeFunction(
 			if !ok {
 				panic(errors.NewUnreachableError())
 			}
-			blockchain.MoveTime(int64(timeDelta.ToInt(invocation.LocationRange)))
+			blockchain.MoveTime(float64(timeDelta) / sema.Fix64Factor)
 			return interpreter.Void
 		},
 	)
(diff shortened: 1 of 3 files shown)


runtime/stdlib/test_emulatorbackend.go
```diff
@@ -726,7 +726,7 @@ func (t *testEmulatorBackendType) newMoveTimeFunction(
 			if !ok {
 				panic(errors.NewUnreachableError())
 			}
-			blockchain.MoveTime(int64(timeDelta.ToInt(invocation.LocationRange)))
+			blockchain.MoveTime(float64(timeDelta) / sema.Fix64Factor)
```
SupunS, line 729: Casting an `int64` to `float64` could lose precision in some cases. We could instead pass `int64` (`Fix64Value` as is) to `MoveTime`, and change the parameter's metric to accept nano-seconds (it currently accepts seconds). But might need to pad-right if we do so, since `Fix64Factor` is 10^8 and nano-seconds factor 10^9. But then there's a risk of padding-right could overflow/underflow. So IDK which one is better.
In reality, these extreme cases would not be even used, so maybe we don't need to worry much, and just documenting might be enough?
@turbolent wdyt?

turbolent: Agreed, we shouldn't use floating point here

Issue #2530, opened 2023-06-05
Allow conversion between constant-size and variable-size arrays


bluesign:
### Issue to be solved

Inspired by recent issue on codeHash in events, I think we should allow some conversion with arrays of different static type. 

```cadence 
// https://run.dnz.dev/snippet/1f8bfbdc77297319

pub fun main(){
    var a : [UInt8; 8] = [1,2,3,4,5,6,7,8]
    var b : [UInt8] = [10,11,12,13,14,15,16,17,18]

    var c : [UInt8] = a as! [UInt8]
    var d : [UInt8; 8] = b as! [UInt8; 8]
}
```

if we don't allow this, only solution is ugly loop. 

### Suggested Solution

Conversion can allow ( with force cast for example ) conversion between 2 types. As force cast has T as right side, I don't think it will create any problems.

turbolent:
Good idea, this could be easily added, PRs are very welcome. 

`as/as?/!` are down-casting operators and do not perform conversion. Instead, the standard library offers conversion functions. The same should be done here, though naming it might be a bit tough.

PratikDhanave:
@j1010001 @bluesign @turbolent   is this issue is resolved ? if yes humble request please can we mark it as closed ?

turbolent:
No, this feature has not been added yet

darkdrag00nv2:
I believe we have two options:

####  Option 1
We could introduce names for the variable and constant sized array types. Then we could declare conversion functions

```cadence
VariableSizedArray.fromConstantSizedArray<C: [T; _]>(param: C) : [T]
ConstantSizedArray.fromVariableSizedArray<C: [T; _]>(param: [T]) : C
```

#### Option 2

Declare two standard library functions:

```cadence
VariableSizedArrayToConstantSizedArray<C: [T; _]>(param: [T]) : C // or ConstantSizedArrayFromVariableSizedArray
ConstantSizedArrayToVariableSizedArray<C: [T; _]>(param: C) : [T] // or VariableSizedArrayFromConstantSizedArray
```
---

In either of these two options, I am not sure if it would be feasible to add the type bound `C : [T; _]`.

bluesign:
Another option:
 .asFixedSizeArray(): T to variable sized arrays. 

And .asVariableSizeArray(): T to fixed sized arrays

Ofc naming can be improved.

turbolent:
Agreed, it might be better to not "pollute" the global namespace and just have functions on the types.

[...]

PR #3016, opened 2024-01-11
Introduce `toVariableSized` to ConstantSizedArray type

darkdrag00nv2:
Work towards #2530

## Description
Introduce `toVariableSized` function to `ConstantSizedArray` type. This function will return a variable sized array whose contents are copies of the original constant sized array.

Since copy of values are involved, this is not available if the inner type of the constant sized array is a resource.
______

### DIFF
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -2571,7 +2571,7 @@ func ArraySliceFunctionType(elementType Type) *FunctionType {
 
 func ArrayToVariableSizedFunctionType(elementType Type) *FunctionType {
 	return NewSimpleFunctionType(
-		FunctionPurityImpure,
+		FunctionPurityView,
 		[]Parameter{},
 		NewTypeAnnotation(&VariableSizedType{
 			Type: elementType,
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -2025,6 +2025,13 @@ Returns a new array with contents in the reversed order.
 Available if the array element type is not resource-kinded.
 `
 
+const ArrayTypeToVariableSizedFunctionName = "toVariableSized"
+
+const arrayTypeToVariableSizedFunctionDocString = `
+Returns a new variable sized array with the same content.
+Available if the array is constant sized and the element type is not resource-kinded.
+`
+
 var insertableEntitledAccess = NewEntitlementSetAccess(
 	[]*EntitlementType{
 		InsertType,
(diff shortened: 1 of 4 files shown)


runtime/sema/type.go
```diff
@@ -2025,6 +2025,13 @@ Returns a new array with contents in the reversed order.
 Available if the array element type is not resource-kinded.
 `
 
+const ArrayTypeToVariableSizedFunctionName = "toVariableSized"
+
+const arrayTypeToVariableSizedFunctionDocString = `
+Returns a new variable sized array with the same content.
```
turbolent, line 2031:
```suggestion
Returns a new variable-sized array with the copy of the contents of the given array.
```

runtime/sema/type.go
```diff
@@ -2533,6 +2569,16 @@ func ArraySliceFunctionType(elementType Type) *FunctionType {
 	)
 }
 
+func ArrayToVariableSizedFunctionType(elementType Type) *FunctionType {
+	return NewSimpleFunctionType(
+		FunctionPurityImpure,
```
turbolent, line 2574: I think it should be possible to make this `view`?

runtime/sema/type.go
```diff
@@ -2533,6 +2569,16 @@ func ArraySliceFunctionType(elementType Type) *FunctionType {
...
+		[]Parameter{},
```
turbolent, line 2575: Can this be `nil`?

runtime/tests/checker/arrays_dictionaries_test.go
```diff
@@ -2478,3 +2478,78 @@ func TestCheckDictionaryFunctionEntitlements(t *testing.T) {
...
+	assert.IsType(t, &sema.ExcessiveArgumentsError{}, errs[0])
+}
+
+func TestCheckVariableSizedArrayToVariableSizedInvalid(t *testing.T) {
+
+	t.Parallel()
+
+	_, err := ParseAndCheck(t, `
+		resource X {}
+
```
turbolent, line 2523: Not needed here

```suggestion
```

turbolent (commented):
Nice!

runtime/sema/type.go
turbolent: cc @dsainati1

dsainati1: Yeah, it should be

darkdrag00nv2: Done, updated.

turbolent (approved):
Nice!

runtime/sema/type.go
SupunS: I think all those usages are nil-safe. For example, `t.Parameters` is always used in for-loops or with `len(t.Parameters)`, both of which handle `nil` gracefully.
There are already function types with `nil` parameters. e.g:
https://github.com/onflow/cadence/blob/08be35302d32825a01001d8ba217fd32d15e8f51/runtime/sema/runtime_type_constructors.go#L27-L31

```suggestion
		nil,
```

SupunS (approved):
Look good!

PR #3009, opened 2024-01-10
Merge range type feature

turbolent:
Closes #2482 
Closes #2886

## Description

Let's merge the great work by @darkdrag00nv2 👏 🎉 

______

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #3017, opened 2024-01-11
Improve code related to new InclusiveRange type

turbolent:
## Description

When reviewing #3009 I noticed a few things that could be improved.

______

### DIFF
--- a/runtime/stdlib/range.go
+++ b/runtime/stdlib/range.go
@@ -80,7 +80,7 @@ var inclusiveRangeConstructorFunctionType = func() *sema.FunctionType {
 			memoryGauge common.MemoryGauge,
 			typeArguments *sema.TypeParameterTypeOrderedMap,
 			astTypeArguments []*ast.TypeAnnotation,
-			astInvocationRange ast.Range,
+			invocationRange ast.HasPosition,
 			report func(error),
 		) {
 			memberType, ok := typeArguments.Get(typeParameter)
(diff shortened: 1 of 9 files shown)


dsainati1 (approved):

runtime/stdlib/range.go
```diff
@@ -89,21 +89,27 @@ var inclusiveRangeConstructorFunctionType = func() *sema.FunctionType {
...
+				if memberType != ty {
+					continue
+				}
+
+				var errorRange ast.HasPosition
+				// If type argument was provided, use its range otherwise fallback to invocation range.
+				if len(astTypeArguments) > 0 {
+					errorRange = astTypeArguments[0]
+				} else {
+					errorRange = invocationRange
```
SupunS, line 103: Could move out of the loop

SupunS (approved):

runtime/stdlib/range.go
```diff
@@ -89,21 +89,27 @@ var inclusiveRangeConstructorFunctionType = func() *sema.FunctionType {
...
 				}
```
turbolent, line 104:
```suggestion
				// If type argument was provided, use its range otherwise fallback to invocation range.
				errorRange := invocationRange
				if len(astTypeArguments) > 0 {
					errorRange = astTypeArguments[0]
				}
```

runtime/stdlib/range.go
turbolent: Done in [c3f04ec](https://github.com/onflow/cadence/pull/3017/commits/c3f04ec8d1bf099d14429ce28c5f3dc1c371dcd3)

PR #3019, opened 2024-01-12
Reserve more keywords for future proofing

turbolent:
## Description

Code written in Cadence 1.0 must stay backward-compatible. At the same time we also want to keep evolving the language.

Conservatively reserve some more keywords we might use in the future. Adding these keywords does not imply we will use them, it just ensures we can, because already code written for 1.0 didn't use them.

______


runtime/parser/keyword.go
```diff
@@ -75,6 +75,22 @@ const (
...
+	KeywordGoto        = "goto"
+	KeywordConst       = "const"
+	KeywordExport      = "export"
+	KeywordThrow       = "throw"
+	KeywordThrows      = "throws"
+	KeywordRequires    = "requires"
+	KeywordWhere       = "where"
+	KeywordFinal       = "final"
+	KeywordInternal    = "internal"
+	KeywordTypealias   = "typealias"
```
dsainati1, line 90: This one seems a bit odd to me, why reserve `typealias` but not `type`?

turbolent: We can also reserve `type`. I just tried to come up with a list, looking at reserved keywords of Swift, JS, Java, etc. and hadn't thought of `type`. It's not intentionally omitted

dsainati1: I think `type` is commonly used to create a type alias (e.g. Go, TypeScript), so we should reserve it just to be safe

SupunS (approved):

runtime/parser/keyword.go
turbolent: I tried adding `type`, but it is used in the standard library, e.g. in `forEachStored` as an argument label. I've added it to the soft keywords set.

dsainati1 (approved):

PR #3023, opened 2024-01-13
Require explicit type annotation for arguments passed to authorized reference parameters

turbolent:
## Description

[FLIP 941: Reference Creation Semantics](https://github.com/onflow/flips/blob/main/cadence/20220516-reference-creation-semantics.md) changed the semantics for creating references, so that an explicit type is no longer required. This improved the confusion around the previous syntax for creating references, `&v as &T`, which was similar to casting. The change made creating references easier and more ergonomic.

However, this is potentially dangerous if the expected type is not local, like in a variable declaration (as used in the FLIP), but defined elsewhere – which is the case for functions, which might be defined in other locations.

For example, consider:

```cadence
// Defined elsewhere, is going to mutate the passed array
fun foo(ints: auth(Mutate) &[Int]) {}

let ints = [1, 2, 3]
// Would implicitly allow mutation
foo(&ints)
```

Change the type checking for
- Invocation arguments
- Assignments 

so that the target type is only used as the expected type when checking the value, if it is not, or does not contain, a reference type.

______


turbolent:
We'll also want to extend this to nested types, e.g. `[auth(Mutate) &Int]`

SupunS (approved):

dsainati1:
While I understand why this change is being proposed, IMO this particular implementation feels a bit arbitrary and "magical". It's important that type inference be predictable to users, since they have little insight into the process, and it can be frustrating that some code seems to require annotations while others don't. One can easily imagine a situation where a developer adds an entitlement to a reference parameter within their own contract, and gets confused when callsites of that function suddenly break downstream. 

IMO if we want to proceed with this change, we should disable "expected type" inference entirely in function parameters (IMO in contravariant positions in general, but parameters are the only ones we have atm as far as I know). This would mean that unentitled reference arguments would still need annotations, but would be significantly more consistent and predictable than making inference dependent on whether the type is entitled or not.

turbolent:
> While I understand why this change is being proposed, IMO this particular implementation feels a bit arbitrary and "magical". It's important that type inference be predictable to users, since they have little insight into the process, and it can be frustrating that some code seems to require annotations while others don't. One can easily imagine a situation where a developer adds an entitlement to a reference parameter within their own contract, and gets confused when callsites of that function suddenly break downstream.

Definitely agreed with having type inference predictable and reducing frustration/confusion why/when type annotations are necessary. Just like type inference in general makes a "best effort" at inferring the type and requiring explicit annotations if it cannot infer the type, I think it is OK to keep trying to require as few type annotations as possible, and only requiring them when needed, because the type cannot be inferred otherwise, – or when it is needed for security reasons. 

Explaining this in the error message (not jet done in the PR, but a great flag) should help with it, e.g. something like "not inferring authorized reference type here, this could be a security issue". 

I think given that references are about access control, which is a critical part of a contract, handling type inference "specially" in the case where special access is granted seems like a good trade-off.

> IMO if we want to proceed with this change, we should disable "expected type" inference for references entirely in function parameters (IMO in contravariant positions in general, but parameters are the only ones we have atm as far as I know). This would mean that unentitled reference arguments would still need annotations, but would be significantly more consistent and predictable than making inference dependent on whether the type is entitled or not.

Changing the behaviour to requiring type an explicit type annotation for any reference expression, independent of if it is authorized/entitled or not, sounds good to me, but I don't really see why we would want to burden the developer with it, given the result is safe. 

I don't have strong feelings here, I'm mostly just interested in how we handle the potentially dangerous case of accidentally granting access, but more curious as to why we would draw the line for type inference there. 

For example, the argument above (paraphrased):

> "for references, we always require an explicit type annotation, e.g. one has to write `&1 as &Int`, even if the type could be inferred from `&1` and is safe"

seems similar to

> "for arrays, we always require an explicit type annotation, e.g. one has to write `[1] as [Int]`, even if the type could be inferred from `[1]` and is safe"

dsainati1:
The difference as I see it, is that IMO type inference should be completely predictable (or as predictable as possible) from syntax alone. A programmer should be able to look at a piece of code, and reason locally about whether or not a type will be inferred for a value based solely on how that value is used syntactically, not needing to worry about the actual or expected type of the value. In this case, we would say that all references appearing in function arguments *must* be explicitly annotated. That's simple. The alternative would be that references appearing in function arguments must be explicitly annotated *if* the function parameter type is entitled. This requires non-local reasoning (i.e. looking at the function definition), and is thus less ergonomic for the developer.

turbolent:
> A programmer should be able to look at a piece of code, and reason locally about whether or not a type will be inferred for a value based solely on how that value is used syntactically, not needing to worry about the actual or expected type of the value. In this case, we would say that all references appearing in function arguments must be explicitly annotated. That's simple, and the developer can see locally, and purely syntactically, that they are passing a reference to a function and thus must annotate it. The alternative would be that references appearing in function arguments must be explicitly annotated if the function parameter type is entitled. This requires non-local reasoning (i.e. looking at the function definition), and is thus less ergonomic for the developer.

That's a really great argument, I'll add that as a comment to the code 👌

turbolent:
@SupunS @dsainati1  Updated now to check both invocation arguments and assignments, and check if the type is or contains a reference type.

runtime/tests/checker/attachments_test.go
```diff
@@ -4022,7 +4022,7 @@ func TestCheckAttachmentBaseNonMember(t *testing.T) {
 			access(all) attachment A for R {
 				access(all) let bases: [&R]
 				init() {
-					self.bases = []
+					self.bases = [] as [&R]
```
dsainati1, line 4025: How difficult would it be to allow assignments to `self` without an explicit type annotation?

turbolent: Should be possible I think, have not tried yet. I think we'd have to analyze the LHS of the assignment and see if the expression is a member expression accessing `self`

dsainati1 (approved):
Nice!

runtime/tests/checker/attachments_test.go
turbolent: Improved in [53e7cc1](https://github.com/onflow/cadence/pull/3023/commits/53e7cc146de36d6b0c3aa5616a952c8ea7763126)

turbolent:
Updated the behaviour for assignments, so assignments to self members do not need a type annotation, and also updated the tests. 

Should be good to go now

SupunS (approved):

PR #3026, opened 2024-01-17
Fix the entitlements migration

turbolent:
## Description

Fix the entitlements migration by not adding authorizations to reference types which were previously converted from deprecated primitive static types. Otherwise it incorrectly migrates e.g. `PublicAccount` to `auth(Storage, Contracts, Keys, Inbox, Capabilities) &Account` (via `PublicAccount` to `&Account` through the account types migration).

Also:
- Add `StaticType.IsDeprecated` to enable the fix above
- Improve `PrimitiveStaticType.SemaType()` to not just return `nil`, but panic for deprecated types. This prevents e.g. accidentally constructing invalid types like "`&nil`" when converting types boxing deprecated primitive types.
- Improve cap con migration tests, as they rely on the cases in `PrimitiveStaticType.SemaType` to convert the deprecated account types `PublicAccount` and `AuthAccount` to sema types

______

### DIFF
--- a/migrations/entitlements/migration.go
+++ b/migrations/entitlements/migration.go
@@ -47,8 +47,26 @@ func (EntitlementsMigration) Name() string {
 // * `ConvertToEntitledType(T) ---> T`
 // where Entitlements(I) is defined as the result of T.SupportedEntitlements()
 func ConvertToEntitledType(t sema.Type) (sema.Type, bool) {
+
 	switch t := t.(type) {
 	case *sema.ReferenceType:
+
+		// Do NOT add authorization for sema types
+		// that were converted from deprecated primitive static types
+		switch t.Type {
+
+		case sema.AccountType,
+			sema.Account_ContractsType,
+			sema.Account_KeysType,
+			sema.Account_InboxType,
+			sema.Account_StorageCapabilitiesType,
+			sema.Account_AccountCapabilitiesType,
+			sema.Account_CapabilitiesType,
+			sema.AccountKeyType:
+
+			return t, false
+		}
+
 		switch t.Authorization {
 		case sema.UnauthorizedAccess:
 			innerType, convertedInner := ConvertToEntitledType(t.Type)
--- a/migrations/entitlements/migration.go
+++ b/migrations/entitlements/migration.go
@@ -19,6 +19,8 @@
 package entitlements
 
 import (
+	"fmt"
+
 	"github.com/onflow/cadence/migrations"
 	"github.com/onflow/cadence/runtime/interpreter"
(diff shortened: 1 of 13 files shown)


migrations/entitlements/migration.go
```diff
@@ -135,6 +135,10 @@ func ConvertValueToEntitlements(
 		staticType = v.StaticType(inter)
 	}
 
+	if staticType.IsDeprecated() {
+		return nil
+	}
```
turbolent, line 140: The actual fix

runtime/interpreter/primitivestatictype.go
```diff
@@ -664,6 +647,17 @@ func (t PrimitiveStaticType) SemaType() sema.Type {
 		return sema.CapabilitiesMappingType
 	case PrimitiveStaticTypeAccountMapping:
 		return sema.AccountMappingType
+
+	case PrimitiveStaticTypeAuthAccount: //nolint:staticcheck
+		// deprecated, but needed for migration purposes
+		return sema.FullyEntitledAccountReferenceType
+	case PrimitiveStaticTypePublicAccount: //nolint:staticcheck
+		// deprecated, but needed for migration purposes
+		return sema.AccountReferenceType
```
turbolent, line 656: Needed for cap cons migration

migrations/entitlements/migration_test.go
```diff
@@ -2108,3 +2117,85 @@ func TestMigrateDictOfWithTypeValueKey(t *testing.T) {
...
+				},
+				value,
+				inter,
+			)
+			require.NoError(t, err)
+			require.NotNil(t, newValue)
+
+			result, err := ConvertValueToEntitlements(inter, newValue)
+			require.NoError(t, err)
+			require.Nilf(t, result, "expected no migration, but got %s", result)
```
turbolent, line 2190: @dsainati1 Where in the entitlements migration do we need to add the check if the given `sema.Type` to be migrated should be ignored because it got migrated from a deprecated primitive static type (e.g. `&Account` should not be migrated)?

dsainati1: I think in `ConvertToEntitledType` you should be able to just return the input unchanged if it is a deprecated type that we should ignore.

turbolent: I had initially tried that at the top and it didn't work, but realized that it must be specifically added in the reference type case which adds the authorization

dsainati1 (approved):
Nice! Thanks for improving this

SupunS (approved):
Nice! Thank you for adding these

PR #3027, opened 2024-01-17
Upgrade `exp` package

tarakby:
`exp` package's version in this repo is behind the version in other `onflow` repos. The newer version of `exp` has a breaking change. This causes issues when updating dependencies in other repos. 

This PR: 
 - updates `exp` dependency
 - updates `[mpz](https://github.com/SaveTheRbtz/mph)` package after updating its `exp` version in https://github.com/SaveTheRbtz/mph/pull/2 (also having a breaking change).
 - solves some breaking changes


tarakby:
There is a build issue in some workflows: the breaking change in `mph` still causes an issue. My guess is that the other repos involved in the workflows need to be updated too (for instance flow-emulator). 
Surprisingly, my update to `mph` solving the breaking change was branched off `v0.1.1` while the highest `mph` version is `v0.1.2`.

.github/workflows/downstream.yml
```diff
@@ -36,17 +36,14 @@ jobs:
...
-
       - name: Update Cadence
         run: go mod edit -replace github.com/onflow/cadence=github.com/${{ github.event.pull_request.head.repo.full_name || github.repository }}@${{ github.event.pull_request.head.sha || github.sha }}
 
       - name: Tidy up
         run: go mod tidy
 
       - name: Build
-        run: go build -v --tags relic ./...
+        run: go build -v --tags ./...
```
turbolent, line 46:
```suggestion
        run: go build -v ./...
```

turbolent (approved):

turbolent:
Can we maybe fork `mph` at the same time into the `onflow` org, so we can more easily update it in the future?

tarakby:
> Can we maybe fork mph at the same time into the onflow org, so we can more easily update it in the future?

We could, though similar issues can happen with any third-party package and we can't fork all of them. We needed the PR to the third-party repo in this case because:
 1. there was a breaking change in a standard library (`exp`) so that we couldn't build the third-party package.
 2. the third-party package didn't upgrade in a long time. 
 My guess is that those two reasons together don't happen often.

SupunS (approved):

tarakby:
@turbolent I have merged this PR to unblock the stable-cadence work in `flow-go`, but we can still fork the `mph` work later if we want to.

PR #3028, opened 2024-01-17
Introduce `toConstantSized` to `VariableSizedArray` type

### LINKED ISSUE #2530: Allow conversion between constant-size and variable-size arrays

darkdrag00nv2:
Closes #2530

## Description

Introduce `toConstantSized` function to `VariableSizedArray` type. This function will return a constant-sized array whose contents are copies of the original variable-sized array.

Since copy of values are involved, this is not available if the inner type of the variable-sized array is a resource.

______

### DIFF
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -2131,6 +2131,13 @@ Returns a new variable-sized array with the copy of the contents of the given ar
 Available if the array is constant sized and the element type is not resource-kinded.
 `
 
+const ArrayTypeToConstantSizedFunctionName = "toConstantSized"
+
+const arrayTypeToConstantSizedFunctionDocString = `
+Returns a new constant-sized array with the copy of the contents of the given array.
+Available if the array is constant sized and the element type is not resource-kinded.
+`
+
 var insertMutateEntitledAccess = NewEntitlementSetAccess(
 	[]*EntitlementType{
 		InsertType,
@@ -2497,6 +2504,32 @@ func getArrayMembers(arrayType ArrayType) map[string]MemberResolver {
 				)
 			},
 		}
+
+		members[ArrayTypeToConstantSizedFunctionName] = MemberResolver{
+			Kind: common.DeclarationKindFunction,
+			Resolve: func(memoryGauge common.MemoryGauge, identifier string, targetRange ast.Range, report func(error)) *Member {
+				elementType := arrayType.ElementType(false)
+
+				if elementType.IsResourceType() {
+					report(
+						&InvalidResourceArrayMemberError{
+							Name:            identifier,
+							DeclarationKind: common.DeclarationKindFunction,
+							Range:           targetRange,
+						},
+					)
+				}
+
+				return NewFunctionMember(
+					memoryGauge,
+					arrayType,
+					insertableEntitledAccess,
+					identifier,
+					ArrayToConstantSizedFunctionType(elementType),
+					arrayTypeToConstantSizedFunctionDocString,
+				)
+			},
+		}
 	}
 
 	if _, ok := arrayType.(*ConstantSizedType); ok {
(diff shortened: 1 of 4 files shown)


runtime/sema/type.go
```diff
@@ -2678,6 +2711,62 @@ func ArrayToVariableSizedFunctionType(elementType Type) *FunctionType {
...
+				if len(astTypeArguments) > 0 {
+					errorRange = astTypeArguments[0]
+				}
+
+				report(&InvalidTypeArgumentError{
+					TypeArgumentName: typeParameter.Name,
+					Range:            ast.NewRangeFromPositioned(memoryGauge, errorRange),
+					Details: fmt.Sprintf(
+						"Type argument for toConstantSized must be [%s; _]",
+						elementType,
```
turbolent, line 2762:
```suggestion
						"Type argument for %s must be [%s; _]",
						sema.ArrayTypeToConstantSizedFunctionName,
						elementType,
```

turbolent (approved):
Nice!

turbolent:
@onflow/cadence could you please have a look? it would be great to get this into the next release, it will make using the EVM API much easier

dsainati1 (approved):

runtime/sema/type.go
```diff
@@ -2497,6 +2504,32 @@ func getArrayMembers(arrayType ArrayType) map[string]MemberResolver {
...
+				}
+
+				return NewFunctionMember(
+					memoryGauge,
+					arrayType,
+					insertableEntitledAccess,
+					identifier,
+					ArrayToConstantSizedFunctionType(elementType),
+					arrayTypeToConstantSizedFunctionDocString,
+				)
```
SupunS, line 2530: This should be a public function, since it doesn't mutate the existing one.
```suggestion
				return NewPublicFunctionMember(
					memoryGauge,
					arrayType,
					identifier,
					ArrayToConstantSizedFunctionType(elementType),
					arrayTypeToConstantSizedFunctionDocString,
				)
```

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -11295,6 +11295,225 @@ func TestInterpretArrayToVariableSized(t *testing.T) {
...
+					},
+					common.ZeroAddress,
+					value1,
+					value2,
+					value3,
+				),
+			),
+		)
+	})
+}
```
SupunS, line 11515: I think this also work on array references, correct? Can you maybe add a test?

I didn't realize this earlier: but since we already have `toVariableSized` function, with the introduction of this, one can get a dereferenced copy of an array, by going the round trip. The catch here is  dereferenced-copy feature doesn't work for all types of arrays, where as this would work for all non-resource typed arrays

runtime/sema/type.go
turbolent: Good catch, didn't notice this during review.

Opened #3031 to fix `toVariableSized`

runtime/tests/interpreter/interpreter_test.go
SupunS: Having `toVariableSized` alone itself is also not ideal, it in a way, dereferences the array.
We need a way to restrict certain functions from being called on references.

runtime/sema/type.go
SupunS: This should also be fixed for `toVariableSized`

runtime/tests/interpreter/interpreter_test.go
turbolent: Sorry, I'm not quite following here, what is the problem?

SupunS: The issue here is, the dereference operator only works for a subset of AnyStruct types (only primitives and collection of primitives).
But one can bypass that restriction by using these two methods, and 'dereference' an array of AnyStruct :
```cadence
fun dereference(arrayRef &[AnyStruct]): [AnyStruct] {
    constantSized = arrayRef.toConstantSized<AnyStruct; n>()
    return constantSized.toVariableSized()
}
```

turbolent: The dereferencing isn't recursive, so that shouldn't be a problem, no?

runtime/sema/type.go
darkdrag00nv2: Thanks!

darkdrag00nv2:
Had to force push after resolving merge conflicts.

runtime/tests/interpreter/interpreter_test.go
SupunS: So based on the discussion here: https://discord.com/channels/613813861610684416/1108479699732152503/1197977922057814117, this should be OK. We may instead update the dereference operator to be consistent.

SupunS (approved):

runtime/tests/interpreter/interpreter_test.go
turbolent: Opened to adjust the reference operator: https://github.com/onflow/cadence/pull/3034

PR #3036, opened 2024-01-20
Fix gaps in migrations

turbolent:
## Description

Fix some gaps and improve tests for migrations:

- `StorageMigration`: 
  - Also migrate `PublishedValue`
  - Test migration of nested values
  - Migrate all storage maps, i.e. also `inbox` and cap cons ones, not just "user accessible" ones (`public`, `private`, `storage`)
  - Test migration of `inbox`, published values with capabilities, and cap cons (account cap cons and storage cap cons)
- Entitlements migration:
  - Also migrate cap cons
  - Improve tests

______


dsainati1 (approved):

migrations/capcons/linkmigration.go
```diff
@@ -187,6 +194,15 @@ func (m *LinkValueMigration) Migrate(
 	), nil
 }
 
+func storageKeyToPathValue(
+	storageKey interpreter.StorageKey,
+	storageMapKey interpreter.StorageMapKey,
+) interpreter.PathValue {
+	domain := common.PathDomainFromIdentifier(storageKey.Key)
+	stringStorageMapKey, _ := storageMapKey.(interpreter.StringStorageMapKey)
```
SupunS, line 202: is the emtpy string a valid key? shouldn't we panic here if the conversion fail?

migrations/capcons/privatemigration.go
```diff
@@ -25,11 +25,13 @@ import (
...
 	storageMap *interpreter.StorageMap,
-	storageKey interpreter.StringStorageMapKey,
-	addressPath interpreter.AddressPath,
+	storageMapKey interpreter.StorageMapKey,
 ) {
-	if addressPath.Path.Domain == common.PathDomainPrivate {
-		storageMap.RemoveValue(inter, storageKey)
+	path := storageKeyToPathValue(storageKey, storageMapKey)
+
+	if path.Domain == common.PathDomainPrivate {
```
SupunS, line 34: just a nit: Could get just the domain here, instead of creating a new PathValue
```suggestion
	domain := common.PathDomainFromIdentifier(storageKey.Key)
	if domain == common.PathDomainPrivate {
```

SupunS (approved):

turbolent:
Just noticed that we also need to migrate the contract values, which are stored in a separate domain: [bf6e74c](https://github.com/onflow/cadence/pull/3036/commits/bf6e74c98a0009d08944eadbe044561bf3548d4e)

migrations/capcons/linkmigration.go
turbolent: The idea here was that we _try_ to convert the storage key and storage map key to a path, if it is one (valid domain, and storage map key must be a string). `PathDomainFromIdentifier` will return `PathDomainUnknown` if the storage key is not a path domain, and ignoring the error result in the path identifier being empty. Effectively, this is exactly the same as the "Go default value" for `PathValue` (all fields, domain and identifier, are empty/have the default value).

I'll make it clearer and return an additional bool

turbolent: Tried to improve it in 478e9e0ebec1256672483b2b2a35a9b0e10a67d7

migrations/capcons/privatemigration.go
turbolent: Improved in 478e9e0ebec1256672483b2b2a35a9b0e10a67d7

SupunS (approved):

Issue #2947, opened 2023-11-17
[EPIC] Enable testing of contract update to Cadence 1.0 in Emulator


j1010001:
### Issue to be solved

Why: Developers need to test their ability to upgrade their existing contract to Cadence 1.0 version. When we release Emulator running Cadence 1.0, this would not be possible because if the upload of a contract that does not parse/typecheck with Cadence 1.0 will fail. 

This step is required to collect data for the next stage, described in https://github.com/onflow/cadence/issues/2865.

```[tasklist]
### Tasks for Emulator 1.0 RC1 release (enables dapp testing):
- [x] Enable devs to deploy old (0.4x-based) Cadence contracts to emulator and then upgrade to C1.0. This will likely require emulator-only contracts.add method which skips type checking and enables any code to be uploaded, and parser that can parse contracts based on both v04 and 1.0 versions to perform the update restrictions. This requires the update checker to be able to parse both v0.42 and 1.0 contract versions.
- [ ] https://github.com/onflow/cadence/pull/3157
- [ ] https://github.com/onflow/cadence/pull/3158
- [ ] https://github.com/onflow/flow-go/pull/5486
- [ ] https://github.com/onflow/flow-go/pull/5479
- [ ] https://github.com/onflow/flow-cli/issues/1433
- [x] Release this emulator build to community - this enables devs to test their contract upgrade to C1.0.
```

j1010001:
Required for: https://github.com/onflow/flow-emulator/issues/468

j1010001:
Can be closed after the new Emulator build is released.

turbolent:
Is the requirement for closing this issue that the updated update checker has been completed, or is that a separate issue/milestone?

j1010001:
> Is the requirement for closing this issue that the updated update checker has been completed, or is that a separate issue/milestone?
> 
> I assume that work is part of #2865, in which case this issue can be considered in-review/done.

No, let's close this when the first version of this is out and create a new issue if the upgrade checker needs adjustments.

SupunS:
I guess we can close this issue now?

turbolent:
Yeah, I think we can close this now, given there's no known work left to be done. 🎉 👏 

We can maybe leave #3098 open until it's been confirmed that the latest CLI preview release works as expected.

PR #3039, opened 2024-01-23
Add v0.42 parser package under old_parser

dsainati1:
Work towards https://github.com/onflow/cadence/issues/2947

In order to have an update checker for v0.42 -> v1.0 contract upgrades, we need to be able to parse both versions of Cadence with the same binary. This ports the parser from v0.42 to v1.0 temporarily. 

Manual changes to the parser to make the code compile are here: https://gist.github.com/dsainati1/a076fc307ffdca3202aa951878a5133d

______


runtime/old_parser/type.go
```diff
@@ -0,0 +1,988 @@
...
+}
+
+func defineReferenceType() {
+	defineType(prefixType{
+		tokenType:    lexer.TokenAmpersand,
+		bindingPower: typeLeftBindingPowerReference,
+		nullDenotation: func(p *parser, right ast.Type, tokenRange ast.Range) ast.Type {
+			return ast.NewReferenceType(
+				p.memoryGauge,
+				nil,
```
dsainati1, line 329: Reference types parsed by the old parser are always created with a `nil` authorization. This should indicate to whatever code we use this for that the reference type was created in an older version of Cadence.

runtime/old_parser/type.go
```diff
@@ -0,0 +1,988 @@
...
+					}
+					if expectType {
+						return nil, p.syntaxError("unexpected comma in restricted type")
+					}
+					if restrictedType == nil {
+						firstNominalType, ok := firstType.(*ast.NominalType)
+						if !ok {
+							return nil, p.syntaxError("non-nominal type in restriction list: %s", firstType)
+						}
+						restrictedType = ast.NewIntersectionType(
```
dsainati1, line 376: Restricted type parsing in the old parser just ignores the `T` in `T{I}`. I think this is okay, but let me know if you think we should do this differently.

runtime/ast/access.go
```diff
@@ -215,6 +215,7 @@ const (
 	AccessContract
 	AccessAccount
 	AccessAll
+	PubSettableLegacy
```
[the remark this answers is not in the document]
SupunS: maybe add `// Deprecated` for these fields, so it's easier to find-and-remove later on.

SupunS (commented):
Were there any changes you had to do for the old parser to make it work?
I guess it only needed the AST changes that you commented above? (thanks for annotating them, makes it a lot easier to review!)

runtime/old_parser/type.go
SupunS: Does it include the `IsAuthorized` flag?

runtime/old_parser/type.go
SupunS: We may need to have the `T` as well. Because `T{I}` and `R{I}` are considered two different types, and certain update checking rules, such as changing fields types, would rely on `T`

runtime/old_parser/type.go
dsainati1: No, the 1.0 Cadence AST doesn't have support for the `IsAuthorized` flag, and I don't really think we need to add it; whether or not a reference was authorized in older versions of Cadence is irrelevant for their semantics w.r.t. migrations, so I don't think it's necessary for us to care about that information when doing upgrade checks.

runtime/old_parser/type.go
dsainati1: I added a legacy field to `IntersectionType` in 6c6947ca7b24591700760b6aef8723a3822db0da to track this.

dsainati1:
> Were there any changes you had to do for the old parser to make it work?

Only a few minor things (changing locations where we created a `*ast.Condition` to `*ast.TestCondition` since that type as changed to an interface, adding unspecified view modifiers to function expressions/declarations, that sort of thing) other than the ones related to the AST changes I pointed out.

runtime/old_parser/type.go
turbolent: Might be worth making this package independent of assumptions on how it will be used, and add it to the AST as a "legacy" field and then users of the package can decide to use it or not

turbolent:
Related to Supun's question above: Could you please post a diff between the code of the old parser in the v0.42 branch and the code that is being added here? That would help with being able ti review the changes that were made to it

dsainati1:
Yep, https://gist.github.com/dsainati1/a076fc307ffdca3202aa951878a5133d has the diff between the fixed code and the original.

runtime/ast/access.go
```diff
@@ -215,6 +215,7 @@ const (
 	AccessContract
 	AccessAccount
 	AccessAll
+	PubSettableLegacy // Deprecated
```
turbolent, line 218: Just like for the other constants, indicate the type by adding it as a prefix: `AccessPubSettableLegacy`

turbolent (approved):
LGTM

runtime/old_parser/type.go
dsainati1: Added this in [110bd77](https://github.com/onflow/cadence/pull/3039/commits/110bd77514aaab13d50cb9eab935e85c026b4da7)

SupunS (approved):

PR #3043, opened 2024-01-24
Use old parser for contract upgrades when legacy upgrade config option is set

dsainati1:
Part of https://github.com/onflow/cadence/issues/2947

This option causes the contract update checker to fall back to using the old parser if the new one fails. 

______

### DIFF
--- a/runtime/stdlib/account.go
+++ b/runtime/stdlib/account.go
@@ -29,6 +29,7 @@ import (
 	"github.com/onflow/cadence/runtime/common"
 	"github.com/onflow/cadence/runtime/errors"
 	"github.com/onflow/cadence/runtime/interpreter"
+	"github.com/onflow/cadence/runtime/old_parser"
 	"github.com/onflow/cadence/runtime/parser"
 	"github.com/onflow/cadence/runtime/sema"
 )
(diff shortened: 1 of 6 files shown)


runtime/stdlib/account.go
```diff
@@ -1593,16 +1594,37 @@ func changeAccountContracts(
 			},
 		)
 
+		var legacyContractUpgrade bool
+		// if we are allowing legacy contract upgrades, fall back to the old parser when the new one fails
+		if !ignoreUpdatedProgramParserError(err) && invocation.Interpreter.SharedState.Config.LegacyContractUpgradeEnabled {
+			legacyContractUpgrade = true
+			oldProgram, err = old_parser.ParseProgram(
+				invocation.Interpreter.SharedState.Config.MemoryGauge,
```
turbolent, line 1602: Maybe refactor `invocation.Interpreter.SharedState.Config` and `invocation.Interpreter.SharedState.Config.MemoryGauge` into variables, and use them in the new code, and in the new parser call above.

turbolent (approved):
Great work!

runtime/stdlib/account.go
```diff
@@ -1585,24 +1586,48 @@ func changeAccountContracts(
...
+			memoryGauge,
 			oldCode,
 			parser.Config{
 				IgnoreLeadingIdentifierEnabled: true,
 			},
 		)
 
+		var legacyContractUpgrade bool
+		// if we are allowing legacy contract upgrades, fall back to the old parser when the new one fails
+		if !ignoreUpdatedProgramParserError(err) && legacyUpgradeEnabled {
```
SupunS, line 1602: why do we need to parse with the new parser first and then fallback?
Can't we pick the correct parser straight-away based on  the `legacyUpgradeEnabled` flag?

dsainati1: I don't think it's necessarily the case that people are going to want exactly one or the other in any given emulator instance; e.g. someone may want to test sequencing updates from v0.42 -> v1.0 with no business logic changes, followed immediately by a second update that applies the business logic changes we told them not to include. IMO this adds more flexibility that enables that use case.

SupunS: makes sense!

SupunS (approved):

Issue #2836, opened 2023-10-03
Use atree's new read-only iterators when mutation of returned values are not required


fxamacker:
### Issue to be solved

New read-only iterators will be available in atree after https://github.com/onflow/atree/pull/345 is merged.

Using the read-only iterators will be more efficient, etc. for use cases that don't need to mutate returned values.

### Context

> @fxamacker Could you please open a tech-debt follow-up issue in the Cadence repo, so that we don't forget to switch Cadence to the new read-only versions where possible?

_Originally posted by @turbolent in https://github.com/onflow/atree/issues/329#issuecomment-1726636215_

### Suggested Solution

Use atree's new read-only iterators when mutation of returned values are not required.

fxamacker:
Closed by #3170

PR #3048, opened 2024-01-26
Atree register inlining for v1.0

turbolent:
## Description

Rebase of `feature/atree-register-inlining-v0.42` on top of `master`.

I attempted merging `master` and resolving conflicts multiple times, but the sheer amount of conflicts and differences made it impossible for me. Rebasing allowed making semantically equivalent changes, step-by-step.

______


turbolent:
- Verified the rebase matches the original by going over 
  ```sh
  git range-diff -w v0.42...feature/atree-register-inlining-v0.42 master...bastian/atree-register-inlining-v1.0
  ```
- Fixed the iteration functions

turbolent:
@fxamacker Could you please have a look at these new cases where a value for `hasNoParentContainer` has to be passed and let me know what needs to be passed? See the remaining TODOs in the diff:

- https://github.com/onflow/cadence/pull/3048/files#diff-424575c293dc2bc3a28aab845829d7656fa83a7be4e026de8906935e23bbd52eR3393
- https://github.com/onflow/cadence/pull/3048/files#diff-424575c293dc2bc3a28aab845829d7656fa83a7be4e026de8906935e23bbd52eR3454
- https://github.com/onflow/cadence/pull/3048/files#diff-424575c293dc2bc3a28aab845829d7656fa83a7be4e026de8906935e23bbd52eR20439

turbolent:
For some reason, the re-hash tests in the migrations panic with an error indicating that the iteration's key cannot be found, but I can't really see why that would be the case. This needs some further investigation

migrations/statictypes/intersection_type_migration_test.go
```diff
@@ -546,6 +546,8 @@ func TestIntersectionTypeRehash(t *testing.T) {
 				false,
 				nil,
 				nil,
+				// TODO:
+				false,
```
fxamacker, line 550:
```suggestion
				true, // dictValue is standalone
```

migrations/statictypes/intersection_type_migration_test.go
```diff
@@ -710,6 +712,8 @@ func TestRehashNestedIntersectionType(t *testing.T) {
 					false,
 					nil,
 					nil,
+					// TODO:
+					false,
```
fxamacker, line 716:
```suggestion
					true, // dictValue is standalone
```

migrations/statictypes/intersection_type_migration_test.go
```diff
@@ -849,6 +853,8 @@ func TestRehashNestedIntersectionType(t *testing.T) {
...
```
fxamacker, line 857:
```suggestion
					true, // dictValue is standalone
```

migrations/string_normalization/migration_test.go
```diff
@@ -245,6 +245,8 @@ func TestStringNormalizingMigration(t *testing.T) {
 			false,
 			nil,
 			nil,
+			// TODO:
+			false,
```
fxamacker, line 249:
```suggestion
			true, // storedValue is standalone
```

migrations/string_normalization/migration_test.go
```diff
@@ -382,6 +384,9 @@ func TestStringValueRehash(t *testing.T) {
 				false,
 				nil,
 				nil,
+				// TODO:
+				false,
```
fxamacker, line 388:
```suggestion
				true, // dictValue is standalone
```

migrations/string_normalization/migration_test.go
```diff
@@ -521,6 +526,8 @@ func TestCharacterValueRehash(t *testing.T) {
...
```
fxamacker, line 530:
```suggestion
				true, // dictValue is standalone
```

runtime/account_test.go
```diff
@@ -2277,9 +2277,15 @@ func TestRuntimePublicKeyPublicKeyField(t *testing.T) {
 		false,
 		nil,
 		nil,
+		// TODO:
+		false,
```
fxamacker, line 2281:
```suggestion
		true, // publicKey is standalone
```

runtime/account_test.go
```diff
@@ -2277,9 +2277,15 @@ func TestRuntimePublicKeyPublicKeyField(t *testing.T) {
...
 	).(*interpreter.CompositeValue)
 
-	publicKey.DeepRemove(inter)
+	publicKey.DeepRemove(
+		inter,
+		// TODO:
+		false,
```
fxamacker, line 2287:
```suggestion
		true, // publicKey is standalone
```

fxamacker (commented):
> @fxamacker Could you please have a look at these new cases where a value for hasNoParentContainer has to be passed and let me know what needs to be passed? See the remaining TODOs in the diff, e.g.:

@turbolent I searched for "TODO" and reviewed search results for new cases where a value for `hasNoParentContainer` has to be passed.

fxamacker:
> `InclusiveRangeStaticType.IsComposite`?

@turbolent Does `InclusiveRangeStaticType` need to implement `atree.TypeInfo` interface?  If not, we don't need to implement `InclusiveRangeStaticType.IsComposite`.

`DecodeTypeInfo` used by atree doesn't appear to handle decoding of `InclusiveRangeStaticType`.

https://github.com/onflow/cadence/blob/027e3c00f1731804b1dcb8d07aef9656be0f4ef9/runtime/interpreter/decode.go#L2145

fxamacker:
@turbolent Do you mind if I push a commit to fix the rehash migration test?

The fix involves multiple files so it would be difficult/confusing to do it in comments compared to pushing a commit.

turbolent:
@fxamacker Of course, please feel free to push commits 👍

turbolent:
@fxamacker Great work fixing the migrations! 👏

fxamacker:
>    Why doesn't CompositeStaticType implement TypeInfo?

@turbolent `CompositeStaticType` doesn't implement `atree.TypeInfo` is because we only creates `atree.OrderedMap` with `compositeTypeInfo` for `CompositeValue`.

https://github.com/onflow/cadence/blob/c8b011058b6a178f66b2d931cbec105606fd643a/runtime/interpreter/value.go#L16665-L16675

fxamacker:
> Out of curiosity: why were the migrations working in v0.42 without those equality functions?

Those equality functions are for Cadence 1.0 migration and are not part of Cadence v0.42 migration with atree inlining.

turbolent:
The only remaining TODO here is to fix one of the subtests ("resource, move from array to array") in `TestNestedContainerMutationAfterMove`, which I skipped for now.

It's been a while, but as far as I can remember I added this test case in the atree register inlining feature to "prepare" for the changes to come, to basically have a regression test for "complex" operations on containers, like moving children between different containers.

I think I need to update it to the new reference semantics in 1.0, i.e. references becoming invalid when the resource is moved.

@SupunS does this sound about right?

turbolent:
Adjusted and unskipped the test case in 6713e43fcdcd0722c92365abbae118adb241a578

SupunS:
adjustment for the test 6713e43fcdcd0722c92365abbae118adb241a578 looks correct 👍

fxamacker (approved):
Nice! :+1:   I reviewed mostly non-test code against the other branch.  I manually looked for any commits that might have been missed because the `git range-diff` had issues (maybe my bad).

There are some places in value.go that can use readonly iterators, such as `ArrayValue.IsImportable()`.  I can open a separate PR for those, which would also address issue #2836.

runtime/storage_test.go
```diff
@@ -1576,11 +1576,14 @@ func TestRuntimeStorageTransfer(t *testing.T) {
 			nonEmptyKeys++
 		}
 	}
-	// 5:
+
+	// TODO: maybe retrieve and compare stored values from 2 accounts
```
dsainati1, line 1580: Is this still unresolved?

dsainati1 (approved):
Took a look at the diff of diffs and it looks similar enough to me.

runtime/storage_test.go
turbolent: Yeah, this TODO was just taken as it is from v0.42. We might want to resolve it to get more confidence before shipping, but it's not critical AFAICS. cc @fxamacker

fxamacker: > Yeah, this TODO was just taken as it is from v0.42. We might want to resolve it to get more confidence before shipping, but it's not critical AFAICS. cc @fxamacker

@turbolent I agree.  We can resolve this TODO in a separate PR.

Issue #1151, opened 2021-09-28
Request for additional math functions: ln(x), log(x,b), pow(x,y), sqrt(x), exp(x)


figs999:
### Issue To Be Solved

Currently cadence does not support any mathematical operations more advanced than basic arithmetic.
When making bonded curves or proportional staking calculations, additional operations are required.

### Suggested Solution

Please add the following mathematical functions:
ln(x), log(x,b), pow(x,y), sqrt(x), exp(x)
Natural Logarithm, Logarithm with arbitrary base, power (x to the y), square root, and exponent (euler's-number to the x)

Functions needed for all applicable numeric types. Currently these functions are being implemented in cadence script by developers, which adds needless overhead to execution and introduces the risk of bugs in the math algorithms. 
It would be more efficient for the script execution runtime to use the math operators available in it's base language.

These would ideally be accessible as language built-in-functions, but could also be placed inside the Numeric types like min and max.

Pow(x,y) could be implemented via an operator, such as in other languages using "x^y" or "x**y".

darkdrag00nv2:
@turbolent 

I think `exp(x)` is not that useful. The return type of `exp(x)` has to be `UFix64` whose maximum value is `184467440737.09551616`.

```go
loge(184467440737.09551616) = 25.940738811884
```

So `exp(x)` would only work for `x <= 25.940738811884` which won't be that useful.

Similar logic for `pow(x,y)` if the return type is a `UFix64`. So, we should probably restrict it to just integers base and exponents and keep the return type as `Int`.

turbolent:
Yeah, we might want to consider adding larger fixed-point types (128, 256, ...), potentially even without a min/max and just a fixed scale ("UFix"/"Fix", like `UInt` and `Int` - question is what the scale would be).

darkdrag00nv2:
yeah, agreed.

Do you see any issues in making the scale flexible for `UFix` as well?  Apart from it no longer being "Fixed point" :slightly_smiling_face: 

Kind of like how `big.Float` allows `SetPrec`. Can have an upper limit on the scale such as 16 or 32 to ensure the computations are feasible.

```cadence
let a = UFix.fromString("10000000000000.22222222", /*scale*/ 16);
```

For computation with different scales - say adding a `UFix` with scale 8 with another with scale 16, we could choose the larger one.

turbolent:
Yeah, I guess that makes sense. It would be best to not reinvent the wheel here and find an existing design and implementation we could follow.

[...]

PR #3055, opened 2024-01-28
Introduce `Sqrt` on `NumberType`

darkdrag00nv2:
Work towards #1151

## Description

Introduce `Sqrt` on `NumberType` defined in a `Math` contract.

```cadence
Sqrt<T: NumberType>(value : T) : UFix64
```

1. Panics with `UnderflowError` if `value < 0`.
2. Panics with `OverflowError` if the result does not fit in a `UFix64`.
______


darkdrag00nv2:
Haven't added any tests yet. Opened it up for some early feedback.

- Are we okay with making this a top level function in the standard library?
- Does the implementation look correct? Mainly the `BigIntSqrt` function in the `runtime/interpreter/big.go` file.
- Should we return a `Optional` and return `Nil` in case of overflows.

turbolent:
Nice work!

> - Are we okay with making this a top level function in the standard library?

Maybe we can introduce a Math library/contract? (similar to existing RLP, BLS, Crypto, etc.; similar to how there's a "math" package in some languages). 
Alternatively we can also define it as a function on all `Number` types, as `.squareRoot()` (how e.g. Swift does it: https://developer.apple.com/documentation/swift/double/squareroot())

darkdrag00nv2:
@turbolent I like both the ideas.

@SupunS @dsainati1 Do you have opinions? I'll go with a separate math contract if both seem fine to you all.

dsainati1:
A `Math` contract seems the most appropriate to me

runtime/interpreter/big.go
```diff
@@ -140,3 +141,31 @@ func BigEndianBytesToSignedBigInt(b []byte) *big.Int {
...
+		res.Mul(res, new(big.Float).SetPrec(256).SetInt(sema.Fix64FactorBig))
+
+		resInt := new(big.Int)
+		res.Int(resInt)
+
+		if !resInt.IsUint64() {
+			panic(OverflowError{
+				LocationRange: locationRange,
+			})
+		}
```
AlexHentschel, line 165: Referring to the issue https://github.com/onflow/cadence/issues/1151
> ln(x), log(x,b), pow(x,y), sqrt(x), exp(x)

In my opinion, these functions are most useful if their output type is a floating point number. This is the overwhelming amount of use cases, the inputs will not permit an integer type output. I am happy with the inputs being integers, float, etc for convenience. Though, the output could be a float. 

There are cases, where the developer knows that the input `x` for `sqrt(x)` satisfies  `x = y^2` for some non-negative integer `y`.  But this is rarely that case, and developers could use functions like `round`, `floor`, `ceiling` to convert the float to an integer.

turbolent: Cadence does not have floating point numbers, only [fixed-point numbers](https://cadence-lang.org/docs/language/values-and-types#fixed-point-numbers). Like for floating point numbers, the precision/scale will only make certain values representable. Over/underflow is automatically checked and leads to a program abort.

darkdrag00nv2: > In my opinion, these functions are most useful if their output type is a floating point number. This is the overwhelming amount of use cases, the inputs will not permit an integer type output. I am happy with the inputs being integers, float, etc for convenience. Though, the output could be a float.

@AlexHentschel I think there is some confusion. The output of the `Sqrt` function is always a fixed-point number (not an integer) which is the closest type we have in Cadence to a floating point number.

`Sqrt<T>` will always return `UFix64` i.e. 64-bit fixed point number.

darkdrag00nv2: I think the confusion is due to the line that you have highlighted.

```
		res.Mul(res, new(big.Float).SetPrec(256).SetInt(sema.Fix64FactorBig))

		resInt := new(big.Int)
		res.Int(resInt)

		if !resInt.IsUint64() {
			panic(OverflowError{
				LocationRange: locationRange,
			})
		}
```

This snippet is not converting the result to an integer. Notice the multiplication by `Fix64FactorBig` (10^8). This is only truncating the output after 8 decimal places which is the level of precision `UFix64` supports.

runtime/interpreter/big.go
```diff
@@ -140,3 +141,31 @@ func BigEndianBytesToSignedBigInt(b []byte) *big.Int {
...
 }
+
+func BigIntSqrt(interpreter *Interpreter, value *big.Int, locationRange LocationRange) UFix64Value {
+	if value.Sign() < 0 {
+		panic(UnderflowError{
+			LocationRange: locationRange,
+		})
+	}
+
+	valueFloat := new(big.Float).SetPrec(256).SetInt(value)
```
AlexHentschel, line 152: I am making an assumption here:
```
By protocol convention, Cadence integer values are guaranteed to
floating-point value with 256 bit precision _without_ truncation. 
```
If that holds, I would suggest to explicitly document it:
```suggestion
	// By protocol convention, Cadence integer values are guaranteed to fit into a
	// floating-point value with 256 bit precision _without_ truncation. 
	valueFloat := new(big.Float).SetPrec(256).SetInt(value)
```

If that doesn't hold, we need to look at the rounding convention (e.g. adopting the [conventions from C](https://wiki.sei.cmu.edu/confluence/display/c/FLP34-C.+Ensure+that+floating-point+conversions+are+within+range+of+the+new+type))

runtime/interpreter/big.go
```diff
@@ -140,3 +141,31 @@ func BigEndianBytesToSignedBigInt(b []byte) *big.Int {
...
+	res := new(big.Float).SetPrec(256).SetMode(big.ToZero).Sqrt(valueFloat)
```
AlexHentschel, line 153: I had to look this up, so maybe its worth a comment:
```suggestion
	// Converting the result to a fixed-point number, we are conceptually converting it to an integer 
	// IEEE 754 specifies different rounding modes https:	//en.wikipedia.org/wiki/IEEE_754#Rounding_rules
	// We follow the "Rationale for International Standard -- Programming Languages -- C", Revision 5.10, April-2003:
	//   > Section 6.3.1.5 Real floating types:
	//   > When a finite value of real floating type is converted to an integer type other than Bool, 
	//   > the fractional part is discarded (i.e., the value is truncated toward zero). If the value
	//   > of the integral part cannot be represented by the integer type, the behavior is undefined.
	// For details, see
	// https:	//wiki.sei.cmu.edu/confluence/display/c/FLP34-C.+Ensure+that+floating-point+conversions+are+within+range+of+the+new+type
	res := new(big.Float).SetPrec(256).SetMode(big.ToZero).Sqrt(valueFloat)
```

runtime/interpreter/big.go
```diff
@@ -140,3 +141,31 @@ func BigEndianBytesToSignedBigInt(b []byte) *big.Int {
...
+
+	valueGetter := func() uint64 {
+		res.Mul(res, new(big.Float).SetPrec(256).SetInt(sema.Fix64FactorBig))
+
+		resInt := new(big.Int)
+		res.Int(resInt)
```
AlexHentschel, line 159: why are we computing this in the getter (presumably multiple times if the value is accessed multiple times)? Can we can we move that outside of `valueGetter`?

runtime/interpreter/big.go
AlexHentschel: Thanks @turbolent for the clarification. I think I am lacking some necessary context here: while I interpreted `UFix64Value` as an integer value type, I realize now that it is a [Rational Number](https://en.wikipedia.org/wiki/Rational_number) representation. That is great. Sorry for my lack of context and thanks for explaining. 

Just to make sure we are on the same page:
https://github.com/onflow/cadence/blob/90b485fe2ada7dc3b752981a22a97bf9f4044d7c/runtime/interpreter/big.go#L167
A fixed-point number essentially means we have some notion of "smallest atomic unit" that cannot be split further. Anything smaller than a "smallest atomic unit" is truncated to zero. This `uint64` counts the number of "smallest atomic unit". Is that correct?

AlexHentschel (commented):
I am less familiar with fixed point numbers, so my confusion maybe it is just my lack of context. Nevertheless, it might be useful to add some documentation making it easier for other people reading to code to follow, even if they are not experts in fixed-point number arithmetics. Please see my suggestions above (specifically [this](https://github.com/onflow/cadence/pull/3055#discussion_r1481969280) and [this comment](https://github.com/onflow/cadence/pull/3055#discussion_r1481988842))

runtime/interpreter/big.go
darkdrag00nv2: Yes, sounds correct to me.

UFix64 represents floating point number up to a fixed precision (8 decimal places).

So if you want to represent Pi in Cadence which is `3.14159265359...`, you could only store it up to 8 decimal places in `UFix64`. So you can only represent it as:

```
let pi = UFix64.fromString("3.14159265");  // notice that I only included 8 decimal places.
```

The underlying value stored inside `UFix64` will be the value multiplied by 10^8 which is `314159265` ("smallest atomic unit" of your explanation).

cc: @turbolent

runtime/interpreter/big.go
darkdrag00nv2: This is a very nice call-out. Made me realize a missing check. Thanks!

> By protocol convention, Cadence integer values are guaranteed to fit
into floating-point values with 256 bit precision _without_ truncation. 

The assumption is true for most Cadence integer types but not for `Int` and `UInt` which are arbitrary precision numbers.

That being said, since the output of `Sqrt` is a UFix64 (max value`184467440737.09551615`), the maximum argument value that the `Sqrt` function can work on is `34028236692093846346337`. See this calculation in `runtime/math_test.go`.

So, we need two things:
1. `Sqrt` function should validate that the input integer is <=  `34028236692093846346337`. I haven't done that yet.
2. Once the limit is in place, we can safely state that all Cadence integer values **supported by the Sqrt function** are guaranteed to fix into floating-point values with 256 bit precision without truncation.

cc: @turbolent FYI

runtime/interpreter/big.go
AlexHentschel: thanks @darkdrag00nv2 for the detailed explanation. Really helpful!

runtime/interpreter/big.go
darkdrag00nv2: Done in 405fed080

runtime/interpreter/big.go
darkdrag00nv2: Added the overflow check and the comment in b2f799b13.

darkdrag00nv2:
Thanks for the review comments @AlexHentschel. Have incorporated your comments. PTAL again, thanks!

AlexHentschel (approved):
👏

runtime/interpreter/big.go
```diff
@@ -140,3 +141,46 @@ func BigEndianBytesToSignedBigInt(b []byte) *big.Int {
...
 }
+
+func BigIntSqrt(interpreter *Interpreter, value *big.Int, locationRange LocationRange) UFix64Value {
+	if value.Sign() < 0 {
+		panic(UnderflowError{
+			LocationRange: locationRange,
+		})
+	}
+
+	if value.Cmp(sema.MaxSquareIntegerBig) == 1 {
```
tarakby, line 152: When calling this check in the case of `Fix64Value` and `UFix64Value`, I think we would exclude some values because `value` is larger than the max, but in reality it represents a number less than `max` (because `value` is multiplied by the factor). Actually, I think all `Fix64` and `UFix64` values are less than the `max`, so the result (square root) can always fit.

If I didn't miss anything and my comment makes sense, it may be better to have a specific function `BigIntSqrtFloat` to process the case of `UFix64` and `Fix64` skipping the overflow (also to add the optimization from my previous comment)

tarakby (commented):
Great work and nice documentation 👌🏼
I left a few comments below.

Issue #3047, opened 2024-01-25
Error messages with authorizations are syntactically invalid


turbolent:
### Current Behavior

Error messages which refer to authorizations may be rendered in a way which is not syntactically valid, e.g.:

```
error: invalid use of interface as type
 --> f3e79be9d83f8f1a30fb98f67317da1625069d6adbe4f51edf470173a5c4f06d:8:41
  |
8 | 							let vaultRef = acc.storage.borrow<auth(Withdraw) &FungibleToken.Vault>(from: /storage/flowTokenVault)
  | 							                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ got `auth(not specified) &FungibleToken.Vault`; consider using `auth(not specified) &{FungibleToken.Vault}`
        	            	
```

```
        	            	error: cannot access `withdraw`: function requires `Withdraw` authorization, but reference only has `not specified` authorization
```

### Expected Behavior

- `auth(not specified) &T` should probably shown as `&T`?
- ```
  has `not specified` authorization
  ```
  should probably be 
  ```
  has no authorization
  ```

### Steps To Reproduce

See above

### Environment

```markdown
- Cadence version:
- Network:
```

turbolent:
cc @dsainati1

turbolent:
Probably need to not skip invalid entitlement types

PR #3056, opened 2024-01-29
Don't skip invalid entitlements when creating errors

dsainati1:
Closes https://github.com/onflow/cadence/issues/3047

When creating an entitlement set where some of the elements are invalid types (i.e. because they are not declared), instead of defaulting to a `not specified` access, create an "invalid" entitlement type in their place for cleaner error messages. 
______

### DIFF
--- a/runtime/sema/checker.go
+++ b/runtime/sema/checker.go
@@ -2119,8 +2119,9 @@ func (checker *Checker) accessFromAstAccess(access ast.Access) (result Access) {
 						)
 					}
 				}
-				semanticEntitlements = append(semanticEntitlements, NewEntitlementType(checker.memoryGauge, checker.Location, "<<INVALID>>"))
-				continue
+				// construct a new "entitlement" type from the nominal type in the AST, but mark it as invalid for better error messages
+				entitlementType = NewEntitlementType(checker.memoryGauge, checker.Location, entitlement.Identifier.Identifier)
+				entitlementType.isInvalid = true
 			}
 			semanticEntitlements = append(semanticEntitlements, entitlementType)
 		}
--- a/runtime/sema/checker.go
+++ b/runtime/sema/checker.go
@@ -2119,8 +2119,8 @@ func (checker *Checker) accessFromAstAccess(access ast.Access) (result Access) {
 						)
 					}
 				}
-				result = PrimitiveAccess(ast.AccessNotSpecified)
-				return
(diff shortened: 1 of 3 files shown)


runtime/sema/checker.go
```diff
@@ -2119,8 +2119,8 @@ func (checker *Checker) accessFromAstAccess(access ast.Access) (result Access) {
 						)
 					}
 				}
-				result = PrimitiveAccess(ast.AccessNotSpecified)
-				return
+				semanticEntitlements = append(semanticEntitlements, NewEntitlementType(checker.memoryGauge, checker.Location, "<<INVALID>>"))
```
turbolent, line 2122: I see how we need a `*EntitlementType` here, and `InvalidType` is not one.

The new `EntitlementType` constructed here does not indicate itself that it is an invalid type. Maybe  add a flag to `EntitlementType` to indicate if it is invalid, and return it from  `EntitlementType.IsInvalidType`.

Instead of the magic string it should then likely also be fine to use the identifier of the AST element (`entitlement`)

runtime/tests/checker/errors_test.go
```diff
@@ -107,3 +108,83 @@ func TestCheckErrorShortCircuiting(t *testing.T) {
...
+					refR.foo()
+					destroy r
+				}
+            `,
+		)
+
+		errs := RequireCheckerErrors(t, err, 2)
+
+		require.IsType(t, &sema.NotDeclaredError{}, errs[0])
+		require.IsType(t, &sema.InvalidAccessError{}, errs[1])
```
turbolent, line 159: Maybe use `require.ErrorAs` here to cast/assert in one place, and avoid the cast below.
Same for the other sub-test

runtime/tests/checker/errors_test.go
```diff
@@ -107,3 +108,83 @@ func TestCheckErrorShortCircuiting(t *testing.T) {
...
+		require.Equal(t, "cannot access `foo`: function requires `E` authorization, but reference only has `<<INVALID>>` authorization", errs[1].(*sema.InvalidAccessError).Error())
```
turbolent, line 160: Please wrap long lines

turbolent (commented):
👍

turbolent (approved):
👌

SupunS (approved):

PR #3057, opened 2024-01-29
Add tests for using keywords as field names

### LINKED ISSUE #2998: Analyze usage of restricted keywords in composite fields on Mainnet
As @bjartek pointed out, some resources might be using field names that will be restricted keywords in Cadence 1.0.
We need to analyze contracts deployed on Miannet to identify those instances so that we can with with the developer to migrate the on-chain data for those cases, if any exist.

SupunS:
Closes #2998

______

### DIFF
--- a/runtime/parser/declaration_test.go
+++ b/runtime/parser/declaration_test.go
@@ -9709,3 +9709,26 @@ func TestParseDeprecatedAccessModifiers(t *testing.T) {
 		)
 	})
 }
+
+func TestParseKeywordsAsFieldNames(t *testing.T) {
+
+	t.Parallel()
+
+	for _, keyword := range []string{
+		"event",
+		"contract",
+		"default",
+	} {
+		keyword := keyword
+
+		t.Run(keyword, func(t *testing.T) {
+			t.Parallel()
+
+			_, errs := testParseDeclarations(fmt.Sprintf(
+				"struct Foo { var %s: String }",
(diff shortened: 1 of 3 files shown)


runtime/parser/declaration_test.go
```diff
@@ -9709,3 +9709,26 @@ func TestParseDeprecatedAccessModifiers(t *testing.T) {
...
+
+	for _, keyword := range []string{
+		"event",
+		"contract",
+		"default",
+	} {
+		t.Run(keyword, func(t *testing.T) {
+			t.Parallel()
+
+			keyword := keyword
```
turbolent, line 9725: Need to be moved before `t.Run`

runtime/tests/checker/composite_test.go
```diff
@@ -2236,3 +2236,35 @@ func TestCheckNativeFieldDeclaration(t *testing.T) {
...
+
+	for _, keyword := range []string{
+		"event",
+		"contract",
+		"default",
+	} {
+		t.Run(keyword, func(t *testing.T) {
+			t.Parallel()
+
+			keyword := keyword
```
turbolent, line 2252: Same as above

turbolent (approved):
Nice!

Maybe this PR can be used to also add this suggestion: https://github.com/onflow/cadence/issues/2998#issuecomment-1915855580

PR #3065, opened 2024-01-31
Fix entitlements migration

turbolent:
## Description

Discovered by @fxamacker in #3048 and fixed there through d0b144b1894802caaf89c5a424cba2fbc552c1e9:
New dictionaries, like other containers (e.g. arrays, which already do), need to be created in the same account as the original (source) value.

Also, defensively check that the owner of the converted value matches that of the original value (if available).

______

### DIFF
--- a/migrations/entitlements/migration.go
+++ b/migrations/entitlements/migration.go
@@ -281,10 +281,11 @@ func ConvertValueToEntitlements(
 			return true
 		})
 
-		return interpreter.NewDictionaryValue(
+		return interpreter.NewDictionaryValueWithAddress(
 			inter,
 			interpreter.EmptyLocationRange,
 			dictionaryStaticType,
+			v.GetOwner(),
 			values...,
 		), nil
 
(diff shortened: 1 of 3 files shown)


migrations/entitlements/migration.go
```diff
@@ -281,10 +281,11 @@ func ConvertValueToEntitlements(
 			return true
 		})
 
-		return interpreter.NewDictionaryValue(
+		return interpreter.NewDictionaryValueWithAddress(
 			inter,
 			interpreter.EmptyLocationRange,
 			dictionaryStaticType,
+			v.GetOwner(),
```
SupunS, line 288: Interesting: should we do the same for all places where dictionary values are created? e.g: in migration framework

SupunS (approved):

migrations/entitlements/migration.go
turbolent: Yeah, I wondered about that too and searched for `NewDictionaryValue*` in `/migrations`. 
However, only the entitlements migration creates new containers, and the cases for other containers (like for arrays) already correctly construct new values with the proper address.

dsainati1 (approved):
😮 thanks for fixing

Issue #2865, opened 2023-10-11
More permissive contract upgrade checker for 1.0 spork


j1010001:

bluesign:
I was checking contract update logic recently, actually it is pretty simple. Maybe listing all restrictions and going over them one by one has some benefit also. Some of them are must-have in any case.

turbolent:
We had some discussions around this, and an alternative to updating the contract update checker could be to remove the requirement of contract updates to pass the checker – instead of statically checking contract updates on-chain, the checker would rather be just an off-chain tool that developers can run by themselves. The contract update functionality would be reduced to just updating the given code as-is. Developers would need to make sure the code is updated correctly.

We would have to go over all rules of the contract update checker (https://cadence-lang.org/docs/1.0/contract-upgrades) and ensure that such changes are gracefully handled at run-time, by ensuring appropriate test cases exist, by adding missing test cases, and by adding missing graceful recovery where missing.

This would allow the network upgrade for Cadence 1.0 to update all contracts, core contracts and staged "user" contracts, to be updated as-is, in one step.

turbolent:
Cadence 1.0 introduces resource destruction events, which allow referring to fields of the reference. 
For example, when the following resource is emitted, the event contains the value of the field `answer`.

```cadence
resource R {
    let answer: Int
    init(answer: Int) {
        self.answer = answer
    }

    event ResourceDestroyed(answer: Int = self.answer)
}
```

However, if the field `answer` got added through a contract update, there might be stored resource in the state which do not have a value for the field yet (the contract update only updates the code, but not stored values).

Though accessing a field which does not have a value is gracefully handled by Cadence, by aborting the program with a user error, this is not acceptable when the resource destruction event is emitted / the resource is destroyed.

Rejecting _any_ member (field) access in a resource destruction event (e.g. `self.answer`) would render those events useless.

@dsainati1 noted that we could consider disabling the contract update checker only _temporarily_ for the Cadence 1.0, and re-enabling it again afterwards.

Also, as discussed in #2170, we might define the semantics of resource destruction events to be that they are not guaranteed to be emitted, but when possible. This could potentially also result in more kinds of expressions that might abort and are currently rejected to be allowed again.

j1010001:
Changing the scope, removing: "collect feedback from community projects (data for checker rules update) - getting their updated contracts, diffing them and evaluating how that affects the upgrade checker."

We will release first version of the upgrade checker asap and deal with any reported issues by opening separate issue per reported problem.

PR #3069, opened 2024-02-02
Contract Update Validator

dsainati1:
Part of https://github.com/onflow/cadence/issues/2865

A more permissive contract upgrade validator for converting v0.42 contracts to v1.0

This includes:
* Relaxing the restriction against removing interface conformances
* Allowing composite field types to be modified, under specific circumstances:
    * A `R{I}` restricted type in a field annotation must be replaced with an `R` type
    * An `&T` reference type may be granted any set of entitlements less permissive than the actual set that `T` supports according to the entitlements migration. 

______


runtime/stdlib/legacy_contract_upgrade_validator.go
```diff
@@ -0,0 +1,359 @@
...
+	// if the new authorization is not an entitlements set, there's nothing to check here
+	if !isEntitlementsSet {
+		return nil
+	}
+
+	switch newReferencedType := newType.Type.(type) {
+	// a lone nominal type must be a composite
+	case *ast.NominalType:
+		expectedAccess := validator.expectedAuthorizationOfComposite(newReferencedType)
+		return validator.requirePermitsAccess(expectedAccess, foundEntitlementSet, newReferencedType)
```
dsainati1, line 224: Currently this allows any set of entitlements with equal or less access permitted than the "actual" access granted by the migration. E.g. if some reference `&T` will be given `auth(E, F)` by the migration, users can update that type in a field to any of `&T`, `auth(E) &T`, `auth(F) &T`, `auth(E | F) &T` or `auth(E, F) &T`. 

Is this okay, or should we be more restrictive and require that users only migrate their field types to exactly what the migration will grant?

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
```diff
@@ -0,0 +1,372 @@
...
+	"fmt"
+
+	"github.com/onflow/cadence/runtime/ast"
+	"github.com/onflow/cadence/runtime/common"
+	"github.com/onflow/cadence/runtime/common/orderedmap"
+	"github.com/onflow/cadence/runtime/errors"
+	"github.com/onflow/cadence/runtime/sema"
+)
+
+type CadenceV042ToV1ContractUpdateValidator struct {
```
SupunS, line 31: I renamed this from `LegacyContractUpdateValidator` to `CadenceV042ToV1ContractUpdateValidator` to be more precise/intuitive about what this validator does.

bluesign:
Is this only for staged updates?

> Relaxing the restriction against removing interface conformances

this may be storage breaking I guess. 

> A R{I} restricted type in a field annotation must be replaced with an R type

related to this, what happens to existing fields like @FungibleToken.Vault types etc ? they need to be able to change to @{FungibleToken.Vault} no ?

SupunS:
> Is this only for staged updates?

Yes.

> > Relaxing the restriction against removing interface conformances
>
> this may be storage breaking I guess.

How come? do you have an example?

bluesign:
> How come? do you have an example?

If I had let's say: [{I}] and put inside A:I , now after update I can still validly keep [{I}] but it will contain A which not implementing I no?

SupunS:
> what happens to existing fields like @FungibleToken.Vault types etc ? they need to be able to change to @{FungibleToken.Vault} no?

Yes, correct.
The changes in this PR are just a subset of all the changes we may need. I am currently trying to mak a list of all potential changes that we might need. I had included the scenario you mentioned in the list/doc, but didn't have a good example, so thank you for bringing up the `@FungibleToken.Vault`!
I will share the doc soon, it is still in draft state

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validation_test.go
```diff
@@ -0,0 +1,1265 @@
...
+                access(all) var a: @{I} 
+                init() {
+                    self.a <- create R()
+                }
+            }
+        `
+
+		err := testContractUpdate(t, oldCode, newCode)
+
+		// This is not allowed because `@R{I}` is converted to `@R`, not `@{I}`
```
SupunS, line 815: I think this should be a valid change for restricted types. i.e: both
- `R{I}` -> `R`
- `R{I}` -> `{I}`

should be valid changes.

SupunS (approved):
Nice!

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
```diff
@@ -0,0 +1,372 @@
...
+
+	// working under the assumption that any new program we are validating already typechecks,
+	// any nominal type must fall into one of three cases:
+	// 1) type qualified by an import (e.g `C.R` where `C` is an imported type)
+	// 2) type qualified by the root declaration (e.g `C.R` where `C` is the root contract or contract interface of the new contract)
+	// 3) unqualified type (e.g. `R`, but declared inside `C`)
+	//
+	// in case 3, we prepend the root declaration identifier with a `.` to the type's string to get its qualified name,
+	// and in 1 and 2 we don't need to do anything
+	typIdentifier := typ.Identifier.Identifier
```
turbolent, line 116: Does this need to handle `ast.NominalType`'s `NestedIdentifiers`?

turbolent (commented):
Great work!

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
turbolent: In case that's the case, let's improve it in a follow-up PR and get this PR in, as it blocks a few other follow-up PRs already.

SupunS:
I'll merge the PR in its current state. Any improvements/fixes can be done in follow-up PR like discussed above.

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
SupunS: not sure. I didn't run into any issues when running this agains the core contracts. But it does looks like we may need to handle `NestedIdentifiers` (or atleast skip if there are already nested identifiers?)

+1 for doing the improvements separately. Need one more approval to merge the PR.

runtime/interpreter/interpreter.go
```diff
@@ -4577,6 +4577,24 @@ func (interpreter *Interpreter) getElaboration(location common.Location) *sema.E
...
+
+	elaborations = map[common.Location]*sema.Elaboration{}
+
+	// Ensure the program for this location is loaded,
+	// so its checker is available
+
+	for location := range interpreter.SharedState.allInterpreters { //nolint:maprange
+		subInterpreter := interpreter.EnsureLoaded(location)
+		if subInterpreter == nil || subInterpreter.Program == nil {
+			return nil
```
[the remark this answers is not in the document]
SupunS: I think it makes sense to panic

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
turbolent: 👍 Could you please add a TODO to #2642?

turbolent (approved):

runtime/interpreter/interpreter.go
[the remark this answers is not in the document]
turbolent: I'll change it. This new function is also more or less contains a duplicate of `getElaboration`, so I'll just call that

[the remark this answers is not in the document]
turbolent: Also noticed the `//nolint:maprange` here is not safe, will fix that as well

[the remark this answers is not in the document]
SupunS: Thank you!

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
SupunS: https://github.com/onflow/cadence/issues/3127

runtime/interpreter/interpreter.go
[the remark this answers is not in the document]
turbolent: Fixed in [65faf29](https://github.com/onflow/cadence/pull/3069/commits/65faf291ca63f273f63b6332601071e8430faefa)

runtime/stdlib/legacy_contract_upgrade_validator.go
turbolent: @SupunS Did we address this in follow up PRs? Do we still need to discuss / address it?

SupunS: ah no, I actually missed this somehow. Created an issue: https://github.com/onflow/cadence/issues/3128

PR #3083, opened 2024-02-07
Allow analysis.Load to continue with parsing/checking errors

### LINKED ISSUE #3087: Allow custom handers for checker & parser errors in analysis.Config
### Issue to be solved

Currently, if `analysis.Load` encounters an error while checking or parsing the program, it will terminate the analysis.  However, this is not always desired and useful diagnostics may be produced by a semantically invalid program.

Notably, the Cadence Language Server effectively deploys its own implementation of `analysis.Load` which will not terminate on these errors, but instead surface them as diagnostics themselves.

It would be useful for the user to have the option to handle these errors and choose whether to continue the analysis, so that if they want this behaviour, they do not need to take the approach of the language server (parsing, checking, constructing analysis.Program) and can just use the official API

### Suggested Solution

Add config fields to handle these errors.

jribbink:
Closes #3087

## Description

Allow for custom checker/parser handlers to be configured in the analysis tool. These errors may be non-fatal & this gives the option to choose whether to suppress them and continue analysis.

Unblocks https://github.com/onflow/cadence-tools/pull/285. Pre-cadence 1.0 code is not semantically valid in cadence 1.0 checker so these errors will be present. In order to run the cadence 1.0 analyzer, a mechanism needs to exist to suppress these errors. This is similar to the behaviour that already exists in the language server.

______

### DIFF
--- a/tools/analysis/analysis_test.go
+++ b/tools/analysis/analysis_test.go
@@ -259,9 +259,106 @@ func TestCheckError(t *testing.T) {
 	require.ErrorAs(t, err, &checkerError)
 }
 
+func TestHandledParserError(t *testing.T) {
+
+	t.Parallel()
+
+	contractAddress := common.MustBytesToAddress([]byte{0x1})
+	contractLocation := common.AddressLocation{
+		Address: contractAddress,
+		Name:    "ContractA",
+	}
+	const contractCode = `
+	  access(all) contract ContractA {
+	    init() {
+	      ???
+	    }
+	  }
+	`
+
+	config := &analysis.Config{
+		Mode: analysis.NeedSyntax,
+		ResolveCode: func(
+			location common.Location,
+			importingLocation common.Location,
+			importRange ast.Range,
+		) ([]byte, error) {
+			switch location {
+			case contractLocation:
+				return []byte(contractCode), nil
+
+			default:
+				require.FailNow(t,
+					"import of unknown location: %s",
+					"location: %s",
+					location,
+				)
+				return nil, nil
+			}
+		},
+		HandleParserError: func(err analysis.ParsingCheckingError, _ *ast.Program) error {
+			return nil
+		},
+	}
+
+	programs, err := analysis.Load(config, contractLocation)
+	require.NoError(t, err)
+
+	var parserError *parser.Error
+	require.ErrorAs(t, programs[contractLocation].LoadError, &parserError)
+}
+
+func TestHandledCheckerError(t *testing.T) {
+
+	t.Parallel()
+
+	contractAddress := common.MustBytesToAddress([]byte{0x1})
+	contractLocation := common.AddressLocation{
+		Address: contractAddress,
+		Name:    "ContractA",
+	}
+	const contractCode = `
+	  access(all) contract ContractA {
+	    init() {
+	      X
+	    }
+	  }
+	`
+
+	config := &analysis.Config{
+		Mode: analysis.NeedTypes,
+		ResolveCode: func(
+			location common.Location,
+			importingLocation common.Location,
+			importRange ast.Range,
+		) ([]byte, error) {
+			switch location {
+			case contractLocation:
+				return []byte(contractCode), nil
+			default:
+				require.FailNow(t,
+					"import of unknown location: %s",
+					"location: %s",
+					location,
+				)
+				return nil, nil
+			}
+		},
+		HandleCheckerError: func(err analysis.ParsingCheckingError, _ *sema.Checker) error {
+			return nil
+		},
+	}
+
+	programs, err := analysis.Load(config, contractLocation)
+	require.NoError(t, err)
+
+	var checkerError *sema.CheckerError
+	require.ErrorAs(t, programs[contractLocation].LoadError, &checkerError)
+}
+
 // Tests that an error handled by the custom error handler is not returned
 // However, it must set LoadError to the handled error so that checkers later importing the program can see it
-func TestHandledCheckerErrorImportedProgram(t *testing.T) {
+func TestHandledLoadErrorImportedProgram(t *testing.T) {
 
 	t.Parallel()
 
@@ -318,9 +415,6 @@ func TestHandledCheckerErrorImportedProgram(t *testing.T) {
 
 	programs, err := analysis.Load(config, contract1Location)
 	require.NoError(t, err)
-
-	require.Error(t, programs[contract1Location].LoadError)
-	require.Error(t, programs[contract2Location].LoadError)
 
 	var checkerError *sema.CheckerError
 	require.ErrorAs(t, programs[contract1Location].LoadError, &checkerError)
--- a/tools/analysis/analysis_test.go
+++ b/tools/analysis/analysis_test.go
@@ -259,6 +259,74 @@ func TestCheckError(t *testing.T) {
 	require.ErrorAs(t, err, &checkerError)
 }
 
+// Tests that an error handled by the custom error handler is not returned
+// However, it must set LoadError to the handled error so that checkers later importing the program can see it
+func TestHandledCheckerErrorImportedProgram(t *testing.T) {
+
+	t.Parallel()
+
+	contract1Address := common.MustBytesToAddress([]byte{0x1})
+	contract1Location := common.AddressLocation{
+		Address: contract1Address,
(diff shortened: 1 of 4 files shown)


tools/analysis/programs.go
```diff
@@ -69,22 +69,53 @@ func (programs Programs) load(
 
 	program, err := parser.ParseProgram(nil, code, parser.Config{})
 	if err != nil {
-		return wrapError(err)
+		// If a parser error handler is set and the error is non-fatal
+		// use handler to handle the error (e.g. to report it and continue analysis)
+		// This will only be used for the entry point program (e.g not an imported program)
+		wrappedErr := wrapError(err)
+		if loadError == nil {
```
turbolent, line 76: Isn't this line always true? `loadError` was never assigned to before

tools/analysis/programs.go
```diff
@@ -69,22 +69,53 @@ func (programs Programs) load(
...
+			loadError = wrappedErr
+		}
+
+		if config.HandleParserError != nil && program != nil && importingLocation == nil {
```
turbolent, line 80: Why is the handler only called when a program exists and when the parsed program was not imported?

It might be surprising for a user of the package that the handler is not always called.

Maybe pass the details (program, importing location) to the handler, so it can determine / filter on its own, if needed. Or at least leave a comment similar to the HandleCheckerError call below.

tools/analysis/programs.go
```diff
@@ -69,22 +69,53 @@ func (programs Programs) load(
...
-			return wrapError(err)
+			wrappedErr := wrapError(err)
+			if loadError == nil {
+				loadError = wrappedErr
+			}
+
+			// If a custom error handler is set, and the error is non-fatal
+			// use handler to handle the error (e.g. to report it and continue analysis)
+			// This will only be used for the entry point program (e.g not an imported program)
+			if config.HandleCheckerError != nil && checker != nil && importingLocation == nil {
```
turbolent, line 102: Same as above: Good that there is a comment explaining here when the handler is called, but it might still be surprising to users of the package.

tools/analysis/programs.go
```diff
@@ -145,6 +176,14 @@ func (programs Programs) check(
 						return nil, err
 					}
 
+					// If the imported program has a load error, return it
+					// This may happen if a program is both imported and the entry point
+					// and has an error that was handled by a custom error handler
+					// However, we still want this error in the import resolution
+					if programs[importedLocation].loadError != nil {
+						return nil, programs[importedLocation].loadError
```
turbolent, line 184: Refactor the common expression, that is tested and then used, into a variable:

```suggestion
					loadError := programs[importedLocation].loadError
					if loadError != nil {
						return nil, loadError
```

turbolent (commented):
👍

tools/analysis/analysis_test.go
```diff
@@ -259,6 +259,168 @@ func TestCheckError(t *testing.T) {
...
+			importRange ast.Range,
+		) ([]byte, error) {
+			switch location {
+			case contractLocation:
+				return []byte(contractCode), nil
+
+			default:
+				require.FailNow(t,
+					"import of unknown location: %s",
+					"location: %s",
```
turbolent, line 293: `FailNow` doesn't format, so the message (second argument) is used as-is, the `%s` format specifier isn't used.

Use `FailNowf` instead

```suggestion
				require.FailNowf(t,
					"import of unknown location: %s",
```

tools/analysis/analysis_test.go
```diff
@@ -259,6 +259,168 @@ func TestCheckError(t *testing.T) {
...
+					location,
+				)
+				return nil, nil
+			}
+		},
+		HandleParserError: func(err analysis.ParsingCheckingError, _ *ast.Program) error {
+			return nil
```
turbolent, line 300: Here and below: Maybe assert the handler was called

tools/analysis/analysis_test.go
```diff
@@ -259,6 +259,168 @@ func TestCheckError(t *testing.T) {
...
+			importRange ast.Range,
+		) ([]byte, error) {
+			switch location {
+			case contractLocation:
+				return []byte(contractCode), nil
+			default:
+				require.FailNow(t,
+					"import of unknown location: %s",
+					"location: %s",
+					location,
```
turbolent, line 342:
```suggestion
				require.FailNowf(t,
					"import of unknown location: %s",
					location,
```

tools/analysis/analysis_test.go
```diff
@@ -259,6 +259,168 @@ func TestCheckError(t *testing.T) {
...
+		) ([]byte, error) {
+			switch location {
+			case contract1Location:
+				return []byte(contract1Code), nil
+			case contract2Location:
+				return []byte(contract2Code), nil
+			default:
+				require.FailNow(t,
+					"import of unknown location: %s",
+					"location: %s",
```
turbolent, line 405:
```suggestion
				require.FailNowf(t,
					"import of unknown location: %s",
```

tools/analysis/programs.go
```diff
@@ -145,12 +172,18 @@ func (programs Programs) check(
...
+
+					// If the imported program had an error while loading, record it
+					loadError = programs[importedLocation].LoadError
 				}
 
-				return sema.ElaborationImport{
+				return &sema.ElaborationImport{
 					Elaboration: elaboration,
-				}, nil
+				}, loadError
```
turbolent, line 186: There shouldn't be any need to return a pointer here.

Also, the common pattern is to either return a value, or an error, not both.

```suggestion
				}
				if loadError != nil {
					return nil, loadError
				}

				return sema.ElaborationImport{
					Elaboration: elaboration,
				}, nil
```

turbolent (commented):
👍

tools/analysis/analysis_test.go
jribbink: ok sounds good, I had just copied this from other tests so I have no switched them all over.  It looks like `FailNow`/`FailNowf` require both a failure message and a formatted message, so I've switched it to this.

`func require.FailNowf(t require.TestingT, failureMessage string, msg string, args ...interface{})`

tools/analysis/analysis_test.go
```diff
@@ -259,6 +256,190 @@ func TestCheckError(t *testing.T) {
...
+	var checkerError *sema.CheckerError
+	require.ErrorAs(t, programs[contract1Location].LoadError, &checkerError)
+	require.ErrorAs(t, programs[contract2Location].LoadError, &checkerError)
+
+	// Validate that parent checker receives the imported program error despite it being handled
+	var importedProgramErr *sema.ImportedProgramError
+	loadErr := programs[contract1Location].LoadError.(analysis.ParsingCheckingError)
+	unwrapedErr := loadErr.Unwrap().(*sema.CheckerError)
+	require.Len(t, unwrapedErr.ChildErrors(), 1)
+	require.ErrorAs(t, unwrapedErr.ChildErrors()[0], &importedProgramErr)
```
turbolent, line 440: `ErrorAs` does all this unwrapping, so all that is needed here is: 

```suggestion
	require.ErrorAs(t, programs[contract1Location].LoadError, &importedProgramErr)
```

tools/analysis/programs.go
```diff
@@ -145,7 +172,17 @@ func (programs Programs) check(
...
 					}
 
-					elaboration = programs[importedLocation].Checker.Elaboration
+					// If the imported program has a checker, use its elaboration for the import
+					if programs[importedLocation].Checker != nil {
+						elaboration = programs[importedLocation].Checker.Elaboration
+					}
+
+					// If the imported program had an error while loading, record it
+					loadError = programs[importedLocation].LoadError
```
turbolent, line 181: Refactor the common subexpressions into single variables:

```suggestion
					program := programs[importedLocation]
					checker := program.Checker

					if checker != nil {
						elaboration = checker.Elaboration
					}

					// If the imported program had an error while loading, record it
					loadError = program.LoadError
```

turbolent (commented):
Looks good now! Just need to lint (remove unused constant) and left a couple suggestions

turbolent (approved):
:+1:

turbolent:
@SupunS Can you please also have a look?

turbolent:
Merging this for now so we can get a release out

PR #3090, opened 2024-02-09
Unify the requirement for all interface types to have to occur in intersection types

turbolent:
## Description

https://discord.com/channels/613813861610684416/621847426201944074/1205259026737266768

- Unify the requirement for all interface types to have to occur in intersection types

  Only requiring resource interface types and struct interface types but not contract interface types to have to occur in intersection types is inconsistent and confusing

- Migrate interface types which do not yet occur in an intersection type into a new intersection type

- Remove the `InterfaceType` run-time type value (`Type`) constructor.

  For example, given a struct interface `SI`, it should not be possible to construct a static type that refers to `SI` outside of an interface type. The checker changes reject `Type<SI>()`, but `InterfaceType("SI")` should also be invalid. 

  Instead, the type can be constructed via `Type<{SI}>()` or `IntersectionType(["SI"])` 

______


SupunS (approved):
Makes a lot of sense!

SupunS:
I guess this would also require a state-migration, for static types.

turbolent:
Argh, you're right, I hadn't thought of that yet. I guess we'd need to add a rule to the static type migration to rewrite contract interface types `CI` to `{CI}`?

turbolent:
@SupunS I added the migration rule and also realized `IntersectionType` can no longer work, so I removed it. Could you please have another look?

migrations/statictypes/statictype_migration.go
```diff
@@ -246,16 +250,36 @@ func (m *StaticTypeMigration) maybeConvertStaticType(staticType interpreter.Stat
...
 
+		// Interface types need to be placed in intersection types
+		if _, ok := parentType.(*interpreter.IntersectionStaticType); !ok {
+			if convertedType == nil {
+				convertedType = interpreter.NewIntersectionStaticType(
+					nil, []*interpreter.InterfaceStaticType{
+						staticType,
+					},
+				)
+			} else if convertedInterfaceType, ok := convertedType.(*interpreter.InterfaceStaticType); ok {
```
SupunS, line 267: why does this second if-condition needed? is it to avoid any accidental double-wrapping of  `IntersectionStatic` types? (e.g: type-converter already returns the correctly wrapped type)

SupunS (commented):
Looks good! just one question/clarification

SupunS (approved):

migrations/statictypes/statictype_migration.go
turbolent: I added comments for the two cases in 2662be331fc51bf2efff26356fbc0c35fc015f93.

Do they answer your question?

SupunS: yes - thank you for the explanation!

I just realized, do we need to do a similar check for the returned value of `m.compositeTypeConverter(staticType)`? e.g: what if a composite type was converted to an interface type? Think we don't have such conversions yet, but just to be future proof?

turbolent: Good idea! We should maybe add that as well and add a test 👍

SupunS: > Think we don't have such conversions yet

actually this is needed when refactoring type-requirements (composite type would become interfaces). e.g: `@FungibleToken.Vault` -->  `@{FungibleToken.Vault}`

SupunS: nice!

turbolent: @SupunS Added in fec43976e36dcbe632077ed981046821acaa647b. Also added tests in general for this intersection type wrapping for both composite and interface type converters

turbolent:
@SupunS Could you please have another look?

SupunS (approved):
Nice! Thanks for the additional tests 👌

PR #3093, opened 2024-02-10
Consider inherited interfaces when checking for member clashes in intersection types

### LINKED ISSUE #3091: CheckIntersectionType: also include interface conformances' members
With interface conformance implemented, the intersection type checking must also consider inherited interfaces.

See the TODO:  https://github.com/onflow/cadence/blob/7fb04f9da0714407dafb7a2aa1ccbf05f541fb3f/runtime/sema/checker.go#L940-L941

turbolent:
Closes #3091

## Description

- Fix order of declaring nested types' explicit interface conformances: setting the explicit interface conformances must occur before checking members
- Consider inherited interfaces when checking for member clashes in intersection types

______

### DIFF
--- a/runtime/sema/checker.go
+++ b/runtime/sema/checker.go
@@ -937,40 +937,45 @@ func CheckIntersectionType(
 
 		// The intersections may not have clashing members
 
-		// TODO: also include interface conformances' members
-		//   once interfaces can have conformances
+		checkClashingMember := func(interfaceType *InterfaceType) {
+			interfaceType.Members.Foreach(func(name string, member *Member) {
 
-		interfaceType.Members.Foreach(func(name string, member *Member) {
-			if previousDeclaringInterfaceType, ok := memberSet[name]; ok {
+				if previousDeclaringInterfaceType, ok := memberSet[name]; ok {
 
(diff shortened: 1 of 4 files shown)


runtime/sema/checker.go
```diff
@@ -937,40 +937,45 @@ func CheckIntersectionType(
 
 		// The intersections may not have clashing members
 
-		// TODO: also include interface conformances' members
-		//   once interfaces can have conformances
+		checkClashingMember := func(interfaceType *InterfaceType) {
```
SupunS, line 940: How does this deal with the delta problem? e.g: having `{A, B}` when `interface A: C {}` and `interface B: C {}` and `C` has a member `foo`.

turbolent: Good point, it needs to handle that as well

turbolent: The function already handled it. It does not matter if the member is inherited or not, all that is checked is that the all members of all interfaces have the same type.

Added a test in cc6258f5267be7344341b791ca60d9bd6ee00554

SupunS (approved):
Nice!

PR #3110, opened 2024-02-15
Improve migration

turbolent:
## Description

- Improve unreachable cases by panicing with more descriptive errors
- Handle `nil` type in `TypeValue` and `PathCapabilityValue`

______

### DIFF
--- a/runtime/sema/type.go
+++ b/runtime/sema/type.go
@@ -7987,6 +7987,10 @@ type IntersectionType struct {
 var _ Type = &IntersectionType{}
 
 func NewIntersectionType(memoryGauge common.MemoryGauge, types []*InterfaceType) *IntersectionType {
+	if len(types) == 0 {
+		panic(errors.NewUnreachableError())
+	}
+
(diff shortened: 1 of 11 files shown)


runtime/sema/type.go
```diff
@@ -7987,10 +7987,6 @@ type IntersectionType struct {
 var _ Type = &IntersectionType{}
 
 func NewIntersectionType(memoryGauge common.MemoryGauge, types []*InterfaceType) *IntersectionType {
-	if len(types) == 0 {
-		panic(errors.NewUnreachableError())
-	}
```
turbolent, line 7992: @onflow/cadence was there a particular reason why we were forbidding empty intersection types?

They are not useful, but I cannot think of a reason why they should not be allowed.

They might be a result of migrating a restricted type without any restrictions.

turbolent: I guess we cannot properly determine resource-kindedness for an empty intersection?

SupunS (approved):

PR #3111, opened 2024-02-15
Improve Cadence-1.0 contract update checker

SupunS:
Work towards of https://github.com/onflow/cadence/issues/2865

Depends on https://github.com/onflow/cadence/pull/3069

## Description

- Requires `AnyStruct{I}` / `AnyResource{I}` to be changed to `{I}`, instead of dropping the `{I}`.
- Requires converting `AnyStruct{}`/`AnyResource{}` (empty set) to `AnyStruct`/`AnyResource`.
- Allow removing events
- Allow changing type-requirements to interfaces. i.e: struct/resource to struct-interface/resource-interface, if they are residing inside a contract interface.

______


runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
```diff
@@ -271,6 +283,19 @@ func (validator *CadenceV042ToV1ContractUpdateValidator) checkTypeUpgradability(
...
+
+		// If the old restricted type is for AnyStruct/AnyResource,
+		// require them to drop the "restricted type".
+		// e.g: `T{I} -> {I}`
+		if restrictedNominalType, isNominal := oldType.LegacyRestrictedType.(*ast.NominalType); isNominal {
+			switch restrictedNominalType.Identifier.Identifier {
+			case "AnyStruct", "AnyResource":
+				break typeSwitch
+			}
+		}
```
turbolent, line 295: Does the code already handle the case where the restricted type is inferred, i.e. `{SI}` is given and it means `AnyStruct{SI}`?

turbolent (commented):
👍

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
SupunS: yeah it should, since the validator works directly on the syntax (and no inferring is involved at this point). I'll add a test just to be sure.

turbolent:
Could you please open an issue for adjusting the state migration in terms of the new `Any*{...}` to `{...}` rule, and add it to the milestone?

turbolent:
@SupunS could you please also add the second rule of #3112 to this PR?

SupunS:
Added in [74b1010](https://github.com/onflow/cadence/pull/3111/commits/74b1010905ea336421dc30dbe455ecde58d6e0c7)

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
SupunS: adding a test for this isn't much useful; both the old-code and the new-code look the same.
e.g: Old code, with inferred-type is
```cadence
access(all) var a: @{I} 
```
And the required update, is also the same (keeping the restrictions only which is also ` @{I} `).

bluesign:
@SupunS Does this cover field type change too? (Ex: var vault: @fungibletoken.vault to var vault: @{fungibletoken.vault} )

SupunS:
@bluesign No, not in this PR

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
```diff
@@ -271,6 +283,24 @@ func (validator *CadenceV042ToV1ContractUpdateValidator) checkTypeUpgradability(
...
+			case "AnyStruct", "AnyResource":
+				if len(oldType.Types) > 0 {
+					break typeSwitch
+				}
+			}
+		}
+
+		// Otherwise require them to drop the "restrictions".
+		// e.g-1: `T{I} -> T`
+		// e.g-2: `AnyStruct{} -> AnyStruct`
```
turbolent, line 303: How is it ensured this is only done in the case of an "owned" type / non-reference type (the parent is not a reference type)?

SupunS: yeah, this one does not; Need to update it to the match the changes we discussed for reference type.

SupunS: Actually, it is not needed for the contract update checker,  rather only needed for the static type migration, because entitlement migration is relying on the restriction set. Entitlement migration would anyway drop the restrictions after converting them to the corresponding entitlement.

So in the two cases, conversions would be:
- Reference type: `&T{I}` -> `auth(E) &T`
- Owned value: `T{I}` -> `T`

Both in which the restrictions are dropped.

SupunS: Then there is a question, what should `&AnyStruct{I}` be converted to? is it `auth(E) &AnyStruct` or `auth(E) &{I}`

The update checker currently forces it to be `auth(E) &{I}`. Do you know how the storage migration handles it?

turbolent: > Actually, it is not needed for the contract update checker, rather only needed for the static type migration, because entitlement migration is relying on the restriction set. 

Ah, right! The rules for rewriting intersection types I had written up in the Notion doc were specifically for the static type migration, considering that the entitlements migration would run after, requiring enough information to be preserved. 

But really, in the end the state migration (static type + entitlements migration) and the contract update checker should have rules for the same outcome at the end.

turbolent: > Then there is a question, what should `&AnyStruct{I}` be converted to? is it `auth(E) &AnyStruct` or `auth(E) &{I}`. The update checker currently forces it to be `auth(E) &{I}`.

I'll see what the entitlements migration currently produces. I guess `AnyStruct{I} -> {I}` should apply, i.e. `&AnyStruct{I}` should be rewritten to `auth(E) &{I}` (instead of just `auth(E) &AnyStruct`), but IDK if that's currently the case

turbolent: I'll fix this in #3118

turbolent:
> Allow changing type-requirements to interfaces. i.e: struct/resource to struct-interface/resource-interface, if they are residing inside a contract interface.

Is it safe to allow this in general? 

For example, given the previous type requirement declaration

```cadence
contract interface CI {
    resource R {}
}
```

existing types might be of the form `@CI.R`. 

If we allow the change of the type declaration to 

```cadence
contract interface CI {
    resource interface R {}
}
```

then all such types `@CI.R` will have to be rewritten to `@{CI.R}`

SupunS:
> then all such types @CI.R will have to be rewritten to @{CI.R}

The static type migration already does that, correct? i.e: wraps with an intersection type if the interface is not already residing within an intersection type.
https://github.com/onflow/cadence/blob/e60051d42a214d2eb15281b19d27e80b1923eefd/migrations/statictypes/statictype_migration.go#L288-L290

turbolent:
Ah, I forgot that we added this general migration already for https://github.com/onflow/cadence/pull/3090, great!

turbolent (approved):
Great work!

PR #3118, opened 2024-02-17
Rewrite owned intersection types in migration

### LINKED ISSUE #3112: Improve migration of intersection types
Update the static type migration to convert _owned_ intersection types

- `T{}` / `Any*{}` to `T/Any*`

  If the intersection type has no interface types, then return the legacy type as-is.

  This prevents the migration from creating an intersection type with no interface types, as static to sema type conversion ignores the legacy type.

- `Any*{A,...}`  -> `{A,...}`

  If the intersection type has no or an AnyStruct/AnyResource legacy type, and has at least one interface type, then return the intersection type without the legacy type.

- `T{A,...}` -> `T`

  If the intersection type has a legacy type, and has at least one interface type, then return the legacy type as-is.

turbolent:
Closes #3112

______


migrations/statictypes/statictype_migration_test.go
```diff
@@ -149,4 +149,186 @@ func TestStaticTypeMigration(t *testing.T) {
 			actual,
 		)
 	})
+
+	t.Run("T{I} -> T", func(t *testing.T) {
```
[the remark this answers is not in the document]
turbolent: Yeah, I'm pretty sure I had added some, but jumping around branches I must have lost them :-( 

Will re-add

migrations/statictypes/statictype_migration.go
```diff
@@ -361,6 +405,62 @@ func (m *StaticTypeMigration) maybeConvertStaticType(staticType, parentType inte
...
 
+func rewriteLegacyIntersectionType(
+	intersectionType *interpreter.IntersectionStaticType,
+	parentType interpreter.StaticType,
+) interpreter.StaticType {
+
+	// Rewrite the intersection type,
+	// if it does not appear in a reference type.
+
+	if _, ok := parentType.(*interpreter.ReferenceStaticType); ok {
```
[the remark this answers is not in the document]
turbolent: Why do we have to handle optionals specially here? Do you have an example?

migrations/statictypes/statictype_migration_test.go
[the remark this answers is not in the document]
turbolent: Added back in [b19c03b](https://github.com/onflow/cadence/pull/3118/commits/b19c03b3fa0897122d7e7b2e4db270407b263d1c)

migrations/statictypes/statictype_migration.go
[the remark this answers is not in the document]
SupunS: like for example, an intersection/restricted type appearing in an optional-reference. i.e: `&T{I}?`
IINM, pre-1.0 could produce references to optionals values, so the types could also have references to optional types?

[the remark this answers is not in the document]
turbolent: Opened https://github.com/onflow/cadence/issues/3122 to handle references to optionals

SupunS (approved):

turbolent:
I tried to add the rules to the entitlements migration in [a2232d9](https://github.com/onflow/cadence/pull/3118/commits/a2232d9b4a717e94977d9d084f655319b1280ceb).

I refactored it and fixed some bugs and gaps along the way. I got the rules working by re-using `RewriteLegacyIntersectionType` from the static type migration, calling it after the entitlement addition is complete. 

However, I noticed there is one crucial exception: https://github.com/onflow/cadence/pull/3118/files#diff-8fed0196f6dd9c6e80c1943403eff29f29b7f4651dded1b7df9ffa1c6449e9afR892-R893: For the particular case `&R{}`, the result should be _unentitled_, because the restriction was empty. 

This is currently difficult to implement, because the static type to sema type conversion looses the information that is necessary. I'll keep fixing this last issue tomorrow

turbolent:
Improved the entitlements migration further in [c10a6eb](https://github.com/onflow/cadence/pull/3118/commits/c10a6eb94043ae99c335e23e724daedc64f8a6f2), unifying the authorization/entitlements calculation for static intersection types and in general for other entitlements-supporting types, and refactoring the migration to be mainly based on static types, only converting to sema types if needed. Before, the migration always converted static to sema, added entitlements, and converted back from sema to static, for the whole type.

This also fixes the last edge case, `&R{} --> &R`.

migrations/migration.go
```diff
@@ -252,18 +252,23 @@ func (m *StorageMigration) MigrateNestedValue(
...
 				keyValuePair{
 					key:   key,
 					value: value,
 				},
 			)
-
-			// continue iteration
-			return true
-		})
+		}
```
turbolent, line 271: I changed this, because the migration of a dictionary with a static type like `T{}` results in a panic, because the dictionary iteration function `Iterate` converts the dictionary's static type to a sema type (`IsResourceKinded`), which results in the invalid  empty intersection type `{}`.

As  dictionary iteration/migration thus no longer loads the dictionary type, `TestMigrationPanics/composite_dictionary` is therefore no longer valid/needed, and I could remove it in 8b2476c0127b5891b0c3174102ec5899331a5b2b.

migrations/migration_test.go
```diff
@@ -948,148 +947,3 @@ func TestContractMigration(t *testing.T) {
 		value,
 	)
 }
-
-func TestMigrationPanics(t *testing.T) {
-
-	t.Parallel()
-
-	t.Run("composite dictionary", func(t *testing.T) {
```
SupunS, line 956: Is the behaviour same for iterating arrays as well (i.e: array of composites)? If not maybe update this test to use an array instead of an dictionary

migrations/migration.go
SupunS: sorry a bit confused: shouldn't the migration produce a valid intersection type? or does it happens after the iteration?

migrations/migration_test.go
SupunS: nevermind, types will not be loaded even for arrays, after the changes in: https://github.com/onflow/cadence/pull/3126.

migrations/migration.go
turbolent: I was also confused at first 😅 Yes, the migration generates a new type for the dictionary. However, the dictionary value still has the old type while it is getting iterated over.

SupunS: aha makes sense, thanks for the explanation

SupunS (approved):

migrations/migration_test.go
turbolent: The migration of arrays was already not loading the type, as it is manually iterating using `Count()` + for-loop with `Get()` calls

Issue #2961, opened 2023-12-01
Reject container mutation while iterating


SupunS:
### Issue to be solved

Mutating a resource-typed container while iterating it is already prohibited at runtime.

It would be good to do the same for non-resource type containers as well, because the current behavior is a bit surprising / not intuitive (though it is safe). More context https://github.com/dapperlabs/cadence-internal/pull/142

### Suggested Solution

Once added, tests added in https://github.com/onflow/cadence/pull/2960 should fail at runtime.

PR #3126, opened 2024-02-22
Prevent container mutation while iterating

SupunS:
Closes https://github.com/onflow/cadence/issues/2961

## Description

NOTE: 
1. Hide whitespace changes when reviewing the PR/diff. There are lot of indentation changes.
2. Many of the changes are for passing the `locationRange` for functions where necessary. I've annotated the important changes that is being added as the main objective of this PR.

______


runtime/interpreter/value.go
```diff
@@ -18514,11 +18572,8 @@ func (v *DictionaryValue) iterateKeys(
...
 		}
 	}
-	if v.IsResourceKinded(interpreter) {
-		interpreter.withMutationPrevention(v.StorageID(), iterate)
-	} else {
-		iterate()
-	}
+
+	interpreter.withMutationPrevention(v.StorageID(), iterate)
 }
```
SupunS, line 18577: call `interpreter.withMutationPrevention` for all values (both resource-typed and non-resource typed values)

runtime/interpreter/value.go
```diff
@@ -1868,18 +1898,19 @@ func (v *ArrayValue) iterate(
...
-		walkChild(element)
-		return true
-	})
+	v.Iterate(
+		interpreter,
+		func(element Value) (resume bool) {
+			walkChild(element)
+			return true
+		},
+		// TODO: Not supposed to panic with container mutation error.
```
SupunS, line 1911: Adding locationRange to these two methods explodes in the code, so I left it to make the changes minimal.
Both here and below, the two functions are not called from user code. Instead these are being used internally for other validations. Hence are not supposed to panic with container mutation errors.

SupunS:
> To check my understanding, this just extends the existing logic for resource containers to struct containers as well?

Yes, precisely.

runtime/interpreter/value.go
turbolent: Yeah, let's do that in a separate PR

runtime/interpreter/value.go
```diff
@@ -1841,25 +1841,55 @@ func (v *ArrayValue) Accept(interpreter *Interpreter, visitor Visitor) {
...
+	locationRange LocationRange,
 ) {
 	iterate := func() {
 		err := atreeIterate(func(element atree.Value) (resume bool, err error) {
 			// atree.Array iteration provides low-level atree.Value,
 			// convert to high-level interpreter.Value
+			elementValue := MustConvertStoredValue(interpreter, element)
+
+			// Each element must be transferred before passing onto the function.
+			transferredElement := elementValue.Transfer(
```
turbolent, line 1883: Are we sure that "lowering" the transfer from the evaluation of `for` loops into the `iterate` function does not somehow affect the semantics of other uses of the `iterate` function (and its callers)?

runtime/interpreter/value.go
```diff
@@ -18187,7 +18235,17 @@ func (v *CompositeValue) ForEach(
...
-		if !function(value) {
+		// Each element must be transferred before passing onto the function.
+		transferredValue := value.Transfer(
+			interpreter,
+			locationRange,
+			atree.Address{},
+			false,
+			nil,
+			nil,
+		)
```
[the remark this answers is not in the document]
turbolent: Same as a above: I guess previously the value passed to the function was not (always) transferred? Does this change potentially negatively affect the behaviour of callers of the function (`CompositeValue.ForEach`)?

turbolent (commented):
Nice!

Changes look good, mostly just concerned about impact on "unrelated" code

SupunS:
Made the transfer optional and only enabled it for the `for` loop: https://github.com/onflow/cadence/pull/3126/commits/797b1255c21c47e84f82bccbc30e3d8ddf06d1be

turbolent (approved):
Nice!

runtime/tests/interpreter/memory_metering_test.go
```diff
@@ -118,10 +118,10 @@ func TestInterpretArrayMetering(t *testing.T) {
 		_, err := inter.Invoke("main")
 		require.NoError(t, err)
 
-		assert.Equal(t, uint64(30), meter.getMemory(common.MemoryKindArrayValueBase))
-		assert.Equal(t, uint64(24), meter.getMemory(common.MemoryKindAtreeArrayDataSlab))
-		assert.Equal(t, uint64(3), meter.getMemory(common.MemoryKindAtreeArrayMetaDataSlab))
-		assert.Equal(t, uint64(9), meter.getMemory(common.MemoryKindAtreeArrayElementOverhead))
+		assert.Equal(t, uint64(26), meter.getMemory(common.MemoryKindArrayValueBase))
```
turbolent, line 121: Why did the metering change here? It looks like we have fewer arrays/data slabs now

SupunS: Because the outer array is not moved/transferred at the loop expression (i.e: in `for value in values`)

Issue #2541, opened 2023-06-06
Cannot take a reference to a nested-optional


SupunS:
### Issue to be solved

Taking a reference to a nested-optional value (double-optional) results in type-checking errors.

```cadence
pub fun main() {
  var dict: {String: Foo?} = {}

  // Type of `dict["foo"]` is `Foo??`

  var ref = &dict["foo"] as &Foo??    // Static Error: cannot create reference. expected reference type, got `&Foo?`
}

pub struct Foo {}
```

This is solved to a certain extent in Stable-Cadence, with the parenthesized types added in https://github.com/onflow/cadence/pull/2140

i.e:
```cadence
  var ref = &dict["foo"] as &(Foo??)    // OK
```

However, the created reference is a reference to an optional (i.e: `&(Foo??)`), but not an optional reference (i.e: `(&Foo)??`) as proposed in [FLIP 722](https://github.com/onflow/flips/blob/main/cadence/20211208-cadence-remove-optional-reference.md)

### Suggested Solution

The issue is, currently the optional unwrapping happens only for one level. 
It would need to recursively unwrap optionals during reference creation.

j1010001:
@SupunS to check if this would require breaking change. There is a workaround.

SupunS:
> i.e:
>
> ```cadence
> var ref = &dict["foo"] as &(Foo??)    // OK
> ```
> However, the created reference is a reference to an optional (i.e: &(Foo??)), but not an optional reference (i.e: (&Foo)??) as > proposed in [FLIP 722](https://github.com/onflow/flips/blob/main/cadence/20211208-cadence-remove-optional-reference.md)

This is a breaking change, since the resulting type of the expression would be different after the change

PR #3132, opened 2024-02-26
Allow creating references to nested optionals

SupunS:
Closes https://github.com/onflow/cadence/issues/2541

This PR added two fixes:
- Allow creating nested-optional references (`&T)??`.
- Reject taking references as `&(T?)` 

______

### DIFF
--- a/runtime/interpreter/interpreter_expression.go
+++ b/runtime/interpreter/interpreter_expression.go
@@ -1335,24 +1335,14 @@ func (interpreter *Interpreter) VisitReferenceExpression(referenceExpression *as
 
 	result := interpreter.evalExpression(referenceExpression.Expression)
 
-	makeReference := func(value Value, typ *sema.ReferenceType) *EphemeralReferenceValue {
-		// if we are currently interpretering a function that was declared with mapped entitlement access, any appearances
-		// of that mapped access in the body of the function should be replaced with the computed output of the map
-		auth := interpreter.getEffectiveAuthorization(typ)
+	return interpreter.createReference(borrowType, result, referenceExpression)
+}
 
-		locationRange := LocationRange{
-			Location:    interpreter.Location,
-			HasPosition: referenceExpression,
-		}
-
-		return NewEphemeralReferenceValue(
-			interpreter,
-			auth,
-			value,
-			typ.Type,
-			locationRange,
-		)
-	}
+func (interpreter *Interpreter) createReference(
+	borrowType sema.Type,
+	value Value,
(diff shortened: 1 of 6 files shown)


runtime/interpreter/interpreter_expression.go
```diff
@@ -1335,24 +1335,14 @@ func (interpreter *Interpreter) VisitReferenceExpression(referenceExpression *as
...
 
-		return NewEphemeralReferenceValue(
-			interpreter,
-			auth,
-			value,
-			typ.Type,
-			locationRange,
-		)
-	}
+func (interpreter *Interpreter) createReference(
```
SupunS, line 1341: This is the same old logic, but made it recursive.

runtime/tests/interpreter/interpreter_test.go
```diff
@@ -11514,53 +11514,6 @@ func TestInterpretArrayToConstantSized(t *testing.T) {
 	})
 }
 
-func TestInterpretOptionalReference(t *testing.T) {
```
SupunS, line 11517: moved to `reference_test.go`

runtime/tests/checker/reference_test.go
```diff
@@ -3804,3 +3804,37 @@ func TestCheckReferenceRequiredTypeAnnotation(t *testing.T) {
...
+		require.NoError(t, err)
+	})
+
+	t.Run("reference to nested optional", func(t *testing.T) {
+		t.Parallel()
+
+		_, err := ParseAndCheck(t, `
+            fun main() {
+                var dict: {String: Foo?} = {}
+                var ref: &(Foo??) = &dict["foo"] as &(Foo??)
```
dsainati1, line 3832: Did we not previously ban references to optionals, and change them to be optional references? IMO the same logic should apply here; taking a reference to a `T??` type should produce a `(&T)??` type.

SupunS: > Did we not previously ban references to optionals, and change them to be optional references?

It was supposed to, but this path wasn't handled. This is actually the second part of the issue: https://github.com/onflow/cadence/issues/2541#issuecomment-1879211932. That one still need to be handled. I will open a follow-up PR.

SupunS: Added the second pat to the same PR, as it turns out to be a simple fix: https://github.com/onflow/cadence/pull/3132/commits/25b0726314c4cd02bd100ae7d41cffa36711f3d3

turbolent (approved):
Nice!

Issue #3096, opened 2024-02-12
Run full Cadence 1.0 migration on TN state


turbolent:
```[tasklist]
### Tasks
- [x] Cadence 1.0 migration: Errors when worker count >1
- [x] Transfer, called by remove, insert, etc., loads types, which might be unmigrated/unsupported: empty intersection types are invalid: https://github.com/onflow/cadence/pull/3138
- [x] Static types of containers need to get migrated too: https://github.com/onflow/cadence/pull/3141
- [x] Ensure nested containers are properly (recursively) migrated: https://github.com/onflow/cadence/pull/3141 and https://github.com/onflow/cadence/pull/3142
- [ ] https://github.com/onflow/flow-go/issues/5485
- [ ] https://github.com/onflow/cadence/issues/3153
- [ ] https://github.com/onflow/cadence/issues/3149
- [x] Clean up TN service account: Prune random beacon history
- [ ] https://github.com/onflow/cadence/pull/3158
- [ ] https://github.com/onflow/cadence/pull/3157
```

turbolent:
For investigating the concurrency issues, I tried building the `util` binary with the race detector on, but it reported no issues:

```sh
CGO_ENABLED=1 \
GOOS=linux \
GOARCH=amd64 \
CC="zig cc -target x86_64-linux-musl" \
CXX="zig c++ -target x86_64-linux-musl" \
go build -v --tags "netgo,osusergo" -race -ldflags "-extldflags=-static"
```

turbolent:
For the empty intersection type loading issue, I opened #3138, which so far has a reproducer for one particular case where the type loading occurs (`DictionaryValue.Insert` in `StorageMigration.MigrateNestedValue`).

I'm not really sure how to proceed here. Disabling type checks, like `checkContainerMutation` isn't sufficient, because e.g. `CompositeValue.Transfer` calls `CompositeValue.IsResourceKinded`, which isn't just a run-time check, but required to distinguish between the struct vs resource code path of transferring (copy vs move).

The best idea I came up so far is rewriting the static type of a container value (e.g. dictionary, array), before migrating the nested value inside of it, i.e. before calling `Remove`/`Insert`, and before the related `Transfer` calls occur, which convert the static type.

turbolent:
I wonder if we can integrate the intersection type rewriting and entitlements migration code directly into `interpreter.ConvertStaticToSemaType`. 

[...]

PR #3144, opened 2024-02-29
Improve migrations

turbolent:
Work towards #3096 and #3149

## Description

Great suggestion by @fxamacker: Perform a atree storage health check after each migration, to ensure individual migrations produce valid outcomes.

This currently fails for some tests:
- `statictypes.TestMigratingNestedContainers/nested_dictionary`:
  ```
  internal error: slabs not referenced from account Storage: [0x4200000000000000.4]
  ```

- `entitlements.TestMigrateSimpleContract`:

  ```
  parent and child are not owned by the same account: child.owner ��������, parent.owner B�������
  ```

### Fixes, improvements, and additions to tests

- Add atree storage health checks to tests after committing migrations - but only if there were no errors
- Enable atree value validation where possible. It is not always possible, because e.g legacy values / static types and dummy static types cause a (correct) mismatch 
- Enable atree storage health checking during tests where possible. It is not always possible, because during the migration storage might be in an invalid state, e.g. when moving a value from an old value to a new one 
- Fix: test values must be transferred to account before adding them to a storage map
- Add more tests for static type migration of nested containers: In addition to dictionaries, also test nested arrays
- Add tests for general storage migration of nested containers: Test nested dictionaries, arrays, and composites. This is similar to nested container tests in static type migration, but also tests composites, which are not migrated by the static type migration
- Fix incorrect locations and qualified identifiers of test types. They did not match the type IDs
- Refactor the static type migration tests, so each test value is independently written, migrated, and asserted. This makes pin-pointing issues easier

### Fixes for migration

The health checks identified several issues, such as missing data or unreferenced data (garbage).

Fix this in general by changing the migrations so that they properly remove old child values from old containers, without transferring them to the stack, and inserting them into the (potentially new) target container, again without transferring them.

The assumption of the container operation functions, e.g. `DictionaryValue.Insert` or `ArrayValue.Remove`, is that the input is on the stack, and the result (if any) is also returned on the stack, even if the container itself is in a non-zero address account.

During the migration, all values are always in the account that is currently migrated, and values are never on the stack.

To properly move child values between or inside containers, add new function `*WithoutTransfer` functions to `DictionaryValue`, `ArrayValue`, and `CompositeValue`, e.g. `DictionaryValue.InsertWithoutTransfer`, and use them in the migration. The existing functions for performing operations on containers, like `DictionaryValue.Insert` still perform the same: they transfer the value from the stack to the container and call into the non-transfer variant of the function (e.g. for insertion), or call into the non-transfer function and transfer the result from the account to the stack.

Finally, introduce `NewWithType` functions for dictionaries and arrays, and use them in the static type and entitlements migration. They create a new container, with the given type, and move all children to the new container. This is currently needed for these migrations, as it is not (yet) possible to change the type of containers.

Once atree exposes functions to change the type of containers directly, we can replace these calls and remove the `NewWithType` again.

Best viewed without whitespace changes.

______

### DIFF
--- a/migrations/migration.go
+++ b/migrations/migration.go
@@ -303,23 +303,6 @@ func (m *StorageMigration) MigrateNestedValue(
 			if newKey == nil {
 				keyToSet = existingKey
 			} else {
-				// Key was migrated.
-				// Remove the old value at the old key.
-				// This old value will be inserted again with the new key, unless the value is also migrated.
-				existingKey = legacyKey(existingKey)
-				oldValue := dictionary.Remove(
-					m.interpreter,
-					emptyLocationRange,
-					existingKey,
-				)
-
-				if _, ok := oldValue.(*interpreter.SomeValue); !ok {
-					panic(errors.NewUnexpectedError(
-						"failed to remove old value for migrated key: %s",
-						existingKey,
-					))
-				}
-
 				keyToSet = newKey
 			}
 
@@ -330,7 +313,25 @@ func (m *StorageMigration) MigrateNestedValue(
 				valueToSet = newValue
 			}
 
-			dictionary.Insert(
+			existingKey = legacyKey(existingKey)
+			existingKeyStorable, existingValueStorable := dictionary.RemoveWithoutTransfer(
+				m.interpreter,
+				emptyLocationRange,
+				existingKey,
+			)
+
+			if existingKeyStorable == nil {
+				panic(errors.NewUnexpectedError(
+					"failed to remove old value for migrated key: %s",
+					existingKey,
+				))
+			}
+
+			interpreter.StoredValue(m.interpreter, existingValueStorable, m.storage).
+				DeepRemove(m.interpreter)
+			m.interpreter.RemoveReferencedSlab(existingValueStorable)
+
+			dictionary.InsertWithoutTransfer(
 				m.interpreter,
 				emptyLocationRange,
 				keyToSet,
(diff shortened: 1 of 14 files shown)


migrations/migration.go
```diff
@@ -303,34 +318,35 @@ func (m *StorageMigration) MigrateNestedValue(
...
 
 			if newValue == nil {
 				valueToSet = existingValue
 			} else {
 				// Value was migrated
 				valueToSet = newValue
+
+				interpreter.StoredValue(m.interpreter, existingValueStorable, m.storage).
+					DeepRemove(m.interpreter)
+				m.interpreter.RemoveReferencedSlab(existingValueStorable)
```
SupunS, line 346: Shouldn't this be done despite the 'value' was migrated or not? i.e: key could be migrated and the old value may now needs to be inserted at a different key?

SupunS (commented):
Looks good!

migrations/migration.go
turbolent: We should only delete the data for the value, if there is a new value. If we are re-inserting the value for a different key, then removing the data for the value would loose the data.

Note also the change above: We're now _always_ removing the value from the container (but that does not imply deletion), i.e. after the old key-value pair is no longer in the container, we can then either:
- Insert the existing value under the new key. We shouldn't delete the value 
- Insert the new value under the (old or new) key. We need to delete the old value

You can try making the change, and I tried so myself yesterday: If the old value is always deleted, then atree will report the missing data (slab).

SupunS (approved):

migrations/migration.go
SupunS: Thanks for the explanation!

PR #3154, opened 2024-03-06
Improve errors in state migration

turbolent:
Work towards #3096 

## Description

Introduce a new error type, `MigrationError`, which captures migration location information.

Include a stack trace when reporting it for recovered errors

______

### DIFF
--- a/migrations/capcons/migration_test.go
+++ b/migrations/capcons/migration_test.go
@@ -70,16 +70,9 @@ type testMigration struct {
 	migration     string
 }
 
-type testMigrationError struct {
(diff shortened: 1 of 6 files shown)


SupunS (approved):

turbolent:
@SupunS Thank you for the quick review! Changed my mind in the approach and extended this a bit. I think the new approach will be better, as it will allow us to debug thanks to the stack trace, but also it will not stop the whole migration just when a Go runtime error is encountered.

SupunS (approved):

Issue #2826, opened 2023-09-28
[stable-cadence] Improve type inference for authorized references


SupunS:
### Issue to be solved

When there are auth references with a common entitlement, the type-inferring infers the supertype type as `AnyStruct`.

```cadence
access(all) 
fun main(
    a: auth(Insert, Mutate) &[String],
    b: auth(Remove, Mutate) &[String]
) {

    // OK: both `a` and `b` have `Mutate` entitlement. 
    let c: [auth(Mutate) &[String]] = [a, b]

    // However, when inferring, type of `d` is inferred as `[AnyStruct]`
    let d = [a, b]

    // Static Error
    let e: [auth(Mutate) &[String]] = d
}
```

### Suggested Solution

It could be improved to consider the common entitlement, and to infer the supertype as `auth(Mutate) &[String]`

j1010001:
nice to have for Stable Cadence

PR #3156, opened 2024-03-07
Improve type inference for authorized references

dsainati1:
Closes https://github.com/onflow/cadence/issues/2826

Adds better computing of the least common supertype of authorized reference types

______

### DIFF
--- a/runtime/sema/type_tags.go
+++ b/runtime/sema/type_tags.go
@@ -733,20 +733,39 @@ func leastCommonAccess(accessA, accessB Access) Access {
 		switch setAccessB.SetKind {
 		case Conjunction:
 			// least common access of two non-disjoint conjunctions is their intersection
-			// e.g. the least common access of (E, F) and (E, G)  is just E
+			// e.g. the least common supertype of (E, F) and (E, G)  is just E
 			intersection := orderedmap.KeySetIntersection(setAccessA.Entitlements, setAccessB.Entitlements)
 			if intersection.Len() != 0 {
 				return NewAccessFromEntitlementSet(intersection, Conjunction)
 			}
 			// if the intersection is completely empty (i.e. the two sets are totally disjoint)
-			// the least common access is the union of the sets converted to a disjunction
-			// e.g. the least common access of E and F is `(E | F)`
-			// and the least common access of `(A, B)` and `(C, D)` is `(A | B | C | D)`
+			// the least common supertype is the union of the sets converted to a disjunction
+			// e.g. the least common supertype of E and F is `(E | F)`
+			// and the least common supertype of `(A, B)` and `(C, D)` is `(A | B | C | D)`
 			union := orderedmap.KeySetUnion(setAccessA.Entitlements, setAccessB.Entitlements)
 			return NewAccessFromEntitlementSet(union, Disjunction)
 
 		case Disjunction:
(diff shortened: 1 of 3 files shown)


runtime/sema/type_tags.go
```diff
@@ -719,6 +720,132 @@ func findSuperTypeFromUpperMask(joinedTypeTag TypeTag, types []Type) Type {
...
+	for _, typ := range types {
+		// 'Never' type doesn't affect the supertype.
+		// Hence, ignore them
+		if typ == NeverType {
+			continue
+		}
+
+		referenceType, ok := typ.(*ReferenceType)
+		if !ok {
+			return getSuperTypeOfDerivedTypes(types)
```
SupunS, line 806: Can maybe directly use `commonSuperTypeOfHeterogeneousTypes` since we know the two types are not the same 'kind'?

runtime/sema/type_tags.go
```diff
@@ -719,6 +720,132 @@ func findSuperTypeFromUpperMask(joinedTypeTag TypeTag, types []Type) Type {
...
+
+		references = append(references, referenceType)
+
+		if prevReferenceType == nil {
+			prevReferenceType = referenceType
+			continue
+		}
+
+		if !referenceType.Type.Equal(prevReferenceType.Type) {
+			return getSuperTypeOfDerivedTypes(types)
```
SupunS, line 817: Same here, maybe could use `commonSuperTypeOfHeterogeneousTypes`?

SupunS (approved):
Awesome!

runtime/sema/type_tags.go
```diff
@@ -719,6 +720,132 @@ func findSuperTypeFromUpperMask(joinedTypeTag TypeTag, types []Type) Type {
...
+			return leastCommonAccess(accessB, accessA)
+		case Disjunction:
+			// least common access of two disjunctions is their union
+			// e.g. the least common supertype of (E | F) and (E | G) is (E | F | G)
+			union := orderedmap.KeySetUnion(setAccessA.Entitlements, setAccessB.Entitlements)
+			return NewAccessFromEntitlementSet(union, Disjunction)
+		}
+	}
+
+	return UnauthorizedAccess
```
turbolent, line 788: Should this ever occur? If so, could you please add a test. If not, maybe panic with an unreachable error

runtime/sema/type_tags.go
```diff
@@ -719,6 +720,132 @@ func findSuperTypeFromUpperMask(joinedTypeTag TypeTag, types []Type) Type {
...
+		// 'Never' type doesn't affect the supertype.
+		// Hence, ignore them
+		if typ == NeverType {
+			continue
+		}
+
+		referenceType, ok := typ.(*ReferenceType)
+		if !ok {
+			return commonSuperTypeOfHeterogeneousTypes(types)
+		}
```
turbolent, line 807: Can you please add a test case that covers this case?

turbolent (commented):
Nice!

runtime/sema/type_tags.go
dsainati1: I think this is unreachable, as we should only get into this function if the mask is a reference type

turbolent (approved):

PR #3170, opened 2024-03-13
Use atree readonly iterators when mutation of values is not needed

### LINKED ISSUE #2836: Use atree's new read-only iterators when mutation of returned values are not required

fxamacker:
Closes #2836

Atree readonly iterators are more performant than non-readonly iterators.  This PR uses readonly iterators when feasible to prevent performance regression.

______

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -1973,12 +1973,14 @@ func (v *ArrayValue) Concat(interpreter *Interpreter, locationRange LocationRang
 
 	first := true
 
-	firstIterator, err := v.array.Iterator()
+	// Use ReadOnlyIterator here because new ArrayValue is created with elements copied (not removed) from original value.
+	firstIterator, err := v.array.ReadOnlyIterator()
 	if err != nil {
 		panic(errors.NewExternalError(err))
 	}
 
-	secondIterator, err := other.array.Iterator()
+	// Use ReadOnlyIterator here because new ArrayValue is created with elements copied (not removed) from original value.
+	secondIterator, err := other.array.ReadOnlyIterator()
 	if err != nil {
 		panic(errors.NewExternalError(err))
 	}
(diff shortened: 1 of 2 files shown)


dsainati1 (approved):

SupunS (approved):

runtime/interpreter/value.go
```diff
@@ -17936,9 +17943,7 @@ func (v *CompositeValue) ForEachFieldName(
 	f func(fieldName string) (resume bool),
 ) {
 	iterate := func(fn atree.MapElementIterationFunc) error {
-		return v.dictionary.IterateKeys(
-			StringAtreeValueComparator,
-			StringAtreeValueHashInput,
+		return v.dictionary.IterateReadOnlyKeys(
```
turbolent, line 17946: Not sure if it's guaranteed that all uses of `CompositeValue.ForEachFieldName` are only read-only.

runtime/interpreter/value.go
```diff
@@ -18633,12 +18638,8 @@ func (v *DictionaryValue) IterateKeys(
 	locationRange LocationRange,
 	f func(key Value) (resume bool),
 ) {
-	valueComparator := newValueComparator(interpreter, locationRange)
-	hashInputProvider := newHashInputProvider(interpreter, locationRange)
 	iterate := func(fn atree.MapElementIterationFunc) error {
-		return v.dictionary.IterateKeys(
-			valueComparator,
-			hashInputProvider,
+		return v.dictionary.IterateReadOnlyKeys(
```
turbolent, line 18642: Same here: Not sure that `f` in all uses of `DictionaryValue.IterateKeys` are always read-only

turbolent (commented):
Nice!

How does this compare to the v0.42 branch/variant?

runtime/interpreter/value.go
fxamacker: @turbolent Good catch! I reverted to use non-readonly iterator here (and below for other comment).

runtime/interpreter/value.go
fxamacker: Good catch! I reverted to use non-readonly iterator here.

turbolent (approved):
Nice!

PR #3176, opened 2024-03-14
Proper type checking for resource use after validation on two-value transfers

dsainati1:
Closes https://github.com/dapperlabs/cadence-internal/issues/169
Closes https://github.com/dapperlabs/cadence-internal/issues/134

Second value transfers were not being properly checked in the same order as the interpreter was executing them, leading to missed errors for invalidated resource use. This fixes the checking to match the interpreter. 

______


runtime/tests/checker/access_test.go
```diff
@@ -1849,7 +1849,7 @@ func TestCheckAccessImportGlobalValueVariableDeclarationWithSecondValue(t *testi
 		},
 	)
 
-	errs := RequireCheckerErrors(t, err, 5)
+	errs := RequireCheckerErrors(t, err, 7)
```
dsainati1, line 1852: This is a small but unfortunate regression of re-evaluating the middle value of the two-value transfer; the resource capturing error gets reported twice. I think we could just get rid of this by deduplicating the errors list.

runtime/sema/check_variable_declaration.go
```diff
@@ -173,6 +199,11 @@ func (checker *Checker) visitVariableDeclarationValues(declaration *ast.Variable
 				declaration.SecondTransfer,
 				true,
 			)
+
+			if recordedResourceInvalidation != nil {
+				checker.resources.RemoveTemporaryMoveInvalidation(recordedResourceInvalidation.resource, recordedResourceInvalidation.invalidation)
+			}
+			checker.VisitExpression(declaration.Value, expectedValueType)
```
SupunS, line 206: Could we do something similar to the swap statement is checked, where the expression is visited once as the target/lhs of as assignment (using `checker.visitAssignmentValueType(declaration.Value)`), and then the next time it is visited as the value-expression/rhs?

dsainati1: Yes, technically, but because the expression is necessarily a variable, member or index expression, these two checks are semantically identical.

runtime/sema/checker.go
```diff
@@ -1559,16 +1559,19 @@ func (checker *Checker) recordResourceInvalidation(
 	switch expression.(type) {
 	case *ast.MemberExpression:
 
-		if accessedSelfMember == nil ||
-			!checker.allowSelfResourceFieldInvalidation {
+		if (accessedSelfMember == nil ||
+			!checker.allowSelfResourceFieldInvalidation) &&
+			invalidationKind != ResourceInvalidationKindMoveTemporary {
```
SupunS, line 1564: Curious why this additional check was needed? is there any particular test that is failing without it?

SupunS (approved):

runtime/sema/checker.go
dsainati1: Yeah, if you remove this then the checking for `let x <- r.y <- e` will fail. Normally a field member cannot be invalidated in this way, which is what this check is designed to prevent. If we don't add this case, then attempting to invalidate the middle `e1` in `let x <- e1 <- e2` will fail, since in this example `e1` is a member access and can't be invalidated. However, it's okay in this case because the invalidation is temporary (hence the `Temporary` move kind), so I just added an extra condition here to not report an erroneous error for this specific case.

turbolent (approved):
Great work!

PR #3181, opened 2024-03-18
Resource Reference Invalidation Improvements

dsainati1:
Closes https://github.com/dapperlabs/cadence-internal/issues/177

This properly invalidates optional resource references, and properly handles the case where an array or dictionary containing an invalid reference is logged or otherwise iterated on. 

______


runtime/interpreter/interpreter_expression.go
```diff
@@ -382,6 +382,10 @@ func (interpreter *Interpreter) evalExpression(expression ast.Expression) Value
 }
 
 func (interpreter *Interpreter) checkInvalidatedResourceOrResourceReference(value Value, hasPosition ast.HasPosition) {
+	if interpreter == nil {
+		return
+	}
```
turbolent, line 387: Why would the receiver be `nil` here?

runtime/interpreter/value.go
```diff
@@ -1940,6 +1940,7 @@ func (v *ArrayValue) iterate(
 			// atree.Array iteration provides low-level atree.Value,
 			// convert to high-level interpreter.Value
 			elementValue := MustConvertStoredValue(interpreter, element)
+			interpreter.checkInvalidatedResourceOrResourceReference(elementValue, locationRange)
```
turbolent, line 1943: Is adding the calls to `checkInvalidatedResourceOrResourceReference` in the iteration functions (here for arrays, and below for dictionaries and composites) the main fix?

turbolent:
Nice!

runtime/interpreter/value.go
dsainati1: Yes, essentially

runtime/interpreter/interpreter_expression.go
dsainati1: There's a very specific possible sequence of calls that can get here: if `checkInvalidatedResourceOrResourceReference` is called from inside an array/dictionary/composite iteration (as is introduced in this PR), which is triggered by a call to `MeteredString`, which itself is triggered by a call to `RecursiveString` or `String`. While it makes sense to add the full interpreter as an argument to `MeteredString`, requiring the same for calls to `String` or `RecursiveString` is not reasonable, so they use a `nil` interpreter if they delegate to `MeteredString`.

turbolent (approved):
Great work! 👏

SupunS (approved):

Issue #3098, opened 2024-02-12
QA full Cadence 1.0 state migration in Emulator


turbolent:
Goal: Someone (in 4D?) can take a test contract, take snapshot based on v0.42 and run migration on emulator and validate the upgraded contract works correctly.

depends on: https://github.com/onflow/flow-cli/issues/1433

SupunS:
I have already done this testing, so guess we can mark this as done.

Maybe one last thing to test is the final CLI build (for the milestone release), once it is out.

turbolent:
Nice work doing some initial testing Supun!

Eventually, it would be good to do a final test before the release. 

There are still some improvements to the state migration outstanding: https://github.com/onflow/flow-go/pull/5479 and we should re-test once those landed in the Emulator.

It would also be great to test the state migration on a "larger" / "real" project. I had asked Dev Ex on Discord here: https://discord.com/channels/613813861610684416/1123314820763111465/1210727371905171496

SupunS:
ah nice!

turbolent:
Forgot to mention / just to be clear: Ideally this isn't done by "us" (Cadence team), but maybe by someone familiar with such an example project. cc @j1010001 to maybe connect with Dev Ex / the community

nvdtf:
I tried testing from CLI branch using `ExampleNFT`. When I run `... migrate state --db-path=./pre-1.0-state/emulator.sqlite --save-report=./reports --contracts="ExampleNFT"` I get:
``` 
4:37PM INF created account migrations inner_migrations=1 nWorker=1
4:37PM INF Sorting payloads by address payloads=256 workers=1
4:37PM INF Sorted. Finding account boundaries in sorted payloads duration=0s payloads=256
4:37PM INF Done grouping payloads by account accounts=12 duration=0s
4:37PM ERR could not migrate account error="failed to find all contract registers that need to be changed for address f8d6e0586b0a20c7:\n  - RandomBeaconHistory\n" address=f8d6e0586b0a20c7 migration=*migrations.ChangeContractCodeMigration migration_index=0
4:37PM INF waiting for migration workers to finish
4:37PM INF Top longest migrations top_longest_migrations=[]
4:37PM INF closing migration migration=*migrations.ChangeContractCodeMigration migration_index=0
[...]

PR #3184, opened 2024-03-20
Relax interface conformance changes in contract update validator

SupunS:
Working towards https://github.com/onflow/cadence/issues/3098#issuecomment-2007977094

## Description

Includes:
- [Consider inherited interfaces when checking for conformances](https://github.com/onflow/cadence/commit/6d56a062dd9ee14e1cb84fa57544f4860415208c)
  - It is OK to remove the explicit conformance to an interface, as long as it is implicitly available via inheritance.
- [Consider custom composite/interface type change rules](https://github.com/onflow/cadence/commit/a78ac56a86e2c074ea51041b60bce7ffb76bde32)
  - We already allow defining custom rules for static type changes. Some of these are also applicable for the conformance changes. 
  - e.g: `MetadataViews.Resolver` is changed to `ViewResolver.Resolver`.
  - So consider these custom rules as well, when checking for conformance changes.

______

### DIFF
--- a/runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
+++ b/runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
@@ -97,7 +97,12 @@ func (validator *CadenceV042ToV1ContractUpdateValidator) Validate() error {
 	validator.TypeComparator.expectedIdentifierImportLocations = collectImports(validator, underlyingValidator.oldProgram)
 	validator.TypeComparator.foundIdentifierImportLocations = collectImports(validator, underlyingValidator.newProgram)
 
-	checkDeclarationUpdatability(validator, oldRootDecl, newRootDecl)
+	checkDeclarationUpdatability(
+		validator,
+		oldRootDecl,
+		newRootDecl,
+		validator.checkConformanceV1,
+	)
 
 	if underlyingValidator.hasErrors() {
 		return underlyingValidator.getContractUpdateError()
@@ -231,7 +236,7 @@ func (validator *CadenceV042ToV1ContractUpdateValidator) requireEqualAccess(
 }
 
 func (validator *CadenceV042ToV1ContractUpdateValidator) expectedAuthorizationOfComposite(composite *ast.NominalType) sema.Access {
-	// if this field is set, we are currently upgrading a formerly legacy restricted type into a reference to a composite
+	// If this field is set, we are currently upgrading a former legacy restricted type into a reference to a composite
 	// in this case, the expected entitlements are based not on the underlying composite type,
 	// but instead the types previously in the restriction set
 	if validator.currentRestrictedTypeUpgradeRestrictions != nil {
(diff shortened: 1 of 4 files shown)


runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
```diff
@@ -46,11 +47,20 @@ func NewCadenceV042ToV1ContractUpdateValidator(
 	contractName string,
 	provider AccountContractNamesProvider,
 	oldProgram *ast.Program,
-	newProgram *ast.Program,
+	newProgram *interpreter.Program,
```
SupunS, line 50: Accept an `interpreter.Program`, because we also need the elaboration of the current program

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
```diff
@@ -152,17 +167,24 @@ func (validator *CadenceV042ToV1ContractUpdateValidator) idAndLocationOfQualifie
...
 
-	foundLocations := validator.TypeComparator.foundIdentifierImportLocations
+	newImportLocations := validator.TypeComparator.foundIdentifierImportLocations
+	oldImportLocations := validator.TypeComparator.expectedIdentifierImportLocations
 
-	if typIdentifier != rootIdentifier && foundLocations[typIdentifier] == nil {
-		qualifiedString = fmt.Sprintf("%s.%s", rootIdentifier, qualifiedString)
-		return common.NewTypeIDFromQualifiedName(nil, location, qualifiedString), location
+	// Here we only need to find the qualified type ID.
+	// So check in both old imports as well as in new imports.
```
SupunS, line 175: In migrations, we also allow changing types to a new type that is in a different import location. e.g: from `MetadataViews.Resolver` to `ViewResolver.Resolver`.
So when computing the type ID for the old type, it will not be found in the new import locations (i.e: `foundIdentifierImportLocations`). So we need to check in the old import locations (`expectedIdentifierImportLocations`) as well.

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
```diff
@@ -499,6 +533,92 @@ func (validator *CadenceV042ToV1ContractUpdateValidator) checkDeclarationKindCha
...
+		for index, newConformance := range newConformances {
+			newConformanceNominalType := semaConformanceToASTNominalType(newConformance)
+
+			err := oldConformance.CheckEqual(newConformanceNominalType, validator)
+
+			var customRuleChecked, customRuleValid bool
+			if err != nil {
+				customRuleChecked, customRuleValid =
+					validator.checkUserDefinedTypeCustomRules(oldConformance, newConformanceNominalType)
+			}
```
SupunS, line 577: If the usual type equality fails, also check in the special rules to see if this type change is a valid one.
The value migrations (static type migration etc.) already use these rules to migrate value. With this change, contract update validator becomes consistent with the value migrations.

(Rest of the logic is same as the old contract update validator)

dsainati1 (approved):

Issue #3188, opened 2024-03-25
Run existing tests for C1.0 contract update validator


SupunS:
### Issue to be solved

The existing contract update validation tests are only ran for the old validator.

### Suggested Solution

Also run them for the updated C1.0 contract update validator

Issue #3186, opened 2024-03-25
Contract update fails if there were enums as type-requirements


SupunS:
### Current Behavior

When updating a contract that has an enum as a type-requirement, it complains about mismatching conformance.

e.g: Assume both the old code and the new code are the same:
```cadence
contract interface CI { 
    enum E { ... }
}

contract C: CI {
  enum E { ... }
}
```

Then updating (with the same code) complains about mismatching conformance for `C.E`

### Expected Behavior

Should allow updating it to the same code.

### Steps To Reproduce

See above

### Environment

```markdown
- Cadence version: current master
- Network: N/A
```

PR #3191, opened 2024-03-25
Fix conformance checking for enums in contract update validator

SupunS:
Closes https://github.com/onflow/cadence/issues/3186
Closes https://github.com/onflow/cadence/issues/3188

______


runtime/contract_update_validation_c1_test.go
SupunS, line 1: Unfortunately there is no easy way to parameterize an entire test file in Go to re-run with different parameters (different validators). So I just copied over the existing tests, and changed the validator only for these new (copied) tests

runtime/stdlib/account.go
```diff
@@ -1595,31 +1596,33 @@ func changeAccountContracts(
...
+		var oldProgram *ast.Program
 
-		var legacyContractUpgrade bool
-		// if we are allowing legacy contract upgrades, fall back to the old parser when the new one fails
-		if !ignoreUpdatedProgramParserError(err) && legacyUpgradeEnabled {
-			legacyContractUpgrade = true
+		// It is not always possible to determine whether the old code is pre-1.0 or not,
+		// only based on the parser errors. Therefore, always rely on the flag only.
+		// If the legacy contract upgrades are enabled, then use the old parser.
+		if legacyUpgradeEnabled {
```
SupunS, line 1604: This is an unrelated change, but realized that the existing condition here used  to rely on parser errors to determine whether this is a legacy-code update. 
However, not all pre-1.0 codes may have parser errors (majority of them would have checking errors, but probably not parser errors always). Therefore, make this simpler, and always use the config flag to determine whether this is a legacy-code update or not. Developers can restart the emulator with/without flag to switch between the use-cases.

runtime/contract_update_validation_c1_test.go
turbolent: Would it be possible to create a testing helper function `func testValidators(test validartorTester)`, where `type validatorTester func(﻿testDeployAndUpdate func(...), useContractUpdateValidator bool)`, which runs the given `test` function twice, once for the default validator, and once for the 1.0 validator? 

Hope it makes sense what I mean, can elaborate if needed. Also happy to open a PR against this one with the changes

runtime/stdlib/account.go
turbolent: Good catch! Agreed

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
```diff
@@ -538,12 +540,27 @@ func (validator *CadenceV042ToV1ContractUpdateValidator) checkConformanceV1(
...
+
+	// NOTE 1: Here it is assumed enums will always have one and only one conformance.
 	// This is enforced by the checker.
-	// Therefore, below check for multiple conformances is only applicable
-	// for non-enum type composite declarations. i.e: structs, resources, etc.
+	//
+	// NOTE 2: If one declaration is an enum, then other is also an enum at this stage.
+	// This is enforced by the validator (in `checkDeclarationUpdatability`), before calling this function.
+	if newDecl.Kind() == common.CompositeKindEnum {
+		err := oldConformances[0].CheckEqual(newDecl.Conformances[0], validator)
```
turbolent, line 551: Is the update checker always only run if checking of the new (and old?) contract succeeded? If not, maybe still check the lengths of the conformances before indexing.

turbolent (commented):
Fixes for enums look good! It would be nice if we could avoid the code duplication for the tests

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
SupunS: existing code should have been type checked already when deploying. The new code gets type checked before running the validator. So the size should already be validated for both programs.
Can still add a check though, just to be safe.

dsainati1 (approved):

runtime/contract_update_validation_c1_test.go
SupunS: refactored in https://github.com/onflow/cadence/pull/3191/commits/595e5a35f9efc1e2f21965d7f5a91f2744aac402

turbolent (approved):
Great work! Thank you for refactoring the tests 👌

Issue #3182, opened 2024-03-19
Enums in Contract Interfaces cannot be updated to Cadence 1.0


dsainati1:
### Issue to be solved

Enums were previously allowed in contract interfaces as a type requirement, but with the removal of type requirements they are no longer allowed to appear there. However, there is no current migration path for any old enums of this kind, as there are no enum interfaces to update them to, and the update validator does not allow the removal of a type declaration. 

### Suggested Solution

We should either allow enum declarations to be removed from contract interfaces (which should be safe because the enum will be declared in all contracts that conform to that interface) or come up with an alternative data migration to support this case.

PR #3197, opened 2024-03-26
Allow removal of enum declarations from contract interfaces for C1.0 upgrade

dsainati1:
Closes https://github.com/onflow/cadence/issues/3182

This allows enum declarations to be removed from contract interfaces as part of the Cadence 1.0 upgrade. While in general this is not a safe operation to permit, analysis of mainnet contracts indicates that there are a small number of existing declarations of this kind, none of which are used in fields that themselves cannot be easily removed. 

______

### DIFF
--- a/runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validation_test.go
+++ b/runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validation_test.go
@@ -189,6 +189,13 @@ func getSingleContractUpdateErrorCause(t *testing.T, err error, contractName str
 	return updateErr.Errors[0]
 }
 
+func assertMissingDeclarationError(t *testing.T, err error, declName string) bool {
+	var missingDeclError *stdlib.MissingDeclarationError
+	require.ErrorAs(t, err, &missingDeclError)
+
+	return assert.Equal(t, declName, missingDeclError.Name)
+}
+
 func getContractUpdateError(t *testing.T, err error, contractName string) *stdlib.ContractUpdateError {
 	require.Error(t, err)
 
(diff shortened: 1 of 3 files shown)


runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
```diff
@@ -533,6 +533,27 @@ func (validator *CadenceV042ToV1ContractUpdateValidator) checkDeclarationKindCha
 	return false
 }
 
+func (validator *CadenceV042ToV1ContractUpdateValidator) checkNestedDeclarationRemoval(
```
turbolent, line 536: Nice!

turbolent (approved):
Looks good!

Could you please also add or check we already have a test that it is not possible to add a new enum declaration?

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validation_test.go
```diff
@@ -2066,3 +2073,51 @@ func TestInterfaceConformanceChange(t *testing.T) {
...
+            }
+        `
+
+		const newCode = `
+            access(all) contract Test {}
+        `
+
+		err := testContractUpdate(t, oldCode, newCode)
+
+		// This is not allowed because `@R{I}` is converted to `@R`, not `@{I}`
```
SupunS, line 2097:
```suggestion
```

SupunS (approved):

dsainati1:
> Could you please also add or check we already have a test that it is not possible to add a new enum declaration?

Why should this not be possible? This should be allowed even by the regular validator.

turbolent:
> > Could you please also add or check we already have a test that it is not possible to add a new enum declaration?
> 
> Why should this not be possible? This should be allowed even by the regular validator.

We can only allow the removal if we also forbid the addition, otherwise an incompatible declaration could be reintroduced, circumventing the update checking rules for enums.

dsainati1:
Oh, yes in that case the new type checking will prevent enums from being added to interfaces implicitly; those contracts won't type check so there's no need to forbid it in the validator.

Issue #3192, opened 2024-03-25
State migration fails in some cases


turbolent:
### Current Behavior

The state migration still runs into some errors. In particular, the log for running the migration on the full TN state (https://www.notion.so/flowfoundation/State-Migration-for-Crescendo-Release-56fe231eb56a4a5383f919ffe444d109?pvs=4#679a28687bbf49e9b3b52697820bc7ae) revealed the following problems:

---

- **Problem:** During the static type migration, values are encountered which have nil types, even though they should not.

  For example:

    ```
    6:40AM ERR failed to run StaticTypeMigration in account b87e00279f5bc9c6, domain   storage, key ChildAccountTag: internal error: unexpected static type: <nil>
    goroutine 1685 [running]:
    runtime/debug.Stack()
    	/opt/homebrew/Cellar/go@1.20/1.20.14/libexec/src/runtime/debug/stack.go:24 +0x65
    github.com/onflow/cadence/runtime/errors.NewUnexpectedError({0x6880db?, 0x1f0c61b?},   {0xf3fc43aaf8?, 0xab6ce0?, 0x2d94e60?})
    	/Users/bastian/go/pkg/mod/github.com/onflow/cadence@v1.0.0-preview.14/runtime/errors/  errors.go:156 +0x4b
    github.com/onflow/cadence/migrations/statictypes.(*StaticTypeMigration).  maybeConvertStaticType(0xf3fc43abc0?, {0x0?, 0x0?}, {0xaa4008?, 0xff1bbabe00?})
    	/Users/bastian/go/pkg/mod/github.com/onflow/cadence@v1.0.0-preview.14/migrations/  statictypes/statictype_migration.go:450 +0x889
    github.com/onflow/cadence/migrations/statictypes.(*StaticTypeMigration).  maybeConvertStaticType(0xff1bbab840, {0xaa4008?, 0xff1bbabe00?}, {0xaa4110?,   0xff99da3180?})
    	/Users/bastian/go/pkg/mod/github.com/onflow/cadence@v1.0.0-preview.14/migrations/  statictypes/statictype_migration.go:185 +0x265
    github.com/onflow/cadence/migrations/statictypes.(*StaticTypeMigration).  maybeConvertStaticType(0xff1bbab840, {0xaa4110?, 0xff99da3180?}, {0x0?, 0x0?})
[...]

PR #3199, opened 2024-03-27
Only encode reference static type in legacy format if it was decoded as such

turbolent:
Work towards #3192 

## Description

When removing an existing key from a dictionary, a "legacy key" is used, which encodes the key in its original form.

However, only do so if the key was actually decoded in legacy form.

In particular, reference static types may have been already migrated, e.g. the entitlements migration adds authorization with an entitlement set, which then should not be encoded in legacy form, i.e. with an authorization flag.

______

### DIFF
--- a/migrations/entitlements/migration_test.go
+++ b/migrations/entitlements/migration_test.go
@@ -3124,6 +3124,7 @@ func TestRehash(t *testing.T) {
 			interpreter.UnauthorizedAccess,
 			newCompositeType(),
 		)
+		refType.HasLegacyIsAuthorized = true
 		refType.LegacyIsAuthorized = true
 
 		legacyRefType := &migrations.LegacyReferenceType{
--- a/migrations/legacy_reference_type.go
+++ b/migrations/legacy_reference_type.go
@@ -35,6 +35,11 @@ type LegacyReferenceType struct {
 var _ interpreter.StaticType = &LegacyReferenceType{}
 
 func (t *LegacyReferenceType) ID() common.TypeID {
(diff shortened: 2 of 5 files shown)


dsainati1 (approved):

SupunS (commented):
Good catch! I'm wondering whether we need to solve this generically. As in, when removing the key, should we maybe first try removing with the existing key, and if it fails, then fallback and try remove with the legacy key. That will of course add some overhead, but it will also allow running the migration multiple time, in future, I guess?

turbolent:
I think the solution already solves this "generically" / supports running the migration multiple times: Originally, when the reference type is decoded for the "first" time, and it has a legacy authorization flag, then the legacy key wrapper will encode it as such again. A migration may then replace it with a reference type that has a new authorization. When the migration is run again, it is decoded with the new authorization, and the legacy key wrapper will use the new encoding. But maybe I'm missing something. It's a bit hard to test

SupunS (approved):

turbolent:
@SupunS please let me know if you still think there's something we need to handle here / there's a flaw in my reasoning. My plan was to merge this and see how it works. We might still need to refine this further. 

Also, I had a look at the other "legacy wrappers", but AFAICS they do not have a similar problem. Again, I might be missing something

SupunS:
ah no, that makes sense. My concern was mostly with the rest of the legacy wrappers, but I had a look at it them again, and I also felt that they would seamlessly work. We can revisit only if we run into something similar, otherwise no need to try and fix something that is not broken :D

PR #3059, opened 2024-01-30
Extend type bounds

turbolent:
Work towards https://github.com/dapperlabs/cadence-internal/issues/188

## Description

See https://github.com/dapperlabs/cadence-internal/issues/188#issuecomment-1906752366

Currently, type bounds for type parameters may only be subtype relationships.

Generalize the type bounds feature to support:
- Subtype relations
- Strict subtype relations
- Supertype relations
- Strict supertype relations
- Conjunctions of the above

```[tasklist]
### TODO
- [ ] Serialization in type ID / syntax
- [ ] Import/export
``` 
______

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #3203, opened 2024-03-28
Properly handle `Never` type argument to `Capabilities.get`

dsainati1:
Closes https://github.com/dapperlabs/cadence-internal/issues/176

Properly return a `nil` value in this case rather than panicking. 

______

### DIFF
--- a/runtime/capabilities_test.go
+++ b/runtime/capabilities_test.go
@@ -236,6 +236,13 @@ func TestRuntimeCapability_borrowAndCheck(t *testing.T) {
               }
 
               access(all)
+              fun testNever() {
+                  let path = /public/r
+                  assert(self.account.capabilities.get<Never>(path) == nil)
+                  assert(self.account.capabilities.exists(path))
+              }
+
+              access(all)
               fun testSwap(): Int {
                  let ref = self.account.capabilities.get<&R>(/public/r)!.borrow()!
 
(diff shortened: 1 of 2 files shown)


turbolent (approved):
Thank you for looking into this!

We can definitely merge this fix, but maybe also want to improve this further: Currently `Never` is only rejected at run-time, we maybe want to already reject this statically.

We currently do not have the means in the language to express the constraint that the type parameter should have a lower bound. However, we have a mechanism for it in the checker: `FunctionType.TypeArgumentsCheck`. Maybe extend this solution by adding such a static check (potentially in a separate PR).

We have a similar case in / report for `revertibleRandom`, see https://github.com/dapperlabs/cadence-internal/issues/188. In a comment of the issue I noted how it would be nice to extend the type bounds to express this in a way where it is also communicated to users: https://github.com/dapperlabs/cadence-internal/issues/188#issuecomment-1906752366. I had opened https://github.com/onflow/cadence/pull/3059 a while back with a PoC of this idea, maybe have a look, I'd love to hear your thoughts!

dsainati1:
Yeah this makes sense. What is left to do on https://github.com/onflow/cadence/pull/3059? I can pick this up

turbolent:
@dsainati1 sounds good! I tried to leave TODOs in both the PR description (see tasklist) as well as in the code. We can maybe discuss it in the next Impl. Sync

SupunS (approved):

PR #3204, opened 2024-03-28
Add additional fields to account key added event

### LINKED ISSUE #3146: emit account key information in AccountKeyAdded event
### Issue to be solved

the old `AccountKeyAddedFromByteArrayEventType` event contained a publickey field which is a byte encoded value of `runtimeAccountPublicKeyWrapper` [injected dependency defined here](https://github.com/onflow/flow-go/blob/master/model/flow/account_encoder.go#L36) With this event payload one is able to identify all information of the public key that was added to this account except for index. 
the new `AccountKeyAddedFromPublicKeyEventType` event only contains the raw public key itself with no additional account key information. 

### Suggested Solution

emit the following information in `AccountKeyAddedFromPublicKeyEventType`:
- account address
- public key raw value
- sig algo
- sign algo
- weight
- keyIndex

rrrkren:
Closes #3146

## Description

adding the following information to `flow.AccountKeyAdded` event:
- hash algorithm
- weight
- key index

______

### DIFF
--- a/runtime/account_test.go
+++ b/runtime/account_test.go
@@ -539,17 +539,17 @@ func TestRuntimeAuthAccountKeysAdd(t *testing.T) {
 	require.NoError(t, err)
 
 	assert.EqualValues(t, key0Weight, key0AddedEvent.Fields[2])
-	assert.EqualValues(t, sema.UFix64Type.ID(), key0AddedEvent.Fields[2].Type().ID())
+	assert.EqualValues(t, sema.UFix64TypeName, key0AddedEvent.Fields[2].Type().ID())
 	assert.EqualValues(t, key1Weight, key1AddedEvent.Fields[2])
 
 	// key hash algorithm
 	assert.EqualValues(t, sema.HashAlgorithmSHA3_256, key0AddedEvent.Fields[3].(cadence.Enum).Fields[0])
-	assert.EqualValues(t, sema.HashAlgorithmType.ID(), key0AddedEvent.Fields[3].Type().ID())
+	assert.EqualValues(t, sema.HashAlgorithmTypeName, key0AddedEvent.Fields[3].Type().ID())
 	assert.EqualValues(t, sema.HashAlgorithmSHA2_256, key1AddedEvent.Fields[3].(cadence.Enum).Fields[0])
 
 	// key index
 	assert.EqualValues(t, cadence.NewInt(0), key0AddedEvent.Fields[4])
-	assert.EqualValues(t, sema.IntType.ID(), key0AddedEvent.Fields[4].Type().ID())
+	assert.EqualValues(t, sema.IntTypeName, key0AddedEvent.Fields[4].Type().ID())
 	assert.EqualValues(t, cadence.NewInt(1), key1AddedEvent.Fields[4])
 }
 
--- a/runtime/account_test.go
+++ b/runtime/account_test.go
@@ -225,7 +225,7 @@ type accountTestEnvironment struct {
 func newAccountTestEnv() accountTestEnvironment {
 	storage := newTestAccountKeyStorage()
 	rt := NewTestInterpreterRuntime()
-	rtInterface := getAccountKeyTestRuntimeInterface(storage)
+	rtInterface := newAccountKeyTestRuntimeInterface(storage)
 
 	addPublicKeyValidation(rtInterface, nil)
 
@@ -471,15 +471,19 @@ func TestRuntimeAuthAccountKeysAdd(t *testing.T) {
 
 	rt := NewTestInterpreterRuntime()
 
-	pubKey := newBytesValue([]byte{1, 2, 3})
+	pubKey1 := []byte{1, 2, 3}
+	pubKey1Value := newBytesValue(pubKey1)
+
+	pubKey2 := []byte{4, 5, 6}
+	pubKey2Value := newBytesValue(pubKey2)
 
 	const code = `
-       transaction(publicKey: [UInt8]) {
+       transaction(publicKey1: [UInt8], publicKey2: [UInt8]) {
            prepare(signer: auth(BorrowValue) &Account) {
                let acct = Account(payer: signer)
                acct.keys.add(
                    publicKey: PublicKey(
-                       publicKey: publicKey,
+                       publicKey: publicKey1,
                        signatureAlgorithm: SignatureAlgorithm.ECDSA_P256
                    ),
                    hashAlgorithm: HashAlgorithm.SHA3_256,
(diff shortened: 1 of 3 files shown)


rrrkren:
@turbolent I did not make the changes you requested wrt `AccountKeyRemoved` because I was not sure what's the safest way to remove an event. Also the `AccountKeyRemoved` currently uses a field called `publicKey` to indicate the index of the publicKey.... which clashes with the `publicKey` field we have in `AccountKeyAdded`. So the consistency game is already off.....

runtime/stdlib/flow.go
```diff
@@ -209,6 +209,21 @@ var AccountEventContractParameter = sema.Parameter{
 	TypeAnnotation: sema.StringTypeAnnotation,
 }
 
+var AccountEventHashAlgorithmParameter = sema.Parameter{
+	Identifier:     "hashAlgorithm",
+	TypeAnnotation: sema.UInt8TypeAnnotation,
```
rrrkren, line 214: not sure if this is the correctly to identify a enum

turbolent: Enums are essentially structs with an `rawValue` field. It's probably better to emit an event with more type information, i.e.

```suggestion
	TypeAnnotation: sema.HashAlgorithmTypeAnnotation,
```

The code in `newAccountKeysAddFunction` and in the test already is correct, only the type information here is wrong.

runtime/stdlib/flow.go
```diff
@@ -209,6 +209,21 @@ var AccountEventContractParameter = sema.Parameter{
...
+}
+
+var AccountEventKeyWeightParameter = sema.Parameter{
+	Identifier:     "weight",
+	TypeAnnotation: sema.IntTypeAnnotation,
```
turbolent, line 219: It's unfortunate that we used `UFix64` for the weight in the Cadence API, in `AccountKey` and in `keys.add`, but under the hood it's just an `int`.

Still, to maybe be consistent, use `UFix64` here.

runtime/stdlib/account.go
```diff
@@ -610,7 +610,8 @@ func newAccountKeysAddFunction(
 				panic(err)
 			}
 
-			hashAlgo := NewHashAlgorithmFromValue(inter, locationRange, invocation.Arguments[1])
+			hashAlgoValue := invocation.Arguments[1]
+			hashAlgo := NewHashAlgorithmFromValue(inter, locationRange, hashAlgoValue)
 			weightValue, ok := invocation.Arguments[2].(interpreter.UFix64Value)
```
turbolent, line 615: While at it, maybe separate the handling of the arguments a bit more:

```suggestion

			weightValue, ok := invocation.Arguments[2].(interpreter.UFix64Value)
```

turbolent (commented):
Nice! Just a couple minor things

turbolent:
> @turbolent I did not make the changes you requested wrt `AccountKeyRemoved` because I was not sure what's the safest way to remove an event. Also the `AccountKeyRemoved` currently uses a field called `publicKey` to indicate the index of the publicKey.... which clashes with the `publicKey` field we have in `AccountKeyAdded`. So the consistency game is already off.....

Yeah, looks like the definition of `AccountKeyRemoved` would also need some clean up before adding more information, `publicKey` is not a public key, but an index, and it is the index of the account key.

turbolent:
@rrrkren once this is merged, could you please also update the docs in https://github.com/onflow/cadence-lang.org/tree/main/versioned_docs/version-1.0 ?

runtime/account_test.go
```diff
@@ -507,19 +515,42 @@ func TestRuntimeAuthAccountKeysAdd(t *testing.T) {
...
+	// key weight
+	key0Weight, err := cadence.NewUFix64("100.0")
+	require.NoError(t, err)
+	key1Weight, err := cadence.NewUFix64("0.0")
+	require.NoError(t, err)
+
+	assert.EqualValues(t, key0Weight, key0AddedEvent.Fields[2])
+	assert.EqualValues(t, sema.UFix64TypeName, key0AddedEvent.Fields[2].Type().ID())
+	assert.EqualValues(t, key1Weight, key1AddedEvent.Fields[2])
+
```
turbolent, line 544: Also add an assertion for the type of the third field of the second key

runtime/account_test.go
```diff
@@ -507,19 +515,42 @@ func TestRuntimeAuthAccountKeysAdd(t *testing.T) {
...
+	// key hash algorithm
+	assert.EqualValues(t, sema.HashAlgorithmSHA3_256, key0AddedEvent.Fields[3].(cadence.Enum).Fields[0])
+	assert.EqualValues(t, sema.HashAlgorithmTypeName, key0AddedEvent.Fields[3].Type().ID())
+	assert.EqualValues(t, sema.HashAlgorithmSHA2_256, key1AddedEvent.Fields[3].(cadence.Enum).Fields[0])
+
```
turbolent, line 549: Also add an assertion for the type of the fourth field of the second key

runtime/account_test.go
```diff
@@ -507,19 +515,42 @@ func TestRuntimeAuthAccountKeysAdd(t *testing.T) {
...
+	// key index
+	assert.EqualValues(t, cadence.NewInt(0), key0AddedEvent.Fields[4])
+	assert.EqualValues(t, sema.IntTypeName, key0AddedEvent.Fields[4].Type().ID())
+	assert.EqualValues(t, cadence.NewInt(1), key1AddedEvent.Fields[4])
 }
```
turbolent, line 554: Also add an assertion for type of the the fifth field of the second key

runtime/account_test.go
```diff
@@ -507,19 +515,42 @@ func TestRuntimeAuthAccountKeysAdd(t *testing.T) {
...
+		key0AddedEvent.Type().ID(),
 	)
+
+	// key weight
+	key0Weight, err := cadence.NewUFix64("100.0")
+	require.NoError(t, err)
+	key1Weight, err := cadence.NewUFix64("0.0")
+	require.NoError(t, err)
+
+	assert.EqualValues(t, key0Weight, key0AddedEvent.Fields[2])
```
turbolent, line 541: Here an below: Use `assert.Equal` to also assert types are correct

runtime/account_test.go
```diff
@@ -507,19 +515,42 @@ func TestRuntimeAuthAccountKeysAdd(t *testing.T) {
...
 
+	key0AddedEvent := storage.events[1]
+	key1AddedEvent := storage.events[2]
+
 	assert.EqualValues(t,
 		stdlib.AccountKeyAddedFromPublicKeyEventType.ID(),
-		storage.events[1].Type().ID(),
+		key0AddedEvent.Type().ID(),
 	)
+
```
turbolent, line 534: While at it, could you maybe please also add assertions for the first and second field?

turbolent (commented):
Looks good!

Just a couple small nits for the tests

turbolent (approved):

turbolent:
@onflow/cadence PTAL 🙏

SupunS (approved):

PR #3211, opened 2024-04-01
Improve Cadence 1.0 state migration

turbolent:
Work towards #3192 

## Description

- Follow-up on #3144: Add health check to remaining tests. This identified some incorrect preparation in the tests
- When a legacy value or type is involved in the dictionary keys, not only do values need to encode properly and types need to produce a proper ID, but also the equality functions need to handle comparison between legacy and non-legacy values/types
- As noted by @bluesign, dictionaries may contain keys that may "conflict" during the migration: for example, the dictionary might contain keys for both normalized and unnormalized strings, characters, and intersection types before the migration – if one key is migrated to the other, the migration should not silently overwrite the new key (which already existed)
- Some additional clean up. For example, do not use `t.Run` for test "steps". Still, use immediately-invoked functions to scope variables, so it is enforced that assertions/loads re-load from serialized state, and not in-memory values created during preparation

______

### DIFF
--- a/migrations/statictypes/account_type_migration_test.go
+++ b/migrations/statictypes/account_type_migration_test.go
@@ -789,7 +789,7 @@ func TestAccountTypeInNestedTypeValueMigration(t *testing.T) {
 						interpreter.NewUnmeteredCompositeField("field1", storedAccountTypeValue),
 						interpreter.NewUnmeteredCompositeField("field2", interpreter.NewUnmeteredStringValue("hello")),
 					},
-					common.Address{},
+					common.ZeroAddress,
 				)
 			},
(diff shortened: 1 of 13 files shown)


migrations/statictypes/account_type_migration_test.go
```diff
@@ -1206,233 +1206,250 @@ func TestAccountTypeRehash(t *testing.T) {
```
turbolent, line 1206: Best viewed with whitespace off. Refactors the "monolithic" test case which tested all types at once into individual subtests.

The new dictionary key conflict detection failed the test, because e.g. `AuthAccount.Keys` and `PublicAccount.Keys` now both map to `Account.Keys`, but the dictionary contained both old keys/types

SupunS (approved):

Issue #3081, opened 2024-02-07
Fixes of v0.42.8-patch.4 do not work on master


turbolent:
### Current Behavior

The fixes of v0.42.8-patch.4 work well on v0.42 (https://github.com/onflow/cadence/pull/3076), but they don't work fully on master: https://github.com/onflow/cadence/pull/3079

### Expected Behavior

All tests should pass

### Steps To Reproduce

Run tests on WIP PR branch #3079

### Environment

```markdown
- Cadence version:
- Network:
```

PR #3079, opened 2024-02-07
Port fixes of v0.42.8-patch.4

turbolent:
Closes #3081

## Description

Port of #3076 to `master`

Had to make some adjustments to make the fixes and tests compatible with Cadence 1.0 in d62622767ad85c729893a1d2dcc0c26abecdf979
 
______

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #3213, opened 2024-04-02
Convert references based on borrow type

SupunS:
Work towards https://github.com/onflow/cadence/issues/3081

### Description

In Cadence `v0.42`, the `valueType` (which is a `ReferenceStaticType`) included not only the referenced-type, but also the "borrowed type" of the value. So if the borrowed-type/referenced-type was different than the expected-type, then the conversion happened. This was added in https://github.com/dapperlabs/cadence-internal/pull/201.

However, in Cadence `1.0`, the the `ReferenceStaticType` (and therefore `valueType`) no longer contains the borrow-type. So even if the borrowed-type was different, the existing check would see the `valueType` and the `expectedType` as equal, and the conversion wouldn't happen. So this change here is to also consider the borrowed-type, just like it used to, but now the borrow-type has to be taken from the `ReferenceValue`, because `ReferenceStaticType` no longer has it.

The actual fix is https://github.com/onflow/cadence/pull/3213#discussion_r1548741022.

______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -2150,44 +2150,62 @@ func (interpreter *Interpreter) convert(value Value, valueType, targetType sema.
 		}
 
 	case *sema.ReferenceType:
-		if !valueType.Equal(unwrappedTargetType) {
-			// transferring a reference at runtime does not change its entitlements; this is so that an upcast reference
-			// can later be downcast back to its original entitlement set
-
-			// check defensively that we never create a runtime mapped entitlement value
-			if _, isMappedAuth := unwrappedTargetType.Authorization.(*sema.EntitlementMapAccess); isMappedAuth {
-				panic(UnexpectedMappedEntitlementError{
-					Type:          unwrappedTargetType,
-					LocationRange: locationRange,
-				})
-			}
-
-			switch ref := value.(type) {
-			case *EphemeralReferenceValue:
+		targetAuthorization := ConvertSemaAccessToStaticAuthorization(interpreter, unwrappedTargetType.Authorization)
+		switch ref := value.(type) {
+		case *EphemeralReferenceValue:
+			if interpreter.shouldConvertReference(ref, valueType, unwrappedTargetType, targetAuthorization) {
+				checkMappedEntitlements(unwrappedTargetType, locationRange)
 				return NewEphemeralReferenceValue(
 					interpreter,
-					ConvertSemaAccessToStaticAuthorization(interpreter, unwrappedTargetType.Authorization),
+					targetAuthorization,
 					ref.Value,
 					unwrappedTargetType.Type,
 					locationRange,
 				)
+			}
 
-			case *StorageReferenceValue:
+		case *StorageReferenceValue:
+			if interpreter.shouldConvertReference(ref, valueType, unwrappedTargetType, targetAuthorization) {
+				checkMappedEntitlements(unwrappedTargetType, locationRange)
 				return NewStorageReferenceValue(
 					interpreter,
-					ConvertSemaAccessToStaticAuthorization(interpreter, unwrappedTargetType.Authorization),
+					targetAuthorization,
 					ref.TargetStorageAddress,
 					ref.TargetPath,
 					unwrappedTargetType.Type,
 				)
-
-			default:
-				panic(errors.NewUnexpectedError("unsupported reference value: %T", ref))
 			}
+
+		default:
+			panic(errors.NewUnexpectedError("unsupported reference value: %T", ref))
 		}
 	}
 
 	return value
+}
+
+func (interpreter *Interpreter) shouldConvertReference(
+	ref ReferenceValue,
+	valueType sema.Type,
+	unwrappedTargetType *sema.ReferenceType,
+	targetAuthorization Authorization,
+) bool {
+	if !valueType.Equal(unwrappedTargetType) {
+		return true
+	}
+
+	return !ref.BorrowType().Equal(unwrappedTargetType.Type) ||
+		!ref.GetAuthorization().Equal(targetAuthorization)
+}
+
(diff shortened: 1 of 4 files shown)


runtime/interpreter/interpreter.go
```diff
@@ -2150,46 +2150,52 @@ func (interpreter *Interpreter) convert(value Value, valueType, targetType sema.
...
-			if _, isMappedAuth := unwrappedTargetType.Authorization.(*sema.EntitlementMapAccess); isMappedAuth {
-				panic(UnexpectedMappedEntitlementError{
-					Type:          unwrappedTargetType,
-					LocationRange: locationRange,
-				})
-			}
+		targetAuthorization := ConvertSemaAccessToStaticAuthorization(interpreter, unwrappedTargetType.Authorization)
+		switch ref := value.(type) {
+		case *EphemeralReferenceValue:
+			if !(ref.BorrowedType.Equal(unwrappedTargetType.Type) && ref.Authorization.Equal(targetAuthorization)) {
```
SupunS, line 2156: Instead of `valueType` type (`ReferenceStaticType`) which is always the same as the type of the referenced value, use the "borrow type" of the value to determine whether this value should be converted to the target type or not.

runtime/tests/interpreter/attachments_test.go
```diff
@@ -1989,107 +1988,6 @@ func TestInterpretAttachmentSelfAccessMembers(t *testing.T) {
 	require.NoError(t, err)
 }
 
-func TestInterpretAttachmentMappedMembers(t *testing.T) {
```
SupunS, line 1992: This run-time test is failing, which I'm not sure why. There is a corresponding checker test, so this is mostly a second layer of defence.

runtime/interpreter/interpreter.go
dsainati1: Should we also not convert if the `valueType` is different, even if the `borrowedType` is the same? E.g. if we have two references that both point to an underlying `R` value, but one has an `AnyResource` borrow type and the other one is `R`? Why don't we need to convert in that case?

runtime/tests/interpreter/attachments_test.go
dsainati1: These previously failed at runtime with an invalid transfer error, which I think will no longer occur because the value being transferred is now being converted at runtime with these changes. So this removal makes sense to me.

runtime/interpreter/interpreter.go
SupunS: That's the thing, `valueType` is no longer the borrow type. SO in your example, even if the borrow types are different, `valueType` remains the same.

We can still keep the old check just to be safe.

dsainati1: Yeah, I think we should convert if any of {`BorrowedType`, `Authorization`, or `valueType`} are different

runtime/interpreter/interpreter.go
```diff
@@ -2150,46 +2150,64 @@ func (interpreter *Interpreter) convert(value Value, valueType, targetType sema.
...
+	valueType sema.Type,
+	unwrappedTargetType *sema.ReferenceType,
+	targetAuthorization Authorization,
+) bool {
+	if !valueType.Equal(unwrappedTargetType) {
+		return true
+	}
+
+	return !ref.BorrowType().Equal(unwrappedTargetType.Type) ||
+		!ref.GetAuthorization().Equal(targetAuthorization)
```
dsainati1, line 2198:
```suggestion
	return !valueType.Equal(unwrappedTargetType) ||
	!ref.BorrowType().Equal(unwrappedTargetType.Type) ||
		!ref.GetAuthorization().Equal(targetAuthorization)
```

dsainati1 (approved):

runtime/interpreter/interpreter.go
SupunS: I kept these two separate for better readability, particularly because they check two different things..

turbolent:
@SupunS Could you please add a description of what the PR does and why? I tried to have a look at the PR, but I don't quite see what the changes are and why they're needed (what the problem is)

SupunS:
@turbolent  https://github.com/onflow/cadence/pull/3213#discussion_r1548741022 has a description on the change. I put it as a comment, because then it is easier to point to where the actual fix is. I'll copy this over to the description as well.

SupunS:
Updated the description

runtime/interpreter/interpreter.go
```diff
@@ -2150,46 +2150,64 @@ func (interpreter *Interpreter) convert(value Value, valueType, targetType sema.
...
-			if _, isMappedAuth := unwrappedTargetType.Authorization.(*sema.EntitlementMapAccess); isMappedAuth {
-				panic(UnexpectedMappedEntitlementError{
-					Type:          unwrappedTargetType,
-					LocationRange: locationRange,
-				})
-			}
-
-			switch ref := value.(type) {
-			case *EphemeralReferenceValue:
+		targetAuthorization := ConvertSemaAccessToStaticAuthorization(interpreter, unwrappedTargetType.Authorization)
```
turbolent, line 2153: Is there a way to delay this as long as needed and only compute it if needed?

turbolent (approved):
It somewhat makes sense to me what was done before and how it needs to be adjusted for Cadence 1.0 👍 

The TODO/Skip is left because we still need to investigate why the test fails, but still want to unblock #3079 and get that in before?

runtime/interpreter/interpreter.go
SupunS: this converted `targetAuthorization` is needed to perform the check

SupunS:
> The TODO/Skip is left because we still need to investigate why the test fails, but still want to unblock https://github.com/onflow/cadence/pull/3079 and get that in before?

Yes. I already figured out the cause, but needed to figure out a solution. It probably needs some extra work, so don't want to block the remaining fixes until we figure that one out.

turbolent (approved):

PR #3225, opened 2024-04-05
Handle missing slabs caused by dictionary keyed by enums bug 

### LINKED ISSUE #3220: Storage migration encounters non-existent slabs
## Description

Both the atree storage health check that is run on the state **before** the migration, and the migration itself, encounter slabs that reference other slabs that cannot be found.

For example:

```
ERR failed to run StorageMigration in account c843c1f5a4805c3a, domain contract, key EnumTest: slab (0x0.34) not found: slab not found for stored value
goroutine 1786 [running]:
runtime/debug.Stack()
	/opt/homebrew/Cellar/go@1.20/1.20.14/libexec/src/runtime/debug/stack.go:24 +0x65
github.com/onflow/cadence/migrations.(*StorageMigration).MigrateNestedValue.func1()
	/Users/bastian/go/pkg/mod/github.com/onflow/cadence@v1.0.0-preview.14/migrations/migration.go:178 +0xd4
panic({0x495520, 0x1337dfd11d0})
	/opt/homebrew/Cellar/go@1.20/1.20.14/libexec/src/runtime/panic.go:884 +0x213
github.com/onflow/cadence/runtime/interpreter.DictionaryIterator.Next({0xf8b0c4f128?}, {0x0, 0x0})
	/Users/bastian/go/pkg/mod/github.com/onflow/cadence@v1.0.0-preview.14/runtime/interpreter/value.go:18808 +0x105
github.com/onflow/cadence/migrations.(*StorageMigration).MigrateNestedValue(0xe23d837b50, {{0x66954e, 0x8}, {0xc8, 0x43, 0xc1, 0xf5, 0xa4, 0x80, 0x5c, ...}}, ...)
```

Another example is [`/storage/FanTopPermission` in `0x5d63c34d7f05e5a4`](https://f.dnz.dev/0x5d63c34d7f05e5a4/raw/storage/FanTopPermission)

In particular, the referenced non-existent slabs all seem to have address 0x0, i.e. the temporary/"stack" account.

While investigating the issue, this seems to be a result of bug fixed in #1565.

## Suggested Solution

Detect such broken values, and if they are dictionaries keyed by an enum, "repair" them, by skipping broken entries when iterating over dictionaries in the storage migration.

turbolent:
Closes #3220 

## Description

- Add a new tool that a allows converting files in flow-go's "payloads file format" to CSV. This is mainly useful for writing tests, and avoids having to import flow-go to read such files
- Allow skipping resource loss checks in the interpreter. This reduces unnecessary overhead, and the need for the contract code of the migrated value to type-check.
  We should enable this in the interpreter(s) used for the migration.
- Improve the migration code to detect the case where missing slabs are encountered while migrating a dictionary that is keyed by enums. 
  In that case, "repair" the state by replaing the dictionary with a new empty one, which has the same owner and type.
  

Best viewed with whitespace ignored.

TODO:

- [ ] `atree.SlabNotFoundError.storageID` is currently not exported, so we need to consider the error message.
  Export the field in atree. cc @fxamacker 
- [ ] The removal of the broken dictionary / ordered map and all of its slab is incomplete. 
  Currently, the deep removal of the dictionary will also encounter the missing slabs. Ignoring the error is insufficient, as `PopIterate` will stop iteration over the remaining slabs of the ordered map, leaving unreferenced slabs in the state. 
  Ideally, atree would have means to clean/remove an ordered map, even if it encounters errors like missing slabs. cc @fxamacker 

______


turbolent:
Given that not only the Cadence 1.0 migration, but also the atree register inlining migration will encounter this problem, we might want to rework this PR so the recovery is performed on a "lower level", i.e. not on a Cadence level, but on atree level

PR #3226, opened 2024-04-05
Check resource loss in all variable assignments

turbolent:
## Description

Instead of only checking for resource loss in `identifierExpressionGetterSetter`, check in all cases where the value of a variable is updated, and an existing value is dropped.

______

### DIFF
--- a/runtime/interpreter/variable.go
+++ b/runtime/interpreter/variable.go
@@ -34,7 +34,7 @@ func (v *Variable) GetValue() Value {
 }
 
 func (v *Variable) SetValue(interpreter *Interpreter, locationRange LocationRange, value Value) {
-	existingValue := v.value
+	existingValue := v.GetValue()
 	if existingValue != nil {
 		interpreter.checkResourceLoss(existingValue, locationRange)
 	}
--- a/runtime/interpreter/variable.go
+++ b/runtime/interpreter/variable.go
@@ -33,7 +33,11 @@ func (v *Variable) GetValue() Value {
 	return v.value
 }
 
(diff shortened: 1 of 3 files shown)


runtime/interpreter/variable.go
```diff
@@ -33,7 +33,11 @@ func (v *Variable) GetValue() Value {
 	return v.value
 }
 
-func (v *Variable) SetValue(value Value) {
+func (v *Variable) SetValue(interpreter *Interpreter, locationRange LocationRange, value Value) {
+	existingValue := v.value
```
SupunS, line 37: should we maybe call `GetValue()` here? Not sure if this is practically possible, but in theory, `v.value` could be `nil` if the value wasn't accessed before (`v.getter()` wasn't called), right?

turbolent: I initially called `GetValue` but that caused an infinite recursion for lazy variables (e.g. globals), so I felt like just using/checking the "materialized" value is OK, and not eagerly computing the value even though it was never requested.

Even though, I looked into it a bit more and changed it in [6161745](https://github.com/onflow/cadence/pull/3226/commits/6161745fc47c07d04379477e1d2004481a40f747). The infinite recursion is for the special case of contract values: They are declared as lazy variables (have a getter), but the getter itself needs to set the variable during the getting (see comment), because the code initializing the contract variable might already refer to it. I replaced the `SetValue` call with a manual reset of the getter and setting of the variable value.

SupunS (approved):

dsainati1 (approved):

PR #3235, opened 2024-04-09
Introduce special self-variable

SupunS:
Closes https://github.com/dapperlabs/cadence-internal/issues/192

## Description

Core idea:
- Introduce a special variable for `self`, which not only holds the value, but also an explicitly created reference to the value.
- This explicitly created reference represents the implicit reference behaviour of self-values. It is used for reference tracking.
- When the value is accessed, always check the reference for any invalidations. This is similar to what is done in bound functions.

Implementation details:
- Introduced an interface for `Variable`. Rename the existing implementation to `SimpleVariable`.
- Add `interpreter` as a parameter for `Variable.GetValue()` method. Majority of the changes in the PR are related to this change (updating tests, etc.).
- Transactions can also become a `self` value. However, creating a reference to them is not possible (transactions don't have a type associated with them) / not useful (they cannot be moved around). So had to special case to not to create the explicit references for transaction values.

______

### DIFF
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -923,7 +923,7 @@ func (interpreter *Interpreter) declareVariable(identifier string, value Value) 
 }
 
 // declareSelfVariable declares a special "self" variable in the latest scope
-func (interpreter *Interpreter) declareSelfVariable(value Value) Variable {
+func (interpreter *Interpreter) declareSelfVariable(value Value, locationRange LocationRange) Variable {
 	identifier := sema.SelfIdentifier
 
 	// If the self variable is already a reference (e.g: in attachments),
@@ -940,11 +940,10 @@ func (interpreter *Interpreter) declareSelfVariable(value Value) Variable {
 	}
 
 	// NOTE: semantic analysis already checked possible invalid redeclaration
-	variable := NewSelfVariableWithValue(interpreter, value)
+	variable := NewSelfVariableWithValue(interpreter, value, locationRange)
 	interpreter.setVariable(identifier, variable)
 
-	// TODO: add proper location info
-	interpreter.startResourceTracking(value, variable, identifier, nil)
+	interpreter.startResourceTracking(value, variable, identifier, locationRange)
 
 	return variable
 }
@@ -1025,6 +1024,11 @@ func (interpreter *Interpreter) evaluateDefaultDestroyEvent(
 	declarationInterpreter.activations.PushNewWithParent(declarationActivation)
 	defer declarationInterpreter.activations.Pop()
 
+	locationRange := LocationRange{
+		Location:    interpreter.Location,
+		HasPosition: eventDecl,
+	}
+
 	var self MemberAccessibleValue = containingResourceComposite
 	if containingResourceComposite.Kind == common.CompositeKindAttachment {
 		var base *EphemeralReferenceValue
@@ -1035,14 +1039,11 @@ func (interpreter *Interpreter) evaluateDefaultDestroyEvent(
 		}
 		supportedEntitlements := entitlementSupportingType.SupportedEntitlements()
 		access := sema.NewAccessFromEntitlementSet(supportedEntitlements, sema.Conjunction)
-		locationRange := LocationRange{
-			Location:    interpreter.Location,
-			HasPosition: eventDecl,
-		}
+
 		base, self = attachmentBaseAndSelfValues(declarationInterpreter, access, containingResourceComposite, locationRange)
 		declarationInterpreter.declareVariable(sema.BaseIdentifier, base)
 	}
-	declarationInterpreter.declareSelfVariable(self)
+	declarationInterpreter.declareSelfVariable(self, locationRange)
 
 	for _, parameter := range parameters {
 		// "lazily" evaluate the default argument expressions.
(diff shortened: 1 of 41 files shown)


runtime/runtime_test.go
```diff
@@ -10207,9 +10207,6 @@ func TestRuntimeIfLetElseBranchConfusion(t *testing.T) {
 
 func TestResourceLossViaSelfRugPull(t *testing.T) {
 
-	// TODO: Disabled temporarily
-	t.SkipNow()
-
```
SupunS, line 10212: This test now passes

runtime/interpreter/variable.go
```diff
@@ -44,16 +65,59 @@ func (v *Variable) SetValue(interpreter *Interpreter, locationRange LocationRang
...
+
+	// Create an explicit reference to represent the implicit reference behavior of 'self' value.
+	// Authorization doesn't matter, we just need a reference to add to tracking.
+	selfRef := NewEphemeralReferenceValue(interpreter, UnauthorizedAccess, value, semaType, EmptyLocationRange)
+
+	return &SelfVariable{
+		value:   value,
+		selfRef: selfRef,
+	}
+}
```
SupunS, line 102: The actual fix is to introduce this special self variable, which internally hold an explicit reference to the actual value.
This reference is tracked and gets updated via usual reference tracking.
Then in the getter below, check for the reference validity before returning the value.

runtime/interpreter/simplecompositevalue.go
```diff
@@ -39,6 +39,8 @@ type SimpleCompositeValue struct {
 	TypeID   sema.TypeID
 	// FieldNames are the names of the field members (i.e. not functions, and not computed fields), in order
 	FieldNames []string
+
+	isTransaction bool
```
SupunS, line 43: This is an unfortunate side effect. Needed a way to distinguish between transaction values and other composite values, because references cannot be created for transaction values, as they don't have a valid runtime type.
This doesn't need to be encoded/decoded, since transaction values doesn't get stored.

If there are any suggestions for a better way to do this, would love to hear.

runtime/interpreter/interpreter.go
```diff
@@ -921,6 +922,33 @@ func (interpreter *Interpreter) declareVariable(identifier string, value Value)
 	return variable
 }
 
+// declareSelfVariable declares a special "self" variable in the latest scope
+func (interpreter *Interpreter) declareSelfVariable(value Value) Variable {
```
SupunS, line 926: Use this method in places where the self variable get declared

dsainati1: I think there are a couple places this PR misses where `self` is created that probably should be updated to use this function:

* https://github.com/onflow/cadence/blob/2bedb730926930bc77970bde22c13079e79d11c3/runtime/interpreter/value.go#L18276 where we create the `self` value in an attachment function.
* https://github.com/onflow/cadence/blob/2bedb730926930bc77970bde22c13079e79d11c3/runtime/interpreter/value_function.go#L355 where bound functions create `self` values

SupunS: For attachments, `self` is already a reference, so we don't really need to create another reference. We can use `declareSelfVariable` there as well, but it will internally calls the normal `declareVariable` function (because it is already a reference).

For bound functions, that place is where the 'explicit reference' for the receiver gets created for reference tracking (i.e: the existing solution for bound functions, on which this solution is also based-on).

dsainati1: Ah, okay nice.

turbolent (approved):
Great work fixing this! 👏👏👏

dsainati1 (approved):

runtime/interpreter/interpreter.go
```diff
@@ -921,6 +922,33 @@ func (interpreter *Interpreter) declareVariable(identifier string, value Value)
...
+		if value.isTransaction {
+			return interpreter.declareVariable(identifier, value)
+		}
+	}
+
+	// NOTE: semantic analysis already checked possible invalid redeclaration
+	variable := NewSelfVariableWithValue(interpreter, value)
+	interpreter.setVariable(identifier, variable)
+
+	// TODO: add proper location info
```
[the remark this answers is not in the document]
SupunS: this is coming from the existing `declareVariable` method: https://github.com/onflow/cadence/blob/6187b93364d9047a9d9e8464f9ec3c520588600c/runtime/interpreter/interpreter.go#L919-L920

I tried passing proper location (`ast.HasPosition`), but this method is being called from places where `ast.HasPosition` is not available. Probably why it was left like that in the original method as well.

runtime/interpreter/simplecompositevalue.go
SupunS: added https://github.com/onflow/cadence/pull/3235/commits/10649294381ac284df0625639c5f0f2694feba99

runtime/interpreter/interpreter.go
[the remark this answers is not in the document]
SupunS: Fixed in https://github.com/onflow/cadence/pull/3235/commits/c627cdab48c89324f470165bb2da6d7214f146e9

runtime/interpreter/variable.go
```diff
@@ -44,16 +65,59 @@ func (v *Variable) SetValue(interpreter *Interpreter, locationRange LocationRang
...
+var _ Variable = &SelfVariable{}
+
+func NewSelfVariableWithValue(interpreter *Interpreter, value Value) Variable {
+	common.UseMemory(interpreter, variableMemoryUsage)
+
+	semaType := interpreter.MustSemaTypeOfValue(value)
+
+	// Create an explicit reference to represent the implicit reference behavior of 'self' value.
+	// Authorization doesn't matter, we just need a reference to add to tracking.
+	selfRef := NewEphemeralReferenceValue(interpreter, UnauthorizedAccess, value, semaType, EmptyLocationRange)
```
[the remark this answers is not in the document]
SupunS: Fixed in https://github.com/onflow/cadence/pull/3235/commits/c627cdab48c89324f470165bb2da6d7214f146e9

PR #3236, opened 2024-04-10
Clear `inAssignment` when recursing into subexpressions

dsainati1:
This was not properly being cleared, meaning accesses nested inside assignments were able to escalate entitlements

______

### DIFF
--- a/runtime/sema/check_assignment.go
+++ b/runtime/sema/check_assignment.go
@@ -268,42 +268,45 @@ func (checker *Checker) accessedSelfMember(expression ast.Expression) *Member {
 	return member
 }
 
-func (checker *Checker) visitAssignmentValueType(
-	targetExpression ast.Expression,
-) (targetType Type) {
-
+func (checker *Checker) withAssignment(b bool, f func() Type) Type {
 	inAssignment := checker.inAssignment
-	checker.inAssignment = true
+	checker.inAssignment = b
 	defer func() {
 		checker.inAssignment = inAssignment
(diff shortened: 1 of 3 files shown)


turbolent (approved):
Nice!

runtime/sema/check_assignment.go
```diff
@@ -344,7 +347,12 @@ func (checker *Checker) visitIndexExpressionAssignment(
 	indexExpression *ast.IndexExpression,
 ) (elementType Type) {
 
-	elementType = checker.visitIndexExpression(indexExpression, true)
+	// in an statement like `ref.foo[i] = x`, the entire statement itself
+	// is an assignment, but the evaluation of the index expression itself (i.e. `ref.foo`)
+	// is not, so we temporarily clear the `inAssignment` status here before restoring it later.
+	elementType = checker.withAssignment(false, func() Type {
+		return checker.visitIndexExpression(indexExpression, true)
+	})
```
SupunS, line 355: For the member expressions, the same check is done in the general method `visitMember` (not in the assignment specific method `visitMemberExpressionAssignment`). That makes sense, since it will handle all cases.

Similarly here, wouldn't it be better to do this change in the `visitIndexExpression` generic method, so it'l handle for all cases, similar to member-expressions?
Also, I think here what needs wrapping is the visiting of `indexExpression.TargetExpression`, not the `indexExpression` itself, right?

SupunS (approved):

PR #3251, opened 2024-04-17
Improve supported entitlements

turbolent:
## Description

- `SupportedEntitlements()` may return `nil`, e.g. by `OptionalType`. Handle the `nil` return value of all calls (some already handled it)
- Avoid allocating empty maps if not necessary

______

### DIFF
--- a/runtime/sema/check_composite_declaration.go
+++ b/runtime/sema/check_composite_declaration.go
@@ -72,16 +72,21 @@ func (checker *Checker) checkAttachmentMemberAccess(
 	baseType Type,
 	supportedBaseEntitlements *EntitlementOrderedSet,
 ) {
-	var requestedEntitlements *EntitlementOrderedSet = &orderedmap.OrderedMap[*EntitlementType, struct{}]{}
+	var requestedEntitlements *EntitlementOrderedSet
+
 	switch memberAccess := member.Access.(type) {
 	case EntitlementSetAccess:
 		requestedEntitlements = memberAccess.Entitlements
+
 	// if the attachment field/function is declared with mapped access, the domain of the map must be a
 	// subset of the supported entitlements on the base. This is because the attachment reference
 	// will never be able to possess any entitlements other than these, so any map relations that map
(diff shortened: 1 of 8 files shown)


dsainati1 (approved):

runtime/sema/check_composite_declaration.go
```diff
@@ -2074,9 +2081,21 @@ func (checker *Checker) checkDefaultDestroyEventParam(
...
 	// make `self` and `base` available when checking default arguments so the fields of the composite are available
 	// as this event is emitted when the resource is destroyed, these values should be fully entitled
-	fullyEntitledAccess := NewAccessFromEntitlementSet(containerType.SupportedEntitlements(), Conjunction)
-	checker.declareSelfValue(fullyEntitledAccess, containerType, containerDeclaration.DeclarationDocString())
-	if compositeContainer, isComposite := containerType.(*CompositeType); isComposite && compositeContainer.Kind == common.CompositeKindAttachment {
+	supportedEntitlements := containerType.SupportedEntitlements()
+	fullyEntitledAccess := UnauthorizedAccess
+	if supportedEntitlements != nil && supportedEntitlements.Len() > 0 {
+		fullyEntitledAccess = NewAccessFromEntitlementSet(supportedEntitlements, Conjunction)
+	}
```
SupunS, line 2088: Seems this pattern is/has-to-be used every time the constructor `NewAccessFromEntitlementSet` is called. Would it makes sense to move this logic to the `NewAccessFromEntitlementSet` itself?

turbolent: Mostly just tried to make the code handle all cases and make it consistent, leaving existing cases as-is, but yeah, we can also simplify the case where `NewAccessFromEntitlementSet` is called. It already handles `Len() == 0`, but we can also add the nil check there.

turbolent: @SupunS Pushed a few commits to improve this

SupunS (approved):

PR #3266, opened 2024-04-18
Improve supported entitlements and entitlements inferrence in migration

### LINKED ISSUE #3253: Entitlement inference in entitlement migration grants too many entitlements
[Discord thread](https://discord.com/channels/613813861610684416/621847426201944074/1229833883340767233 )

context:
- Entitlement migration pulls all entitlements that a migrated type needs
- Disjunction access (e.g function needs access(Owner | Deposit)) is translated to conjunction authorization (auth(Owner, Deposit))

This is high priority, because we want to migrate capabilities with as few entitlements as possible, to minimize the chances of the migration causing safety issues for the migrated contracts. It is difficult to determine subset of entitlements actually needed.
  - Potential solutions:
    - Improve entitlement inference in entitlement migration in general 
    - Don't include Owner entitlement in NFT. But other developers' contracts might have similar problem
    - Special case NFT in entitlements migration. But other developers' contracts might have similar problem
 
 Impact: NFTs, majority of contracts

turbolent:
Closes #3253

## Description

- Introduce a new Go type `EntitlementSet` (maybe needs a more descriptive name, suggestions very welcome), which is a conjunction of entitlement types, and disjunctions of entitlements. For example, it can represent `{A, B, (C | D)}`. This allows representing more complex entitlement sets than access/auth entitlement sets can (only conjunction or disjunction of entitlements, e.g. `{A, B}` OR `{A | B}`)
- Switch `EntitlementSupportingType.SupportedEntitlements` to return `EntitlementSet`. 

  Implementations of the function combine sub-sets – for example, composite type's implementation merges the entitlement sets of all members, and intersection type's implementation merges the entitlement sets of all interface types' supported entitlements. Switching the return type to a set type which is able to represent more complex sets results in less "information loss" in recursive calls. 

  As a result, when translating the resulting entitlement set to an access, it is possible to get a disjunction now, which was not previously possible.

Given `EntitlementSupportingType.SupportedEntitlements` is used in the entitlements migration and the contract update checker to infer the entitlements/auth, these changes should hopefully result in authorizations that are  less permissive than before.

______

### DIFF
--- a/runtime/sema/check_composite_declaration.go
+++ b/runtime/sema/check_composite_declaration.go
@@ -107,20 +107,35 @@ func (checker *Checker) checkAttachmentMembersAccess(attachmentType *CompositeTy
 	var supportedBaseEntitlements *EntitlementOrderedSet
 	baseType := attachmentType.GetBaseType()
 	if base, ok := attachmentType.GetBaseType().(EntitlementSupportingType); ok {
-		supportedBaseEntitlements = base.SupportedEntitlements()
+		// TODO:
+		access := base.SupportedEntitlements().Access()
+		if access, ok := access.(EntitlementSetAccess); ok {
+			supportedBaseEntitlements = access.Entitlements
+		}
 	}
 	if supportedBaseEntitlements == nil {
 		supportedBaseEntitlements = &orderedmap.OrderedMap[*EntitlementType, struct{}]{}
 	}
 
-	attachmentType.EffectiveInterfaceConformanceSet().ForEach(func(intf *InterfaceType) {
-		intf.Members.Foreach(func(_ string, member *Member) {
-			checker.checkAttachmentMemberAccess(attachmentType, member, baseType, supportedBaseEntitlements)
+	attachmentType.EffectiveInterfaceConformanceSet().
+		ForEach(func(interfaceType *InterfaceType) {
+			interfaceType.Members.Foreach(func(_ string, member *Member) {
+				checker.checkAttachmentMemberAccess(
+					attachmentType,
+					member,
+					baseType,
+					supportedBaseEntitlements,
+				)
+			})
 		})
-	})
 
 	attachmentType.Members.Foreach(func(_ string, member *Member) {
-		checker.checkAttachmentMemberAccess(attachmentType, member, baseType, supportedBaseEntitlements)
+		checker.checkAttachmentMemberAccess(
+			attachmentType,
+			member,
+			baseType,
+			supportedBaseEntitlements,
+		)
 	})
 
 }
@@ -2081,7 +2096,7 @@ func (checker *Checker) checkDefaultDestroyEventParam(
 
 	// make `self` and `base` available when checking default arguments so the fields of the composite are available
 	// as this event is emitted when the resource is destroyed, these values should be fully entitled
-	fullyEntitledAccess := NewAccessFromEntitlementSet(containerType.SupportedEntitlements(), Conjunction)
+	fullyEntitledAccess := containerType.SupportedEntitlements().Access()
 
 	checker.declareSelfValue(
 		fullyEntitledAccess,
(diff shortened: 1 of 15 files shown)


runtime/sema/entitlementset.go
```diff
@@ -0,0 +1,191 @@
...
+	if other.Disjunctions != nil {
+		other.Disjunctions.
+			Foreach(func(_ string, disjunction *EntitlementOrderedSet) {
+				s.AddDisjunction(disjunction)
+			})
+	}
+}
+
+// Minimize minimizes the entitlement set.
+// It removes disjunctions that contain entitlements
```
dsainati1, line 122: How much extra work would it be to remove "redundant" disjunctions as well? I.e., if we have an `EntitlementSet` that contains `(A | B)` and `(A | B | C)`, the minimal representation of this set would be `(A | B)`.

runtime/sema/entitlementset.go
```diff
@@ -0,0 +1,191 @@
...
+		})
+	}
+
+	if entitlements == nil {
+		return UnauthorizedAccess
+	}
+
+	return EntitlementSetAccess{
+		Entitlements: entitlements,
+		SetKind:      Conjunction,
```
dsainati1, line 189: To check: the behavior of this new set should be exactly the same as the old behavior except in the case where there are no conjunctions and only one disjunction? I.e. some composite type with functions `foo` and `bar` with `access(A)` and `access(B | C)` would still return `(A, B, C)` for its supported entitlements?

If we are committed to this behavior IMO we should provide some kind of warning to users when this happens, so that they are aware that their staged contract may be vulnerable to privilege escalation as a result of this behavior.

runtime/sema/check_composite_declaration.go
```diff
@@ -107,20 +107,35 @@ func (checker *Checker) checkAttachmentMembersAccess(attachmentType *CompositeTy
 	var supportedBaseEntitlements *EntitlementOrderedSet
 	baseType := attachmentType.GetBaseType()
 	if base, ok := attachmentType.GetBaseType().(EntitlementSupportingType); ok {
-		supportedBaseEntitlements = base.SupportedEntitlements()
+		// TODO:
```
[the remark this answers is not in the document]
turbolent: I wasn't quite sure how `supportedBaseEntitlements` is used after. If convering to an access and then just a full set of entitlements (basically what was done before) is sufficient, then I'll remove the TODO

runtime/sema/entitlementset.go
turbolent: I mainly avoided that (and also minimizing on all operations except the final access conversion) for performance reasons, as e.g. such a minimization would grow with the number of entitlements, disjunctions, and entitlements in the disjunctions. We can maybe add it for "reasonable simple" cases.

runtime/sema/entitlementset.go
turbolent: > the behavior of this new set should be exactly the same as the old behavior except in the case where there are no conjunctions and only one disjunction?

Correct. It's not a big improvement, but at least improves things a bit. This particular case/set can actually be the result of multiple different situations in the program, e.g. `access(A | B)` occurring on multiple members of a composite, or interface types of an intersection.

> some composite type with functions `foo` and `bar` with `access(A)` and `access(B | C)` would still return `(A, B, C)` for its supported entitlements?

Correct, there is nothing we can do in such a case, because we cannot represent it in an access/auth.
IDK if it's safe and if we want to minimize disjunctions further, e.g. `(A | B), (B | C)` into `B`.

> If we are committed to this behavior IMO we should provide some kind of warning to users when this happens, so that they are aware that their staged contract may be vulnerable to privilege escalation as a result of this behavior.

Yes, we should do this in any case, with this PR or not: Developers should check that the entitlement migration does not grant more than needed.

turbolent:
@onflow/cadence PTAL 🙏

SupunS (approved):

runtime/sema/type.go
```diff
@@ -4957,27 +4956,69 @@ func (t *CompositeType) MemberMap() *StringMemberOrderedMap {
...
-			case EntitlementSetAccess:
-				set.SetAll(access.Entitlements)
+// TODO: improve naming
+func newCompositeOrInterfaceSupportedEntitlementSet(
+	members *StringMemberOrderedMap,
+	effectiveInterfaceConformanceSet *InterfaceSet,
+) *EntitlementSet {
+	set := &EntitlementSet{}
+
+	// First pass: Handle maps and conjunctions
```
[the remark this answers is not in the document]
turbolent: Good point, added an explanation in 8f666bb6e8fb6551d7d52a04e710a47eb1e9b913

Issue #3248, opened 2024-04-16
Optional Reference Types and References to Optional Types are represented with the same TypeID


dsainati1:
### Current Behavior

There is currently an ambiguity in the creation of type IDs between references to optionals and optional references. I.e., a `ReferenceType(OptionalType(T))` will be rendered as `&T?`, and `OptionalType(ReferenceType(T))` is also rendered as `&T?`. 

### Expected Behavior

We should produce non-ambiguous TypeIDs for these two cases.

TODOs:
  * What should the representation for these be?
  * How and when should we migrate existing TypeIDs to the new format?
  * Check whether or not the typeIDs of these values are stored here

### Steps To Reproduce

N/A

### Environment

```markdown
See https://github.com/onflow/cadence/pull/3232
```

dsainati1:
This blocks https://github.com/onflow/cadence/pull/3232

PR #3272, opened 2024-04-19
Fix optional type ID ambiguity and migrate types

dsainati1:
Closes https://github.com/onflow/cadence/issues/3248
______


runtime/sema/type.go
```diff
@@ -693,7 +693,7 @@ func (t *OptionalType) QualifiedString() string {
 }
 
 func FormatOptionalTypeID[T ~string](elementTypeID T) T {
-	return T(fmt.Sprintf("%s?", elementTypeID))
+	return T(fmt.Sprintf("(%s)?", elementTypeID))
```
turbolent, line 696: Maybe we want to safe some space here and not have a 3 byte overhead for an optional type. 
Would a "prefix operator" variant be sufficient (non-ambiguous)?

turbolent (commented):
Nice!

runtime/sema/type.go
dsainati1: Are TypeIDs ever user-visible? If not then I think a prefix operator works but if they are then it's a bit hard to read.

turbolent: They do show up at times, mostly for developers, e.g. in the JSON encoding, so yeah, maybe we want to keep it readable

SupunS (commented):
Can you please also add a "Rehash test", similar to: https://github.com/onflow/cadence/blob/922e65cf4a2e23632f3c3c55810dfbabf2c0ade1/migrations/statictypes/intersection_type_migration_test.go#L481-L484.

The idea is to test, if there is an optional-typed `TypeValue` is being used as a key in a dictionary, then retrieving (using the old hash/typeID) and updating that key during the migration should work.

SupunS:
hmm, yes it does (didn't realize it earlier). However, that test is for testing entitlements rehashing, and the optional-type seems more like a coincident/side-effect? It would be good to have a separate test for optionals, so that any change to the entitlement test won't accidentally remove the use of optional type.

dsainati1:
So while I was writing these tests, I realized that I am not actually sure the migration itself is strictly necessary. We had to migrated restricted/intersection and reference static types because how they are encoded and decoded from storage changed, but with optional types even though the TypeID itself is changing how it's printed, the encoding of these values is not changing. So any optional static types that were created and encoded with the old IDs are just going to be loaded from storage the same way, and there is no migration that actually needs to be performed on the data. I realized this while trying to write the test @SupunS was talking about above because I couldn't actually get the migration to run on the test data, as when the data was loaded from storage to be migrated, it was already "migrated" so to speak, in that it was already converted to the new format by virtue of having been decoded.

Am I misunderstanding something about how this works, or do we just not need to do anything here?

turbolent:
@dsainati1 Do you need help here with the rehash test case? I can finish this up

dsainati1:
If you have time that would be super helpful! Otherwise I can get to it after the security stuff

turbolent:
- Simplified the legacy optional type implementation in fe757c3d2e7f4a9b0f8ef7e86898cb132d0cc150
- Forced re-storing/encoding of optional types, even if the inner type has not changed in 80ede1e322f477ee7bf3a0721354845db99f0f71
- Add a rehashing test in e94d7e080d8a042485bbd250c8c36766a7dd039e. Showed that forcing the re-storage/encoding was still missing

Should be ready now. @SupunS could you PTAL?

migrations/statictypes/statictype_migration_test.go
```diff
@@ -1238,3 +1238,168 @@ func TestCanSkipStaticTypeMigration(t *testing.T) {
...
+		optionalType := newOptionalType()
+		typeValue := interpreter.NewUnmeteredTypeValue(optionalType)
+
+		// NOTE: in *new* format
+		assert.Equal(t,
+			common.TypeID("(Int)?"),
+			optionalType.ID(),
+		)
+
+		value, ok := dictValue.Get(inter, locationRange, typeValue)
```
SupunS, line 1396: Just a nit: maybe also check that the dictionary doesn't have other values (no old entries). e.g: size == 1

SupunS (approved):
LGTM!

migrations/statictypes/statictype_migration.go
```diff
@@ -565,7 +569,8 @@ func CanSkipStaticTypeMigration(valueType interpreter.StaticType) bool {
 		return CanSkipStaticTypeMigration(valueType.ElementType())
 
 	case *interpreter.OptionalStaticType:
-		return CanSkipStaticTypeMigration(valueType.Type)
+		// optional types need to be rewritten as a result of changes to TypeID printing
+		return false
```
turbolent, line 573: Not sure we actually need this change. 

The skipping is based on the _value's type_, and an optional value does not have to be migrated.
Instead, we need to migrate values in which an optional static type appears, for example, in typed capability values (`Capability<Int?>`, capability static type), type values (`Type<Int?>()`, primitive static type meta type, etc.)

migrations/statictypes/statictype_migration_test.go
turbolent: Good idea! Added in 55043809b0945178bc76ea0ed7e9ef608478e64a here and in similar tests

migrations/statictypes/statictype_migration.go
turbolent: @SupunS @dsainati1 I'm not 100% about the above, wdyt? If it makes sense, I'll revert the change

SupunS: makes sense to me. should be OK to revert.

dsainati1: Yeah, same

turbolent: 6c04cd4dc0740fc2d28e27cbe144d154aab5196c

turbolent:
@dsainati1 @SupunS Could you PTAL?

turbolent (approved):

SupunS (approved):

Issue #3265, opened 2024-04-18
Distribute migration reports to a central location/URL


jribbink:
### Issue to be solved

In order for tooling to query the status of the last migration report, there must exist some central server/URL which hosts so that the latest migration report can always be found.

(specifically needed for https://github.com/onflow/flow-cli/issues/1508)

### Suggested Solution

Setup a pattern for uploading these to GCP, GitHub, or other so that they may be repeatably queried by tooling & the latest report will always be used.

jribbink:
I think we can probably just use GitHub for this & put the reports in this folder somewhere:

https://github.com/onflow/cadence/tree/e020944c7f187de8bf9bfc3b3293f32abb05510a/migrations_data

As long as we follow the same pattern each time, the CLI should be able to pick these up.

jribbink:
I've created the PR for the CLI to fetch contract update statuses for the latest migration report https://github.com/onflow/flow-cli/pull/1529.

Currently, the CLI assumes the following:
 - `master` branch of `onflow/cadence`
 - files live in `migrations_data/raw/XXXXX-<unix-timestamp>.json` (naming can be anything as long as it ends with `-<unix-timestamp>.json`)

If an alternate format is preferred, I can update the PR accordingly.  If this is acceptable we can try to merge this Flow CLI PR ASAP.  I can't seem to find the JSON report for the last emulated migration, so I'm unable to make a PR for this myself, but if someone else could do this that would be awesome.

Thanks!

(cc @j1010001 )

jribbink:
closed by #3273

PR #3273, opened 2024-04-22
Adding staged contracts migration output JSON from Apr 17 to migratio…

j1010001:
related to https://github.com/onflow/cadence/issues/3265


jribbink (approved):
LGTM, thanks!

One quick quesstion: at the end of the filename, what does "103000" represent so that I can parse/interpret it correctly.

turbolent (commented):
Small nit: Maybe improve the naming an make it explicit this is the _report_ for the staged contracts, not the staged contracts themselves. We might add them in the future, so maybe avoid the potential confusion.

j1010001:
@jribbink I changed the format slightly - I added the timestamp (10-30) after the date. @turbolent I also added "report" to make it clearer, good suggestion!

migrations_data/staged-contracts-report--04-17-2024-10-30-testnet.json
turbolent, line 1: Not sure if the double dash (`--`) is intentional

jribbink, line 1: Maybe we could change the timestamp format to something a bit more cannonical? It's hard to know exactly what each component means or what timezone this is from.

For example, ISO 8601 (based on the UNIX timestamp I saw in a previous PR):
`staged-contracts-report-2024-04-17T20:05:50.000Z-testnet.json`
and
`staged-contracts-report-2024-04-17T20:05:50.000Z-testnet.md`

It doesn't make a difference on my end (I can parse anything as long as I know what it is), but I think this way is most obvious for somebody else looking at the file.

turbolent:
@jribbink Please go ahead with making improvements as you deem appropriate, e.g. defining the naming scheme and renaming the existing report and the new one here, so you can continue on https://github.com/onflow/flow-cli/pull/1529 🙏

turbolent (approved):

turbolent:
@jribbink The colons in the filenames are causing issues in CI, see https://github.com/onflow/cadence/actions/runs/8820926221/job/24215623175#step:10:10. Could you please remove them?

jribbink:
> @jribbink The colons in the filenames are causing issues in CI, see https://github.com/onflow/cadence/actions/runs/8820926221/job/24215623175#step:10:10. Could you please remove/replace them?

I've made a PR to fix this 👍 https://github.com/onflow/cadence/pull/3281

Issue #3282, opened 2024-04-24
Contract update validator doesn't allow valid type updates for fields


SupunS:
### Current Behavior

There are two issues (in the thread https://discord.com/channels/613813861610684416/1232706046246781000/1232789114303021147):
- [x] Doesn't allow updating from `&{MetadataViews.Resolver} ` to `&{ViewResolver.Resolver}`
  - When the restricted type is not present, validator skips the custom type-change rules.
  - https://github.com/onflow/cadence/pull/3283
- [x] Doesn't allow updating from ` Capability<&AuthAccount> ` to ` Capability<auth(Storage, Contracts, Keys, Inbox, Capabilities) &Account>`
  - Account type changes are not allowed/handled in the validator
  - https://github.com/onflow/cadence/pull/3284

### Expected Behavior

Both of these are safe, and must be allowed.

### Steps To Reproduce

See above

### Environment

```markdown
- Cadence version: 1.0
```

PR #3283, opened 2024-04-24
Fix restricted typed field updates

SupunS:
Closes https://github.com/onflow/cadence/issues/3282

## Description

Previously, if there were no restricted type (e.g: `{MetadataViews.Resolver}`), then the update validator directly checked for type-equality. However, in the restriction set, there can be types that needed to be changed according the the special custom rules. e.g: convert the same to `{ViewResolver.Resolver}`

Solution: Visit and check the interface set using the  `checkTypeUpgradability` (rather than the default type-equality), which also checks for the custom rules.

______


dsainati1 (approved):

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
```diff
@@ -338,11 +338,31 @@ typeSwitch:
...
 		if oldType.LegacyRestrictedType == nil {
-			break
+			foundIntersectionType, ok := newType.(*ast.IntersectionType)
+			if !ok {
+				return newTypeMismatchError(oldType, newType)
+			}
+
+			if len(oldType.Types) != len(foundIntersectionType.Types) {
+				return newTypeMismatchError(oldType, newType)
+			}
```
turbolent, line 351: With interface inheritance, I guess there could be cases where the new set is smaller, but has semantically equal "contents"?  Just a side-note, we do not need to address this here

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
```diff
@@ -338,11 +338,31 @@ typeSwitch:
...
+
+			for index, expectedIntersectedType := range oldType.Types {
+				foundType := foundIntersectionType.Types[index]
```
turbolent, line 354: Does that mean the types have to occur in exactly the same order as before?
Definitely OK, but we might need to help developers here in the case where they are not, because e.g. for a case like `{A, B}` -> `{B, A}`, this would fail (right?), but it might not be clear the reason is the order.

turbolent (approved):
Nice! Thanks for fixing this so quickly 👍

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
SupunS: yeah, for now this only allows changing the known "moved" interfaces. I'll see if we can support reducing too, as long as they are semantically equivalent, as a follow-up PR. We would need to make sure it is consistent to how the migration works.

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
SupunS: ah good point! I copied the logic form the existing type-comparator. But yeah, order shouldn't matter now.

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
turbolent: Yeah, let's see how this works and we can investigate if needed 👍

runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
SupunS: added with https://github.com/onflow/cadence/pull/3283/commits/2d443446cae94474b68472441b9e43e557171d00

turbolent (approved):
Nice! 👏

PR #3284, opened 2024-04-25
Handle account types in contract update validator

SupunS:
Work towards https://github.com/onflow/cadence/issues/3282

## Description

&subject

______

### DIFF
--- a/runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
+++ b/runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
@@ -308,26 +308,83 @@ func (validator *CadenceV042ToV1ContractUpdateValidator) checkEntitlementsUpgrad
 	return nil
 }
 
-func (validator *CadenceV042ToV1ContractUpdateValidator) checkTypeUpgradability(oldType ast.Type, newType ast.Type) error {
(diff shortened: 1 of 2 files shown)


runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go
```diff
@@ -393,37 +450,91 @@ typeSwitch:
...
 	// If the new/old type is non-storable,
 	// then changing the type of this field has no impact to the storage.
-	if isNonStorableType(oldType) || isNonStorableType(newType) {
-		return nil
+	// However, fields having non-storable types inside capabilities are in-fact storable.
+	// So, skip only if the non-storable type is a direct field type.
+	if !inCapability {
+		if isNonStorableType(oldType) || isNonStorableType(newType) {
+			return nil
+		}
```
SupunS, line 537: This is a separate issue that I discovered while fixing the original issue: it is only OK to change to/from a non-storable type (hence the field itself is also non-storable), only if the type was a direct field-type. i.e: if the non-storabe type was residing inside a Capability, then the field is in-fact storable, and changing the type is NOT OK.

dsainati1 (approved):

turbolent (approved):
Great work!

PR #3290, opened 2024-04-26
Unexport fields to prevent access by index

### LINKED ISSUE #2952: Remove access to fields by index
### Issue to be solved

Accessing fields by index (instead of by name) can create dependency on sort order.

For example, issue https://github.com/onflow/cadence/issues/2558:
> Some client programs are accessing fields by index rather than name. This created a dependency on the sort order of fields which surfaced when we switched from JSON-CDC to CCF for encoding events because CCF explicitly sorts as specified in CCF specification for deterministic encodings.

PR https://github.com/onflow/cadence/pull/2559 resolved issue 2558 by updating CCF Codec to make CCF Determinisitic Encoding optional.  It was a stop-gap solution to avoid abruptly breaking compatibility for encoded events.

For exported values, not allowing access to fields by index can help prevent Cadence developers from unintentionally creating dependency on sort order.  This can prevent unintended consequences like blocking deterministic encoding, etc.

However, this issue can be viewed as broader topic than CCF, sort order, or access by field index.

### Current Status

CCF encoding is not currently sorting the fields (in order to be compatible with old programs that access fields by index).

- CCF specs ([onflow/ccf](https://github.com/onflow/ccf)) defines how to sort and makes it optional, so each CCF-based format can balance trade-offs by specifying which CCF options they use:
  > CCF defines "Deterministic CCF Encoding Requirements" and makes it optional. 

- CCF codec (onflow/cadence) offers sorting as non-default option, but sorting isn't being used (e.g. for events encoding) due to compatibility reasons mentioned in:
  - #2558
  - #2559
  
I didn't find any programs using the sort option in CCF codec (as of Nov 28, 2023).

### Suggested Solution

Programming languages before 1.0 can benefit from taking away the ability for programmers to (sometimes unintentionally) create dependencies on implementation details of compilers and interpreters.

For example, a few months before Go 1.0 (~12 years ago), https://codereview.appspot.com/5285042/ added a random offset for map iteration and changed Go language specification.  

```diff
Index: doc/go_spec.html
===================================================================
--- a/doc/go_spec.html
+++ b/doc/go_spec.html
@@ -4085,7 +4085,8 @@
 </li>
 
 <li>
-The iteration order over maps is not specified.
+The iteration order over maps is not specified
+and is not guaranteed to be the same from one iteration to the next.
 If map entries that have not yet been reached are deleted during iteration,
 the corresponding iteration values will not be produced. If map entries are
 inserted during iteration, the behavior is implementation-dependent, but the
```

That change to Go intentionally broke some existing Go programs by taking away existing behavior that wouldn't be ideal to officially or unofficially support forever after Go 1.0 release.

In the same way, maybe we can determine if some existing features (accessing field by index) should be removed before Cadence 1.0 release.

EDIT: added Current Status section about CCF for more context.

turbolent:
Closes #2952 

## Description

Unexport `cadence` composite value's fields field, to prevent access by index, and require access by field name instead.

______

### DIFF
--- a/encoding/ccf/ccf_test.go
+++ b/encoding/ccf/ccf_test.go
@@ -11698,11 +11698,9 @@ func TestExportRecursiveType(t *testing.T) {
 
 	testEncode(
 		t,
-		cadence.Resource{
-			Fields: []cadence.Value{
-				cadence.Optional{},
-			},
-		}.WithType(ty),
+		cadence.NewResource([]cadence.Value{
+			cadence.Optional{},
+		}).WithType(ty),
 		[]byte{
 			// language=json, format=json-cdc
 			// {"type":"Resource","value":{"id":"S.test.Foo","fields":[{"name":"foo","value":{"type": "Optional","value":null}}]}}
--- a/encoding/ccf/encode.go
+++ b/encoding/ccf/encode.go
@@ -25,6 +25,7 @@ import (
 	goRuntime "runtime"
 	"sort"
 	"sync"
+	_ "unsafe"
 
 	"github.com/fxamacker/cbor/v2"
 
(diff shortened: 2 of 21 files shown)


turbolent:
@fxamacker Do you think we should also unexport type fields?

SupunS (approved):

fxamacker:
@turbolent Amazing work! :+1:  Thank you for implementing this!

> @fxamacker Do you think we should also unexport type fields?

I will look into #2952 and CCF (hopefully today).

fxamacker:
> @fxamacker Do you think we should also unexport type fields?

@turbolent I think it would be more consistent to unexport type fields and provide a new function `FieldTypes()` that returns `map[string]Type`.  This way, there isn't any implied ordering for either field definitions or field values.  It also gives us flexibility to change implementation details without breaking API.

Thoughts?

fxamacker:
@turbolent BTW, it seems `FieldsMappedByName()` includes attachment as a field if present, while `SearchFieldByName()` only looks up field type definition.  Is it OK for these two functions to differ in this way?

PR #3316, opened 2024-05-06
Fix Cadence 1.0 migration of dictionary values when using atree inlined data

fxamacker:
Closes #3288

This PR fixes `MigrateNestedValue()` for dictionary values.

Previously, `MigrateNestedValue()` migrates dictionary by using readonly iterator and migrating values in place.

This PR migrates keys first using read-only iterator and then migrates values using mutable iterator.
______

- [ ] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [ ] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

Issue #3288, opened 2024-04-26
Testnet state migration preview.22 health check after migration error


j1010001:
flow-go build: v0.34.0-crescendo-preview.15-atree-inlining
migration doc: https://www.notion.so/flowfoundation/TN-state-migration-test-Atree-inlining-C1-0-Apr-24-preview-15-71afa024d73042da9c2bc020cb734850?pvs=4

error:
[migration-inlined-payloads-apr24-preview15-check2-output.txt](https://github.com/onflow/cadence/files/15133679/migration-inlined-payloads-apr24-preview15-check2-output.txt)

fxamacker:
@j1010001 status of this is "in-progress" (since yesterday afternoon).
- :thread: https://discord.com/channels/613813861610684416/1234652799380488262/1234654088730378270

fxamacker:
@turbolent I looked into this today and it looks like the problem is in Cadence 1.0 migration's `MigrateNestedValue` function. 

https://github.com/onflow/cadence/blob/e29e0f3802cfdc373cfc0acbb6af24860d06000d/migrations/migration.go#L165-L165

It seems when dictionary element is migrated, `MigrateNestedValue()` returns `nil` as migrated value.  So parent container is unchanged even though nested element is modified.

https://github.com/onflow/cadence/blob/e29e0f3802cfdc373cfc0acbb6af24860d06000d/migrations/migration.go#L363-L365

With testnet data:
- dictionary (`0xa47a2d3a3b7e9133.14553`) is inlined and removed from storage as result of migration
- `MigrateNestedValue()` returns `nil` (as migrated value) for this dictionary even though it is migrated
- parent dictionary (`0xa47a2d3a3b7e9133.16828`) thinks child element isn't migrated (`nil`) so it still retain the old data (slab ID storable  `0xa47a2d3a3b7e9133.14553`) even though the child element should be reset.
- given this, storage health check returns `slab not found` error for `0xa47a2d3a3b7e9133.14553`.

I am going to look into this more tomorrow.  Please let me know if you have some questions.

turbolent:
Thank you for the investigation @fxamacker! 

I was able to reproduce this in flow-go, and started working on reproducer in the Cadence repo here: https://github.com/onflow/cadence/compare/feature/atree-register-inlining-v1.0...bastian/debug-lost-slab-during-migration. So far I'm not able to reproduce it

fxamacker:
[...]

PR #3318, opened 2024-05-07
Improve dictionary migration

turbolent:
Work towards #3288 

## Info

This PR is a result of investigating and fixing #3288, the broken migration of dictionaries in the atree register inlining version, https://github.com/onflow/cadence/tree/feature/atree-register-inlining-v1.0.

Kudos to @fxamacker for investigating the issue, finding the cause, and coming up with this non-intrusive solution! 👏 
This PR mostly just ports #3316 to `master`, so that the atree register inlining feature branch only contains additional changes required for atree register inlining, and the branch does not drift too far off of master.

The PR is best reviewed by viewing each commit individually.

## Description

- 02ebd1eeca6dae80a6611f1fc12ee9fd5006abdd: Fix MigrateNestedValue() for dictionary value

  The original bug was the use of the read-only iterator when migrating dictionary keys and values. Using the read-only iterator is wrong, as the migration might mutate dictionary keys and values.

  However, simply switching from the read-only iterator to the iterator that allows mutations of children is not possible, as the iterator that allows mutations to children does not support reading old dictionary keys that will get a different hash value in Cadence 1.0. 

  @fxamacker realized that we can split the migration of dictionaries from the current "interleaved" approach of iterating over all key-value pairs, into two phases, one that migrates all keys, and then another that migrates all values. 

  - In the first phase which migrates the keys, we can use the read-only iterator, which supports loading old keys which potentially have different hash values in Cadence 1.0. Using a read-only iterator is safe, as keys are immutable anyways, and migrations should never mutate them anyways (see below for further improvements to this)
  - In the second phase which migrates the values, we keep using the mutating iterator

  This was implemented in #3316, but on top of the atree register inlining feature branch. This PR ports the commit, 02ebd1eeca6dae80a6611f1fc12ee9fd5006abdd, to `master`. Even though the change isn't strictly needed on `master`, porting it anyways keeps the approach of how dictionaries are migrated the same across the different branches.

- 412780e4b461673ba7530a1e0c9641557100cdd2: Add comments

  Explain why the migration of dictionaries is now in two phases.
  Also add a note to not attempt to refactor the conflict handling code – I caught myself attempting to "improve" it, just to realize it was working correctly, even after the changes of the previous commit.

- 274cfcee76f71356b6a2dd1a433a51446134a56e: Refactors the dictionary migration phases into separate functions

  The `MigrateNestedValue` function had become very long, so this commit refactors the dictionary case of the switch into two functions. The code got moved as-is and no changes were made to it.

- 76e337cb384c8a27dc19a24598ff0d00c510c534: Adds a test case for migrating dictionaries with enum keys

  We did not previously have a test case for it, so add one to ensure this is still supported.

- 5d2501ae0b9c598e3ead4148f1dd749b45d26655: Prevents mutation of dictionary keys

  #3288 was hard to debug and fix, because the invalid mutation of child values returned from the read-only mutation did not produce any errors. Only the defensive storage health check after the migration identified there was something wrong.

  Improve the migration of nested values by adding assertions that ensure that migrations never mutate dictionary keys. A mutation may replace a dictionary key (e.g. enum with a different enum), but it may never mutate a dictionary key in place (e.g. change the raw value of an enum).

  Add a test that has such an invalid migration, and ensure it is immediately rejected.

- c4bed076a1e3249bdf9f1da6be98818435e5b9f7: Avoid variable shadowing

  Consistently use `newValue` as the name of the variable that is the migrated version of `existingValue`, a child values that is being migrated. Avoid shadowing the final result variable of the whole function, `migratedValue`.

______

### DIFF
--- a/migrations/migration.go
+++ b/migrations/migration.go
@@ -259,6 +259,10 @@ func (m *StorageMigration) MigrateNestedValue(
 				continue
 			}
 
+			// We should only check if we're allowed to mutate if we actually are going to mutate,
+			// i.e. if newValue != nil. It might be the case that none of the values need to be migrated,
+			// in which case we should not panic with an error that we're not allowed to mutate
+
 			if !allowMutation {
 				panic(errors.NewUnexpectedError(
 					"mutation not allowed: attempting to migrate array element at index %d: %s",
@@ -315,6 +319,10 @@ func (m *StorageMigration) MigrateNestedValue(
 			if newValue == nil {
 				continue
 			}
+
+			// We should only check if we're allowed to mutate if we actually are going to mutate,
+			// i.e. if newValue != nil. It might be the case that none of the values need to be migrated,
+			// in which case we should not panic with an error that we're not allowed to mutate
 
 			if !allowMutation {
 				panic(errors.NewUnexpectedError(
@@ -487,6 +495,10 @@ func (m *StorageMigration) migrateDictionaryKeys(
 		if newKey == nil {
 			continue
 		}
+
+		// We should only check if we're allowed to mutate if we actually are going to mutate,
+		// i.e. if newKey != nil. It might be the case that none of the keys need to be migrated,
+		// in which case we should not panic with an error that we're not allowed to mutate
 
 		if !allowMutation {
 			panic(errors.NewUnexpectedError(
@@ -676,6 +688,10 @@ func (m *StorageMigration) migrateDictionaryValues(
 		if newValue == nil {
 			continue
 		}
+
+		// We should only check if we're allowed to mutate if we actually are going to mutate,
+		// i.e. if newValue != nil. It might be the case that none of the values need to be migrated,
+		// in which case we should not panic with an error that we're not allowed to mutate
 
 		if !allowMutation {
 			panic(errors.NewUnexpectedError(
--- a/migrations/migration.go
+++ b/migrations/migration.go
@@ -318,30 +318,22 @@ func (m *StorageMigration) MigrateNestedValue(
 			key, value interpreter.Value
 		}
 
-		// Read the keys first, so the iteration wouldn't be affected
-		// by the modification of the nested values.
-		var existingKeysAndValues []keyValuePair
+		// Migrate keys first.
 
-		iterator := dictionary.Iterator()
+		var existingKeys []interpreter.Value
 
-		for {
-			key, value := iterator.Next(nil)
-			if key == nil {
-				break
-			}
+		dictionary.IterateKeys(
+			inter,
+			func(key interpreter.Value) (resume bool) {
 
-			existingKeysAndValues = append(
-				existingKeysAndValues,
-				keyValuePair{
-					key:   key,
-					value: value,
-				},
-			)
-		}
+				existingKeys = append(existingKeys, key)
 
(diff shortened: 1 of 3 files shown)


migrations/migration.go
```diff
@@ -559,6 +449,262 @@ func (m *StorageMigration) MigrateNestedValue(
...
+		if newKey == nil {
+			continue
+		}
+
+		if !allowMutation {
+			panic(errors.NewUnexpectedError(
+				"mutation not allowed: attempting to migrate dictionary key: %s",
+				existingKey,
+			))
+		}
```
SupunS, line 496: Can take this out of the loop (?)

migrations/migration.go
```diff
@@ -559,6 +449,262 @@ func (m *StorageMigration) MigrateNestedValue(
...
+		if newValue == nil {
+			continue
+		}
+
+		if !allowMutation {
+			panic(errors.NewUnexpectedError(
+				"mutation not allowed: attempting to migrate dictionary value: %s",
+				existingValue,
+			))
+		}
```
SupunS, line 685: same here: could take it out of the loop, since it will report only one error anyway (exits at the first error)

SupunS (approved):

migrations/migration.go
turbolent: We should only check if we're allowed to mutate if we actually are going to mutate, i.e. if `newValue != nil`. It might be the case that none of the values need to be migrated, in which case we should not panic with an error that we're not allowed to mutate

SupunS: ah that makes sense!

turbolent: Great question though! I added comments in 13e8e6b1193b02a172b64e2e021f96568aeaf44a

turbolent:
Added one last small improvement: c4bed076a1e3249bdf9f1da6be98818435e5b9f7: Avoid variable shadowing

Consistently use `newValue` as the name of the variable that is the migrated version of `existingValue`, a child values that is being migrated. Avoid shadowing the final result variable of the whole function, `migratedValue`.

SupunS (approved):

migrations/migration.go
```diff
@@ -559,6 +457,270 @@ func (m *StorageMigration) MigrateNestedValue(
...
+	dictionary.IterateKeys(
+		inter,
+		func(key interpreter.Value) (resume bool) {
+
+			existingKeys = append(existingKeys, key)
+
+			// Continue iteration
+			return true
+		},
+	)
```
fxamacker, line 481: We need to use a different iterator function here for cadence v1.0 + atree inlining migration because `dictionary.IterateKeys()` in branch `feature/atree-register-inlining-v1.0` uses mutable iterators.  

https://github.com/onflow/cadence/blob/4ef654197cf9a73413549349b3ddbd12dd6cccf3/runtime/interpreter/value.go#L18898-L18916

fxamacker (commented):
@turbolent Nice! Thanks for porting #3316 to master and adding mutation checks etc.  

I only left a comment to use a different iterator function for cadence v1.0 + atree inlining migration because `dictionary.IterateKeys()` in branch `feature/atree-register-inlining-v1.0` uses mutable iterators.

PR #3322, opened 2024-05-08
Make migration error stacktrace configurable

SupunS:
## Description

The stacktrace generation upon errors (type loading errors due un-migrated contracts, etc) is consuming a lot of CPU. This not only slows down the migration, but also shadows any actual bottlenecks in migration logic (when profiled). Therefore turn it off by default, and make it configurable so it can be enabled if needed. 

______

### DIFF
--- a/migrations/migration.go
+++ b/migrations/migration.go
@@ -55,6 +55,7 @@ type StorageMigration struct {
 	name                   string
 	address                common.Address
 	dictionaryKeyConflicts int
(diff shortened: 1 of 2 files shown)


turbolent (approved):
Good idea! 👍 

The stack trace was mostly useful at the beginning of development, where we still had a lot of edge cases to handle, but by now we shouldn't have any cases left where it is constantly needed.

PR #3329, opened 2024-05-08
Add/improve caching in static type and entitlements migration

### LINKED ISSUE #3326: Cache results of migrations, e.g. static types, entitlements, etc.

turbolent:
Closes #3326

## Description

- Improve caching in the entitlements migration: Always cache the result, even if the outcome is that no migration of the type is necessary, i.e. the result is `nil`
- Add caching to the static type migration
- Refactor the tests so that individual test cases are run independently.

  There might be "conflicts" between the test cases, in the sense that one test case defines a type `X` as something (e.g. an interface), while another test case defines it as something else (e.g. a composite type).

  When test cases are combined and all run using a single migration, the added caching will result in incorrect results.

Best viewed with whitespace disabled in the diff.

______

### DIFF
--- a/migrations/statictypes/statictype_migration.go
+++ b/migrations/statictypes/statictype_migration.go
@@ -177,19 +177,16 @@ func (m *StaticTypeMigration) maybeConvertStaticType(
 	// Parse of the migration, e.g. the intersection type migration depends on the parent type.
 	// For example, `{Ts}` in `&{Ts}` is migrated differently from `{Ts}`.
 
-	if parentType == nil {
+	migratedTypeCache := m.migratedTypeCache
+	staticTypeID := staticType.ID()
 
-		migratedTypeCache := m.migratedTypeCache
-		staticTypeID := staticType.ID()
-
-		if migratedType, exists := migratedTypeCache.Get(staticTypeID); exists {
-			return migratedType
-		}
-
-		defer func() {
-			migratedTypeCache.Set(staticTypeID, resultType)
-		}()
+	if cachedType, exists := migratedTypeCache.Get(staticTypeID); exists {
+		return cachedType.StaticType
 	}
+
+	defer func() {
+		migratedTypeCache.Set(staticTypeID, resultType, nil)
+	}()
 
 	switch staticType := staticType.(type) {
 	case *interpreter.ConstantSizedStaticType:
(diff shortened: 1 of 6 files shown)


migrations/statictypes/statictype_migration.go
```diff
@@ -154,7 +165,30 @@ func (m *StaticTypeMigration) Migrate(
...
+	staticType interpreter.StaticType,
+	parentType interpreter.StaticType,
+) (
+	resultType interpreter.StaticType,
+) {
+	// Consult the cache and cache the result
+	// at the root of the migration,
+	// i.e. when the parent type is nil.
+
+	if parentType == nil {
```
SupunS, line 178: Any particular reason to read from the cache only if the type is the root? If there are two types, say `[T]` and `{String: T}`, isn't it OK/safe to read `T` from the cache?

SupunS (approved):

turbolent:
Idea: We could maybe extend the static type cache to have the value be a tuple `(StaticType, error)`, instead of just `StaticType`. In the entitlements migration we could then also cache errors.

migrations/statictypes/statictype_migration.go
turbolent: Initially I didn't have this filter in place and got confused why the tests failed.

The reason is that the intersection type migration depends on the parent type, `{Ts}` in `&{Ts}` is migrated differently from `{Ts}`.

I'll extend the comment with that context.

turbolent:
Added the caching of errors in [81c7de9](https://github.com/onflow/cadence/pull/3329/commits/81c7de92fcb44e889768497a6d9bad69d628f565)

dsainati1 (approved):

migrations/statictypes/statictype_migration.go
```diff
@@ -177,20 +177,17 @@ func (m *StaticTypeMigration) maybeConvertStaticType(
 	// Parse of the migration, e.g. the intersection type migration depends on the parent type.
 	// For example, `{Ts}` in `&{Ts}` is migrated differently from `{Ts}`.
 
-	if parentType == nil {
+	migratedTypeCache := m.migratedTypeCache
+	staticTypeID := staticType.ID()
```
SupunS: is this not needed now? 🤔

turbolent: I've added it back in [21f06f7](https://github.com/onflow/cadence/pull/3329/commits/21f06f73fc3d5364d49f0a08064cd79015045780). I can't come up with an example where it is needed, but neither can I convince myself there isn't one. Let's conservatively do it only for the root type, where we can always be sure it is correct. As a side-effect, it might also reduce traffic on and the size of the cache.

SupunS (approved):

PR #3339, opened 2024-05-14
Check dictionary keys when checking the ability for dereferencing

SupunS:
Port https://github.com/dapperlabs/cadence-internal/pull/229

## Description

This is currently inconsistent because enums in other places are not allowed to be dereferenced. Also makes the existing check vulnerable for future mishaps, if some new type (e.g: builtin type) is allowed to be used as a dictionary key in future, and forgets to update this logic in `IsPrimitiveOrContainerOfPrimitive`

______


turbolent (approved):
👍

PR #3340, opened 2024-05-14
Use parent to determine whether an expression is target of an invocation

SupunS:
Port https://github.com/dapperlabs/cadence-internal/pull/220

Closes https://github.com/dapperlabs/cadence-internal/issues/213

Also lays the foundation for https://github.com/onflow/cadence/pull/3341

## Description

Concept is same as https://github.com/dapperlabs/cadence-internal/pull/146, where parent is tracked to determine whether the current expression is an invoked-expression.

However, unlike in https://github.com/dapperlabs/cadence-internal/pull/146, only record the parent temporarily as the checker visit the expression, and then reset it. Doesn't store the parent-child mapping in the elaboration. This way, there is no performance impact:
```
                                                 │   before    │               after               │
                                                 │   sec/op    │   sec/op     vs base              │
CheckContractInterfaceFungibleTokenConformance-8   65.28µ ± 0%   65.58µ ± 0%  +0.46% (p=0.009 n=6)

                                                 │    before    │               after                │
                                                 │     B/op     │     B/op      vs base              │
CheckContractInterfaceFungibleTokenConformance-8   57.77Ki ± 0%   57.78Ki ± 0%  +0.03% (p=0.002 n=6)

                                                 │   before    │             after              │
                                                 │  allocs/op  │  allocs/op   vs base           │
CheckContractInterfaceFungibleTokenConformance-8   1.022k ± 0%   1.022k ± 0%  ~ (p=1.000 n=6) ¹
```

______

### DIFF
--- a/runtime/sema/check_member_expression.go
+++ b/runtime/sema/check_member_expression.go
@@ -168,7 +168,7 @@ func (checker *Checker) visitMember(expression *ast.MemberExpression, isAssignme
 		// is an assignment, but the evaluation of the accessed exprssion itself (i.e. `a.b`)
 		// is not, so we temporarily clear the `inAssignment` status here before restoring it later.
 		accessedType = checker.withAssignment(false, func() Type {
-			return checker.VisitExpression(accessedExpression, nil)
+			return checker.VisitExpression(accessedExpression, expression, nil)
 		})
 	}()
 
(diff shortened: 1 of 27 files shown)


runtime/sema/check_member_expression.go
```diff
@@ -345,16 +345,21 @@ func (checker *Checker) visitMember(expression *ast.MemberExpression, isAssignme
 	//
 	// This would result in a bound method for a resource, which is invalid.
 
-	if !checker.inInvocation &&
```
turbolent, line 348: Can `inInvocation` be removed now?

turbolent (approved):
Great work! 👏 

I wonder if we can also refactor other existing flags, like `inAssignment` (in a follow-up PR)

runtime/sema/check_member_expression.go
SupunS: I tried removing that, but there were some tests failing, and seemed there is still one use-case where it is required to know whether the current expression is inside an invocation, but not necessarily the immediate child. We could maybe revisit that, but would probably need some additional work.

turbolent: Sounds good!

SupunS:
> I wonder if we can also refactor other existing flags, like inAssignment (in a follow-up PR)

Yeah, would be great if we can remove these other flags as well. I'll create an issue for this.

PR #3341, opened 2024-05-14
Improve static validation of contract/transaction moves

SupunS:
Closes https://github.com/dapperlabs/cadence-internal/issues/217

## Description

Depends on https://github.com/onflow/cadence/pull/3340

______


runtime/sema/check_expression.go
```diff
@@ -46,9 +46,50 @@ func (checker *Checker) VisitIdentifierExpression(expression *ast.IdentifierExpr
...
+			if parent.TargetExpression == IdentifierExpression {
+				return
+			}
+		}
+
+		checker.report(
+			&InvalidMoveError{
+				Name:            variable.Identifier,
+				DeclarationKind: declarationKind,
+				Pos:             IdentifierExpression.StartPosition(),
```
turbolent, line 76:
```suggestion
func (checker *Checker) checkVariableMove(identifierExpression *ast.IdentifierExpression, variable *Variable) {

	reportMaybeInvalidMove := func(declarationKind common.DeclarationKind) {
		// If the parent is member-access or index-access, then it's OK.
		// e.g: `v.foo`, `v.bar()`, `v[T]` are OK.
		switch parent := checker.parent.(type) {
		case *ast.MemberExpression:
			// TODO: No need for below check? i.e: should always be true
			if parent.Expression == identifierExpression {
				return
			}
		case *ast.IndexExpression:
			// Only `v[foo]` is OK, `foo[v]` is not.
			if parent.TargetExpression == identifierExpression {
				return
			}
		}

		checker.report(
			&InvalidMoveError{
				Name:            variable.Identifier,
				DeclarationKind: declarationKind,
				Pos:             identifierExpression.StartPosition(),
```

turbolent (approved):
Great work! Much better to check this consistently in one place 👌

runtime/runtime_test.go
```diff
@@ -10634,123 +10634,6 @@ func TestRuntimeNonPublicAccessModifierInInterface(t *testing.T) {
 	require.Len(t, conformanceErr.MemberMismatches, 2)
 }
 
-func TestRuntimeMoveSelfVariable(t *testing.T) {
```
SupunS, line 10637: These fails at the checker now. There are checker tests added for this particular scenario.

turbolent (approved):

PR #3342, opened 2024-05-14
Wrap host-function typed member-values with a bound-function

SupunS:
Work towards https://github.com/dapperlabs/cadence-internal/issues/215

## Description

⚠️  The diff is best viewed with whitespaces off.

Wraps host-function with a bound function, when accessing members of containers (arrays, dictionaries, and composites), if they are function-typed members.

This is already done for user-defined functions (interpreted functions) of composites. So this PR is basically extending the same for native functions (host functions) as well, and for other types of containers.

### Note:
I refactored the code to create bound functions for all native instance-methods. Unfortunately it exploded in the code, and had to refactor quite a bit.
There are a couple of places (e.g: RLP methods, HashAlogrithm methods) where creating a bound function is not possible because the interpreter is not available. However they are OK for now, since the underlying values are not resources and would have no impact on behaviour. i.e: conceptually wrong, but functionality wise no difference.
Left TODOs in these places, we can try to refactor these later.

On a side note, I also renamed "NewHostFunction" to "NewStaticHostFunction", to avoid naming chaos (bound, unbound, unmetered-bound, unmetered-unbound, etc.)

______


runtime/interpreter/interpreter.go
```diff
@@ -1187,14 +1188,20 @@ func (declarationInterpreter *Interpreter) declareNonEnumCompositeValue(
 
 	var initializerFunction FunctionValue
 	if declaration.Kind() == common.CompositeKindEvent {
-		initializerFunction = NewHostFunctionValue(
+		// Initializer is a static function.
+		initializerFunction = NewStaticHostFunctionValue(
```
turbolent, line 1192: Should this be a bound function?

turbolent (approved):
Really great work! 👏

turbolent (approved):

runtime/interpreter/interpreter.go
SupunS: hmm, I initially thought this is same as the constructor. But seems not? In that case, yeah, this can be a bound function.

PR #3345, opened 2024-05-14
Add back transaction `execute` tests

SupunS:
Port of https://github.com/dapperlabs/cadence-internal/pull/225 by @dsainati1

## Description

This adds back a number of tests for destructor type checking and interpreting that were removed when destructors were removed, except rewritten to use transaction execute blocks. The majority of the removed tests relied on certain properties of destructors that don't apply to execute blocks, in particular the ability to re-enter and invoke destructors from within other functions. This doesn't apply to execute blocks, which can only be entered once from the top level, and as such those tests could not be ported. However, tests that didn't rely on that behavior and only made use of the relaxed checker rules for nested resource moves were ported.

This uncovered a runtime crash with recursive transfers, which luckily was fixed by @SupunS's work in https://github.com/onflow/cadence/pull/3342, and so this PR is based on top of that one. It also depends on https://github.com/onflow/cadence/pull/3331.

______

### DIFF
--- a/runtime/runtime_test.go
+++ b/runtime/runtime_test.go
@@ -9787,12 +9787,11 @@ func TestRuntimePreconditionDuplication(t *testing.T) {
 	var checkerErr *sema.CheckerError
 	require.ErrorAs(t, err, &checkerErr)
 
-	errs := checker.RequireCheckerErrors(t, checkerErr, 4)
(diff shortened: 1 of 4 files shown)


turbolent (approved):
👍

Issue #3355, opened 2024-05-16
Access of enum-typed field on reference should maybe return non-reference


turbolent:
🧵 https://discord.com/channels/613813861610684416/621847426201944074/1240549428452921399

In Cadence 1.0, accessing a field of a reference returns a reference. This change was introduced in https://github.com/onflow/flips/blob/main/cadence/20230517-member-access-semnatics.md.

Enums aren't primitives, they are composite values, therefore accessing an enum-typed field of a reference returns a reference to the enum.

Though this behaviour is correct, it is obviously unnecessarily inconvenient: enums are immutable, getting a copy is safe. We might want to improve how the FLIP applies to enums in particular, e.g. return a copy in this case.

PR #3357, opened 2024-05-17
Return a copy for enum-typed members, during member-access via references

SupunS:
Closes https://github.com/onflow/cadence/issues/3355

## Description

Enums are immutable, so returning a copy during member access via references is safe.

______

### DIFF
--- a/runtime/tests/checker/entitlements_test.go
+++ b/runtime/tests/checker/entitlements_test.go
@@ -1615,6 +1615,29 @@ func TestCheckBasicEntitlementMappingAccess(t *testing.T) {
 
 		assert.NoError(t, err)
 	})
+
+	t.Run("enum field", func(t *testing.T) {
+		t.Parallel()
+
+		_, err := ParseAndCheck(t, `
+            enum Status: Int {
+                case On
+                case Off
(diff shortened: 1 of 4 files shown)


runtime/tests/checker/entitlements_test.go
```diff
@@ -1615,6 +1615,29 @@ func TestCheckBasicEntitlementMappingAccess(t *testing.T) {
...
+            struct interface S {
+                // enum cases are public.
+                // Hence, using entitlement mappings for enum-typed variables are pointless.
+                // So reject this statically.
+                access(mapping M) let status: Status
+            }
+        `)
+
+		errs := RequireCheckerErrors(t, err, 1)
+		require.IsType(t, &sema.InvalidMappedEntitlementMemberError{}, errs[0])
```
SupunS, line 1639: This is a side effect of the change, which I think makes sense. But if we think otherwise, I can only make the change to the member-access expression and for-loop only (which is going to be a bit uglier, but still can do).

turbolent (approved):
Nice, thank your for picking this up!

Great to see it was easy to change :ok_hand:

PR #3368, opened 2024-05-23
Decrease move operator precedence relative to cast operators

dsainati1:
Closes https://github.com/onflow/cadence/issues/3372

Currently the `<-` move operator has higher precedence than casting operators, meaning that expressions like `return <-r as! @T` must be manually parenthesized as `return <-(r as! @T)`, or they will not typecheck. This lowers the move operator's precedence in the expression parser so that the unparenthesized version parses in a way that will typecheck. 
______


dsainati1:
cc @turbolent 

This passes our tests, what would be the procedure for checking this against staged contracts to see if we break anything?

SupunS (approved):

turbolent (approved):

Issue #3366, opened 2024-05-23
Migration problems


turbolent:
See https://discord.com/channels/613813861610684416/1108479699732152503/1241110592153063619

We recently created two migration results:
- `migrationtestnet1-may8-inlined-preview20`: migrated with an older, unoptimized version of Cadence/flow-go
- `migrationtestnet1-may8-inlined-preview23`: migrated with the latest, optimized version of Cadence/flow-go

These results were both generated from the same TN snapshot (right, @j1010001?), but produced different state commitments.

As a result, I developed a [flow-go util command](https://github.com/onflow/flow-go/pull/5970) to compare two states.

The diffing uncovered several reasons for the different state commitments, but also uncovered several problems:

- Some capabilities using `LockedTokens.TokenAdmin` have different authorizations:

  ```json
  {
    "Address": "50f4b24a66fc3efb",
    "Domain": "storage",
    "Kind": "cadence_value_diff",
    "Msg": "values differ: *interpreter.PathCapabilityValue vs *interpreter.PathCapabilityValue",
    "Trace": "storage[flowTokenHolder].tokenManager",
    "OldValue": "Capability<auth(A.9a0766d93b6608b7.FungibleToken.Withdraw) &A.95e019a17d0e23d7.LockedTokens.LockedTokenManager>(address: 0x0a4b7be95de5c4f7, path: /storage/lockedTokenManager)",
    "NewValue": "Capability<auth(A.9a0766d93b6608b7.FungibleToken.Withdraw, A.95e019a17d0e23d7.LockedTokens.UnlockTokens) &A.95e019a17d0e23d7.LockedTokens.LockedTokenManager>(address: 0x0a4b7be95de5c4f7, path: /storage/lockedTokenManager)",
    "OldValueStaticType": "Capability<auth(A.9a0766d93b6608b7.FungibleToken.Withdraw) &A.95e019a17d0e23d7.LockedTokens.LockedTokenManager>",
    "NewValueStaticType": "Capability<auth(A.9a0766d93b6608b7.FungibleToken.Withdraw, A.95e019a17d0e23d7.LockedTokens.UnlockTokens) &A.95e019a17d0e23d7.LockedTokens.LockedTokenManager>"
  }
  ```

  This is because a new entitlement was added to the `LockedTokenManager` contract, see https://github.com/onflow/flow-core-contracts/pull/429/files#diff-a200545fac9017d8d015d11d56aa104f41e0a09088a2fb5f8eabb7006001528dR85

[...]

PR #3375, opened 2024-05-28
Remove incorrect caching from migrations

turbolent:
Work towards #3366 

## Description

Caching is currently incorrect. For example, the entitlements migration should not cache using just the type ID, as e.g. it does not include an intersection type's potential legacy restricted type. 

Remove caching for now. We can bring it back once/if we figure out how to cache correctly.

______


dsainati1:
If the caching is incorrect because of the typeID computation for intersection types missing the legacy type, can we just add the legacy type to the typeID computation instead of removing caching entirely?

dsainati1 (approved):

SupunS (approved):

Issue #3210, opened 2024-04-01
Allow removing type declarations from contracts


SupunS:
### Issue to be solved

Currently, the contract update validator prevents removing type declarations from a contract, to prevent it from getting re-added with a different structure. However, it would be useful to have a way to remove type definitions and force users of those contracts to start using a new type instead. This could be even more useful with cadence 1.0.

#### Suggestion I:

One possible solution is to introduce a pragma to tombstone a type definition. e.g: `#removedType(T)`.
One can remove a type and add this pragma to indicate that this type was removed. It will prevent re-adding a new type with the same name in future. The pragma itself is not removable once it is aded.

**Cons:**
 - The problem with just removing a type is that, any stored value belongs to this type would become unusable. This does not cause any harm to the network though (which was the original intention of preventing type removals).
 - How would migrations handle these missing types?

#### Suggestion II:

Alternatively, it is possible to provide a way to remove and replace a type with another, particularly during the Cadence 1.0 upgrades. e.g: `#replacedType(T, R)`, where `R` is the new type to replace with. Similar to the suggested `removedType` pragma, this will prevent any re-additions of a new type with the same name. 
 - During the state migrations, all usages of `T` would be replaced with type `R`, including static types of values, conformances, etc.
 - This is in a way, providing the user also the ability to define type-change rules similar to what we have done for the system contracts.
 - Migrations will first have to collect all these replacements pairs first, before do the actual migrations

This would be only supported if `T` and `R` satisfy particular conditions, such as:

- `T` and `R` has to be of the same kind. i.e: both are resources / both are resource-interfaces / etc.
- The new type `R` should have at-most the set of fields as the old type `T`. It's ok to have less number of fields, but no any additional fields. The filed names and types of the fields should be the same. (This is similar to the existing field update restrictions during contract updates).
- So this is basically similar to allowing renaming a type (in addition to what is already supported in existing contract updates) 
- Anything else??

j1010001:
Not breaking, does not make sense to do this post C1.0 migration

j1010001:
Implementing Suggestion 1 would unblock staging of the FiatToken contract, which is now a blocker for Crescendo launch. Increasing priority to urgent.

turbolent:
@dsainati1 Could you please document this? Both in the language reference, as well as a short update to the forum post https://forum.flow.com/t/update-on-cadence-1-0/5197

PR #3376, opened 2024-05-28
Allow types to be removed in contract updates

dsainati1:
Closes https://github.com/onflow/cadence/issues/3210

Allow removing types during contract updates if they are replaced with a `#removedType(T)` pragma. This special pragma works as follows:
1) It cannot be removed once added
2) A declaration for a type `T` cannot be added to a program that has a `#removedType(T)` present 
3) The pragma is scoped to the declaration in which it appears, so within a contract declaration `#removedType(T)` prevents a new nested type called `T` from being added, and would permit an existing nested type called `T` to be removed

______


runtime/stdlib/contract_update_validation.go
```diff
@@ -215,6 +217,30 @@ func (validator *ContractUpdateValidator) hasErrors() bool {
 	return len(validator.errors) > 0
 }
 
+func collectRemovedTypePragmas(validator UpdateValidator, pragmas []*ast.PragmaDeclaration) (removedTypes []ast.Identifier) {
```
turbolent, line 221: Maybe make the result set. It is only used for inclusion tests (contains), right? Maybe the AST identifier's identifier (string) is enough?

Also consider that `ast.Identifier` may have nested identifiers. Those should be maybe rejected in `collectRemovedTypePragmas`. e.g. `#removedType(Foo.Bar)`

turbolent (commented):
Nice!

runtime/stdlib/contract_update_validation.go
dsainati1: > Also consider that ast.Identifier may have nested identifiers. Those should be maybe rejected in collectRemovedTypePragmas. e.g. #removedType(Foo.Bar)

Yeah, it's already rejected. `ast.NominalType` can have nested identifiers, but `ast.Identifer` is just a wrapper around a string, i.e. no nested identifiers. We check in `collectRemovedTypePragmas` that the argument to `#removedType` is an identifier, restricting it to just `Foo`-shaped things

turbolent: Ah right, I confused `ast.Identifer` with `ast.NominalType` 👍

runtime/stdlib/contract_update_validation.go
```diff
@@ -215,6 +216,32 @@ func (validator *ContractUpdateValidator) hasErrors() bool {
 	return len(validator.errors) > 0
 }
 
+func collectRemovedTypePragmas(validator UpdateValidator, pragmas []*ast.PragmaDeclaration) map[string]struct{} {
+	removedTypes := map[string]struct{}{}
+
+	for _, pragma := range pragmas {
+		invocationExpression, isInvocation := pragma.Expression.(*ast.InvocationExpression)
+		if !isInvocation || len(invocationExpression.Arguments) != 1 {
```
turbolent, line 224: Maybe report a dedicated error for `len(invocationExpression.Arguments) != 1`. The current code will just silently ignore something like `#removedType(A, B)`. This would need to get moved after the check if the pragma is `removedType`.

go.mod
```diff
@@ -55,6 +55,7 @@ require (
 	github.com/x448/float16 v0.8.4 // indirect
 	github.com/zeebo/assert v1.3.0 // indirect
 	github.com/zeebo/blake3 v0.2.3 // indirect
+	golang.org/dl v0.0.0-20240507154152-429173cd35bf // indirect
```
turbolent, line 58: Is this needed?

dsainati1: No, not sure why this is here

turbolent (dismissed):
Looks good! Just one final comment regarding adding some additional tests

runtime/stdlib/contract_update_validation.go
```diff
@@ -334,18 +374,46 @@ func (validator *ContractUpdateValidator) oldTypeID(oldType *ast.NominalType) co
...
 	checkConformance checkConformanceFunc,
 ) {
 
+	// process pragmas first, as they determine whether types can later be removed
+	oldRemovedTypes := collectRemovedTypePragmas(validator, oldDeclaration.DeclarationMembers().Pragmas())
+	removedTypes := collectRemovedTypePragmas(validator, newDeclaration.DeclarationMembers().Pragmas())
+
+	// #typeRemoval pragmas cannot be removed, so any that appear in the old program must appear in the new program
+	// they can however, be added, so use the new program's type removals for the purposes of checking the upgrade
+	for oldRemovedType, _ := range oldRemovedTypes { //nolint:maprange
```
turbolent, line 403: Sorry, just noticed that this loop has side-effects (reporting), so this needs to be changed.

Maybe switch to an `OrderedSet` to get both deterministic ordering and fast set inclusion tests.

SupunS (approved):
Nice!

runtime/stdlib/contract_update_validation.go
turbolent:
```suggestion
		if !isInvocation {
```

Can you please also add tests for the error reported below (`if len(invocationExpression.Arguments) != 1 {`), e.g. for the case where length is 0 or 2?

turbolent (approved):

PR #3381, opened 2024-05-29
Improve dictionary value migration

turbolent:
## Description

#3316 improved the migration of dictionaries to work around the limitation of the mutating iterator in the inlining atree, which re-hashes keys while iterating and thus fails to load old / unmigrated values, by splitting the migration of dictionaries into two phases, one for keys, one for values.

This fixed the problem in the first phase, migrating the keys in `StorageMigration.migrateDictionaryKeys()`, by using a read-only iterator, which doesn't have the re-hasing problem that the mutating iterator in the inlining version of atree has.

However, we still have the problem in the second phase, migrating the values in `StorageMigration.migrateDictionaryValues()`, where we use a mutating iterator to re-iterate over the dictionary.

Gather all keys in the first phase and use them in the second phase, instead of re-iterating.

This doesn't fix anything on `master`, but will fix the iteration problem in the atree register inlining feature branch. Having the same code on `master` avoids unnecessary drift.

Thanks to @SupunS for the great idea in https://discord.com/channels/613813861610684416/1108479699732152503/1245434532572958791 🙏 

______

### DIFF
--- a/migrations/entitlements/migration_test.go
+++ b/migrations/entitlements/migration_test.go
@@ -3564,17 +3564,6 @@ func TestUseAfterMigrationFailure(t *testing.T) {
 		)
 	}
 
-	entitlementSetAuthorization := sema.NewEntitlementSetAccess(
-		[]*sema.EntitlementType{
-			sema.NewEntitlementType(
-				nil,
-				fooAddressLocation,
-				"E",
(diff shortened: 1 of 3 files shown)


SupunS (approved):
👏

PR #3386, opened 2024-05-31
Improve dictionary key migration

turbolent:
## Description

Context: https://discord.com/channels/613813861610684416/1108479699732152503/1245434532572958791

Revert part of #3381. This approach does not work: When a new migration runs that does not migrate, after a previous migration that did migrate, then we can't just use the legacy key as a lookup – the key has not been migrated in the current migration, but it is in migrated form.

Improve the migration of dictionary keys that have a legacy representation: Introduce a new migration, `TypeKeyMigration`, that re-stores `Type` values as-is, if they appear as dictionary keys. This migration is meant to be run after the existing migrations that update `Type` values, i.e. the static type migration and entitlements migration. The migration ensures that type values used as dictionary keys are stored under their new hash, even if a prior migration failed to do so. That way the key-value pair stays accessible, and the mutating iterator of the inlining-version of atree does not fail when iterating over the migrated dictionary.

I tested this will also work on `feature/atree-register-inlining-v1.0` and in flow-go `master.

______

### DIFF
--- a/migrations/migration.go
+++ b/migrations/migration.go
@@ -22,6 +22,8 @@ import (
 	"fmt"
 	"runtime/debug"
 
+	"github.com/onflow/atree"
+
 	"github.com/onflow/cadence/runtime"
 	"github.com/onflow/cadence/runtime/common"
 	"github.com/onflow/cadence/runtime/errors"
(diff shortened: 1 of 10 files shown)


migrations/statictypes/statictype_migration.go
```diff
@@ -586,6 +587,9 @@ func CanSkipStaticTypeMigration(valueType interpreter.StaticType) bool {
 	case interpreter.PrimitiveStaticType:
 
 		switch valueType {
+		case interpreter.PrimitiveStaticTypeMetaType:
+			return false
+
```
turbolent, line 592: This isn't strictly needed as it is the default, but I added it for clarity

migrations/migration.go
```diff
@@ -37,11 +39,19 @@ type ValueMigration interface {
...
 	) (newValue interpreter.Value, err error)
 	CanSkip(valueType interpreter.StaticType) bool
 	Domains() map[string]struct{}
 }
 
+type ValueMigrationPosition uint8
+
+const (
+	ValueMigrationPositionOther ValueMigrationPosition = iota
+	ValueMigrationPositionDictionaryKey
```
turbolent, line 52: We can add more specific ones here, but they are not needed at the momement

dsainati1 (approved):

SupunS (approved):
👍

fxamacker (approved):
LGMT! :+1:  Thanks for adding a migration step for all legacy keys!

PR #3387, opened 2024-06-03
Don't short circuit references on member access when entitlements are involved

dsainati1:
When a struct-typed field is accessed on a reference to a composite, we return a reference to that struct instead of the actual struct itself. However, when the struct-typed field has type `AnyStruct` and the underlying value of that field is itself a reference, we don't create a doubly nested reference, instead just returning the underlying value directly. This is normally fine, but when the type being created has entitlements (usually as the result of a mapping) this can cause entitlements to be granted to the resulting reference that it shouldn't have. This disables this short circuiting when if the reference type being created should be entitled. 

Thanks to @bluesign for the report

______

### DIFF
--- a/runtime/interpreter/interpreter_expression.go
+++ b/runtime/interpreter/interpreter_expression.go
@@ -304,15 +304,25 @@ func (interpreter *Interpreter) memberExpressionGetterSetter(
 // e.g.1: Given type T, this method returns &T.
 // e.g.2: Given T?, this returns (&T)?
 func (interpreter *Interpreter) getReferenceValue(value Value, resultType sema.Type, locationRange LocationRange) Value {
+
+	// `resultType` is always an [optional] reference.
+	// This is guaranteed by the checker.
+	referenceType, ok := sema.UnwrapOptionalType(resultType).(*sema.ReferenceType)
+	if !ok {
+		panic(errors.NewUnreachableError())
+	}
+
 	switch value := value.(type) {
 	case NilValue, ReferenceValue:
 		// Reference to a nil, should return a nil.
 		// If the value is already a reference then return the same reference.
 		// However, we need to make sure that this reference is actually a subtype of the resultType,
 		// since the checker may not be aware that we are "short-circuiting" in this case
+		// Additionally, it is only safe to "compress" reference types like this when the desired
+		// result reference type is unauthorized
 
 		staticType := value.StaticType(interpreter)
-		if !interpreter.IsSubTypeOfSemaType(staticType, resultType) {
+		if referenceType.Authorization != sema.UnauthorizedAccess || !interpreter.IsSubTypeOfSemaType(staticType, resultType) {
 			panic(InvalidMemberReferenceError{
 				ExpectedType:  resultType,
 				ActualType:    interpreter.MustConvertStaticToSemaType(staticType),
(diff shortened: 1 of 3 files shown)


runtime/interpreter/interpreter_expression.go
```diff
@@ -304,15 +304,25 @@ func (interpreter *Interpreter) memberExpressionGetterSetter(
...
 		// Reference to a nil, should return a nil.
 		// If the value is already a reference then return the same reference.
 		// However, we need to make sure that this reference is actually a subtype of the resultType,
 		// since the checker may not be aware that we are "short-circuiting" in this case
+		// Additionally, it is only safe to "compress" reference types like this when the desired
+		// result reference type is unauthorized
 
 		staticType := value.StaticType(interpreter)
-		if !interpreter.IsSubTypeOfSemaType(staticType, resultType) {
+		if referenceType.Authorization != sema.UnauthorizedAccess || !interpreter.IsSubTypeOfSemaType(staticType, resultType) {
```
SupunS, line 325: Should we not allow if there are entitlements, but there are the same/equivalent?

runtime/tests/interpreter/member_test.go
```diff
@@ -1197,4 +1197,97 @@ func TestInterpretNestedReferenceMemberAccess(t *testing.T) {
...
+                init(_ a: AnyStruct){
+                    self.ref = a
+                }
+            }
+        
+            fun test() {
+                let t = T()
+                let pubTRef = &t as &T
+                var s = &S(pubTRef) as auth(E) &S
+                var tRef = s.ref as! auth(E) &T
```
SupunS, line 1224: Can you add a test for the same, but instead of `pubTRef = &t as &T`, with an entitled reference; `pubTRef = &t as auth(E) &T`?

I guess that should pass?
i.e: doing `s.ref as! auth(E) &T` in two steps as
```cadence
var member = s.ref
var tRef =  member as! auth(E) &T
```
already works? (not sure)

runtime/interpreter/interpreter_expression.go
dsainati1: No, this is exactly the scenario that resulted in the issue in the first place. Different entitlements are already disallowed by the subtyping check, but in the example that motivated this, the entitlement mapping is pre-computed to ask for the entitlements in the output of the map.

runtime/tests/interpreter/member_test.go
dsainati1: No, this also must fail. Here the static type of `member` is `&AnyStruct` after the `s.ref` access. We cannot allow this to be downcast to `auth(E) &T` because entitlements are erased at runtime, so there is no way to regain the `E` entitlement after it is lost. I added a test for this case though.

turbolent (commented):
The description specifically refers to structs. Is this only a problem for structs and the problem does not exist / is already handled for resources?

dsainati1:
> The description specifically refers to structs. Is this only a problem for structs and the problem does not exist / is already handled for resources?

Yes, because references are not a subtype of `AnyResource` you cannot run into this problem for resources, (`&R` would not be a subtype of `&AnyStruct` either) it only exists for structs.

runtime/tests/interpreter/member_test.go
SupunS: ah right, `member` is `&AnyStruct` (not `AnyStruct` which is what I was thinking) 👍 . Makes sense

Issue #2679, opened 2023-07-24
Non-deterministic error strings caused by debug stack traces


peterargue:
### Current Behavior

I'm working on adding a commitment to the protocol state for the `TransactionResults` produced by execution nodes. The results include the errors encountered while executing the transaction. To produce a deterministic commitment to the results, there must be a canonical representation of any errors that are included.

Some cadence runtime error conditions throw an `UnexpectedError` which contains a debug stack trace:
https://github.com/onflow/cadence/blob/181924e3b5ceebae628ae82299a19dbf3fc60aba/runtime/errors/errors.go#L142-L146

A common example is when computation limits are exceeded. e.g.
```
Execution failed:
error: internal error: failed to store slab 0x6320abf93dc53317.560373: set value failed: [Error Code: 1110] computation exceeds limit (9999)
goroutine 15210758782 [running]:
runtime/debug.Stack()
	/usr/local/go/src/runtime/debug/stack.go:24 +0x65
github.com/onflow/cadence/runtime/errors.NewUnexpectedErrorFromCause(...)
...
```

While these are convenient for debugging, the error struct and resulting string are non-deterministic since the backtrace includes the goroutine id and pointer values.

It may be possible to strip the `Stack` variable out of the error in the FVM before building the final error struct, but that may get complicated depending on error wrapping and serialization along the error path.

### Expected Behavior

Ideally, stack traces would not be included in production error messages.

### Steps To Reproduce

Send a transaction with computation limit set to 1. This will generate an error that includes a backtrace.

### Environment

```markdown
- Cadence version: v0.39.14
- Network: mainnet
```

turbolent:
We should put this behind a feature flag, disabled by default, but enabled in e.g. the Emulator

PR #3392, opened 2024-06-04
Remove stacktrace from errors

SupunS:
Work towards https://github.com/onflow/cadence/issues/2679

## Description

I tried putting this behind a flag, but unfortunately there are 300+ places where this internal error is created, and many of them don't have the access to the runtime/interpreter/checker objects to get access to the "flag". Passing down the flag for all these 300+ call-sites exponentially grows and explodes in the code base. I don't really think changing that much just for a debug trace worth the effort, given this won't be useful for the smart contract developers (only we, the language developers will use it, but in such cases we would anyway be debugging the code, and won't really rely on the stack-trace).

Anyway, if we think otherwise: that the stack trace is strictly necessary, then I can start going down the rabbit hole, but trying to keep it simple here.

______

### DIFF
--- a/runtime/errors/errors.go
+++ b/runtime/errors/errors.go
@@ -20,7 +20,6 @@ package errors
 
 import (
 	"fmt"
-	"runtime/debug"
 
 	"golang.org/x/xerrors"
 )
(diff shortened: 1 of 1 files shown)


turbolent (approved):
Agreed, removing the stack trace is probably fine. 👍 

This is a good first step towards making internal errors more deterministic.

Just to note though, this won't fully resolve the non-determinism for internal errors, as the error message might also differ based on Go compiler version, and by nature the error message might contain non-constant / input-dependent data (e.g. pointers)

Issue #3297, opened 2024-04-29
Epic - Migration optimization


j1010001:
Collection of all migration optimization issues

```[tasklist]
### Optimizations
- [ ] https://github.com/onflow/flow-go/issues/6132
```

PR #3396, opened 2024-06-05
Cache results of static types migration and entitlements migration

turbolent:
Work towards #3297

## Description

In #3375 we removed the broken caching of results of the static types migration and entitlements migration.

Bring back the caching, but use the encoding of the static type instead of the type ID. The encoding of a static type isn't ambiguous or lacking any details compared to the type ID. I double checked that legacy /  pre-1.0 information in static types, like the legacy restricted type in an intersection type (previously restriction type), is not only decoded properly for the migration, but also encoded again with the legacy information (which isn't the case for type IDs).

Disable the caching by default, so the default behaviour is always correct. In flow-go, this feature will be enabled through a flag (again, off by default).

I tested this approach in flow-go, and a run with caching always produces the same new state commitment as a non-caching run.

______

### DIFF
--- a/migrations/cache.go
+++ b/migrations/cache.go
@@ -1,0 +1,79 @@
+/*
+ * Cadence - The resource-oriented smart contract programming language
+ *
+ * Copyright Dapper Labs, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


dsainati1 (approved):

SupunS (approved):
Looks good! My only concern/question is what the performance would be for converting the static-type to bytes (encoding) vs creating a new static type. I guess we can only find out by running the actual migration with it. So good idea to have behind a flag.

PR #3404, opened 2024-06-07
Prevent storage reference to another reference

turbolent:
Closes https://github.com/dapperlabs/cadence-internal/issues/235

## Description

Based on #3398. Thanks for @dsainati1 for the initial investigation.

Nested references are not allowed. 

The creation of an ephemeral reference, e.g. through a reference expression, already prevents this (`NewUnmeteredEphemeralReferenceValue` aborts with `NestedReferenceError`).

Also prevent storage references to references. Given that the target of a storage reference is a path and the value stored in the path may change, check the referenced value is not a reference value on each dereference, not just when the storage reference is created.

______

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -20601,6 +20601,13 @@ func (v *StorageReferenceValue) dereference(interpreter *Interpreter, locationRa
 		return nil, nil
 	}
 
+	if reference, isReference := referenced.(ReferenceValue); isReference {
+		panic(NestedReferenceError{
+			Value:         reference,
+			LocationRange: locationRange,
+		})
+	}
+
 	if v.BorrowedType != nil {
 		staticType := referenced.StaticType(interpreter)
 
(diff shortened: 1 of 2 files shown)


dsainati1 (approved):

runtime/interpreter/value.go
```diff
@@ -20601,6 +20601,13 @@ func (v *StorageReferenceValue) dereference(interpreter *Interpreter, locationRa
 		return nil, nil
 	}
 
+	if reference, isReference := referenced.(ReferenceValue); isReference {
+		panic(NestedReferenceError{
+			Value:         reference,
+			LocationRange: locationRange,
+		})
+	}
```
SupunS, line 20609: Can we do this check at the time of the creation? i.e: in the constructor `NewUnmeteredStorageReferenceValue`

SupunS: ah never mind, just saw the reason for doing it here.

SupunS (approved):

PR #3406, opened 2024-06-10
Add access top type to model inaccessible access for identity maps

dsainati1:
Closes https://github.com/dapperlabs/cadence-internal/issues/236

The `Identity` entitlement mapping has a codomain (output space) that is theoretically infinite, as any number of possible entitlements can be produced from it. This means that in order to be safe when initializing a field that is mapped with the `Identity`, or a mapping that includes it, the reference used to initialize this field must be authorized to an impossible access, an infinite set of entitlements. Otherwise, the `Identity` mapping could be used to escalate entitlements, as seen in one of the tests added in this PR.

This PR introduces a new `None` access to represent this impossible set, and ensures that this permits no other access besides itself (functionally the top type of the access lattice, the same way that `all` is the bottom). Initializing a reference with the `Identity` map type requires this top access value, which can only be obtained via immediately creating a reference to an owned value. This way we ensure that `Identity`-mapped fields can never be initialized with a reference that has a limited entitlement set. 

______

### DIFF
--- a/runtime/interpreter/encode.go
+++ b/runtime/interpreter/encode.go
@@ -224,8 +224,14 @@ const (
 	CBORTagUnauthorizedStaticAuthorization
 	CBORTagEntitlementMapStaticAuthorization
 	CBORTagEntitlementSetStaticAuthorization
-	CBORTagInclusiveRangeStaticType
 	CBORTagInaccessibleStaticAuthorization
+
+	_
+	_
+	_
+	_
+
+	CBORTagInclusiveRangeStaticType
 
 	// !!! *WARNING* !!!
 	// ADD NEW TYPES *BEFORE* THIS WARNING.
--- a/runtime/interpreter/encode.go
+++ b/runtime/interpreter/encode.go
@@ -225,6 +225,7 @@ const (
 	CBORTagEntitlementMapStaticAuthorization
 	CBORTagEntitlementSetStaticAuthorization
 	CBORTagInclusiveRangeStaticType
+	CBORTagInaccessibleStaticAuthorization
 
 	// !!! *WARNING* !!!
 	// ADD NEW TYPES *BEFORE* THIS WARNING.
@@ -1310,6 +1311,17 @@ func (t Unauthorized) Encode(e *cbor.StreamEncoder) error {
 	err := e.EncodeRawBytes([]byte{
 		// tag number
 		0xd8, CBORTagUnauthorizedStaticAuthorization,
+	})
+	if err != nil {
+		return err
+	}
+	return e.EncodeNil()
+}
+
+func (t Inaccessible) Encode(e *cbor.StreamEncoder) error {
+	err := e.EncodeRawBytes([]byte{
+		// tag number
+		0xd8, CBORTagInaccessibleStaticAuthorization,
 	})
 	if err != nil {
 		return err
--- a/runtime/interpreter/interpreter.go
+++ b/runtime/interpreter/interpreter.go
@@ -4991,7 +4991,7 @@ func (interpreter *Interpreter) mapMemberValueAuthorization(
 
 		default:
 			var access sema.Access
-			if mappedAccess.Type == sema.IdentityType {
+			if mappedAccess.Type.IncludesIdentity {
 				access = sema.AllSupportedEntitlements(resultingType)
 			}
 
--- a/runtime/interpreter/statictype.go
+++ b/runtime/interpreter/statictype.go
@@ -643,11 +643,11 @@ var FullyEntitledAccountAccess = ConvertSemaAccessToStaticAuthorization(nil, sem
 func (Unauthorized) isAuthorization() {}
 
 func (Unauthorized) String() string {
-	return ""
+	return "Unauthorized"
 }
 
-func (Unauthorized) MeteredString(_ common.MemoryGauge) string {
-	return ""
+func (a Unauthorized) MeteredString(_ common.MemoryGauge) string {
+	return "Unauthorized"
 }
 
 func (Unauthorized) ID() TypeID {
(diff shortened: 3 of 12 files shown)


runtime/interpreter/encode.go
```diff
@@ -225,6 +225,7 @@ const (
 	CBORTagEntitlementMapStaticAuthorization
 	CBORTagEntitlementSetStaticAuthorization
 	CBORTagInclusiveRangeStaticType
+	CBORTagInaccessibleStaticAuthorization
```
turbolent, line 228: We might want to clean this up while we still can: Given there is no on-chain state stored yet for `InclusiveRange`, we can change the tag, separating it from the authorization kinds, and leaving some space for future additions and changes in this "group" of tags:

```suggestion
	CBORTagInaccessibleStaticAuthorization
	_
	_
	_
	_
	CBORTagInclusiveRangeStaticType
```

I arbitrarily chose 4, but we can use fewer or more

runtime/interpreter/interpreter.go
```diff
@@ -4991,7 +4991,7 @@ func (interpreter *Interpreter) mapMemberValueAuthorization(
 
 		default:
 			var access sema.Access
-			if mappedAccess.Type == sema.IdentityType {
+			if mappedAccess.Type.IncludesIdentity {
```
turbolent, line 4994: Could you please add a comment to `IncludesIdentity` and explain for when it is set and how it is used?

runtime/interpreter/statictype.go
```diff
@@ -659,6 +660,29 @@ func (Unauthorized) Equal(auth Authorization) bool {
...
 }
 
+type Inaccessible struct{}
+
+var InaccessibleAccess Authorization = Inaccessible{}
+
+func (Inaccessible) isAuthorization() {}
+
+func (Inaccessible) String() string {
+	return ""
```
turbolent, line 670: Would it be possible to return a description here? At the same time, could `Unauthorized` above also get a non-empty string representation?

runtime/interpreter/statictype.go
```diff
@@ -1049,6 +1073,9 @@ func ConvertSemaAccessToStaticAuthorization(
 		if access.Equal(sema.UnauthorizedAccess) {
 			return UnauthorizedAccess
 		}
+		if access.Equal(sema.PrimitiveAccess(ast.AccessNone)) {
```
turbolent, line 1076: Given the repeated expressions converting AST to sema primitive accesses (e.g. here `sema.PrimitiveAccess(ast.AccessNone)`), it might make sense to define equivalents for the AST accesses (e.g. `ast.AccessNone`) in `sema`, e.g. `const AccessNone =PrimitiveAccess(ast.AccessNone)`, and then replace all occurrences

runtime/sema/access.go
```diff
@@ -458,6 +464,9 @@ func (a PrimitiveAccess) Equal(other Access) bool {
 }
 
 func (a PrimitiveAccess) PermitsAccess(otherAccess Access) bool {
+	if ast.PrimitiveAccess(a) == ast.AccessNone {
+		return PrimitiveAccess(ast.AccessNone) == otherAccess
```
turbolent, line 468: Avoid Yoda condition:

```suggestion
		return otherAccess == PrimitiveAccess(ast.AccessNone)
```

runtime/tests/checker/entitlements_test.go
```diff
@@ -6080,7 +6080,7 @@ func TestCheckIdentityMapping(t *testing.T) {
 
                 init() {
                     let x = X()
-                    self.x1 = &x as auth(A, B, C) &X
```
turbolent, line 6083: Would it be possible to keep the existing test and only adjust the assertion (if needed), and add the modified version as a new test?

runtime/tests/checker/entitlements_test.go
```diff
@@ -6191,6 +6191,83 @@ func TestCheckIdentityMapping(t *testing.T) {
...
+	t.Run("initializer with owned value", func(t *testing.T) {
+		t.Parallel()
+
+		_, err := ParseAndCheck(t, `
+            entitlement X
+
+            struct S {
+                access(mapping Identity) let x: [String]
+                init(_ str: [String]) {
+                    self.x = str // this should be possible, as the string array is owned here
```
turbolent, line 6223: Nice!

runtime/tests/checker/entitlements_test.go
```diff
@@ -6191,6 +6191,83 @@ func TestCheckIdentityMapping(t *testing.T) {
...
+	t.Run("initializer", func(t *testing.T) {
+		t.Parallel()
+
+		_, err := ParseAndCheck(t, `
+            entitlement X
+
+            struct S {
+                access(mapping Identity) let x: auth(mapping Identity) &String
+                init(_ str: auth(X) &String) {
+                    self.x = str // this should not be possible, as Identity may grant any entitlement, not just X
```
turbolent, line 6204: Nice!

runtime/tests/checker/entitlements_test.go
```diff
@@ -6191,6 +6191,83 @@ func TestCheckIdentityMapping(t *testing.T) {
...
+		t.Parallel()
+
+		_, err := ParseAndCheck(t, `
+            entitlement X
+
+            struct S {
+                access(mapping Identity) let x: auth(mapping Identity) &String
+                init(_ str: String) {
+                    self.x = &str // this should be possible, as we own the string and thus inference is able to give the &str 
+                    // reference the appropriate type
```
turbolent, line 6241: Nice!

runtime/tests/interpreter/entitlements_test.go
```diff
@@ -3024,7 +3024,7 @@ func TestInterpretIdentityMapping(t *testing.T) {
 
                 init() {
                     let x = X()
-                    self.x1 = &x as auth(A, B, C) &X
```
turbolent, line 3027: Same comment as above

turbolent (commented):
Great work! 👏 

The behaviour in the tests is looking really good 👌 It's great to see that we can specifically reject references and prevent rights escalation, while at the same time allow support owned value.

SupunS (approved):
LGTM. great tests!

turbolent (approved):
Great work! 👏

PR #3417, opened 2024-06-12
Import contracts as references

SupunS:
## Description

Implements the changes proposed in https://github.com/onflow/flips/pull/277

______

### DIFF
--- a/runtime/interpreter/interpreter_import.go
+++ b/runtime/interpreter/interpreter_import.go
@@ -84,6 +84,9 @@ func (interpreter *Interpreter) importResolvedLocation(resolvedLocation sema.Res
 
 	for _, name := range names {
 		variable := variables[name]
+		if variable == nil {
+			continue
+		}
 
 		value := variable.GetValue(interpreter)
 
--- a/runtime/interpreter/interpreter_import.go
+++ b/runtime/interpreter/interpreter_import.go
@@ -88,19 +88,24 @@ func (interpreter *Interpreter) importResolvedLocation(resolvedLocation sema.Res
 			continue
 		}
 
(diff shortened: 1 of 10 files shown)


runtime/interpreter/interpreter_import.go
```diff
@@ -83,6 +84,33 @@ func (interpreter *Interpreter) importResolvedLocation(resolvedLocation sema.Res
 
 	for _, name := range names {
 		variable := variables[name]
+		if variable == nil {
+			continue
+		}
+
+		value := variable.GetValue(interpreter)
```
turbolent, line 91: Not sure of it matters, but maybe it's good to keep the "time of evaluation" of the imported variables. 

Variables may be either have a value, or are lazy and have a function that produces the value. `GetValue` forces the production of the value. 

Previously, the variable was imported as-is, potentially not evaluated yet, and thus evaluated on first use. Now, the variable is eagerly evaluated at import time.

AFAICS there shouldn't be any part in Cadence's implementation that will provide an unevaluated / lazy variable during import, but I'm not 100% sure.

Maybe create a new lazy variable with a closure that will perform the evaluation of the imported variable and the code below.

turbolent (approved):
Great work!

Good to see the implementation is very simple, only a couple lines in checker and interpreter 👌 

Just one comment regarding the interpreter.

runtime/interpreter/interpreter_import.go
SupunS: done in f69da859e9d2bdc14bd0090f86fec19765ff0c4d

dsainati1 (approved):

turbolent (approved):

PR #3432, opened 2024-06-20
Remove access to field slices from composite and interface types

### LINKED ISSUE #2952: Remove access to fields by index

turbolent:
Closes #2952

## Description

Similar to how access to field value slices was from composite values was previously removed in #3290, also remove access to field slices of composite and interface types.

______


values.go
```diff
@@ -1521,16 +1521,64 @@ func NewMeteredKeyValuePair(gauge common.MemoryGauge, key, value Value) KeyValue
...
+func getCompositeFieldValues(composite Composite) []Value { //nolint:unused
 	return composite.getFieldValues()
 }
 
+// SearchFieldByName searches for the field with the given name in the composite type,
+// and returns the value of the field, or nil if the field is not found.
+//
+// WARNING: This function performs a linear search, so is not efficient for accessing multiple fields.
+// Prefer using FieldsMappedByName if you need to access multiple fields.
+func SearchFieldByName(v Composite, fieldName string) Value {
```
turbolent, line 1544: Moved as-is from `types.go`

values.go
```diff
@@ -1521,16 +1521,64 @@ func NewMeteredKeyValuePair(gauge common.MemoryGauge, key, value Value) KeyValue
...
+
+	for i, field := range fields {
+		if field.Identifier == fieldName {
+			return fieldValues[i]
+		}
+	}
+	return nil
+}
+
+func FieldsMappedByName(v Composite) map[string]Value {
```
turbolent, line 1560: Moved as-is from `types.go`

values.go
```diff
@@ -1521,16 +1521,64 @@ func NewMeteredKeyValuePair(gauge common.MemoryGauge, key, value Value) KeyValue
...
 type Composite interface {
 	Value
-	GetFields() []Field
+
+	isComposite()
+	getFields() []Field
 	getFieldValues() []Value
+
+	SearchFieldByName(fieldName string) Value
+	FieldsMappedByName() map[string]Value
```
turbolent, line 1530: Feedback from users: Expose existing global functions with the same name as methods on `Composite` for convenience.

fxamacker (approved):
Thanks for removing access to field types!  LGTM!

With both field types and values unexported, is it safe from Cadence perspective to sort fields in CCF now?

values.go
```diff
@@ -1521,16 +1521,64 @@ func NewMeteredKeyValuePair(gauge common.MemoryGauge, key, value Value) KeyValue
...
+
+	fieldsMap := make(map[string]Value, len(fields))
+	for i, fieldValue := range fieldValues {
+		var fieldName string
+		if i < len(fields) {
+			fieldName = fields[i].Identifier
+		} else if attachment, ok := fieldValue.(Attachment); ok {
+			fieldName = interpreter.AttachmentMemberName(attachment.Type().ID())
+		}
+		fieldsMap[fieldName] = fieldValue
```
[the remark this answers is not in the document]
turbolent: I guess so? I'll add a panic just in case, good idea

SupunS (approved):

PR #3449, opened 2024-07-03
Allow borrowing of capability with subtype

### LINKED ISSUE #3080: Allow borrowing a concrete reference type to a capability linked as an interface type
### Issue to be solved

In Cadence 1.0, with the addition of entitlements, references that are restricted to an interface type can be cast to the concrete type if it is known. Because of this, it would be nice to have the ability to borrow a concrete reference type directly from a capability that is only linked as an interface type instead of having to borrow and then cast.

The inability to do this may cause issues with some standard transactions. For example, many apps use a transaction like [this one](https://github.com/onflow/flow-nft/blob/standard-v2/transactions/setup_account_from_nft_reference.cdc) to set up an account based of only a reference to a generic NFT, without having imported the contract first. The transaction only links the new collection as `NonFungibleToken.Receiver`. 

Many other transactions try to borrow the public collection as its concrete type though, `ExampleToken.Collection`. Because this isn't possible, many of these transactions would fail. If they were able to borrow it directly, then this wouldn't be a problem. This allows us to standardize more simple transactions that don't need boilerplate for borrowing as a reference before casting. We can always just borrow directly.

### Suggested Solution

For a capability created like this:
```
        // create a public capability for the collection
        let collectionCap = signer.capabilities.storage.issue<&{NonFungibleToken.Receiver}>(
                collectionData.storagePath
            )
        signer.capabilities.publish(collectionCap, at: publicPath)
```

Assuming `ExampleNFT.Collection` implements `NonFungibleToken.Receiver`, the reference should be borrowable like this:
```
        // borrow a public reference to the receivers collection
        let receiverCap = recipient.capabilities.get<&ExampleNFT.Collection>(collectionData.publicPath)
            ?? panic("Could not get the recipient's Receiver Capability")
```

turbolent:
Closes #3080

## Description

Extend borrowing functions (`borrow`, `check`) to allow borrowing a capability also with a subtype, not just a supertype.

This is valid, because a reference borrowed with the issued type may also be downcasted, since https://github.com/onflow/flips/blob/main/cadence/20221214-auth-remodel.md.

Ensure though that the requested type's auth/access is not more permissive than the capability's.

cc @dsainati1 

______

### DIFF
--- a/runtime/stdlib/account.go
+++ b/runtime/stdlib/account.go
@@ -3522,7 +3522,7 @@ func canBorrow(
 		return false
 	}
 
-	// Ensure the wanted borrow type must be a subtype or supertype of the capability borrow type
+	// Ensure the wanted borrow type is a subtype or supertype of the capability borrow type
 
 	return sema.IsSubType(wantedBorrowType.Type, capabilityBorrowType.Type) ||
 		sema.IsSubType(capabilityBorrowType.Type, wantedBorrowType.Type)
--- a/runtime/stdlib/account.go
+++ b/runtime/stdlib/account.go
@@ -3509,6 +3509,25 @@ func newAccountCapabilitiesUnpublishFunction(
 	}
 }
 
+func canBorrow(
+	wantedBorrowType *sema.ReferenceType,
+	capabilityBorrowType *sema.ReferenceType,
+) bool {
+
+	// Ensure the wanted borrow type is not more permissive than the capability borrow type
+
+	if !wantedBorrowType.Authorization.
+		PermitsAccess(capabilityBorrowType.Authorization) {
+
+		return false
+	}
+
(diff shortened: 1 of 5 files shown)


runtime/stdlib/account.go
```diff
@@ -3509,6 +3509,25 @@ func newAccountCapabilitiesUnpublishFunction(
...
+
+	// Ensure the wanted borrow type is not more permissive than the capability borrow type
+
+	if !wantedBorrowType.Authorization.
+		PermitsAccess(capabilityBorrowType.Authorization) {
+
+		return false
+	}
+
+	// Ensure the wanted borrow type must be a subtype or supertype of the capability borrow type
```
turbolent, line 3525:
```suggestion
	// Ensure the wanted borrow type is a subtype or supertype of the capability borrow type
```

SupunS (approved):
Nice!

PR #3455, opened 2024-07-08
Add a String.contains function

turbolent:
## Description

Add a new string function which determines substring containment.

______

### DIFF
--- a/runtime/interpreter/value.go
+++ b/runtime/interpreter/value.go
@@ -1703,6 +1703,48 @@ func (v *StringValue) ForEach(
 	}
 }
 
+func (v *StringValue) IsBoundaryStart(start int) bool {
+	v.prepareGraphemes()
+	return v.isGraphemeBoundaryStartPrepared(start)
+}
+
+func (v *StringValue) isGraphemeBoundaryStartPrepared(start int) bool {
+
+	for {
+		boundaryStart, _ := v.graphemes.Positions()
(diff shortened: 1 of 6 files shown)


runtime/interpreter/value.go
```diff
@@ -1688,6 +1703,44 @@ func (v *StringValue) ForEach(
...
+				if end == expectedEnd {
+					return TrueValue
+				} else if end > expectedEnd {
+					return FalseValue
+				}
+
+				if !v.graphemes.Next() {
+					break
+				}
+			}
```
SupunS, line 1732: Asking out of curiosity, why would this inner loop be needed? Do you know any example?

SupunS (approved):

runtime/interpreter/value.go
turbolent: Great question! The tests contain an example: in ` \u{1F476}\u{1F3FB}`, ` \u{1F476}` is NOT a substring -- though the boundary start matches, the end does not.

BTW, I optimized the function in [1fcafe0](https://github.com/onflow/cadence/pull/3455/commits/1fcafe0f7bdba56e719f7a5dfd70a90cb3fcb165), but it still contains the grapheme cluster boundary end check.

SupunS: oh  see, thank you!

SupunS (approved):

Issue #3459, opened 2024-07-10
Emit events for capability controller operations


turbolent:
### Issue to be solved

Capability controller operations, such as issuing, should emit an event

### Suggested Solution

- `StorageCapabilityControllers.issue`: 
  `flow.StorageCapabilityControllerIssued(id: UInt64, address: Address, type: Type, path: StoragePath)`
- `AccountCapabilityControllers.issue`: 
  `flow.AccountCapabilityControllerIssued(id: UInt64, address: Address, type: Type)`

turbolent:
Do these proposed events above look good?

What other events should be emitted?

@austinkline @bjartek @joshuahannan @sisyphusSmiling Would love to hear your thoughts!

joshuahannan:
Those look good I think! Could we have some for deleting also? That might be useful

turbolent:
@joshuahannan good point!

How about:

- `StorageCapabilityControllerDeleted(id: UInt64, address: Address)`
- `AccountCapabilityControllerDeleted(id: UInt64, address: Address)`

turbolent:
Do we need an event for StorageCapabilityController.retarget`?

sisyphusSmiling:
I think the events and associated actions proposed are great. Only suggestion would be to consider emitting `type: String` instead of `type: Type` based on how the value is defined in v2 FT/NFT standard events.

> How about:
>
> StorageCapabilityControllerDeleted(id: UInt64, address: Address)
> AccountCapabilityControllerDeleted(id: UInt64, address: Address)

I also wonder if `tag` would be useful here. If so, we might also consider truncating at a certain length if there's concern around emitting garbage that would otherwise be too computationally expensive at scale.

turbolent:
> Is it possible to include the type in the `StorageCapabilityControllerDeleted` event?

We could, but it might be unecessary? We currently have the type in `InboxValuePublished`, but also omit it in `InboxValueUnpublished`. If we add it, we should also add it to `AccountCapabilityControllerDeleted`.

> Only suggestion would be to consider emitting type: String instead of type: Type based on how the value is defined in v2 FT/NFT standard events.

I'd prefer keeping strong types and avoid "stringly" typed code. It is also more consistent with existing event types, like `InboxValuePublished`.

turbolent:
As of #3464 the following events are emitted:

For the following functions, emit the new event:


- `StorageCapabilityControllers.issue`:
[...]

PR #3460, opened 2024-07-10
Emit events when capability controllers are issued

turbolent:
Work towards #3459

## Description

For the following functions, emit the new event:

- `StorageCapabilityControllers.issue`:
   `flow.StorageCapabilityControllerIssued(id: UInt64, address: Address, type: Type, path: StoragePath)`
- `AccountCapabilityControllers.issue`:
   `flow.AccountCapabilityControllerIssued(id: UInt64, address: Address, type: Type)`

______

### DIFF
--- a/migrations/capcons/migration_test.go
+++ b/migrations/capcons/migration_test.go
@@ -36,16 +36,37 @@ import (
 	"github.com/onflow/cadence/runtime/tests/utils"
 )
 
-type testAccountIDGenerator struct {
-	ids map[common.Address]uint64
+type testCapConIssueHandler struct {
+	ids    map[common.Address]uint64
+	events []cadence.Event
 }
 
-func (g *testAccountIDGenerator) GenerateAccountID(address common.Address) (uint64, error) {
+var _ stdlib.CapabilityControllerIssueHandler = &testCapConIssueHandler{}
+
+func (g *testCapConIssueHandler) GenerateAccountID(address common.Address) (uint64, error) {
 	if g.ids == nil {
 		g.ids = make(map[common.Address]uint64)
 	}
 	g.ids[address]++
 	return g.ids[address], nil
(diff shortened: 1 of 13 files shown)


migrations/capcons/migration_test.go
```diff
@@ -1818,6 +1921,11 @@ func TestLinkMigration(t *testing.T) {
 					capabilityID: 2,
 				},
 			},
+			expectedEvents: []string{
+				`flow.AccountCapabilityControllerIssued(id: 1, address: 0x0000000000000001, type: Type<auth(Capabilities,Contracts,Inbox,Keys,Storage)&Account>())`,
+				// TODO: fix type
+				`flow.AccountCapabilityControllerIssued(id: 2, address: 0x0000000000000001, type: Type<&auth(Capabilities,Contracts,Inbox,Keys,Storage)&Account>())`,
```
turbolent, line 1927: It looks like there's a bug in the link migration and the test cases for it. I'll fix this in a separate PR

migrations/capcons/migration_test.go
```diff
@@ -1436,6 +1517,10 @@ func TestLinkMigration(t *testing.T) {
 					capabilityID: 2,
 				},
 			},
+			expectedEvents: []string{
```
SupunS, line 1520: Are these events get emitted during the actual migration as well?
(was thinking if that could be an overhead for the migration)

SupunS (approved):
LGTM!

migrations/capcons/migration_test.go
turbolent: In the tests here, I'm collecting the events for the assertions. In the real migration code in flow-go, we can just stub out `CapabilityControllerIssueHandler`'s `EmitEvent` and make it a noop.

PR #3465, opened 2024-07-12
Changed data type of account key index to uin32

janezpodhostnik:
Work towards onflow/flow-go#6204

## Description

Changed data type of account key index to uin32.

There are still a few casts in the code, that I did not know how to get rid of.

______

### DIFF
--- a/runtime/stdlib/account.go
+++ b/runtime/stdlib/account.go
@@ -657,7 +657,7 @@ func newAccountKeysAddFunction(
 
 type AccountKey struct {
 	PublicKey *PublicKey
-	KeyIndex  int
+	KeyIndex  uint32
 	Weight    int
 	HashAlgo  sema.HashAlgorithm
 	IsRevoked bool
@@ -669,8 +669,8 @@ type AccountKeyProvider interface {
 	BLSPoPVerifier
 	Hasher
 	// GetAccountKey retrieves a key from an account by index.
-	GetAccountKey(address common.Address, index int) (*AccountKey, error)
-	AccountKeysCount(address common.Address) (uint64, error)
+	GetAccountKey(address common.Address, index uint32) (*AccountKey, error)
+	AccountKeysCount(address common.Address) (uint32, error)
 }
 
 func newAccountKeysGetFunction(
@@ -699,7 +699,7 @@ func newAccountKeysGetFunction(
 				var err error
 				var accountKey *AccountKey
 				errors.WrapPanic(func() {
-					accountKey, err = provider.GetAccountKey(address, index)
+					accountKey, err = provider.GetAccountKey(address, uint32(index))
 				})
 
 				if err != nil {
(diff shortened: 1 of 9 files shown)


runtime/stdlib/account.go
```diff
@@ -699,7 +699,7 @@ func newAccountKeysGetFunction(
 				var err error
 				var accountKey *AccountKey
 				errors.WrapPanic(func() {
-					accountKey, err = provider.GetAccountKey(address, index)
+					accountKey, err = provider.GetAccountKey(address, uint32(index))
```
janezpodhostnik, line 702: not sure how to do the cast here safely.

runtime/stdlib/account.go
```diff
@@ -894,7 +894,7 @@ func newAccountKeysRevokeFunction(
 				var err error
 				var accountKey *AccountKey
 				errors.WrapPanic(func() {
-					accountKey, err = handler.RevokeAccountKey(address, index)
+					accountKey, err = handler.RevokeAccountKey(address, uint32(index))
```
janezpodhostnik, line 897: same here.

turbolent:
I had a look at the referenced issue and it sort of makes sense in flow-go. However, do we want to enforce limits of Flow in Cadence? It would be nice to keep Cadence agnostic of particular chain-limits. Even without the proposed changes, it should be possible to check/reject invalid indices.

janezpodhostnik:
> It would be nice to keep Cadence agnostic of particular chain-limits

does this mean you would prefer cadence still having a `uint64` for the key index? `Uint64` seems like a big data type for key indexes.

turbolent:
Maybe it's good to discuss this a bit more synchronously, maybe as part of the next implementation sync? https://www.notion.so/flowfoundation/Cadence-Implementation-Sync-Notes-Agenda-cf6890eb8a9f4692a8c08611856a5af5?pvs=4#e78f967d7c814ea8b0178e83d39b2e4b. Can have it earlier, don't need to wait for a week. Can also have a separate sync about it

runtime/stdlib/account.go
turbolent: Fixed in [63abfd5](https://github.com/onflow/cadence/pull/3465/commits/63abfd5ec7b0d11bf142eead83591bf2e6c83ebb)

runtime/stdlib/account.go
turbolent: Fixed in [63abfd5](https://github.com/onflow/cadence/pull/3465/commits/63abfd5ec7b0d11bf142eead83591bf2e6c83ebb)

turbolent:
@janezpodhostnik As discussed, I fixed the Cadence `Int` to Go `uint32` conversion in [63abfd5](https://github.com/onflow/cadence/pull/3465/commits/63abfd5ec7b0d11bf142eead83591bf2e6c83ebb).

@SupunS PTAL 🙏

turbolent (approved):

SupunS (approved):

PR #3471, opened 2024-07-16
Update Fungible token transfer benchmark

SupunS:
## Description

Update the benchmark so that we can compare the benchmarks between Cadence versions (`v0.42` vs `v1.0`)

Updated to the latest code of FT contract and the associated transactions, taken from https://github.com/onflow/flow-core-contracts. Had to modify the FT contract a bit to remove the dependency to `MetadataViews` and `Burner` Contracts (removed some methods), as those are not needed for the token transfer transaction.

______

### DIFF
--- a/runtime/ft_test.go
+++ b/runtime/ft_test.go
@@ -19,8 +19,6 @@
 package runtime_test
 
 import (
-	"encoding/hex"
-	"fmt"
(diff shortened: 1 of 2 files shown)


turbolent (approved):
Nice! Thanks for updating these

PR #3476, opened 2024-07-18
adding report for TN migration Jul 17

j1010001:


SupunS:
Hmm, seems theres a race condition in test. Not from this PR, but rather something that is already there in the existing code.
I'm going to re-trigger the CI, so we can get this PR in. 
The CI log with the errors can be found here: [CI_test_log.txt](https://github.com/user-attachments/files/16285251/7_Test.txt)

Issue #3480, opened 2024-07-19
Handle broken contracts


turbolent:
### Issue to be solved

After the update to Cadence 1.0 some contracts will be broken.

It should be possible to keep using them at least in a read-only form.

### Suggested Solution

https://www.notion.so/flowfoundation/Keep-values-with-types-of-broken-contracts-working-7a57ddf83a50456da6851ed1f65e26a9?pvs=4#c69daabfde7a4134be73f70b1ffea58f

turbolent:
FLIP: https://github.com/onflow/flips/pull/283

j1010001:
Remaining - handling of non-fungible token collections

PR #3482, opened 2024-07-19
Handle broken contracts

turbolent:
Work towards #3480 

## Description

Implements the program recovery as described in FLIP https://github.com/onflow/flips/pull/283

______


runtime/ft_test.go
```diff
@@ -766,3 +768,263 @@ func BenchmarkRuntimeFungibleTokenTransfer(b *testing.B) {
 
 	utils.RequireValuesEqual(b, nil, mintAmountValue, sum)
 }
+
+// TODO:
```
turbolent, line 772: The goal is to uncomment and use this version of the ExampleToken and have the transaction still work

SupunS (approved):
Nice!

runtime/ft_test.go
```diff
@@ -766,3 +770,408 @@ func BenchmarkRuntimeFungibleTokenTransfer(b *testing.B) {
...
+	vaultDeclaration := contractDeclaration.Members.CompositesByIdentifier()[fungibleTokenVaultTypeIdentifier]
+	if vaultDeclaration == nil {
+		return false
+	}
+
+	// Check if the Vault resource has a balance field
+	balanceFieldDeclaration := getField(vaultDeclaration, fungibleTokenVaultTypeBalanceFieldName)
+	if balanceFieldDeclaration == nil {
+		return false
+	}
```
SupunS, line 909: Do we need to check these? Can we assume if a contract is deployed, then it must be a valid contract (and hence, given it conforms to the FungibleToken, it should have all these fields)?

runtime/ft_test.go
```diff
@@ -766,3 +770,408 @@ func BenchmarkRuntimeFungibleTokenTransfer(b *testing.B) {
...
+	const transaction = `
+      import FungibleToken from 0x1
+      import ExampleToken from 0x1
+
+      transaction {
+
+          let vault: @ExampleToken.Vault
+
+          prepare(signer: auth(LoadValue) &Account) {
+              self.vault <- signer.storage.load<@ExampleToken.Vault>(from: /storage/exampleTokenVault)!
```
SupunS, line 1147: Could we add some more operations to the test? e.g:
- getting the type (`self.vault.getType()`)
- casting to `FungibleToken.Vault`
- Storing the read value back to storage
- Calling a function on the `self.vault`
- etc.

runtime/ft_test.go
turbolent: Yes, I guess we could maybe make these checks more lenient. Not 100% sure if that could introduce loopholes

runtime/ft_test.go
turbolent: Good idea! Will add these 👍

turbolent: Added in [7197c69](https://github.com/onflow/cadence/pull/3482/commits/7197c6908705c98e8cc791c342efb2da6cfe9c72) and [acd782a](https://github.com/onflow/cadence/pull/3482/commits/acd782a9a053181c46194063913a0bbb81f46abb)

turbolent:
For anyone looking at this PR: This PR only adds support for the program recovery functionality as a runtime interface function. It does NOT provide an implementation. An implementation of the function is only provided for demonstratation / testing purposes in a test case. Merging the PR does not change any behaviour for Cadence programs.

PR #3494, opened 2024-07-25
Return user error when CCF encodes attachment field

fxamacker:
The CCF encoder currently requires field values and types to match.

However, composite values that have attachment field don't satisfy this requirement because:
- composite field values INCLUDE attachment field values
- composite field types EXCLUDE attachment field types

Currently, the encoder rejects this mismatch with a panic.

This commit makes CCF encoder return a user error rather than panic for the caller to handle when this mismatch is encountered.

Closes https://github.com/dapperlabs/cadence-internal/issues/241
NOTE:  The problem fixed by this PR is related to a feature that is not yet deployed to mainnet.
______


turbolent (approved):
Great work! 👏

Issue #824, opened 2021-04-21
AuthAccount link should return linked path capability


bluesign:
### Problem

```
var cap = acct.link<&String>(/private/secret, target: /storage/secret)!
log(cap)

```

Returns: 

`Capability<&String>(address: 0x1, path: /storage/secret)`

I guess it should return: (otherwise unlink still allowing access)

`Capability<&String>(address: 0x1, path: /private/secret)`

PR #3503, opened 2024-07-30
Migrate capability values targeting storage paths

turbolent:
## Description

Capabilities may incorrectly target a storage path, due to an incorrect implementation of the `link` function in an early version of Cadence (#824). This bug was fixed in #865, and subsequent capabilities created by the link function have correct targets, but state of networks (e.g. TN and MN) may still store capabilities with incorrect targets.

Such incorrect capabilities can still be borrowed currently, so they need to be migrated properly.

Migrate such capabilities by issuing a new storage capability controller.

Also, include the capability ID in the report for migrated path capability values, as now the capability ID might have been issued during the capability value migration, and there was no prior link to cap con migration which can be cross-referenced.

______


migrations/capcons/capabilitymigration.go
```diff
@@ -67,65 +70,107 @@ func (m *CapabilityValueMigration) Migrate(
...
+	inter *interpreter.Interpreter,
 	_ migrations.ValueMigrationPosition,
 ) (
 	interpreter.Value,
 	error,
 ) {
 
+	// Migrate path capabilities to ID capabilities
+	if pathCapabilityValue, ok := value.(*interpreter.PathCapabilityValue); ok { //nolint:staticcheck
+		return m.migratePathCapabilityValue(pathCapabilityValue, storageKey, inter)
```
turbolent, line 82: Diff looks a bit odd at the bottom of the file, but all that changed is refactoring the switch, which only has single case, into a separate function.

SupunS (approved):

PR #3505, opened 2024-07-31
Add `Type.isRecovered` field

turbolent:
Work towards #3480 

## Description

Feedback on onflow/flips#283 has been that it would be great to have a way to determine if a type (of a value) has been recovered, i.e. if the type was defined in a recovered program. For example, this allows avoiding function calls, which will always fail (abort execution) for such values.

Add a new field `isRecovered: Bool` to the Cadence type `Type` (run-time types), which is true iff the type is defined in a recovered program.

Implement the feature by adding a new `IsRecovered` field to the elaboration, which is set if program recovery (`RecoverProgram`) succeeds.

______

### DIFF
--- a/runtime/environment.go
+++ b/runtime/environment.go
@@ -505,6 +505,8 @@ func (e *interpreterEnvironment) parseAndCheckProgramWithRecovery(
 		return program, elaboration, err
 	}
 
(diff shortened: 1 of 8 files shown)


SupunS (approved):
Good idea!

Issue #308, opened 2020-08-10
Retain comments in AST


turbolent:
### Issue To Be Solved

Currently the AST only retains comments for declarations if they are docstrings (`///` line comments, a and `/**` block comments).

To be able to use the AST for formatting, it must retain also non-docstring comments.
 
Retain comments for:

- Declarations
- Statements
- Expressions 

### Suggest A Solution

- [ ] Add AST elements for comments:
  - [ ] Block comment
  - [ ] Line comment
- [ ] Add a leading/trailing comments property
  - [ ] Declarations: Line and block comments
  - [ ] Statements: Line and block comments
  - [ ] Expressions: Block comments
- [ ] Add comments to AST elements when parsing
  - [ ] Declarations
  - [ ] Statements
  - [ ] Expressions

SupunS:
I think we may have to capture the comments at a further granular level. Because a comment can exist at any place where whitespace can occur, not only at declarations/statements/expressions level, and we would have to retain those as well.
e.g. ` let /*Im a comment*/ x: 5`

A solution would be to:
- Capture doc comments only at the declaration level. (we can even omit this, and only go with the below one)
- Capture other comments (and any trivia including whitespace, newlines, etc, if needed) at the token level.

Roslyn (C#) compiler follows a similar approach: https://github.com/dotnet/roslyn/blob/main/docs/wiki/Roslyn-Overview.md#syntax-trivia
They also divide the whitespace/comments between two consecutive tokens, so that comments are attached to the correct node/token. And moving that node around (code organizing/formatting), would also move the comment along with it.

e.g: See the bellow comments, which is the 'usual' way followed by most people.
```rust
// (1) The comment immediately before the decl is usually about 'x' decl
let x: 5  // (2) A comment on the same line also talks about 'x' decl (or about 5)
// (3) Similarly, this is a comment about 'y' decl
let y: 8  // (4) This is also a comment about 'y' decl
```
In the above, comments (1) and (2) would be a part of x-var-decl. Moving `let x: 5` around would also move the comments (1) and (2) along with it.
Similarly, (3) and (4) would be a part of y-var-decl.

lkadian:
hi @SupunS Great comment, agree with your proposed solution.
Have you started work on this issue? I ask this because I had actually worked on this but never got around to completing it. I was planning to put up a PR soon, so would just like to confirm if you are also working on it so that we can avoid duplicate work!

SupunS:
[...]

Issue #243, opened 2020-07-17
Remove old parser


turbolent:
## Context

Once the new parser is proven stable, remove the old one.

## Definition of Done

- [x] Use the new parser in the REPL
- [x] Remove the old parser package

PR #3509, opened 2024-08-04
Retain comments in AST

bartolomej:
## Description

POC for preserving comments in Cadence AST. When complete, this should enable us to make a fully working a pretty printer for Cadence.

This is a follow-up to my discussion in Discord some time ago: https://discord.com/channels/613813861610684416/1108479699732152503/1240749220357607475

Closes https://github.com/onflow/cadence/issues/308

## Design

I decided to start tracking trivia tokens in the form of leading/trailing trivia structs attached to Token's as discussed in https://github.com/onflow/cadence/issues/308#issuecomment-809040244.

I believe this could simplify some parsing logic, as the lexer already computes which tokens are associated with which comments + it removes the need to skip comments when parsing.

At the AST level, I added leading/trailing comments fields to some AST nodes. The other trivia types such as spaces or newlines are just discarded for now, as I don't think we'd need them. Maybe that also means we don't even need to track them at the lexer level, but I still think that is a good think for checking correctness in unit testing if nothing else.

Note that I've currently implemented this only for FunctionDeclaration and Block AST nodes, but I'll keep expanding that to other AST nodes.

## Notes

- This change will touch quite a lot of the parsing code, as we'd need to update how we parse declaration/statement/expression (and possibly other) AST nodes to attach leading/trailing comments to them. These changes shouldn't be difficult to implement but are pretty sensitive regardless, so updating and expanding the test suite will be important.
- We should also test that the comments in the real-world code on mainnet/testnet are preserved as expected + possibly add a runtime check to see if all the comments are correctly preserved when doing prettifying, similar as it's done in prettier: https://github.com/prettier/prettier/blob/251ecabbdee509954f7b0d33b38da9ec4ae93ad8/src/main/comments/print.js#L206-L222
- I'm tracking some of the leftover work with `TODO(preserve-comments)` in code
- I'm not sure if old parser is still being used at all. If not, should that package be removed from source code (I saw there was already some related work here, but the source wasn't removed yet: https://github.com/onflow/cadence/issues/243)? For now, I'm updating it so that the tests pass.

## Definition of Done
* [x]  Add AST elements for comments:
	* [x]  Block comment
	* [x]  Line comment
* [ ]  Add a leading/trailing comments property
	* [ ]  Declarations: Line and block comments
	* [ ]  Statements: Line and block comments
	* [ ]  Expressions: Block comments
* [ ]  Add comments to AST elements when parsing
	* [ ]  Declarations (WIP)
	* [ ]  Statements
	* [ ]  Expressions
* [ ]  Tests

## Description

______


RZhang05:
Is this still being worked on? If not, I may try to finish what remains

bartolomej:
> Is this still being worked on? If not, I may try to finish what remains

Hi, yes I'm still working on this and plan to finish it including some follow up work, but I've just been busy with other stuff in the past few weeks.

SupunS:
hey @bartolomej , we recently restructured the cadence repo and as part of that, moved all the subdirectories under `runtime` package (e.g: `runtime/ast`, `runtime/interpreter`, etc) to the root/top level. Please take a pull and update the branch. Github should be able to identify the changes to the moved files, and correctly map them to the new locations. So hopefully there shouldn't be much conflicts.

Let me know if you run into any issues/conflicts, I can help with resolving those.

parser/function.go
```diff
@@ -197,14 +222,20 @@ func parseParameter(p *parser, expectDefaultArgument bool) (*ast.Parameter, erro
...
 		typeAnnotation,
 		defaultArgument,
-		startPos,
+		startToken.StartPos,
+		ast.Comments{
+			Leading: append(
+				startToken.Comments.PackToList(),
+				colonToken.Comments.PackToList()...,
+			),
+		},
```
turbolent: Just to clarify my understand: If we have comments for the argument label / parameter name and the colon, then pack them together and store them for the whole parameter. 

So if we for example have 
```cadence
fun test(foo: Int) { // ...
```
and it is commented like so:

```cadence
fun test(/* 1 */ foo /* 2 */ : /* 3 */ Int) { // ...
```

we will group and store `1 2 3` as the leading comment for the parameter, right?

I think this is OK for now. Eventually we might also want to include the comments when there is an explicit argument label, e.g. in:
```cadence
fun test(foo bar: Int) { // ...
```
(this is a function where `bar` is the parameter name available in `test`, and which must be called with argument label `foo:`)

Eventually we might also want track the comments for the individual parts (argument label, parameter name, colon, type) and not bundle them, but we can do that later, as it's unlikely to be needed much.

ast/function_declaration_test.go
```diff
@@ -585,14 +587,17 @@ func TestSpecialFunctionDeclaration_MarshalJSON(t *testing.T) {
...
+				Leading: []*Comment{NewComment(nil, []byte("///test"))},
+			},
+			StartPos: Position{Offset: 34, Line: 35, Column: 36},
 		},
 	}
 
 	actual, err := json.Marshal(decl)
 	require.NoError(t, err)
 
+	// TODO(preserve-comments): Do we need to include comments in the JSON AST?
```
turbolent: I'd expect the JSON output to have `Comments`, given it was added to the Go struct 🤔

parser/comment_test.go
```diff
@@ -193,55 +193,145 @@ func TestParseBlockComment(t *testing.T) {
 			errs,
 		)
 	})
+}
 
-	t.Run("invalid content", func(t *testing.T) {
-
-		t.Parallel()
+func TestParseWhileStatementComment(t *testing.T) {
```
turbolent: Should this maybe go into `statement_test.go`?

parser/lexer/lexer_test.go
```diff
@@ -1644,53 +2657,14 @@ func TestLexBlockComment(t *testing.T) {
...
+						SpaceOrError: MissingCommentEndError{},
 						Range: ast.Range{
-							StartPos: ast.Position{Line: 1, Column: 18, Offset: 18},
-							EndPos:   ast.Position{Line: 1, Column: 19, Offset: 19},
+							StartPos: ast.Position{Line: 1, Column: 20, Offset: 20},
+							EndPos:   ast.Position{Line: 1, Column: 20, Offset: 20},
 						},
 					},
-					Source: "*/",
+					Source: "\000",
```
turbolent: How come the source is the null character here?

parser/type.go
```diff
@@ -179,15 +181,16 @@ func parseNominalTypeRemainder(p *parser, token lexer.Token) (*ast.NominalType,
 		// Skip the dot
 		p.next()
 
-		nestedToken := p.current
+		tok := p.current
+		nestedToken = &tok
```
turbolent: Is it OK to override this on each loop iteration? Is the idea to only keep track of the last?

turbolent (commented):
Great work! Had another pass through the code with some suggestions. Will review the tests next

parser/function.go
bartolomej: @turbolent Yes, you are exactly right! 

I decided to simplify comments tracking for such (I believe) edge cases and just bundle them together, instead of potentially having to introduce separate fields/structures for each syntax token in the AST, which would introduce some more compexity.

I agree that this is probably ok for now. Better to come back and complicate it more later when/if needed.

parser/lexer/lexer.go
```diff
@@ -384,16 +490,6 @@ func IsIdentifierRune(r rune) bool {
 		r == '_'
 }
 
-func IsValidIdentifier(s string) bool {
```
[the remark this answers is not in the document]
bartolomej: Ooh yeah that make sense, will undo this change!

I guess I assumed it was likely used by some old parsing logic, and just removed it without thinking about it too much.

Maybe I'll add a comment around that, even though such mistakes would be caught by some kind of integration tests sooner or later.

ast/function_declaration_test.go
bartolomej: It was added to the struct, but I haven't explicitly updated the json serialization logic (MarshalJSON) of AST nodes yet: https://github.com/bartolomej/cadence/blob/14834e4c604b3be9a6f07e25a11876ac9944163e/ast/function_declaration.go#L197-L216

parser/type.go
bartolomej: Yeah so this was previously an inline variable. I moved it out of the loop precisely because of that - to have access to the last token outside the loop.

I added a comment around that to the `nestedToken` var, but I can clarify it further a bit.

parser/lexer/lexer_test.go
bartolomej: I guess I thought this is expected, since the error occurs at EOF (which doesn't have a position within the source code).

But I guess the idea here is to track the location of the last (unconsumed) token, which makes sense actually, 

Will update this.

PR #3510, opened 2024-08-08
Add migration report for aug 7

zhangchiqing:


j1010001 (approved):

bluesign:
> Stats: 745 contracts staged, 702 successfully upgraded, 43 failed to upgrade (4 failed with FiatToken)

not sure if this is correct.

SupunS (approved):

SupunS:
@bluesign 
> not sure if this is correct.

Why do you think so? The number seems consistent with the previous stats 🤔

bluesign:
@SupunS hmm I compared with my staging.dnz.dev ( then I searched for  ✅ in .md file )

SupunS:
oh I see. I'll see why the numbers are different. Thanks for flagging!

j1010001:
I checked again and @bluesign is right, I can see 797 staged, 41 errors, 756 successfuly migrated. I'll update the report

PR #3514, opened 2024-08-09
Report and skip storage caps with missing borrow type

SupunS:
## Description

If the storage capability was issued without a type, then do not migrate such values. Report and skip it.
This is an edge-case that can be improved later.

______

### DIFF
--- a/migrations/capcons/capabilitymigration.go
+++ b/migrations/capcons/capabilitymigration.go
@@ -37,16 +37,6 @@ type CapabilityMigrationReporter interface {
 		accountAddress common.Address,
 		addressPath interpreter.AddressPath,
 	)
-	MissingBorrowType(
-		accountAddress common.Address,
-		addressPath interpreter.AddressPath,
-	)
-	IssuedStorageCapability(
-		accountAddress common.Address,
-		addressPath interpreter.AddressPath,
-		borrowType *interpreter.ReferenceStaticType,
-		capabilityID interpreter.UInt64Value,
-	)
 }
 
 // CapabilityValueMigration migrates all path capabilities to ID capabilities,
--- a/migrations/capcons/capabilitymigration.go
+++ b/migrations/capcons/capabilitymigration.go
@@ -37,6 +37,16 @@ type CapabilityMigrationReporter interface {
 		accountAddress common.Address,
 		addressPath interpreter.AddressPath,
 	)
+	MissingBorrowType(
+		accountAddress common.Address,
+		addressPath interpreter.AddressPath,
+	)
+	IssuedStorageCapability(
+		accountAddress common.Address,
+		addressPath interpreter.AddressPath,
+		borrowType *interpreter.ReferenceStaticType,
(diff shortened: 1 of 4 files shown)


migrations/capcons/capabilitymigration.go
```diff
@@ -37,6 +37,16 @@ type CapabilityMigrationReporter interface {
...
+	MissingBorrowType(
+		accountAddress common.Address,
+		addressPath interpreter.AddressPath,
+	)
+	IssuedStorageCapability(
+		accountAddress common.Address,
+		addressPath interpreter.AddressPath,
+		borrowType *interpreter.ReferenceStaticType,
+		capabilityID interpreter.UInt64Value,
+	)
```
turbolent, line 49: Maybe move these into a separate interface, this reporter is for the capability value migration, whereas the new functions are for the capabilities issuing function, a related, but different "migration"

migrations/capcons/capabilitymigration.go
```diff
@@ -37,6 +37,16 @@ type CapabilityMigrationReporter interface {
 		accountAddress common.Address,
 		addressPath interpreter.AddressPath,
 	)
+	MissingBorrowType(
+		accountAddress common.Address,
+		addressPath interpreter.AddressPath,
+	)
+	IssuedStorageCapability(
```
turbolent, line 44: Maybe rename `IssuedStorageCapability` to indicate a capability *controller* was issued, e.g. `IssuedStorageCapabilityController`

migrations/capcons/storagecapmigration.go
```diff
@@ -73,14 +73,26 @@ func (m *StorageCapMigration) CanSkip(valueType interpreter.StaticType) bool {
 
 func IssueAccountCapabilities(
```
turbolent, line 74: I wonder if we should move / replicate this function in flow-go, so we have more and easier control there over how to handle cases like e.g. a missing borrow type. Once we decide to e.g. no longer skip, we'd have to update this logic here and tag yet another Cadence release. Not sure, just a side-note

turbolent (approved):
Nice!

migrations/capcons/capabilitymigration.go
SupunS: Refactored in https://github.com/onflow/cadence/pull/3514/commits/97ddc3129f3cb6a6ed90f24887f77c21042ddcd1

migrations/capcons/capabilitymigration.go
SupunS: Refactored in https://github.com/onflow/cadence/pull/3514/commits/97ddc3129f3cb6a6ed90f24887f77c21042ddcd1

migrations/capcons/storagecapmigration.go
SupunS: yeah had the same thought - it would definitely be easier. But then again, it's like any-other migration (cap-con, link migration, etc), the migration logic is in Cadence, and only the flow-go makes use of it.

PR #3522, opened 2024-08-13
Improve storage path capability migration reporting

SupunS:
## Description

Include the storage location/path where the capability itself is stored, in addition to the target path.

______

### DIFF
--- a/migrations/capcons/capabilities.go
+++ b/migrations/capcons/capabilities.go
@@ -19,6 +19,7 @@
 package capcons
 
 import (
+	"fmt"
 	"sync"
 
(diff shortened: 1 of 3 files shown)


migrations/capcons/capabilities.go
```diff
@@ -28,18 +29,33 @@ import (
 type AccountCapability struct {
 	Path       interpreter.PathValue
```
turbolent, line 30: Maybe rename this to e.g. `TargetPath` to avoid this getting confused as e.g. the source path

turbolent (approved):

PR #3524, opened 2024-08-13
Add capability stored path info to the report

SupunS:
## Description

Follow up on https://github.com/onflow/cadence/pull/3522. Add the collected additional info to the report.

______

### DIFF
--- a/migrations/capcons/capabilitymigration.go
+++ b/migrations/capcons/capabilitymigration.go
@@ -39,7 +39,8 @@ type CapabilityMigrationReporter interface {
 	)
 	MissingBorrowType(
 		accountAddress common.Address,
-		addressPath interpreter.AddressPath,
+		targetPath interpreter.AddressPath,
+		storedPath Path,
 	)
 }
 
--- a/migrations/capcons/capabilitymigration.go
+++ b/migrations/capcons/capabilitymigration.go
@@ -38,9 +38,8 @@ type CapabilityMigrationReporter interface {
 		addressPath interpreter.AddressPath,
 	)
 	MissingBorrowType(
-		accountAddress common.Address,
 		targetPath interpreter.AddressPath,
-		storedPath Path,
+		storedPath interpreter.AddressPath,
 	)
 }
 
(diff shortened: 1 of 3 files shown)


migrations/capcons/storagecapmigration.go
```diff
@@ -129,18 +131,14 @@ func IssueAccountCapabilities(
 			borrowType = capabilityBorrowType.(*interpreter.ReferenceStaticType)
 
 		} else {
-			if _, _, ok := untypedCapabilityMapping.Get(addressPath); ok {
-				continue
-			}
-
```
SupunS, line 135: Removed this, because me might want to get a count on how many such capabilities are there, even if they point to the same storage location.

migrations/capcons/capabilitymigration.go
```diff
@@ -39,7 +39,8 @@ type CapabilityMigrationReporter interface {
 	)
 	MissingBorrowType(
 		accountAddress common.Address,
-		addressPath interpreter.AddressPath,
+		targetPath interpreter.AddressPath,
+		storedPath Path,
```
turbolent, line 43: Maybe combine `accountAddress` and `storedPath` into a `interpreter.AddressPath`. Same below

migrations/capcons/storagecapmigration.go
turbolent: I don't think we can remove this, as otherwise one controller per capability is issued.
Maybe add separate reporting before the bail-out

SupunS: oh I see!

SupunS: Updated it to always log (https://github.com/onflow/cadence/pull/3524/commits/0f12301ee4a7ee4cb8db24aca684d66a6e17440b). Then any cap that did not get inferred can be found by diffing the two logs entries.

migrations/capcons/capabilitymigration.go
SupunS: Updated in 0f12301ee4a7ee4cb8db24aca684d66a6e17440b

turbolent (approved):
Nice!

PR #3532, opened 2024-08-15
Check if CBOR tag number is reserved by atree before using it to encode Cadence values

fxamacker:
Using tag numbers reserved for atree internal use can result in data in storage that cannot be decoded.

To reduce risks, this PR checks if CBOR tag number is reserved by atree for internal use before using it to encode Cadence values as elements in atree containers.

As of Aug 15, 2024:
- Atree reserves CBOR tag numbers [240, 255] for internal use.
- Cadence uses CBOR tag numbers [128, 230] for encoding Cadence values as elements in atree containers.

When new tag numbers are needed in Atree, we will use higher tag numbers first from Atree's reserved range.  In Cadence, we will use lower tag numbers first from its own (different) reserved range. This allows Atree and Cadence more flexibility in case we need to revisit allocation of adjacent unused ranges for Atree and Cadence.

While at it, also used new function `SlabID.Address()` to simply related code.

______

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/kr/pretty v0.3.1
 	github.com/leanovate/gopter v0.2.9
 	github.com/logrusorgru/aurora/v4 v4.0.0
-	github.com/onflow/atree v0.8.0-rc.5
+	github.com/onflow/atree v0.8.0-rc.5.0.20240815140907-141ec5ca3ce2
 	github.com/rivo/uniseg v0.4.4
 	github.com/schollz/progressbar/v3 v3.13.1
 	github.com/stretchr/testify v1.9.0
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/kr/pretty v0.3.1
 	github.com/leanovate/gopter v0.2.9
 	github.com/logrusorgru/aurora/v4 v4.0.0
-	github.com/onflow/atree v0.8.0-rc.5
+	github.com/onflow/atree v0.8.0-rc.5.0.20240815140907-141ec5ca3ce2
```
turbolent, line 16: Maybe tag a new version

turbolent (approved):
Good idea, nice!

Issue #3536, opened 2024-08-15
Cadence migration is non-deterministic since crescendo RC3.


j1010001:
RC3 produces consistent results. RC4 crashes on MN state. RC5 produces non-deterministic results. So [somewhere between RC3-RC5](https://github.com/onflow/flow-go/compare/v0.37.0-crescendo-RC3...v0.37.0-crescendo-RC5) we have added source of non-determinism to the migration.

fxamacker:
Finished initial investigation and extracted 3 files with payloads that can be used by Cadence team for creating a reproducer for the discrepency with `IDCapabilityValue.ID`.

If creating and resolving the reproducer doesn't fix the non-determinism, then I can look for additional issues. :pray:

PR #3542, opened 2024-08-20
Make capabilities iteration ordered

SupunS:
Closes https://github.com/onflow/cadence/issues/3536

## Description

A more real-world test is added in https://github.com/onflow/flow-go/pull/6375

______

### DIFF
--- a/migrations/capcons/capabilities.go
+++ b/migrations/capcons/capabilities.go
@@ -19,6 +19,7 @@
 package capcons
 
 import (
+	"cmp"
 	"fmt"
 	"strings"
 	"sync"
@@ -41,7 +42,8 @@ type Path struct {
 }
 
 type AccountCapabilities struct {
-	Capabilities []AccountCapability
+	capabilities []AccountCapability
+	sortOnce     sync.Once
 }
 
 func (c *AccountCapabilities) Record(
@@ -50,8 +52,8 @@ func (c *AccountCapabilities) Record(
 	storageKey interpreter.StorageKey,
 	storageMapKey interpreter.StorageMapKey,
 ) {
-	c.Capabilities = append(
-		c.Capabilities,
+	c.capabilities = append(
+		c.capabilities,
 		AccountCapability{
 			TargetPath: path,
 			BorrowType: borrowType,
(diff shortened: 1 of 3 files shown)


migrations/capcons/capabilities.go
```diff
@@ -60,6 +63,34 @@ func (c *AccountCapabilities) Record(
...
+
+			if pathA.Domain == pathB.Domain {
+				return strings.Compare(pathA.Identifier, pathB.Identifier)
+			}
+
+			if pathA.Domain < pathB.Domain {
+				return -1
+			}
+
+			return +1
```
turbolent, line 83: Can maybe use Go 1.22's new `cmp.Or` and `cmp.Compare`, see e.g. https://pkg.go.dev/cmp#example-Or-Sort

migrations/capcons/capabilities.go
```diff
@@ -60,6 +63,34 @@ func (c *AccountCapabilities) Record(
 	)
 }
 
+func (c *AccountCapabilities) ForEach(
```
turbolent, line 66: Maybe rename the function to indicate that it also has the side-effect of sorting, and maybe also add a comment to make it clear

turbolent (approved):
Nice!

migrations/capcons/capabilities.go
SupunS: good idea! added in https://github.com/onflow/cadence/pull/3542/commits/f712fdea18aa174f9073a26156c7d101f891faba

migrations/capcons/capabilities.go
SupunS: added in https://github.com/onflow/cadence/pull/3542/commits/f712fdea18aa174f9073a26156c7d101f891faba

migrations/capcons/capabilities.go
```diff
@@ -60,6 +65,38 @@ func (c *AccountCapabilities) Record(
...
+	c.sort()
+	for _, accountCapability := range c.capabilities {
+		if !f(accountCapability) {
+			return
+		}
+	}
+}
+
+func (c *AccountCapabilities) sort() {
+	c.sortOnce.Do(
```
turbolent, line 82: Maybe consider resetting the `sync.Once` in `Record` to avoid surprises when `Record` is called in between two `ForEachSorted` calls.

SupunS: hmm maybe just keeping a boolean flag might be better in that case? resetting `sync.Once` every time `Record` is called is an overkill.

turbolent (approved):
👍

PR #3554, opened 2024-08-28
Fix contract update error when old program has errors

turbolent:
## Description

Fixes the error encountered in https://testnet.flowdiver.io/tx/5480ce5f9f81a976be1d2232c4d33ebef02f02204a6e4dbace8319d4917c819a.

When reporting a contract update error, the error is not always for the new program:
When updating, we parse the old program, and even though it was at some point valid, it may no longer is.

Currently the new code is used, and the parser error for the old program is not wrapped with the location of the old program, so instead the location performing the update (most of the time the transaction location) is used and its code is used for error pretty printing. This effectively results in the transaction code being used when pretty printing parsing errors for the old program.

Use the correct code, i.e. the old program's code, and the correct location for the error, i.e the old program's location and not the transaction location.

______

### DIFF
--- a/runtime/contract_update_test.go
+++ b/runtime/contract_update_test.go
@@ -28,6 +28,7 @@ import (
 	. "github.com/onflow/cadence/runtime"
 	"github.com/onflow/cadence/runtime/common"
 	"github.com/onflow/cadence/runtime/interpreter"
+	"github.com/onflow/cadence/runtime/stdlib"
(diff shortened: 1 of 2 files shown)


SupunS (approved):
I have a deja-vu, didn't we already fix this? (or something similar?)

turbolent:
@SupunS maybe? As far as I can remember, we fixed some related issues before. This one is special because of v1.0 is now reporting parsing errors for v0.42 programs

PR #3562, opened 2024-08-31
Handle unmigrated path capabilities and path links in Account.capabilities functions

turbolent:
## Description

Path links which cannot be migrated stay as-is (see `LinkValueMigration.Migrate`, when `getPathCapabilityFinalTarget` returns failure). Path capabilities which cannot be migrated stay as-is (see `CapabilityValueMigration.migratePathCapabilityValue`, when controller ID is missing from mapping).

As result, we cannot assume capability values in general and values stored in public paths are always ID capability values (`IDCapabilityValue`), they might be path capability values (`PathCapabilityValue`) and path link values (`PathLinkValue`).

- Add a `CapabilityValue.Address()` function to easily get the address of `IDCapabilityValue` and `PathCapabilityValue`.

  I didn't do this for the `Type` and `ID` fields, because I realized it would require a lot more code to change. It still would be nice to also add getter functions for those fields, in a follow up PR, later

- Fix the export of path capability values. They don't always have a borrow type.
  
  I noticed this when writing the tests, as publishing a capability emits an event which includes the capability, which leads to an export of the event and the capability it contains.

- Handle path capability values and path link values in `Account.capabilities.get`, `publish`, and `unpublish` 

______

### DIFF
--- a/runtime/convertValues.go
+++ b/runtime/convertValues.go
@@ -713,8 +713,12 @@ func exportPathCapabilityValue(
 	v *interpreter.PathCapabilityValue, //nolint:staticcheck
 	inter *interpreter.Interpreter,
 ) (cadence.Capability, error) {
-	borrowType := inter.MustConvertStaticToSemaType(v.BorrowType)
-	exportedBorrowType := ExportMeteredType(inter, borrowType, map[sema.TypeID]cadence.Type{})
+	var exportedBorrowType cadence.Type
+
+	if v.BorrowType != nil {
+		borrowType := inter.MustConvertStaticToSemaType(v.BorrowType)
+		exportedBorrowType = ExportMeteredType(inter, borrowType, map[sema.TypeID]cadence.Type{})
+	}
 
 	capability := cadence.NewMeteredCapability(
 		inter,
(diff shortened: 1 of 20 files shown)


runtime/convertValues.go
```diff
@@ -713,13 +713,17 @@ func exportPathCapabilityValue(
 	v *interpreter.PathCapabilityValue, //nolint:staticcheck
 	inter *interpreter.Interpreter,
 ) (cadence.Capability, error) {
-	borrowType := inter.MustConvertStaticToSemaType(v.BorrowType)
-	exportedBorrowType := ExportMeteredType(inter, borrowType, map[sema.TypeID]cadence.Type{})
+	var exportedBorrowType cadence.Type
+
+	if v.BorrowType != nil {
```
SupunS, line 718: Do we need to do the same for `IDCapabilityValue` above?

SupunS (approved):

runtime/convertValues.go
turbolent: Not sure 🤔 I assumed that all ID capabilities have a type now, but maybe that's not actually the case?

PR #3568, opened 2024-09-03
Bring changelog up-to-date

turbolent:
## Description

Use github.com/rhysd/changelog-from-release to automatically generate the changelog from the repository's releases.

Might be nice to have this as a GitHub Action.
 
______

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,217 +1,4646 @@
-# v0.19.1 (2021-09-13)
+<a name="v1.0.0-preview.52"></a>
+# [v1.0.0-preview.52](https://github.com/onflow/cadence/releases/tag/v1.0.0-preview.52) - 31 Aug 2024
+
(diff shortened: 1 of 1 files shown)


SupunS (commented):
Could we only have the change-log for the most recent release, and then maybe have a link to the releases section of the github for everything else? Having the full change-log since day-zero can be a really long list, and this file would keep growing. We already have these changes in the release notes for each release.

turbolent:
Though most users and maintainers will likely use the GitHub releases page, it is still valuable to have all release notes inside of the codebase itself. For example, the project might at some point move to another platform.

This is also commonly done in other projects, for example:
- Swift: https://github.com/swiftlang/swift/blob/main/CHANGELOG.md
- Rust: https://github.com/rust-lang/rust/blob/master/RELEASES.md
- Kotlin: https://github.com/JetBrains/kotlin/blob/master/ChangeLog.md

SupunS (approved):

PR #3570, opened 2024-09-03
Add Type.address and Type.contractName

turbolent:
## Description

Extend `Type` by adding two new fields:

- ```cadence
  /// The address of the type, if it was declared in a contract deployed to an account
  let address: Address?
  ```

- ```cadence
  /// The contract name of the type, if it was declared in a contract
  let contractName: String?
  ```

Currently, getting this information requires manually extracting the information from the type identifier, e.g. https://github.com/onflow/flow-evm-bridge/blob/ca95d452aa89ba17ca79776592ff9fd7a9d7f804/cadence/contracts/bridge/FlowEVMBridgeUtils.cdc#L947-L960

______

### DIFF
--- a/runtime/interpreter/value_type.go
+++ b/runtime/interpreter/value_type.go
@@ -172,6 +172,31 @@ func (v TypeValue) GetMember(interpreter *Interpreter, _ LocationRange, name str
 		}
 
 		return AsBoolValue(elaboration.IsRecovered)
+
+	case sema.MetaTypeAddressFieldName:
+		staticType := v.Type
+		if staticType == nil {
+			return Nil
+		}
+
+		location, _, err := common.DecodeTypeID(interpreter, string(staticType.ID()))
+		if err != nil || location == nil {
+			return Nil
+		}
+
+		addressLocation, ok := location.(common.AddressLocation)
+		if !ok {
+			return Nil
+		}
+
+		addressValue := NewAddressValue(
+			interpreter,
+			addressLocation.Address,
+		)
+		return NewSomeValueNonCopying(
+			interpreter,
+			addressValue,
+		)
 	}
 
 	return nil
--- a/runtime/interpreter/value_type.go
+++ b/runtime/interpreter/value_type.go
@@ -181,8 +181,16 @@ func (v TypeValue) GetMember(interpreter *Interpreter, _ LocationRange, name str
 			return Nil
 		}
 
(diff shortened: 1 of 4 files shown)


turbolent:
cc @sisyphusSmiling

sisyphusSmiling:
This would be great!

runtime/interpreter/value_type.go
```diff
@@ -172,6 +174,67 @@ func (v TypeValue) GetMember(interpreter *Interpreter, _ LocationRange, name str
...
+	case sema.MetaTypeAddressFieldName:
+		staticType := v.Type
+		if staticType == nil {
+			return Nil
+		}
+
+		location, _, err := common.DecodeTypeID(interpreter, string(staticType.ID()))
+		if err != nil || location == nil {
+			return Nil
+		}
```
SupunS, line 187: rather than decoding the type-ID, it would be efficient to check whether `v.Type` is either the `CompositeStaticType`/`InterfaceStaticType` and return the locations from the type itself (and nil, otherwise)

runtime/interpreter/value_type.go
```diff
@@ -172,6 +174,67 @@ func (v TypeValue) GetMember(interpreter *Interpreter, _ LocationRange, name str
...
+				common.NewStringMemoryUsage(contractNameLength),
+				func() string {
+					return qualifiedIdentifier[0:contractNameLength]
+				},
+			)
+
+			return NewSomeValueNonCopying(interpreter, contractNameValue)
+
+		default:
+			return Nil
```
SupunS, line 235: shouldn't this return a Go `nil`?

turbolent: 🤔 

The field is optional (`String?`), so we need to ensure to return a Cadence `nil` (i.e. `interpreter.Nil`) in all case (see also the two other failure cases above)

runtime/interpreter/value_type.go
turbolent: Good idea! Changed it here and for the other field below in d90d44960c54be16ec68785ad5946f042ab8df3e

runtime/interpreter/value_type.go
SupunS: ah my bad, I mistook it for the outer switch-case's default value (i.e: non existing member). This makes sense.

SupunS (approved):

PR #3571, opened 2024-09-04
Add Crescendo migration report for mainnet24

j1010001:


dida42-prog:
Downstream dependencies / flow-go: It's failing, which could indicate dependency issues that need to be resolved before approval.

Downstream dependencies / Emulator: It's also failing, suggesting that the emulator is not passing some test or check.

Congratulations on the work, team!

turbolent (approved):

SupunS (approved):

PR #3575, opened 2024-09-09
Change index expression's start position

### LINKED ISSUE #718: Change indexing expression start position to
### Problem

For historic reasons, index expressions' start position is currently the opening square bracket and not the indexed expression's start position.

This leads to incorrect error ranges and thus confusing errors for users.

### Definition of Done

Change the index expression AST node:
- [ ] Add `IndexingStartPos` field to indicate the position of the opening square bracket
- [ ] Change the start position to the start position of the indexed expression

RZhang05:
Closes #718

## Description

This PR adds a test which throws an error involving index expression different from the type mismatch error addressed in #719. Index expression's start position was changed from opening square bracket to indexed expression's start position.

______

### DIFF
--- a/runtime/account_test.go
+++ b/runtime/account_test.go
@@ -2676,6 +2676,49 @@ func TestRuntimeGetAuthAccount(t *testing.T) {
 
 		assert.IsType(t, &sema.NotDeclaredError{}, errs[0])
 	})
+
+	t.Run("update names index expression start position", func(t *testing.T) {
+		t.Parallel()
+
+		rt := NewTestInterpreterRuntime()
+
+		script := []byte(`
+            transaction {
+                prepare(signer: auth(Mutate) &Account) {
+                    signer.contracts.names[0] = "baz"
+                }
+            }
+        `)
+
+		runtimeInterface := &TestRuntimeInterface{
+			OnGetSigningAccounts: func() ([]Address, error) {
+				return []Address{{42}}, nil
+			},
+			OnGetAccountContractNames: func(_ Address) ([]string, error) {
+				return []string{"foo", "bar"}, nil
+			},
+		}
+
+		nextTransactionLocation := NewTransactionLocationGenerator()
+
+		err := rt.ExecuteTransaction(
+			Script{
+				Source: script,
+			},
+			Context{
+				Interface: runtimeInterface,
+				Location:  nextTransactionLocation(),
+			},
+		)
+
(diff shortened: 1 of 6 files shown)


runtime/parser/expression.go
```diff
@@ -1262,7 +1262,7 @@ func defineIndexExpression() {
 				firstIndexExpr,
 				ast.NewRange(
 					p.memoryGauge,
-					token.StartPos,
+					left.StartPosition(),
```
SupunS, line 1265: now that `token` parameter is not used anymore, we can omit declaring this unused parameter by using the underscore (`_`) as the parameter name.
i.e: in the function signature above:
```go
func(p *parser, _ lexer.Token, left ast.Expression) (ast.Expression, error) {
```

runtime/account_test.go
```diff
@@ -2676,6 +2676,49 @@ func TestRuntimeGetAuthAccount(t *testing.T) {
 
 		assert.IsType(t, &sema.NotDeclaredError{}, errs[0])
 	})
+
+	t.Run("update names index expression start position", func(t *testing.T) {
```
SupunS, line 2680: Nice work adding this test!

Tests in the `runtime` package are usually "integration tests" in a way. As in, it setup the environments, create transactions, etc. For this change, given we only need to test the interpreting behaviour, moving this test to the interpreter unit tests under `runtime/tests/interpreter` (`indexing_test.go` file) might be a better option.

SupunS (commented):
First contribution! 👏  Thank you for opening a PR to fix this!

Looks like there are some tests failing, due to the updated position info. Can you please update those tests as well? https://github.com/onflow/cadence/actions/runs/10775433087/job/29882238324?pr=3575
You can also run `make test` from the root directory of the repo, to run all tests locally.

runtime/parser/expression.go
RZhang05: `token` is now omitted.

runtime/account_test.go
RZhang05: Given there are multiple equivalent tests in `runtime/tests/interpreter/interpreter_test.go` which cover the case of errors being thrown using the update index expression positions, this test has been removed.

SupunS: make sense 👍

SupunS (approved):
Nice work! 🎉

turbolent (approved):
Great work on your first PR! 👏

PR #3580, opened 2024-09-11
Unwrap optionals with dynamic cast

### LINKED ISSUE #3183: Dynamic casting should unwrap optional values
### Why / Impact

Developer convenience

### Current Behavior

Dynamic casting (failable casting `as?`, and force casting `as!`) an optional value (potentially type-erased to an existential) to a non-optional type currently fails (e.g. failable casting results in `nil`).

### Expected Behavior

I expected (but perhaps shouldn't have) that an optional value being casted into a subtype it conforms to would unwrap it. I'm really not sure if this is a bug or if it's expected though

### Steps To Reproduce

I've made a repro contract/transactions to demonstrate

https://github.com/austinkline/cadence-burner-bug

### Environment

```markdown
- Cadence version:Version: v1.15.0-cadence-v1.0.0-preview.12, Commit: c582e63baf4581e4cf021a9b1917c198ca5b6065
- Network: emulator
```

RZhang05:
Closes #3183

## Description

Dynamic casting now unwraps optionals following the behaviour of Swift.

______

### DIFF
--- a/runtime/interpreter/interpreter_expression.go
+++ b/runtime/interpreter/interpreter_expression.go
@@ -1337,16 +1337,17 @@ func (interpreter *Interpreter) VisitCastingExpression(expression *ast.CastingEx
 		// so we don't substitute them.
 		valueSemaType := interpreter.SubstituteMappedEntitlements(interpreter.MustSemaTypeOfValue(value))
 		valueStaticType := ConvertSemaToStaticType(interpreter, valueSemaType)
+		isUnwrappable := interpreter.IsUnwrappable(valueStaticType, expectedType)
 		isSubType := interpreter.IsSubTypeOfSemaType(valueStaticType, expectedType)
 
 		switch expression.Operation {
 		case ast.OperationFailableCast:
-			if !isSubType {
+			if !isSubType && !isUnwrappable {
 				return Nil
 			}
 
 		case ast.OperationForceCast:
-			if !isSubType {
+			if !isSubType && !isUnwrappable {
 				locationRange := LocationRange{
 					Location:    interpreter.Location,
 					HasPosition: expression.Expression,
@@ -1361,6 +1362,11 @@ func (interpreter *Interpreter) VisitCastingExpression(expression *ast.CastingEx
 
 		default:
 			panic(errors.NewUnreachableError())
+		}
+
+		if isUnwrappable {
+			// dynamic cast now unboxes optionals
+			value = interpreter.Unbox(locationRange, value)
 		}
 
 		// The failable cast may upcast to an optional type, e.g. `1 as? Int?`, so box
--- a/runtime/interpreter/interpreter_expression.go
+++ b/runtime/interpreter/interpreter_expression.go
@@ -1335,19 +1335,24 @@ func (interpreter *Interpreter) VisitCastingExpression(expression *ast.CastingEx
 		// thus this is the only place where it becomes necessary to "instantiate" the result of a map to its
 		// concrete outputs. In other places (e.g. interface conformance checks) we want to leave maps generic,
 		// so we don't substitute them.
+
+		// if the target is anystruct or anyresource we want to preserve optionals
+		if !(expectedType == sema.AnyStructType || expectedType == sema.AnyResourceType) {
+			// otherwise dynamic cast now always unboxes optionals
+			value = interpreter.Unbox(locationRange, value)
+		}
 		valueSemaType := interpreter.SubstituteMappedEntitlements(interpreter.MustSemaTypeOfValue(value))
 		valueStaticType := ConvertSemaToStaticType(interpreter, valueSemaType)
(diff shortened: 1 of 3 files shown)


runtime/interpreter/interpreter_expression.go
```diff
@@ -1335,6 +1335,12 @@ func (interpreter *Interpreter) VisitCastingExpression(expression *ast.CastingEx
 		// thus this is the only place where it becomes necessary to "instantiate" the result of a map to its
 		// concrete outputs. In other places (e.g. interface conformance checks) we want to leave maps generic,
 		// so we don't substitute them.
+
+		// if the target is anystruct or anyresource we want to preserve optionals
+		if !(expectedType == sema.AnyStructType || expectedType == sema.AnyResourceType) {
```
SupunS, line 1340: We might have to also unwrap `expectedType` type before performing this check.
For e.g: If the targetType is an optional of AnyStruct (`AnyStruct?`), then it will pass the check and will unbox the optional value, which it shouldn't. i.e:
```cadence
let x: Int??? = 42
let y: AnyStruct = x as! AnyStruct?
```

runtime/tests/interpreter/dynamic_casting_test.go
```diff
@@ -3870,3 +3870,177 @@ func TestInterpretDynamicCastingReferenceCasting(t *testing.T) {
...
+			t,
+			inter,
+			interpreter.NewUnmeteredSomeValueNonCopying(
+				interpreter.NewUnmeteredSomeValueNonCopying(
+					interpreter.NewUnmeteredIntValueFromInt64(42),
+				),
+			),
+			inter.Globals.Get("y").GetValue(inter),
+		)
+	})
```
SupunS, line 4045: Tests are great! Thank you for covering all these different cases.
Can you please also add a test for resources, and `AnyResource` as well?

SupunS (commented):
Looks good! Nice to see the fix is kept simple 👌

runtime/interpreter/interpreter_expression.go
RZhang05: Good point, added this unwrapping.

runtime/tests/interpreter/dynamic_casting_test.go
RZhang05: Added some tests for resources covering the expected use cases.

SupunS (approved):
LGTM!

runtime/tests/interpreter/dynamic_casting_test.go
```diff
@@ -3870,3 +3870,365 @@ func TestInterpretDynamicCastingReferenceCasting(t *testing.T) {
...
+		require.NoError(t, err)
+		require.Equal(t, interpreter.NewUnmeteredStringValue("hello"), result)
+	})
+
+	t.Run("return nested optional as?", func(t *testing.T) {
+		t.Parallel()
+
+		code := `
+			let x: Int??? = 42
+			let y: Int?? = x as? Int?
```
turbolent, line 4008: Can you please also add a test case for this scenario (`Int???` casted to `Int?`) with the `as!` operator?

runtime/tests/interpreter/dynamic_casting_test.go
```diff
@@ -3870,3 +3870,365 @@ func TestInterpretDynamicCastingReferenceCasting(t *testing.T) {
...
+			inter.Globals.Get("y").GetValue(inter),
+		)
+	})
+
+	t.Run("return nested optional as!", func(t *testing.T) {
+		t.Parallel()
+
+		code := `
+			let x: Int??? = 42
+			let y: Int?? = x as! Int??
```
turbolent, line 4030: Can you please also add a test case for this scenario (`Int???` casted to `Int??`) with the `as?` operator?

turbolent (approved):
Great work!

+1 on what @SupunS said, it's great that the implementation of this improvement is so simple.

Initially I was surprised this worked, I had assumed we would need to unbox only to the requested "level" of nesting. However, after looking at the rest of the implementation, this works because we box up the value again to the requested level again below (`value = interpreter.ConvertAndBox(locationRange, value, valueSemaType, expectedType`) 👍

turbolent:
Related: Swift's dynamic casting is actually way more complicated then I thought https://github.com/swiftlang/swift/blob/main/docs/DynamicCasting.md 😅

runtime/tests/interpreter/dynamic_casting_test.go
RZhang05: Added.

runtime/tests/interpreter/dynamic_casting_test.go
RZhang05: Added.

turbolent (approved):
Great work!

PR #3582, opened 2024-09-12
Always report errors against the new program during contract update validation

### LINKED ISSUE #3574: Improve contract update error printing positions
### Current Behavior

During the contract update validation, errors can be reported for the **old** program in two occasions:
- When parsing the old program
- When validating the new program against the old one

The first case (parser errors) now uses the correct (old) code for pretty-printing after the fix: https://github.com/onflow/cadence/pull/3554.

However, it is still possible to report errors using the old positions during validation (2nd case above). For e.g:
- Getting root declaration reports the error using the old declaration's position
- Validating the type-removal pragma, also uses old code's position info for reporting errors

Given the update validator doesn't fail on the first error, having errors with position info from both old and new codes is problematic, because the pretty-printer can only use one of the two versions of code (either the new or the old code) at a given time, and will point to incorrect positions during pretty-printing the errors.

### Expected Behavior

Error reporting probably need to use new code's position info, always.

### Steps To Reproduce

N/A

### Environment

```markdown
- Cadence version: master
- Network:
```

SupunS:
Closes #3574

## Description

Having a mix of errors from both old and new programs could result in incorrect positions being reported.
So always report errors only against new program's positions. 

______

### DIFF
--- a/runtime/contract_update_validation_test.go
+++ b/runtime/contract_update_validation_test.go
@@ -3666,3 +3666,101 @@ func TestTypeRemovalPragmaUpdates(t *testing.T) {
 		},
 	)
 }
+
+func TestRuntimeContractUpdateErrorsInOldProgram(t *testing.T) {
+
+	t.Parallel()
+
+	testWithValidatorsAndTypeRemovalEnabled(t,
+		"invalid #removedType pragma in old code",
+		func(t *testing.T, config Config) {
+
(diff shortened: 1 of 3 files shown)


turbolent (approved):
Nice! Thanks for fixing this. Great to see we can reuse `OldProgramError` here

Issue #3579, opened 2024-09-11
Add string formatting function 


bluesign:
### Issue to be solved

context: https://discord.com/channels/613813861610684416/1162086721471647874/1283181721562517607

It would be nice to have some basic string formatting.

SupunS:
Based on the discord discussion issue can be have two parts:
- [ ] Introduce a `Stringer` interface.
  - Make all string-convertible types are subtypes of this newly added interface
  - See the existing `HashableStruct` type for inspiration.
- [ ] Support string templates, where formatting can be inlined in the string-literal itself.
  - See [Kotlin's sprint templates](https://kotlinlang.org/docs/strings.html#string-templates) and [Swift's string interpolation](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/stringsandcharacters/#String-Interpolation) for some ideas.

bluesign:
> If the Stringer implementation is native to the language, then they could be safely used in this example

@jribbink I agree user defined Stringer is dangerous (I raised some concerns on this before), but also a quick note, in general for your use case, native Stringer can be also dangerous. 

In general you need 3 properties; deterministic, uniqueness and immutable ( more like crypto hashing ) I think output of ‘log’ lacks some.

SupunS:
Thanks @jribbink and @bluesign for the feedback! Given the concerns, we opened a FLIP for the `Stringer` interface specifically: https://github.com/onflow/flips/pull/294 to discuss these concerns further and find a solution that works. We can also discuss in one of the upcoming working groups.

PR #3585, opened 2024-09-19
Limited string templates: $identifier

RZhang05:
Working towards #3579

## Description

This PR introduces a simple string template in the form of the character `$` followed by an identifier (`"$foo"`), with plans to support the syntax of `$` followed by an expression in curly braces later.

______

### DIFF
--- a/runtime/ast/expression.go
+++ b/runtime/ast/expression.go
@@ -220,6 +220,63 @@ func (*StringExpression) precedence() precedence {
 	return precedenceLiteral
 }
 
+// StringTemplateExpression
+
+type StringTemplateExpression struct {
+	Values      []string
+	Expressions []Expression
+	Range
+}
+
+var _ Expression = &StringTemplateExpression{}
+
+func NewStringTemplateExpression(gauge common.MemoryGauge, values []string, exprs []Expression, exprRange Range) *StringTemplateExpression {
+	common.UseMemory(gauge, common.StringExpressionMemoryUsage)
+	return &StringTemplateExpression{
+		Values:      values,
+		Expressions: exprs,
+		Range:       exprRange,
+	}
+}
+
+var _ Element = &StringExpression{}
+var _ Expression = &StringExpression{}
+
+func (*StringTemplateExpression) ElementType() ElementType {
+	return ElementTypeStringExpression
+}
+
+func (*StringTemplateExpression) isExpression() {}
+
+func (*StringTemplateExpression) isIfStatementTest() {}
+
+func (*StringTemplateExpression) Walk(_ func(Element)) {
+	// NO-OP
+}
+
+func (e *StringTemplateExpression) String() string {
+	return Prettier(e)
+}
+
+func (e *StringTemplateExpression) Doc() prettier.Doc {
+	return prettier.Text(QuoteString("String template"))
+}
+
+func (e *StringTemplateExpression) MarshalJSON() ([]byte, error) {
+	type Alias StringTemplateExpression
+	return json.Marshal(&struct {
+		*Alias
+		Type string
+	}{
+		Type:  "StringTemplateExpression",
+		Alias: (*Alias)(e),
+	})
+}
+
+func (*StringTemplateExpression) precedence() precedence {
+	return precedenceLiteral
+}
+
 // IntegerExpression
 
 type IntegerExpression struct {
--- a/runtime/ast/expression.go
+++ b/runtime/ast/expression.go
@@ -243,7 +243,7 @@ var _ Element = &StringExpression{}
 var _ Expression = &StringExpression{}
 
 func (*StringTemplateExpression) ElementType() ElementType {
-	return ElementTypeStringExpression
+	return ElementTypeStringTemplateExpression
 }
 
 func (*StringTemplateExpression) isExpression() {}
--- a/runtime/ast/expression.go
+++ b/runtime/ast/expression.go
@@ -231,6 +231,7 @@ type StringTemplateExpression struct {
 var _ Expression = &StringTemplateExpression{}
 
 func NewStringTemplateExpression(gauge common.MemoryGauge, values []string, exprs []Expression, exprRange Range) *StringTemplateExpression {
+	// STRINGTODO: change to be similar to array memory usage?
 	common.UseMemory(gauge, common.StringExpressionMemoryUsage)
 	return &StringTemplateExpression{
 		Values:      values,
--- a/runtime/interpreter/interpreter_expression.go
+++ b/runtime/interpreter/interpreter_expression.go
@@ -957,6 +957,34 @@ func (interpreter *Interpreter) VisitStringExpression(expression *ast.StringExpr
 	return NewUnmeteredStringValue(expression.Value)
 }
 
+func (interpreter *Interpreter) VisitStringTemplateExpression(expression *ast.StringTemplateExpression) Value {
+	values := interpreter.visitExpressionsNonCopying(expression.Expressions)
+
+	templateExpressionTypes := interpreter.Program.Elaboration.StringTemplateExpressionTypes(expression)
+	argumentTypes := templateExpressionTypes.ArgumentTypes
+
+	var copies []Value
+
+	count := len(values)
+	if count > 0 {
+		copies = make([]Value, count)
+		for i, argument := range values {
+			argumentType := argumentTypes[i]
+			argumentExpression := expression.Expressions[i]
+			locationRange := LocationRange{
+				Location:    interpreter.Location,
+				HasPosition: argumentExpression,
+			}
+			copies[i] = interpreter.transferAndConvert(argument, argumentType, sema.StringType, locationRange)
+		}
+	}
+
+	result := ""
+
+	// NOTE: already metered in lexer/parser
+	return NewUnmeteredStringValue(result)
+}
+
 func (interpreter *Interpreter) VisitArrayExpression(expression *ast.ArrayExpression) Value {
 	values := interpreter.visitExpressionsNonCopying(expression.Values)
 
--- a/runtime/interpreter/interpreter_expression.go
+++ b/runtime/interpreter/interpreter_expression.go
@@ -20,6 +20,7 @@ package interpreter
 
 import (
 	"math/big"
+	"strings"
 	"time"
 
 	"github.com/onflow/atree"
@@ -960,29 +961,28 @@ func (interpreter *Interpreter) VisitStringExpression(expression *ast.StringExpr
 func (interpreter *Interpreter) VisitStringTemplateExpression(expression *ast.StringTemplateExpression) Value {
 	values := interpreter.visitExpressionsNonCopying(expression.Expressions)
 
-	templateExpressionTypes := interpreter.Program.Elaboration.StringTemplateExpressionTypes(expression)
-	argumentTypes := templateExpressionTypes.ArgumentTypes
+	templatesType := interpreter.Program.Elaboration.StringTemplateExpressionTypes(expression)
+	argumentTypes := templatesType.ArgumentTypes
 
-	var copies []Value
-
-	count := len(values)
-	if count > 0 {
-		copies = make([]Value, count)
-		for i, argument := range values {
-			argumentType := argumentTypes[i]
-			argumentExpression := expression.Expressions[i]
-			locationRange := LocationRange{
-				Location:    interpreter.Location,
-				HasPosition: argumentExpression,
+	var builder strings.Builder
+	for i, str := range expression.Values {
+		builder.WriteString(str)
+		if i < len(values) {
+			// STRINGTODO: is this how the conversion should happen?
+			s := values[i].String()
+			switch argumentTypes[i] {
+			case sema.StringType:
+				// remove quotations
+				s = s[1 : len(s)-1]
+				builder.WriteString(s)
+			default:
+				builder.WriteString(s)
 			}
-			copies[i] = interpreter.transferAndConvert(argument, argumentType, sema.StringType, locationRange)
 		}
 	}
 
-	result := ""
-
-	// NOTE: already metered in lexer/parser
-	return NewUnmeteredStringValue(result)
+	// STRINGTODO: already metered as a string constant in parser?
+	return NewUnmeteredStringValue(builder.String())
 }
 
 func (interpreter *Interpreter) VisitArrayExpression(expression *ast.ArrayExpression) Value {
--- a/runtime/parser/expression.go
+++ b/runtime/parser/expression.go
@@ -1167,14 +1167,15 @@ func defineStringExpression() {
 				literal = p.tokenSource(curToken)
 				length = len(literal)
 
-				if length >= 1 && literal[0] == '"' {
-					literal = literal[1:]
-					length = len(literal)
-				}
-
+				// this order of removal matters in case the token is the single quotation "
 				if length >= 1 && literal[length-1] == '"' {
 					literal = literal[:length-1]
+					length = len(literal)
 					missingEnd = false
+				}
+
+				if length >= 1 && literal[0] == '"' {
+					literal = literal[1:]
 				}
 
 				parsedString := parseStringLiteralContent(p, literal)
--- a/runtime/parser/expression.go
+++ b/runtime/parser/expression.go
@@ -433,19 +433,6 @@ func init() {
 		},
 	})
 
-	defineExpr(literalExpr{
-		tokenType: lexer.TokenString,
-		nullDenotation: func(p *parser, token lexer.Token) (ast.Expression, error) {
-			literal := p.tokenSource(token)
-			parsedString := parseStringLiteral(p, literal)
-			return ast.NewStringExpression(
-				p.memoryGauge,
-				parsedString,
-				token.Range,
-			), nil
-		},
-	})
-
 	defineExpr(prefixExpr{
 		tokenType:    lexer.TokenMinus,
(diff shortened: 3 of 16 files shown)


RZhang05:
Opened for early feedback.

- Need to add more tests across all stages of the feature
- Implementation questions marked with `STRINGTODO` in the code
- Further consider handling (or lack of handling) of error cases in the lexer and parser

runtime/parser/lexer/lexer.go
```diff
@@ -49,6 +49,14 @@ type position struct {
 	column int
 }
 
+type LexerMode int
+
+const (
+	NORMAL = iota
+	STR_IDENTIFIER
+	STR_EXPRESSION
```
SupunS, line 57: Looks like `STR_EXPRESSION` is not set anywhere. So it actually on requires two modes? Maybe just use a boolean flag in that case. e.g: `inStringTemplate` or something similar.

runtime/parser/lexer/lexer.go
```diff
@@ -414,6 +424,11 @@ func (l *lexer) scanString(quote rune) {
 				l.backupOne()
 				return
 			}
+		case '$':
+			// string template, stop and set mode
+			l.backupOne()
+			l.mode = STR_IDENTIFIER
+			return
```
SupunS, line 431: We might also need a way to escape  the`$` sign when needed. e.g: not all `$` signs in a string are interpolations.
e.g: I may want to keep the $-sign as-is in a string literal such as: `"withdrawn amount: $5.00"`

We could use `\` for escaping, like for other cases. Then I would have to re-write the literal as `"withdrawn amount: \$5.00"`

RZhang05: This should already be supported in the lexer and requires a small change in the parser. Added.

runtime/parser/lexer/lexer.go
RZhang05: For this PR, `STR_EXPRESSION` could be removed and switched to a boolean flag instead, future extension for `${expression}` would require this to be added back potentially. With some limitations on syntax the changes for `${expression}` should not be too significant as in #3587.

runtime/sema/check_string_template_expression.go
```diff
@@ -0,0 +1,64 @@
...
+		argumentTypes = make([]Type, elementCount)
+
+		for i, element := range stringTemplateExpression.Expressions {
+			valueType := checker.VisitExpression(element, stringTemplateExpression, elementType)
+
+			argumentTypes[i] = valueType
+
+			// All number types, addresses, path types, bool and strings are supported in string template
+			if IsSubType(valueType, NumberType) || IsSubType(valueType, TheAddressType) ||
+				IsSubType(valueType, PathType) || IsSubType(valueType, StringType) || IsSubType(valueType, BoolType) {
```
SupunS, line 42: Maybe extract this logic to a separate method for better readability. For things like `TheAddressType`, `StringType`, and `BoolType`, we can directly switch on the type (doesn't need to check sub-typing).
Also need to handle `Character` type

runtime/sema/check_string_template_expression.go
```diff
@@ -0,0 +1,64 @@
...
+				checker.checkResourceMoveOperation(element, valueType)
+			} else {
+				checker.report(
+					&TypeMismatchWithDescriptionError{
+						ActualType:              valueType,
+						ExpectedTypeDescription: "a type with built-in toString() or bool",
```
SupunS, line 48: Why `bool` needs to be special cased? Can't we treat it also as a type with `toString()` method?
Maybe a more generic description like "a string convertible type" might fit all?

runtime/sema/check_string_template_expression.go
```diff
@@ -0,0 +1,64 @@
...
+
+		for i, element := range stringTemplateExpression.Expressions {
+			valueType := checker.VisitExpression(element, stringTemplateExpression, elementType)
+
+			argumentTypes[i] = valueType
+
+			// All number types, addresses, path types, bool and strings are supported in string template
+			if IsSubType(valueType, NumberType) || IsSubType(valueType, TheAddressType) ||
+				IsSubType(valueType, PathType) || IsSubType(valueType, StringType) || IsSubType(valueType, BoolType) {
+				checker.checkResourceMoveOperation(element, valueType)
```
SupunS, line 43: We might not need to check for `checkResourceMoveOperation`, given the resource types are not allowed in the template expressions (already filtered above)

runtime/parser/lexer/lexer.go
```diff
@@ -49,6 +49,13 @@ type position struct {
 	column int
 }
 
+type LexerMode int
+
+const (
+	NORMAL = iota
+	STR_IDENTIFIER
```
SupunS, line 56: Suggestion: Maybe rename this to a more generic one like `stringInterpolation`, since more expressions will be supported in future.

btw, Go's constants also follow a Camel-case naming style (because starting with a upper-case letter means it becomes a global constant)

runtime/parser/lexer/lexer.go
```diff
@@ -49,6 +49,13 @@ type position struct {
 	column int
 }
 
+type LexerMode int
+
+const (
+	NORMAL = iota
```
SupunS, line 55:
```suggestion
type LexerMode uint8

const (
	NORMAL LexerMode = iota
```

runtime/parser/expression_test.go
```diff
@@ -6055,6 +6055,270 @@ func TestParseStringWithUnicode(t *testing.T) {
...
+			[]error{
+				&SyntaxError{
+					Message: "expected identifier got: add()",
+					Pos:     ast.Position{Offset: 12, Line: 2, Column: 11},
+				},
+			},
+			errs,
+		)
+	})
+}
```
SupunS, line 6320: Can you please also add some tests for templates with unbalanced parenthesis?

runtime/parser/expression.go
```diff
@@ -1144,6 +1132,108 @@ func defineNestedExpression() {
 	)
 }
 
+func defineStringExpression() {
```
SupunS, line 1135: Would it help to emit a different token as the starting token of the string template, other than a `lexer.TokenString`.
I was thinking, in a template string like `"hello \(name), welcome!"`, if we can emit a new token, say `lexer.TokenStringTemplatePart` for the first `"hello` part (and all subsequent string parts), then we can maybe keep the parsing logic of the strings and string templates separate?

runtime/parser/lexer/state.go
```diff
@@ -118,6 +129,17 @@ func rootState(l *lexer) stateFn {
 			return numberState
 		case '"':
 			return stringState
+		case '\\':
+			if l.mode == STR_IDENTIFIER {
+				r = l.next()
+				switch r {
+				case '(':
+					l.emitType(TokenStringTemplate)
+					l.openBrackets++
```
SupunS, line 138: So after reaching `\(`, it continues to lex in the root-state. So AFAIU, lexer doesn't limit what can be put inside the interpolation (only the parser does, correct?). Curious how does the lexer handles nested string templates? Something like `"outer template \( "inner template with \(foo)") end"`. Could you maybe add some tests?

runtime/interpreter/interpreter_expression.go
```diff
@@ -957,6 +958,32 @@ func (interpreter *Interpreter) VisitStringExpression(expression *ast.StringExpr
...
+	templatesType := interpreter.Program.Elaboration.StringTemplateExpressionTypes(expression)
+	argumentTypes := templatesType.ArgumentTypes
+
+	var builder strings.Builder
+	for i, str := range expression.Values {
+		builder.WriteString(str)
+		if i < len(values) {
+			// this is equivalent to toString() for supported types
+			s := values[i].String()
+			switch argumentTypes[i] {
```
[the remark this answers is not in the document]
RZhang05: This is a good point, it should not switch on type.

runtime/sema/check_string_template_expression.go
RZhang05: Currently, bool does not support a `toString()` method so the error would be a little inaccurate. As well I think it is unclear what a "string convertible type" refers to if a more generic description was chosen.

runtime/parser/expression.go
RZhang05: We could do this and keep the logic separate. However, I think handling the parsing of both is quite natural and does not affect the complexity of string templates.

runtime/parser/lexer/state.go
RZhang05: Tests added in [bda6b28](https://github.com/onflow/cadence/pull/3585/commits/bda6b28fb29cff50e635a3e9d94b926290001b33). For `"outer template \( "inner template with \(foo)") end"`, the parser will identify `"outer template \( "` as a string which is invalid and will error with `missing ')'`.

runtime/sema/check_string_template_expression.go
```diff
@@ -0,0 +1,66 @@
...
+			}
+		}
+	}
+
+	checker.Elaboration.SetStringTemplateExpressionTypes(
+		stringTemplateExpression,
+		StringTemplateExpressionTypes{
+			ArgumentTypes: argumentTypes,
+		},
+	)
```
SupunS, line 63: `StringTemplateExpressionTypes` are no longer read/used, so no need to actually set these.

SupunS (commented):
Looks good! Just some minor cleanups

runtime/parser/expression.go
```diff
@@ -1144,6 +1132,108 @@ func defineNestedExpression() {
...
+
+			for curToken.Is(lexer.TokenString) {
+				literal = p.tokenSource(curToken)
+				length = len(literal)
+
+				// remove quotation marks if they exist
+				if curToken == startToken {
+					literal = literal[1:]
+					length = len(literal)
+				}
```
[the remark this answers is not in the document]
RZhang05: Removed the first `length` assignment instead since its unused.

runtime/sema/elaboration.go
```diff
@@ -480,6 +485,21 @@ func (e *Elaboration) SetStringExpressionType(expression *ast.StringExpression,
...
+	return e.stringTemplateExpressionTypes[expression]
+}
+
+func (e *Elaboration) SetStringTemplateExpressionTypes(expression *ast.StringTemplateExpression, types StringTemplateExpressionTypes) {
+	if e.stringTemplateExpressionTypes == nil {
+		e.stringTemplateExpressionTypes = map[*ast.StringTemplateExpression]StringTemplateExpressionTypes{}
+	}
+	e.stringTemplateExpressionTypes[expression] = types
+}
+
```
SupunS, line 502: Unused?

Can remove `type StringTemplateExpressionTypes struct` altogether

SupunS (approved):
Great work! 🎉

runtime/ast/expression.go
```diff
@@ -220,6 +220,68 @@ func (*StringExpression) precedence() precedence {
 	return precedenceLiteral
 }
 
+// StringTemplateExpression
+
+type StringTemplateExpression struct {
+	Values      []string
+	Expressions []Expression
```
turbolent, line 227: How are string templates represented by these fields? Is the assumption that they are interleaved, i.e., `V[0] + E[0] + ... + E[N-1] + V[N]`? So for example, 
- would `"a\(b)c"` be `Values: []string{"a", "c"}, Expressions: []Expression{b}`, and
- would `"\(a)b\(c)"` be `Values: []string{"", "b", ""}, Expressions: []Expression{a, c}`? 

Could you please add some more tests for these and similar cases?

runtime/ast/expression.go
```diff
@@ -220,6 +220,68 @@ func (*StringExpression) precedence() precedence {
...
+func (*StringTemplateExpression) Walk(_ func(Element)) {
+	// NO-OP
+}
+
+func (e *StringTemplateExpression) String() string {
+	return Prettier(e)
+}
+
+func (e *StringTemplateExpression) Doc() prettier.Doc {
+	return prettier.Text(QuoteString("String template"))
```
turbolent, line 267: This needs to reproduce the expression as a literal. If this is not supported yet as part of this PR, maybe just leave a `// TODO` and `panic("not implemented")`

runtime/parser/lexer/lexer.go
```diff
@@ -49,6 +49,13 @@ type position struct {
 	column int
 }
 
+type lexerMode uint8
+
+const (
+	normal lexerMode = iota
+	stringInterpolation
```
turbolent, line 56: Prefix with the type name, i.e.
```suggestion
	lexerModeNormal lexerMode = iota
	lexerModeStringInterpolation
```

runtime/sema/check_string_template_expression.go
```diff
@@ -0,0 +1,67 @@
...
+
+	// visit all elements
+
+	var elementType Type
+
+	elementCount := len(stringTemplateExpression.Expressions)
+
+	var argumentTypes []Type
+	if elementCount > 0 {
+		argumentTypes = make([]Type, elementCount)
```
turbolent, line 47: The slice allocated here does not seem to be used, it can probably be removed

turbolent (commented):
Great work!

turbolent:
Can you please update the title to reflect the current proposal (e.g. description)?

runtime/ast/expression.go
RZhang05: > Is the assumption that they are interleaved, i.e., `V[0] + E[0] + ... + E[N-1] + V[N]`?

Yes they are assumed to be interleaved and this is enforced by the parser. 

> Could you please add some more tests for these and similar cases?

Added the above tests and one more.

runtime/ast/expression.go
RZhang05: Left it as a `//TODO` with `panic` for this PR.

runtime/ast/expression.go
```diff
@@ -220,6 +220,73 @@ func (*StringExpression) precedence() precedence {
...
+func (*StringTemplateExpression) ElementType() ElementType {
+	return ElementTypeStringTemplateExpression
+}
+
+func (*StringTemplateExpression) isExpression() {}
+
+func (*StringTemplateExpression) isIfStatementTest() {}
+
+func (*StringTemplateExpression) Walk(_ func(Element)) {
+	// NO-OP
```
turbolent, line 259: This should walk over the child expressions (`e.Expressions`). See how this is implemented for e.g. array expressions (`ArrayExpression.Walk`), with the `walkExpressions` helper function

runtime/ast/expression.go
turbolent: Makes sense! Can you please leave an explanation in the code, maybe also with these examples?

It might also be good to assert this structure in the constructor (`NewStringTemplateExpression`), to prevent illegal state

runtime/ast/expression.go
```diff
@@ -220,6 +220,73 @@ func (*StringExpression) precedence() precedence {
...
+}
+
+func (e *StringTemplateExpression) String() string {
+	return Prettier(e)
+}
+
+func (e *StringTemplateExpression) Doc() prettier.Doc {
+	if len(e.Expressions) == 0 {
+		return prettier.Text(e.Values[0])
```
turbolent, line 268: The result of the pretty printing function must produce source code, so the string component here must be quoted, so e.g. the input `"hello"` get's parsed into `hello`, and then should be reproduced here as `"hello"` again.

You can see how this is done for normal string literals in `StringExpression.Doc`.

runtime/tests/checker/string_test.go
```diff
@@ -697,3 +697,94 @@ func TestCheckStringCount(t *testing.T) {
...
+		errs := RequireCheckerErrors(t, err, 1)
+
+		assert.IsType(t, &sema.TypeMismatchWithDescriptionError{}, errs[0])
+	})
+
+	t.Run("invalid, array", func(t *testing.T) {
+		t.Parallel()
+
+		_, err := ParseAndCheck(t, `
+			let x :[AnyStruct] = ["tmp", 1]
```
turbolent, line 749:
```suggestion
			let x: [AnyStruct] = ["tmp", 1]
```

runtime/sema/check_string_template_expression.go
```diff
@@ -0,0 +1,66 @@
...
+func (checker *Checker) VisitStringTemplateExpression(stringTemplateExpression *ast.StringTemplateExpression) Type {
+
+	// visit all elements
+
+	var elementType Type
+
+	elementCount := len(stringTemplateExpression.Expressions)
+
+	if elementCount > 0 {
+		argumentTypes := make([]Type, elementCount)
```
turbolent, line 46: This slice is populated, but never used. From what I can see, it should be possible to just remove it:
```suggestion
```

runtime/sema/check_string_template_expression.go
```diff
@@ -0,0 +1,66 @@
...
+
+		for i, element := range stringTemplateExpression.Expressions {
+			valueType := checker.VisitExpression(element, stringTemplateExpression, elementType)
+
+			argumentTypes[i] = valueType
```
turbolent, line 51: See above: The slice is populated here, but does not seem to be used/read anywhere.

```suggestion
```

turbolent (commented):
Great progress! Only a couple minor issues left. Once resolved, this should be ready to get merged 👏

turbolent (approved):
Great work! 👏

runtime/parser/expression.go
SupunS: @RZhang05 can you maybe create an issue for this as a future improvement? Then we can resolve this and merge the PR.

PR #3589, opened 2024-09-25
Improve `result` variable

### LINKED ISSUE #774: BUG: Using `result` as a variable name causes a checker error
### Problem

The checker doesn't allow using the name `result` as a variable name, 

### Steps to Reproduce

Use this script:
```
pub fun main(): Int {
  let result = 1

  return result
}
```
See the error: 
```
error: cannot redeclare result: `result` is already declared
```

### Acceptance Criteria

I'm not sure if this is intentional or not. it might have something to do with the result of the script being checked in a post condition or something. Even if it is intentional, the error message should probably be improved to make it more clear

turbolent:
Closes #774

## Description

- Only declare the `result` variable if post conditions are declared
- Provide a better, dedicated error, with notes

Screenshot 2024-09-25 at 2 08 44 PM

Best viewed without whitespace changes

______

### DIFF
--- a/runtime/runtime_test.go
+++ b/runtime/runtime_test.go
@@ -11480,3 +11480,30 @@ func TestRuntimeStorageEnumAsDictionaryKey(t *testing.T) {
 		loggedMessages,
 	)
 }
+
+func TestResultRedeclared(t *testing.T) {
+
+	t.Parallel()
+
+	runtime := NewTestInterpreterRuntime()
+
+	script := []byte(`
+      access(all) fun main(): Int { let result = 1; return result }
+    `)
+
+	runtimeInterface := &TestRuntimeInterface{}
+
+	nextScriptLocation := NewScriptLocationGenerator()
+
+	_, err := runtime.ExecuteScript(
+		Script{
+			Source: script,
+		},
+		Context{
+			Interface: runtimeInterface,
+			Location:  nextScriptLocation(),
+		},
+	)
+	require.NoError(t, err)
+
+}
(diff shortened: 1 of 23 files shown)


bluesign:
Never thought before but is there any value of post condition accessing local scope of the function ?  I think I haven't seen any usage like that, also feels a bit strange when I think about it. 

Global scope ( and self ), input and output feels should be enough, if that is the case, maybe pushing function body when level down can be some solution.

SupunS (commented):
> Only declare the result variable if post conditions are declared

How does this work with inherited post conditions? Can you maybe add some tests?

turbolent:
@SupunS Good idea! Added a test case in fea8a9a739b31d55761018a0306b1df29d89d500. Was that what you had in mind? Anything else you think that should be tested?

runtime/runtime_test.go
```diff
@@ -11480,3 +11480,86 @@ func TestRuntimeStorageEnumAsDictionaryKey(t *testing.T) {
...
+
+          access(all)
+          struct S: SI {
+
+              access(all)
+              fun test(): Int {
+                  let result = 1
+                  return result
+              }
+          }
```
SupunS, line 11542: Thank you for adding the test! 

Maybe return a different value than the local variable `result`, to make sure the post condition picks-up the returned value, not the local variable.

```suggestion
          struct interface SI {
              access(all)
              fun test(): Int {
                  post {
                      result == 2
                  }
              }
          }

          access(all)
          struct S: SI {

              access(all)
              fun test(): Int {
                  let result = 1
                  return 2  // return a different value than the local variable
              }
          }
```

SupunS (approved):

runtime/runtime_test.go
turbolent: Great idea!

PR #3594, opened 2024-09-26
Introduce StructStringer

RZhang05:
Working towards #3579

## Description

Implement https://github.com/onflow/flips/pull/294

Add a simple `StructStringer` struct interface. It is intended in the future that all types which implement `StructStringer` can be used in string templates.

______

- [x] Targeted PR against `feature/string-templates` branch


SupunS:
@RZhang05 is this ready to be reviewed? or are there any pending changes?

RZhang05:
> @RZhang05 is this ready to be reviewed? or are there any pending changes?

It is ready, note that there are some questions left in the code marked with TODO however.

runtime/sema/type.go
```diff
@@ -7825,6 +7837,12 @@ func checkSubTypeWithoutEquality(subType Type, superType Type) bool {
 					IsSubsetOf(typedSubType.EffectiveInterfaceConformanceSet())
 			}
 
+			// STRINGERTODO: other options? how to make existing simple types
+			// conform to an intersection type?
```
turbolent, line 7841: It isn't so much about conforming these types to an intersection type, but an interface, `Stringer`.

At the moment we don't have a means to specify a subtype relationship for simple types, so we might want to add that, i.e. generalize this solution a bit.

Note that the intersection type might contain multiple interfaces, e.g. might not just be `{Stringer}`, so checking the first element is not sufficient.

There's also no need, and it is not sufficient, to compare the names only, it is better to check for pointer equality here.

runtime/tests/interpreter/stringer_test.go
```diff
@@ -0,0 +1,78 @@
...
+	result, err := inter.Invoke("test")
+	require.NoError(t, err)
+
+	RequireValuesEqual(
+		t,
+		inter,
+		interpreter.NewUnmeteredStringValue("1"),
+		result,
+	)
+}
```
SupunS, line 78: Can you also add some tests for
- Using `Stringer` type as a 'value'. e.g: `var v = Type<Stringer>()`
- Casting to `Stringer` type. e.g: `somevalue as {Stringer}`

runtime/sema/type.go
```diff
@@ -9534,3 +9552,34 @@ func init() {
...
+
+		nestedTypes.Foreach(func(_ string, nestedType Type) {
+			nestedInterfaceType, ok := nestedType.(*InterfaceType)
+			if !ok {
+				return
+			}
+
+			interfaceTypes = append(interfaceTypes, nestedInterfaceType)
+		})
+	}
```
turbolent, line 9584: It might be good to refactor this into a function, and also use it in the `init` block initializing `NativeCompositeTypes`

runtime/sema/stringer.go
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+package sema
+
+import (
+	"github.com/onflow/cadence/runtime/ast"
+	"github.com/onflow/cadence/runtime/common"
+)
+
+const StringerTypeName = "Stringer"
```
turbolent, line 26: Unfortunately we cannot define a single interface that users could implement for both structs and resources. To begin with we're only adding a struct interface, but we might also want to provide another resource interface in the future. We might want to consider naming this so it indicates that it is only for structs, e.g. `StructStringer`

turbolent (commented):
Nice work!

runtime/sema/stringer.go
```diff
@@ -0,0 +1,59 @@
...
+
+var StringerType = func() *InterfaceType {
```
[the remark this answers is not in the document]
RZhang05: Done in [5f16333](https://github.com/onflow/cadence/pull/3594/commits/5f163333eb1290df11bd7e1443b08ea03eb8a3c0), added support for simple interfaces.

runtime/sema/type.go
```diff
@@ -7825,6 +7837,12 @@ func checkSubTypeWithoutEquality(subType Type, superType Type) bool {
 					IsSubsetOf(typedSubType.EffectiveInterfaceConformanceSet())
 			}
 
+			// STRINGERTODO: other options? how to make existing simple types
+			// conform to an intersection type?
+			if typedSuperType.Types[0].Identifier == StringerTypeName {
+				return IsStringerType(subType)
+			}
+
```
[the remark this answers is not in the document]
RZhang05: Not sure if this applies since `StringerType` is an interface and from my understanding it only makes sense to check an interface conformance as part of an intersection type, not directly as a subtype (e.g. `var x: Stringer = 1` is invalid).

runtime/tests/interpreter/stringer_test.go
RZhang05: Add casting test in [ab157ad](https://github.com/onflow/cadence/pull/3594/commits/ab157ad8e4b0fb15147493aca0beab5d08be0203). Using `Type<Stringer>` as a value is invalid since its an interface, is the test meant to check for that error?

runtime/sema/type.go
```diff
@@ -4899,6 +4900,17 @@ func IsHashableStructType(t Type) bool {
 	}
 }
 
+// which simple types conform to stringer interface (except Bool?)
```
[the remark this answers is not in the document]
RZhang05: I believe `Bool` does not have a `toString` method which may be confusing.

runtime/sema/type.go
RZhang05: > At the moment we don't have a means to specify a subtype relationship for simple types, so we might want to add that, i.e. generalize this solution a bit.

In [5f16333](https://github.com/onflow/cadence/pull/3594/commits/5f163333eb1290df11bd7e1443b08ea03eb8a3c0), I added the functionality for simple types to conform to an interface.

Not sure the best approach in general for all string convertible types to conform to `Stringer` since there are types such as `TheAddressType` and `NumberType` which are not `SimpleType`s so I've left the edge case check in for those.

runtime/sema/type.go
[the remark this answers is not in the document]
SupunS: The `checkSubTypeWithoutEquality` method is a recursive function, so for each type in an interface set `{I1, I2, Stringer}` (or a simple case where `{Stringer}`), it will call the same method recursively, by each interface-type as the super-type. So it makes sense to have `Stringer` type at the top level of the switch-case. Any regular interface types are also at the top level of the switch case, see [sema/type.go#L7927](https://github.com/onflow/cadence/blob/aa514550ad5eca99db4f43bd99f643f4ef79bdad/runtime/sema/type.go#L7927).

> (e.g. var x: Stringer = 1 is invalid).

This is a separate check done at [a different place](https://github.com/onflow/cadence/blob/aa514550ad5eca99db4f43bd99f643f4ef79bdad/runtime/sema/checker.go#L2340). It prevents someone from using an interface type outside of a interface-set/intersection type.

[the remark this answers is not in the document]
RZhang05: Currently, moving the check to the top level of the switch-case does not work, perhaps I'm misunderstanding but it does not seem to call itself recursively once it enters [sema/type.go#L7724](https://github.com/onflow/cadence/blob/aa514550ad5eca99db4f43bd99f643f4ef79bdad/runtime/sema/type.go#L7724).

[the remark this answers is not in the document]
SupunS: Oh I see the problem.
I think then we can extend the changes you already added/mentioned in https://github.com/onflow/cadence/pull/3594#discussion_r1785154740, to also other non-simple types like `AddressType` and `NumberType`.

i.e: introduce a new Go interface, something like

```golang
type ConformingType interface {
	Type
	EffectiveInterfaceConformanceSet() *InterfaceSet
}
```
And make sure `SimpleType`, `AddressType` and `NumberType` all implement the new interface `ConformingType`.

Then in then [sema/type.go#L7776](https://github.com/onflow/cadence/blob/aa514550ad5eca99db4f43bd99f643f4ef79bdad/runtime/sema/type.go#L7776), add a new `case ConformingType:`, and there, perform the subset check, similar to the composite type (i.e: `typedSuperType.EffectiveIntersectionSet().IsSubsetOf(typedSubType.EffectiveInterfaceConformanceSet())`). Might be able to even combine the two, by making `CompositeType` also implement `ConformingType` at Go level.

runtime/tests/interpreter/stringer_test.go
SupunS: oh my bad, meant to say `var v = Type<{Stringer}>()`

runtime/sema/type.go
RZhang05: Added a `ConformingType` interface as suggested by https://github.com/onflow/cadence/pull/3594#discussion_r1785276861.

runtime/sema/gen/main.go
```diff
@@ -164,6 +164,9 @@ type typeDecl struct {
 	memberDeclarations []ast.Declaration
 	nestedTypes        []*typeDecl
 	hasConstructor     bool
+
+	// used in simpleType generation
+	conformances []string
```
turbolent, line 169: Maybe consider making this more type-safe by using a `[]common.TypeID` instead

runtime/sema/gen/main.go
```diff
@@ -736,8 +741,151 @@ func (g *generator) VisitCompositeDeclaration(decl *ast.CompositeDeclaration) (_
...
-	panic("interface declarations are not supported")
+func (g *generator) VisitInterfaceDeclaration(decl *ast.InterfaceDeclaration) (_ struct{}) {
+	compositeKind := decl.CompositeKind
+	switch compositeKind {
+	case common.CompositeKindStructure,
+		common.CompositeKindResource,
+		common.CompositeKindContract:
+		break
+	default:
+		panic(fmt.Sprintf("%s declarations are not supported", compositeKind.Name()))
```
turbolent, line 752:
```suggestion
		panic(fmt.Sprintf("%s interface declarations are not supported", compositeKind.Name()))
```

runtime/sema/gen/main.go
```diff
@@ -736,8 +741,151 @@ func (g *generator) VisitCompositeDeclaration(decl *ast.CompositeDeclaration) (_
 	return
 }
 
-func (*generator) VisitInterfaceDeclaration(_ *ast.InterfaceDeclaration) struct{} {
-	panic("interface declarations are not supported")
+func (g *generator) VisitInterfaceDeclaration(decl *ast.InterfaceDeclaration) (_ struct{}) {
```
turbolent, line 744: Is there anything significantly different from `VisitCompositeDeclaration`? If the code generation is pretty much the same, maybe refactor it into a common function

runtime/sema/gen/main.go
```diff
@@ -1591,6 +1739,9 @@ func simpleTypeLiteral(ty *typeDecl) dst.Expr {
 	//	Comparable:    false,
 	//	Exportable:    false,
 	//	Importable:    false,
+	//  comformances:  []*InterfaceType {
+	//      StructStringer,
```
turbolent, line 1743:
```suggestion
	//      StructStringerType,
```

runtime/sema/gen/main.go
```diff
@@ -2069,6 +2240,100 @@ func compositeTypeLiteral(ty *typeDecl) dst.Expr {
 	}
 }
 
+func interfaceTypeExpr(ty *typeDecl) dst.Expr {
```
turbolent, line 2243: Same as above: Is there anything significantly different from the composite-related functions (`compositeTypeExpr`, `compositeTypeLiteral`)? If the code generation is pretty much the same, maybe refactor it into a common function

runtime/sema/struct_stringer.cdc
```diff
@@ -0,0 +1,7 @@
+/// StructStringer is an interface implemented by all the string convertible structs.
+access(all) 
+struct interface StructStringer {
```
turbolent, line 3: Nice!

runtime/sema/type.go
```diff
@@ -312,6 +312,12 @@ func TypeActivationNestedType(typeActivation *VariableActivation, qualifiedIdent
 	return ty
 }
 
+// allow all types to specify interface conformances
```
turbolent, line 315:
```suggestion
// ConformingType is a type that can conform to interfaces
```

runtime/sema/type.go
```diff
@@ -1190,14 +1196,24 @@ type NumericType struct {
...
 var _ SaturatingArithmeticType = &NumericType{}
 
 func NewNumericType(typeName string) *NumericType {
-	return &NumericType{name: typeName}
+	return &NumericType{
+		name: typeName,
+		conformances: []*InterfaceType{
+			StructStringerType,
+		},
+	}
```
turbolent, line 1216: Maybe instead of creating a new, same conformance set for each numeric type, maybe consider not storing the conformance set (remove the new fields here), and return a global from `EffectiveInterfaceConformanceSet`, see below

runtime/sema/type.go
```diff
@@ -1375,6 +1391,21 @@ func (*NumericType) CheckInstantiated(_ ast.HasPosition, _ common.MemoryGauge, _
...
+
+func (t *NumericType) initializeEffectiveInterfaceConformanceSet() {
+	t.effectiveInterfaceConformanceSetOnce.Do(func() {
+		t.effectiveInterfaceConformanceSet = NewInterfaceSet()
+
+		for _, conformance := range t.conformances {
+			t.effectiveInterfaceConformanceSet.Add(conformance)
+		}
+	})
+}
```
turbolent, line 1407: Follow-up from above:

```suggestion

var numericTypeEffectiveInterfaceConformanceSet *InterfaceSet

func init() {
	numericTypeEffectiveInterfaceConformanceSet = NewInterfaceSet()
	numericTypeEffectiveInterfaceConformanceSet.Add(StructStringerType)
}

func (t *NumericType) EffectiveInterfaceConformanceSet() *InterfaceSet {
	return numericTypeEffectiveInterfaceConformanceSet
}
```

runtime/sema/type.go
```diff
@@ -7244,11 +7276,18 @@ const AddressTypeName = "Address"
 
 // AddressType represents the address type
 type AddressType struct {
```
turbolent, line 7278: Same as above for `NumericType`

runtime/sema/type.go
```diff
@@ -9499,10 +9553,57 @@ func (t *EntitlementMapType) CheckInstantiated(_ ast.HasPosition, _ common.Memor
...
+				continue
+			}
+
+			nestedTypes.Foreach(func(_ string, nestedType Type) {
+				nestedCompositeType, ok := nestedType.(*CompositeType)
+				if !ok {
+					return
+				}
+
+				types = append(types, nestedCompositeType)
```
turbolent, line 9580: Add all nested types (e.g. interface types), not just composite types

```suggestion
				types = append(types, nestedType)
```

runtime/sema/type.go
```diff
@@ -9499,10 +9553,57 @@ func (t *EntitlementMapType) CheckInstantiated(_ ast.HasPosition, _ common.Memor
...
+				continue
+			}
+
+			nestedTypes.Foreach(func(_ string, nestedType Type) {
+				nestedInterfaceType, ok := nestedType.(*InterfaceType)
+				if !ok {
+					return
+				}
+
+				types = append(types, nestedInterfaceType)
```
turbolent, line 9596: Same as above

```suggestion
				types = append(types, nestedType)
```

runtime/sema/type.go
```diff
@@ -9499,10 +9553,57 @@ func (t *EntitlementMapType) CheckInstantiated(_ ast.HasPosition, _ common.Memor
...
+			})
+		}
+
```
turbolent, line 9599: Maybe add a `default` case and `panic(errors.NewUnreachableError())`

runtime/sema/gen/main.go
turbolent, line 744: Please add a test case to `sema/gen/testdata`

turbolent (commented):
Great work!

runtime/sema/gen/main.go
RZhang05: Changed to an `InterfaceType` for simplicity

turbolent:
@SupunS Was this closed by accident?

SupunS:
oh, my bad. I was cleaning up some of my old branches, and must have accidentally deleted the base branch! Thought it was protected. Will revert it.

turbolent:
No worries at all, I was just not sure about the feature branches 👍

sema/path_type.go
```diff
@@ -48,6 +51,9 @@ var StoragePathType = &SimpleType{
 	Comparable:    false,
 	Exportable:    true,
 	Importable:    true,
+	conformances: []*InterfaceType{
+		StructStringerType,
+	},
```
turbolent, line 56: The conformance list here for `StoragePath` and for the other subtypes of `Path` below, should not be necessary, as the subtypes should implicitly conform to the interface list as well.

I guess removing them wouldn't work though, because `SimpleType.EffectiveInterfaceConformanceSet` doesn't take subtyping relationships into account yet (and we do not have that available there, only implicitly in the `IsSubtype` function)

tests/checker/stringer_test.go
```diff
@@ -0,0 +1,75 @@
+/*
+ * Cadence - The resource-oriented smart contract programming language
+ *
+ * Copyright Dapper Labs, Inc.
```
turbolent, line 4:
```suggestion
 * Copyright Flow Foundation
```

turbolent (approved):
Great work!

sema/gen/main.go
```diff
@@ -473,6 +499,8 @@ func (g *generator) VisitCompositeDeclaration(decl *ast.CompositeDeclaration) (_
 	// Check if the declaration is explicitly marked to be generated as a composite type.
 	if _, ok := g.leadingPragma["compositeType"]; ok {
 		generateSimpleType = false
+	} else if !isCompositeType {
```
SupunS, line 502: Nit: inverting this flag to say `isInterfaceType` feels more intuitive

sema/type.go
```diff
@@ -9511,10 +9541,49 @@ func (t *EntitlementMapType) CheckInstantiated(_ ast.HasPosition, _ common.Memor
 	// NO-OP
 }
 
+func extractNativeTypes(
+	types []Type,
+) {
```
SupunS, line 9546: Need to return the resulting `slice` after all the appends, because the `append` function allocates a new slice (unlike maps). So any appends made inside this function would be lost.

SupunS: ah right!

SupunS (approved):

Issue #2566, opened 2023-06-12
Split `runtime/interpreter/value.go` into multiple files


darkdrag00nv2:
### Issue to be solved

The `runtime/interpreter/value.go` file is huge. Even the GitHub syntax highlighting gets disabled on PR reviews because of the huge size.

### Suggested Solution

We should consider splitting the file into different files. A decent split can be:

1. `value_collections.go`
2. `value_numbers.go`
3. `value_options.go`
.. More if needed ..

`value_test.go` would have to be split accordingly.

turbolent:
Agreed, it's grown significantly and new value types have already been added in new files.

Logistically, the plan is to merge `feature/stable-cadence` into `master` soon, when all maybe features have been merged (e.g. Entitlements). Once that is done, and before other major PRs are started which might touch `value.go`, we should split it up 👍

PR #3609, opened 2024-10-10
Split `value.go` into multiple files

SupunS:
Closes https://github.com/onflow/cadence/issues/2566

## Description

This is simply moving the existing code as-is. Nothing else was changed.

We could probably do the same for some more files, like `interpreter.go`, `runtime_test.go`, etc. Can do that in follow-up PRs.

______


turbolent (approved):
🎉

PR #3616, opened 2024-10-17
Update source compatibility suite for latest cadence v1.0.1

### LINKED ISSUE #3608: Operationalize Source compatibility suite
## Goal

- Ensure that unit tests of ecosystem projects (e.g. [Flow core contracts](https://github.com/onflow/flow-core-contracts/)) pass for new Cadence releases
- Once done, we don't need #2503 
- Ask community for any repos they would like to be added to the compatibility check

## Scope

- Clone Git repo of repository of the ecosystem project
- Update the configuration to use latest Cadence version
- Run tests and ensure they pass

## Context

- Current code lives in https://github.com/onflow/cadence/tree/master/compat
- Current code rewrites Cadence code (e.g. updates import statements), and parses and typechecks it. 
  This is hard to maintain/brittle, we might want to remove this functionality
- Some projects, like the [Flow core contracts](https://github.com/onflow/flow-core-contracts/), have their tests are written in Go and use Cadence as a Go dependency
- Some projects use the [Cadence Testing framework](https://cadence-lang.org/docs/testing-framework)

RZhang05:
Closes #3608

## Description

- Removes the parsing and typechecking functionality
- Run go tests on `flow-go v0.37.17`, `cadence v1.0.1` and pass

______


compat/main.py
```diff
@@ -122,17 +111,17 @@ class GoTest:
     path: str
     command: str
 
-    def run(self, working_dir: Path, prepare: bool, go_test_ref: Optional[str]) -> bool:
-        if go_test_ref:
-            replacement = f'github.com/onflow/cadence@{go_test_ref}'
-        else:
-            replacement = shlex.quote(str(Path.cwd().parent.absolute()))
+    def run(self, working_dir: Path, prepare: bool, cadence_version: str, flowgo_version: str) -> bool:
```
turbolent, line 114: Maybe keep the Cadence version optional, and keep the original `else` branch. The idea here is that when not specifying a particular version, the default is to use the working copy of Cadence, by using the absolute path to it. This is useful for local development

compat/main.py
```diff
@@ -200,6 +189,11 @@ def from_dict(cls, data: Dict):
 
     def _clone(self, working_dir: Path):
         if working_dir.exists():
+            for root, dirs, files in os.walk(working_dir):  
+                for dir in dirs:
+                    os.chmod(os.path.join(root, dir), stat.S_IRWXU)
+                for file in files:
+                    os.chmod(os.path.join(root, file), stat.S_IRWXU)
```
turbolent, line 196: Why do we have to make all files executable? That might be dangerous and should not be necessary.

Maybe just make it readable and writable (and not executable):

```suggestion
                for file in files:
                    os.chmod(os.path.join(root, file), stat.S_IRUSR | stat.S_IWUSR)
```

compat/main.py
```diff
@@ -122,17 +111,17 @@ class GoTest:
...
-            replacement = f'github.com/onflow/cadence@{go_test_ref}'
-        else:
-            replacement = shlex.quote(str(Path.cwd().parent.absolute()))
+    def run(self, working_dir: Path, prepare: bool, cadence_version: str, flowgo_version: str) -> bool:
+        cadence_replacement = f'github.com/onflow/cadence@{cadence_version}'
+        flowgo_replacement = f'github.com/onflow/flow-go@{flowgo_version}'
 
         with cwd(working_dir / self.path):
             if prepare:
+                logger.info("Editing dependencies")
```
turbolent, line 120: Good idea to add some logging 👍

compat/main.py
```diff
@@ -4,7 +4,7 @@
 import html
 import json
 import logging
-import os
+import os, stat
```
turbolent, line 7: Maybe put this on a separate line to follow existing style

turbolent (commented):
Great work! I didn't realize we already had the go mod replace functionality

compat/main.py
```diff
@@ -122,17 +112,28 @@ class GoTest:
...
+            cadence_replacement = f'github.com/onflow/cadence@{cadence_version}'
+        else:
+            cadence_replacement = Path.cwd().parent.absolute().resolve().as_posix()
+
+        if flowgo_version:
+            flowgo_replacement = f'github.com/onflow/flow-go@{flowgo_version}'
         else:
-            replacement = shlex.quote(str(Path.cwd().parent.absolute()))
+            # default flow-go version is 0.37.16 the one corresponding to cadence v1.0.0
+            flowgo_replacement = f'github.com/onflow/flow-go@v0.37.16'
```
turbolent, line 125: This will quickly get out of date and will have to get updated.

It might be better to just make this required. Even better, we could also automate this: By leveraging Go's package manager proxy, we can fetch the latest release: Fetching https://proxy.golang.org/github.com/onflow/flow-go/@latest we can use the `Version` field.

compat/main.py
```diff
@@ -122,17 +112,28 @@ class GoTest:
...
     command: str
 
-    def run(self, working_dir: Path, prepare: bool, go_test_ref: Optional[str]) -> bool:
-        if go_test_ref:
-            replacement = f'github.com/onflow/cadence@{go_test_ref}'
+    def run(self, working_dir: Path, prepare: bool, cadence_version: Optional[str], flowgo_version: Optional[str]) -> bool:
+        if cadence_version:
+            cadence_replacement = f'github.com/onflow/cadence@{cadence_version}'
+        else:
+            cadence_replacement = Path.cwd().parent.absolute().resolve().as_posix()
```
turbolent, line 119: This probably needs to be quoted (see original), in case the path has spaces

turbolent (commented):
Now that the parsing and typechecking functionality has been removed, can you please remove the relevant parts from the configuration files, i.e. the `files:` property?

compat/main.py
RZhang05: Replaced this with `go get ...flow-go@latest`.

turbolent: Ah right, given we just run go get anyway, that works and is much simpler 👌

turbolent (approved):
Thank you for removing the unused code! Could you please also remove the unused parts from the configuration files in https://github.com/onflow/cadence/blob/0bb1b6964b9ce61a29f0ce4359c169d724ee9813/compat/suite ?
For example: https://github.com/onflow/cadence/blob/0bb1b6964b9ce61a29f0ce4359c169d724ee9813/compat/suite/flow-core-contracts.yaml#L10-L139

PR #3631, opened 2024-10-18
Update contracts-compatibility-checker

SupunS:
Depends on https://github.com/onflow/flow-go/pull/6571

## Description

For each PR, the contracts backward-compatibility checker runs with the latest changes (~HEAD). The CI always replace the dependency to be the latest, however, if the tool wasn't updated to the latest cadence changes, then it is only caught at the time of CI. 

Therefore, make the tool to always use the latest cadence, so any changes that would break the tool is caught locally, when running the tests.

______


turbolent (approved):
Nice!

tools/compatibility-check/contracts_checker.go
```diff
@@ -117,7 +117,10 @@ func (c *ContractsChecker) analyze(
 	config *analysis.Config,
 	locations []common.Location,
 ) {
-	programs := make(analysis.Programs, len(locations))
+	programs := analysis.Programs{
+		Programs:                  make(map[common.Location]*analysis.Program, len(locations)),
+		CryptoContractElaboration: config.CryptoContractElaboration,
```
SupunS, line 122: This currently doesn't work, since the tool doesn't have the access to Crypto contract's elaboration. Not sure what the correct solution is though. Maybe we would have to get it from FVM (adds a dependency to fvm).

turbolent: Yeah, we'll likely need to pull in FVM, which is not ideal, but needed if we want to check contracts exactly like they are checked on-chain

turbolent: @SupunS I can help with / take this over if you want to

SupunS: I pushed this fix https://github.com/onflow/cadence/pull/3631/commits/cfd6412917aee4a9d78265a8eccc935b7bb2daba, to only convert the `Crypto` import to a proper address-qualified import, so that it will resolve Crypto as any other contract. But unfortunately that requires the Crypto contract to be deployed on chain. So this will still be broken until the next network upgrades.

SupunS: @turbolent if you know an easy way to get the elaboration from the FVM directly, that would be great!

SupunS: For now, I provided the `Crypto` contract's code manually for the tool (https://github.com/onflow/cadence/pull/3631/commits/871780ed35abc62eb3fe5795d68ee39e6e4d23fa). Left a TODO: to revert it once the contract is deployed on-chain.

SupunS:
CI check (https://github.com/onflow/cadence/actions/runs/11448933364/job/31853489202?pr=3631), is failing at "checking contracts using master" step because the tool's code on master is outdated, due to the changes of this PR. But the workflow passes for "checking contracts using head of the PR" (as expected), and the `Crypto` contract is properly resolved there. So the CI-check would be back to normal once this PR is merged.

PR #3638, opened 2024-10-22
Add built-in CCF contract with encode function

turbolent:
## Description

Add a new built-in contract `CCF`, which allows encoding an exportable value to .

The parameter of `encode` is a reference, so it can accept both structs and resources, and supports move-semantics of resources (i.e. does not consume the passed resource-kinded value)

For now this is a PoC showing how this could be done.

```[tasklist]
### TODO
- [ ] Determine interest and use-cases
- [ ] Add more tests
- [ ] Decode function?
``` 

______


turbolent:
Unfortunately there's currently a cyclic import between `encoding/ccf` and `stdlib` at the moment, which we'll need to resolve

PR #3625, opened 2024-10-18
Support expressions in string template

RZhang05:
Working towards https://github.com/onflow/cadence/issues/3579

## Description

This PR extends the work in #3585, as planned to support expressions. 

______

- [x] Targeted PR against `feature/string-templates` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [ ] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [ ] Added appropriate labels

PR #3662, opened 2024-10-30
Support expressions in string template

RZhang05:
Working towards https://github.com/onflow/cadence/issues/3579
## Description

This PR extends the work in https://github.com/onflow/cadence/pull/3585, as planned to support expressions.

Reopening of #3625

______

### DIFF
--- a/interpreter/misc_test.go
+++ b/interpreter/misc_test.go
@@ -12509,4 +12509,53 @@ func TestInterpretStringTemplates(t *testing.T) {
 			inter.Globals.Get("x").GetValue(inter),
 		)
 	})
+
+	t.Run("func", func(t *testing.T) {
+		t.Parallel()
+
+		inter := parseCheckAndInterpret(t, `
+			let add = fun(): Int {
+				return 2+2
+			}
+			let x: String = "\(add())"
+		`)
+
+		AssertValuesEqual(
+			t,
+			inter,
+			interpreter.NewUnmeteredStringValue("4"),
+			inter.Globals.Get("x").GetValue(inter),
+		)
+	})
+
+	t.Run("ternary", func(t *testing.T) {
+		t.Parallel()
+
+		inter := parseCheckAndInterpret(t, `
+			let z = false
+			let x: String = "\(z ? "foo" : "bar" )"
+		`)
+
+		AssertValuesEqual(
+			t,
+			inter,
+			interpreter.NewUnmeteredStringValue("bar"),
+			inter.Globals.Get("x").GetValue(inter),
+		)
+	})
+
+	t.Run("nested", func(t *testing.T) {
+		t.Parallel()
+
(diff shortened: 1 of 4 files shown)


turbolent (approved):
Nice!

parser/expression_test.go
```diff
@@ -6496,6 +6505,45 @@ func TestParseStringTemplate(t *testing.T) {
 
 		AssertEqualWithDiff(t, expected, actual)
 	})
+
+	t.Run("valid, extra closing paren", func(t *testing.T) {
```
SupunS, line 6509:
```suggestion
	t.Run("invalid, extra closing paren", func(t *testing.T) {
```

parser/expression_test.go
```diff
@@ -6293,24 +6284,42 @@ func TestParseStringTemplate(t *testing.T) {
...
+		_, errs := testParseExpression(`
+			"\(add"
+		`)
+
+		var err error
+		if len(errs) > 0 {
+			err = Error{
+				Errors: errs,
+			}
+		}
```
SupunS, line 6303: Instead of creating a new error, we can simply check whether the `errs` are empty. i.e:
```suggestion
		require.NotEmpty(t, errs)
```

Same for other similar places as well.

interpreter/misc_test.go
SupunS, line 1: Do we have any tests for nested string-templates (i.e: string-template inside a string-template)? if not, could you please add one?

SupunS (commented):
Nice!

interpreter/misc_test.go
RZhang05: There is one here [expression_test.go#L6335](https://github.com/onflow/cadence/blob/master/parser/expression_test.go#L6335).

parser/expression_test.go
SupunS: ah right

interpreter/misc_test.go
SupunS: So only the string-templates are not supported as expression inside a string-template, and all other expressions are supported? Was that intentionally rejected, or a side effect of how the parser works? Sorry I vaguely remember discussing about it before, but don't really remember the details.

RZhang05: It is intentionally rejected, this simplifies the lexer to not have to deal with recursively nested states.

SupunS: great!

parser/expression_test.go
```diff
@@ -6496,6 +6463,45 @@ func TestParseStringTemplate(t *testing.T) {
...
+
+		require.NoError(t, err)
```
SupunS, line 6482:
```suggestion
		require.Empty(t, errs)
```

SupunS (approved):
Nice!

SupunS (approved):

PR #3680, opened 2024-11-14
[Account Storage Maps] Refactor migration

turbolent:
Depends on #3678 

## Description

- Simplify the bookkeeping in V2: only keep a simple map, an ordered map is unnecessary
- Simplify the domain register migration to allow migration of a single account
- For now, bring back the automatic migration of written-to accounts during commit.
  We'll keep discussing this in the upcoming sync with the Execution team.

______

### DIFF
--- a/runtime/storage.go
+++ b/runtime/storage.go
@@ -49,7 +49,7 @@ type Storage struct {
 
 	// cachedV1Accounts contains the cached result of determining
 	// if the account is in storage format v1 or not.
-	cachedV1Accounts *orderedmap.OrderedMap[common.Address, bool]
+	cachedV1Accounts map[common.Address]bool
 
 	// contractUpdates is a cache of contract updates.
 	// Key is StorageKey{contract_address, contract_name} and value is contract composite value.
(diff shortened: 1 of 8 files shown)


runtime/storage.go
```diff
@@ -379,81 +379,95 @@ func (s *Storage) commit(inter *interpreter.Interpreter, commitContractUpdates b
...
+		inter,
+		s.memoryGauge,
+		getDomainStorageMap,
+	)
+
+	for pair := s.cachedV1Accounts.Oldest(); pair != nil; pair = pair.Next() {
+		address := pair.Key
+		isV1 := pair.Value
+
+		if !isV1 || !s.PersistentSlabStorage.HasUnsavedChanges(atree.Address(address)) {
```
turbolent, line 450: For now I hard-coded the predicate. We can discuss this in the upcoming sync

runtime/storage.go
```diff
@@ -379,81 +379,95 @@ func (s *Storage) commit(inter *interpreter.Interpreter, commitContractUpdates b
...
+			continue
+		}
+
+		accountStorageMap, err := migrator.MigrateAccount(address)
+		if err != nil {
+			return err
+		}
+
+		// TODO: is this all that is needed?
```
turbolent, line 459: The update of the bookkeeping is the same as before: Make V2 aware of the new account storage map, so it is not re-loaded, and consider the account no longer as V1. From what I can see this should be sufficient, but it's good to double check this is sufficient. @fxamacker

SupunS (approved):
Similar to the other PR, I don't have a lot of context on the business logic. So I mostly reviewed for the general Go code.

turbolent:
Merging into #3678 to have all changes needed to pass CI in one PR, as discussed in https://github.com/onflow/cadence/pull/3678#issuecomment-2479961625

Issue #3693, opened 2024-11-22
Make existing interpreter values reusable in VM


turbolent:
We are evaluating how Cadence programs can be executed using an alternative approach, namely compilation (code generation). Programs executed using the alternative approach must behave exactly like the current approach, interpretation using the interpreter.

Currently the implementation of the majority of the built-in functionality of Cadence is part of the package implementing the interpreter, `interpreter`. The implementation is also tightly coupled to the interpreter.

If we can find a way to decouple the built-in functionality from the interpreter, we can reuse it for the alternative execution approach, and do not have to re-implement the functionality, which is a lot of work, and very difficult to get correct, as it must function exactly like the existing solution.

Decouple and define (an) interface(s) for the following components, so they can be reused outside of the interpreter:

- Values (`interpreter.Value` and implementations)
- Static types (`interpreter.StaticType` and implementations)
- Encoding (`encode.go` and `decode.go`)

This is difficult, as values are tightly coupled to the interpreter (most functions of the values get passed the full interpreter).

We might want to refactor the code gradually instead of all parts all at once.

`interpreter.Value` methods depend on the `Interpreter` for the following categories of operations, for example:

- Metering of computation: `ReportComputation`
- Metering of memory usage: `MeterMemory` 
- Get configuration: for example to determine if feature is enabled (e.g. tracing), invoke callbacks, etc.
- Tracing: e.g. `reportDictionaryValueDestroyTrace`
- Resource tracking: e.g. `withResourceDestruction`
- Reference tracking: e.g. `invalidateReferencedResources`
- Function value invocation: e.g. `invokeFunctionValue`
- Storage writes: e.g. `WriteStored`, `RemoveReferencedSlab`
- Storage reads: e.g. `ReadStored`
- Ensure invariants: e.g. `withMutationPrevention`
- Atree validation: e.g. `maybeValidateAtreeValue`, `maybeValidateAtreeStorage`
- Static type conversion and validation: e.g. `MustConvertStaticToSemaType`, `ExpectType`, `checkContainerMutation`
- etc.

SupunS:
I think we can close this one now?

turbolent:
I don't quite know if this is necessary for making the interpreter values reusable in the VM, but I had a quick look at https://github.com/onflow/cadence/blob/5c376e1efef3111cc6a57b0403fe1690ff351515/interpreter/value.go, and it looks like there are a few functions that need an interpreter left:
[...]

PR #3698, opened 2024-11-26
Decouple interpreter values from interpreter

turbolent:
Work towards #3693

## Description

Demonstrate how we could decouple methods of `interpreter.Value` types from `Interpreter`.

In particular, define interfaces for the functionality that is required by implementations of the `Value.StaticType` method and change the parameter type from `Interpreter` to the new combined interface (for now called `StaticTypeGetter`, for lack of a better name – suggestions are very welcome).

Most implementations of `Value.StaticType` only need to perform memory metering, so need a `common.MemoryGauge`. However, `StorageReferenceValue.StaticType` dereferences, which requires access to storage and a means to convert a static type to a sema type.

If this approach looks good, we can gradually refactor other methods of `Value` in a similar way.
Suggestions for improvements or alternatives are very welcome!

______


jsproz (approved):

interpreter/interface.go
SupunS, line 1: Nice! Having only the 'minimal' set of functionality + having them as interfaces makes the life a lot easier 👌

interpreter/value.go
```diff
@@ -159,30 +159,37 @@ type MemberAccessibleValue interface {
 	SetMember(interpreter *Interpreter, locationRange LocationRange, name string, value Value) bool
 }
 
+type ValueComparisonContext interface {
+	common.MemoryGauge
+	ValueStaticTypeContext
```
SupunS, line 164: I guess `ValueStaticTypeContext` is included because they are needed for container equality (getting children from containers)?

SupunS (approved):
Great!

interpreter/value.go
turbolent: No, this is actually needed for some implementations of the comparison functions, like `Less`, especially in implementations of number value types, like `IntValue`, mainly to get the static type information for producing useful error messages, e.g.:

```go
func (v IntValue) Less(context ValueComparisonContext, other ComparableValue, locationRange LocationRange) BoolValue {
	o, ok := other.(IntValue)
	if !ok {
		panic(InvalidOperandsError{
			Operation:     ast.OperationLess,
			LeftType:      v.StaticType(context),
			RightType:     other.StaticType(context),
			LocationRange: locationRange,
		})
	}
    // ...
```

PR #3707, opened 2024-12-02
Machine readable syntax specification

### LINKED ISSUE #3686: Machine Readable Syntax Specification
### Issue to be solved

Create a machine readable specification for the Cadence syntax that can be used to generate a compiler frontend.

### Suggested Solution

- [x] 1. A implementation can be generated from the specification and it parses the majority of contracts and transactions on MainNet.
- [x] 2. The specification will have a section describing the notation used in the specification.

jsproz:
Closes #3686

## Description

Added a machine readable syntax specification for the Cadence language and a Syntax Notation document.

______

### DIFF
--- a/docs/syntax/Cadence.syntax
+++ b/docs/syntax/Cadence.syntax
@@ -1,0 +1,1482 @@
+
+%scanner
+
+*
+<spaceToken>:
+        whitespace
+    ;
+
+whitespace:
+        1*whitespaceItem
+    ;
+
+whitespaceItem:
+        singleLineComment
+    |   multipleLineComment
+    |   whitespaceCharacter
+    |   lineBreak
+    ;
+
+whitespaceCharacter:
+        U+0000,U+0009,U+000B-U+000C,U+0020
+    ;
+
+lineBreak:
+        "\r" "\n"
+    |   "\r"
+    |   "\n" ;
+
+singleLineComment:
+        "/" "/" *1singleLineCommentText lineBreak
+    ;
+
+singleLineCommentText:
+        1*singleLineCommentTextItem
+    ;
+
+singleLineCommentTextItem:
+        ~"\n\r" ;
+
+multipleLineComment:
+        multipleLineCommentStart *1multipleLineCommentText multipleLineCommentEnd
+    ;
+
+multipleLineCommentStart:
+        "/" "*" ;
+
+multipleLineCommentEnd:
+        "*" "/" ;
+
+multipleLineCommentText:
+        1*multipleLineCommentTextItem
+    ;
+
+multipleLineCommentTextItem:
+        multipleLineCommentTextCharacter
+    |   multipleLineComment
+    ;
+
+multipleLineCommentTextCharacter:
+        ~multipleLineCommentStartOrEnd
+    ;
+
+multipleLineCommentStartOrEnd:
+        multipleLineCommentStart
+    |   multipleLineCommentEnd
+    ;
+
+<identifierToken>:
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz" *identifierFollower
+    ;
+
+identifierFollower:
+        "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz" ;
+
+<fixedPointToken>:
+        "0123456789" *"0123456789_" "." "0123456789" *"0123456789_"
+    ;
+
+<decimalToken>:
+        "0"
+    |   "123456789" *"0123456789_" ;
+
+<binaryToken>:
+        "0" "b" 1*"01_" ;
+
+<octalToken>:
+        "0" "o" 1*"01234567_" ;
+
+<hexadecimalToken>:
+        "0" "x" 1*"0123456789ABCDEFabcdef_" ;
+
+<invalidNumberToken>:
+        "0" "acdefghijklmnpqrstuvwyzABCDEFGHIJKLMNOPQRSTUVWXYZ" "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ_"
+    ;
+
+<stringToken>:
+        "\"" *quotedTextItem "\"" ;
+
+quotedTextItem:
+        quotedTextItemCharacter
+    |   escapedCharacter
+    ;
+
+quotedTextItemCharacter:
+        ~nonquotedTextItemCharacter
+    ;
+
+nonquotedTextItemCharacter:
+        "\"\r\n\\" ;
+
+escapedCharacter:
+        escapedUnicodeScalar
+    |   escapedCCharacter
+    ;
+
+escapedUnicodeScalar:
+        "\\" "u" "{" 1*8unicodeScalarDigit "}" ;
+
+unicodeScalarDigit:
+        "0123456789abcdefABCDEF" ;
+
+escapedCCharacter:
+        "\\" "0\\tnr\"'" ;
+
+%parser
+
+*
+program:
+         *declaration( declarationSeparator ) $
+    ;
+
+declarationSeparator:
+        semicolon
+    |   empty
+    ;
+
+empty:
+    ;
+
+semicolon:
+        ";"
+    ;
+
+declaration:
+        pragmaDeclaration
+    |   variableDeclaration
+    |   functionDeclaration
+    |   importDeclaration
+    |   eventDeclaration
+    |   structCompositeDeclaration
+    |   resourceCompositeDeclaration
+    |   contractCompositeDeclaration
+    |   enumCompositeDeclaration
+    |   structInterfaceDeclaration
+    |   resourceInterfaceDeclaration
+    |   contractInterfaceDeclaration
+    |   enumInterfaceDeclaration
+    |   entitlementDeclaration
+    |   entitlementMappingDeclaration
+    |   attachmentDeclaration
+    |   transactionDeclaration
+    ;
+
+attachmentDeclaration:
+        access attachmentKeyword nonReservedIdentifier forKeyword nominalType conformances "{" membersAndNestedDeclarations "}"
+    ;
+
+transactionDeclaration:
+        transactionKeyword *1parameterList "{" fields *1prepareFunctionDeclaration *1preConditions *1executeOrPostConditions "}"
+    ;
+
+executeOrPostConditions:
+        executeOrPostConditionsExecute
+    |   executeOrPostConditionsExecutePostConditions
+    |   executeOrPostConditionsPostConditions
+    |   executeOrPostConditionsPostConditionsExecute
+    ;
+
+executeOrPostConditionsExecute:
+        execute
+    ;
+
+executeOrPostConditionsExecutePostConditions:
+        execute postConditions
+    ;
+
+executeOrPostConditionsPostConditions:
+        postConditions
+    ;
+
+executeOrPostConditionsPostConditionsExecute:
+        postConditions execute
+    ;
+
+execute:
+        identifier block
+    ;
+
+importDeclaration:
+        importKeyword *1importDeclarationFromPhrase importDeclarationSource
+    ;
+
+importDeclarationFromPhrase:
+        1*identifier( "," ) fromKeyword
+    ;
+
+importDeclarationSource:
+        stringLiteral
+    |   hexadecimalLiteral
+    |   identifier
+    ;
+
+access:
+        accessPriv
+    |   accessPub
+    |   accessSpecified
+    |   accessNotSpecified
+    ;
+
+accessNotSpecified:
+    ;
+
+accessPriv:
+        privKeyword
+    ;
+
+accessPub:
+        pubKeyword *1accessPubSet
+    ;
+
+accessPubSet:
+        "(" setKeyword ")"
+    ;
+
+accessSpecified:
+        accessKeyword "(" accessAccessValue ")"
+    ;
+
+accessAccessValue:
+        selfKeyword
+    |   contractKeyword
+    |   accountKeyword
+    |   allKeyword
+    |   accessAccessValueEntitlement
+    |   accessAccessValueEntitlementMapping
+    ;
+
+accessAccessValueEntitlement:
+        accessAccessValueEntitlementConjunction
+    |   accessAccessValueEntitlementDisjunction
+    ;
+
+accessAccessValueEntitlementConjunction:
+        1*fullType( "," )
+    ;
+
+accessAccessValueEntitlementDisjunction:
+        1*fullType( "|" )
+    ;
+
+accessAccessValueEntitlementMapping:
+        mappingKeyword fullType
+    ;
+
+conformances:
+        conformancesTypes
+    |   conformancesEmpty
+    ;
+
+conformancesTypes:
+        ":" 1*nominalType( "," )
+    ;
+
+conformancesEmpty:
+    ;
+
+field:
+        letField
+    |   varField
+    |   unspecifiedField
+    ;
+
+letField:
+        access letKeyword nonReservedIdentifier ":" typeAnnotation
+    ;
+
+varField:
+        access varKeyword nonReservedIdentifier ":" typeAnnotation
+    ;
+
+unspecifiedField:
+        access nonReservedIdentifier ":" typeAnnotation
+    ;
+
+fields:
+        *fieldsEntry
+    ;
+
+fieldsEntry:
+        field *1semicolon
+    ;
+
+structCompositeDeclaration:
+        access structKeyword nonReservedIdentifier conformances "{" membersAndNestedDeclarations "}"
+    ;
+
+resourceCompositeDeclaration:
+        access resourceKeyword nonReservedIdentifier conformances "{" membersAndNestedDeclarations "}"
+    ;
+
+contractCompositeDeclaration:
+        access contractKeyword nonReservedIdentifier conformances "{" membersAndNestedDeclarations "}"
+    ;
+
+enumCompositeDeclaration:
+        access enumKeyword nonReservedIdentifier conformances "{" *semicolonEnumerationDeclarationCase "}"
+    ;
+
+structInterfaceDeclaration:
+        access structKeyword interfaceKeyword nonReservedIdentifier *1interfaceInheritance "{" membersAndNestedDeclarations "}"
+    ;
+
+resourceInterfaceDeclaration:
+        access resourceKeyword interfaceKeyword nonReservedIdentifier *1interfaceInheritance "{" membersAndNestedDeclarations "}"
+    ;
+
+contractInterfaceDeclaration:
+        access contractKeyword interfaceKeyword nonReservedIdentifier *1interfaceInheritance "{" membersAndNestedDeclarations "}"
+    ;
+
+enumInterfaceDeclaration:
+        access enumKeyword interfaceKeyword nonReservedIdentifier *1interfaceInheritance "{" *semicolonEnumerationDeclarationCase "}"
+    ;
+
+interfaceInheritance:
+        ":" 1*fullType( "," )
+    ;
+
+membersAndNestedDeclarations:
+        *membersAndNestedDeclarationsEntry
+    ;
+
+membersAndNestedDeclarationsEntry:
+        memberOrNestedDeclaration declarationSeparator
+    ;
+
+memberOrNestedDeclaration:
+        field
+    |   initFunctionDeclaration
+    |   destroyFunctionDeclaration
+    |   prepareFunctionDeclaration
+    |   functionDeclaration
+    |   structCompositeDeclaration
+    |   resourceCompositeDeclaration
+    |   contractCompositeDeclaration
+    |   enumCompositeDeclaration
+    |   structInterfaceDeclaration
+    |   resourceInterfaceDeclaration
+    |   contractInterfaceDeclaration
+    |   enumInterfaceDeclaration
+    |   eventDeclaration
+    |   pragmaDeclaration
+    |   entitlementDeclaration
+    |   entitlementMappingDeclaration
+    |   attachmentDeclaration
+    ;
+
+initFunctionDeclaration:
+        access *1viewKeyword initKeyword parameterList *1functionBlock
+    ;
+
+destroyFunctionDeclaration:
+        access *1viewKeyword destroyKeyword parameterList *1functionBlock
+    ;
+
+prepareFunctionDeclaration:
+        access *1viewKeyword prepareKeyword parameterList *1functionBlock
+    ;
+
+functionDeclaration:
+        access *1viewKeyword funKeyword nonReservedIdentifier parameterList *1functionDeclarationTypeAnnotation *1functionBlock
+    ;
+
+functionDeclarationTypeAnnotation:
+        ":" typeAnnotation
+    ;
+
+eventDeclaration:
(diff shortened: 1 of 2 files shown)


docs/syntax/Syntax Notation.md
```diff
@@ -0,0 +1,255 @@
...
+        singleLineComment
+    |   multipleLineComment
+    |   whitespaceCharacter
+    |   lineBreak
+    ;
+```
+
+### Set Rule
+
+Set rules consist of a sequence of intervals separated by a comma (`,`). These intervals can either represent a single Unicode code point or a range of code points, separated by a hyphen (`-`).
```
turbolent, line 38: Aren't alternations with `|` also sets? Could we maybe just consistently use `|` to avoid unnecessarily adding a new operator uncommon in other notions

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+        ~multipleLineCommentStartOrEnd
+    ;
+
+multipleLineCommentStartOrEnd:
+        multipleLineCommentStart
+    |   multipleLineCommentEnd
+    ;
+
+<identifierToken>:
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz" *identifierFollower
```
turbolent, line 69: Is it possible to express these as ranges? e.g. `[A-Z_a-z]` like in flex?

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+        "0123456789abcdefABCDEF" ;
+
+escapedCCharacter:
+        "\\" "0\\tnr\"'" ;
+
+%parser
+
+*
+program:
+         *declaration( declarationSeparator ) $
```
turbolent, line 129: Could we make it more explicit that the notation here means "separated by"? The parentheses look like grouping (like in other notations), and at first glance it looks like they are unnecessary and can be removed.

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+
+resourceInterfaceDeclaration:
+        access resourceKeyword interfaceKeyword nonReservedIdentifier *1interfaceInheritance "{" membersAndNestedDeclarations "}"
+    ;
+
+contractInterfaceDeclaration:
+        access contractKeyword interfaceKeyword nonReservedIdentifier *1interfaceInheritance "{" membersAndNestedDeclarations "}"
+    ;
+
+enumInterfaceDeclaration:
```
turbolent, line 331: We might want to indicate that enum interfaces are illegal, but this rule exists to produce better error messages. We might in general make this syntax definition as strict as possible, so maybe just remove the rule

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+
+importDeclarationSource:
+        stringLiteral
+    |   hexadecimalLiteral
+    |   identifier
+    ;
+
+access:
+        accessPriv
+    |   accessPub
```
turbolent, line 215: `pub`, `pub(set)`, and `priv` got removed in Cadence 1.0

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+fullTypeReferenceAuthNoEntitlements:
+        authKeyword
+    ;
+
+fullTypeReferenceAuthMapping:
+        authKeyword "(" mappingKeyword nominalType ")"
+    ;
+
+fullTypeOptional:
+        . optional
```
turbolent, line 516: The `.` looks like "any character" here, like in other notations. Maybe leave a comment

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+
+transferMoveForced:
+        moveForced
+    ;
+
+transferMove:
+        move
+    ;
+
+expression:
```
turbolent, line 810: The precedence numbers are an implementation detail of the current parser implementation, maybe remove them from the document as they are not necessary and probably just confusing. The rules are already structured in a way where they encode the precedence rules.

turbolent (commented):
Nice work!

docs/syntax/Cadence.syntax
jsproz: This language does not have a grouping operator.

docs/syntax/Cadence.syntax
jsproz: I was thinking we should have two versions of the syntax. A strict grammar that is allowed by the compiler, and a more general, internal syntax that is used to produce useful error messages.

docs/syntax/Cadence.syntax
jsproz: This syntax is focused on duplicating the behavior of the interpreter's parser.  These would produce errors in the semantic checks. That should also be part of the specification.

docs/syntax/Cadence.syntax
jsproz: The `.` and `_` were chosen to represent adjacent and not-adjacent. The language for the parser section does not match individual characters. It matches tokens.

docs/syntax/Cadence.syntax
jsproz: That was the intention. They were there only to help define the grammar. I forgot to remove them.

docs/syntax/Cadence.syntax
turbolent: In that case, could you please leave a comment above, with an equivalent rule in somehting like EBNF notation? Same for the comments above (adding a comment for ranges)

docs/syntax/Cadence.syntax
turbolent: That's a good idea. To begin with we might at least indicate which variant of the syntax this is

docs/syntax/Cadence.syntax
turbolent: Maybe at least leave a comment documenting this decision, e.g.

`# Invalid since Cadence 1.0, so not part of the strict syntax`

docs/syntax/Cadence.syntax
turbolent: Right, that's OK. It might not be obvious to readers unfamiliar with this custom notation, and given the `.` notation means something else in other commonly used notations, maybe help the reader here and point out that this does not mean "any character" but "adjacent"

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+    ;
+
+fullType:
+        fullTypeNormal
+    |   fullTypeNested
+    ;
+
+fullTypeNormal:
+        *1fullTypeReference innerType *fullTypeOptional
+    ;
```
turbolent, line 486: Maybe encode the precedence rules for types in nested rules, like it is done for expressions, see the `typeLeftBindingPower*` constants. For example, references bind tighter than optionals

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+
+baseType:
+        nominalType
+    |   functionType
+    |   variableSizedType
+    |   constantSizedType
+    |   dictionaryType
+    ;
+
+typeRestrictions:
```
turbolent, line 548: Restriction types got removed in Cadence 1.0. Instead, there are only intersection types, which have this syntax, but not as a "suffix" to a type. For example, before a type `T` could be restricted with `T{...}`, where `...` is a list of **zero** or more nominal types. Now there is just the syntax for intersection types `{...}` where `...` is a list of **one** or more nominal types.

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+contractInterfaceDeclaration:
+        access contractKeyword interfaceKeyword nonReservedIdentifier *1interfaceInheritance "{" membersAndNestedDeclarations "}"
+    ;
+
+enumInterfaceDeclaration:
+        access enumKeyword interfaceKeyword nonReservedIdentifier *1interfaceInheritance "{" *semicolonEnumerationDeclarationCase "}"
+    ;
+
+interfaceInheritance:
+        ":" 1*fullType( "," )
```
turbolent, line 336: Only nominal types are allowed

turbolent (commented):
Did another pass over types as requested. Mainly the precedence rules need to be encoded properly, and the only other thing is the intersection types.

Also did another pass over all declarations, and left a few minor comments, especially regarding what types are allowed.

Will do another pass over expressions tomorrow.

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+orExpression:  // 30 '||'
+        1*andExpression( "||" )
+    ;
+
+andExpression:  // 40
+        1*comparisonExpression( "&&" )
+    ;
+
+comparisonExpression:  // 50
+        1*2nilCoalescingExpression( comparisonOp )
```
turbolent, line 839: Though type checking might reject nested comparison expressions, parsing still allows them.
So this should not have an upper bound in the multiplicity:
```suggestion
        1*nilCoalescingExpression( comparisonOp )
```

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+
+nullDenotation:
+        identifierExpression
+    ;
+
+leftDenotation:
+        forceUnwrapOp
+    |   accessExpression
+    |   invocationExpression
+    ;
```
turbolent, line 910: "null denotation" and "left denotation" are terminology of Pratt's Top Down Operator Precedence parsing technique, i.e. are an implementation detail of the current Go parser. They shouldn't be used as rule names.

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+
+castingExpressionCast:
+        castingOp typeAnnotation
+    ;
+
+prefixExpression:  // 150
+        *1prefixOp postfixExpression
+    ;
+
+postfixExpression:  // 160
```
turbolent, line 890: Force unwrapping is on this precedence level

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+leftDenotation:
+        forceUnwrapOp
+    |   accessExpression
+    |   invocationExpression
+    ;
+
+accessExpression:  // 170
+        memberExpression
+    |   indexExpression
+    ;
```
turbolent, line 915: member expression, indexing expression, and invocation expression are all part of the same precedence and should probably be in the same rule. As noted above, force unwrapping has lower precedence (needs to be in a "higher" rule)

turbolent (commented):
Did another pass reviewing the expression parsing. Looks good, a couple notes.

Note that some expressions need newline differentiation to resolve ambiguities when expressions are used as statements, see the "(* if no line terminator ahead *)" comments in the expression rules of the EBNF grammar, like https://github.com/onflow/cadence/blob/53642201c5274fe864a06455f0b8bab18ca7032a/docs/cadence.ebnf#L457.

For example:

```
foo(bar) // invocation
```

```
foo    // identifier expression
(bar)  // parenthesized expression
```

docs/syntax/Syntax Notation.md
jsproz: The `|` allows for defining separate productions to the same rule. Only one production can match. It does not define members of a set.

docs/syntax/Cadence.syntax
jsproz: Ranges of characters can be defined using unicode notation (`U+XXXX-U+XXXX` ) in set rules. This is a possible feature that can be added in the future is there is enough demand for it.

docs/syntax/Cadence.syntax
jsproz: I can add some comments with EBNF equivalents where it is practical.

docs/syntax/Cadence.syntax
jsproz: Comments were added.

docs/syntax/Cadence.syntax
jsproz: Added a comment.

docs/syntax/Cadence.syntax
jsproz: Done.

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+destroyFunctionDeclaration:
+        access *1viewKeyword destroyKeyword parameterList *1functionBlock
+    ;
+
+prepareFunctionDeclaration:
+        access *1viewKeyword prepareKeyword parameterList *1functionBlock
+    ;
+
+functionDeclaration:
+        access *1viewKeyword funKeyword nonReservedIdentifier parameterList *1functionDeclarationTypeAnnotation *1functionBlock
```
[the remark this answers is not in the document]
jsproz: I think this can be handled downstream in the semantic checks. Those rules can be more explicit about when the functionBlock is required.

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+        "(" *parameter( "," ) *1comma ")"
+    ;
+
+parameter:
+        parseLabeledParameter
+    |   parseUnlabeledParameter
+    ;
+
+parseLabeledParameter:
+        parameterPublicName parameterLocalName ":" typeAnnotation *1defaultValue
```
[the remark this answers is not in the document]
jsproz: Done.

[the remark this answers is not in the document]
jsproz: Done.

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+        1*fullType( "|" )
+    ;
+
+accessAccessValueEntitlementMapping:
+        mappingKeyword fullType
+    ;
+
+conformances:
+        conformancesTypes
+    |   conformancesEmpty
```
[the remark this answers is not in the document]
jsproz: Done.

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+contractCompositeDeclaration:
+        access contractKeyword nonReservedIdentifier conformances "{" membersAndNestedDeclarations "}"
+    ;
+
+enumCompositeDeclaration:
+        access enumKeyword nonReservedIdentifier conformances "{" *semicolonEnumerationDeclarationCase "}"
+    ;
+
+structInterfaceDeclaration:
+        access structKeyword interfaceKeyword nonReservedIdentifier *1interfaceInheritance "{" membersAndNestedDeclarations "}"
```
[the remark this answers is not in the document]
jsproz: Done.

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+unspecifiedField:
+        access nonReservedIdentifier ":" typeAnnotation
+    ;
+
+fields:
+        *fieldsEntry
+    ;
+
+fieldsEntry:
+        field *1semicolon
```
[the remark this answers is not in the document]
jsproz: Done.

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+entitlementMappingTarget:
+        fullType
+    ;
+
+entitlementIncludeType:
+        includeKeyword nominalType
+    ;
+
+semicolonEnumerationDeclarationCase:
+        semicolon
```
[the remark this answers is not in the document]
jsproz: `declarationSeparator` is essentially an optional semicolon. `semicolon` is a required semicolon. They are not equivalent. There are subtle differences for each of these examples, and the rule was written to match the go parser. In some cases a series of semicolons can be parsed. In other cases an item in a list can be optionally followed by a semicolon. In still other cases there is a separated list that is optionally followed by a trailing semicolon.

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+        entitlementMappingSource "->" entitlementMappingTarget
+    ;
+
+entitlementMappingSource:
+        fullType
+    ;
+
+entitlementMappingTarget:
+        fullType
+    ;
```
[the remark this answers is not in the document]
jsproz: Done.

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+        1*fullType( "," )
+    ;
+
+accessAccessValueEntitlementDisjunction:
+        1*fullType( "|" )
+    ;
+
+accessAccessValueEntitlementMapping:
+        mappingKeyword fullType
+    ;
```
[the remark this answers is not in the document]
jsproz: Done.

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+    |   entitlementDeclaration
+    |   entitlementMappingDeclaration
+    |   attachmentDeclaration
+    ;
+
+initFunctionDeclaration:
+        access *1viewKeyword initKeyword parameterList *1functionBlock
+    ;
+
+destroyFunctionDeclaration:
```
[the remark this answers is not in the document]
jsproz: Added comment.

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+quotedTextItemCharacter:
+        ~nonquotedTextItemCharacter
+    ;
+
+nonquotedTextItemCharacter:
+        "\"\r\n\\" ;
+
+escapedCharacter:
+        escapedUnicodeScalar
+    |   escapedCCharacter
```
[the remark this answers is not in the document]
jsproz: Escaped C character -> `escapedCCharacter`

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1482 @@
...
+nominalTypeTemplate:
+        "<" *1typeAnnotationList ">"
+    ;
+
+typeAnnotationList:
+        1*typeAnnotation( "," )
+    ;
+
+functionType:
+        funKeyword "(" *1functionTypeParameters ")" *1functionTypeReturn
```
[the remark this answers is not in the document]
jsproz: Added optional `view`.

docs/syntax/Cadence.syntax
jsproz: Done.

docs/syntax/Cadence.syntax
jsproz: Done.

docs/syntax/Cadence.syntax
jsproz: Renamed.

docs/syntax/Cadence.syntax
jsproz: Removed precedence level in comment.

jsproz: The`leftDenotation` was changed to `postfixOperation`. These are parsed into an array and are evaluated left to right. They are not in a tree structure so precedence has no use in this case.

docs/syntax/Cadence.syntax
jsproz: See previous note.

docs/syntax/Cadence.syntax
jsproz: Making this change causes some of the contracts to fail because they still use `T{...}`.

docs/syntax/Cadence.syntax
jsproz: `typeLeftBindingPowerIntersection` was never used.

docs/syntax/Syntax Notation.md
turbolent: Yeah, I get how the parser generator has this feature, but for tokens it isn't really semantically different, no?         

Could this just be:

```
whitespaceCharacter:
    U+0000
  | U+0009
  | U+000B-U+000C
  | U+0020
  ;
```

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1483 @@
...
+
+whitespaceItem:
+        singleLineComment
+    |   multipleLineComment
+    |   whitespaceCharacter
+    |   lineBreak
+    ;
+
+whitespaceCharacter:
+        U+0000,U+0009,U+000B-U+000C,U+0020
```
turbolent, line 21: U+0000 is not valid whitespace, it should be removed:

```suggestion
        U+0009,U+000B-U+000C,U+0020
```

docs/syntax/Cadence.syntax
turbolent: Maybe handle it the same as above: https://github.com/onflow/cadence/pull/3707#discussion_r1873819904

jsproz: Added a comment.

docs/syntax/Cadence.syntax
jsproz: Removed.

docs/syntax/Syntax Notation.md
jsproz: The tool doesn't support escape sequences for vertical tab and form feed. The unicode expression is needed in this case, and the tool only supports one unicode expression per rule. It cannot be used in alternatives like in your example.
I added a comment for clarity.

docs/syntax/Cadence.syntax
jsproz: I believe the current syntax handles the precedence appropriately. We can test this and make changes in the future as we refine it.

docs/syntax/Cadence.syntax
turbolent: Thanks for adding comments for `pub`. Can you please also add comments for  `pub(set), and `priv`?

docs/syntax/Cadence.syntax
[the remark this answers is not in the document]
turbolent: I don't quite follow. Why are those escape sequences called "C character"?

docs/syntax/Cadence.syntax
turbolent: Though the rule parses all types, e.g. `&T?`, it leaves the meaning of it ambiguous (`(&T)?`). Can you please leave a TODO to split this rule up and to encode the precedence rules, just like we do it for expressions?

turbolent (approved):
Looks good! Only a couple comments to resolve and then we can merge what we have so far. 

@SupunS Could you please take a final look?

docs/syntax/Cadence.syntax
jsproz: Done.

docs/syntax/Cadence.syntax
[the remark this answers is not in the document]
jsproz: Second C removed.

docs/syntax/Cadence.syntax
jsproz: Done.

docs/syntax/Cadence.syntax
```diff
@@ -0,0 +1,1485 @@
...
+    |   inKeyword
+    |   removeKeyword
+    |   selfKeyword
+    |   setKeyword
+    |   toKeyword
+    |   typeKeyword
+    |   viewKeyword
+    ;
+
+accessKeyword:
```
SupunS, line 1283: Aren't these keywords also tokens? i.e: shouldn't be `<accessKeyword>`?

SupunS: @jsproz in case if ^this got buried under other threads

PR #3713, opened 2024-12-11
[Compiler+VM PoC] Refactor bytecode encoding

turbolent:
## Description

Move the bytecode encoding out of the compiler, for now into the `opcode` package (we might want to rename it? or add a new package?).

Also, abstract the instruction emission from the compiler behind an interface. To begin with, only the bytecode emission is supported. Eventually we can also implement the interface to produce in-memory Go structs, because in a first milestone of the compiled execution environment, it will likely be unnecessary to encode to bytecode, because it is not stored. I named this interface `CodeGen` for a lack of better word, maybe "assembler" is a more fitting name? 

In the next PR I will move the bytecode decoding out of the VM and move it next to the encoding.

Finally, we can then look into generating bytecode encoding and decoding functions from a single source of truth.

______

### DIFF
--- a/bbq/compiler/codegen.go
+++ b/bbq/compiler/codegen.go
@@ -31,21 +31,21 @@ type CodeGen interface {
 	EmitFalse()
 	EmitDup()
 	EmitDrop()
-	EmitGetConstant(index uint16)
+	EmitGetConstant(constantIndex uint16)
 	EmitJump(target uint16) int
 	EmitJumpIfFalse(target uint16) int
 	PatchJump(offset int, newTarget uint16)
 	EmitReturnValue()
 	EmitReturn()
-	EmitGetLocal(index uint16)
-	EmitSetLocal(index uint16)
-	EmitGetGlobal(index uint16)
-	EmitSetGlobal(index uint16)
+	EmitGetLocal(localIndex uint16)
+	EmitSetLocal(localIndex uint16)
+	EmitGetGlobal(globalIndex uint16)
+	EmitSetGlobal(globalIndex uint16)
 	EmitGetField()
 	EmitSetField()
 	EmitGetIndex()
 	EmitSetIndex()
-	EmitNewArray(index uint16, size uint16, isResource bool)
+	EmitNewArray(typeIndex uint16, size uint16, isResource bool)
 	EmitIntAdd()
 	EmitIntSubtract()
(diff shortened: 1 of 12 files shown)


bbq/compiler/codegen.go
```diff
@@ -0,0 +1,242 @@
...
+	EmitReturn()
+	EmitGetLocal(index uint16)
+	EmitSetLocal(index uint16)
+	EmitGetGlobal(index uint16)
+	EmitSetGlobal(index uint16)
+	EmitGetField()
+	EmitSetField()
+	EmitGetIndex()
+	EmitSetIndex()
+	EmitNewArray(index uint16, size uint16, isResource bool)
```
SupunS, line 48: Just to be more explicit on what does this index refers to (since other instructions also index to local-vars, constant pool, etc.)
```suggestion
	EmitNewArray(typeIndex uint16, size uint16, isResource bool)
```

bbq/compiler/codegen.go
```diff
@@ -0,0 +1,242 @@
...
+	EmitIntDivide()
+	EmitIntMod()
+	EmitEqual()
+	EmitNotEqual()
+	EmitIntLess()
+	EmitIntLessOrEqual()
+	EmitIntGreater()
+	EmitIntGreaterOrEqual()
+	EmitUnwrap()
+	EmitCast(index uint16, kind opcode.CastKind)
```
SupunS, line 61: Same here, and some more instructions below, like `EmitTransfer`, `EmitNewRef`, `EmitNew`
```suggestion
	EmitCast(typeIndex uint16, kind opcode.CastKind)
```

SupunS (approved):
Nice! Thanks for refactoring!

bbq/compiler/codegen.go
turbolent: Good point, improved in [d02289b](https://github.com/onflow/cadence/pull/3713/commits/d02289b09a0f92570eabead6c3751b3b1cfa05ef)

bbq/compiler/codegen.go
turbolent: Improved here and in the other cases in [d02289b](https://github.com/onflow/cadence/pull/3713/commits/d02289b09a0f92570eabead6c3751b3b1cfa05ef)

PR #3716, opened 2024-12-14
[Compiler+VM POC] Generate the code for the instructions

turbolent:
## Description

Based on @jsproz's idea, generate the Go code for the instructions (type, encoding, decoding, etc.) from a single specification file. This is similar to what we had done in the old compiler PoC (https://github.com/onflow/cadence/blob/f5e27df4e8802f6881ca28125b87dbdf12a2c1e8/compiler/wasm/gen/main.go), but uses https://github.com/dave/dst to generate pretty Go code.

In the future we can maybe extend the instruction specification file with the effects of the instructions.
Maybe to start with at least stack operations (how many values, of what type, are popped from the stack; and what values, of what type, are pushed to the stack)

______


SupunS (approved):
👏

PR #3717, opened 2024-12-16
Import aliasing

### LINKED ISSUE #1171: Support referencing duplicate contracts (deployed to different accounts) in cadence context
### Issue To Be Solved

Here's an example: https://play.onflow.org/2d6bf889-ac32-4de4-9a79-a25ddf1d5236?type=tx&id=73c0651f-3aba-4150-81ea-5ccaa4b20c61

The same (duplicate code) contract deployed to 0x02 and 0x03, and in the transaction, I'm trying to reference the two duplicated contracts:  
image
However, it gives an error saying "cannot redeclare contract: FUSD is already declared".

I think this restriction here doesn't make any sense. Yes the contract code is duplicated, but they're deployed to different accounts so can carry with different states / storage spaces. 

We should allow referencing these contracts in the cadence environment, either through import .. as ... alias, or other developer-friendly ways. (See suggested solution)

### Suggested Solution

1. Import .. as .. alias:
```js
import FUSD as FUSD1 from 0x02
import FUSD as FUSD2 from 0x03

transaction {
  prepare(acct: AuthAccount) {
    let typeFUSD = Type<FUSD1>()
    let typeKibble = Type<FUSD2>()
    log(typeFUSD == typeKibble)
    log(typeFUSD.identifier)
    log(typeKibble.identifier)
  }
}
```
2. Maybe other developer-friendly way? 

### Additional Comments:
* Why I think this feature is necessary? 
Let's say here's a scenario like: since each uniswap trading pool shares the same code template from UniswapPair.cdc and deployed as UniswapPair contract. Now say we have 2 pools: FlowUSDT pool and FUSDUSDT pool. Due to the above mentioned error, current developers / arbitrageurs cannot reference these 2 pools in the same contract / transaction code. 
(I know Blocto workarounds this by naming the two pools with different names - but this is silly! Duplicating code adds technical debt and maintainance efforts with no benefit -  it also lost permissionless pool deployment and composability then.)

I think this feature will be good for any pool-based defi projects build on flow, and will be beneficial to flow ecosystem composibility as a whole.

* Any comments against this feature is also welcome, let's debate on the pros and cons then :).

RZhang05:
Closes #1171 

## Description

Introduces import aliases.

______

### DIFF
--- a/ast/import_test.go
+++ b/ast/import_test.go
@@ -40,6 +40,9 @@ func TestImportDeclaration_MarshalJSON(t *testing.T) {
 				Pos:        Position{Offset: 1, Line: 2, Column: 3},
 			},
 		},
+		Aliases: map[string]string{
+			"foo": "bar",
+		},
 		Location:    common.StringLocation("test"),
 		LocationPos: Position{Offset: 4, Line: 5, Column: 6},
 		Range: Range{
@@ -63,7 +66,9 @@ func TestImportDeclaration_MarshalJSON(t *testing.T) {
                     "EndPos": {"Offset": 3, "Line": 2, "Column": 5}
                 }
             ],
-			"Aliases": null,
+			"Aliases": {
+				"foo": "bar"
+			},
             "Location": {
                 "Type": "StringLocation",
                 "String": "test"
--- a/ast/import_test.go
+++ b/ast/import_test.go
@@ -63,6 +63,7 @@ func TestImportDeclaration_MarshalJSON(t *testing.T) {
                     "EndPos": {"Offset": 3, "Line": 2, "Column": 5}
                 }
             ],
+			"Aliases": null,
             "Location": {
                 "Type": "StringLocation",
                 "String": "test"
--- a/interpreter/import_test.go
+++ b/interpreter/import_test.go
@@ -446,7 +446,7 @@ func TestInterpretImportWithAlias(t *testing.T) {
 		  import a as a2 from 0x2
 
           access(all) fun test(): Int {
-              return a() + a()
+              return a1() + a2()
           }
         `,
 		ParseAndCheckOptions{
--- a/interpreter/import_test.go
+++ b/interpreter/import_test.go
@@ -542,3 +542,95 @@ func TestInterpretImportWithAlias(t *testing.T) {
 		value,
 	)
 }
+
+func TestInterpretImportAliasGetType(t *testing.T) {
+
+	t.Parallel()
+
+	address := common.MustBytesToAddress([]byte{0x1})
+
+	importedChecker, err := ParseAndCheckWithOptions(t,
+		`
+		access(all) struct Foo {
+		}
+		`,
+		ParseAndCheckOptions{
+			Location: common.AddressLocation{
+				Address: address,
+				Name:    "",
+			},
+		},
+	)
+	require.NoError(t, err)
+
+	importingChecker, err := ParseAndCheckWithOptions(t,
+		`
+		import Foo as Bar from 0x1
+
+		access(all) fun test(): String {
+			var bar: Bar = Bar()
+			return bar.getType().identifier
+		}
+        `,
+		ParseAndCheckOptions{
+			Config: &sema.Config{
+				LocationHandler: func(identifiers []ast.Identifier, location common.Location) (result []sema.ResolvedLocation, err error) {
+
+					for _, identifier := range identifiers {
+						result = append(result, sema.ResolvedLocation{
+							Location: common.AddressLocation{
+								Address: location.(common.AddressLocation).Address,
+								Name:    identifier.Identifier,
+							},
+							Identifiers: []ast.Identifier{
+								identifier,
+							},
+						})
+					}
+					return
+				},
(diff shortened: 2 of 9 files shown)


ast/import_test.go
```diff
@@ -63,6 +63,7 @@ func TestImportDeclaration_MarshalJSON(t *testing.T) {
                     "EndPos": {"Offset": 3, "Line": 2, "Column": 5}
                 }
             ],
+			"Aliases": null,
```
turbolent, line 66: Could you please test the case where the aliases are set? You can do that by adding some dummy/arbitrary aliases in the test declaration above (https://github.com/onflow/cadence/pull/3717/files#diff-5133e9d8f2ff8054c388461bfe306367de04a0ed58979bdcc84e3cd0463dd910R36) and asserting the serialization here

interpreter/import_test.go
```diff
@@ -401,3 +401,144 @@ func TestInterpretResourceConstructionThroughIndirectImport(t *testing.T) {
 		resourceConstructionError.CompositeType,
 	)
 }
+
+// TestInterpretImportWithAlias shows importing two funs of the same name from different addresses
```
turbolent, line 405: 👌

interpreter/interpreter_import.go
```diff
@@ -32,13 +32,13 @@ func (interpreter *Interpreter) VisitImportDeclaration(declaration *ast.ImportDe
...
 	for _, resolvedLocation := range resolvedLocations {
-		interpreter.importResolvedLocation(resolvedLocation)
+		interpreter.importResolvedLocation(resolvedLocation, &declaration.Aliases)
 	}
 
 	return nil
 }
 
-func (interpreter *Interpreter) importResolvedLocation(resolvedLocation sema.ResolvedLocation) {
+func (interpreter *Interpreter) importResolvedLocation(resolvedLocation sema.ResolvedLocation, aliases *map[string]string) {
```
turbolent, line 41: Go `map`s already have reference semantics (a value with a pointer), so there is no need to pass it by pointer. Maybe also chop down long parameter lists over multiple lines:

```suggestion
func (interpreter *Interpreter) importResolvedLocation(
	resolvedLocation sema.ResolvedLocation,
	aliases map[string]string,
) {
```

(also need to adjust the call-site above, and the dereference below)

parser/declaration.go
```diff
@@ -645,13 +645,14 @@ func parsePragmaDeclaration(p *parser) (*ast.PragmaDeclaration, error) {
 //
 //	importDeclaration :
 //	    'import'
-//	    ( identifier (',' identifier)* 'from' )?
+//	    ( identifier (as identifier)? (',' identifier (as identifier)?)* 'from' )?
```
turbolent, line 648: The `as` keyword is a token, so it needs to be quoted:

```suggestion
//	    ( identifier ('as' identifier)? (',' identifier ('as' identifier)?)* 'from' )?
```

sema/check_import_declaration.go
```diff
@@ -113,7 +113,7 @@ func (checker *Checker) resolveLocation(identifiers []ast.Identifier, location c
 	return locationHandler(identifiers, location)
 }
 
-func (checker *Checker) importResolvedLocation(resolvedLocation ResolvedLocation, locationRange ast.Range) {
+func (checker *Checker) importResolvedLocation(resolvedLocation ResolvedLocation, locationRange ast.Range, aliases *map[string]string) {
```
turbolent, line 116: Same as above, the map can be passed as-is, no need to pass it by pointer

sema/check_import_declaration.go
```diff
@@ -302,6 +304,7 @@ func (checker *Checker) importElements(
 	valueActivations *VariableActivations,
 	requestedIdentifiers []ast.Identifier,
 	availableElements *StringImportElementOrderedMap,
+	aliases *map[string]string,
```
turbolent, line 307: Same as above, the map can be passed as-is, no need to pass it by pointer

sema/import_test.go
```diff
@@ -824,3 +824,237 @@ func TestCheckImportContract(t *testing.T) {
...
+			},
+		)
+
+		errs := RequireCheckerErrors(t, err, 1)
+
+		notDeclaredError := &sema.NotDeclaredError{}
+		assert.ErrorAs(t, errs[0], &notDeclaredError)
+
+	})
+
```
turbolent, line 1059: Can you please add a test checking that `getType().identifier` still returns the original type name, and does not use the alias?

turbolent (commented):
Great work!

Just a few minor suggestions, but overall looks great! Good to see the main change is simple 👌

ast/import_test.go
RZhang05: Replaced the null case with the case where the aliases are set.

sema/import_test.go
RZhang05: Added here https://github.com/onflow/cadence/pull/3717/files#diff-501230d8bbe927ccf68df43c9ce8dbb3cc6cde1355abb293bd87bbfbbf3fbee6R546.

turbolent (approved):
Nice!

PR #3736, opened 2025-01-24
Update to Go 1.23

turbolent:
## Description

Just like [in flow-go](https://github.com/onflow/flow-go/pull/6922), update to Go 1.23.

Go 1.24 is releasing next month, [which will make 1.22 unsupported.](https://endoflife.date/go)

______


SupunS (approved):

turbolent:
@SupunS Had to fix new lint errors, see [b81e414](https://github.com/onflow/cadence/pull/3736/commits/b81e4145a6cd34621b598c717b520a7658e5f578)

sema/type.go
```diff
@@ -1615,324 +1615,293 @@ func (*FixedPointNumericType) CheckInstantiated(_ ast.HasPosition, _ common.Memo
 
 // Numeric types
 
-var (
```
turbolent, line 1618: gofmt is acting up here when using method call chains in a `var` block: it applies the correct indentation for the wrapped method calls, then undoes the indentation for whatever reason.

Changing the large `var` block to individual var declarations (automatically refactored using Goland) makes gofmt produce the correct indentations and makes the linter happy

parser/declaration.go
```diff
@@ -910,7 +910,10 @@ func parseHexadecimalLocation(p *parser) common.AddressLocation {
 	address, err := common.BytesToAddress(rawAddress)
 	if err != nil {
 		// Any returned error is a syntax error. e.g: Address too large error.
-		p.reportSyntaxError(err.Error())
```
SupunS, line 913: Do you know what was the problem with the old way?

SupunS (approved):

parser/declaration.go
turbolent: Looks like the linting got smarter: `reportSyntaxError()` has a "printf"-style signature `message string, params ...any`, and in turn calls `syntaxError()`, which has a similar signature, and that in turn calls `NewSyntaxError`, which again has a similar signature, and eventually calls `fmt.Sprintf(message, params...)`.

So `err.Error()` is used here as a format string, but really should be used as-is as the message.

SupunS: hah, a bit too smart! 😆

PR #3741, opened 2025-01-28
[Compiler] Support accessing `result` variable inside post conditions

SupunS:
Work towards #3742

## Description

`result` variable gets passed as a parameter (last parameter) to the generated functions (for conditions).
e.g:

```cadence
struct interface A {
    access(all) view fun $A.test.postConditions(_ a1: T1, _ a2: T2, _ result: T3): Void {
        ...
    }
}
```

______


bbq/compiler/compiler_test.go
```diff
@@ -33,6 +33,8 @@ func TestCompileRecursionFib(t *testing.T) {
 
 	t.Parallel()
 
+	t.SkipNow()
```
SupunS, line 36: Skipping these for now, since the generated code's structure is still changing (e.g: code generation for returns, etc.), and having to change these tests each and every time is not worth the hassle. We can enable these at the end, once the code generation become somewhat finalized.

ast/variable_declaration.go
```diff
@@ -151,7 +151,12 @@ func (d *VariableDeclaration) Doc() prettier.Doc {
 		)
 	}
 
-	valueDoc := d.Value.Doc()
+	var valueDoc prettier.Doc
```
SupunS, line 154: This is probably temporary. Needed this to make it easy to debug the generated code for now.
See the comment below (in `desugar.go`) as to why the "value" can be `nil` here.

bbq/compiler/desugar.go
```diff
@@ -247,58 +202,286 @@ func (d *Desugar) VisitFunctionDeclaration(declaration *ast.FunctionDeclaration)
...
+		ast.NewIdentifier(
+			d.memoryGauge,
+			tempResultVariableName,
+			pos,
+		),
+		declaration.ReturnTypeAnnotation,
+
+		// TODO: What should be the value?
+		//  We just need the variable to be defined. Value is assigned later.
+		nil,
```
SupunS, line 295: Here we only declare the temp variable `var $_result: T`. Then each return statement (eventually) will assign the return value to this `$_result` variable. So this is an uninitialized variable at this point. Maybe is there a better way to handle this?

bbq/compiler/desugar.go
```diff
@@ -382,6 +587,45 @@ func (d *Desugar) generateConditionsFunction(
...
+				pos,
+			),
+
+			ast.NewTypeAnnotation(
+				d.memoryGauge,
+				false,
+
+				// TODO: Pass the proper type here. It can be looked-up from the elaboration (see above).
+				//  However, will have to convert the `sema.Type` to `ast.Type`.
+				//  Currently this `ast.Type` is not needed for the compiler, hence passing a dummy type.
```
SupunS, line 611: Here the AST type is not really used by the compiler, so this is probably fine for now. But eventually we would need to pass the proper type, just for consistency.

bluesign:
doesn't also locals need to be passed?

ast/variable_declaration.go
turbolent: Maybe leave a comment, so we do not revert this at some point

bbq/compiler/compiler_test.go
turbolent: I guess that's fine, as long as we keep the "integration tests" in the VM, i.e. testing executing the compiled code still works.

sema/check_function.go
```diff
@@ -455,6 +456,8 @@ func (checker *Checker) visitWithPostConditions(
 				resultType = returnType
 			}
 
+			checker.Elaboration.SetResultVariableType(enclosingElement, resultType)
```
turbolent, line 459: Can we just use the return type of the function type instead?

turbolent (approved):
Nice!

sema/check_function.go
SupunS: Unfortunately no, for resources, the result variable's type becomes a reference.
We could theoretically re-calculate the type at the compiler (i.e: if it's a resource, then use the reference type), but it would be less error-prone to do the calculation once (which is already done at the checker), and then cache and re-use it

turbolent: Ah, I see! In that case it makes sense to store it 👍

ast/variable_declaration.go
SupunS: Added https://github.com/onflow/cadence/pull/3741/commits/9e2d8404367c1cfa12e2e511d2e03ccd263a19c5

bbq/compiler/compiler_test.go
SupunS: Created an issue https://github.com/onflow/cadence/issues/3758 so we don't forget.

PR #3745, opened 2025-01-29
[Compiler] Improve VM stack operations

turbolent:
## Description

The VM has some helper functions for the operand stack, e.g. `pop2` and `pop3`, which pop 2 and 3 values from the stack respectively. These functions currently return the values in reverse order, which is a bit confusing and error prone. For example, the implementation of the `SetIndex` instruction in `vm.opSetIndex` gets the values in reverse order.

Change the order of the return valuesto reflect the order on the stack.

Also, add tests and add a `peekN` helper, as well as tests.

______

### DIFF
--- a/bbq/vm/vm.go
+++ b/bbq/vm/vm.go
@@ -103,24 +103,32 @@ func (vm *VM) pop() Value {
 	return value
 }
 
-// pop2 removes and returns the top two value of the stack.
-// It is efficient than calling `pop` twice.
-func (vm *VM) pop2() (Value, Value) {
+// pop2NonResource removes and returns the top two non-resource values from the stack:
+// N-2, and N-1, where N is the number of elements on the stack.
+// It is efficient than calling `pop` twice, in the case of non-resource values.
+func (vm *VM) pop2NonResource() (Value, Value) {
 	lastIndex := len(vm.stack) - 1
-	value1, value2 := vm.stack[lastIndex], vm.stack[lastIndex-1]
-	vm.stack[lastIndex], vm.stack[lastIndex-1] = nil, nil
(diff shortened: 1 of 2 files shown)


bbq/vm/vm.go
```diff
@@ -465,9 +471,11 @@ func opInvokeDynamic(vm *VM, ins opcode.InstructionInvokeDynamic) {
 	var functionValue = vm.lookupFunction(compositeType.Location, qualifiedFuncName)
 
 	parameterCount := int(functionValue.Function.ParameterCount)
-	arguments := vm.stack[stackHeight-parameterCount:]
+	arguments := vm.peekN(parameterCount)
 	vm.pushCallFrame(functionValue, arguments)
-	vm.dropN(parameterCount)
+	vm.dropN(len(arguments))
+
+	// TODO: drop the receiver?
```
SupunS, line 478: `parameterCount` already includes the receiver. see: https://github.com/onflow/cadence/blob/6ce5b49adbfc6c730589286b8f512e68d8642f39/bbq/compiler/compiler.go#L1208-L1210

So we don't have to explicitly drop it.

On a related note: All tests do this sanity check on the stack size, at the end of the execution, just to make sure no garbage values are left on stack: https://github.com/onflow/cadence/blob/6ce5b49adbfc6c730589286b8f512e68d8642f39/bbq/vm/test/utils.go#L631-L634

SupunS (approved):

bbq/vm/vm.go
turbolent: Ah, nice! 👌  Thanks for clarifying!

```suggestion
	// We do not need to drop the receiver, as the parameter count given in the instruction already includes it
```

Issue #3742, opened 2025-01-29
Compiler Milestone 2: More realistic FT subset


SupunS:
The goal of this milestone is to support most of the language features needed for the FT transfer transaction and contracts.
 
```[tasklist]
### FT (everything expect closures)
- [ ]  Default functions - @Supun
- [ ]  Pre-/post-conditions (difficult, complex, e.g. linking/chaining, default functions, linearization, etc.) - @Supun
- [ ]  Casting (force-casting `as!`; Optional casting `as?` doesn’t seem necessary, but maybe also implement while at it)
- [ ]  Fixed-point literal (compile to constant pool + load)
- [ ]  Event emission (simple, try to reuse infrastructure (e.g. export) of runtime)
- [ ]  If-let (for `FlowToken`)
- [ ]  Switch statement (for `FlowToken` metadata standard impl.)
- [ ]  Dictionary literal (compile + add VM instruction; replicate what we have for arrays)
- [ ]  Some of the comparison operators (<, ≤, etc.)
- [ ]  Fixed-point numbers (`UFix64` for balance of Vault; wait until interpreter refactor)
```
- Not needed:
        - Entitlements: purely static
        - Swap
        - Assignment with second value

While at it, also add some of the fundamental language features:
- [ ] Control flow:
    - if statement with optional (if-let)
    - switch statement (just compiler)
    - for-loop statement (more effort, given iteration over different iterators, iteration over references)
- [ ] Double check we are erroring for unimplemented functionality in compiler

turbolent:
@SupunS We might also want to add scoping to the list, as there are a few TODOs in the compiler for that

turbolent:
@jsproz Can you please add your name to the item(s) you're working on? That way we can avoid accidental duplicate work. Thank you!

SupunS:
> We might also want to add scoping to the list, as there are a few TODOs in the compiler for that

I wasn't sure if we actually need any scoping at the compiler level, other than functions. Didn't run into any case where we would need such. So left those alone for now. Maybe we can create an issue, but might not actually need to do anything there, and all it need could be to clean-up the TODO s.

bluesign:
one item I saw (I know you didnt implement yet scoping but as a note ) function argument shadowing and post conditions I saw recently when playing with your PR.

turbolent:
@SupunS Not sure, but from what I can see, it looks like all comparison operators (<, ≤, etc.) are already implemented in the compiler and the VM. Was there anything left to be done there?

[...]

PR #3763, opened 2025-02-07
[Compiler] Support compilation for before statements

SupunS:

## Description

Before functions/statements are treated similar to pre/post conditions.

### For concrete type functions
Before-statements are inlined and added to the beginning of the function.

```cadence
struct Test {
    var i: Int

    init() {
        self.i = 2
    }

    fun test() {
        post {
            print(before(self.i).toString())
        }
        self.i = 8
    }
}
```

becomes:
```cadence
struct Test {
    var i: Int
    
    init() {
        self.i = 2
    }
    
    fun test() {
        var �exp0 = self.i    // Inlined before-statement
        self.i = 8
        if !print(�exp0.toString()) {    // Post condition refers to the variable declared above
            panic("pre/post condition failed")
        }
        return
    }
}
```
### For interface methods
- Before-statements are extracted out as a separate function, and are invoked explicitly by the implementation.
- Post condition accept the before-statement results are parameters.
```cadence
struct interface Foo {
    var i: Int

    fun test() {
        post {
            print(before(self.i).toString())
        }
    }
}

struct interface Bar: Foo {
    var i: Int

    fun test() {
        post {
            print(before(self.i + 3).toString())
        }
    }
}


struct Test: Bar {
    var i: Int

    init() {
        self.i = 2
    }

    fun test() {
        self.i = 8
    }
}
```

Would eventually become something like:
```cadence
struct interface Foo {
    var i: Int
    
    access(all)
    view fun $Foo.test.�exp0(): AnyStruct {    // Before statements get extracted out to a separate function
        return self.i
    }
    
    access(all)
    view fun $Foo.test.postConditions(_ �exp0: AnyStruct): Void {  // Post condition accept the before-statement results as parameters
        if !print(�exp0.toString()) {
            panic("pre/post condition failed")
        }
        return
    }
}

struct interface Bar: Foo {
    var i: Int
    
    access(all)
    view fun $Bar.test.�exp1(): AnyStruct {
        return self.i + 3
    }
    
    access(all)
    view fun $Bar.test.postConditions(_ �exp1: AnyStruct): Void {
        if !print(�exp1.toString()) {
            panic("pre/post condition failed")
        }
        return
    }
}

struct Test: Bar {
    var i: Int
    
    init() {
        self.i = 2
    }
    
    fun test() {
        var $_before0 = self.$Foo.test.�exp0()  // Implementation would call the before-statement-functions at the start of the implemented function
        var $_before1 = self.$Bar.test.�exp1()
        self.i = 8
        self.$Foo.test.postConditions($_before0)  // And the results are passed as argments
        self.$Bar.test.postConditions($_before1)
        return
    }
}
```

______


turbolent:
Great work! I haven't looked too deep into the code yet, but how does this handle the case where the post-conditions in the interfaces refer to different expressions, e.g.

```cadence
struct interface Foo {
    var i: Int

    fun test() {
        post {
            before(self.i) == 2
        }
    }
}

struct interface Bar: Foo {
    var j: Int

    fun test() {
        post {
            before(self.j) == 3
        }
    }
}


struct Test: Bar {
    var i: Int
    var j: Int

    init() {
        self.i = 2
        self.j = 3
    }

    fun test() {
        self.i = 4
        self.j = 5
    }
}
```

In the the example above the two interfaces referred to the same expression (`self.i`).

I guess there's no potential for confusion, as the IDs are distinct (`$Foo.test.�exp�0` vs `$Bar.test.�exp�1`), and it's also ensured that the separate compilation ensures confusion even if e.g. `Foo` gets additional `before` expressions (I assume `$Foo.test.�exp�1`, etc.?) How are the IDs generated?

SupunS:
> but how does this handle the case where the post-conditions in the interfaces refer to different expressions, e.g.

It's pretty much the same. Regardless of the type of the expression, whatever the expression that is inside `before(xyz)` will become a function of the same interface. 
i.e: Turn it into a function which returns `xyz`
```cadence
struct interface Foo {
    var i: Int
    
    access(all)
    view fun $Foo.test.�exp�0(): AnyStruct {
        return self.i
    }
    ...
}

struct interface Bar: Foo {
    var j: Int
    
    access(all)
    view fun $Bar.test.exp1(): AnyStruct {
        return self.j
    }
    ...
}
```

And then in the post condition, `before(xyz)` is replaced by the passed-in argument/parameter.

> I guess there's no potential for confusion, as the IDs are distinct ($Foo.test.�exp�0 vs $Bar.test.�exp�1), and it's also ensured that the separate compilation ensures confusion even if e.g. Foo gets additional before expressions (I assume $Foo.test.�exp�1, etc.?) How are the IDs generated?

Yes, the IDs are distinct (the counter is per-type). Before-extractor in type-checker generate these counters (`�exp�1`), the compiler only make it type/method-qualified (appends the type/method name info). Additionally, the type info (`MemberAccessInfo`) of these invocations (say `self.$Foo.test.�exp�0()`) points to the correct interface type (`Foo`) and that is what the compiler use. So the type name prefix in the generated ID is more like a second-level insurance (and to improve debug-ability), but not really a necessity.

The generated complete code for your example is:

```cadence
struct interface Foo {
    var i: Int
    
    access(all)
    view fun $Foo.test.exp0(): AnyStruct {
        return self.i
    }
    
    access(all)
    view fun $Foo.test.postConditions(_ exp0: AnyStruct): Void {
        if !(exp0 == 2) {
            panic("pre/post condition failed")
        }
        return
    }
}

struct interface Bar: Foo {
    var j: Int
    
    access(all)
    view fun $Bar.test.exp1(): AnyStruct {
        return self.j
    }
    
    access(all)
    view fun $Bar.test.postConditions(_ exp1: AnyStruct): Void {
        if !(exp1 == 3) {
            panic("pre/post condition failed")
        }
        return
    }
}

struct Test: Bar {
    var i: Int
    
    var j: Int
    
    init() {
        self.i = 2
        self.j = 3
    }
    
    fun test() {
        var $_before0 = self.$Foo.test.exp0()
        var $_before1 = self.$Bar.test.exp1()
        var $_result =  
        self.i = 4
        self.j = 5
        self.$Foo.test.postConditions($_before0)
        self.$Bar.test.postConditions($_before1)
        return
    }
}
```

SupunS:
Happy to walk-through the idea in detail in the next implementation sync

jsproz:
It seems like the pre-conditions and the post-conditions should be called in opposite orders.

```cadence
    fun test() {
        var $_before0 = self.$Foo.test.�exp�0()
        var $_before1 = self.$Bar.test.�exp�1()
        var $_result =  
        self.i = 4
        self.j = 5
        self.$Bar.test.postConditions($_before1)
        self.$Foo.test.postConditions($_before0)
        return
    }
```

And the order in which they are called should be well documented. I believe that C++ and Swift use different orders for constructors and destructors. Should the pre-conditions of the grandparent be called before or after the parent's pre-conditions?

jsproz:
Should the grandparents post-condition even be called if the parents condition fails?

SupunS:
@jsproz conditions are linearized in a "depth-first pre-ordered" manner. It is already documented under the interface section: https://cadence-lang.org/docs/language/interfaces#linearizing-conditions

> Should the grandparent's post-condition even be called if the parent's condition fails?

Given a failed condition would panic the runtime, it will terminate the execution, and would not call any conditions to follow (grandparent's post-condition in this case)

jsproz:
The language reference does indicate the post-conditions are executed in the reverse order:

"Similarly, for post-conditions, the same linearization of interfaces would be used, and the post-conditions are executed in the **reverse order**."

SupunS:
This PR only adds support for before-statements compilation, on top of the existing framework.
The follow-up PR https://github.com/onflow/cadence/pull/3764, inlines everything, including what this PR add, as well what is already there (associated with function-conditions).

We can also merge #3764 into this, so we don't have to review the outdated code.

bbq/compiler/desugar.go
```diff
@@ -502,7 +516,26 @@ func (d *Desugar) desugarPostConditions(
...
 			}
 
-			resultVarType, resultVarExist := d.elaboration.ResultVariableType(inheritedFunc.functionDecl.FunctionBlock)
+			// Result variable must be looked-up in the corresponding elaboration.
+			resultVarType, resultVarExist := inheritedFunc.elaboration.ResultVariableType(inheritedFunc.functionDecl.FunctionBlock)
+
+			// If the inherited function has before-statements, then add an invocation
+			// to call the generated before-statements-function of the interface.
+			//
+			// IMPORTANT: inherited before-functions must be visited before the
```
turbolent, line 525: 👍

turbolent (approved):
Great work!

It would be nice to also have some compiler (unit) tests for this functionality. We can do that in a follow-up PR, added it to #3773 for now to track it.

SupunS:
Yeah, I'll add some compiler tests around these once both the PR's are merged (didn't add those before, until the compilation  approach is somewhat finalized) 👍

PR #3764, opened 2025-02-12
[Compiler] Inline inherited conditions instead generating a separate function

SupunS:

## Description

Inline inherited-conditions at the concrete-type's function (by copying them over) instead of generating a separate function at the interface type.

No change to the default functions. They continue to be generated at the interface type.

______

### DIFF
--- a/ast/variable_declaration.go
+++ b/ast/variable_declaration.go
@@ -153,11 +153,13 @@ func (d *VariableDeclaration) Doc() prettier.Doc {
 
 	// Desugared `result` variable is uninitialized at first,
 	// hence would not have a value.
-	var valueDoc prettier.Doc
+	var valueDoc, transferDoc prettier.Doc
 	if d.Value != nil {
 		valueDoc = d.Value.Doc()
+		transferDoc = d.Transfer.Doc()
 	} else {
 		valueDoc = prettier.Line{}
+		transferDoc = prettier.Line{}
 	}
 
 	var valuesDoc prettier.Doc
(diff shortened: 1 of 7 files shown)


ast/variable_declaration.go
```diff
@@ -153,11 +153,13 @@ func (d *VariableDeclaration) Doc() prettier.Doc {
 
 	// Desugared `result` variable is uninitialized at first,
 	// hence would not have a value.
```
turbolent, line 155:
```suggestion
	// hence would not have a value or transfer.
```

bbq/compiler/compiler.go
```diff
@@ -575,6 +575,18 @@ func (c *Compiler[_]) VisitContinueStatement(_ *ast.ContinueStatement) (_ struct
...
+	// If statements can be coming from inherited conditions.
+	// If so, use the corresponding elaboration.
+	stmtElaboration, ok := c.ExtendedElaboration.conditionsElaborations[statement]
+	if ok {
+		prevElaboration := c.ExtendedElaboration
+		c.ExtendedElaboration = stmtElaboration
+		defer func() {
+			c.ExtendedElaboration = prevElaboration
+		}()
+	}
```
turbolent, line 588: Maybe we can extract this into a "wrapper" function which takes a function that is invoked while the extended elaboration is effective, e.g. something like 

```go
func (c *Compiler[_]) withConditionExtendedElaboration(statement ast.Statement, f func()) {
	stmtElaboration, ok := c.ExtendedElaboration.conditionsElaborations[statement]
	if ok {
		prevElaboration := c.ExtendedElaboration
		c.ExtendedElaboration = stmtElaboration
		defer func() {
			c.ExtendedElaboration = prevElaboration
		}()
	}
	f()
}
```

turbolent (approved):
Great work!

bbq/compiler/compiler.go
SupunS: Good idea! Added in 48c8488da8a147e455775bebe77bc6396a63bf97

PR #3766, opened 2025-02-13
Decouple interpreter values from interpreter – Part 4

turbolent:
Work towards #3693

## Description

Move the core functionality of the value implementations of `Bool` and `Int` out of the `interpreter` package and into a new `values` package. This demonstrates how we could decouple at least some of the more basic values from the interpreter and use them in the VM.

If this approach looks good, we can also refactor the other number value types, and some other primitive value types like e.g. `String`.

I'm happy with the `values` package itself, but honestly not really happy with how the integration looks in the `interpreter` package's `BoolValue` and `IntValue` types, especially all the unwrapping and wrapping, casting, type checking, error handling, etc. Maybe this is just temporary, and once we have refactored the value types in the `interpreter` package fully from the interpreter this ugliness will go away.

Another idea I had was not to create a whole new `values.Value` type hierarchy with methods, but instead just have global functions, e.g. instead of having `IntValue.LessEqual(other ComparableValue)`, just have `IntLessEqual(IntValue, IntValue)`.

Alternatively we would have to *duplicate* this code in the VM, which seems worse to me (e.g. potential for mismatch).

Let's discuss this in the Implemention Sync.

______

### DIFF
--- a/values/value_int.go
+++ b/values/value_int.go
@@ -68,7 +68,7 @@ func NewUnmeteredIntValueFromBigInt(value *big.Int) IntValue {
 
 var _ Value = IntValue{}
 var _ EquatableValue = IntValue{}
-var _ ComparableValue = IntValue{}
+var _ ComparableValue[IntValue] = IntValue{}
 var _ NumberValue[IntValue] = IntValue{}
 var _ IntegerValue[IntValue] = IntValue{}
 var _ atree.Storable = IntValue{}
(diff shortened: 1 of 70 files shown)


values/value_int.go
```diff
@@ -0,0 +1,387 @@
...
+			return new(big.Int).Neg(v.BigInt)
+		},
+	)
+}
+
+func (v IntValue) Plus(gauge common.MemoryGauge, other NumberValue) (NumberValue, error) {
+	o, ok := other.(IntValue)
+	if !ok {
+		return nil, InvalidOperandsError{}
+	}
```
SupunS, line 97: We could get rid f some of these type-casts by making the interfaces (`NumberValue`, `IntegerValue`) generic. e.g: https://github.com/onflow/cadence/compare/bastian/decouple-values-from-interpreter-4...supun/bind-interface-to-type?expand=1#diff-4ba970fa0343b83d635e65319f40746490752a9e69d40072f87221672ce16c81R23-R29

But the downside is, the interface, say `IntegerValue`, can only be used to define the structure of all the integer values. e.g:
```cadence
var _ IntegerValue[IntValue] = IntValue{} // Ensures the `IntValue` implements `IntegerValue`
```
But cannot be actually used in the code, because `IntegerValue` interface type would always have to be used with a `T`, making the type too-sepicific.

```cadence
var someValue Value = ...

 integerValue := someValue.(IntegerValue[IntValue]) // This becomes pointless since `IntegerValue[IntValue]` is not generic enough
```

SupunS (approved):

jsproz (approved):

values/value_int.go
turbolent: This is great! As discussed offline, this improvement was opened as #3790, and it + follow-up improvement [5004f33](https://github.com/onflow/cadence/pull/3766/commits/5004f33e1d3a6ef612c2d36384b15c26c8f5b4ec) got merged.

turbolent:
@SupunS Could you please have a final look at the added commits? 🙏 

Once merged, I'll also refactor `UFix64` value for #3791, and then we can sync master into the compiler branch, and start using the new `values` package in the VM

SupunS (approved):
Looks great!

PR #3767, opened 2025-02-13
[Compiler] Support basic casting operations in VM

SupunS:

## Description

Adds basic support for type-casting in the VM.

Conversion for values based on type (e.g: `Int16` to `Int32`, etc.) is not yet supported (left a TODO). It makes more sense to add those once the interpreter values are re-used. 

______


turbolent (approved):
Nice!

I guess the core functions of the implementation, like `ConvertAndBox`, `Unbox`, `BoxOptional` are "copies" of the one in the interpreter? Maybe we can refactor them (in a follow-up) so they only exist in one place so we can share the logic between interpreter and VM, and there is no chance their behaviour drifts in different directions?

SupunS:
Yes. It's not possible to refactor yet, because the values are different, and also interpreter's functions rely on `sema.Type` and the VM's functions rely on `StaticType`s. For the latter, I think the better approach is probably to refactor both to use StaticTypes. But again for that, we need to figure out a way / refactor the entitlements to also use StaticTypes.

I've added a item to our laundry-list for the next milestone, for adding those.

SupunS:
Like discussed, I split the different variants of casts in to separate instructions: https://github.com/onflow/cadence/pull/3767/commits/c76e75a8b823d8541f43bc7a9e78930c75876dcf.

Can you please have another look on the last commit? 🙏

turbolent (approved):
Nice! Looks great 👍

bbq/opcode/print_test.go
```diff
@@ -101,7 +101,7 @@ func TestPrintInstruction(t *testing.T) {
 
 		"New kind:CompositeKind(258) typeIndex:772": {byte(New), 1, 2, 3, 4},
 
-		"Cast typeIndex:258 kind:3": {byte(Cast), 1, 2, 3},
+		"SimpleCast typeIndex:258": {byte(SimpleCast), 1, 2, 3},
```
turbolent, line 105: Can you please also add tests for the other two new casting instructions?

turbolent:
Opened https://github.com/onflow/cadence/pull/3771 to also add some test cases for the compilation

turbolent (approved):
👌

PR #3768, opened 2025-02-13
[Compiler] Compile and execute emit statement

turbolent:
Work towards #3742

## Description

Compile `emit` statements to a new `EmitEvent` instruction, and execute the instruction by invoking a config callback.

______

### DIFF
--- a/bbq/vm/vm.go
+++ b/bbq/vm/vm.go
@@ -804,12 +804,29 @@ func (vm *VM) run() {
 			opNot(vm)
 		case opcode.InstructionUnwrap:
 			opUnwrap(vm)
+		case opcode.InstructionEmitEvent:
+			onEmitEvent(vm, ins)
 		default:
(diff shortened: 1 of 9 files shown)


bbq/vm/vm.go
```diff
@@ -813,6 +815,21 @@ func (vm *VM) run() {
 	}
 }
 
+func onEmitEvent(vm *VM, ins opcode.InstructionEmitEvent) {
+	eventType := vm.loadType(ins.TypeIndex).(*interpreter.CompositeStaticType)
```
SupunS, line 819: Maybe move this to after the if condition `if onEventEmitted == nil { }`. i.e: We wouldn't need to load the type if the callback is not set.

SupunS (approved):
👌

bbq/vm/vm.go
turbolent: Good idea! Done in 9ee137fd6056cdf0edbc7e53094f16a08f8073ba

Issue #3769, opened 2025-02-14
Compiler Milestone 3: Realistic FT and Flow token transfer for book-keeping functions


SupunS:
### Objective

Complete functionality to be able to execute FT transfer transaction (functional parity with interpreter for FT transfer) as part of the book-keeping functions (Account balance check, Storage limit check, Fee deduction)

## Key result

Compiler and VM are able to run real FT and Flow Token contracts.

## DACI

| Role | Assigned |
| ------ | ---------|
| Driver |  technical: @SupunS , EM: @j1010001  |
| Approver | @dete |
| Consulted| @turbolent, @Kay-Zee  |
| Informed| Flow Engineering team |

### Task breakdown

In order of complexity and priority:

- Compiler + VM:
  - [x] Closures @turbolent 

- Entitlements (subset needed for FT; only access modifier?) @SupunS 
  - [x] Support member-access with entitlement mappings. (e.g: `access(M) var field`)

- Standard library functionality / function injection (mostly, no work in compiler / VM. Should be fairly easy:
  - [x] `Type`
  - [x] `Address`
  - [x] `Account.Storage` (e.g. `copy`)
  - [x] `Account.Capabilities` (e.g. `borrow`)
  - [x] Number conversion functions (e.g. `UFix64()`)

- Computed fields:
  - [x] Contract `account` field @turbolent 
  - [x] Resource `owner` field @turbolent 
  - [x] Resource `uuid` field 

### Tech-debt
- [x] https://github.com/onflow/cadence/issues/3758
- [x] https://github.com/onflow/cadence/issues/3693
- [x] https://github.com/onflow/cadence/issues/3770
- [x] Research how to run existing interpreter and runtime tests with the compiler/VM (focus on interpreter tests, runtime tests later) https://github.com/onflow/cadence/pull/3813

### Optimizations
- [x] Disable pre-loading types in the VM for executable (unnecessarily loads all types, even if they are not used)
- [x] Maybe produce an un-encoded/decoded executable directly from the compiler, to avoid encoding-decoding times (e.g: for static types, etc).

### Done

- [x] Boolean operators @turbolent 
- [x] Bitwise operators @turbolent 
- [x] Conditional expression @turbolent 
- [x] Unary operators (`*`, `-`) @turbolent 
- [x] Transaction pre/post conditions @SupunS 
- [x] Force-unwrap @turbolent 

### Remaining

Identified by https://github.com/onflow/cadence/pull/3860

- [x] Necessary data structure methods
[...]

PR #3793, opened 2025-02-25
[Compiler] Compile for-loops

### LINKED ISSUE #3784: Add scoping to the compiler
Scoping might be need for local variables. e.g: shadowing inside blocks, (optional binding in if-statement?), (loop variables of for-loops?), etc.

SupunS:
Work towards #3742
Closes #3784

## Description

A for-loop get compiled similar to a while-loop, with an iterator.
```cadence
for e in array {
}
```
Compiles similar to a:
```cadence
var iterator = array.iterator() 
while(iterator.hasNext()) {
    var e = iterator.next()
}
```

The resulting `iterator` is a value (hence can be put on the stack), but is strictly for internal usages.

### Other Notes:
Here, instead of using function-calls for iterator-related operations, three dedicated instructions are used.

| function-call | Instruction |
|---| ---|
|`array.iterator()` | `Iterator` |
|`array.hasNext()` | `IteratorHasNext` |
|`array.next()` | `IteratorNext` |

This is done to reduce the effort on both compilation (i.e: less code to desugar), as well as runtime/vm (i.e: executing a function call vs single instruction.)
______

### DIFF
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -229,14 +229,14 @@ func (c *Compiler[_]) addStringConst(str string) *constant {
 	return c.addConstant(constantkind.String, []byte(str))
 }
 
-func (c *Compiler[_]) intConstLoad(i int64) {
-	constant := c.addIntConst(i)
+func (c *Compiler[_]) intConstLoad(intKind constantkind.ConstantKind, i int64) {
+	constant := c.addIntConst(intKind, i)
 	c.codeGen.Emit(opcode.InstructionGetConstant{ConstantIndex: constant.index})
 }
 
-func (c *Compiler[_]) addIntConst(i int64) *constant {
+func (c *Compiler[_]) addIntConst(intKind constantkind.ConstantKind, i int64) *constant {
 	data := leb128.AppendInt64(nil, i)
-	return c.addConstant(constantkind.Int, data)
+	return c.addConstant(intKind, data)
 }
 
 func (c *Compiler[_]) emitJump(target int) int {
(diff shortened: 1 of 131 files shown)


bbq/compiler/compiler.go
```diff
@@ -229,6 +229,16 @@ func (c *Compiler[_]) addStringConst(str string) *constant {
...
 
+func (c *Compiler[_]) intConstLoad(i int64) {
+	constant := c.addIntConst(i)
+	c.codeGen.Emit(opcode.InstructionGetConstant{ConstantIndex: constant.index})
+}
+
+func (c *Compiler[_]) addIntConst(i int64) *constant {
+	data := leb128.AppendInt64(nil, i)
+	return c.addConstant(constantkind.Int, data)
+}
```
turbolent, line 240: Nice! Could we maybe use this in `VisitIntegerExpression` while at it?

turbolent (approved):
Great work! 👍

bbq/compiler/compiler.go
SupunS: Good idea! Added in 3121eda24c101d3de9be148ec32184a28a4b451f

bbq/compiler/compiler.go
```diff
@@ -931,13 +931,7 @@ func (c *Compiler[_]) VisitNilExpression(_ *ast.NilExpression) (_ struct{}) {
...
 	integerType := c.ExtendedElaboration.IntegerExpressionType(expression)
 	constantKind := constantkind.FromSemaType(integerType)
-
-	// TODO:
-	var data []byte
-	data = leb128.AppendInt64(data, expression.Value.Int64())
-
-	constant := c.addConstant(constantKind, data)
-	c.codeGen.Emit(opcode.InstructionGetConstant{ConstantIndex: constant.index})
+	c.intConstLoad(constantKind, expression.Value.Int64())
```
turbolent, line 934: Small nit: Might be good to leave the TODO here, because we still need to support larger types in the future. In any case, added it to #3769, so we don't forget

turbolent (approved):

PR #3798, opened 2025-02-26
[Compiler] Produce an un-encoded program by the compiler

SupunS:

## Description

Currently, the compiler produces a program where all types are encoded (converted to bytes), and in the VM, these types are decoded back again. However, given compilation happens on the same place, this encoding and then decoding of types is an unnecessary overhead.

Therefore, updated the compiler such that it produce the decoded type, and the VM can use it directly without having to decode. This is similar to emitting instructions vs bytecode.

Micro-benchmarks showed that this reduced about 17% of the compile+execution time for FT transfer.
```
             │ ft_existing │            ft_optimized            │
             │   sec/op    │   sec/op     vs base               │
FTTransfer-8   88.97µ ± 3%   73.54µ ± 2%  -17.34% (p=0.002 n=6)

             │ ft_existing  │            ft_optimized             │
FTTransfer-8   65.76Ki ± 0%   55.75Ki ± 0%  -15.22% (p=0.002 n=6)

             │ ft_existing │            ft_optimized            │
             │  allocs/op  │  allocs/op   vs base               │
FTTransfer-8   1.268k ± 0%   1.033k ± 0%  -18.53% (p=0.002 n=6)
```

We could potentially do the same for constants as well.

______

### DIFF
--- a/bbq/compiler/typegen.go
+++ b/bbq/compiler/typegen.go
@@ -16,16 +16,35 @@
  * limitations under the License.
  */
 
-package bbq
+package compiler
 
 import (
+	"github.com/onflow/cadence/bbq"
 	"github.com/onflow/cadence/interpreter"
 )
 
-type StaticType = interpreter.StaticType
+type TypeGen[T any] interface {
+	CompileType(staticType bbq.StaticType) T
+}
 
-func StaticTypeFromBytes(bytes []byte) (StaticType, error) {
-	dec := interpreter.CBORDecMode.NewByteStreamDecoder(bytes)
-	typeDecoder := interpreter.NewTypeDecoder(dec, nil)
-	return typeDecoder.DecodeStaticType()
+type DecodedTypeGen struct {
+}
+
+var _ TypeGen[bbq.StaticType] = &DecodedTypeGen{}
+
+func (d DecodedTypeGen) CompileType(staticType bbq.StaticType) bbq.StaticType {
+	return staticType
+}
+
+type EncodedTypeGen struct {
+}
+
+var _ TypeGen[[]byte] = &EncodedTypeGen{}
+
+func (d EncodedTypeGen) CompileType(staticType bbq.StaticType) []byte {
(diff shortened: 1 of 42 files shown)


bbq/compiler/typegen.go
```diff
@@ -0,0 +1,50 @@
...
+	CompileType(staticType bbq.StaticType) T
+}
+
+type DecodedTypeGen struct {
+}
+
+var _ TypeGen[bbq.StaticType] = &DecodedTypeGen{}
+
+func (d DecodedTypeGen) CompileType(staticType bbq.StaticType) bbq.StaticType {
+	return staticType
```
turbolent, line 36: 👌

bbq/vm/test/interpreter_test.go
```diff
@@ -799,21 +799,21 @@ func BenchmarkInterpreterFTTransfer(b *testing.B) {
...
+			b,
+			realFlowTokenTransferTransaction,
+			txLocation(),
+			ParseCheckAndInterpretOptions{
+				Config:        interConfig,
+				CheckerConfig: checkerConfig,
+			},
+		)
+		require.NoError(b, err)
+
```
turbolent, line 816: Was this change, now also measuring parsing and checking, intentional?

turbolent (approved):
Great idea and work! 👏 

Makes a lot of sense to apply the same optimization to types as we already previously made for the generated programs ("in-memory" instructions, and byte code).

Maybe we can do the same also for constants? We don't have to do it right away, and can just track it as potential future optimization

bbq/vm/test/interpreter_test.go
SupunS: Yes - I updated both the interpreter-test and the vm-test to benchmark all phases (including parsing and checking), to see where we stand. At the moment there is about a ~13-15% difference: compile+VM is slower by ~13-15%, because of the compilation overhead (previously was ~28-30%). I think we should be on-par with the existing interpreter with some more low-hanging optimizations and the re-use of interpreter values.

PR #3801, opened 2025-02-26
[Compiler] Compile transaction pre/post conditions

SupunS:

## Description

Include the transaction pre/post conditions into the execute function, so that the existing condition desgaring logic would handle and generate code. 

______

### DIFF
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -565,6 +565,10 @@ func (c *Compiler[_, _]) compileBlock(block *ast.Block) {
 }
 
 func (c *Compiler[_, _]) compileFunctionBlock(functionBlock *ast.FunctionBlock) {
+	if functionBlock == nil {
+		return
+	}
+
 	// Function conditions must have been desugared to statements.
 	// So there shouldn't be any condition at this point.
 	if functionBlock.PreConditions != nil ||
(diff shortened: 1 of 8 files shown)


turbolent:
@SupunS See "Needed for more realistic FT" in #3769. These might not be all, but at least the ones I noticed.

bbq/compiler/compiler.go
```diff
@@ -565,11 +565,18 @@ func (c *Compiler[_, _]) compileBlock(block *ast.Block) {
...
+	}
+
 	// Function conditions must have been desugared to statements.
 	// So there shouldn't be any condition at this point.
-	if functionBlock != nil {
-		c.compileBlock(functionBlock.Block)
+	if functionBlock.PreConditions != nil ||
+		functionBlock.PostConditions != nil {
+		panic(errors.NewUnreachableError())
 	}
```
turbolent, line 577: Good idea to add assertions like this one 👍

bbq/compiler/desugar.go
```diff
@@ -1151,11 +1149,55 @@ func (d *Desugar) VisitTransactionDeclaration(transaction *ast.TransactionDeclar
...
+			}
+
+			executeFuncType := sema.NewSimpleFunctionType(
+				sema.FunctionPurityImpure,
+				nil,
+				sema.VoidTypeAnnotation,
+			)
+
+			d.elaboration.SetFunctionDeclarationFunctionType(executeFunc, executeFuncType)
+		}
```
turbolent, line 1190: Is there maybe a way to avoid this unnecessary allocation of an empty function?

turbolent (approved):
Nice!

bbq/compiler/desugar.go
SupunS: The function is created only if there are conditions (and no execute block/function). Since conditions are just statements, they anyway going to need a wrapper function.

turbolent: Ah, I see. Maybe we can allocate the function declaration once and reuse it? Maybe we can at least reuse the type?

turbolent (approved):
Nice!

PR #3803, opened 2025-02-27
[Compiler] Compile force expression

turbolent:
Work towards #3769

## Description

- Compile force operator / expression (postfix `!`) to `Unwrap` instruction.
- Change the implementation of the `Unwrap` instruction in the VM to panic on `nil`

______

### DIFF
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -1455,9 +1455,10 @@ func (c *Compiler[_, _]) VisitReferenceExpression(expression *ast.ReferenceExpre
 	return
 }
 
-func (c *Compiler[_, _]) VisitForceExpression(_ *ast.ForceExpression) (_ struct{}) {
-	// TODO
-	panic(errors.NewUnreachableError())
+func (c *Compiler[_, _]) VisitForceExpression(expression *ast.ForceExpression) (_ struct{}) {
+	c.compileExpression(expression.Expression)
+	c.codeGen.Emit(opcode.InstructionUnwrap{})
+	return
 }
 
 func (c *Compiler[_, _]) VisitPathExpression(expression *ast.PathExpression) (_ struct{}) {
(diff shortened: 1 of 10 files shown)


bbq/vm/vm.go
```diff
@@ -797,9 +797,11 @@ func opNot(vm *VM) {
...
 func opUnwrap(vm *VM) {
 	value := vm.peek()
-	if someValue, ok := value.(*SomeValue); ok {
-		value = someValue.value
-		vm.replaceTop(value)
+	switch value := value.(type) {
+	case *SomeValue:
+		vm.replaceTop(value.value)
+	case NilValue:
+		panic(ForceNilError{})
```
SupunS, line 804: Maybe panic in the `default` case with an unreachable-error

SupunS (approved):

bbq/vm/vm.go
turbolent: Added in [ff2aec9](https://github.com/onflow/cadence/pull/3803/commits/ff2aec9ee3658020ce360f316016d0cd1570b90e).

This made me realize that we actually allow force-unwrapping of non-optional values, so I adjusted compilation to support it (omit unwrap if not needed) in [a55dfe5](https://github.com/onflow/cadence/pull/3803/commits/a55dfe5e0563d8d68684000987cd7416f9f65544)

bbq/compiler/compiler.go
```diff
@@ -1457,7 +1457,13 @@ func (c *Compiler[_, _]) VisitReferenceExpression(expression *ast.ReferenceExpre
 
 func (c *Compiler[_, _]) VisitForceExpression(expression *ast.ForceExpression) (_ struct{}) {
 	c.compileExpression(expression.Expression)
-	c.codeGen.Emit(opcode.InstructionUnwrap{})
+
+	// Unwrap if optional
+	valueType := c.ExtendedElaboration.ForceExpressionType(expression)
+	if _, ok := valueType.(*sema.OptionalType); ok {
+		c.codeGen.Emit(opcode.InstructionUnwrap{})
+	}
```
SupunS, line 1465: Wonder if it is safe to determine this at compile time. For e.g.: there can be optional values assign to `AnyStruct` and then trying to unwrap.
```cadence
var a: Int? = 5
var b: AnyStruct = a
var c = b!
```

turbolent: You're probably right, we should match this exactly to what the interpreter is doing.

turbolent: Should be fixed by e96db5fa3213d00d0cce61c43723a929f91007a1

SupunS (approved):

PR #3809, opened 2025-03-04
Fix get-contracts workflow

turbolent:
## Description

We want to be able to trigger `get-contracts` to be manually. This is done by `workflow_dispatch`, not `workflow_call`.

______

### DIFF
--- a/.github/workflows/get-contracts.yml
+++ b/.github/workflows/get-contracts.yml
@@ -1,7 +1,7 @@
 name: Get contracts
 
 on:
-  workflow_call:
+  workflow_dispatch:
     inputs:
       chain:
         required: true


bluesign:
will it also trigger automatically ?

turbolent:
No, this is just a helper that allows us to get the current contracts, when we need them. I guess we/you could also run it periodically, e.g. get a snapshot every hour, if there's demand (and Find Labs is OK with it, because we're using their API). 

This action is actually just extracted from the existing backward compatibility checking, that runs on each PR, and also fetches all contracts: https://github.com/onflow/cadence/actions/workflows/compatibility-check.yml. That job makes the checking results and also the contracts available as artifacts. However, that one caches the contracts once per day: https://github.com/onflow/cadence/blob/master/.github/workflows/compatibility-check-template.yml#L51

SupunS (approved):

PR #3813, opened 2025-03-11
Run existing interpreter tests with the compiler and VM

SupunS:

## Description

This way, it only needs to switch to the new test-utility method.

I only did it for the `interpreter/while_test.go` as it was one of the simplest ones. If this sounds like a good idea, then we can gradually expand this to other tests.

We might need to abstract/expose a few more things as we go, for other complex tests. But the idea would be the same.

______


interpreter/while_test.go
```diff
@@ -24,14 +24,15 @@ import (
...
+	"github.com/onflow/cadence/test_utils"
 	. "github.com/onflow/cadence/test_utils/interpreter_utils"
 )
 
 func TestInterpretWhileStatement(t *testing.T) {
 
 	t.Parallel()
 
-	inter := parseCheckAndInterpret(t, `
+	invokable := test_utils.ParseCheckAndPrepare(t, `
```
SupunS, line 35: This is all we need to change in the test files: switch to the new API

test_utils/test_utils.go
```diff
@@ -0,0 +1,375 @@
...
+	CheckerConfig      *sema.Config
+	HandleCheckerError func(error)
+}
+
+var compile = false
+
+func init() {
+	// TODO: Set the env-variable and re-run the tests.
+	// compile = os.Getenv("COMPILE") == "true"
+}
```
SupunS, line 33: This is disabled by default, for now. I'll open a follow-up PR to enable them

test_utils/test_utils.go
```diff
@@ -0,0 +1,375 @@
...
+		vmInstance:  vmInstance,
+		config:      vmConfig,
+		Interpreter: vmConfig.Interpreter(),
+	}
+
+}
+
+// Below helper functions were copied as-is from `misc_test.go`.
+// Idea is to eventually use the below functions everywhere, and remove them from `misc_test.go`,
+// so that the `misc_test.go` would contain only tests.
```
SupunS, line 89: Similar here, I'll open a follow-up PR to refactor all existing tests to use the API in this file (still using the interpreter, not the compiler), and remove them from `misc_test.go`.

I didn't do it here because it will make this PR huge, and also is not related to "using the compiler in existing tests" (i.e: this is an unrelated refactor)

interpreter/while_test.go
turbolent: This is great! Maybe use a dot import for `test_utils` like we have for `interpreter_utils`

test_utils/interpreter_utils/values.go
```diff
@@ -49,20 +49,20 @@ func AssertValueSlicesEqual(t testing.TB, inter *interpreter.Interpreter, expect
 	return true
 }
 
-func ValuesAreEqual(inter *interpreter.Interpreter, expected, actual interpreter.Value) bool {
+func ValuesAreEqual(context interpreter.ValueComparisonContext, expected, actual interpreter.Value) bool {
```
turbolent, line 52: 👌

test_utils/test_utils.go
turbolent: We can maybe make this a flag, similar to how we have for the smoke tests (see `runSmokeTests`)

turbolent (approved):
Great idea and great work!

This will allow us to incrementally run more and more existing tests for interpreter and VM, and will save us from having to duplicate tests. Nice 👌

test_utils/test_utils.go
SupunS: For the moment, maybe we can use the flag, but in the makefile, we would have to enable each package separately. This might also makes sense, since we don't want to run all the tests with the compiler (e.g: type checker tests, tools-tests, etc.). Will update and see.

interpreter/while_test.go
SupunS: done in ecdbe8e3a484ed1d2f9f8c5623d2fb17d8f198b5

test_utils/test_utils.go
SupunS: Added a temporary solution in https://github.com/onflow/cadence/pull/3813/commits/ecdbe8e3a484ed1d2f9f8c5623d2fb17d8f198b5

turbolent (approved):
Nice! 👏

Issue #3584, opened 2024-09-13
Optimize Account Storage Format (reduce domain payloads count by ~29%, etc.)


fxamacker:
## Why (Objective)

UPDATE: With Jan 31, 2025 mainnet state, this optimization reduces state size by 11.1%, eliminates 28.2% of mtrie vertexes (leaf and non-leaf nodes), and also reduces payload count by 28.2%.  Our success criteria was to reduce state size by 10-12%, payload count by 25-30%, and eliminate 25-30% of mtrie vertexes. :tada:

Reducing payload count and mtrie vertexes has benefits beyond EN memory reduction.  E.g. last year, when [Atree inlining](https://github.com/onflow/atree/pull/429) in [Atree v0.8](https://github.com/onflow/atree/releases/tag/v0.8.0) reduced payload count by 61% and eliminated 61% of mtrie vertexes, it improved transaction speed by ~7% and @bluesign reported TinyAN bootstrap time reduced by 5x.

We can reduce payload count and mtrie vertex count by:
- replacing all domain payloads per account with one account payload
- using account storage map (atree payload) referencing domain storage map 
 
We can further reduce execution state size by using atree register inlining feature added in [Atree v0.8.0](https://github.com/onflow/atree/releases/tag/v0.8.0).

This improvement can be deployed via HCU and implements first on-the-fly migration to avoid causing further downtime.

The full impact on memory reduction (TBD) will be realized after all the accounts eventually migrate all their non-atree domain registers to a single combined atree payload per account.

NOTE: Memory reduction is only 1 benefit from reducing payload count.

## How will we measure success (Key Results) ?

After all the non-atree domain registers are migrated to 1 atree map per account we expect:
- **reduce execution state size by 10-12% (~40GB as of Sept. 2024 mainnet state)**
- reduce payload count by 20-30% (by combining payloads)
- reduce mtrie nodes by 20-30% (by eliminating mtrie nodes)
- reduce EN memory use by TBD (usually better than state size reduction)

Requires HCU.
If we use on-the-fly migration, then the full impact won't be realized until all accounts are migrated.

Requires "cleanup" migration for accounts that were not migrated on-the-fly.

## Estimate

~3dev/weeks to get ready for deployment.

## DACI

| Role | Assigned |
| ------ | ---------|
| Driver | Technical: @fxamacker , EM: @j1010001|
| Approver | @dete  |
| Consulted| @ramtinms, @zhangchiqing, @turbolent, @janezpodhostnik |
[...]

PR #3814, opened 2025-03-13
Remove support for account storage format V1

turbolent:
Follow-up to #3584

## Description

Now that all production networks were migrated to storage format V2, we can remove format V1.

However, to avoid having to deploy this change in a HCU / avoid making this change produce a different execution result, we need to keep the existing logic / reads to determine the storage format.

If an account in format V1 is detected, abort execution with a new error.

______


runtime/storage.go
```diff
@@ -176,23 +162,6 @@ func (s *Storage) GetDomainStorageMap(
...
-			domain,
-			createIfNotExists,
-		)
-
-		return
-	}
-
-	// StorageFormatV2 is enabled.
-
 	// Check if cached account format is available.
```
turbolent, line 165: Unfortunately we won't be able to remove the read of the domain register / V1 reads here, because that will produce a different execution result, i.e. require a HCU.

fxamacker (approved):
Nice!  Thanks for removing code related to storage format V1.

Since there are still register reads related to storage format V1, I opened an issue #3815 as a reminder for next HCU.

turbolent:
@SupunS Could you please take a look? We could try including this in the upcoming HCU

PR #3816, opened 2025-03-20
Support compilation for conditional returns

### LINKED ISSUE #3772: Conditional returns compilation
At the moment, the compiler only supports the return statement at the end of the function.

Needs to add support for conditional-returns. e.g:
```cadence
fun main(_ a: Int): Int {
    if a > 5 {
        return 1
    }

    return 2
}

### PR DESCRIPTION
**Pull request by SupunS**
Closes #3772

## Description

**Design:**

When a return statement is being compiled, if the enclosing function:
- Doesn't have post conditions, then return in-place (same as before).
- Has post conditions:
  - Instead of emitting a return, jump to the start of the post conditions (post conditions are at the bottom of the function, after all the regular statements)
  - At the end of post conditions, emit the return.

______

### CODE REVIEW

File: bbq/compiler/desugar.go [hunk 1]
```diff
@@ -621,7 +535,16 @@ func (d *Desugar) desugarCondition(condition ast.Condition) ast.Statement {
...
+	desugaredDecl := d.desugarDeclaration(declaration.FunctionDeclaration).(*ast.FunctionDeclaration)
+	if desugaredDecl == declaration.FunctionDeclaration {
+		return declaration
+	}
+
+	return ast.NewSpecialFunctionDeclaration(
+		d.memoryGauge,
+		declaration.Kind,
+		desugaredDecl,
+	)
```
SupunS, line 547: Unrelated fix: Initializers wasn't getting desugared (hence conditions in initializers wasn't working)

bbq/compiler/desugar.go
```diff
@@ -633,10 +556,15 @@ func (d *Desugar) VisitCompositeDeclaration(declaration *ast.CompositeDeclaratio
...
+	prevInheritedFuncsWithConditions := d.inheritedFuncsWithConditions
+	prevEnclosingInterfaceType := d.enclosingInterfaceType
+
 	d.inheritedFuncsWithConditions = d.inheritedFunctionsWithConditions(compositeType)
+	d.enclosingInterfaceType = nil
+
 	defer func() {
-		d.inheritedFuncsWithConditions = prevInheritedFuncs
+		d.inheritedFuncsWithConditions = prevInheritedFuncsWithConditions
+		d.enclosingInterfaceType = prevEnclosingInterfaceType
```
SupunS, line 567: Also an unrelated fix, but needed for the FT test. Composite types can be inside interfaces, in the form of events. And events have initializer functions, that should be desugar-ed/compiled. This path was uncovered by the desugar-ing of initializers (fix) above.

bbq/compiler/compiler.go
```diff
@@ -52,8 +52,13 @@ type Compiler[E, T any] struct {
 	usedImportedGlobals []*global
 	controlFlows        []controlFlow
 	currentControlFlow  *controlFlow
+	returns             []returns
+	currentReturn       *returns
 	staticTypes         []T
 
+	postConditionsIndices map[ast.Declaration]int
+	postConditionsIndex   int
```
turbolent, line 60: Could you please add a comment explaining the purpose of these new fields? Is it a "statement index"?

bbq/compiler/compiler.go
```diff
@@ -577,17 +601,75 @@ func (c *Compiler[_, _]) compileDeclaration(declaration ast.Declaration) {
...
+	case common.DeclarationKindFunction, common.DeclarationKindInitializer:
+		if c.hasPostConditions() {
+			c.pushReturns()
+		}
+	}
+
+	for index, statement := range block.Statements {
+		// Once the post conditions are reached, patch all the previous return statements
+		// to jump to the current index (i.e: update them to jump to the post conditions).
+		if index == c.postConditionsIndex {
```
turbolent, line 620: How are functions with nested blocks handled? `compileFunctionBlock` sets the `postConditionsIndex` field, but then here in `compileBlock` it seems to be checked for all blocks of the function.

bbq/compiler/compiler.go
```diff
@@ -612,16 +705,67 @@ func (c *Compiler[_, _]) compileExpression(expression ast.Expression) {
 
 func (c *Compiler[_, _]) VisitReturnStatement(statement *ast.ReturnStatement) (_ struct{}) {
 	expression := statement.Expression
+
+	// There can be five different variations of return values:
```
turbolent, line 709: Great comment👌

bbq/compiler/desugar.go
turbolent: Good catch! Could you please add a test case for this? (maybe there is one already, I might have missed it)

bbq/compiler/desugar.go
turbolent: Same as above: Is this fix already tested?

turbolent (commented):
Great work! 👏 The updated compiler tests look much cleaner 👌 

The logic looks good, just wondering a bit about how `postConditionsIndices` / `postConditionsIndex` works

bbq/compiler/desugar.go
SupunS: Yeah, there is already a test for this (a test with conditions in the initializer - the desugaring only has an impact if there are conditions)

bbq/compiler/compiler.go
SupunS: That's a good point! It should only patch the jumps at the function-block level, but not at the nested block levels. Currently it does both, and the same jump get patched twice (incorrectly), but since the outer-most patch is always the function-block (a "sandwich"), it ends-up producing the correct jump-index eventually. 

Fixed it in https://github.com/onflow/cadence/pull/3816/commits/e9b789ea18187043dc63f0dc37eda4f8eb5316dc, to only patch the returns for function-blocks, but not for nested/non-function blocks.
However, it's not actually test-able, since the end result is the same.

Great catch!

bbq/compiler/compiler.go
SupunS: I had added the comment in the desugar, where this was used. Copied the same here: https://github.com/onflow/cadence/pull/3816/commits/e0c6f34241118d3c50ff1df30783b32beaaaff5c

bbq/compiler/compiler.go
```diff
@@ -52,8 +52,17 @@ type Compiler[E, T any] struct {
...
+	returns             []returns
+	currentReturn       *returns
 	staticTypes         []T
 
+	// postConditionsIndices keeps track of where the post conditions start (i.e: index of the statement in the block),
+	// for each function. This is used by the compiler to patch the jumps for return statements.
+	postConditionsIndices map[ast.Declaration]int
+
+	// postConditionsIndex is the statement-index of the post-conditions for the current function.
+	postConditionsIndex int
```
turbolent, line 64: Does this assume the post conditions were added by the desugaring/rewrite to the end of the function's block?

bbq/vm/test/vm_test.go
```diff
@@ -5311,5 +5210,115 @@ func TestCompileForce(t *testing.T) {
 		require.NoError(t, err)
 		assert.Equal(t, vm.NewIntValue(42), actual)
 	})
+}
+
+func TestReturnStatements(t *testing.T) {
```
turbolent, line 5215: Could you please add (a) test case(s) using the `result` variable in the post condition? (maybe they exist already in another test function)

SupunS: There is a bunch of tests under the post-conditions tests: https://github.com/onflow/cadence/blob/bde5c821fbef69e6ccb00cf80f49850b5d6298bd/bbq/vm/test/vm_test.go#L3631-L3734

bbq/compiler/compiler.go
SupunS: Correct. This `postConditionsIndices` is populated by/during the desugar/rewrite

turbolent: Makes sense 👍 

Could you please add this detail to the comment? It seems essential to understanding where the information comes from

bbq/vm/test/vm_test.go
turbolent: Nice!

bbq/compiler/compiler.go
SupunS: Improved the comment here: https://github.com/onflow/cadence/pull/3816/commits/49b91a2a4f9916b9d6d6ce5d23c3c6debe0ab9ff

bbq/compiler/compiler.go
```diff
@@ -591,17 +622,86 @@ func (c *Compiler[_, _]) compileDeclaration(declaration ast.Declaration) {
...
-func (c *Compiler[_, _]) compileBlock(block *ast.Block) {
+func (c *Compiler[_, _]) compileBlock(block *ast.Block, enclosingDeclKind common.DeclarationKind) {
 	locals := c.currentFunction.locals
 	locals.PushNewWithCurrent()
 	defer locals.Pop()
 
-	for _, statement := range block.Statements {
-		c.compileStatement(statement)
+	if c.shouldPatchReturns(enclosingDeclKind) {
+		c.pushReturns()
```
turbolent, line 631: The push is gated by `shouldPatchReturns`, but the pop below is gated by effectively `(enclosingDeclKind == common.DeclarationKindFunction || enclosingDeclKind == common.DeclarationKindInitializer) && hasPostConditions`. That matches `shouldPatchReturns` at the moment, but might get out-of-sync. Maybe ensure the corresponding pop for the push with a `defer`

bbq/compiler/desugar.go
```diff
@@ -71,10 +72,11 @@ func NewDesugar(
 		checker:                      checker,
 		importsSet:                   map[common.Location]struct{}{},
 		inheritedFuncsWithConditions: map[string][]*inheritedFunction{},
+		postConditionIndices:         map[ast.Declaration]int{},
 	}
 }
 
-func (d *Desugar) Run() *ast.Program {
+func (d *Desugar) Run() (*ast.Program, map[ast.Declaration]int) {
```
turbolent, line 79: Maybe name the return values, especially for the second return value it is not clear what it is just based off of the type

turbolent (approved):
Great work! 👏

bbq/compiler/compiler.go
SupunS: Good idea! Added in https://github.com/onflow/cadence/pull/3816/commits/fc87b300480b695dd05bfd2ea0f267fa47ff8a84

turbolent (approved):
👌

PR #3817, opened 2025-03-21
[Compiler] Refactor `vm.Value` and related operations to be independent of the `vm.Config`

SupunS:
Work towards https://github.com/onflow/cadence/issues/3693

## Description

Similar to the refactor in the interpreter, avoid the `vm.Config` from being passed into value related functions (`StaticType`, `Transfer`, etc.) at the VM side. Instead, pass only what is needed for each function, as an interface (re-use interpreter's interfaces wherever possible).

In the follow-up PR https://github.com/onflow/cadence/pull/3819, similar interfaces are defined at the interpreter (e.g: `ValueTransferContext`) and decoupled more of `interpreter.Value`s.

The idea is to gradually decouple values of both sides (both vm and interpreter), and make bring the interfaces defined in both sides to an exact match, which will allow us to switch and replace vm-values from the interpreter-values.

______

### DIFF
--- a/bbq/vm/config.go
+++ b/bbq/vm/config.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/onflow/cadence/bbq/commons"
 	"github.com/onflow/cadence/common"
+	"github.com/onflow/cadence/errors"
 	"github.com/onflow/cadence/interpreter"
 	"github.com/onflow/cadence/sema"
 	"github.com/onflow/cadence/stdlib"
@@ -58,6 +59,10 @@ type Config struct {
 }
 
 var _ ReferenceTracker = &Config{}
+var _ StaticTypeContext = &Config{}
+var _ TransferContext = &Config{}
+var _ interpreter.StaticTypeConversionHandler = &Config{}
+var _ interpreter.ValueComparisonContext = &Config{}
 
 func NewConfig(storage interpreter.Storage) *Config {
 	return &Config{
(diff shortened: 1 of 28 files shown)


bbq/vm/config.go
```diff
@@ -102,6 +111,107 @@ func (c *Config) Interpreter() *interpreter.Interpreter {
...
+
+func (c *Config) IsSubType(subType interpreter.StaticType, superType interpreter.StaticType) bool {
+	inter := c.Interpreter()
+	return inter.IsSubType(subType, superType)
+}
+
+func (c *Config) IsSubTypeOfSemaType(staticSubType interpreter.StaticType, superType sema.Type) bool {
+	inter := c.Interpreter()
+	return inter.IsSubTypeOfSemaType(staticSubType, superType)
+}
```
SupunS, line 185: These functions are later removed (in follow-up PRs), as the interpreter functions were re-used.  So this is kind of temporary.

bbq/vm/value.go
```diff
@@ -22,28 +22,58 @@ import (
...
 		address atree.Address,
 		remove bool,
 		storable atree.Storable,
 	) Value
 	String() string
 }
 
+type StaticTypeContext = interpreter.ValueStaticTypeContext
+
+type StorageContext interface {
```
SupunS, line 43: These (and below) contexts/interfaces are also removed in follow-up PRs, as the interpreter functions/values were re-used. So this is kind of temporary.

PR #3821, opened 2025-03-24
[Compiler] Decouple `Value.MeteredString` method from the interpreter

SupunS:
Work towards https://github.com/onflow/cadence/issues/3693

Depends on https://github.com/onflow/cadence/pull/3819

## Description

Removes the `Interpreter` parameter from the `Value.MeteredString` method.

______

### DIFF
--- a/interpreter/interface.go
+++ b/interpreter/interface.go
@@ -148,7 +148,6 @@ func (n NoOpStringContext) MeterMemory(_ common.MemoryUsage) error {
 
 func (n NoOpStringContext) WithMutationPrevention(_ atree.ValueID, f func()) {
 	f()
-	return
 }
 
 func (n NoOpStringContext) ReadStored(_ common.Address, _ common.StorageDomain, _ StorageMapKey) Value {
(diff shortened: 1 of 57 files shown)


interpreter/interface.go
```diff
@@ -19,9 +19,12 @@
 package interpreter
```
turbolent, line 19: Should the changes to the `interpreter` package maybe also be applied on `master`?

turbolent (approved):
Nice! Should we maybe have the changes on `master` first and then merge into the compiler feature branch?

interpreter/interface.go
SupunS: I was thinking about this. Since the refactoring needed changes from both interpreter and vm side, it was easier to do the refactor on the compiler branch (same for follow-up PRs). What I thought was to once the refactoring is done, get the diff of all the interpreter/stdlib packages and apply it on master.

turbolent: That's a good idea! Yeah, it might be good to just port the changes in a single PR 👍 We can just create a patch with all commits of the compiler branch, and filter to only include changed to the interpreter package

PR #3830, opened 2025-03-28
[Compiler] Fix FT test in vm

SupunS:

## Description

Because of the way the linker works in the vm, all member functions are defined independent of the parent/enclosing value. Thus, we have to explicitly pass down information as part of the value, if they are needed to be accessed within the function.

______


turbolent (commented):
How does the interpreter handle this in comparison?

SupunS:
That's a good question. In the interpreter, these are captured in a closure: https://github.com/onflow/cadence/blob/61b9ab66ab5e8c09112936e1f8560b3605af4df9/stdlib/account.go#L3886-L3891

The downside of doing so is the function is always bound to the enclosing value. So a new function-value is needed for every parent-value. In contrast, the vm only creates one function-value for the entire vm. VM functions are static in a way.

interpreter/simplecompositevalue.go
```diff
@@ -43,6 +43,11 @@ type SimpleCompositeValue struct {
 	// This is used for distinguishing between transaction values and other composite values.
 	// TODO: maybe cleanup if there is an alternative/better way.
 	isTransaction bool
+
+	// Metadata is a property bag to carry internal data
+	// that are not visible to cadence users.
+	// TODO: any better way to pass down information?
+	metadata map[string]Value
```
turbolent, line 50: Maybe we can name this "private fields"?

turbolent (approved):
👍

interpreter/simplecompositevalue.go
SupunS: Renamed https://github.com/onflow/cadence/pull/3830/commits/7f6a2f40be22bfe147003c76d18999615e506049

PR #3831, opened 2025-03-31
[Compiler] Add remaining `Account.Storage` stdlib functions to the VM

SupunS:

## Description

Also enabled account related interpreter tests (`interpreter/account_test.go`)

______


bbq/vm/vm.go
```diff
@@ -1230,6 +1233,13 @@ func (vm *VM) StackSize() int {
 	return len(vm.stack)
 }
 
+func (vm *VM) Reset() {
+	vm.stack = nil
+	vm.locals = nil
+	vm.callstack = nil
+	vm.ipStack = nil
```
turbolent, line 1240: Small performance optimization: Maybe we can just reslice instead of assigning `nil`, to re-use the potentially allocated slices and avoid having to re-allocate later?

Maybe we should assert the stack is empty?

turbolent (approved):
Nice!

bbq/vm/vm.go
SupunS: Good idea. I'll update.

As for checking the stack size: it may not always empty. For example, if the invocation panicked in the middle of the execution, there would be leftovers in the stack.

SupunS: Updated in https://github.com/onflow/cadence/pull/3831/commits/ab34810fe4b04a978be9318e25ad01b50a9ba2ec

PR #3834, opened 2025-04-02
[Compiler] Compile simple function expressions

turbolent:
Work towards #3769

## Description

Add support for simple function expressions that do not close over variables.

- Add new instruction `NewClosure` which pushes a function value for the given function index onto the stack
- Compile function expressions by compiling function and emitting `NewClosure` instruction

______

### DIFF
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -224,9 +224,13 @@ func (c *Compiler[E, T]) addFunction(name string, parameterCount uint16) *functi
 
 	function := newFunction[E](name, parameterCount, isCompositeFunction)
 	c.functions = append(c.functions, function)
+
+	return function
+}
+
+func (c *Compiler[E, T]) targetFunction(function *function[E]) {
 	c.currentFunction = function
 	c.codeGen.SetTarget(&function.code)
-	return function
 }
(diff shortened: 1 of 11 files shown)


bbq/compiler/compiler.go
```diff
@@ -1589,9 +1591,42 @@ func (c *Compiler[_, _]) VisitBinaryExpression(expression *ast.BinaryExpression)
 	return
 }
 
-func (c *Compiler[_, _]) VisitFunctionExpression(_ *ast.FunctionExpression) (_ struct{}) {
-	// TODO
-	panic(errors.NewUnreachableError())
+func (c *Compiler[_, _]) VisitFunctionExpression(expression *ast.FunctionExpression) (_ struct{}) {
+	// TODO: desugar function expressions
```
SupunS, line 1595: I guess until they are desugar-ed, the pre/post conditions wouldn't be supported?

SupunS (approved):
Nice!

bbq/compiler/compiler.go
turbolent: Correct. I think I tried it and an assertion triggered that the function still had conditions

PR #3839, opened 2025-04-03
[Compiler] Compile simple inner functions

turbolent:
Depends on #3834
Work towards #3769

______

### DIFF
--- a/bbq/vm/test/vm_test.go
+++ b/bbq/vm/test/vm_test.go
@@ -5450,3 +5450,23 @@ func TestFunctionExpression(t *testing.T) {
 	assert.Equal(t, interpreter.NewUnmeteredIntValueFromInt64(6), actual)
 
 }
+
+func TestInnerFunction(t *testing.T) {
+
+	t.Parallel()
+
+	actual, err := compileAndInvoke(t,
+		`
+          fun test(): Int {
+              fun addOne(_ x: Int): Int {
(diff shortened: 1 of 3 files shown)


bbq/vm/test/vm_test.go
```diff
@@ -5450,3 +5450,23 @@ func TestFunctionExpression(t *testing.T) {
...
+
+func TestInnerFunction(t *testing.T) {
+
+	t.Parallel()
+
+	actual, err := compileAndInvoke(t,
+		`
+          fun test(): Int {
+              fun addOne(_ x: Int): Int {
+                  return x + 1
```
SupunS, line 5462: Maybe also add some test(s) with pre/post conditions + conditional returns. I think it should seamlessly work.
If not, we can leave it for now - I'll have a look why it's not working.

SupunS (approved):
Nice how simple the solution is 👌

bbq/vm/test/vm_test.go
turbolent: It looks like both function expressions and inner functions are not desugared yet. Testing both triggers:
```go
	// Function conditions must have been desugared to statements.
	// So there shouldn't be any condition at this point.
	if functionBlock.PreConditions != nil ||
		functionBlock.PostConditions != nil {
	}
```

PR #3840, opened 2025-04-03
[Compiler] Work towards implementing closures

turbolent:
Work towards #3769 
Depends on #3839 

## Description

Start implementing closures.

- Allow referring to outer locals in nested functions (inner functions and functions expressions), by keeping a stack of locals across all nested functions, instead of per function
- Centralize the code generation of storing to variables (`emitVariableStore`), just how it already was for loading variables (`emitVariableLoad`)
- Centralize the code generation for declaring locals
- Keep track of local and function depth. This is needed later for determining if loading/storing locals or cells
- Clean up: Remove the unused `isCompositeFunction` field of functions

______

- [ ] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #3842, opened 2025-04-08
[Compiler] Compile upvalues

turbolent:
Work towards #3769 

## Description

Continue implementing closures, by completing the compilation part. 

References to outer variables (free variables) are compiled to upvalues, like in Lua. See e.g. 
- https://www.cs.tufts.edu/~nr/cs257/archive/roberto-ierusalimschy/closures-draft.pdf
- https://www.jucs.org/jucs_11_7/the_implementation_of_lua/jucs_11_7_1159_1176_defigueiredo.html#sec5
- https://craftinginterpreters.com/closures.html

The main benefit compared to the previously planned approach similar to CPython is that we do not need an additional pass to detect free variables in the compiler or checker (last point):

> - [...] no impact on the performance of programs that do not use non-local variables.
> - [...] provide[s] acceptable performance for imperative programs, where side effects (e.g., assignments) are the norm.
> - [...] compatible with the standard execution model for procedural languages, where variables live in activation records allocated in an array-based stack.
> - [...] amenable to a one-pass compiler that generates code on the fly, without intermediate representations.

Changes of this PR:
- Revert part of #3840, parts of [7f340f8](https://github.com/onflow/cadence/pull/3840/commits/7f340f871c30c211f4e70662daf599cdcfc8aa10) and [3109d7d](https://github.com/onflow/cadence/pull/3840/commits/3109d7da06b59b34209a38b3c4f3554a783ba920), as they were unnecessary
- Add new instructions `GetUpvalue` and `SetUpvalue`
- Add upvalue array operand to `NewClosure` instruction
- Until implemented in the VM, abort VM when a closure has upvalues. i.e. this work has no impact on simple inner functions and function expressions which do not refer to outer variables

Unrelated cleanup: remove support for string operands from instruction code generator. All instructions which used to use string operands were previously refactored to refer to a constant index instead.

______

### DIFF
--- a/bbq/compiler/compiler_test.go
+++ b/bbq/compiler/compiler_test.go
@@ -4161,7 +4161,15 @@ func TestCompileFunctionExpressionOuterVariableUse(t *testing.T) {
 			opcode.InstructionSetLocal{LocalIndex: xLocalIndex},
 
 			// let inner = fun(): Int { ...
-			opcode.InstructionNewClosure{FunctionIndex: 1},
+			opcode.InstructionNewClosure{
+				FunctionIndex: 1,
+				Upvalues: []opcode.Upvalue{
+					{
+						TargetIndex: xLocalIndex,
+						IsLocal:     true,
+					},
+				},
+			},
 			opcode.InstructionTransfer{TypeIndex: 1},
 			opcode.InstructionSetLocal{LocalIndex: innerLocalIndex},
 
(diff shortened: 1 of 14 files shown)


bbq/compiler/compiler_test.go
```diff
@@ -4261,3 +4291,396 @@ func TestCompileInnerFunctionOuterVariableUse(t *testing.T) {
...
+	functions := comp.ExportFunctions()
+	require.Equal(t, len(program.Functions), len(functions))
+
+	{
+		const (
+			// aLocalIndex is the local index of the variable `a`, which is the first local variable
+			aLocalIndex = iota
+			// bLocalIndex is the local index of the variable `b`, which is the second local variable
+			bLocalIndex
+			// middleLocalIndex is the local index of the variable `b`, which is the third local variable
```
SupunS, line 4504:
```suggestion
			// middleLocalIndex is the local index of the variable `middle`, which is the third local variable
```

SupunS (approved):
👌

PR #3850, opened 2025-04-09
[Compiler] Add UUID to resources

turbolent:
Work towards #3769 

## Description

When creating a new resource, generate a UUID using the config's UUID handler, and inject it into the resource as a field.

______

### DIFF
--- a/bbq/vm/vm.go
+++ b/bbq/vm/vm.go
@@ -30,7 +30,6 @@ import (
 	"github.com/onflow/cadence/errors"
 	"github.com/onflow/cadence/interpreter"
 	"github.com/onflow/cadence/runtime"
-	"github.com/onflow/cadence/sema"
 )
 
 type VM struct {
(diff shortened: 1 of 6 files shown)


bbq/vm/vm.go
```diff
@@ -695,20 +696,57 @@ func opNew(vm *VM, ins opcode.InstructionNew) {
...
-		nil,
+		compositeFields,
 		// Newly created values are always on stack.
 		// Need to 'Transfer' if needed to be stored in an account.
 		common.ZeroAddress,
 	)
 	vm.push(value)
 }
 
+func newCompositeValueFields(vm *VM, compositeKind common.CompositeKind) (fields []interpreter.CompositeField) {
```
SupunS, line 715: Maybe only pass the `UUIDHandler` instead of the entire vm

SupunS: Just a nit: can we move this function to a separate file, say `vm/value_composite.go` and have all composite value related utilities in it (similar to `vm/value_account.go`, `vm/value_capability.go ). Then we can keep `vm.go` to only contain functions specific to the vm/vm-operations.

turbolent: Good idea moving this to a separate file and keeping `vm.go` small and focused. 👌 

The UUID handler is only needed for resources, and a future version of the function will very likely need other functionality when it will support other composite kinds. I've changed it to just pass the config for now.

SupunS (approved):

PR #3858, opened 2025-04-11
[Compiler] Support entitlements in compiler/vm

SupunS:

## Description

Entitlements in static-types is already working, since the `interpreter.Value`s and `interpreter.StaticType`s are re-used in the vm.

The only new functionality needed was to support entitlement mappings in fields/field-access, particularly for references.

While at it, also add the support for returning references for member-access/index-access, via references.

______

### DIFF
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -1341,7 +1341,7 @@ func (c *Compiler[_, _]) VisitInvocationExpression(expression *ast.InvocationExp
 			panic(errors.NewUnreachableError())
 		}
 
-		typeName := TypeName(memberInfo.AccessedType)
+		typeName := commons.TypeQualifier(memberInfo.AccessedType)
 		var funcName string
 
 		invocationType := memberInfo.Member.TypeAnnotation.Type.(*sema.FunctionType)
@@ -1419,20 +1419,6 @@ func isDynamicMethodInvocation(accessedType sema.Type) bool {
 		return true
 	default:
 		return false
-	}
-}
-
-func TypeName(typ sema.Type) string {
-	switch typ := typ.(type) {
-	case *sema.ReferenceType:
-		return TypeName(typ.Type)
-	case *sema.IntersectionType:
-		// TODO: Revisit. Probably this is not needed here?
-		return TypeName(typ.Types[0])
-	case *sema.CapabilityType:
-		return interpreter.PrimitiveStaticTypeCapability.String()
-	default:
-		return typ.QualifiedString()
 	}
 }
 
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -1465,6 +1465,8 @@ func (c *Compiler[_, _]) VisitMemberExpression(expression *ast.MemberExpression)
 	}
 
 	if memberAccessInfo.IsOptional {
+		// TODO: Complete the optional-chaining implementations.
+		//  e.g: Need a nil check, since unwrap panics on nil
 		c.codeGen.Emit(opcode.InstructionUnwrap{})
 	}
 
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -1469,6 +1469,9 @@ func (c *Compiler[_, _]) VisitMemberExpression(expression *ast.MemberExpression)
 	}
 
 	constant := c.addStringConst(expression.Identifier.Identifier)
+
+	// TODO: remove member if `isNestedResourceMove`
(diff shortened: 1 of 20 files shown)


bbq/compiler/compiler.go
```diff
@@ -1472,10 +1458,33 @@ func (c *Compiler[_, _]) loadTypeArguments(expression *ast.InvocationExpression)
...
 	c.codeGen.Emit(opcode.InstructionGetField{
 		FieldNameIndex: constant.index,
 	})
+
+	// Return a reference, if the member is accessed via a reference.
+	// This is pre-computed at the checker.
+	if memberAccessInfo.ReturnReference {
+		index := c.getOrAddType(memberAccessInfo.ResultingType)
+		c.codeGen.Emit(opcode.InstructionNewRef{TypeIndex: index})
+	}
```
SupunS, line 1484: This is the only thing needed to support the  behaviour for entitlement-mappings on references. Its nice that the type-checker does most of the work on computing the result types.

This also works in the same way in the interpreter. i.e: This now matches the `memberExpressionGetterSetter ` of the interpreter: https://github.com/onflow/cadence/blob/285de336ba10bcd929c83307e620a9c525db188a/interpreter/interpreter_expression.go#L245
Left the TODOs for the unimplemented functionalities. We would get to them once we start enabling more tests.

interpreter/entitlements_test.go
SupunS, line 1: Enabled all entitlements-related interpreter tests, except the ones that uses attachment.

bbq/compiler/compiler.go
```diff
@@ -1472,17 +1458,59 @@ func (c *Compiler[_, _]) loadTypeArguments(expression *ast.InvocationExpression)
...
+
+	// Return a reference, if the member is accessed via a reference.
+	// This is pre-computed at the checker.
+	if memberAccessInfo.ReturnReference {
+		index := c.getOrAddType(memberAccessInfo.ResultingType)
+		c.codeGen.Emit(opcode.InstructionNewRef{
+			TypeIndex:  index,
+			IsImplicit: true,
+		})
+	}
```
SupunS, line 1487: This (i.e: create a reference for the resulting value) is the only thing needed to support the  behaviour for entitlement-mappings on references. Its nice that the type-checker does most of the work on computing the result types.

This also works in the same way in the interpreter. i.e: This now matches the `memberExpressionGetterSetter ` of the interpreter: https://github.com/onflow/cadence/blob/285de336ba10bcd929c83307e620a9c525db188a/interpreter/interpreter_expression.go#L245
Left the TODOs for the unimplemented functionalities. We would get to them once we start enabling more tests.

bbq/compiler/compiler.go
```diff
@@ -1472,17 +1458,59 @@ func (c *Compiler[_, _]) loadTypeArguments(expression *ast.InvocationExpression)
...
 	c.codeGen.Emit(opcode.InstructionGetIndex{})
+
+	indexExpressionTypes, ok := c.ExtendedElaboration.IndexExpressionTypes(expression)
+	if !ok {
+		panic(errors.NewUnreachableError())
+	}
+
+	// Return a reference, if the element is accessed via a reference.
+	// This is pre-computed at the checker.
+	if indexExpressionTypes.ReturnReference {
```
SupunS, line 1506: Same here for index expressions. In reality there are no entitlements associated in this case (no way to define mappings for array elements). It just returns a reference.

bbq/compiler/compiler.go
```diff
@@ -1472,17 +1458,59 @@ func (c *Compiler[_, _]) loadTypeArguments(expression *ast.InvocationExpression)
...
 func (c *Compiler[_, _]) VisitMemberExpression(expression *ast.MemberExpression) (_ struct{}) {
 	c.compileExpression(expression.Expression)
+
+	memberAccessInfo, ok := c.ExtendedElaboration.MemberExpressionMemberAccessInfo(expression)
+	if !ok {
+		panic(errors.NewUnreachableError())
+	}
+
+	if memberAccessInfo.IsOptional {
+		c.codeGen.Emit(opcode.InstructionUnwrap{})
```
turbolent, line 1468: It looks like this implements optional chaining. The `unwrap` instructions panics on `nil`, so we probably need to emit a nil test / branch here. Feel free to just leave a TODO here and we can implement this in a follow up PR

turbolent (approved):
Great work! 👏

bbq/compiler/compiler.go
SupunS: Yeah, I added what is only needed for the tests to work, so it doesn't fully implement the optional chaining feature. It's already captured in the milestone-X. Will just leave a todo for now

PR #3876, opened 2025-04-15
[Compiler] Improve compilation of static method invocations

turbolent:
Work towards #3769 

## Description

Dynamic method invocations unwrap the receiver if it is a reference. Likewise, invocations of static methods should do the same.

Enable this by differentiating method invocations from non-method invocations, by keeping the `Invoke` instruction for non-method invocations, and introducing a new `InvokeMethodStatic` instruction. While at it, also rename `InvokeDynamic` to `InvokeMethodDynamic` to indicate that the arguments include a receiver.

To summarize, we now have three invocation instructions:
- `Invoke`: Function value on stack, no receiver
- `InvokeMethodStatic`: Function value on stack, receiver
- `InvokeMethodDynamic`: Function as operands, receiver

This also cleans up having to unwrap the receiver manually in some native function implementations.

For whatever reason, the "stack to stack" resource reference invalidation test now fails. I tried to look into it, but could not see how it was working before – the other subtests of the test where already commented out. I guess we properly have to address resource reference invalidation for all cases now.

______

### DIFF
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -1425,6 +1425,11 @@ func (c *Compiler[_, _]) VisitInvocationExpression(expression *ast.InvocationExp
 
 		// Receiver is loaded first. So 'self' is always the zero-th argument.
 		c.compileExpression(invokedExpr.Expression)
+
+		if _, ok := memberInfo.AccessedType.(*sema.ReferenceType); ok {
+			c.codeGen.Emit(opcode.InstructionDeref{})
+		}
+
 		// Load arguments
 		c.loadArguments(expression)
 
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -1390,8 +1390,8 @@ func (c *Compiler[_, _]) VisitInvocationExpression(expression *ast.InvocationExp
 		//if invocationType.IsConstructor {
 		//}
 
-		// Load arguments
-		c.loadArguments(expression)
+		// Compile arguments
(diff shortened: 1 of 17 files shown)


bbq/compiler/compiler.go
```diff
@@ -1425,6 +1425,11 @@ func (c *Compiler[_, _]) VisitInvocationExpression(expression *ast.InvocationExp
 
 		// Receiver is loaded first. So 'self' is always the zero-th argument.
 		c.compileExpression(invokedExpr.Expression)
+
+		if _, ok := memberInfo.AccessedType.(*sema.ReferenceType); ok {
+			c.codeGen.Emit(opcode.InstructionDeref{})
+		}
```
SupunS, line 1431: Would it be better to do this at runtime, like `opInvokeDynamic` used to be? Wonder if the accessed-type could be a non-reference at compile-time, but the runtime-value be a reference. For e.g: `getType()` on an `AnyStruct` variable, where the actual value is a reference.

SupunS: Also, should we handle optional here?

bbq/opcode/instructions.yml
```diff
@@ -286,9 +286,28 @@
   controlEffects:
     - call:
 
-- name: "invokeDynamic"
+- name: "invokeMethodStatic"
```
SupunS, line 289: Good idea to include "method" in the instruction to indicate that this is used for object methods 👌

SupunS (approved):
Nice!

PR #3877, opened 2025-04-16
[Compiler] Fix condition compiling

SupunS:

## Description

There were three issues: 
1. Similar to if-statements, emit-statements also could be coming from the inherited interfaces. Thus, when compiling emit-statements, the proper elaboration must be used. (✅ Fixed)
2.  If the inherited condition refers to an unqualified-type (e.g: event, `emit Event()`), then the global lookup fail for the concrete type when compiling this inherited condition. (✅ Fixed). 
    - This is because the event is a local type definition for the interface. Concrete type doesn't have the access.\
    - The fix was to re-write the inherited emit to be type-qualified. i.e: `emit Event()` → `emit Contract.Event()` 
3. Similar to events, if the inherited condition refers to a function/imported symbol that is not available to the concrete type (can happen if the interface calls `A.foo()` inside the condition and interface has an import to `A`. But the concrete type that inherits the same condition does not import `A`. Then `A.foo` cannot be found to the compiler.) (❌ Not fixed yet). 
    - This is not needed for the FT, so we can merge the PR in the current state, and fix this third issue separately.
    - Created https://github.com/onflow/cadence/issues/3878

______

### DIFF
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -1013,8 +1013,8 @@ func (c *Compiler[_, _]) VisitEmitStatement(statement *ast.EmitStatement) (_ str
 	c.withConditionExtendedElaboration(
 		statement,
 		func() {
-			eventType := c.ExtendedElaboration.EmitStatementEventType(statement)
 			c.compileExpression(statement.InvocationExpression)
+			eventType := c.ExtendedElaboration.EmitStatementEventType(statement)
 			typeIndex := c.getOrAddType(eventType)
 			c.codeGen.Emit(opcode.InstructionEmitEvent{
 				Type: typeIndex,
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -1008,12 +1008,19 @@ func (c *Compiler[_, _]) VisitForStatement(statement *ast.ForStatement) (_ struc
 }
 
 func (c *Compiler[_, _]) VisitEmitStatement(statement *ast.EmitStatement) (_ struct{}) {
-	c.compileExpression(statement.InvocationExpression)
-	eventType := c.ExtendedElaboration.EmitStatementEventType(statement)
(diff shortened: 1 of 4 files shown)


bbq/compiler/desugar.go
```diff
@@ -533,12 +533,115 @@ func (d *Desugar) desugarCondition(condition ast.Condition) ast.Statement {
...
-		return emitStmt
+
+		if !isInherited {
+			return emitStmt
+		}
+
+		// If the condition is inherited, then re-write
+		// the emit statement to be type-qualified.
+		// i.e: `emit Event()` will be re-written as `emit Contract.Event()`.
+		// Otherwise, the compiler can't find the symbol.
```
SupunS, line 544: I'm not supper happy having to re-write it this way. If there are better ways to do it, we should definitely explore those. I am out of ideas though.

bbq/compiler/compiler.go
```diff
@@ -1008,12 +1008,19 @@ func (c *Compiler[_, _]) VisitForStatement(statement *ast.ForStatement) (_ struc
...
 func (c *Compiler[_, _]) VisitEmitStatement(statement *ast.EmitStatement) (_ struct{}) {
-	c.compileExpression(statement.InvocationExpression)
-	eventType := c.ExtendedElaboration.EmitStatementEventType(statement)
-	typeIndex := c.getOrAddType(eventType)
-	c.codeGen.Emit(opcode.InstructionEmitEvent{
-		Type: typeIndex,
-	})
+	// Emit statements can be coming from inherited conditions.
+	// If so, use the corresponding elaboration.
+	c.withConditionExtendedElaboration(
```
SupunS, line 1013: This is the fix for the issue (1) mentioned in the description.

turbolent (approved):
Nice!

Unfortunately I also don't have a better idea than type qualifying, I guess it's fine for now

Issue #3878, opened 2025-04-16
Fix inherited conditions compilation


SupunS:
Compilation fails if an inherited condition refers to a function/imported-symbol that is not available to the concrete type

This can happen, say, the interface calls `A.foo()` inside the condition and interface has an import to A. But the concrete type that inherits the same condition does not import A. Then A.foo cannot be found to the compiler.

i.e:
```cadence
contract A {
    struct TestStruct {}
}
```

```cadence
import A from 0x1

contract interface B {

    resource interface VaultInterface {

        fun getBalance(): Int {
            pre { self.test(A.TestStruct()) }  // use 'A.TestStruct'
        }

        view fun test(_ a: A.TestStruct): Bool {
           return true
        }
    }
    
}
```

```cadence
import B from 0x2

contract C {

    resource Vault: B.VaultInterface {
        // The inherited condition uses 'A.TestStruct'. But here 'A' is not imported to contract 'C'.
        // So the global lookup fails, during compiling.
        fun getBalance(): Int {
            return self.balance
        }
    }
}
```

Also refer https://github.com/onflow/cadence/pull/3877 for more details.

SupunS:
Fixed in https://github.com/onflow/cadence/pull/3884

PR #3884, opened 2025-04-17
[Compiler] Fix inherited conditions compilation

SupunS:
Closes https://github.com/onflow/cadence/issues/3878

## Description

Compiling of inherited conditions so far assumed that they do not refer to local/program-only variable/symbols. This is true for all cases, except for imports. As in the example in https://github.com/onflow/cadence/issues/3878, inherited conditions can refer to imports that are only available to the original program.

Therefore, when compiling inherited conditions, always find and add the transitive dependencies (i.e: imports of the inherited program) to the current program. So the compiler can resolve these globals. It is safe to always add these, since the compiler only include these added imports in the final compiled-program only if they were ever used. Any unused imports will be discarded. 

While at it, I also refactored the test-utils in `vm/test` to extract out the utils related to compilation only, and move it to a common place. This makes it easier to re-use those utils in the compiler tests.

______


bbq/test-utils/utils.go
SupunS, line 1: Moved these functions as-is, from `vm/test/utils`

bbq/compiler/desugar.go
```diff
@@ -464,7 +464,26 @@ var panicFuncInvocationTypes = sema.InvocationExpressionTypes{
...
+	// i.e: transitive dependencies of the concrete type.
+	// Therefore, add those transitive dependencies to the current compiling program.
+	// They will only be added to the final compiled program, if those are used in the code.
+	if inheritedFrom != nil {
+		elaboration, err := d.config.ElaborationResolver(inheritedFrom.Location)
+		if err != nil {
+			panic(err)
+		}
+
+		// Order doesn't matter, since imports are only added to globals in the order of usage.
```
turbolent, line 479: Does this rely on `addImport` only having "order-independent" side-effects? It looks like that might not be the case, as it appends to `d.newImports`.

bbq/compiler/desugar.go
```diff
@@ -1028,22 +1047,22 @@ func (d *Desugar) interfaceDelegationMethodCall(
 }
 
 func (d *Desugar) addImport(location common.Location) {
-	interfaceLocation, isAddressLocation := location.(common.AddressLocation)
+	addressLocation, isAddressLocation := location.(common.AddressLocation)
```
turbolent, line 1050: Unrelated to this PR: Is there a particular reason why we only handle address imports? Other imports seem to be ignored. Is that a TODO?

turbolent (commented):
Nice!

bbq/compiler/desugar.go
SupunS: ah no, even though it appends to `d.newImports` (and eventually to import declarations of the program), when the compiler visits these import declarations, all it does is populate another map with imported symbols: https://github.com/onflow/cadence/blob/0747673c026bad007a37f70fd7893af9df2d1c1f/bbq/compiler/compiler.go#L2136

This is again order-independent.

The order is only determined by the order in which these imported symbols are used in the code. i.e: their index is set when the symbol is used in the program for the first time: https://github.com/onflow/cadence/blob/0747673c026bad007a37f70fd7893af9df2d1c1f/bbq/compiler/compiler.go#L175-L181

So the order of import declarations doesn't really matter in this case.

bbq/compiler/desugar.go
SupunS: Good question! I haven't run into to a scenario yet where the location is different. i.e: if the inherited codes are from a different program, I think it's always a deployed contract (given the locations are "resolved" by the time we reach here), and hence it's an address-location. But probably good to add some assertion here 👌

SupunS: Good question! I haven't run into to a scenario yet where the location is different. i.e: if the inherited codes are from a different program, I think it's always a deployed contract (given the locations are "resolved" by the time we reach here), and hence it's an address-location. But probably good to add some assertion here 👌

SupunS: Added in https://github.com/onflow/cadence/pull/3884/commits/0c2c4dea2264126516d461d3d4b1fc234a6bcafe

turbolent (approved):
Great work!

As mentioned in the comment, it might be good to make/keep the output of the desugaring (and compilation) deterministic, but we can also address that in a follow-up PR

bbq/compiler/desugar.go
SupunS: Makes sense. Sorted the transitive-imports before adding them here: https://github.com/onflow/cadence/pull/3884/commits/684d094d5d628a671f4298bf0a5a54cd533d78f3

turbolent (approved):
Nice! Thank you for improving the desugared imports

Issue #3863, opened 2025-04-14
Member methods as function pointers


SupunS:

SupunS:
Added in https://github.com/onflow/cadence/pull/3886

PR #3886, opened 2025-04-18
[Compiler] Support use of member-functions as function-pointers

SupunS:
Closes https://github.com/onflow/cadence/issues/3863

## Description

The changes done in this PR are:
 - Refactor `interpreter.Value` implementations to separate the function-members from the other (e.g: field) members. They were extracted out to a new method `Value.GetMethod`. First commit f66380c. (this accounts for the majority of the changes in the PR)
 - Then delegate the function member lookup to the `InvocationContext`. 
   - For e.g: `GetMember` method of all functions now calls to `InvocationContext.GetMethod`. 
   - Then the `InvocationContext` implementation decides how to return the value.
   - In the interpreter, it will return an  `interpreter.FunctionValue` (i.e: delegates to previously extracted out `Value.GetMethod`)
   - In the VM, it will lookup the function value from the globals of the relevant program, similar to how `InvokeMethodDynamic` works.
- Introduce a new `BoundFunctionPointerValue` (similar to `interpreter.BoundFunctionValue`), to hold the implicit receiver argument.

______


bbq/compiler/extended_elaboration.go
```diff
@@ -318,3 +352,17 @@ func (e *ExtendedElaboration) FunctionExpressionFunctionType(expression *ast.Fun
...
+	return e.elaboration.InterfaceTypeDeclaration(interfaceType)
+}
+
+// Elaboration returns the underlying elaboration.
+// IMPORTANT: Only use the original elaboration for type-checking use-cases.
+// It is safe to use in type-checker, since the type checker doesn't rely on
+// extra information added by the desugar phase.
+// Never use it in the compiler or desugar, as it may not include the
+// type information added during the desugar phase.
+func (e *ExtendedElaboration) Elaboration() *sema.Elaboration {
```
turbolent, line 366: Thanks for adding a great comment! Maybe "extended elaboration" can be renamed to something like "desugared elaboration", an elaboration of a desugared program, to make this more obvious

interpreter/value_accountcapabilitycontroller.go
```diff
@@ -241,6 +235,22 @@ func (v *AccountCapabilityControllerValue) GetMember(context MemberAccessibleCon
...
+
+	return context.GetMethod(v, name, locationRange)
+}
+
+func (v *AccountCapabilityControllerValue) GetMethod(
+	context MemberAccessibleContext,
+	locationRange LocationRange,
+	name string,
+) FunctionValue {
+	switch name {
```
turbolent, line 248: Not sure if this is needed, but better to have this here, like in `GetMember`, in case `GetMethod` is invoked outside of `GetMember`:

```suggestion
	// NOTE: check if controller is already deleted
	v.checkDeleted()

	switch name {
```

interpreter/simplecompositevalue.go
```diff
@@ -151,6 +151,15 @@ func (v *SimpleCompositeValue) GetMember(context MemberAccessibleContext, locati
 	return nil
 }
 
+func (v *SimpleCompositeValue) GetMethod(
+	context MemberAccessibleContext,
+	locationRange LocationRange,
+	name string,
+) FunctionValue {
+	// TODO:
```
turbolent, line 159: I guess `SimpleCompositeValue.Fields` should also be split into fields and methods, there might be some uses of `SimpleCompositeValue` where methods are stored in `Fields`

interpreter/value_pathcapability.go
```diff
@@ -180,6 +180,14 @@ func (v *PathCapabilityValue) GetMember(context MemberAccessibleContext, locatio
 	return nil
 }
 
+func (v *PathCapabilityValue) GetMethod(
+	_ MemberAccessibleContext,
+	_ LocationRange,
+	_ string,
+) FunctionValue {
+	return nil
```
turbolent, line 188: Do the `borrow` and `check` functions from `GetMember` have to be moved here?

interpreter/value_ephemeral_reference.go
```diff
@@ -138,6 +138,15 @@ func (v *EphemeralReferenceValue) GetMember(context MemberAccessibleContext, loc
 	return getMember(context, v.Value, locationRange, name)
 }
 
+func (v *EphemeralReferenceValue) GetMethod(
+	context MemberAccessibleContext,
+	locationRange LocationRange,
+	name string,
+) FunctionValue {
+	// TODO:
+	return nil
```
turbolent, line 147: Does this have to dereference like in `GetMember` and call `GetMethod` on the dereferenced value?

interpreter/value_storage_reference.go
```diff
@@ -219,6 +219,15 @@ func (v *StorageReferenceValue) GetMember(context MemberAccessibleContext, locat
 	return member
 }
 
+func (v *StorageReferenceValue) GetMethod(
+	context MemberAccessibleContext,
+	locationRange LocationRange,
+	name string,
+) FunctionValue {
+	// TODO:
+	return nil
```
turbolent, line 228: Same as for ephemeral references: Does this have to dereference like in `GetMember` and call `GetMethod` on the dereferenced value?

turbolent (commented):
Great work!

interpreter/value_ephemeral_reference.go
SupunS: Turned out that the references behave a bit differently. `GetMember` already returns the members of the dereferenced-value. But for references, the builtin-functions (such `getType`, `isInstance`) must also be returned on the dereferenced-value, but the `GetMember` implementations of other values doesn't really return those (for other values they are returned at the interpreter, when visiting member-access expression.)

So had to explicitly return the builtin functions here, for references. Added in:
https://github.com/onflow/cadence/pull/3886/commits/a08e05adbb877a96f4ec4ae8726624b5e097bec1

interpreter/value_storage_reference.go
SupunS: Same as https://github.com/onflow/cadence/pull/3886#discussion_r2054734558

interpreter/value_pathcapability.go
SupunS: I left these out since PathCapability values are deprecated. Anywa added them in https://github.com/onflow/cadence/pull/3886/commits/51ad86e22d8fd774fc1200158e426767cb57f28f for completion.

bbq/compiler/extended_elaboration.go
SupunS: Renamed in https://github.com/onflow/cadence/pull/3886/commits/44541b68ac26c329a804a59533ec706182097509

interpreter/simplecompositevalue.go
SupunS: Yes. Will open a follow-up PR, as it will need some refactoring for many builtin values, and this PR is already big.

SupunS: Added in https://github.com/onflow/cadence/pull/3890

interpreter/interpreter.go
```diff
@@ -5213,6 +5208,17 @@ func getMember(context MemberAccessibleContext, self Value, locationRange Locati
 	return result
 }
 
+func getBuiltinFunctionMember(context MemberAccessibleContext, self Value, identifier string) FunctionValue {
```
turbolent, line 5211: Nice!

turbolent (approved):
Great work!

Issue #3864, opened 2025-04-14
Pass proper type for the bounds-function-values of derived types


SupunS:
e.g:
- `Capability.borrow` function. See: https://github.com/onflow/cadence/blob/f01b80e67bbfdd6e7f9ddb9d1794e6aeb4c41d14/bbq/vm/value_capability.go#L37
- Array, Dictionary methods
- etc.

SupunS:
Completed in https://github.com/onflow/cadence/pull/3907

PR #3897, opened 2025-04-24
[Compiler] Derive the function type from the receiver for bound-function values

SupunS:
Work towards https://github.com/onflow/cadence/issues/3864

## Description

As discussed, this solution was built on top of the fact that, in VM:
- Direct function invocations (e.g: static functions such as `panic(...)`, and bound functions such as `array.append(...)`) does not need runtime typing, since they are only being invoked (but not moved around as values / not being treated as values).
- Only function pointers (e.g: `var panicFunc = panic` or `var appendFunc = array.append`) needs a runtime type, because now they can be moved around as values.

Therefore, create functions with a type-getter if the type is not pre-computable. And then only compute the type when they are being wrapped with a function pointer.

______


bbq/vm/vm.go
```diff
@@ -657,69 +672,39 @@ func opInvokeMethodDynamic(vm *VM, ins opcode.InstructionInvokeMethodDynamic) {
...
 }
 
 func invokeFunction(
 	vm *VM,
 	functionValue Value,
-	hasExplicitReceiver bool,
+	explicitArgumentsCount int,
+	arguments []Value,
 	typeArguments []bbq.StaticType,
 ) {
```
SupunS, line 706: Simplified this `invokeFunction` method, by passing the arguments rather than trying to figure out the arguments inside this function for different kinds of function-values/invocations. The invocation instruction is responsible for passing the correct arguments. Reduces the unnecessary checks.

bbq/vm/vm.go
```diff
@@ -627,25 +625,42 @@ func opGetIndex(vm *VM) {
...
+
+	explicitArgumentsCount := int(ins.ArgCount)
+	arguments := vm.peekN(explicitArgumentsCount)
+
+	// If the function is a pointer to an object-method, then the receiver is implicitly captured.
+	if boundFunction, isBoundFUnction := functionValue.(*BoundFunctionPointerValue); isBoundFUnction {
+		functionValue = boundFunction.Method
+		receiver := unwrapReceiver(vm.context, boundFunction.Receiver)
+		arguments = append([]Value{receiver}, arguments...)
+	}
```
SupunS, line 638: Now the instruction is responsible for passing the arguments to the `invokeFunction` method. See the below comment for details.

bbq/vm/vm.go
```diff
@@ -851,33 +829,20 @@ func getStringConstant(vm *VM, index uint16) string {
...
-		atree.Address{},
-		false,
-		nil,
-		nil,
+	value := vm.peek()
+	valueType := value.StaticType(context)
 
-		// TODO: Pass the correct flag here
-		false,
+	transferredValue := interpreter.TransferAndConvert(
```
SupunS, line 838: Updated to match the interpreter: Transfer should not only transfer, but also should convert if necessary.

bbq/compiler/compiler.go
```diff
@@ -1293,8 +1293,11 @@ func (c *Compiler[_, _]) VisitArrayExpression(array *ast.ArrayExpression) (_ str
 		panic(errors.NewDefaultUserError("invalid array expression"))
 	}
 
+	elementExpectedType := arrayTypes.ArrayType.ElementType(false)
+
 	for _, expression := range array.Values {
 		c.compileExpression(expression)
+		c.emitTransfer(elementExpectedType)
```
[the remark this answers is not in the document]
turbolent: Great catch!

bbq/vm/value_function.go
```diff
@@ -30,7 +30,8 @@ import (
 
 type FunctionValue interface {
 	interpreter.FunctionValue
-	GetParameterCount() int
+	HasDerivedType() bool
+	DerivedFunctionType(receiver Value, context interpreter.TypeConverter) *sema.FunctionType
```
turbolent, line 34: Maybe add comments here explaining what "derived" means. Documenting the cases with some examples, like discussed in the Impl. Sync, might help

bbq/vm/value_function.go
```diff
@@ -352,62 +398,78 @@ func (v BoundFunctionPointerValue) Accept(
...
+		v.initializeFunctionType(context)
+	}
+
+	return v.functionType
+}
+
+func (v *BoundFunctionPointerValue) initializeFunctionType(context interpreter.TypeConverter) {
+	method := v.Method
+	// The type of the native function could be either pre-known (e.g: `Integer.toBigEndianBytes()`),
+	// Or would needs to be derived based on the receiver (e.g: `[Int8].append()`).
```
turbolent, line 464: 👌

sema/type.go
```diff
@@ -2076,11 +2076,15 @@ type ArrayType interface {
 	isArrayType()
 }
 
+const ArrayTypeFirstIndexFunctionName = "firstIndex"
```
turbolent, line 2079: 👌 

Thank you for addressing this tech-debt! In a follow up PR we can maybe use these in the interpreter

turbolent (approved):
Great work! 👏

bbq/vm/value_function.go
SupunS: Good idea! Added some documentation: https://github.com/onflow/cadence/pull/3897/commits/29dbed0f9e45ee8a562ce934b9dbcd89083aa8f6

While at it, also renamed the function to `HasGenericType`, since it seems to be more intuitive.

sema/type.go
SupunS: I think it's already used in the interpreter as well.

Issue #3804, opened 2025-02-27
Compiler Milestone 4


turbolent:
### Issue to be solved

Functionality that needs to be implemented to match parity with the interpreter. For now, an unprioritized laundry list, we need to go over this after finishing #3769

### Features
- [ ] Source location info (for debugging)
- [ ] Remaining/full values
- [ ] Range value
- [ ] Dictionaries in for-loop
- [ ] All integer types
- [ ] All fixed-point types (atm, `Fix64`)
- [ ] Member functions of built-in values
- [ ] Swap statement
- [ ] Assignment / variable declaration / if-let with second value
- [ ] Force assignment
- [ ] Optional arguments

### Simple features

- [ ] String templates
- [ ] Implicit boxing
- [ ] Global variables (lazy)
- [ ] Resource invalidation (validate / finish; maybe already complete)

### Harder features

- [ ] Entitlements
- [ ] Attachments

### Optimization 

- [ ] Similar to static-types^, bypass encoding-decoding of constants
- [ ] Add ‘resource-ness’ (flag) to the static-type. Might need a state migration. (maybe not needed for the initial version)
  - Move resource-ness from instruction to static type (currently instructions have resource-nes operand)

PR #3908, opened 2025-04-29
[Compiler] Fix function invocation

SupunS:

## Description

Includes couple of fixes related to invocation compilation:
- Fix invoking a function-typed field. e.g. `foo.bar()`, where `bar` is a function-typed field, rather than a method of `foo`.
  - This should be invoked in the same way function pointers are being invoked.
  - i.e: Should use the `Invoke` instruction, and should not be invoked using `InvokeMethodStatic`/`InvokeMethodDynamic`.
- Change the expression evaluation order of invocation expressions to match that of interpreter.
  - Invoked expression must be evaluated first, and then the arguments.

An unrelated fix:
- Always transfer the elements in for-loop.

______


interpreter/misc_test.go
```diff
@@ -39,316 +39,30 @@ import (
...
 	"github.com/onflow/cadence/sema"
 	"github.com/onflow/cadence/stdlib"
+	"github.com/onflow/cadence/test_utils"
 	. "github.com/onflow/cadence/test_utils/common_utils"
 	. "github.com/onflow/cadence/test_utils/interpreter_utils"
 	. "github.com/onflow/cadence/test_utils/runtime_utils"
 	. "github.com/onflow/cadence/test_utils/sema_utils"
 )
 
-type ParseCheckAndInterpretOptions struct {
```
SupunS, line 48: These functions were replicated to `github.com/onflow/cadence/test_utils` with a hope to remove it from here, in a previous PR. So remove the implementation form there, but keeps names as-is just to avoid having to refactor all the tests (by converting them to variables with the same name) .

interpreter/misc_test.go
```diff
@@ -39,316 +39,30 @@ import (
 	"github.com/onflow/cadence/runtime"
 	"github.com/onflow/cadence/sema"
 	"github.com/onflow/cadence/stdlib"
+	"github.com/onflow/cadence/test_utils"
```
turbolent, line 42: Maybe make this a dot-import, like with the other test utils packages

turbolent (approved):
Great work! 👏

interpreter/misc_test.go
SupunS: Not possible yet, because some duplicate types are also defined in this file (`misc_test.go`) as well. Would need to refactor all tests that uses them and then we can safely remove the duplicates from here. Will do that in a separate PR.

Issue #3856, opened 2025-04-11
Compiler Milestone 3: FVM integration & execution of book-keeping functions


j1010001:
### Objective

Integrate compiler + VM into the execution stack and use it for execution of book-keeping functions (account balance check, storage limit check, fee deduction), which are under execution node control.

## Key result

Compiler + VM is integrated with execution stack, alongside existing interpreter.
Interpreter is still used for transaction execution.
Compiler + VM is used for execution of book-keeping functions.

## DACI

| Role | Assigned |
| ------ | ---------|
| Driver |  technical: @SupunS , EM: @j1010001  |
| Approver | @dete |
| Consulted| @turbolent, @Kay-Zee  |
| Informed| Flow Engineering team |

### Task breakdown

Read-only functions
- [ ] Account balance check
- [ ] Storage limit check

Read/Write function
- [ ] Fee deduction

j1010001:
Slack thread covering the profiling / performance: https://flow-foundation.slack.com/archives/C07RZKJH679/p1748296354800019

PR #3914, opened 2025-04-30
[Compiler] Improve and fix event emission

turbolent:
Work towards #3856 

## Description

Avoid the unnecessary allocation of a full composite value for an event. Also, fix event values/arguments, they were previously not stored in the event composite value.

______

### DIFF
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -1025,11 +1025,22 @@ func (c *Compiler[_, _]) VisitEmitStatement(statement *ast.EmitStatement) (_ str
 	c.withConditionElaboration(
 		statement,
 		func() {
-			c.compileExpression(statement.InvocationExpression)
+			invocationExpression := statement.InvocationExpression
+			arguments := invocationExpression.Arguments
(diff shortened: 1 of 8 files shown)


SupunS (approved):
Nice!

We could also skip compiling the constructor of the event-type definition (not needed for this PR; maybe in a follow-up PR).

turbolent:
Good idea @SupunS! Done in [d80fd35](https://github.com/onflow/cadence/pull/3914/commits/d80fd354ae4ad455624a7fea74e103e11d928a50)

turbolent:
Hmm, looks like the last commit broke something. The failures look odd. Do we maybe have to adjust linking / globals? 🤔

turbolent:
I reverted the change for now and added it to #3804

SupunS:
Ah found it! We also need to skip reserving global variables for the now-skipped event-type members. OK to push the change?

SupunS:
Opened https://github.com/onflow/cadence/pull/3915

SupunS (approved):

turbolent:
@SupunS Thank you for looking into this and finding the fix! I'll merge this and then we can merge #3915

PR #3917, opened 2025-05-01
[Compiler] Enable conditions test

SupunS:

## Description

While enabling tests, had to do various fixes:
1. Adding support for `Character` constants.
2. Support compiling locally defined contracts (contracts defined in scripts, needed for tests)
3. Support compiling locally defined events (again, needed only for tests)
4. Compile conditions in initializers
______

### DIFF
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -301,6 +301,14 @@ func (c *Compiler[_, _]) addStringConst(str string) *Constant {
 	return c.addConstant(constant.String, []byte(str))
 }
 
+func (c *Compiler[_, _]) emitCharacterConst(str string) {
+	c.emitGetConstant(c.addCharacterConst(str))
+}
+
+func (c *Compiler[_, _]) addCharacterConst(str string) *Constant {
+	return c.addConstant(constant.Character, []byte(str))
+}
+
 func (c *Compiler[_, _]) emitIntConst(i int64) {
 	c.emitGetConstant(c.addIntConst(i))
 }
@@ -415,7 +423,7 @@ func (c *Compiler[E, T]) Compile() *bbq.Program[E, T] {
 		c.compileDeclaration(declaration)
 	}
 
-	contract, _ := c.exportContract()
+	contract := c.exportContract()
 
 	compositeDeclarations := c.Program.CompositeDeclarations()
 	variableDeclarations := c.Program.VariableDeclarations()
@@ -636,41 +644,40 @@ func (c *Compiler[_, _]) exportVariables(variableDecls []*ast.VariableDeclaratio
 }
 
 func (c *Compiler[_, _]) contractType() (contractType sema.CompositeKindedType) {
-	contractDecl := c.Program.SoleContractDeclaration()
-	if contractDecl != nil {
-		contractType = c.DesugaredElaboration.CompositeDeclarationType(contractDecl)
-		return
-	}
-
-	interfaceDecl := c.Program.SoleContractInterfaceDeclaration()
-	if interfaceDecl != nil {
-		contractType = c.DesugaredElaboration.InterfaceDeclarationType(interfaceDecl)
-		return
+	// In tests, there can be contracts along with functions.
+	// TODO: Is it needed to handle having more than one contract?
+	compositeDeclarations := c.Program.CompositeDeclarations()
+	for _, declaration := range compositeDeclarations {
+		if declaration.Kind() == common.CompositeKindContract {
+			return c.DesugaredElaboration.CompositeDeclarationType(declaration)
+		}
 	}
 
 	return nil
 }
 
-func (c *Compiler[_, _]) exportContract() (*bbq.Contract, sema.CompositeKindedType) {
+func (c *Compiler[_, _]) exportContract() *bbq.Contract {
 	var location common.Location
 	var name string
 
 	contractType := c.contractType()
 	if contractType == nil {
-		return nil, nil
+		return nil
 	}
-
-	_, isInterface := contractType.(*sema.InterfaceType)
 
 	location = contractType.GetLocation()
 	name = contractType.GetIdentifier()
 
-	addressLocation := location.(common.AddressLocation)
+	var addressBytes []byte
+	addressLocation, ok := location.(common.AddressLocation)
+	if ok {
+		addressBytes = addressLocation.Address.Bytes()
+	}
+
 	return &bbq.Contract{
-		Name:        name,
-		Address:     addressLocation.Address[:],
-		IsInterface: isInterface,
-	}, contractType
+		Name:    name,
+		Address: addressBytes,
+	}
 }
 
 func (c *Compiler[_, _]) compileDeclaration(declaration ast.Declaration) {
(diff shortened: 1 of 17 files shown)


bbq/compiler/compiler.go
```diff
@@ -1855,7 +1862,17 @@ func (c *Compiler[_, _]) VisitFunctionExpression(expression *ast.FunctionExpress
...
+
+	switch stringType {
+	case sema.CharacterType:
+		c.emitCharacterConst(expression.Value)
+	case sema.StringType:
+		c.emitStringConst(expression.Value)
+	default:
+		panic(errors.NewUnreachableError())
+	}
+
```
SupunS, line 1875: (1) Adding support for Character constants

bbq/compiler/compiler.go
```diff
@@ -636,41 +644,40 @@ func (c *Compiler[_, _]) exportVariables(variableDecls []*ast.VariableDeclaratio
...
-	if interfaceDecl != nil {
-		contractType = c.DesugaredElaboration.InterfaceDeclarationType(interfaceDecl)
-		return
+	// In tests, there can be contracts along with functions.
+	// TODO: Is it needed to handle having more than one contract?
+	compositeDeclarations := c.Program.CompositeDeclarations()
+	for _, declaration := range compositeDeclarations {
+		if declaration.Kind() == common.CompositeKindContract {
+			return c.DesugaredElaboration.CompositeDeclarationType(declaration)
+		}
```
SupunS, line 653: (2) Support compiling locally defined contracts (contracts defined in scripts, needed for tests)

bbq/compiler/desugar.go
```diff
@@ -571,7 +571,17 @@ func (d *Desugar) desugarCondition(condition ast.Condition, inheritedFrom *sema.
...
-		// If the condition is inherited, then re-write
+		// Get the contract in which the event was declared.
+		eventType := d.elaboration.EmitStatementEventType(emitStmt)
+		declaredContract := declaredContractType(eventType)
+
+		// If the event is not declared in a contract, then it's a local type (e.g: in a script).
+		// Ne need to change anything in that case.
+		if declaredContract == nil {
+			return emitStmt
+		}
```
SupunS, line 582: (3) Support compiling locally defined events (again, needed only for tests)

bbq/compiler/desugar.go
```diff
@@ -765,35 +779,51 @@ func (d *Desugar) VisitCompositeDeclaration(declaration *ast.CompositeDeclaratio
...
-				functionDecl:        functionDecl,
-				rewrittenConditions: rewrittenConditions,
-				elaboration:         elaboration,
-			})
-			inheritedFunctions[name] = funcs
+		// Special functions (e.g: initializer can also have conditions)
+		specialFunctions := interfaceDecl.Members.SpecialFunctions()
+		for _, specialFunctionDecl := range specialFunctions {
+			functionDecl := specialFunctionDecl.FunctionDeclaration
+			addInheritedFunction(functionDecl, elaboration, interfaceType)
```
SupunS, line 826: (4) Compile conditions in initializers

turbolent (approved):
Great work! 👏

turbolent:
@SupunS Argh, sorry about the conflicting PR! I can help resolve this

bbq/compiler/compiler.go
```diff
@@ -636,41 +644,40 @@ func (c *Compiler[_, _]) exportVariables(variableDecls []*ast.VariableDeclaratio
...
-	contractDecl := c.Program.SoleContractDeclaration()
-	if contractDecl != nil {
-		contractType = c.DesugaredElaboration.CompositeDeclarationType(contractDecl)
-		return
-	}
-
-	interfaceDecl := c.Program.SoleContractInterfaceDeclaration()
-	if interfaceDecl != nil {
-		contractType = c.DesugaredElaboration.InterfaceDeclarationType(interfaceDecl)
-		return
```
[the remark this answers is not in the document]
SupunS: This is actually not needed; it was added during early stages. The "exported contract" is used to populate the contract value during linking. However, the interfaces are skipped already there because they don't have a value / a value is never loaded.

So this is just a cleanup to not to export the contract-interfaces at all.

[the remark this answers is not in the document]
turbolent: Nice!

bbq/vm/vm.go
```diff
@@ -165,17 +165,26 @@ func (vm *VM) pop3() (Value, Value, Value) {
 func (vm *VM) peekN(count int) []Value {
 	stackHeight := len(vm.stack)
 	startIndex := stackHeight - count
-	return vm.stack[startIndex:]
+
+	// Make a copy, since the slice can get mutated, since the stack is reused.
+	values := make([]interpreter.Value, count)
+	copy(values, vm.stack[startIndex:])
```
SupunS, line 171: @turbolent FYI

turbolent: Are we running into any problems without the copy? We should maybe inspect when `peekN`/`popN` are used. If we only need a temporary slice, then allocating a full copy might be quite expensive. 

From what I can see, the `peekN` function is only used for constructing arrays and dictionaries, which only need the slice temporarily, and there shouldn't be any stack operations during the construction of those. We might want to rename the function to make this more clear (e.g. `unstablePeekN`, `temporarilyPeekN` or something like that).

`popN` seems to be used for invocation(-like) operations (getting arguments). In those cases we should theoretically also only extract the individual arguments out of the slice and then not use it afterwards.

However, if we ever do hold onto such a temporary slice for too long, its problematic. So maybe a defensive copy is justified? 

Is this change needed for the remainder of the PR, enabling the condition tests? If not, maybe we can open a separate issue/PR for this change.

turbolent (commented):
Changes look good, great to see the tests pass!

Not quite sure about the defensive copy, we might want to address this separately, if possible.

bbq/vm/vm.go
SupunS: This was needed for the `popN` use of opEmitEvent, where the resulting slice (arguments) are not immediately used. While looking into it, I also realized after the `popN` in native function invocations, also the slice is passed as-is to the native function, and sometimes we re-enter the execution inside the native function : https://github.com/onflow/cadence/blob/5f41f6034396d8b8909a8b1c50157b88951f79fc/bbq/vm/vm.go#L1175-L1183.

Further I wasn't sure about the `peekN` - it is currently only used in constructing arrays and dictionaries, true, but it creates it with a iterator, which capture the slice in a closure - again it seems not immediately used?

So thought it would be better to do the safe thing first and then look into optimizing later. WDYT?

SupunS: I think it's always safe to not copy the arguments for compiled function invocations, since each arguments is then added/appended to the locals 👍 

For native function invocations, it depends: So far I don't think any function keep the arguments slice around as is: each argument is extracted out separately and used. So maybe it's also OK to not-copy for native function invocations too.

SupunS: Reverted the change for now https://github.com/onflow/cadence/pull/3917/commits/f7b006f53f02fb8b3d253993a1aed01030bfa14a, and only kept copying of event fields.

SupunS:
Sounds good, reverted the changes for now: https://github.com/onflow/cadence/pull/3917/commits/f7b006f53f02fb8b3d253993a1aed01030bfa14a

We can maybe discuss this further and address separately

bbq/vm/vm.go
turbolent: Sounds good to me! Good catch with noticing that it is definitely needed for events 👍 We should maybe also defensively do it for native functions, and can skip it for compiled functions

turbolent (approved):
👏

PR #3918, opened 2025-05-01
[Compiler] Fix argument compilation: transfer to parameter type, not argument type

turbolent:
Work towards #3856 

## Description

When transferring arguments, transfer to the *parameter type*, not the argument type.

Also rename the misleading field name in the elaboration.

IDK if the desugaring here is correct: https://github.com/onflow/cadence/pull/3918/files#diff-136d20e34597268ae788b3f11e31e23c55aeba34f27dd523babc79360bd08c17R1051-R1052.
Do we have to use the argument types here? How do we get them?

______

### DIFF
--- a/bbq/compiler/desugar.go
+++ b/bbq/compiler/desugar.go
@@ -465,6 +465,9 @@ var panicFuncInvocationTypes = sema.InvocationExpressionTypes{
 	ArgumentTypes: []sema.Type{
 		sema.StringType,
 	},
+	ParameterTypes: []sema.Type{
+		sema.StringType,
+	},
 }
 
 func (d *Desugar) desugarCondition(condition ast.Condition, inheritedFrom *sema.InterfaceType) ast.Statement {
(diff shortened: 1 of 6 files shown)


bbq/compiler/desugar.go
```diff
@@ -1044,8 +1047,10 @@ func (d *Desugar) interfaceDelegationMethodCall(
 	}
 
 	invocationTypes := sema.InvocationExpressionTypes{
-		ReturnType:    funcType.ReturnTypeAnnotation.Type,
-		ArgumentTypes: funcType.ParameterTypes(),
+		ReturnType: funcType.ReturnTypeAnnotation.Type,
+		// TODO:
+		ArgumentTypes:  funcType.ParameterTypes(),
```
SupunS, line 1052: Think this is safe and correct. Here we are simply delegation call, inside the synthetic-wrapper function we created. Something like:
```cadence
fun defaultFunc(a1: T1, a2: T2): R {
    return Interface.defaultFunc(a1, a2)
}
```

Where `defaultFunc` wrapper is created using the same parameter types as `Interface.defaultFunc`. So the argument types to the invocation `Interface.defaultFunc` is same as the parameter type `defaultFunc`/`Interface.defaultFunc`.

SupunS (approved):
🙏

bbq/compiler/desugar.go
SupunS: Also see: https://github.com/onflow/cadence/blob/6bc99e707156ce628141e0192e36cf240cb0c659/bbq/compiler/desugar.go#L846-L855

turbolent: Ah, I see, great! Replaced the TODO with a comment with this explanation, thanks! c07a3a54c814fbb69db1302a2d9371c33714cf38

PR #3920, opened 2025-05-02
[Compiler] Compile global variables

SupunS:

## Description

Global variables are compiled as functions (with the same name). Then when linking, the getter of the `Variable` is set to this compiled synthetic function.

```cadence
var a = "hello, world!"
```

is compiled as 

```cadence
fun a(): T {
    return "hello, world!"
}
```

and the linker set this function `a` as the getter for the global variable `a`.
______

### DIFF
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -45,15 +45,16 @@ type Compiler[E, T any] struct {
 
 	compositeTypeStack *Stack[sema.CompositeKindedType]
 
-	functions           []*function[E]
-	constants           []*Constant
-	Globals             map[string]*Global
-	importedGlobals     map[string]*Global
-	usedImportedGlobals []*Global
-	controlFlows        []controlFlow
-	currentControlFlow  *controlFlow
-	returns             []returns
-	currentReturn       *returns
+	functions             []*function[E]
+	globalVariableGetters []*function[E]
+	constants             []*Constant
+	Globals               map[string]*Global
+	importedGlobals       map[string]*Global
+	usedImportedGlobals   []*Global
+	controlFlows          []controlFlow
+	currentControlFlow    *controlFlow
+	returns               []returns
+	currentReturn         *returns
 
 	types         []sema.Type
 	compiledTypes []T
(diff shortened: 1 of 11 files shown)


bbq/compiler/compiler.go
```diff
@@ -476,7 +529,9 @@ func (c *Compiler[_, _]) reserveGlobalVars(
 	interfaceDecls []*ast.InterfaceDeclaration,
 ) {
 	for _, declaration := range variableDecls {
-		c.addGlobal(declaration.Identifier.Identifier)
+		variableName := declaration.Identifier.Identifier
+		c.addGlobal(variableName)
+		c.globalVariablesSet[variableName] = struct{}{}
```
turbolent, line 534: Could we use `Compiler.Globals` instead of introducing a new set?

bbq/vm/test/vm_test.go
```diff
@@ -7845,3 +7845,111 @@ func TestSaturatingArithmetic(t *testing.T) {
 		result,
 	)
 }
+
+func TestGlobalVariables(t *testing.T) {
```
turbolent, line 7849: Could you please add some tests that have side effects, like calling a function (e.g. logging), and assert the evaluation order?

bbq/compiler/compiler.go
turbolent, line 1: Not sure if it already is, but we need to implement this aspect of global variables like we have in the interpreter: https://github.com/onflow/cadence/blob/64af7e495b1c5feb89b4abf0e2cb7cf7b69a58de/interpreter/interpreter.go#L631-L699, i.e. all global variables are lazy, but forced evaluated at program start. 

We could do this either here in the compiler, by emitting code that loads all globals in the program initialization function, or by doing that in the VM.

bbq/vm/test/vm_test.go
SupunS: Added in https://github.com/onflow/cadence/pull/3920/commits/37ea957ecec57cc9a7c9cd39c69b64dbbef7073d

bbq/vm/test/vm_test.go
```diff
@@ -7845,3 +7845,190 @@ func TestSaturatingArithmetic(t *testing.T) {
...
+				`"invoked initializeA"`,
+				`"invoked initializeB"`,
+				`"exiting initializeA"`,
+
+				`"invoked test"`,
+			},
+			logs,
+		)
+	})
+}
```
turbolent, line 8034: Could you please also add a test case which tests that the global variables are all force-evaluated at program start? Maybe just a variant of the last test where `test` is a no-op, and the global variable has a side-effect of logging.

turbolent (approved):
Great work on the recent commits, the improvements look good to me! 👏

bbq/vm/test/vm_test.go
SupunS: Added on 5885c1d381a38fe33ea0c591aebb5865998d04a8

Issue #3838, opened 2025-04-02
Implement Iterators for String and Reference values


SupunS:
### Issue to be solved

Currently iterator is only implemented for arrays. Also need the same for String and Reference values. Iterators are used in the compiler/vm.

SupunS:
Implemented in https://github.com/onflow/cadence/pull/3923

Issue #3922, opened 2025-05-05
Compiler Milestone 5: All interpreter tests for existing functionality


turbolent:
### Issue to be solved

Switch all interpreter tests for the features _already implemented in the compiler/VM_ to be optionally run with the compiler/VM. We don't need to write new tests, we already have interpreter tests for all features.

This will give us assurance that the features are working properly, and like in the interpreter.

When tests for those implemented features depend on other features not yet implemented, consider also implementing the required feature, if reasonably low effort. For example, in the past, tests used global variables, which were not implemented yet.

Also, try to address some of the technical debt along the way, to enable a future milestones to be implementable more easily.

### Tests of implemented features

- [ ] ...

### Technical Debt

- [ ] Don't use interpreter config in VM
- [ ] Move stdlib definitions out of compiler/VM

Kay-Zee:
Currently at 60% running

Kay-Zee:
~180 remaining

turbolent:
@Kay-Zee Where is this number coming from?

turbolent:
@SupunS clarified that the remaining tests are for features yet to be implemented in the compiler/VM, and all tests for features implemented so far in the compiler/VM have been updated.

Therefore, this milestone is done.

PR #3923, opened 2025-05-05
[Compiler] Run for-statement tests with compiler

SupunS:

Also closes https://github.com/onflow/cadence/issues/3838
______

### DIFF
--- a/interpreter/for_test.go
+++ b/interpreter/for_test.go
@@ -38,7 +38,7 @@ func TestInterpretForStatement(t *testing.T) {
 
 	t.Parallel()
 
-	inter := parseCheckAndInterpret(t, `
+	inter := parseCheckAndPrepare(t, `
        fun test(): Int {
            var sum = 0
            for y in [1, 2, 3, 4] {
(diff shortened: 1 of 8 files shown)


interpreter/for_test.go
```diff
@@ -335,6 +335,7 @@ func TestInterpretForStatementCapturing(t *testing.T) {
 
 	t.Parallel()
 
+	// TODO: Use compiler (parseCheckAndPrepare)
```
SupunS, line 338: This test captures loop-variable as a closure. It seems, here it is expected to be behaved differently to other closures / everywhere else. Not sure how to implement this, so left it for now. Would be good to discuss/handle this separately.

interpreter/for_test.go
```diff
@@ -546,6 +547,7 @@ func TestInterpretEphemeralReferencesInForLoop(t *testing.T) {
 	t.Run("Mutating reference to resource array", func(t *testing.T) {
 		t.Parallel()
 
+		// TODO: Use compiler (need mutation-while-iterating validation)
```
SupunS, line 550: Mutation while iteration is not straightforward to implement. Hence left this for now, and added as an item for https://github.com/onflow/cadence/issues/3804

interpreter/for_test.go
turbolent: Yeah, this still needs to be implemented. It is already tracked as "Captured loop variables" in #3804

turbolent (approved):
Nice!

interpreter/for_test.go
SupunS: oh I see, great!

PR #3926, opened 2025-05-07
[Compiler] Run `interpreter/misc_test.go` with compiler

SupunS:

## Description

There were some more tests remaining in the file that cannot be enabled due to missing features. We can get to them in another round.

______

### DIFF
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -1734,7 +1734,7 @@ func (c *Compiler[_, _]) compileMethodInvocation(
 
 func (c *Compiler[_, _]) compileMethodInvocationArguments(
 	invokedExpr *ast.MemberExpression,
-	arguments ast. Arguments,
+	arguments ast.Arguments,
 	memberInfo sema.MemberAccessInfo,
 	invocationTypes sema.InvocationExpressionTypes,
 ) {
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -1744,6 +1744,9 @@ func (c *Compiler[_, _]) compileMethodInvocationArguments(
 	// Unwrap the target, if the member access is via optional chaining.
 	if memberInfo.IsOptional {
 		c.codeGen.Emit(opcode.InstructionUnwrap{})
+
+		// TODO: Implement the remaining parts of optional-chaining.
+		// e.g: early returning with nil, if the target is nil.
 	}
 
 	// Compile arguments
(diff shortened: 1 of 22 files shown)


test_utils/test_utils.go
```diff
@@ -89,10 +89,6 @@ func (v *VMInvokable) InitializeContract(contractName string, arguments ...inter
 func ParseCheckAndPrepare(tb testing.TB, code string, compile bool) Invokable {
 	tb.Helper()
 
-	if !compile {
-		return ParseCheckAndInterpret(tb, code)
-	}
-
```
SupunS, line 95: Redundant: `ParseCheckAndPrepareWithOptions` already does this

bbq/compiler/compiler.go
```diff
@@ -2486,6 +2506,10 @@ func (c *Compiler[_, _]) emitTransfer(targetType sema.Type) {
 
 	lastInstruction := c.codeGen.LastInstruction()
 
+	// TODO: Revisit the below logic: last instruction may not always be the
+	//  actually executed last instruction, in case where branching is present.
+	//  e.g: conditional-expression
```
turbolent, line 2511: Good catch! Could you please comment out the whole incorrect optimization?

bbq/compiler/compiler.go
```diff
@@ -1730,6 +1732,24 @@ func (c *Compiler[_, _]) compileMethodInvocation(
...
+	memberInfo sema.MemberAccessInfo,
+	invocationTypes sema.InvocationExpressionTypes,
+) {
+	// Receiver is loaded first. So 'self' is always the zero-th argument.
+	c.compileExpression(invokedExpr.Expression)
+
+	// Unwrap the target, if the member access is via optional chaining.
+	if memberInfo.IsOptional {
+		c.codeGen.Emit(opcode.InstructionUnwrap{})
+	}
```
turbolent, line 1747: Does this function implement optional chaining (e.g. `v?.m(a, b, ...)`)? Do we need to handle `v == nil` here? In that case the result is `nil`, and neither arguments nor invocation should occur

bbq/compiler/compiler.go
```diff
@@ -1730,6 +1732,24 @@ func (c *Compiler[_, _]) compileMethodInvocation(
 	}
 }
 
+func (c *Compiler[_, _]) compileMethodInvocationArguments(
```
turbolent, line 1735: Good idea to refactor this into a common function 👍

turbolent (commented):
Great work! 👏

bbq/compiler/compiler.go
SupunS: I didn't implement the full optional-chaining feature; The related tests are also still disabled. Added only this fix to get some more unrelated tests working.

I think it would be better to implement the optional-chaining feature separately. We can handle these cases there. WDYT?

bbq/compiler/compiler.go
SupunS: I checked the remaining optimizations, and they seemed to make sense. e.g: skipping transfer on constant loading happens if the target type is same as the loaded constant. That's why I left those as-is for now.

We can revisit them again, and disable only if they are actually problematic?

bbq/compiler/compiler.go
turbolent: So does this basically implement enough for the non-nil case of optional chaining to work?
We can/should definitely implement optional chaining in a separate PR if possible 👍 
Could you please leave a TODO here?

bbq/compiler/compiler.go
turbolent: Don't the other optimizations suffer from the same problem, that a jump could have targeted the transfer, and the previous instruction is not the previously executed instruction?

Maybe I'm misunderstanding why only the nil case is problematic / commented out

SupunS: Problem with the `nil` case is that it doesn't check the expected/target type. So if the target type is a super type, say `T?`, then then the other branch (if exists) can have a different value, and need tarnsfering/conversion.

On the other cases, since the target type is checked and is optimized only if the value type and the target type is the same, then even if there is another path, the value coming from that path would also be the same.

Anyways, I'll comment out the entire block if we are having doubts. 👍

SupunS: Disabled in https://github.com/onflow/cadence/pull/3926/commits/067f37ff3813dc754bacfa45c79c2031248e440a

bbq/compiler/compiler.go
SupunS: > So does this basically implement enough for the non-nil case of optional chaining to work?

Yeah, that's sounds about right.

Added a TODO for the full feature: https://github.com/onflow/cadence/pull/3926/commits/067f37ff3813dc754bacfa45c79c2031248e440a

turbolent (approved):
👌

PR #3933, opened 2025-05-08
[Compiler] Pass `vm.Context` instead of `vm.Config` to contract function handler

SupunS:
## Description

Just a cleanup: `ContractValueHandler` used to accept a `vm.Config` which only has limited configurations. Instead pass  a `vm.Context` that provides more functionality.

This also eliminates the need for creating a new context in `(*vmEnvironment).loadContractValue` method.
______

### DIFF
--- a/bbq/vm/context.go
+++ b/bbq/vm/context.go
@@ -236,7 +236,7 @@ func (c *Context) GetMethod(
 
 	qualifiedFuncName := commons.TypeQualifiedName(semaType, name)
 
-	method := c.lookupFunction(location, qualifiedFuncName)
+	method := c.GetFunction(location, qualifiedFuncName)
 	if method == nil {
 		return nil
 	}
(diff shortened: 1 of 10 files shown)


bbq/vm/context.go
```diff
@@ -246,3 +246,10 @@ func (c *Context) GetMethod(
 		method,
 	)
 }
+
+func (c *Context) GetFunction(
+	location common.Location,
+	name string,
+) interpreter.FunctionValue {
+	return c.lookupFunction(location, name)
```
turbolent, line 254: Given that it forwards the call as-is, could we just inline `lookupFunction` here and replace all callers of `lookupFunction` with `GetFunction`?

turbolent (approved):
Nice!

bbq/vm/context.go
SupunS: yah, that makes sense. Added in https://github.com/onflow/cadence/pull/3933/commits/92bcca9567b6c92f5e52776144fb7bd077750ad9

PR #3938, opened 2025-05-13
[Compiler] Run `resources-test` with the compiler

SupunS:

______


sema/check_unary_expression.go
```diff
@@ -137,6 +137,16 @@ func (checker *Checker) VisitUnaryExpression(expression *ast.UnaryExpression) Ty
...
 
+		// Store the target type in the elaboration.
+		// Target type is the expected type of the expression.
+		// However, it may not always be present (e.g: `var v2 <- v1`).
+		// In that case, use the actual type of the expression as the target type.
+		targetType := expectedType
+		if targetType == nil {
+			targetType = valueType
+		}
+		checker.Elaboration.SetMoveExpressionTypes(expression, targetType)
```
turbolent, line 148: Why do we need this in the compiler, but not in the interpreter? Isn't the unary `<-` just syntactic sugar and basically just a no-op?

SupunS: Needed that for the resource invalidation. But you are correct, maybe we might not need the full transfer, only yhe invalidation part.

SupunS: As per the offline discussion, we decided to treat the move-operator as a transfer. Added an item in https://github.com/onflow/cadence/issues/3804 for optimizing/removing any potential duplicate transfers that get's added by this change.

turbolent (approved):

PR #3939, opened 2025-05-13
[Compiler] Compile position info as a line-number table

SupunS:

## Description

Embed the position info in the bytecode as a separate line-number table. This follows a similar approach to Java's [LineNumberTable attribute](https://docs.oracle.com/javase/specs/jvms/se10/html/jvms-4.html#jvms-4.7.12).

Currently there is one table per-function (as opposed to one table per-file in Java) because the instructions are indexed per-function.

### Next up:
At the moment, to all the interpreter code that were re-used by the VM (where almost all the errors are thrown), a `LocationRange` instance must be passed in as an argument. However, creating an `LocationRange` for each operation is extremely/unacceptably expensive in VM.

Therefore, we would need to refactor all places where an `LocationRange` is passed, to accept a "location range getter" instead, so that the `LocationRange` is created only if an error occurred (i.e: only on-demand).
______

### DIFF
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -75,6 +75,9 @@ type Compiler[E, T any] struct {
 	// postConditionsIndex is the statement-index of the post-conditions for the current function.
 	postConditionsIndex int
 
+	lastChangedPosition ast.Position
+	currentPosition     ast.Position
+
 	// Cache alike for compiledTypes and constants in the pool.
 	typesInPool     map[sema.TypeID]uint16
 	constantsInPool map[constantsCacheKey]*Constant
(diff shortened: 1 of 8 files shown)


turbolent (approved):
Great work! 👏 

Thank you for already adding support for this, it's very useful during development.

Not sure, but do we need to make the current positions a "stack", e.g. by `defer`ing a restore of the previous position? For example, when compiling a statement like `a = b + c`, the compilation of the right-hand expression will overwrite the position of the assignment to the left-hand expression (RHS is compiled first)

SupunS:
@turbolent That's a good point! Handled the nested/recursive case in: https://github.com/onflow/cadence/pull/3939/commits/f75e4ac93df00c7ff34b3c5e93492e4c3e67cb44

turbolent (approved):
Nice! 👌

PR #3974, opened 2025-05-26
Ensure unbounded loops in parser eventually make progress

turbolent:
## Description

Follow up to https://github.com/onflow/cadence-internal/pull/335: Defensively check that all unbounded loops eventually make progress, i.e. the token offset changed during iteration.

Inspired by the same technique used in the Swift parser: https://github.com/swiftlang/swift-syntax/blob/29fd4ccb861e39520fea2eeb2157dd42a31d8880/Sources/SwiftParser/LoopProgressCondition.swift

The backward compatibility check [showed](https://github.com/onflow/cadence/actions/runs/15261918089/job/42921201721?pr=3974) that the current implementation of block comments is not compatible with this check: The lexer potentially emitted empty block content tokens, which are correctly skipped by the block comment parser, but this looks like a stall to the progress check. Therefore, this PR also improves the lexer to not emit empty block content tokens. A reproducer was added in https://github.com/onflow/cadence/pull/3974/files#diff-7e00f62438b5e2e571684438e3d48cead624ae24dde2fa6e37f6b0de40aa79f3R31

Also add a command to lex source, which helps with debugging the lexer.

______

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -453,9 +453,14 @@ func (p *parser) parseTrivia(options triviaOptions) (containsNewline bool, docSt
 		}
 	}()
 
-	var atEnd, insideLineDocString bool
+	var insideLineDocString bool
 
-	for !atEnd {
+	var (
+		atEnd    bool
+		progress parserProgress
+	)
+	for !atEnd && p.checkProgress(&progress) {
+
 		switch p.current.Type {
 		case lexer.TokenSpace:
 			space, ok := p.current.SpaceOrError.(lexer.Space)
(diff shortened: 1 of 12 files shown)


RZhang05 (approved):

parser/parser.go
```diff
@@ -577,6 +582,29 @@ func (p *parser) endAmbiguity() {
...
+
+// checkProgress checks that the parser has made progress since it was called last with this parserProgress.
+func (p *parser) checkProgress(progress *parserProgress) bool {
+	parserOffset := p.current.StartPos.Offset
+
+	if !progress.initialized {
+		progress.initialized = true
+		progress.offset = parserOffset
+		return true
+	}
```
SupunS, line 598: Could we do the initialization when the `parserProgress` was created? So that it doesn't need to check it here for every iteration..

SupunS (approved):
Nice idea!

parser/parser.go
turbolent: That's a good idea, yeah. I was trying to make the "zero value" of the type usable, so that we wouldn't have to ensure it is initialized (e.g. forget to call a constructor function), but that might be more efficient than checking for initialization on each call 👍

turbolent: Improved this in c5ac940fa176fa8a3ceead949d5fdd01a2bb66e8

SupunS (approved):
Nice!

Issue #3976, opened 2025-05-27
Compiler Milestone 6: More features and tech debt


turbolent:
### Features
- [ ] Optional chaining
- [ ] Enums
- [ ] Member functions of built-in values / "static" functions of built-in types
  - [ ] Integer / fixed-point types `toBigEndianBytes`, `fromBigEndianBytes`, `fromString`
  - [ ] `Address.fromString`, `Address.toBytes`, `Address.fromBytes`
- [ ] Number type min/max fields
- Analyze contracts (e.g. with regex) to determine which features to prioritize. Canidates
  - Assignment / variable declaration / if-let with second value ?
  - Force assignment ?

### Technical Debt

- [ ] Move stdlib definitions out of compiler/VM
  - https://github.com/onflow/cadence/blob/a6f1444a833a91378956cf3601aba51e01d3c476/bbq/vm/native_functions.go#L85-L155
  - cadence/bbq/vm/native_functions.go

PR #3982, opened 2025-05-28
[Compiler] Implement `Number` types' methods and static functions 

turbolent:
Work towards #3976 

## Description

- Port #3977 to the compiler feature branch
- Improve compilation of method invocations on type functions (e.g. `Int.fromString`)
- Fix invocation of native functions from compiled functions: An invocation like `opt.map(Int)` invokes the native function `Int` from compiled code. Such an invocation was running the VM, which executed unrelated code.
- Implement all `Number` types'
  - `toBigEndianBytes` methods
  - `fromBigEndianBytes` and `fromString` static functions

______

### DIFF
--- a/bbq/vm/test/vm_test.go
+++ b/bbq/vm/test/vm_test.go
@@ -8285,3 +8285,25 @@ func TestGlobalVariables(t *testing.T) {
 		)
 	})
 }
+
+func TestUserInvokesNativeFunction(t *testing.T) {
+
+	t.Parallel()
+
+	result, err := CompileAndInvoke(t,
+		`
+          fun test(): Int? {
+              let opt: UInt8? = 1
+              return opt.map(Int)
+          }
+        `,
+		"test",
+	)
+	require.NoError(t, err)
+	require.Equal(t,
+		interpreter.NewUnmeteredSomeValueNonCopying(
+			interpreter.NewUnmeteredIntValueFromInt64(1),
+		),
+		result,
+	)
+}
--- a/bbq/vm/vm.go
+++ b/bbq/vm/vm.go
@@ -89,7 +89,15 @@ func NewVM(
 	}
 
 	// Delegate the function invocations to the vm.
-	context.invokeFunction = vm.invokeExternally
+	context.invokeFunction = func(function Value, arguments []Value) (Value, error) {
+		// invokeExternally runs the VM, which is incorrect for native functions.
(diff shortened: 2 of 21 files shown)


bbq/vm/test/vm_test.go
```diff
@@ -8285,3 +8285,25 @@ func TestGlobalVariables(t *testing.T) {
...
+
+func TestUserInvokesNativeFunction(t *testing.T) {
+
+	t.Parallel()
+
+	result, err := CompileAndInvoke(t,
+		`
+          fun test(): Int? {
+              let opt: UInt8? = 1
+              return opt.map(Int)
```
turbolent, line 8297: Without the fix in https://github.com/onflow/cadence/pull/3982/files#diff-da1a470674f415c97427d68ce713bd999c2386eaf8ee9bef093c5532ddeae73bR92-R100, this would fail in a very confusing way: The function would result in a doubly-nested optional. While investigating what caused this, I eventually realized that the invocation of the native function also kept running the compiled code after it, which here happens to be a `Transfer` instruction of the return value, which would lead to an additional boxing.

bbq/vm/vm.go
```diff
@@ -89,7 +89,15 @@ func NewVM(
...
 	// Delegate the function invocations to the vm.
-	context.invokeFunction = vm.invokeExternally
+	context.invokeFunction = func(function Value, arguments []Value) (Value, error) {
+		// invokeExternally runs the VM, which is incorrect for native functions.
+		if function, ok := function.(NativeFunctionValue); ok {
+			result := function.Function(vm.context, nil, arguments...)
+			return result, nil
+		}
+
+		return vm.invokeExternally(function, arguments)
```
SupunS, line 99: Good catch!
We might have to move this to the `invokeExternally` function, because it also handles the case where native function is wrapped with a bound-function-pointer.
i.e: `vm.run()` only if the actual function is a non-native function

SupunS (approved):

bbq/vm/vm.go
turbolent: Do you maybe have an example for when that would be the case?

turbolent: Though we might also need to fix `invokeExternally`, it might be a good idea to keep this shortcut for native functions here, as it avoids a push and pop from the stack.

turbolent: Added an item to #3804

PR #3986, opened 2025-05-29
[Compiler] Improve native functions

turbolent:
Work towards #3976 

## Description

- Make native function values pointer types. They are rather large, so avoid copying
- Fix invocation of bound function pointers (like `Int.fromString`, `String.fromCharacters`, etc.) created from methods of "type functions" (like `Int`, `String`, etc.) by implementing `MemberAccessibleValue`
- Add support for fields to native functions
- Add `min`/`max` fields to type functions of `Number` types (e.g. `Int8.max`)

______


SupunS (approved):
👍

PR #4019, opened 2025-06-16
[Compiler] Implement String Templates

### LINKED ISSUE #4016: String templates

RZhang05:
Closes #4016

## Description

Introduces an instruction [`stringTemplate`](https://github.com/onflow/cadence/pull/4019/files#diff-a7e7636a4b05289e752951aa362b98876e14076283893b01d7cb34a1f80bcf8d) which implements a string template by popping two arrays off the stack `values` and `expressions` and generating a string. Logic is identical to the interpreter. 

Added compiler, vm tests. Enabled interpreter tests.

______

### DIFF
--- a/bbq/compiler/compiler_test.go
+++ b/bbq/compiler/compiler_test.go
@@ -7928,57 +7928,148 @@ func TestCompileStringTemplate(t *testing.T) {
 
 	t.Parallel()
 
-	checker, err := ParseAndCheck(t, `
-        fun test() {
-            let str = "2+2=\(2+2)"
-        }
-    `)
-	require.NoError(t, err)
+	t.Run("simple", func(t *testing.T) {
+		t.Parallel()
 
-	comp := compiler.NewInstructionCompiler(
-		interpreter.ProgramFromChecker(checker),
-		checker.Location,
-	)
-	program := comp.Compile()
+		checker, err := ParseAndCheck(t, `
+			fun test() {
+				let str = "2+2=\(2+2)"
+			}
+		`)
+		require.NoError(t, err)
 
-	functions := program.Functions
-	require.Len(t, functions, 1)
+		comp := compiler.NewInstructionCompiler(
+			interpreter.ProgramFromChecker(checker),
+			checker.Location,
+		)
+		program := comp.Compile()
 
-	assert.Equal(t,
-		[]opcode.Instruction{
-			// let str = "2+2=\(2+2)"
-			opcode.InstructionStatement{},
-			opcode.InstructionGetConstant{Constant: 0x0},
-			opcode.InstructionGetConstant{Constant: 0x1},
-			opcode.InstructionGetConstant{Constant: 0x2},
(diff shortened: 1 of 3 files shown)


bbq/compiler/compiler_test.go
```diff
@@ -7923,3 +7923,62 @@ func TestCompileSwapIndex(t *testing.T) {
...
+	assert.Equal(t,
+		[]opcode.Instruction{
+			// let str = "2+2=\(2+2)"
+			opcode.InstructionStatement{},
+			opcode.InstructionGetConstant{Constant: 0x0},
+			opcode.InstructionGetConstant{Constant: 0x1},
+			opcode.InstructionGetConstant{Constant: 0x2},
+			opcode.InstructionGetConstant{Constant: 0x2},
+			opcode.InstructionAdd{},
+			opcode.InstructionStringTemplate{ValueSize: 0x2, ExprSize: 0x1},
```
turbolent, line 7956: Is the second value, the empty string, caused by the design decision that expressions are always "sandwiched" between strings? So is there a relationship between these two sizes, `valueSize == exprSize + 1`? If so, maybe one of the instruction operands can be removed, e.g. only keep the `exprSize` operand

bbq/vm/test/vm_test.go
```diff
@@ -9131,3 +9131,37 @@ func TestGetAuthAccount(t *testing.T) {
...
+		interpreter.ProgramFromChecker(checker),
+		checker.Location,
+	)
+	program := comp.Compile()
+
+	vmConfig := &vm.Config{}
+	vmInstance := vm.NewVM(TestLocation, program, vmConfig)
+
+	result, err := vmInstance.InvokeExternally("test")
+	require.NoError(t, err)
```
turbolent, line 9157: Maybe use the `CompileAndInvoke` helper here, if possible

bbq/vm/vm.go
```diff
@@ -1377,6 +1377,34 @@ func opDeref(vm *VM) {
...
+			switch e := expressions[i].(type) {
+			case *interpreter.StringValue:
+				builder.WriteString(e.Str)
+			case interpreter.CharacterValue:
+				builder.WriteString(e.Str)
+			default:
+				builder.WriteString(e.String())
+			}
+		}
+	}
```
turbolent, line 1403: Instead of copy-and-pasting this logic from `Interpreter.VisitStringTemplateExpression`, refactor the logic out of that method into a public function, and then call it from here.

That way the behaviour of the interpreter and VM stay identical when the feature is changed / it is less likely that the behaviour will become inconsistent.

turbolent (commented):
Nice!

bbq/compiler/compiler_test.go
RZhang05: Yeah the relationship between `valueSize` and `exprSize` is enforced by the parser so we can get rid of `valueSize` 👍

SupunS (approved):
Nice!

Issue #4015, opened 2025-06-13
Compiler Milestone 10: Remaining known feature gaps (some exceptions)


j1010001:
## Goal 

Complete the remaining, _known_ missing features, except for contract deployment and attachments, that are needed for executing user transactions and scripts.

## Features

- [x] https://github.com/onflow/cadence/issues/4016 @RZhang05 
- [x] `Capability.check` @turbolent 
- [x] ~Dictionaries in for-loop~ @SupunS 
- [x] Conditions in function expressions and inner functions @SupunS 
- [x] Captured loop variables @turbolent 
- [x] Void expression @turbolent 
- [x] Move stdlib definitions out of compiler/VM @turbolent 
- [x] Stdlib types  @turbolent 
  - [x] `PublicKey`
  - [x] `HashAlgorithm`
  - [x] `SignatureAlgorithm`
  - [x] `StorageCapabilityController`
  - [x] `AccountCapabilityController`
  - [x] `DeployedContract`
  - [ ] ?
- [x] Finish Stdlib @turbolent 
  - [x] `revertibleRandom`
  - [x] `getCurrentBlock`, `getBlock`, etc.
  - [x] `Account` constructor
  - [x] `Account.Storage` 
  - [x] `Account.Capabilities`
  - [x] `Account.Keys`
  - [x] `Account.StorageCapabilities`
  - [x] `Account.AccountCapabilities`
  - [x] `Account.Inbox`
  - [x] `RLP` contract
  - [x] `BLS` contract
  - [x] `Account.Contracts.get`
  - [x] `Account.Contracts.borrow` 
  - [ ] ?
- [x] `flow.*` types
- [x] Runtime defensive checks (to be in parity with interpreter):
  - [x] Receiver invalidation for method static-invocations. i.e: `InvokeMethodStatic` (needed for `TestInterpretInvalidRecursiveTransferInExecute`). @SupunS 
  - [x] Member access type validation (needed for: `TestInterpretMemberAccessType`)
  - [x] Fix import of contract values, need reference (in linker)

PR #4020, opened 2025-06-17
[Compiler] Close upvalues for locals of for loops

turbolent:
Work towards #4015

## Description

Closures capturing variables introduced by for loops, e.g. `i` and `x` in `for i, x in ["a", "b", "c"] { ...` should keep the value at the time of the loop iteration, not the final assigned value (each loop iteration re-assigns the current index and current loop element to the variables).

Achieve this by adding a new instruction `CloseUpvalue` which closes an open upvalue for a local index. 
Emit the instruction just before a new iteration starts, i.e. when the loop jumps to the start of the loop because the end of the loop was reached, and when the user explicitly jumps to the start of the loop with a `continue` statement.

This is the same approach as in Lua / in the Lua VM.

______

### DIFF
--- a/bbq/vm/vm.go
+++ b/bbq/vm/vm.go
@@ -810,6 +810,16 @@ func opSetUpvalue(vm *VM, ins opcode.InstructionSetUpvalue) {
 	}
 }
 
+func opCloseUpvalue(vm *VM, ins opcode.InstructionCloseUpvalue) {
+	absoluteLocalsIndex := int(vm.callFrame.localsOffset) + int(ins.Local)
+	openUpvalues := vm.callFrame.openUpvalues
+	upvalue := openUpvalues[absoluteLocalsIndex]
+	if upvalue != nil {
+		upvalue.closed = vm.locals[absoluteLocalsIndex]
+		delete(openUpvalues, absoluteLocalsIndex)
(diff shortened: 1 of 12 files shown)


bbq/vm/vm.go
```diff
@@ -810,6 +810,16 @@ func opSetUpvalue(vm *VM, ins opcode.InstructionSetUpvalue) {
 	}
 }
 
+func opCloseUpvalue(vm *VM, ins opcode.InstructionCloseUpvalue) {
+	absoluteLocalsIndex := int(vm.callFrame.localsOffset) + int(ins.Local)
+	openUpvalues := vm.callFrame.openUpvalues
+	upvalue := openUpvalues[absoluteLocalsIndex]
+	if upvalue != nil {
+		upvalue.closed = vm.locals[absoluteLocalsIndex]
+		delete(openUpvalues, absoluteLocalsIndex)
```
SupunS, line 819: Not sure, could deleting the value (as opposed to just clearing / setting `nil`) affect the indexes for the rest of the up-values?

SupunS: I see. Thanks for the explanation!

SupunS (approved):

PR #4027, opened 2025-06-17
[Compiler] Compile conditions in function-expressions and inner-functions

SupunS:
Work towards https://github.com/onflow/cadence/issues/4015

## Description

Inner-functions and function-expressions are visited separately, during the compilation. 

i.e: During the initial desugar run, only the top level declarations are desugared (because we only need to deugar them). The only non-declarations that are needed to be desugared is the inner-functions/function-expression; thus, rather than trying to visit the entire AST all the way down to expressions at the desugar-phase, here it will only visit/desugar them when they are needed in the compiler.

This is done to avoid visiting all nested statements and expressions during the desugar-phase which could add an overhead, as well as because reconstructing the entire AST-branch which involves the statement/expression can be error prone. Also such re-construction of all nested statements/expressions needs to make sure elaboration is populated with proper type information for the modified AST-nodes.

Note: The alternative approach is to visit them during the desugar-phase, similar to top-level declarations. But like mentioned above, it is tedious and error-prone. A sample implementation of this alternative approach can be found at: https://github.com/onflow/cadence/compare/feature/compiler...supun/func-conditions#diff-136d20e34597268ae788b3f11e31e23c55aeba34f27dd523babc79360bd08c17R1674-R2235 

______


bbq/compiler/desugar.go
```diff
@@ -131,22 +175,72 @@ func (d *Desugar) VisitVariableDeclaration(declaration *ast.VariableDeclaration)
...
-	preConditions := d.desugarPreConditions(
+	functionType := d.elaboration.FunctionDeclarationFunctionType(declaration)
+
+	// If this is an interface-method without a body,
+	// then do not generate a function for it.
+	if d.enclosingInterfaceType != nil && !funcBlock.HasStatements() {
+		return nil
+	}
+
+	modifiedFuncBlock := d.desugarFunctionBlock(
```
SupunS, line 189: The diff here looks complex/more than what it actually is. The actual change here was just to move the function body desugaring logic to `desugarFunctionBlock` function, so that it can be re-used for functio expressions as well.

turbolent (approved):
Great work! 👏

PR #4052, opened 2025-06-25
[Compiler] Attachments #1/3

RZhang05:
Working towards #4028 

## Description

Supports compilation of four main attachment features, attach, remove, index and declaration. This sample code illustrates the supported use cases. For more information on supported use cases see enabled tests in [interpreter/attachments_test](https://github.com/onflow/cadence/blob/26ecc2ce99fbbef77209a2a15b4d11625a76399e/interpreter/attachments_test.go).

```cadence
resource R {}
attachment A for R {
	let x: Int
	init(x: Int) {
		self.x = x
	}
	fun foo(): Int { return self.x }
}
fun test(): Int {
	let r <- create R()
	let r2 <- attach A(x: 4) to <-r
	remove A from r2
	let r3 <- attach A(x: 3) to <-r2
	let i = r3[A]?.foo()!
	destroy r3
	return I
}
```

### Remove and index
These are the most straightforward, as in the interpreter they call the `RemoveTypeKey` and `GetTypeKey` functions respectively. See [interpreter/VisitIndexExpression](https://github.com/onflow/cadence/blob/master/interpreter/interpreter_expression.go#L1088) and [interpreter/VisitRemoveStatement](https://github.com/onflow/cadence/blob/master/interpreter/interpreter_statement.go#L460). For `RemoveTypeKey` it also performs the resource destruction if necessary just like the interpreter. May need to be updated to remove the base value. 

### Declaration
Also straightforward, the same as `VisitCompositeDeclaration`.  Note not refactored into common function because there are fields that do not exist on `CompositeLikeDeclaration`. 

### Attach implementation

This consists of several steps, referencing the interpreter implementation, and may need to be updated in the future mostly to support referencing `base` in the body of attachments. 

The main steps are:
1. Call attachment constructor with base reference as implicit argument. 
2. Transfer base. 
3. Call `SetTypeKey`

### Misc changes
There are other changes in vm and compiler to support the fact that `self` in attachments is a reference unlike other composites. 

### Future work

See #4028.

______


bbq/compiler/compiler.go
```diff
@@ -765,6 +789,26 @@ func (c *Compiler[_, _]) reserveFunctionGlobals(
...
+
+		members := declaration.Members
+
+		c.reserveFunctionGlobals(
+			compositeType,
+			members.SpecialFunctions(),
+			members.Functions(),
+			members.Composites(),
+			members.Interfaces(),
+			members.Attachments(),
```
SupunS, line 811: Reserving of global variables must be done in the same order as they are visited at the top-level (in `Compile()` method ) as well as when visiting as members of other composites (`compileCompositeMembers`). Given in other two places (`Compile()` and `compileCompositeMembers()`) attachments are visited after interfaces, we need to visit them in the same order here. Otherwise index lookup can get messed-up.

Unfortunately I couldn't come up with a better way to enforce this visiting-order in a non-heuristic way. If you have any ideas please feel free to add/suggest 🙏

SupunS (commented):
Great work with splitting the big feature into smaller, yet testable ones, and adding them in batches. Makes it a lot easier to review 🙏 

~I reviewed up to compiling the attachment-declarations. Will continue to review the remaining code in a second round.~

bbq/compiler/compiler_test.go
```diff
@@ -8848,6 +8848,80 @@ func TestCompileInnerFunctionConditions(t *testing.T) {
...
+				opcode.InstructionJumpIfNil{Target: 0x1e},
+				opcode.InstructionGetLocal{Local: 0x3},
+				opcode.InstructionUnwrap{},
+				opcode.InstructionGetMethod{Method: 0x7},
+				opcode.InstructionInvokeMethodStatic{TypeArgs: []uint16(nil), ArgCount: 0x0},
+				opcode.InstructionJump{Target: 0x1f},
+				opcode.InstructionNil{},
+				opcode.InstructionUnwrap{},
+				opcode.InstructionTransferAndConvert{Type: 0x3},
+				opcode.InstructionReturnValue{},
```
SupunS, line 8918: Can you please split this up a little more and add comments explaining in a bit more granular way?

bbq/opcode/instructions.yml
```diff
@@ -937,3 +937,50 @@
     push:
       - name: "string"
         type: "string"
+
+# Attachment instructions
+
+- name: "getTypeKey"
```
SupunS, line 943: Reg. naming: `getTypeIndex` might better complement the other existing instruction `getIndex` (given both are index accesses, with the syntax `A[B]`). Same for the set/remove instructions below.

bbq/compiler/compiler.go
```diff
@@ -3229,9 +3312,34 @@ func (c *Compiler[_, _]) compileEnumCaseDeclaration(
...
-	panic(errors.NewUnreachableError())
+func (c *Compiler[_, _]) VisitAttachmentDeclaration(declaration *ast.AttachmentDeclaration) (_ struct{}) {
+	// Similar to VisitCompositeDeclaration
+	// Not combined because need to access fields not accessible in CompositeLikeDeclaration
+	compositeType := c.DesugaredElaboration.CompositeDeclarationType(declaration)
+
+	c.compositeTypeStack.push(compositeType)
+	defer func() {
+		c.compositeTypeStack.pop()
+	}()
```
turbolent, line 3323: The call inside the immediately invoked invoked function is already a function call, so can be used in `defer` as-is:
```suggestion
	defer c.compositeTypeStack.pop()
```

bbq/compiler/compiler_test.go
```diff
@@ -8848,6 +8848,95 @@ func TestCompileInnerFunctionConditions(t *testing.T) {
...
+				opcode.InstructionSetLocal{Local: 0x0},
+
+				// STATEMENT: return s[A]?.foo()!
+				opcode.InstructionStatement{},
+				opcode.InstructionGetLocal{Local: 0x0},
+				// access A on s: s[A], returns attachment reference as optional
+				opcode.InstructionGetTypeIndex{Type: 0x2},
+				opcode.InstructionSetLocal{Local: 0x3},
+				opcode.InstructionGetLocal{Local: 0x3},
+				opcode.InstructionJumpIfNil{Target: 0x1e},
```
turbolent, line 8923: Here and below: Maybe use decimal offsets to make it easier to see/debug the targets

bbq/vm/vm.go
```diff
@@ -1008,6 +1008,12 @@ func loadTypeArguments(vm *VM, typeArgs []uint16) []bbq.StaticType {
 func maybeDereference(context interpreter.ValueStaticTypeContext, value Value) Value {
 	switch typedValue := value.(type) {
 	case *interpreter.EphemeralReferenceValue:
+		// Do not dereference attachments, so that the receiver is a reference as expected.
+		if val, ok := typedValue.Value.(*interpreter.CompositeValue); ok {
+			if val.Kind == common.CompositeKindAttachment {
+				return value
+			}
+		}
```
turbolent, line 1016: Given this is specific to references to receivers, let's maybe rename this function to `maybeDereferenceReceiver`, so it is not accidentally used for other purposes unrelated to receivers

turbolent (commented):
Great work! 👏

interpreter/interpreter_expression.go
```diff
@@ -1557,7 +1557,20 @@ func CreateReferenceValue(
...
+		baseAccess := baseType.SupportedEntitlements().Access()
+		auth := ConvertSemaAccessToStaticAuthorization(context, baseAccess)
 
+		return NewEphemeralReferenceValue(
+			context,
+			auth,
+			value,
+			MustSemaTypeOfValue(value, context).(*sema.CompositeType),
+			locationRange,
+		)
```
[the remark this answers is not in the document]
SupunS: I think we can create a reference-type using the `baseType`, and pass it to the `NewRef` instruction (instead of directly passing `baseType`) in the attach-expression, during compilation:
https://github.com/onflow/cadence/blob/9a8797ff0eba3a30f934f71895b2a3c233951b5a/bbq/compiler/compiler.go#L3373-L3376

bbq/vm/vm.go
```diff
@@ -1652,6 +1723,12 @@ func (vm *VM) run() {
 			opStatement(vm)
 		case opcode.InstructionTemplateString:
 			opStringTemplate(vm, ins)
+		case opcode.InstructionGetTypeIndex:
+			opGetTypeKey(vm, ins)
+		case opcode.InstructionSetTypeIndex:
+			opSetTypeKey(vm, ins)
+		case opcode.InstructionRemoveTypeIndex:
+			opRemoveTypeKey(vm, ins)
```
SupunS, line 1731: Nit: maybe rename the function names to also use the term "Index", instead of "Key", to be consistent with the instruction names.

SupunS (approved):
Nice work! 👏

turbolent (approved):
Great work! 👏

PR #4075, opened 2025-07-09
[Compiler] Use the correct environment for runtime tests

SupunS:

## Description

Noticed that some of the runtime tests were still running only using the interpreter, eventhough they were also expected to run with the VM. This was because the executors were ignoring the flag, if the environment was provided explicitly.

______

### DIFF
--- a/bbq/vm/test/interpreter_test.go
+++ b/bbq/vm/test/interpreter_test.go
@@ -641,8 +641,6 @@ func BenchmarkRuntimeFungibleTokenTransfer(b *testing.B) {
 		},
 	}
 
-	environment := runtime.NewBaseInterpreterEnvironment(runtime.Config{})
-
 	nextTransactionLocation := NewTransactionLocationGenerator()
 
 	// Deploy contract interfaces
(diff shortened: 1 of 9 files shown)


runtime/sharedstate_test.go
```diff
@@ -130,7 +130,8 @@ func TestRuntimeSharedState(t *testing.T) {
 				Interface:   runtimeInterface,
 				Location:    nextTransactionLocation(),
 				Environment: environment,
-				UseVM:       *compile,
+				// TODO:
+				//UseVM:       *compile,
```
SupunS, line 134: This test started to fail when running with the VM. Would need to look into that separately.

bbq/vm/test/interpreter_test.go
```diff
@@ -641,8 +641,6 @@ func BenchmarkRuntimeFungibleTokenTransfer(b *testing.B) {
 		},
 	}
 
-	environment := runtime.NewBaseInterpreterEnvironment(runtime.Config{})
-
```
SupunS, line 645: Avoid passing the environment explicitly., The test util will create the correct environment based on the `compile` flag.

SupunS: ah I see. Added the envrionement back: https://github.com/onflow/cadence/pull/4075/commits/8d3357059bd8be060a6127a996a1aa1b7fe97edd

Issue #4059, opened 2025-07-04
Compiler Milestone 11 - Execution of user transactions - closing feature gaps


j1010001:
## Goal (Objective)

Run Cadence execution runtime using compiler + VM executes user transactions for 4 weeks with no failures or differences in execution results (compared with interpreter with fees disabled).

**Scope**
It is unclear  how much effort will be required to close all the existing gaps to execute user transaction execution, this is discovery + implementation of found gaps.

- [ ] All runtime tests work with compiler/VM
- [ ] All flow-go tests (FVM, EN, VN, etc.) work with compiler/VM

## Key results

- Increasing number of user transactions are successfully executed using compiler + VM, eventually reaching 100% success rate over 4-week period.

## DACI
| Role | Assigned |
| ------ | ---------|
| Driver | Technical: @SupunS , EM: @j1010001  |
| Approver | @Dete |
| Consulted| @turbolent  |
| Informed| Flow engineering team |

## Functionality

- [ ] Swap of resources in index expression (move target out while evaluating index)
  - https://github.com/onflow/cadence-internal/pull/182

turbolent:
@SupunS moved the list of failing tests here instead of keeping it in #3804

turbolent:
@SupunS The resource bound function tests are in `TestRuntimeStorageReferenceBoundFunction`

SupunS:
Seems to work now: https://github.com/onflow/cadence/pull/4102

turbolent:
@onflow/cadence Adding tracing support to the VM so we can debug interpreter/VM differences

turbolent:
We've done several test runs, documented in https://www.notion.so/flowfoundation/Testing-Cadence-compiler-VM-with-interpreter-2381aee1232480a3820dd9b1fd9c8f8d, and are now able to execute hundreds of thousands of transactions matching the interpreter results, so we're considering this issue complete until we find new problems in our continued tests.

PR #4089, opened 2025-07-14
Interpreter: Evaluate and transfer invocation, array, and dictionary arguments immediately

turbolent:
Work towards #4059 

## Description

Evaluation of arguments of invocation, array, and dictionary expressions currently differs between interpreter and VM, e.g. `{a: 1, b: 2}`:

Interpreter:
  1. Evaluate all key-value pairs of the expression (eval `a`, eval `1`, eval `b`, eval `2`)
  2. Transfer (copy) all results to the stack (copy `a`, copy `1`, copy `b`, copy `2`)

VM:
  1. Evaluate each key-value pair and transfer (copy) it (eval `a`, copy `a`; eval `1`, copy `1`; ...)

Iteration order for atree ordered maps is based on the slab ID, which is generated sequentially. As a result, the value IDs and thus the iteration order for atree ordered maps (used for Cadence dictionaries and composite values) might differ.

Change the evaluation order of the interpreter to match the VM. Changing the evaluation order in the interpreter to match the VM is easier and more efficient than changing the evaluation order in the VM:
The interpreter currently needs to allocate a slice for all evaluates values before they are all transferred,
which is unnecessary if the value is transferred immediately after evaluation. Instead of introducing the same overhead to the VM, optimize the interpreter.

However:

⚠️ NOTE: This change might result in different execution results, as event field values in events potentially end up having a different (but still deterministic) iteration order.

I managed to come up with an interpreter test for invocations, as the arguments can be intercepted. For arrays and dictionaries though this is not possible. If anyone has ideas, please let me know!

______


runtime/test-export-json-deterministic.json
turbolent, line 1: Same contents, but different order

SupunS (approved):
LGTM!

PR #4090, opened 2025-07-14
[Compiler] Fix expression evaluation order of second value assignments

SupunS:

## Description

In the compiler, the sub-expressions in the middle expression of second value assignments (i.e: `b` in `var a <- b <- c`) used to evaluate twice, once during the first assignment `a <- b` and a second time around during the second value assignment `b <- c`. Instead, evaluate `b` only once and keep the result in a temp-local variable, when needed.

This is also the same order/way in which the interpreter evaluate the expressions.

______


turbolent (approved):
Great work! 👏

PR #4097, opened 2025-07-15
[Compiler] Create an implicit reference for self variable

SupunS:

## Description

When invoking a function, the `self`/receiver' reference is captured in two places:
- In the bound-function (used to validate the invocation, at the time of invocation)
- in the local-variables of the function activation (used to validate access to `self` variable inside a function)

The first one had been added in both the interpreter and in VM. However, the second one was only available in the interpreter. This PR adds the same for the VM as well.

In the interpreter, since local variables are kept as `Variables`, this implicit reference is kept in a special-type of a variable, called `SelfVariable`: https://github.com/onflow/cadence/blob/3362ceda22e3d7129fa2de3590a0cd07cacd2f59/interpreter/variable.go#L104-L107

However, in the VM, there are no local-variables, all locals are just values. Therefore, introduced a new value specific to the VM, called `ImplicitReferenceValue`, which pretty much mimics the functionality of `SelfVariable` in the interpreter.

---
 Also enable following tests:
- `TestInterpretSwapInSameArray`
- `TestResourceLossViaSelfRugPull`
- `TestRuntimeStorageReferenceBoundFunctionConfusion`
______


bbq/vm/vm.go
```diff
@@ -923,7 +930,10 @@ func opInvokeMethodDynamic(vm *VM, ins opcode.InstructionInvokeMethodDynamic) {
 	// Load arguments
 	arguments := vm.popN(int(ins.ArgCount))
 	receiver := arguments[ReceiverIndex]
-	arguments[ReceiverIndex] = maybeDereference(vm.context, receiver)
+	arguments[ReceiverIndex] = NewImplicitReferenceValue(
+		vm.context,
+		maybeDereference(vm.context, receiver),
+	)
```
SupunS, line 936: Similar to creating a `SelfVariable` in the interpreter, when putting the receiver in local-variables, here create an `ImplicitReferenceValue` (maybe rename to `ReceiverValue`?) and put that into locals.

interpreter/value_storage_reference.go
```diff
@@ -230,21 +230,11 @@ func (v *StorageReferenceValue) GetMember(context MemberAccessibleContext, locat
...
-		)
-		return boundFunction
-	}
-
-	return member
+	return context.MaybeUpdateStorageReferenceMemberReceiver(
+		v,
+		referencedValue,
+		member,
+	)
```
SupunS, line 237: The member's receiver was updated if the member is a bound function. However, interpreter and VM has two implementations for bound functions. Hence, delegate this receiver-update logic to the corresponding implementation (i.e: to the context)

bbq/vm/value_deployedcontract.go
```diff
@@ -45,7 +45,7 @@ func init() {
 				var receiver interpreter.Value
 
 				// arg[0] is the receiver. Actual arguments starts from 1.
-				receiver, args = args[ReceiverIndex], args[TypeBoundFunctionArgumentOffset:] // nolint:staticcheck
+				receiver, _ = SplitReceiverAndArgs(context, args)
```
turbolent, line 48: The idea in the original code was to still overwrite `args`, even if it is not used, just in case it will be used in the future. It might be good to keep doing it, and even consider doing this everywhere else, to avoid future bugs/misuses of the arguments slice

PR #4122, opened 2025-07-25
Improve tracing in interpreter, add support for tracing to VM

turbolent:
Work towards #4059 

## Description

- Improve tracing in the interpreter: Remove some unnecessary traces and info, remove code duplication, and refactor it so it can be used in the VM.
- Replace stubbed tracing implementation with implementation from interpreter
- Enable interpreter tracing tests to be run with VM
- Add runtime tests for tracing 

______


interpreter/value_dictionary.go
```diff
@@ -730,21 +730,6 @@ func (v *DictionaryValue) MeteredString(context ValueStringContext, seenReferenc
...
-
-		defer func() {
-			context.ReportDictionaryValueGetMemberTrace(
-				typeInfo,
-				count,
-				name,
-				time.Since(startTime),
-			)
-		}()
-	}
```
SupunS, line 747: Is this no longer needed?

SupunS (approved):
Nice cleanup!

interpreter/value_dictionary.go
turbolent: I removed it for now because it was inconsistent with the other containers traces, and didn't seem very useful (e.g. dictionaries have few member methods, and it's not very interesting if e.g. `keys` or `values` is accessed).  We can add it back for all at some point if needed.

PR #4033, opened 2025-06-19
[Interpreter] Import Aliasing

RZhang05:
Work towards #1171 

FLIP: https://github.com/onflow/flips/issues/314

Reopening of https://github.com/onflow/cadence/pull/3717, accidentally closed due to deletion of fork.

## Description

Introduces import aliases.

______

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [ ] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #4125, opened 2025-07-30
[Compiler] Import Aliasing

RZhang05:
Working towards #1171 
Depends on #4199 

## Description

This is implemented as a compiler only change (no vm, linker). The changes can be summarized as the addition of two maps, `globalAliasTable` and `globalRemoveAddressTable`, [link](https://github.com/onflow/cadence/pull/4125/files#diff-47622d5063a3f0b0ac981d4c81306aaf1394fe8dfbf64795d100e9023644e5b4R100).

`globalAliasTable` maps an alias back to its original type. If member functions did not use the original qualified typename to load the global method then there would be no need for the other map and we would not have to address qualify the typename. However, member functions do use the original qualified typename (e.g. `Foo.bar`) so we need to be able to differentiate between two identical names (but different implementations). This is why the table instead maps to an address qualified typename.

`globalRemoveAddressTable` is a simplification, instead of string parsing the address qualified typenames, they are stored along with the original qualified typename at creation time in this map. 

### Summary

In the compiler, each aliased import (e.g. `Foo1`) is mapped to an address qualified global (e.g. `A.001.Foo`). When global loads are encountered, the `globalAliasTable` is checked to retrieve the correct global (`Foo1` -> `A.001.Foo`). Global functions may be address qualified (e.g. `A.001.Foo.bar`), this is checked by looking in the `globalRemoveAddressTable` map. 

Linking only requires that we use the correct original typenames instead of the address qualified ones, so when we export imports for the vm and linker, we use the `globalRemoveAddressTable` to revert all the address qualified typenames back to their original names. 

______

### DIFF
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -3335,18 +3335,15 @@ func (c *Compiler[_, _]) VisitPragmaDeclaration(_ *ast.PragmaDeclaration) (_ str
 
 func (c *Compiler[_, _]) VisitImportDeclaration(declaration *ast.ImportDeclaration) (_ struct{}) {
 
+	// Resolve and add globals from transitive imports.
+
+	// TODO: Is it possible to reuse the resolved locations from the elaboration?
+	// resolvedLocations := c.DesugaredElaboration.elaboration.ImportDeclarationResolvedLocations(declaration)
+
 	var identifiers []ast.Identifier
-	// generate a map for aliases because when creating globals we need to associate each name with its alias
-	// each alias can be used multiple times, for example: for functions of the aliased type
-	aliases := make(map[string]string)
 	for _, imp := range declaration.Imports {
 		identifiers = append(identifiers, imp.Identifier)
-		if imp.Alias.Identifier != "" {
-			aliases[imp.Identifier.Identifier] = imp.Alias.Identifier
-		}
 	}
-
-	// Resolve and add globals from transitive imports.
 
 	resolvedLocations, err := commons.ResolveLocation(
 		c.Config.LocationHandler,
@@ -3356,6 +3353,8 @@ func (c *Compiler[_, _]) VisitImportDeclaration(declaration *ast.ImportDeclarati
 	if err != nil {
 		panic(err)
 	}
+
+	aliases := c.DesugaredElaboration.elaboration.ImportDeclarationAliases(declaration)
 
 	for _, location := range resolvedLocations {
 		c.addGlobalsFromImportedProgram(location.Location, aliases)
--- a/sema/check_import_declaration.go
+++ b/sema/check_import_declaration.go
@@ -126,7 +126,7 @@ func (checker *Checker) declareImportDeclaration(
 		}
 	}
 
-	checker.Elaboration.SetImportDeclarationsResolvedLocations(declaration, resolvedLocations)
+	checker.Elaboration.SetImportDeclarationResolvedLocations(declaration, resolvedLocations)
 	checker.Elaboration.SetImportDeclarationAliases(declaration, aliases)
 
 	for _, resolvedLocation := range resolvedLocations {
--- a/sema/elaboration.go
+++ b/sema/elaboration.go
@@ -865,7 +865,7 @@ func (e *Elaboration) ImportDeclarationResolvedLocations(declaration *ast.Import
 	return e.importDeclarationResolvedLocations[declaration]
 }
 
-func (e *Elaboration) SetImportDeclarationsResolvedLocations(
+func (e *Elaboration) SetImportDeclarationResolvedLocations(
 	declaration *ast.ImportDeclaration,
 	locations []ResolvedLocation,
 ) {


bbq/compiler/desugar.go
```diff
@@ -1647,16 +1647,6 @@ func (d *Desugar) VisitImportDeclaration(declaration *ast.ImportDeclaration) ast
 		panic(err)
 	}
 
-	for _, resolvedLocation := range resolvedLocations {
```
RZhang05, line 1650: Consider 
```
import Foo as Foo1 from 0x01
import Foo as Foo2 from 0x01
```
This does not seem like a very useful use case, but is supported currently in the interpreter. This requires that we allow locations to be imported multiple times here.

bbq/compiler/compiler.go
```diff
@@ -174,9 +190,11 @@ func newCompiler[E, T any](
...
 			elements: make([]sema.CompositeKindedType, 0),
 		},
-		codeGen:             codeGen,
-		typeGen:             typeGen,
-		postConditionsIndex: -1,
+		codeGen:                  codeGen,
+		typeGen:                  typeGen,
+		postConditionsIndex:      -1,
+		globalAliasTable:         make(map[string]string),
+		globalRemoveAddressTable: make(map[string]string),
```
turbolent, line 197: Maybe try to allocate these lazily the first time they are needed

bbq/compiler/compiler.go
```diff
@@ -3281,13 +3328,28 @@ func (c *Compiler[_, _]) VisitImportDeclaration(declaration *ast.ImportDeclarati
...
 }
 
-func (c *Compiler[_, _]) addGlobalsFromImportedProgram(location common.Location) {
+// Applies type aliases to a given string.
+// Generalized to support type qualified function names
+// e.g. Foo1.bar -> A.001.Foo.bar
+func (c *Compiler[_, _]) applyTypeAliases(name string) string {
+	parts := strings.Split(name, ".")
+	for i, part := range parts {
+		alias, ok := c.globalAliasTable[part]
```
turbolent, line 3343: It looks like the alias table is applied to all parts, but it should probably only be applied to the first part. 

For example, assume there's a contract `Foo` deployed at `0x1` which has a nested type `Foo1`:

```cadence
contract Foo {
  struct Foo1 {}
}
```

and assume a program that imports it:

```cadence
import Foo as Foo1 from 0x1

// ... use Foo1.Foo1

```

then the function should rewrite `Foo1.Foo1` should only be rewritten to `A.0…1.Foo.Foo1`. However, it looks like also the nested types/parts get rewritten, which would produce e.g. incorrect `A.0…1.Foo.A.0…1.Foo`

bbq/compiler/compiler.go
```diff
@@ -3298,21 +3360,53 @@ func (c *Compiler[_, _]) addGlobalsFromImportedProgram(location common.Location)
...
+		name := contract.Name
+		alias, ok := aliases[name]
+		if ok {
+			// we want a table pointing from the alias -> address qualifier (Foo1 -> A.0001.Foo)
+			addressQualifiedName := string(location.TypeID(nil, name))
+			c.globalAliasTable[alias] = addressQualifiedName
+
+			c.globalRemoveAddressTable[addressQualifiedName] = name
+			name = addressQualifiedName
+		}
```
turbolent, line 3372: This is repeated below for imported variables and functions. Consider refactoring this into a reusable function

bbq/compiler/desugar.go
turbolent: Good catch! Maybe we can also simplify this and remove support for this in the implementation of import aliasing in the interpreter, e.g. by rejecting this statically in the type checker

turbolent (commented):
Great work!

bbq/compiler/compiler.go
RZhang05: Changed and added a test in https://github.com/onflow/cadence/pull/4125/commits/5ccf03de1723eb5639a03081bfa753cb19c139c0

SupunS (approved):
LGTM!

turbolent:
Might be good to address https://github.com/onflow/cadence/pull/4125#discussion_r2277533108 in #4033, and then this can be merged, see  https://github.com/onflow/cadence/issues/1171#issuecomment-3238468998

bbq/compiler/desugar.go
RZhang05: >  remove support for this in the implementation of import aliasing in the interpreter, e.g. by rejecting this statically in the type checker

Is this suggesting we specifically don't allow a type to be aliased multiple times?

RZhang05: Done in #4199

bbq/compiler/compiler.go
```diff
@@ -3273,8 +3336,14 @@ func (c *Compiler[_, _]) VisitPragmaDeclaration(_ *ast.PragmaDeclaration) (_ str
...
 	var identifiers []ast.Identifier
+	// generate a map for aliases because when creating globals we need to associate each name with its alias
+	// each alias can be used multiple times, for example: for functions of the aliased type
+	aliases := make(map[string]string)
 	for _, imp := range declaration.Imports {
 		identifiers = append(identifiers, imp.Identifier)
+		if imp.Alias.Identifier != "" {
+			aliases[imp.Identifier.Identifier] = imp.Alias.Identifier
+		}
 	}
```
turbolent, line 3347: With #4199 the alias map is now available from the elaboration so we can avoid re-allocating here, I'll push up a change

turbolent: Done in [1857524](https://github.com/onflow/cadence/pull/4125/commits/18575245d49ea5b597515fe76662fb3cc0c58d4b)

turbolent (approved):
Great work! 👏

bbq/compiler/compiler.go
```diff
@@ -3272,13 +3335,16 @@ func (c *Compiler[_, _]) VisitPragmaDeclaration(_ *ast.PragmaDeclaration) (_ str
 
 func (c *Compiler[_, _]) VisitImportDeclaration(declaration *ast.ImportDeclaration) (_ struct{}) {
 
+	// Resolve and add globals from transitive imports.
+
+	// TODO: Is it possible to reuse the resolved locations from the elaboration?
+	// resolvedLocations := c.DesugaredElaboration.elaboration.ImportDeclarationResolvedLocations(declaration)
```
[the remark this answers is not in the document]
turbolent: Resolving for now so we can merge this PR, added as tech debt to #3804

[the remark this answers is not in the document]
turbolent: Yeah, we are not going to use the compiler yet, so it's OK to keep this as a TODO and already merge this.

Issue #4062, opened 2025-07-07
Improve Cadence Errors to Support LLM Efficiency


briandoyle81:
### Issue to be solved

Cadence errors, are not sufficiently informative for LLMs to solve linter errors on their own, particularly for pre-1.0 keywords.  IE:

```
pub fun foo() {}
```

Results in an error: `'pub' is no longer a valid access keyword`

This isn't sufficient information for an LLM to solve on it's own, and I don't think it includes a link to more information.

### Suggested Solution

Improve the error messages, link to the documentation, and ideally provide suggested fixes, so that tools like Cursor and other agent-based tools can automatically fix these errors. Agents are even able to read the referenced link and include that information in their solutions.

#### High priority

Primary focus is on on improving _existing_ errors, in particular their messages and secondary messages, and adding links to the documentation, as well as migration notes for errors related to pre-Cadence 1.0.

The documentation links and migration notes should be surfaced in the language server (https://github.com/onflow/cadence-tools/issues/484).

#### Lower priority

- Add suggested fixes for errors where possible
- Show migration note and documentation link in "pretty"/"console" output

### Out of scope

- Adding error codes
- Adding additional errors

### Errors for features that lack documentation

- [x] Member access / dot operator / optional chaining
- [x] Pragmas
- [x] Generics (e.g. type instantiations, function invocation type arguments, etc.) 
- [x] Complete list of usages and rules for all punctuation characters: (), {}, [], <>, `, ., |, (comma)
- [x] Whitespace rules
- [x] Improve documentation for casting/downcasting
- [ ] Explicit list/statement of where resources can be declared (https://cadence-lang.org/docs/language/resources)
- [x] Fixed point literal scale/range (sema.InvalidFixedPointLiteralRangeError, sema.InvalidFixedPointLiteralScaleError)
- [ ] Documentation of how contract-level fields work (https://cadence-lang.org/docs/language/contracts)
- [ ] Comparison rules, what can and cannot be compared (https://cadence-lang.org/docs/language/operators/comparison-operators)
- [x] Indexing

Some features / errors are internal / non-user facing and should not be documented because they are not part of the "official" language:

- `native`/`static` modifiers

turbolent:
All PRs are merged now. 

@briandoyle81 Could you please do another pass over the errors in https://github.com/onflow/cadence/blob/master/parser/errors.go and https://github.com/onflow/cadence/blob/master/sema/errors.go ?

briandoyle81:
[...]

PR #4146, opened 2025-08-13
Improve declaration parsing errors

turbolent:
Work towards #4062 

## Description

Continue improving errors, mainly focusing on parsing of declarations. Refactor most of the remaining missing/wrong token errors into dedicated error types, in particular mismatched open/closing parens/braces.

While at it, clean up the declaration parsing tests. Apologies the PR got so large.

______

### DIFF
--- a/cmd/errors/errors.go
+++ b/cmd/errors/errors.go
@@ -23,6 +23,18 @@ func generateErrors() []namedError {
 				DestructorRange: placeholderRange,
 			},
 		},
+		{"parser.DeclarationMissingClosingBraceError",
+			&parser.DeclarationMissingClosingBraceError{
+				Kind:     placeholderDeclarationKind,
+				GotToken: placeholderToken,
+			},
+		},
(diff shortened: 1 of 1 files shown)


parser/errors.go
```diff
@@ -2601,6 +2639,117 @@ func (*InvalidPubModifierError) DocumentationLink() string {
 	return "https://cadence-lang.org/docs/language/access-control"
 }
 
+// MissingPubClosingParenError is reported when a pub(set) modifier is missing a closing parenthesis.
+type MissingPubClosingParenError struct {
```
RZhang05, line 2643: Just a thought, since `pub` is not supported it makes sense to me that the only errors that should be reported if `pub` or `priv` is encountered are the ones that explicitly say that its not supported like `PubAccessError`. We could still attempt to parse the rest of the program to find more errors, but not report other syntax errors like this related to `pub` or `priv` parsing. 🤔

PR #4159, opened 2025-08-16
Add and improve suggested fixes

turbolent:
⚠️ Depends on #4158
Work towards #4062 

## Description

Add more suggested fixes for errors. Improve suggested fixes by moving tokens left, so they are "attached" to the previous token.

______

### DIFF
--- a/ast/position.go
+++ b/ast/position.go
@@ -135,12 +135,30 @@ func NewUnmeteredRange(startPos, endPos Position) Range {
 	}
 }
 
-func (e Range) StartPosition() Position {
-	return e.StartPos
+func (r Range) StartPosition() Position {
+	return r.StartPos
 }
 
-func (e Range) EndPosition(common.MemoryGauge) Position {
-	return e.EndPos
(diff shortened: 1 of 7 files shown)


ast/position.go
```diff
@@ -135,12 +135,30 @@ func NewUnmeteredRange(startPos, endPos Position) Range {
...
+func (r Range) EndPosition(common.MemoryGauge) Position {
+	return r.EndPos
+}
+
+// AttachLeft moves the range left until it reaches a non-whitespace character.
+func (r Range) AttachLeft(code string) Range {
+	var offset int
+	for i := r.StartPos.Offset - 1; i >= 0; i-- {
+		switch code[i] {
+		case ' ', '\t':
```
RZhang05, line 151: Maybe we can use `unicode.IsSpace` here, for cases with new lines like:

```cadence
let v = {"hi", true
let c = test
```

would currently suggest

```cadence
let v = {"hi", true
}let c = test
```

turbolent: Good idea! Opened a follow-up PR to avoid conflicts: #4161

Issue #4058, opened 2025-07-04
Metering optimization


j1010001:
**Objective**: Metering (computation and memory), avoid call to FVM each time (see https://github.com/onflow/cadence/issues/3992)

VM environment directly metering VS going through runtime.Interface / ExternalInterface is ~25%!

**Key Results**:
- [ ] Discuss design + Estimate the effort with execution team
- [ ] write benchmark
- [ ] measure the improvement with metering stubbed-out
- [ ] refactor the metering (move to Cadence)

## DACI
| Role | Assigned |
| ------ | ---------|
| Driver | Technical: @turbolent  , EM: @j1010001|
| Approver | @dete  |
| Consulted| @SupunS, @RZhang05  |
| Informed| Flow Protocol engineering |

PR #4166, opened 2025-08-21
Refactor memory metering

turbolent:
Work towards #4058

Moved from https://github.com/onflow/cadence-internal/pull/344

## Description

Refactor the metering of memory: Instead of calling into the embedder (e.g. FVM) via the runtime interface, have it optionally pass a memory gauge in the context. This should hopefully reduce the overhead for memory metering a bit.

______

### DIFF
--- a/common/metering.go
+++ b/common/metering.go
@@ -35,6 +35,14 @@ type MemoryGauge interface {
 	MeterMemory(usage MemoryUsage) error
 }
 
+type FunctionMemoryGauge func(usage MemoryUsage) error
+
+var _ MemoryGauge = FunctionMemoryGauge(nil)
+
+func (f FunctionMemoryGauge) MeterMemory(usage MemoryUsage) error {
+	return f(usage)
+}
+
 type ComputationUsage struct {
 	Kind      ComputationKind
 	Intensity uint64
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -75,8 +75,8 @@ type ExternalNonError struct {
 	Recovered any
 }
 
-func NewExternalNonError(recovered error) ExternalError {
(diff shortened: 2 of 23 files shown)


RZhang05 (approved):

common/metering.go
```diff
@@ -52,36 +52,6 @@ type ComputationGauge interface {
 	MeterComputation(usage ComputationUsage) error
 }
 
-// Gauge combines a memory and computation gauge.
-// Metering-sites can use this combined interface
-// if they need to do both memory and computation metering.
-type Gauge interface {
```
turbolent, line 50: Removed the combined gauge, because it adds another layer of indirection, and thus overhead: Both memory gauge and computation gauge are optional, so the implementations of the methods `MeterMemory` and `ComputationGauge` would actually have to check for `nil` each time they are called.

common/metering.go
```diff
@@ -52,28 +60,6 @@ type Gauge interface {
 	ComputationGauge
 }
 
-// CombinedGauge is a Gauge, i.e. it allows metering both memory and computation,
-// by delegating to an independent memory gauge and independent computation gauge.
-//
-// It is mostly just a convenience/helper type, which is useful for e.g. tests,
-// where one has an independent (test) memory gauge, and an independent (test) computation gauge.
-type CombinedGauge struct {
```
turbolent, line 60: Remove `CombinedGauge`, as there are two issues:
- It is incomplete: Users may assume that the passed in gauges may be `nil`, but the current implementation does not allow this. Both `MeterMemory` and `MeterComputation` need to be implemented, and they need to check for `nil`
- Such a correct implementation adds another indirection, i.e. overhead

Issue #4171, opened 2025-05-12
Pretty printing crashing on incomplete AST nodes


jribbink:
When I create an incomplete dictionary type, such as:

```cadence
access(all) contract Foo {
    access(all) let dict: @{String: }
    init() {}
}
```

The language server will crash with the following error:

```
goroutine 45 [running]:
github.com/onflow/cadence/ast.(*DictionaryType).Doc(0x1400132b4f0)
	github.com/onflow/cadence@v1.3.4/ast/type.go:412 +0x40
github.com/onflow/cadence/ast.Prettier({0x1058f7000, 0x1400132b4f0})
	github.com/onflow/cadence@v1.3.4/ast/prettier.go:29 +0x40
github.com/onflow/cadence/ast.(*DictionaryType).String(0x140013bfd40?)
	github.com/onflow/cadence@v1.3.4/ast/type.go:398 +0x28
github.com/onflow/cadence-tools/lint.(*cadenceV1Analyzer).AnalyzeAll.(*cadenceV1Analyzer).analyzeRemovedTypeIdentifiers.func2(0x1400138fc50)
	github.com/onflow/cadence-tools/lint@v1.2.1/cadence_v1_analyzer.go:171 +0x34
github.com/onflow/cadence-tools/lint.(*cadenceV1Analyzer).inspectTypeAnnotations.func1(0x140000a6c48?)
	github.com/onflow/cadence-tools/lint@v1.2.1/cadence_v1_analyzer.go:279 +0x90
github.com/onflow/cadence-tools/lint.(*cadenceV1Analyzer).inspectTypeAnnotations.func3({0x10591a580?, 0x14001eae5a0})
	github.com/onflow/cadence-tools/lint@v1.2.1/cadence_v1_analyzer.go:309 +0x174
github.com/onflow/cadence/ast.(*Inspector).Preorder(0x14002009728, {0x140000a6d08?, 0x7f?, 0x102cbb7b8?}, 0x140000a6cf0)
	github.com/onflow/cadence@v1.3.4/ast/inspector.go:87 +0x8c
github.com/onflow/cadence-tools/lint.(*cadenceV1Analyzer).inspectTypeAnnotations(0x14002009728?, 0x1400024ddd8?)
	github.com/onflow/cadence-tools/lint@v1.2.1/cadence_v1_analyzer.go:295 +0x110
github.com/onflow/cadence-tools/lint.(*cadenceV1Analyzer).analyzeRemovedTypeIdentifiers(...)
[...]

Issue #529, opened 2021-01-14
Make the parser resilient


SupunS:
### Issue To Be Solved

Currently the parser terminates upon certain syntax errors with a panic. Once this happens, syntax analyzing stops for the remainder of the program, and the user would not get any further syntax errors, even if there's any.

### (Optional): Suggest A Solution

Solution is to replace the places where the parser panics, with a non-greedy/non-exhaustive recovery approach, and continue the parsing.

Such recovered nodes may leave to regression errors at the semantic analyzer. Hence any recovered node should not be analyzed by the semantic analyzer.

PR #4173, opened 2025-08-26
Handle partial AST nodes when pretty printing

turbolent:
Work towards #4171 

## Description

The parser may produce partial AST nodes, i.e. parts may be `nil`, increasingly so as we make the parser more resilient (#529). 

Partial AST nodes are for example printed in tooling like the linter.

Start handling `nil` in all type AST nodes and some others. Produce an empty document in such cases.
While at it, also add missing tests.

Best viewed [with whitespace changes disabled](https://github.com/onflow/cadence/pull/4173/files?w=1)

______

### DIFF
--- a/ast/attachment.go
+++ b/ast/attachment.go
@@ -123,16 +123,9 @@ func (d *AttachmentDeclaration) Doc() prettier.Doc {
 	var doc prettier.Concat
 
 	if d.Access != AccessNotSpecified {
-		var accessDoc prettier.Doc
-		if d.Access == nil {
-			accessDoc = prettier.Text("")
-		} else {
-			accessDoc = d.Access.Doc()
-		}
-
 		doc = append(
 			doc,
-			accessDoc,
+			docOrEmpty(d.Access),
 			prettier.HardLine{},
 		)
 	}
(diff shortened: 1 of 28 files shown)


ast/attachment.go
```diff
@@ -123,9 +123,16 @@ func (d *AttachmentDeclaration) Doc() prettier.Doc {
 	var doc prettier.Concat
 
 	if d.Access != AccessNotSpecified {
+		var accessDoc prettier.Doc
+		if d.Access == nil {
+			accessDoc = prettier.Text("")
+		} else {
+			accessDoc = d.Access.Doc()
+		}
```
SupunS, line 131: Instead of doing the if-`nil` check everywhere manually, could we move the `nil` check to a utility method, and call that to get the `prettier.Doc` of a given element.
```suggestion
		accessDoc := getDoc(d.Access)
```
where 
```Go
func getDoc(element ast.Element)
	if element == nil {
		return prettier.Text("")
	}

	return element.Doc()
}
```
Might have to update/introduce the interface to have a `Doc` method that is available for all AST nodes/elements.

SupunS (approved):

ast/attachment.go
SupunS: nice!

Issue #4172, opened 2025-08-25
Unusual rounding behaviour (Fix64, Int128, Int256, Int)


dete:
### Current Behavior

Numeric types in Cadence have no documented rounding behaviour when performing division operations (or fractional multiplication operations), but appear to truncate in most cases:

```Cadence
Int64(5) / Int64(2) == Int64(2)
Int64(-5) / Int64(2) == Int64(-2)
Int64(-5) / Int64(-2) == Int64(2)
UFix64(0.00000005) / UFix64(2.0) == UFix64(0.00000002)
UFix64(0.00000005) * UFix64(0.5) == UFix64(0.00000002)
```

However, on negative inputs, the types `Fix64`, `Int128`, `Int256`, and `Int` provide unexpected results:

```Cadence
Int128(5) / Int128(2) == Int128(2)
Int128(-5) / Int128(2) == Int128(-3)
Int128(-5) / Int128(-2) == Int128(-2)
Int128(-5) / Int128(-2) == Int128(3)
Fix64(0.00000005) / Fix64(2.0) == Fix64(0.00000002)
Fix64(-0.00000005) / Fix64(2.0) == Fix64(-0.00000003)
Fix64(-0.00000005) / Fix64(-2.0) == Fix64(0.00000003)
```

This is quite strange rounding behaviour, which appears to be due to using the `Int.Div()` function in Go's `math.big` library which (somewhat surprisingly!) [uses Euclidean division](https://pkg.go.dev/math/big#Int.Div). I would expect truncation in all cases.

### Expected Behavior

I would propose that we change the rounding behaviour for negative values to match `Int64` and truncate in all cases, this should be easily achieved by switching to the `Int.Quo()` function (which seems to do what we want).

We should also explicitly document that all numeric types truncate (or "round toward zero") on division (and multiplication for fixed-point types).

### Steps To Reproduce

This script:
```Cadence
access(all) fun main(): [Int64] {
    return [Int64(5) / Int64(2), Int64(-5) / Int64(2), Int64(5) / Int64(-2), Int64(-5) / Int64(-2)]
}
```

returns `[2, -2, -2, 2]`

While this script (identical to the above, except for the types):
```Cadence
access(all) fun main(): [Int128] {
    return [Int128(5) / Int128(2), Int128(-5) / Int128(2), Int128(5) / Int128(-2), Int128(-5) / Int128(-2)]
}
```

returns `[2, -3, -2, 3]`

### Environment

```markdown
[...]

PR #4184, opened 2025-08-27
Use truncation instead of rounding for `big.Int` division

SupunS:
Fix https://github.com/onflow/cadence/issues/4172

Implements FLIP: https://github.com/onflow/flips/pull/344

## Description

Use truncation division (`big.Int.Quo()`) instead of rounding, for number types backed by `big.Int`, to be consistent with the rest of the number types that are backed by Go's primitive values.

______

### DIFF
--- a/interpreter/fixedpoint_test.go
+++ b/interpreter/fixedpoint_test.go
@@ -24,8 +24,10 @@ import (
 	"math/big"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/onflow/cadence/ast"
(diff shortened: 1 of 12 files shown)


RZhang05 (approved):

SupunS:
Going to merge this PR, so that we can get a preview release (and then a CLI build) out for people to test the new behaviour.

Merging the implementation doesn't mean this change is final, but rather is still very much subject to the outcome of the FLIP https://github.com/onflow/flips/pull/344. If the sentiment is not positive on the FLIP, we can revert the PR. It's much more easier to revert, than having to manage separate branches for all the downstream dependencies needed the CLI preview release.

PR #4199, opened 2025-09-03
Improve import aliasing

turbolent:
Work towards #1171

## Description

Improve import aliasing.

Most importantly reject:
- Re-importing the same element multiple times (e.g. `import a as b, a as c from 0x1`, `import a from 0x1; import a as b from 0x1`, etc)
- Re-use of aliases (e.g. `import a as b, c as b from 0x1`)

Reject these both when the imports are part of a single import declaration or if they are split across multiple declarations.

While at it, also improve and add missing functionality:
- Improve the import declaration AST element by storing the identifiers and imports together. Keeping the identifiers for the aliases allows us to refer to their positions, and keeping them in a slice instead of a map allows us to detect duplicates
- Take aliases into account when pretty printing import declarations
- Take import aliases into account in contract update validation

______

### DIFF
--- a/sema/check_import_declaration.go
+++ b/sema/check_import_declaration.go
@@ -406,10 +406,7 @@ func (checker *Checker) importElements(
 			if !ok {
 				continue
 			}
-			alias, ok := aliases[name]
-			if ok {
-				name = alias
-			}
+
 			elements.Set(name, element)
 			found[identifier] = true
 			explicitlyImported[name] = identifier
@@ -448,6 +445,10 @@ func (checker *Checker) importElements(
 				}
 			}
 
+			alias, ok := aliases[name]
+			if ok {
+				name = alias
+			}
 			_, err := valueActivations.declare(variableDeclaration{
 				identifier: name,
 				ty:         elementType,
--- a/sema/import_test.go
+++ b/sema/import_test.go
@@ -1041,9 +1041,10 @@ func TestCheckImportAlias(t *testing.T) {
 			},
 		)
 
-		errs := RequireCheckerErrors(t, err, 1)
+		errs := RequireCheckerErrors(t, err, 2)
 
 		assert.IsType(t, &sema.DuplicateImportAliasError{}, errs[0])
+		assert.IsType(t, &sema.RedeclarationError{}, errs[1])
 	})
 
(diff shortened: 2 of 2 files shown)


sema/import_test.go
```diff
@@ -925,19 +917,27 @@ func TestCheckImportAlias(t *testing.T) {
...
 		)
 
-		require.NoError(t, err)
+		errs := RequireCheckerErrors(t, err, 1)
+
+		assert.IsType(t, &sema.DuplicateImportError{}, errs[0])
 	})
 
-	t.Run("invalid duplicate aliases", func(t *testing.T) {
+	t.Run("invalid multiple alias of same element, same import", func(t *testing.T) {
```
turbolent, line 925: Here and below: The diff looks like a change, but really is just adding another variant of the subtest above: same import declaration instead of multiple import declarations.

RZhang05 (approved):
Nice, thank you! #4125 will have to be updated on top of these changes

RZhang05:
> @RZhang05 I the conflicts are not too bad, I can help with that

Should be done already 👍

sema/check_import_declaration.go
```diff
@@ -45,15 +45,55 @@ func (checker *Checker) VisitImportDeclaration(declaration *ast.ImportDeclaratio
...
+
+				if _, ok := usedAliases[aliasName]; ok {
+					checker.report(
+						&DuplicateImportAliasError{
+							Alias: imp.Alias,
+						},
+					)
+				} else {
+					usedAliases[aliasName] = struct{}{}
+				}
```
SupunS, line 93: Do we report errors for cases where there's an alias with the same name as the imported identifier (not a duplicate alias).
e.g:
```cadence
import Foo as Bar from 0x1
import Bar from 0x2
```

turbolent: I just checked and though this is handled, the single declaration variant `import Foo as Bar, Bar from 0x1` isn't handled. I'll add tests and fix it

turbolent: Fixed in [5a003ac](https://github.com/onflow/cadence/pull/4199/commits/5a003ac19096f012d1d94d01fe1f0baa9215cc0b). Problem was that the alias was applied to early

SupunS (approved):
👌

PR #4208, opened 2025-09-09
[Compiler] Always transfer on move operator. Remove redundent transfers

SupunS:

## Description

Reverts https://github.com/onflow/cadence/pull/4193, and instead, avoid the redundant transfer operation, when unary move-operator is used in expressions. This is required for the runtime resource invalidation defensive check to work in the VM.

This is done by:
 - Always transfer the value during assignments. e.g: `var x <- r`.
   - Because `<-` is not treated as a unary operator during assignments, in the AST.
   - Therefore always transfer.
   - Same to other operations that are similar to 'assignments'. e.g: Swap, etc.
 - During non-assignments, skip if the value is a resource.
   - Because in non-assignments, resource values are always used with a unary-move operator. And it does the transfer.
   - e.g: `return <- r`, `foo(<-)`.

______


SupunS:
Ran the [test-10](https://www.notion.so/flowfoundation/Cadence-compiler-VM-manual-comparison-Test-10-2741aee1232480d4bdedf124c7147e14) (100 blocks from for block ID `5628d11b47d52120aa388d225895825f879f809628a4f23dcad5dffed72cfdac`), and found no mis-matches.
```
INF Compared 100 blocks: 100 matched, 0 mismatched
INF Compared 531 transactions: 531 matched, 0 mismatched
```

Will run some more blocks to see if there are any, but seems all good at this point.

SupunS:
[Test-9](https://www.notion.so/flowfoundation/Cadence-compiler-VM-manual-comparison-Test-9-2731aee1232480f1aaebf6f3b8cec5f3) also looks good:
```
INF Compared 100 blocks: 100 matched, 0 mismatched
INF Compared 547 transactions: 547 matched, 0 mismatched
```

SupunS:
Yeah, we already have an item for this unification in the Milestone X; Was planing to update the interpreter once we are all OK with this approach, and have tested enough to make sure it produces the same results. Now that this seems to work, I will open a follow-up PR.

turbolent:
> Yeah, we already have an item for this unification in the Milestone X; Was planing to update the interpreter once we are all OK with this approach, and have tested enough to make sure it produces the same results. Now that this seems to work, I will open a follow-up PR.

Sounds good! I'll give this another review today 👍

bbq/compiler/compiler_test.go
```diff
@@ -3491,13 +3491,14 @@ func TestCompileFunctionConditions(t *testing.T) {
 			[]opcode.Instruction{
 				opcode.InstructionStatement{},
 
-				// $_result = x
+				// $_result <- x
 				opcode.InstructionStatement{},
 				opcode.InstructionGetLocal{Local: xIndex},
+				opcode.InstructionTransfer{},
```
turbolent, line 3497: I'm not quite sure, but it looks like this is adding an additional transfer to the program, not moving it to a different place. Given it is a resource and it is "stack-to-stack", the transfer is probably not causing a difference in slab allocations here, but is that also the case for non-resources and non-stack-to-stack transfers? 🤔

turbolent (approved):
Great work! 👏

bbq/compiler/compiler_test.go
SupunS: Good point! The transfer here is correct, but it should not transfer again when returning the `$_result`. Fixed it in https://github.com/onflow/cadence/pull/4208/commits/dbe7851ae5a6427195a9414f54b08a893bceaf56

PR #4220, opened 2025-09-15
[Compiler] Simplify imported globals linking

SupunS:

## Description

Previously, linker used to link imported globals based on the `Imports` section of the `bbq.Program`. This is because `bbq.Program.Globals` didn't have enough information needed for linking. However, with https://github.com/onflow/cadence/pull/4211,  `Globals` now has more information that is sufficient to link the globals.

- Therefore, switch to linking the imported globals also using `bbq.Program.Globals` instead of `bbq.Program.Imports`.
- This further eliminates the implicit dependency of order of imports between compiler and the VM.
- Also eliminates the need for recursive linking, as the `bbq.Program.Globals` already includes all symbols/values needed to execute the program, including transitive dependencies.
- After this, `bbq.Program.Imports` are only used by the compiler.

______

### DIFF
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -48,16 +48,16 @@ type Compiler[E, T any] struct {
 
 	compositeTypeStack *Stack[sema.CompositeKindedType]
 
-	functions          []*function[E]
-	globalVariables    []*globalVariable[E]
-	constants          []*Constant
-	Globals            map[string]bbq.Global
-	globalImports      *activations.Activation[GlobalImport]
-	importedGlobals    []bbq.Global
-	controlFlows       []controlFlow
-	currentControlFlow *controlFlow
-	returns            []returns
-	currentReturn      *returns
+	functions           []*function[E]
+	globalVariables     []*globalVariable[E]
+	constants           []*Constant
+	Globals             map[string]bbq.Global
+	importedGlobals     *activations.Activation[GlobalImport]
+	usedImportedGlobals []bbq.Global
+	controlFlows        []controlFlow
+	currentControlFlow  *controlFlow
+	returns             []returns
+	currentReturn       *returns
 
 	types         []sema.Type
 	compiledTypes []T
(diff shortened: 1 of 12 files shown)


bbq/compiler/compiler.go
```diff
@@ -159,7 +159,19 @@ func NewInstructionCompilerWithConfig(
 
 type GlobalImport struct {
 	Location common.Location
-	Name     string
+	// Need to maintain both "qualified" and "unqualified" names for a given import,
+	// because when type-aliasing is used, imported name becomes qualified.
+	// However, the same import must use the unqualified name when linking.
+	// TODO: We can simplify this by always using qualified names for all imports.
```
SupunS, line 165: I wasn't sure the reason to not use address-qualified names for all imports (only the imports with aliases use the address-qualified names). @RZhang05 do you maybe remember? Was it more of an optimization (for e.g: serialization)?

bbq/compiler/compiler.go
```diff
@@ -3439,12 +3455,17 @@ func (c *Compiler[_, _]) applyTypeAliases(name string) string {
...
+
 	parts := strings.Split(name, ".")
 	part := parts[0]
 	alias, ok := c.globalAliasTable[part]
-	if ok {
-		parts[0] = alias
+
+	// If no alias found, then return the original name.
+	if !ok {
+		return name
```
SupunS, line 3465: Unrelated change: avoid the string-joining, if no aliases are found.

bbq/vm/linker.go
```diff
@@ -30,9 +30,6 @@ import (
 )
 
 type LinkedGlobals struct {
-	// context shared by the globals in the program.
-	executable *ExecutableProgram
```
SupunS, line 34: was no longer used

bbq/vm/linker.go
```diff
@@ -46,83 +43,10 @@ func LinkGlobals(
...
-
-		var indexedGlobals *activations.Activation[Variable]
-
-		if importLocation == nil {
-			if context.BuiltinGlobalsProvider == nil {
-				indexedGlobals = DefaultBuiltinGlobals()
-			} else {
-				indexedGlobals = context.BuiltinGlobalsProvider(location)
-			}
-		} else {
```
SupunS, line 62: Moved this import-linking logic to a separate function below

bbq/vm/linker.go
```diff
@@ -176,24 +100,107 @@ func LinkGlobals(
...
+			)
+			globals[i] = importedGlobal
+
+			// Don't need to add to the `indexedGlobals`, since, like the below comment says,
+			// importer/caller doesn't need to know globals of nested imports
 		}
 	}
 
-	// add imported globals to the end of the globals
-	globals = append(globals, importedGlobals...)
```
SupunS, line 185: This was still causing to have some implicit dependency of order, between the compiler and VM. e.g: As mentioned in the PR description imports **were** linked through `bbq.Program.Import`, but needed them to be in the same order as `bbq.Program.Globals`, because the indexes uses the later.

Instead, directly link using the `bbq.Program.Global` (above), now that it has sufficient information.

bbq/compiler/compiler.go
RZhang05: I think @turbolent left this comment, but I think it was just easier to leave the previous functionality as is (and not address qualify imports without aliases).

SupunS: Nice! yeah, we can do this as a cleanup later on. I'll make a note of it.

PR #4225, opened 2025-09-18
Fix `invokeMethodDynamic`

turbolent:
## Description

Copy the arguments slice before getting the member. Getting the member might result in implicit invocations, which mutate the stack.

I still need to find a minimal reproducer and add a test case for it.

As the TODO for `opInvokeMethodDynamic` points out, we might want to just get rid of it / replace it with a combination of other instructions.
 
______

### DIFF
--- a/bbq/vm/vm.go
+++ b/bbq/vm/vm.go
@@ -917,8 +917,12 @@ func opInvokeMethodDynamic(vm *VM, ins opcode.InstructionInvokeDynamic) {
 	// Load type arguments
 	typeArguments := loadTypeArguments(vm, ins.TypeArgs)
 
-	// Load arguments
-	arguments := vm.popN(int(ins.ArgCount))
+	// Load arguments.
+	// Make a copy, since the slice can get mutated, since the stack is reused.
+	// This can happen in case getting the member below triggers execution of other functions.
+	argCount := int(ins.ArgCount)
+	arguments := make([]Value, argCount)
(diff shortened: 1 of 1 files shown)


turbolent:
> What was the reason for this reversion https://github.com/onflow/cadence/commit/f7b006f53f02fb8b3d253993a1aed01030bfa14a?

I guess it was an optimization to avoid always defensively copying, which adds quite some overhead. I actually don't think it's needed for the event emission 🤔

bbq/vm/vm.go
```diff
@@ -917,8 +917,12 @@ func opInvokeMethodDynamic(vm *VM, ins opcode.InstructionInvokeDynamic) {
 	// Load type arguments
 	typeArguments := loadTypeArguments(vm, ins.TypeArgs)
 
-	// Load arguments
-	arguments := vm.popN(int(ins.ArgCount))
+	// Load arguments.
+	// Make a copy, since the slice can get mutated, since the stack is reused.
+	// This can happen in case getting the member below triggers execution of other functions.
```
SupunS: Good catch! I wonder how a member-access/ a nested function call could change the existing levels of the stack though 🤔, because the nested function calls should only use the stack from current index/position onwards. The bug / problem could actually be there, rather than here.

SupunS (approved):
LGTM! 
I'll try to address the TODO for spitting this to two methods. The last time I tried, there was some test failures, but did't investigate further.

PR #4248, opened 2025-10-01
[Compiler] Unify natives.

### LINKED ISSUE #4216: Unify interpreter and VM host/native functions
(currently lots of duplication, mostly argument extraction)

RZhang05:
Closes #4216

## Description

Targets uses of `NewBoundHostFunctionValue`, `NewStaticHostFunctionValue`, `NewUnmeteredStaticHostFunctionValue`, `NewNativeFunctionValue` and `NewNativeFunctionValueWithDerivedType`, replacing them with the unified version. The goal is to eventually replace all uses of these and then we can rename the unified functions to remove the `unified` part. 

This PR changes the majority of native/host functions with exceptions for testing.

For reviewing, I would suggest looking at both the original function implementation (both vm and interpreter) and the new unified one, doesn't show well in the diff so probably look at the diff and in the codebase at the same time. Most of the changes are quite simple (argument extraction), some modify the logic to accommodate differences between interpreter and vm. 

______

### DIFF
--- a/stdlib/account.go
+++ b/stdlib/account.go
@@ -264,7 +264,13 @@ func UnifiedGetAuthAccountFunction(handler AccountHandler) interpreter.UnifiedNa
 				panic(errors.NewUnreachableError())
 			}
 
-			return NewAccountReferenceValue(context, handler, accountAddress, referenceType.Authorization, locationRange)
+			return NewAccountReferenceValue(
+				context,
+				handler,
+				accountAddress,
+				referenceType.Authorization,
+				locationRange,
+			)
 		},
 	)
 }
@@ -2670,7 +2676,13 @@ func UnifiedGetAccountFunction(handler AccountHandler) interpreter.UnifiedNative
 			args ...interpreter.Value,
 		) interpreter.Value {
 			accountAddress := interpreter.AssertValueOfType[interpreter.AddressValue](args[0])
-			return NewAccountReferenceValue(context, handler, accountAddress, interpreter.UnauthorizedAccess, locationRange)
+			return NewAccountReferenceValue(
+				context,
+				handler,
+				accountAddress,
+				interpreter.UnauthorizedAccess,
+				locationRange,
+			)
 		},
 	)
 }
@@ -5063,7 +5075,12 @@ func unifiedAccountAccountCapabilitiesGetControllersFunction(
 	) interpreter.Value {
 		address := interpreter.GetAddress(receiver, addressPointer)
 
-		return accountAccountCapabilitiesGetControllers(context, address, locationRange, handler)
+		return accountAccountCapabilitiesGetControllers(
+			context,
+			address,
+			locationRange,
+			handler,
+		)
 	}
 }
 
@@ -5177,7 +5194,13 @@ func unifiedAccountAccountCapabilitiesForEachControllerFunction(
 
 		address := interpreter.GetAddress(receiver, addressPointer)
 
-		return AccountCapabilitiesForEachController(context, address, functionValue, locationRange, handler)
+		return AccountCapabilitiesForEachController(
+			context,
+			address,
+			functionValue,
+			locationRange,
+			handler,
+		)
 	}
 }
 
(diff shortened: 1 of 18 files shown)


interpreter/unified_function.go
```diff
@@ -52,12 +52,21 @@ func NewInterpreterTypeParameterGetter(memoryGauge common.MemoryGauge, typeParam
...
-	return ConvertSemaToStaticType(i.memoryGauge, i.NextSema())
+	semaType := i.NextSema()
+	if semaType == nil {
+		return nil
+	}
+	return ConvertSemaToStaticType(i.memoryGauge, semaType)
 }
 
 func (i *InterpreterTypeParameterGetter) NextSema() sema.Type {
 	current := i.typeParameterTypes.Oldest()
```
turbolent, line 63: Good catch that this needs to handle `nil` / optional type parameters/arguments!

While looking at this I also realized that the function isn't keeping state right now: It should store the next pair, initialized to `Oldest`, and call `Next` on the pair to obtain the next one (if any), just like the loops currently iterating over the ordered set (`for pair := orderedMap.Oldest(); pair != nil; pair = pair.Next() { ... }`)

turbolent: Right, I think there are no functions with multiple type parameters at the moment. 

[5d96d9f](https://github.com/onflow/cadence/pull/4248/commits/5d96d9f492987335fd1981b0e40fdb0b2e863ed8)
looks good! 👍 

To make sure that we will handle functions with multiple type parameters correctly once we add any, could you please add an interpreter test (which also works with the compiler by using `parseCheckAndPrepare`) which injects a native function with multiple type parameters, and the function iterates over all type arguments.

RZhang05: Added in https://github.com/onflow/cadence/pull/4248/commits/48b901a5d64fe46f6bcbe69217fcfff02a00a8ef

turbolent (commented):
Great work! 👏 🧹 ✨

bbq/vm/value_capability.go
```diff
@@ -62,37 +61,25 @@ func init() {
...
 
-				return interpreter.CapabilityBorrow(
-					context,
-					typeParameter,
-					address,
-					capabilityID,
-					capabilityBorrowType,
-					EmptyLocationRange,
-				)
+				unifiedFunc := interpreter.UnifiedCapabilityBorrowFunction(address, capabilityID, capabilityBorrowType)
```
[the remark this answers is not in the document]
RZhang05: No, it can be refactored, I thought it might be cleaner this way, but to be consistent with the other refactored functions that have similar interpreter vs vm differences, these can be changed.

turbolent (approved):
Great work!

stdlib/account.go
```diff
@@ -255,72 +247,45 @@ var GetAuthAccountFunctionType = func() *sema.FunctionType {
...
-			)
-
-			return NewAccountReferenceValue(
-				inter,
-				handler,
-				accountAddress,
-				authorization,
-				locationRange,
-			)
+			return NewAccountReferenceValue(context, handler, accountAddress, referenceType.Authorization, locationRange)
```
SupunS, line 267:
```suggestion
			return NewAccountReferenceValue(
				context,
				handler,
				accountAddress,
				referenceType.Authorization,
				locationRange,
			)
```

stdlib/account.go
```diff
@@ -2875,51 +2660,37 @@ var GetAccountFunctionType = sema.NewSimpleFunctionType(
...
+	return interpreter.UnifiedNativeFunction(
+		func(
+			context interpreter.UnifiedFunctionContext,
+			locationRange interpreter.LocationRange,
+			typeParameterGetter interpreter.TypeParameterGetter,
+			receiver interpreter.Value,
+			args ...interpreter.Value,
+		) interpreter.Value {
+			accountAddress := interpreter.AssertValueOfType[interpreter.AddressValue](args[0])
+			return NewAccountReferenceValue(context, handler, accountAddress, interpreter.UnauthorizedAccess, locationRange)
```
SupunS, line 2673:
```suggestion
			return NewAccountReferenceValue(
				context,
				handler,
				accountAddress,
				interpreter.UnauthorizedAccess,
				locationRange,
			)
```

SupunS (approved):
Nice!

PR #4249, opened 2025-10-02
Set error location range in interpreter like in VM

turbolent:
Work towards #3804 

## Description

Location ranges for errors are provided by the VM by assigning it to the error *when it is reported*.

The interpreter passes the location range everywhere *in case an error is reported*. Threading the location range to deep call stacks is tedious, amounts to thousands of lines of code, and makes many function parameter lists longer.

Demonstrate that by using the method used in the VM also in the interpreter, we can remove the location range passing code for panics, and therefor eventually **all location range passing code**. This is accomplished by tracking the evaluated expression in addition to the currently evaluated statement.

In a follow-up PR I started removing a large amount of location range passing code (statements and function parameters), currently already amounting to savings of 2kLOC.

______


RZhang05 (approved):
good idea 👍

Issue #3691, opened 2024-11-21
Avoid conversion from `StaticType` to `sema.Type` during runtime sub-type checking


SupunS:
### Issue to be solved

In `Interpreter.IsSubType(...)` function, both the sub-type and the super-type gets converted from `StaticType` to `sema.Type` (see below). This adds an unnecessary overhead.

https://github.com/onflow/cadence/blob/b09231fe0216ecb024fbfc909f213bed53a2d3bb/interpreter/interpreter.go#L4066
https://github.com/onflow/cadence/blob/b09231fe0216ecb024fbfc909f213bed53a2d3bb/interpreter/interpreter.go#L4039

### Suggested Solution

Having two separate sub-type checking implementation for runtime (static types) could eleiminate this overhead. We could define the sub-typing rules in a declarative format, and then generate the two implementations from that, to make them consistant/in-sync.

PR #4262, opened 2025-10-06
Generate subtype checking rules from a DSL

SupunS:

## Description

🚧 WIP

______


tools/subtype-gen/README.md
```diff
@@ -0,0 +1,320 @@
...
+
+Represents a condition that is always false.
+
+### Not Predicate (`not`)
+
+Negates a nested predicate.
+
+### And Predicate (`and`)
+
+Represent a list of nested predicate that needs to be satisfied.
```
RZhang05:
```suggestion
Represent a list of nested predicates that needs to be satisfied.
```

tools/subtype-gen/README.md
```diff
@@ -0,0 +1,320 @@
...
+```
+
+### Or Predicate (`or`)
+
+Represent a list of nested predicate that could be satisfied.
+An `or` predicate becomes true if at-least one of its nested predicates are also true.
+e.g:
+
+```yaml
+and:
```
RZhang05:
```suggestion
or:
```

tools/subtype-gen/README.md
```diff
@@ -0,0 +1,320 @@
...
+
+```yaml
+  - super: AnyStructType
+    predicate:
+      equals:
+        source: sub  # `sub` is a `Type`
+        target: nil
+```
+
+In the above, `super` variable inside `equals` predicate contains a type-value of the root type `Type`.
```
RZhang05:
```suggestion
In the above, `sub` variable inside `equals` predicate contains a type-value of the root type `Type`.
```

interpreter/subtype_check.go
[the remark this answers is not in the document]
turbolent: Could we maybe keep this file on a separate branch that adds the code generation for the  subtype checking of static type later? That way we can keep the code on the feature branch focused on replacing the subtype checking of sema types

tools/subtype-gen/README.md
```diff
@@ -0,0 +1,320 @@
...
+An `or` predicate becomes true if at-least one of its nested predicates are also true.
+e.g:
+
+```yaml
+and:
+  - P1
+  - P2
+```
+
+### Equals Predicate (`equals`)
```
turbolent: Maybe note here that this corresponds to Go `==`, not e.g. an `Equals` method call

tools/subtype-gen/rules.yaml
```diff
@@ -0,0 +1,433 @@
...
+
+  - super: PathType
+    predicate:
+      subtype:
+        sub: sub
+        super:
+          # Maybe use 'or' predicate here instead?
+          # The difference is 'or' is a predicate containing zero or more inner predicates.
+          # Whereas, 'oneOf' contains zero or more values, not predicates.
+          oneOf: [StoragePathType, CapabilityPathType]
```
turbolent: I only noticed this comment in the code once I had already left the comment below. Agreed that using `or` might be simpler.

tools/subtype-gen/rules.yaml
```diff
@@ -0,0 +1,433 @@
...
+              source: sub
+              type: InterfaceType
+          - setContains:
+              source: sub.EffectiveInterfaceConformanceSet
+              target: super
+        - isParameterizedSubtype:
+            sub: sub
+            super: super
+
+  - super: IntersectionType
```
turbolent: Maybe as a TODO before merging the feature branch: Review and verify these and the other rules, as well as the code that is generated for them, and compare to the original Go code.

tools/subtype-gen/rules.yaml
```diff
@@ -0,0 +1,369 @@
...
+              source: sub
+              type: InterfaceType
+          - setContains:
+              source: sub.EffectiveInterfaceConformanceSet
+              target: super
+
+  - super: IntersectionType
+    predicate:
+      or:
+        - and:
```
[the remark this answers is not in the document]
turbolent: In a follow-up PR we can maybe split this up and use YAML anchors and comments to make this more readable

tools/subtype-gen/rules_parser.go
```diff
@@ -0,0 +1,459 @@
...
+// parseType parses a type with just a name.
+func parseType(typePlaceHolder string) Type {
+	if !strings.HasSuffix(typePlaceHolder, TypePlaceholderSuffix) {
+		panic(fmt.Errorf(
+			"type name %#[1]q is not suffixed with %#[2]q."+
+				" Replace the type-name with `%[1]s%[2]s`",
+			typePlaceHolder,
+			TypePlaceholderSuffix,
+		))
+	}
```
turbolent: 👍

tools/subtype-gen/rules_parser.go
```diff
@@ -0,0 +1,459 @@
...
+func parsePredicate(predicate any) (Predicate, error) {
+	switch v := predicate.(type) {
+	case string:
+		switch v {
+		case "always":
+			return AlwaysPredicate{}, nil
+		case "never":
+			return NeverPredicate{}, nil
+		default:
+			return nil, fmt.Errorf("unsupported string rule: %s", v)
```
turbolent:
```suggestion
			return nil, fmt.Errorf("unsupported string predicate: %s", v)
```

tools/subtype-gen/rules_parser.go
```diff
@@ -0,0 +1,459 @@
...
+				}
+				predicates = append(predicates, predicate)
+			}
+
+			return AndPredicate{Predicates: predicates}, nil
+
+		case "or":
+			or, ok := value.([]any)
+			if !ok {
+				return nil, fmt.Errorf("expected a list of predicates, got %T", value)
```
turbolent: Future idea: Improved error reporting, e.g. by keeping a path when parsing predicates (and rules), similar to #4302. Especially when rules get complicated it might be hard to find in the point in the rules files this error is for.

go.mod
```diff
@@ -17,6 +19,7 @@ require (
 	github.com/leanovate/gopter v0.2.11
 	github.com/logrusorgru/aurora/v4 v4.0.0
 	github.com/onflow/atree v0.11.0
+	github.com/onflow/cadence/tools/subtype-gen v0.0.0-00010101000000-000000000000
```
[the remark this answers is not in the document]
SupunS: Because we need to use this in both `sema` package and the `interpreter` package, we need to have this outside of those two. But yeah, I think I could make this just a sub-package in the main module, rather than a whole separate module.

tools/subtype-gen/rules.yaml
SupunS: Added an item here: https://github.com/onflow/cadence/issues/3691#issuecomment-3439450533

tools/subtype-gen/rules_parser.go
SupunS: Good idea! Added an item here: https://github.com/onflow/cadence/issues/3691#issuecomment-3439450533

tools/subtype-gen/main_test.go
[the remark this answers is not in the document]
SupunS: Yeah, definitely need to add some more tests to it. Added an item here: https://github.com/onflow/cadence/issues/3691#issuecomment-3439450533

RZhang05 (approved):

Issue #4273, opened 2025-10-09
[Compiler] Optimize conversion between static and sema types


RZhang05:
Avoid conversions where possible. Improve caching.

Possible approaches:
- [ ] Re-use static-type conversion results from compiler/ previous executions. See https://github.com/onflow/cadence/pull/4054#discussion_r2172847539
- [ ] #3691
- [ ] Avoid static to sema type conversion when boxing (ConvertAndBox) [moved from https://github.com/onflow/cadence/issues/3804]

PR #4054, opened 2025-06-27
Re-use static-type to sema-type conversion results

SupunS:
Work towards https://github.com/onflow/cadence/issues/3804

## Description

```
             │ without_caching.txt │          with_caching.txt          │
             │       sec/op        │   sec/op     vs base               │
FTTransfer-8           90.51µ ± 3%   77.88µ ± 2%  -13.95% (p=0.002 n=6)

             │ without_caching.txt │          with_caching.txt           │
             │        B/op         │     B/op      vs base               │
FTTransfer-8          54.78Ki ± 0%   47.85Ki ± 0%  -12.65% (p=0.002 n=6)

             │ without_caching.txt │          with_caching.txt          │
             │      allocs/op      │  allocs/op   vs base               │
FTTransfer-8           1.430k ± 0%   1.201k ± 0%  -16.01% (p=0.002 n=6)
```

______

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #4287, opened 2025-10-14
[Compiler] Re-use static to sema conversions in vm.

RZhang05:
Work towards #4273 

## Description

Re-use static to sema conversions from the compiler in the vm. Tried various changes in the linker to no success, tried caching static type ID calls, which was an improvement but affects execution, so not included.

```
              │ ./bbq/vm/test/old-3.txt │   ./bbq/vm/test/optimized-2.txt   │
              │         sec/op          │   sec/op     vs base              │
FTTransfer-14               39.10µ ± 2%   38.43µ ± 2%  -1.73% (p=0.007 n=8)

              │ ./bbq/vm/test/old-3.txt │ ./bbq/vm/test/optimized-2.txt │
              │          B/op           │     B/op      vs base         │
FTTransfer-14              36.93Ki ± 0%   36.93Ki ± 0%  ~ (p=0.660 n=8)

              │ ./bbq/vm/test/old-3.txt │ ./bbq/vm/test/optimized-2.txt │
              │        allocs/op        │ allocs/op   vs base           │
FTTransfer-14                965.0 ± 0%   965.0 ± 0%  ~ (p=1.000 n=8) ¹
```

______

### DIFF
--- a/bbq/compiler/compiler.go
+++ b/bbq/compiler/compiler.go
@@ -106,9 +106,7 @@ type Compiler[E, T any] struct {
 	// used mainly for exporting imports for linking
 	globalRemoveAddressTable map[string]string
 
-	// Passed to the vm to avoid repeated conversions from static-types to sema-types.
-	semaTypeCache map[sema.TypeID]sema.Type
-	addedImports  map[common.Location]struct{}
+	addedImports map[common.Location]struct{}
 }
 
 var _ ast.DeclarationVisitor[struct{}] = &Compiler[any, any]{}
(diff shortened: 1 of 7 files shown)


SupunS (commented):
It's surprising how/why caching doesn't improve the performance, given that first iteration of caching in https://github.com/onflow/cadence/pull/4054 gave us some improvement. 

Given that we already cache it after the very first conversion, maybe that initial conversion overhead is trivial?
Could you please add some more stat: for example, the number of cache hits/misses before and after the improvement?

Additionally, do we also need to cache types coming from the imports / for the imported programs?

PR #4315, opened 2025-10-28
Fix accidental value caching and reuse in native functions when run with VM

turbolent:
## Description

Fix a regression in the VM introduced in the recent refactor of native functions.

Native functions for methods are called slightly differently by the interpreter and VM:
- In the interpreter, the function is only ever called for the same receiver, which is often passed as a closed over parameter
- In the VM, the function is called for all/different receivers

We have to be careful not to cache the receiver or other values in the first call and reuse it in a subsequent call.

Fix the cases that I could find and ensure the behaviour is correct by extending the existing tests with an additional second call of the tested method.

It would be nice to lint such cases and prevent them in the future, but so far I could not find anything existing.

______

### DIFF
--- a/interpreter/value_deployedcontract.go
+++ b/interpreter/value_deployedcontract.go
@@ -20,6 +20,7 @@ package interpreter
 
 import (
 	"github.com/onflow/cadence/common"
+	"github.com/onflow/cadence/errors"
 	"github.com/onflow/cadence/sema"
 )
 
@@ -69,7 +70,6 @@ func NewDeployedContractValue(
 func NewNativeDeployedContractPublicTypesFunctionValue(
 	addressPointer *common.Address,
 	name *StringValue,
-	publicTypes *ArrayValue,
 ) NativeFunction {
 	return func(
 		context NativeFunctionContext,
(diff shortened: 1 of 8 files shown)


RZhang05 (approved):
thank you 🙏

stdlib/publickey.go
```diff
@@ -333,8 +335,10 @@ func NativePublicKeyVerifyPoPFunction(
 	) interpreter.Value {
 		signatureValue := interpreter.AssertValueOfType[*interpreter.ArrayValue](args[0])
 
+		publicKeyValue := constantPublicKeyValue
```
SupunS: Shouldn't we **not** cache the value? 🤔

interpreter/value_deployedcontract.go
```diff
@@ -94,18 +94,18 @@ func NewNativeDeployedContractPublicTypesFunctionValue(
...
 			name = AssertValueOfType[*StringValue](nameFieldValue)
 		} else {
+			// Interpreter provides address and name
 			address = *addressPointer
+			if name == nil {
+				panic(errors.NewUnreachableError())
+			}
 		}
 
-		if publicTypes == nil {
```
turbolent: The result used to be cached for the interpreter, but also accidentally for the VM. As we don't cache in any other places, I thought it was better/cleaner/safer to remove the caching here and always compute the result, no matter the environment.

stdlib/publickey.go
turbolent: Sorry, I don't quite understand the question/concern. `constantPublicKeyValue` is passed by the interpreter, but is `nil` in the VM. So at call-time we use it if it was passed (interpreter), or use the current receiver.

SupunS (approved):

PR #4319, opened 2025-10-30
Add peephole pass with potential optimizations.

RZhang05:
Working towards #4305 

## Description

Adds a peephole pass with sample optimizations hidden behind a feature flag. These optimizations are not confirmed safe and will likely lead to differences between the vm and interpreter.

Also changes `CompiledFunctionValue` to be a pointer type.

Benchmarking impact on compile time of a simple program (https://github.com/onflow/cadence/pull/4319/files#diff-f4d6987e3014a1d17cc52ba737ca0eddb39d0c1ae84eb139334cd620c1fa8926R10488):
```
               │ before.txt  │              after.txt              │
               │   sec/op    │   sec/op     vs base                │
CompileTime-14   10.62µ ± 2%   11.87µ ± 0%  +11.73% (p=0.001 n=10)

CompileTime-14   15.07Ki ± 0%   16.22Ki ± 0%  +7.62% (p=0.000 n=10)

               │ before.txt │             after.txt              │
               │ allocs/op  │ allocs/op   vs base                │
CompileTime-14   173.0 ± 0%   192.0 ± 0%  +10.98% (p=0.000 n=10)
```

______

### DIFF
--- a/bbq/compiler/peephole_pass.go
+++ b/bbq/compiler/peephole_pass.go
@@ -19,9 +19,11 @@
 package compiler
 
 import (
+	"math"
 	"sort"
 
 	"github.com/onflow/cadence/bbq/opcode"
+	"github.com/onflow/cadence/errors"
 	"github.com/onflow/cadence/interpreter"
 )
 
@@ -105,6 +107,10 @@ func (o *PeepholeInstructionStaticTypeOptimizer) patchJumps(optimized []opcode.I
 
 		// patch jump
 		newJumpTarget := jumpTarget + cumShift
+		if newJumpTarget > math.MaxUint16 {
+			//TODO: abort optimization pass instead of panicking
+			panic(errors.NewUnexpectedError("peephole shifted jump target past max uint16"))
+		}
 		opcode.PatchJumpInstruction(optimized, jump, uint16(newJumpTarget))
 	}
 }


bbq/compiler/peephole_patterns.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+var NilTransferAndConvertPattern = PeepholePattern{
+	Name:    "NilTransferAndConvert",
+	Opcodes: []opcode.Opcode{opcode.Nil, opcode.TransferAndConvert},
+	Replacement: func(instructions []opcode.Instruction, compiler *Compiler[opcode.Instruction, interpreter.StaticType]) []opcode.Instruction {
+		nil := instructions[0].(opcode.InstructionNil)
+		return []opcode.Instruction{nil}
+	},
+}
```
SupunS: I think this is similar to the case we ran into last time: https://github.com/onflow/cadence/blob/4fb3feb908bfedb000953e413131c8d9e679197f/bbq/compiler/compiler.go#L4008-L4011. Could you please add a test, or even-better, run all existing tests with the optimizations enabled? That might uncover any edge-cases if present.

bbq/compiler/peephole_pass.go
```diff
@@ -0,0 +1,182 @@
...
+			optimized[jump] = opcode.InstructionJumpIfFalse{Target: uint16(newJumpTarget)}
+		case opcode.InstructionJumpIfTrue:
+			optimized[jump] = opcode.InstructionJumpIfTrue{Target: uint16(newJumpTarget)}
+		case opcode.InstructionJumpIfNil:
+			optimized[jump] = opcode.InstructionJumpIfNil{Target: uint16(newJumpTarget)}
+		}
+	}
+}
+
+func (o *PeepholeOptimizer[E, T]) OptimizeInstructions(instructions []opcode.Instruction) []opcode.Instruction {
```
SupunS: Can you also do a quick benchmark comparison for the compilation times with and without the optimizations? Optimizations adding some compilation overhead is inevitable I guess, but we can chose to only optimize contracts (which are compiled only once and executed multiple times). A simple benchmark might help us getting some rough idea.

bbq/compiler/peephole_patterns.go
RZhang05: I can also add simple control flow / basic block identification to try and get all the tests to pass, then benchmark to see the compile time tradeoff.

bbq/compiler/peephole_pass.go
```diff
@@ -0,0 +1,202 @@
...
+
+func NewPeepholeOptimizer[E, T any](compiler *Compiler[E, T]) PeepholeOptimizer[E, T] {
+	// this map is meant to improve pattern matching performance
+	patternsByOpcode := make(map[opcode.Opcode][]PeepholePattern)
+	for _, pattern := range AllPatterns {
+		if patternsByOpcode[pattern.Opcodes[0]] == nil {
+			patternsByOpcode[pattern.Opcodes[0]] = make([]PeepholePattern, 0)
+		}
+		patternsByOpcode[pattern.Opcodes[0]] = append(patternsByOpcode[pattern.Opcodes[0]], pattern)
+	}
```
turbolent: Is the resulting map (`patternsByOpcode`), or any of the values (`[]PeepholePattern` slices) ever mutated after construction? From what I can see it is not. Does the map differ for each optimizer instance? 

If not, we can construct this lookup table in a global variable and construct it at package initialization time by moving this code into an [`init()` function](https://go.dev/doc/effective_go#init)

bbq/compiler/peephole_pass.go
```diff
@@ -0,0 +1,202 @@
...
+		switch ins := optimized[jump].(type) {
+		case opcode.InstructionJump:
+			jumpTarget = int(ins.Target)
+		case opcode.InstructionJumpIfFalse:
+			jumpTarget = int(ins.Target)
+		case opcode.InstructionJumpIfTrue:
+			jumpTarget = int(ins.Target)
+		case opcode.InstructionJumpIfNil:
+			jumpTarget = int(ins.Target)
+		}
```
turbolent: Maybe extract this into a function in the `opcode` package, e.g. `func JumpTarget(instruction opcode.Instruction) (uint16, bool)`. That way we have it centralized in one place

turbolent: Maybe also handle the case where no case matches, e.g. when a new jump instruction is added later, by adding a `default: panic(errors.NewUnreachableError())`

bbq/compiler/peephole_pass.go
```diff
@@ -0,0 +1,202 @@
...
+		switch optimized[jump].(type) {
+		case opcode.InstructionJump:
+			optimized[jump] = opcode.InstructionJump{Target: uint16(newJumpTarget)}
+		case opcode.InstructionJumpIfFalse:
+			optimized[jump] = opcode.InstructionJumpIfFalse{Target: uint16(newJumpTarget)}
+		case opcode.InstructionJumpIfTrue:
+			optimized[jump] = opcode.InstructionJumpIfTrue{Target: uint16(newJumpTarget)}
+		case opcode.InstructionJumpIfNil:
+			optimized[jump] = opcode.InstructionJumpIfNil{Target: uint16(newJumpTarget)}
+		}
```
turbolent: This is very similar to `InstructionCodeGen.PatchJump`. Maybe we can refactor that existing function and this new code, and define a function `func PatchJump(instructions *[]Opcode, offset int, newTarget uint16)`

bbq/compiler/peephole_pass.go
```diff
@@ -0,0 +1,202 @@
...
+		return jumpATarget < jumpBTarget
+	})
+
+	o.patchJumps(optimized)
+
+	return optimized
+}
+
+func (o *PeepholeInstructionOptimizer[E, T]) Optimize(code []E) []E {
+	if instructions, ok := any(code).([]opcode.Instruction); ok {
```
turbolent: If `PeepholeInstructionOptimizer` only works for `E == opcode.Instruction`, then the `E` type parameter should be removed. Happy to help with that

bbq/vm/vm.go
```diff
@@ -1539,6 +1539,41 @@ func opRemoveTypeIndex(vm *VM, ins opcode.InstructionRemoveTypeIndex) {
...
+	fieldName := getRawStringConstant(vm, fieldNameIndex)
+
+	fieldValue := memberAccessibleValue.GetMember(vm.context, fieldName)
+	if fieldValue == nil {
+		panic(&interpreter.UseBeforeInitializationError{
+			Name: fieldName,
+		})
+	}
+
+	vm.push(fieldValue)
```
turbolent: Would it be possible to reuse the existing components of the instructions, i.e. `opGetLocal` and `opGetField`? We might have to refactor parts of those to make their logic reusable (i.e. avoid a temporary push and pop to the stack)

bbq/vm/value_function.go
```diff
@@ -57,26 +57,26 @@ type CompiledFunctionValue struct {
 	Type       interpreter.FunctionStaticType
 }
 
-var _ Value = CompiledFunctionValue{}
-var _ FunctionValue = CompiledFunctionValue{}
+var _ Value = &CompiledFunctionValue{}
```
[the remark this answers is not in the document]
RZhang05: Unrelated optimization from Milestone X, could be removed from this PR (Make NativeFunctionValue, CompiledFunctionValue, and other large types pointer types).

bbq/compiler/peephole_patterns.go
RZhang05: We can definitely collect all the jump targets before the optimizations.

bbq/compiler/peephole_pass.go
```diff
@@ -0,0 +1,202 @@
...
+		case opcode.InstructionJumpIfTrue:
+			optimized[jump] = opcode.InstructionJumpIfTrue{Target: uint16(newJumpTarget)}
+		case opcode.InstructionJumpIfNil:
+			optimized[jump] = opcode.InstructionJumpIfNil{Target: uint16(newJumpTarget)}
+		}
+	}
+}
+
+func (o *PeepholeInstructionOptimizer[E, T]) OptimizeInstructions(instructions []opcode.Instruction) []opcode.Instruction {
+	optimized := make([]opcode.Instruction, 0, len(instructions))
```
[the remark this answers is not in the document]
turbolent: We could maybe just replace the removed instructions with no-op instructions. Let's leave it as-is for now and record the optimization in the issue as a follow-up TODO/idea.

bbq/compiler/peephole_pass.go
```diff
@@ -0,0 +1,170 @@
...
+	// after the peephole optimizations have been applied
+	bytecodeShifts []BytecodeShiftPair
+
+	// jumps is a list of offsets in the optimized bytecode that need to be patched
+	jumps []int
+
+	// jumpTargets is a set of offsets in the original bytecode that are jump targets
+	jumpTargets map[int]struct{}
+}
+
```
turbolent: Maybe assert here that `PeepholeInstructionStaticTypeOptimizer` implements `PeepholeOptimizer` (once type parameter `T` got removed):

```suggestion

var _ PeepholeOptimizer[opcode.Instruction] = &PeepholeInstructionStaticTypeOptimizer{}

```

bbq/compiler/peephole_pass.go
```diff
@@ -0,0 +1,170 @@
...
+// look up table to improve pattern matching performance
+var PatternsByOpcode map[opcode.Opcode][]PeepholePattern
+
+func init() {
+	PatternsByOpcode = make(map[opcode.Opcode][]PeepholePattern)
+	for _, pattern := range AllPatterns {
+		if PatternsByOpcode[pattern.Opcodes[0]] == nil {
+			PatternsByOpcode[pattern.Opcodes[0]] = make([]PeepholePattern, 0)
+		}
+		PatternsByOpcode[pattern.Opcodes[0]] = append(PatternsByOpcode[pattern.Opcodes[0]], pattern)
```
turbolent: Extract the repeated expression, and the `nil` check / slice allocation is not needed, the `map` lookup returns `nil` by default, and `append` accepts that as the first argument:

```suggestion
		firstOpcode := pattern.Opcodes[0]
		PatternsByOpcode[firstOpcode] = append(PatternsByOpcode[firstOpcode], pattern)
```

bbq/compiler/peephole_pass.go
```diff
@@ -0,0 +1,170 @@
...
+	})
+
+	o.patchJumps(optimized)
+
+	return optimized
+}
+func (o *PeepholeInstructionStaticTypeOptimizer) Optimize(code []opcode.Instruction) []opcode.Instruction {
+	o.bytecodeShifts = make([]BytecodeShiftPair, 0)
+	o.jumps = make([]int, 0)
+	o.jumpTargets = make(map[int]struct{})
```
turbolent: Move the allocation to the constructor function (later we should ideally allocate these lazily), and re-use the values here, instead of throwing away and re-allocating:

```suggestion
	o.bytecodeShifts = o.bytecodeShifts[:0]
	o.jumps = o.jumps[:0]
	clear(o.jumpTargets)
```

bbq/opcode/instruction_jump.go
```diff
@@ -0,0 +1,52 @@
...
+func IsJump(instruction Instruction) bool {
+	switch instruction.(type) {
+	case InstructionJump:
+		return true
+	case InstructionJumpIfFalse:
+		return true
+	case InstructionJumpIfTrue:
+		return true
+	case InstructionJumpIfNil:
+		return true
```
turbolent: Maybe combine all cases: 

```suggestion
	case InstructionJump,
		InstructionJumpIfFalse
		InstructionJumpIfTrue,
		InstructionJumpIfNil:
		
		return true
```

bbq/vm/vm.go
```diff
@@ -1037,17 +1041,15 @@ func opSetField(vm *VM, ins opcode.InstructionSetField) {
 	memberAccessibleValue.SetMember(vm.context, fieldName, fieldValue)
 }
 
-func opGetField(vm *VM, ins opcode.InstructionGetField) {
-	memberAccessibleValue := vm.pop().(interpreter.MemberAccessibleValue)
+func getField(vm *VM, memberAccessibleValue interpreter.MemberAccessibleValue, fieldNameIndex uint16, accessedTypeIndex uint16) Value {
```
turbolent: 👌

bbq/vm/vm.go
```diff
@@ -1539,6 +1549,14 @@ func opRemoveTypeIndex(vm *VM, ins opcode.InstructionRemoveTypeIndex) {
 	}
 }
 
+func opGetFieldLocal(vm *VM, ins opcode.InstructionGetFieldLocal) {
+	local := getLocal(vm, ins.Local)
+	memberAccessibleValue := local.(interpreter.MemberAccessibleValue)
+	fieldValue := getField(vm, memberAccessibleValue, ins.FieldName, ins.AccessedType)
```
turbolent: 👌

PR #4320, opened 2025-10-31
[Subtyping Generator] Fix return statement generation and add tests

SupunS:

## Description

Add tests for the code generation.

While adding tests, I realized there were two bugs with return statement generation:
 1. Switch-cases were not always returning.
 2. When complex rules are added, the negation wasn't get applied to the return. Again, this was only when switch-statements are involved. 

Fixed these by always generating the return statements whenever switch-statements are generated, and updating the nodes "merging" logic to handle multiple/duplicate returns.

______

### DIFF
--- a/tools/subtype-gen/testdata/complex_nested_and_or_test.yaml
+++ b/tools/subtype-gen/testdata/complex_nested_and_or_test.yaml
@@ -6,11 +6,10 @@ rules:
             - mustType:
                 source: sub
                 type: IntersectionType
-            - not:
-                equals:
-                  source: sub.LegacyType
-                  target:
-                    oneOf: [nil, AnyResourceType, AnyStructType, AnyType]
+            - equals:
+                source: sub.LegacyType
+                target:
+                  oneOf: [nil, AnyResourceType, AnyStructType, AnyType]
             - mustType:
                 source: sub.LegacyType
                 type: CompositeType
--- a/tools/subtype-gen/testdata/complex_nested_and_or_with_not_test.golden.go
+++ b/tools/subtype-gen/testdata/complex_nested_and_or_with_not_test.golden.go
@@ -28,21 +28,21 @@ func checkSubTypeWithoutEquality_gen(subType Type, superType Type) bool {
 	case ComplexType:
 		switch typedSubType := subType.(type) {
 		case *IntersectionType:
+
+			// The `not` operator will make this exit-early if matches.
 			switch typedSubType.LegacyType {
 			case nil,
 				AnyResourceType,
 				AnyStructType,
 				AnyType:
-				switch typedSubTypeLegacyType := typedSubType.LegacyType.(type) {
-				case *CompositeType:
-					return true
-				}
-
 				return false
 			}
 
-			return false
-		case *CompositeType:
+			switch typedSubTypeLegacyType := typedSubType.LegacyType.(type) {
+			case *CompositeType:
+				return true
+			}
+
 			return false
 		}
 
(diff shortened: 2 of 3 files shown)


tools/subtype-gen/testdata/complex_nested_and_or_test.golden.go
```diff
@@ -0,0 +1,54 @@
...
+	case ComplexType:
+		switch typedSubType := subType.(type) {
+		case *IntersectionType:
+			switch typedSubType.LegacyType {
+			case nil,
+				AnyResourceType,
+				AnyStructType,
+				AnyType:
+				return false
+			}
```
turbolent: Would the generated code of the `equals` + `oneOf` which is _not_ contained in a `not` like here, result in an additional `default` case which returns false?

tools/subtype-gen/testdata/deeply_nested_predicates_test.yaml
turbolent: Given the complexity of the rules I wasn't quite able to manually verify if the generated code matches the rules

SupunS: Sorry about the complexity. I added it mostly to verify the code generation for a complex scenario, with combinations of `and` and `or` predicates. I manually verified the generated code.

It might be easier to verify by considering parts of the rules. For example, the outer-most `or` predicate as two inner predicates: they corresponds to the two switch-statements generated, in order.

tools/subtype-gen/testdata/return_covariant_predicate_test.yaml
```diff
@@ -0,0 +1,7 @@
+rules:
+  - super: FunctionType
+    predicate:
+      returnCovariant:
+        source: sub
+        target: super
```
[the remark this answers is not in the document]
SupunS: I thought about it, the catch was that the return type could also be `nil`, and the subtype check treats `nil` differently than how `returnCovatiant` treats it. i.e: In `IsSubType` method, if subtype is `nil`, then the result is _always_ `false`. However, for returns, if one of the return-types is `nil` then the other also being `nil` is a valid scenario.

tools/subtype-gen/testdata/complex_nested_and_or_test.golden.go
turbolent: Great! Thank you for the explanation 👍

tools/subtype-gen/testdata/complex_nested_and_or_test.yaml
```diff
@@ -0,0 +1,22 @@
...
+                  target:
+                    oneOf: [nil, AnyResourceType, AnyStructType, AnyType]
+            - mustType:
+                source: sub.LegacyType
+                type: CompositeType
+        - and:
+            - mustType:
+                source: sub
+                type: CompositeType
+            - never
```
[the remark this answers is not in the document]
turbolent: My bad, I had just misread the generated code: the switch case is the first part of the `and`, and the second part (`never`) is the inner `return false`, makes sense 👍 

I wasn't proposing to optimize the generated code, the Go compiler can do that for us, no worries

tools/subtype-gen/testdata/deeply_nested_predicates_test.yaml
turbolent: Oh no worries at all, I just felt bad that I wasn't of much help here. I think it's great to have a more complex test case that checks combinations :100:

RZhang05 (approved):
👏

PR #4322, opened 2025-11-06
[Subtyping Generator] Switch to static-types for runtime subtype checking

### LINKED ISSUE #3691: Avoid conversion from `StaticType` to `sema.Type` during runtime sub-type checking

SupunS:
**⚠️ DO NOT MERGE**: While this improves the performance of the VM, as a side effect, it degrades the performance of the interpreter. So we should only merge/deploy this once we fully switched to the VM. Unfortunately it's not trivial to put these changes behind a feature flag.

Closes #3691

## Description

- Use the generated static-type based subtyping check for the `IsSubType()` function (instead of the subtyping check based on sema-type)
- Remove usages of `MustConvertStaticToSemaType` in VM, and rely on static-types.
- Migrate all stdlib functions (except for a one or two that strictly needs composite type) to only use static-types (replace the usages of sema-type).

### Vs VM in last released version (`v1.8.3`)
```
goos: darwin
goarch: arm64
pkg: github.com/onflow/cadence/runtime
cpu: Apple M1 Pro
                                 │ v1.8.3.txt  │             after.txt              │
                                 │   sec/op    │   sec/op     vs base               │
RuntimeFungibleTokenTransferVM-8   291.3µ ± 9%   273.0µ ± 4%  -6.27% (p=0.002 n=10)

                                 │  v1.8.3.txt  │              after.txt              │
RuntimeFungibleTokenTransferVM-8   171.9Ki ± 0%   160.0Ki ± 0%  -6.92% (p=0.000 n=10)

                                 │ v1.8.3.txt  │              after.txt              │
                                 │  allocs/op  │  allocs/op   vs base                │
RuntimeFungibleTokenTransferVM-8   3.082k ± 0%   2.753k ± 0%  -10.67% (p=0.000 n=10)
```

### Vs VM before switching to static-types (master)
```
goos: darwin
goarch: arm64
pkg: github.com/onflow/cadence/runtime
cpu: Apple M1 Pro
                                 │ before.txt  │             after.txt              │
                                 │   sec/op    │   sec/op     vs base               │
RuntimeFungibleTokenTransferVM-8   285.6µ ± 9%   273.0µ ± 4%  -4.40% (p=0.011 n=10)

RuntimeFungibleTokenTransferVM-8   170.6Ki ± 0%   160.0Ki ± 0%  -6.21% (p=0.000 n=10)

                                 │ before.txt  │             after.txt              │
                                 │  allocs/op  │  allocs/op   vs base               │
RuntimeFungibleTokenTransferVM-8   3.000k ± 0%   2.753k ± 0%  -8.23% (p=0.000 n=10)
```
**_Note: The master branch also contains some optimizations that was ported over. So the above comparison only includes the improvement we get from switching to static-types only._

### Vs Interpreter (master)
```
goos: darwin
goarch: arm64
pkg: github.com/onflow/cadence/runtime
cpu: Apple M1 Pro
                               │ interpreter.txt │           after.txt           │
                               │     sec/op      │   sec/op     vs base          │
RuntimeFungibleTokenTransfer-8       272.8µ ± 2%   273.0µ ± 4%  ~ (p=0.796 n=10)

                               │ interpreter.txt │              after.txt              │
                               │      B/op       │     B/op      vs base               │
RuntimeFungibleTokenTransfer-8      157.3Ki ± 0%   160.0Ki ± 0%  +1.70% (p=0.000 n=10)

                               │ interpreter.txt │             after.txt              │
                               │    allocs/op    │  allocs/op   vs base               │
RuntimeFungibleTokenTransfer-8       2.980k ± 0%   2.753k ± 0%  -7.62% (p=0.000 n=10)
```

## Future work:
- Event emitting still need the sema types (hence the conversion exist). Should try to replace those as well.
- Generated subtype check for static-types still converts to sema types in certain cases (for e.g: conformance information is not available in static-types, so need to convert to sema-type to get these information)
______

### DIFF
--- a/ast/expression.go
+++ b/ast/expression.go
@@ -276,8 +276,25 @@ func (e *StringTemplateExpression) Doc() prettier.Doc {
 		return prettier.Text(QuoteString(e.Values[0]))
 	}
 
-	// TODO: must reproduce expressions as literals
-	panic("not implemented")
+	concat := make(prettier.Concat, 0, 2+len(e.Values)+(3*len(e.Expressions)))
+	concat = append(concat, prettier.Text(`"`))
+	for i, value := range e.Values {
+		var sb strings.Builder
+		QuoteStringInner(value, &sb)
+		concat = append(concat, prettier.Text(sb.String()))
+
+		if i < len(e.Expressions) {
+			e := e.Expressions[i]
+			concat = append(
+				concat,
+				prettier.Text(`\(`),
+				e.Doc(),
+				prettier.Text(`)`),
+			)
+		}
+	}
+	concat = append(concat, prettier.Text(`"`))
+	return concat
 }
 
 func (e *StringTemplateExpression) MarshalJSON() ([]byte, error) {
--- a/ast/expression_extractor.go
+++ b/ast/expression_extractor.go
@@ -146,6 +146,7 @@ type ExpressionExtractor struct {
 	ReferenceExtractor      ReferenceExtractor
 	MemberExtractor         MemberExtractor
 	PathExtractor           PathExtractor
+	IdentifierPrefix        string
 	nextIdentifier          int
 }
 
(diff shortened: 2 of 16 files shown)


SupunS:
>  Is this because in the interpreter we get sema types from the elaboration, and now have to convert them to static types?

 Yes, correct. Before this PR, all the common codes of the Cadence runtime (stdlibs, runtime utility functions, etc.) were using sema-types, and was "optimized for the interpreter".  Now those functions use the static-types, making them optimized for the VM, but the interpreter now have to bear the cost for the converting the sema-types to the static-types.

PR #4327, opened 2025-11-10
[Subtyping Generator] Generate subtype check for runtime static-types

SupunS:

## Description

Includes only the changes to related to the subtype-check "generation". This doesn't integrate the generated function to the runtime. I have opened a separate PR https://github.com/onflow/cadence/pull/4322 to integrate the generated function.

______

### DIFF
--- a/interpreter/interface.go
+++ b/interpreter/interface.go
@@ -638,7 +638,3 @@ func (NoOpStringContext) SemaTypeFromStaticType(_ StaticType) sema.Type {
 func (NoOpStringContext) SemaAccessFromStaticAuthorization(Authorization) (sema.Access, error) {
 	panic(errors.NewUnreachableError())
 }
-
-func (ctx NoOpStringContext) SemaAccessFromStaticAuthorization(auth Authorization) sema.Access {
-	panic(errors.NewUnreachableError())
-}
--- a/interpreter/subtype_check.go
+++ b/interpreter/subtype_check.go
@@ -84,8 +84,16 @@ func IsResourceType(typeConverter TypeConverter, typ StaticType) bool {
 }
 
 func PermitsAccess(typeConverter TypeConverter, superTypeAuth, subTypeAuth Authorization) bool {
-	superTypeAccess := typeConverter.SemaAccessFromStaticAuthorization(superTypeAuth)
-	subTypeAccess := typeConverter.SemaAccessFromStaticAuthorization(subTypeAuth)
+	superTypeAccess, err := typeConverter.SemaAccessFromStaticAuthorization(superTypeAuth)
+	if err != nil {
+		panic(err)
+	}
+
+	subTypeAccess, err := typeConverter.SemaAccessFromStaticAuthorization(subTypeAuth)
+	if err != nil {
+		panic(err)
+	}
+
 	return sema.PermitsAccess(superTypeAccess, subTypeAccess)
 }
 


interpreter/subtype_check.gen.go
```diff
@@ -0,0 +1,417 @@
...
+					return false
+				}
+
+				for i, source := range typedSubTypeParameters {
+					target := typedSuperTypeParameters[i]
+					if
+
+					// Note the super-type is the subtype's parameter
+					// because the parameters are contravariant.
+					!(sema.IsSubType(target.TypeAnnotation.Type, source.TypeAnnotation.Type)) {
```
SupunS: Same here, the `FunctionStaticType` is just a wrapper around `sema.FunctionType`. Therefore the parameter type (i.e: `target.TypeAnnotation.Type`) is a `sema.Type` (not a `StaticType`). So calling `IsSubType` method of the interpreter package (like every other place inside this method) isn't going to work. Needed to call the `IsSubType` method from the `sema` package. So the customizer will update here as well, to the desire code, similarly to the previous case.

bbq/vm/context.go
```diff
@@ -55,7 +55,8 @@ type Context struct {
 	// This cache-alike is maintained per execution.
 	// TODO: Re-use the conversions from the compiler.
 	// TODO: Maybe extend/share this between executions.
-	semaTypeCache map[sema.TypeID]sema.Type
+	semaTypeCache   map[sema.TypeID]sema.Type
+	semaAccessCache map[interpreter.Authorization]sema.Access
```
turbolent: Is this cache only possible/used by the new generated subtype checking code, or is this a general optimization that could also be done on `master`? If the latter, it might be nice to port it over before, it's a great idea. That would keep this PR focused on just adding the generated subtype checking code.

interpreter/interface.go
```diff
@@ -633,3 +634,7 @@ func (ctx NoOpStringContext) IsTypeInfoRecovered(_ common.Location) bool {
 func (ctx NoOpStringContext) SemaTypeFromStaticType(_ StaticType) sema.Type {
 	panic(errors.NewUnreachableError())
 }
+
+func (ctx NoOpStringContext) SemaAccessFromStaticAuthorization(auth Authorization) sema.Access {
```
turbolent: Maybe also remove the other unused receiver variables from the other methods

```suggestion
func (NoOpStringContext) SemaAccessFromStaticAuthorization(auth Authorization) sema.Access {
```

bbq/vm/context.go
SupunS: Ported https://github.com/onflow/cadence/pull/4336

interpreter/statictype.go
```diff
@@ -1063,6 +1107,11 @@ func ConvertSemaToStaticType(memoryGauge common.MemoryGauge, t sema.Type) Static
 
 	case *sema.TransactionType:
 		return ConvertSemaTransactionToStaticTransactionType(memoryGauge, t)
+
+	case *sema.GenericType:
+		// Function types could have generic-typed returns/parameters. e.g: builtin functions.
+		// Since they are not resolved, the type is unknown here.
+		return PrimitiveStaticTypeUnknown
```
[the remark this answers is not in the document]
SupunS: Opened https://github.com/onflow/cadence/issues/4337 to track this improvement.

PR #4332, opened 2025-11-12
[Subtyping Generator] Add tests for type-checker's subtype check

SupunS:
Work towards https://github.com/onflow/cadence/issues/3691

## Description

Add tests to achieve 100% coverage for the `checkSubTypeWithoutEquality` function. i.e: includes a test for every execution path of the function.

Then run both the existing `checkSubTypeWithoutEquality` as well as the generated function `checkSubTypeWithoutEquality_gen`, to ensure the generated function produces the same results as the existing function.

______

- [ ] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [x] Code follows the [standards mentioned here](https://github.com/onflow/cadence/blob/master/CONTRIBUTING.md#styleguides)
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels

PR #4335, opened 2025-11-13
[Subtyping Generator] Test and improve generated runtime subtype check 

SupunS:

Depends on https://github.com/onflow/cadence/pull/4327

## Description

Run the same set of tests added in https://github.com/onflow/cadence/pull/4332 for the generated static-type checking function as well. Fixed the gaps uncovered during tests.

______


runtime/subtype_check_test.go
SupunS: Moved `sema/subtype_check_test.go` to the `runtime` package, so that the generated function for static-types can also be tested at the same time, using the same set of test. Git is showing this as a new file rather than a move, because there were lot of package-name updates within the code (e.g: `IntType` -> `sema.IntType`), and I extracted some pre-computed types to a file level (from test level)

runtime/subtype_check_test.go
```diff
@@ -0,0 +1,1720 @@
...
+)
+
+var location = common.NewStringLocation(nil, "test")
+
+var interfaceType1 = &sema.InterfaceType{
+	Location:      location,
+	Identifier:    "I1",
+	CompositeKind: common.CompositeKindStructure,
+	Members:       &sema.StringMemberOrderedMap{},
+}
```
SupunS: Moved these types to the file-level, since they are re-used in multiple tests

runtime/subtype_check_test.go
```diff
@@ -0,0 +1,1720 @@
...
+		t,
+		result,
+		generatedStaticTypeResult,
+		"generated function in `interpreter` package produced different results for"+
+			" subType=%s, superType=%s: manual=%s, generated=%s",
+		subType,
+		superType,
+		result,
+		generatedStaticTypeResult,
+	)
```
SupunS: New addition: Now this test also tests the `interpreter.CheckSubTypeWithoutEquality_gen(...)` function

encoding/ccf/simpletype_test.go
```diff
@@ -52,7 +52,9 @@ func TestTypeConversion(t *testing.T) {
 	}
 
 	for ty := interpreter.PrimitiveStaticType(1); ty < interpreter.PrimitiveStaticType_Count; ty++ {
-		if !ty.IsDefined() || ty.IsDeprecated() { //nolint:staticcheck
+		if !ty.IsDefined() ||
+			ty.IsDeprecated() || //nolint:staticcheck
+			ty == interpreter.PrimitiveStaticTypeStorable {
```
SupunS: I didn't add support for "exporting" storable static type, because it not something a user can write/use (only used in subtyping checks). But noticed we already support exporting `Any` static type, which is also not something user's can use in their code. Do we maybe want to support the storable-type in a similar way?

sema/subtype_check.gen.go
```diff
@@ -211,7 +211,7 @@ func checkSubTypeWithoutEquality_gen(subType Type, superType Type) bool {
 
 			switch typedSubTypeLegacyType := typedSubType.LegacyType.(type) {
 			case *CompositeType:
-				return typedSubTypeLegacyType == typedSuperType
+				return deepEquals(typedSubTypeLegacyType, typedSuperType)
```
RZhang05: For these `deepEquals` replacements, can you explain why its necessary?

RZhang05 (approved):

encoding/ccf/simpletype_test.go
turbolent: If we're adding a new primitive type we should also add support for converting/exporting it. I'm not quite sure how this would be encountered, but we should handle it properly if it is.

SupunS: Added in https://github.com/onflow/cadence/pull/4335/commits/0c933655e25442d01e09eb04dc0cb44fb367c3fb

PR #4344, opened 2025-11-19
[Subtyping Generator] Always run the generated subtyping check and compare results

SupunS:

## Description

Enable the generated subtype check (for sema package) to be run always, and compared the results with the existing one. The comparison can be toggled using the flag. 

Unfortunately the flag is a package level one (not per checker/vm), because the `IsSubType` function is a static-function, and is being used in 100+ places. Trying to make it per checker/vm adds too much unnecessary tech-debt.

______


turbolent (approved):
Great work! Do you think this is ready for testing now?

PR #4359, opened 2025-11-25
adding sast-sca

JGowsk9:
## Security
-Adding Code Analysis
-Adding Dependency Review
-This will block failing PR's

### DIFF
--- a/.github/workflows/compatibility-check-template.yml
+++ b/.github/workflows/compatibility-check-template.yml
@@ -35,7 +35,7 @@ jobs:
 
       - uses: actions/setup-go@v6
         with:
-          go-version-file: go.mod
+          go-version-file: ./tools/compatibility-check/go.mod
           cache: true
 
       - name: Make output dirs


.github/workflows/code-analysis.yml
turbolent: We already have a CodeQL analysis action here: https://github.com/onflow/cadence/blob/master/.github/workflows/codeql-analysis.yml. Maybe update it (it's using v2, this PR adds v4)

.github/workflows/codeql-analysis.yml
```diff
@@ -23,23 +22,38 @@ jobs:
...
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v2
+      uses: github/codeql-action/init@v4
       with:
         languages: go
+        queries: security-extended
 
     - name: Build
-      run: make build
+      run: CGO_ENABLED=0 go build -mod=vendor -tags=no_cgo ./...
```
turbolent: This should keep using the Makefile
```suggestion
      run: make build
```

turbolent:
This was already auto-merged, fix is #4363

PR #4368, opened 2025-11-28
Compare storage for vm vs interpreter during tests

SupunS:
### 💡Idea:
- When running the test with the compiler, while compiling and preparing the VM, also prepare an interpreter and have it ready.
- Then when invoking the test function with the VM, also invoke the same with the interpreter.
- Then compare the slabs in the two storages.

### ⚠️ Note:
Above happens when the `compile` flag is on for tests, and we already run the tests purely with the interpreter separately as well. This means we are running the tests with the interpreter twice, effectively. We could theoretically remove the separate run of the interpreter tests. But it might need some more work, so kept it as-is for now, 

______

### DIFF
--- a/interpreter/condition_test.go
+++ b/interpreter/condition_test.go
@@ -1192,9 +1192,9 @@ func TestInterpretIsInstanceCheckInPreCondition(t *testing.T) {
 
 	t.Parallel()
 
-	test := func(tt *testing.T, condition string) {
+	test := func(t *testing.T, condition string) {
 
-		inter, err := parseCheckAndPrepareWithOptions(tt,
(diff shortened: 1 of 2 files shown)


test_utils/test_utils.go
```diff
@@ -47,61 +50,31 @@ type ParseCheckAndInterpretOptions struct {
 	HandleCheckerError   func(error)
 }
 
-type VMInvokable struct {
```
SupunS: Move to a separate file

turbolent (approved):
Great work! This is really useful for making sure the compiler/VM matches the interpreter

Issue #4364, opened 2025-11-26
Compiler/VM Optimizations


SupunS:
### Quick wins / Low hanging:

most impactful potentially:
- [ ] #4273 
- [ ] Reusing the VM (vm.NewVM call, need to make linking result reusable, ensure no values are used concurrently)

other:
- [ ] Cache linking results from VM in FVM programs cache. But cannot include interpreter values
- [ ] Make `NativeFunctionValue`, and other large types pointer types
- [ ] #4239

### Long term (more work):
- [ ] #4305
- [ ] Emit VM functions in compiler ([Direct Call Threading](https://www.cs.toronto.edu/~matz/dissertation/matzDissertation-latex2html/node6.html#sec:Direct-Call-Threading))
- [ ] Re-use `getField` results where possible (requires liveness analysis)
- [ ] Add "resource-ness" (flag) to the static-type. Might need a state migration. (maybe not needed for the initial version)
  - Move resource-ness from instruction to static type (currently instructions have resource-nes operand)
- [ ] Optimize conversion function calls
  - [ ] Constant literal argument (e.g. `UInt32(1)`) -> Emit constant load (`GetConstant`). avoids call
  - [ ] Other argument (e.g. `let x: UInt8 = 1; UInt32(x)`) -> Call specific target/source conversion function (`$UInt8ToUInt32(x)`). avoids run-time type dispatch in conversion function
- [ ] concurrent processing in Cadence

PR #4380, opened 2025-12-05
[Compiler] Optimize call frames in the VM

turbolent:
Work towards #4364

## Description

Tried to optimize call frames in the VM:

- Make compiled function value a pointer type
- Make tracing info a zero-sized struct when tracing is disabled (default)
- Reorder struct fields for optimal alignment (using https://github.com/dkorunic/betteralign)

Theoretically this should be more optimal, but running e.g. `BenchmarkRecursionFib`, there doesn't seem to be any difference. Maybe I'm missing something. Opening this PR anyways

______


SupunS (approved):
Yeah, most likely the gains from these optimizations are overshadowed by the other operations. Still makes sense to have these in place 👌

RZhang05 (approved):

PR #4447, opened 2026-02-27
Update to Go 1.26

turbolent:
## Description

______


SupunS (approved):

turbolent:
@SupunS also ran `go fix ./...` and added it to the Makefile and CI, just like flow-go already has/does. 

This automatically cleaned up the code using many of the new features in the recent Go releases.

SupunS (approved):
Nice! Let's just merge it then?

Though I was suggesting to merge this at the end (earlier, during the standup), I then realized doing so would double the work for updating the downstream deps. e.g: Say, for flow-sdk repo, we would require one PR to bump Go version of the repo itself, and another to bump the Cadence version. So just updating Cadence first and then updating the downstream dependencies makes sense / might be easier 👍

turbolent:
Sounds good!

PR #4448, opened 2026-03-04
Optimize transferring array, dict, and composite (e.g., 2x faster for byte arrays)

fxamacker:
Updates https://github.com/onflow/flow-go/issues/8447 https://github.com/onflow/flow-go/issues/8401
Requires https://github.com/onflow/atree/pull/633

This optimization benefits EVM, FVM, etc.  For example, an open yield vault transaction  indirectly uses `Transfer()` more than **2900** times.  More than **1800** of those **2900** `Transfer()` are optimized by this.

This PR optimizes copying arrays or maps that don't require deep copying or slab (aka payload) operations by using Atree's new `Array.Copy()` and `OrderedMap.Copy()`.  For example, copying byte array is 2x faster and uses half the memory allocations.

`EVMAddress` and other frequently used data like hashes currently use byte arrays, so the 2x speedup and reduced memory use for copying them can be worthwhile.  Other data such as Cadence enum is about 1.5x faster to copy.

Currently, `ArrayValue`, `DictionaryValue`, and `CompositeValue` use these Atree functions to copy arrays/maps in `Transfer()`:
- `atree.NewArrayFromBatchData()`
- `atree.NewMapFromBatchData()`

With this PR, Cadence `Transfer()` uses Atree's `Copy()` if:
- The array or map is small (a single slab container).
- All elements are not containers or references to other slabs.
- All elements don't wrap containers or references to other slabs.
- All elements return true from `Storable.CanCopy()`.

Otherwise, `Transfer()` continues to use `NewArrayFromBatchData()` or `NewMapFromBatchData()` for arrays or maps that require deep copy or slab operations.

NOTE: To be conservative, some `Storable` types use the old way of copying but it might be OK to use Atree's `Copy()` for some of them too if the Cadence team determines it would be safe.  See "Caveats" section below for more details.

## Benchmarks

Benchmarks to transfer Cadence [UInt8]:

```
                     │    sec/op    │   sec/op     vs base                │
ByteArrayTransfer-12   1528.5n ± 2%   755.6n ± 2%  -50.57% (p=0.000 n=20)

                     │ before.txt  │              after.txt              │
                     │    B/op     │    B/op      vs base                │
ByteArrayTransfer-12   1551.5 ± 0%   1014.0 ± 0%  -34.64% (p=0.000 n=20)

                     │ before.txt  │             after.txt              │
                     │  allocs/op  │ allocs/op   vs base                │
ByteArrayTransfer-12   14.000 ± 0%   7.000 ± 0%  -50.00% (p=0.000 n=20)
```

Benchmarks to transfer Cadence enum:

```
                      │ before.txt  │              after.txt              │
                      │   sec/op    │   sec/op     vs base                │
EMVAddressTransfer-12   2.692µ ± 1%   2.114µ ± 1%  -21.49% (p=0.000 n=20)

                      │  before.txt  │              after.txt               │
                      │     B/op     │     B/op      vs base                │
EMVAddressTransfer-12   3.163Ki ± 0%   2.407Ki ± 0%  -23.90% (p=0.000 n=20)

                      │ before.txt │             after.txt              │
                      │ allocs/op  │ allocs/op   vs base                │
EMVAddressTransfer-12   36.00 ± 0%   29.00 ± 0%  -19.44% (p=0.000 n=20)
```

Benchmarks to transfer EVMAddress (composite value with [UInt8: 20] field):

```
                │    sec/op    │   sec/op     vs base                │
EnumTransfer-12   1228.0n ± 1%   815.6n ± 3%  -33.59% (p=0.000 n=20)

                │ before.txt  │             after.txt              │
                │    B/op     │    B/op     vs base                │
EnumTransfer-12   1719.0 ± 0%   843.5 ± 1%  -50.93% (p=0.000 n=20)

                │ before.txt │             after.txt              │
                │ allocs/op  │ allocs/op   vs base                │
EnumTransfer-12   19.00 ± 0%   13.00 ± 0%  -31.58% (p=0.000 n=20)

```

### Caveats

To be conservative, `CanCopy()` returns `false` for these `Storable` types, and they continue to use the old way of copying:
- `NonStorable`
- `AccountCapabilityControllerValue`
- `PublishedValue`
- `StorageCapabilityControllerValue`

If needed, the Cadence team can look into these 4 types to determine if some of them are safe to copy their values using the new Atree `Copy()` function.

### TODO

- [x] Cadence team (not me) [needs to add new metering kinds](https://github.com/onflow/cadence/pull/4448#discussion_r2884460527).  Done by @turbolent by adding them to this PR. :+1: 
- [ ] Outside this PR, @janezpodhostnik needs to assign weights to the new metering kinds. :pray: 

______


turbolent (commented):
Great work! Great idea to optimize this common case with a "fast path" 👏 

The changes in the `Transfer` methods look good. Just a few minor TODOs / concerns:
- [x] Improve method naming (see atree PR)
- [x] Resolve question re: large values with separate slab
- [x] Add back computation metering
- [x] Add back tracing

interpreter/value_dictionary.go
```diff
@@ -1417,86 +1407,110 @@ func (v *DictionaryValue) Transfer(
 		)
 		common.UseMemory(context, elementMemoryUse)
 
-		func() {
-			seed := v.dictionary.Seed()
+		isDictSafeToCopy := v.dictionary.IsWithinSingleSlab() && canCopyNonRefSimpleForType(v.Type.KeyType) && canCopyNonRefSimpleForType(v.Type.ValueType)
+		canCopyNonRefSimple := isDictSafeToCopy || v.dictionary.CanCopyNonRefSimple()
+
+		if canCopyNonRefSimple {
```
turbolent: Same question as for arrays above

interpreter/value_array.go
```diff
@@ -1295,72 +1288,92 @@ func (v *ArrayValue) Transfer(
 		common.UseMemory(context, dataSlabs)
 		common.UseMemory(context, metaDataSlabs)
 
-		func() {
+		isArraySafeToCopy := v.array.IsWithinSingleSlab() && canCopyNonRefSimpleForType(v.Type.ElementType())
```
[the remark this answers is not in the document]
fxamacker: > Could you please also add comments for the optimizations at isSingleSlabCopyableArrayType, isSingleSlabCopyableDictType, isSingleSlabEnum?

I added more comments in commit [49c048f](https://github.com/onflow/cadence/pull/4448/commits/49c048f848c47b8356281754b0f2b5d11aef85e6).

turbolent:
Triggering stuck CI by closing and reopening

fxamacker:
Thanks @turbolent for adding new metering kinds.

Outside this PR, @janezpodhostnik will need to assign weights to the new metering kinds. 🙏

turbolent (approved):
Great work! 👏

interpreter/value_capability.go
```diff
@@ -270,3 +270,16 @@ func (v *IDCapabilityValue) ChildStorables() []atree.Storable {
...
+	return true
+}
+
+func (v *IDCapabilityValue) CopyNonRefSimple() (atree.Storable, error) {
+	// Use shallow copy of StaticType since static type isn't expected to be changed.
+	return &IDCapabilityValue{
+		BorrowType: v.BorrowType,
+		address:    v.address,
+		ID:         v.ID,
+	}, nil
```
SupunS: For all values that create a new instance, we might want to memory-meter the newly created value-instance. Just need to switch to using the constructor, e.g: `NewCapabilityValue`, but might have to pass down the `MemoryGauge` as a parameter for this function.

I don't think it's a blocker for this PR, so we can do that separately I guess.

SupunS (approved):
LGTM! 🚀

interpreter/value_capability.go
SupunS: Great, thank you!

SupunS (approved):

PR #4465, opened 2026-04-13
docs: clarify ContractUpdateValidator scope and threat model

j1010001:
## Summary

- Adds a package-level comment to `stdlib/contract_update_validation.go` explaining the validator's narrow purpose: storage compatibility, not a general-purpose interface freeze.
- Explicitly documents what is and isn't enforced, and why — in particular that access modifiers and `let`/`var` mutability are intentionally out of scope.
- Adds a short doc comment on `checkField()` pointing to the package-level explanation.
- Includes a link to the [contract updatability docs](https://cadence-lang.org/docs/language/contract-updatability).

## Motivation

The validator's design intent was not documented, making it easy to misread the absence of access-modifier and mutability checks as oversights rather than deliberate decisions. This change makes the threat model explicit in the source so that readers — including security researchers — have the context needed to correctly assess the validator's behaviour.

## Test plan

- [ ] No logic changes — comment-only diff, no tests required.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/stdlib/contract_update_validation.go
+++ b/stdlib/contract_update_validation.go
@@ -16,6 +16,38 @@
  * limitations under the License.
  */
 
+// Package-level note on the ContractUpdateValidator's security model and scope:
+//
+// The ContractUpdateValidator enforces upgrade safety on contract update transactions
+// (account.contracts.update()). Its purpose is narrow and deliberate: ensure that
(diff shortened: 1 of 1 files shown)


turbolent (approved):
Nice! 

Could we please also add this to the top of https://cadence-lang.org/docs/language/contract-updatability ?

PR #4466, opened 2026-04-14
Add freezeTime and unfreezeTime to Test stdlib Blockchain interface

holyfuchs:
Scheduled transactions execute at a time determined by the emulator clock, which drifts with real wall time. This makes tests that assert on timing-sensitive behavior inherently flaky. 
To fix this, `cadence-tools`' `EmulatorBackend` will implement `FreezeTime`/`UnfreezeTime` implementations, to make them reachable from Cadence test code, they will be added here.


turbolent (approved):
Nice!

PR #4473, opened 2026-04-18
Add pow function to UFix64 and UFix128

turbolent:
Work towards #1151

## Description

Add `pow` function to `UFix64` and `UFix128` fixed-point types, backed by `Pow` methods in https://github.com/onflow/fixed-point:

- `UFix64.pow(_ exponent: Fix64): UFix64`
- `UFix128.pow(_ exponent: Fix128): UFix128`

______

### DIFF
--- a/bbq/vm/builtin_globals.go
+++ b/bbq/vm/builtin_globals.go
@@ -293,6 +293,8 @@ func init() {
 	registerBuiltinCommonTypeBoundFunctions()
 
 	registerBuiltinSaturatingArithmeticFunctions()
+
+	registerBuiltinFixedPointPowFunctions()
 }
 
 func registerBuiltinCommonTypeBoundFunctions() {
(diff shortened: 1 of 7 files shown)


bbq/vm/builtin_globals.go
```diff
@@ -421,6 +423,19 @@ func registerBuiltinTypeSaturatingArithmeticFunctions(t sema.SaturatingArithmeti
 	}
 }
 
+func registerBuiltinFixedPointPowFunctions() {
+	for baseType, funcType := range sema.FixedPointPowFunctionTypes { //nolint:maprange
```
SupunS, line 427: It seems `sema.FixedPointPowFunctionTypes` gets populated only when the `FixedPointNumericType.GetMembers()` is called. Could this be a problem? e.g: the current program doesn't use the `pow` function statically, but an imported program (maybe dynamically) calls to the `pow` function at runtime.

I think its applicable for both the interpreter and vm.

turbolent: Good catch, that didn't occur to me. While looking at it again, I realized we already have an established pattern for saturating arithmetic functions. Improved this in 344a5af580e17a4225fa8fba1baa50a5e7471360

turbolent: Also fixed this in the PR stacked on top of this, which had the same issue: [`fab5f47` (#4478)](https://github.com/onflow/cadence/pull/4478/commits/fab5f4777cbb9b1889d7545183b6035f24c83b69)

SupunS (approved):
Nice!

Issue #4484, opened 2026-04-29
Issues encountered building a formatter


janezpodhostnik:
# Issues encountered building a formatter

Issues in the Cadence AST `Doc()` and `Walk()` methods that affect external formatting tools.

---

## 1. `EntitlementDeclaration.Doc()` uses `HardLine` for access modifiers

`EntitlementDeclaration.Doc()` and `EntitlementMappingDeclaration.Doc()` use `prettier.HardLine{}` after the access modifier, forcing a line break:

```go
// entitlement_declaration.go:111
doc = append(doc, docOrEmpty(d.Access), prettier.HardLine{})
```

This produces:

```cadence
access(all)
entitlement NodeOperator
```

All other declaration types (`FunctionDeclaration`, `CompositeDeclaration`, `FieldDeclaration`, etc.) use `prettier.Line{}`, which keeps the access modifier on the same line when it fits:

```cadence
access(all) fun foo()
access(all) struct Bar {}
```

---

## 2. Move operator `<-` spacing inconsistency

The move operator `<-` is formatted inconsistently across `Doc()` methods:

- `return <-expr` (no space after `<-`)
- `self.x <- expr` (space around `<-`)

This appears to stem from different handling of `TransferOperation` rendering in return statements vs assignment statements.

---

## 3. `InvocationExpression.Walk()` doesn't yield `Argument` wrappers

`InvocationExpression.Walk()` yields `arg.Expression` for each argument, but not the `Argument` struct itself:

```go
func (e *InvocationExpression) Walk(walkChild func(Element)) {
    walkChild(e.InvokedExpression)
    for _, typeArgument := range e.TypeArguments {
        walkChild(typeArgument)
    }
    for _, argument := range e.Arguments {
        walkChild(argument.Expression)  // only the Expression, not the Argument
    }
}
```

Since `Argument` has positional information (`LabelStartPos`, `LabelEndPos`, `TrailingSeparatorPos`), tools that map source positions to AST nodes (e.g., for comment attachment) cannot correctly associate comments that fall between argument labels and their expressions.

For comparison, `ParameterList.Walk()` yields individual `Parameter` elements.

hiSandog:
[...]

PR #4485, opened 2026-04-29
Fix AST Doc() and Walk() issues for formatter support

janezpodhostnik:
## Summary

  Fixes issues in the Cadence AST `Doc()` and `Walk()` methods that affect external formatting tools (#4484).

  - **`HardLine` → `Line` after access modifiers**: All declaration `Doc()` methods used `prettier.HardLine{}` after access modifiers, forcing a line break. Changed to `prettier.Line{}` so
  access modifiers stay on the same line when content fits (e.g. `access(all) fun foo()` instead of a forced line break).
  - **Move operator spacing**: `UnaryExpression.Doc()` produced `return <-expr` with no space after `<-`, while `AssignmentStatement.Doc()` produced `x <- expr` with spaces. Added
  `prettier.Space` after `OperationMove` in `UnaryExpression.Doc()` for consistency.
  - **`Argument` as walkable `Element`**: `InvocationExpression.Walk()` yielded `argument.Expression` directly, skipping the `Argument` wrapper and its positional information (`LabelStartPos`,
   `LabelEndPos`, `TrailingSeparatorPos`). `Argument` now implements the `Element` interface and `Walk()` yields the wrapper, enabling correct source-position mapping for tools like comment
  attachment.
  - **Indent on continuation lines**: `BinaryExpression.Doc()` and `CastingExpression.Doc()` placed operator continuations at the same indent level as the preceding expression. Wrapped
  continuations in `prettier.Indent{}` so operators indent when they wrap to a new line.
  - **Spurious space in anonymous functions**: `FunctionDocument()` unconditionally added a space after `fun`, producing `fun ()` for anonymous functions. Space is now conditional on the
  identifier being present, producing `fun()`.
  - **`EntitlementMappingDeclaration.Doc()` grouping**: Wrapped return value in `prettier.Group{}` for consistency with `EntitlementDeclaration.Doc()`.
  
______


turbolent:
Thank you for improving the walking and formatting 👍 Code style is a very subjective and controversial topic 😄  

> HardLine → Line after access modifiers

This was intentional. When we changed `pub`/`priv` to `access(all)`/`access(self)` I had found function signatures to become very long and hard to read. I felt like keeping the access modifier as an "access annotation" above and keeping the `fun` at the start would be easier to read. 

This hasn't caught on and the majority of the code keeps the access modifier inline, so this change is probably the best way to go forward 👍 

> Move operator spacing

This was also intentional. In the assignment case it is common to use whitespace on both sides of the operator (e.g. also `x = 1` and not `x =1`), and for unary operators it is also common to have no space (e.g. `-1` and not `- 1`). I'd maybe keep this consistent, or we can make it configurable.

> Indent on continuation lines

Not quite sure what this means, do you have an example? IMHO I'd ideally have `a + b` wrap as `a\n    + b`, i.e. keep the operator at the start of the continuation line (unlike e.g. Go, which keeps the operator at the end of the previous/first line)

BTW, I don't know if you have seen it: We have a development tool in https://github.com/onflow/cadence/tree/master/tools/pretty which allows testing the code formatting. Paired with a "watcher" like e.g. https://github.com/mitranim/gow, it allows quickly iterating on the implementation of these function, trying out the formatting and wrapping based on varying max lengths.

turbolent (approved):

turbolent:
Thank you!

PR #4487, opened 2026-05-14
Record occurrences of nested types and imported members

turbolent:
## Description

Needed for https://github.com/onflow/cadence-tools/pull/636

Also record the occurrences of
- Nested types: We already record the occurrence of `C` in `C.S`, also record the occurrence of `S`. 
- Imported members

______

### DIFF
--- a/sema/checker.go
+++ b/sema/checker.go
@@ -1272,6 +1272,10 @@ func (checker *Checker) convertNominalType(t *ast.NominalType) Type {
 			)
 			return InvalidType
 		}
+
+		if checker.PositionInfo != nil && identifier.Identifier != "" {
+			checker.recordNestedTypeReferenceOccurrence(identifier, ty)
+		}
 	}
 
 	return ty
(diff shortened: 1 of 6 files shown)


sema/checker.go
```diff
@@ -1272,6 +1272,10 @@ func (checker *Checker) convertNominalType(t *ast.NominalType) Type {
 			)
 			return InvalidType
 		}
+
+		if checker.PositionInfo != nil && identifier.Identifier != "" {
+			checker.recordNestedTypeReferenceOccurrence(identifier, ty)
```
SupunS, line 1277: Not related to this PR: I guess these information are only used in dev-tooling? Was wondering if it would be a good idea to put these behind a "dev-mode" flag? (just exploring the idea - not that we need to do it now)

SupunS (approved):

sema/checker.go
turbolent: Correct, the position info is only needed for tooling. It is basically already gated: `Checker.PositionInfo` is only initialized if `Config.PositionInfoEnabled` is enabled, and only tooling enables it

SupunS: ah nice! I didn't realize it is already gated.

SupunS (approved):

PR #4495, opened 2026-06-05
[Compiler] Separate attachment-base passing from arguments

SupunS:

## Description

So far, `base` reference for attachment functions/constructor was passed-in as part of the arguments list. This had complicated things, since `base` needed to be special cased in several places. Simplify this by separating the `base` from normal arguments, and handle the same way as `self` is being passed into functions.

Also fixes an inconsistency where `base` was placed **before** the normal arguments for attachment bound-functions, where as `base` was placed **after** the normal arguments for attachment constructors. Again, this was both complicating things, as well as inconsistent. So improved this by always placing the `base` before the normal arguments, regardless of the scenario.

______


turbolent (approved):
Great work fixing and cleaning this up!

PR #4497, opened 2026-06-10
[Compiler] Run computation metering tests with compiler/vm

SupunS:

## Description

Enables the computation metering tests with the compiler and VM.

______

### DIFF
--- a/interpreter/account_test.go
+++ b/interpreter/account_test.go
@@ -643,7 +643,7 @@ func testAccountWithErrorHandler(
 	} else {
 
 		// NOTE: test code for VM above
-		inter, err := parseCheckAndInterpretWithOptions(t,
+		inter, err := parseCheckAndInterpretWithOptions(t, //nolint:staticcheck
 			code,
 			ParseCheckAndInterpretOptions{
 				ParseAndCheckOptions: &ParseAndCheckOptions{
--- a/interpreter/computation_metering_test.go
+++ b/interpreter/computation_metering_test.go
@@ -37,22 +37,8 @@ import (
 	. "github.com/onflow/cadence/test_utils/sema_utils"
 )
 
-// The tests in this file run on both execution engines: the tree-walking
-// interpreter and, when `-compile=true` is set, the bytecode compiler + VM
-// (see parseCheckAndPrepareWithOptions and the `compile` flag).
-//
-// The two engines do not always meter computation identically. Rather than
-// changing production metering, these tests treat each engine's current
-// behaviour as correct and, where they differ, assert the compiler/VM's
-// expected usages in a `if *compile { ... }` branch. The known, systematic
-// reasons for the differences are:
-//
-//   - Function invocations: the interpreter does not meter the entry-point
-//     invocation (the `Invoke` call from the test) nor invocations made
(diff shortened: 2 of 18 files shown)


interpreter/computation_metering_test.go
```diff
@@ -37,6 +37,44 @@ import (
...
+//   - Function invocations: the interpreter does not meter the entry-point
+//     invocation (the `Invoke` call from the test) nor invocations made
+//     internally by built-in functions, while the VM meters every `Invoke`
+//     instruction. The VM therefore reports additional FunctionInvocation
+//     usages, and meters the invocation after evaluating the call's arguments
+//     (the interpreter meters it before), which also changes the ordering.
+//
+//   - Global variable/constant declarations: the interpreter meters the
+//     top-level declaration as a Statement when evaluated, whereas the VM's
+//     global initializers are not metered as statements.
```
SupunS, line 45: I need to check how metering works for global variables in compiler. Will open a follow-up PR is there's any gap.

interpreter/account_test.go
```diff
@@ -643,7 +643,7 @@ func testAccountWithErrorHandler(
 	} else {
 
 		// NOTE: test code for VM above
-		inter, err := parseCheckAndInterpretWithOptions(t,
+		inter, err := parseCheckAndInterpretWithOptions(t, //nolint:staticcheck
```
turbolent, line 646: Why do we have to ignore the linter here and below? Is this now flagged as deprecated?

interpreter/computation_metering_test.go
```diff
@@ -2592,12 +3726,23 @@ func TestInterpretComputationMeteringIntegerParsing(t *testing.T) {
 		_, err = inter.Invoke("test")
 		require.NoError(t, err)
 
-		AssertEqualWithDiff(t,
-			[]common.ComputationUsage{
+		var expectedUsages []common.ComputationUsage
+		if *compile {
+			expectedUsages = []common.ComputationUsage{
+				{Kind: common.ComputationKindFunctionInvocation, Intensity: 1},
```
turbolent: Here and in the other tests: why do we get one more function invocation in the VM? If it is more accurate, maybe we can change metering in the interpreter to match it?

turbolent (approved):
Nice, thanks for adding all the compiler/VM assertions 👍

interpreter/account_test.go
SupunS: Yes, deprecated this API so it won't get used accidentl;y

interpreter/computation_metering_test.go
SupunS: The additional invocation is for the entry-point (external invocation in tests). In the VM all invocations goes through a single metered-function. Whereas, in the interpreter, metering happens at the visitor method `VisitInvocationExpression`, and the test entry-point function get invoked outside of the visitor method.

SupunS: Added metering to the entry-point: https://github.com/onflow/cadence/pull/4497/changes/1d1f8e86636bcdd50b0ee962a58d0a665ad0c6da

turbolent (approved):
Nice!

PR #4505, opened 2026-06-17
[Compiler] Fix before-statements parameter binding during desugaring

SupunS:

## Description

Interface functions and their concrete implementations may have different parameter names. When a before-statement defined in the interface refers to the interface function's parameter name, it should be bound to the concrete function's parameter name when desugaring. 

### Bug 1:
Currently, before-statements does not bind the parameters, hence resulting in a compilation failure due to mismatching names (lookup fails).

This has ben already handled for `pre` and `post` conditions. This PR adds the same parameter binding to the before-statements as well.

### Bug 2:
Parameter bindings were mapping the interface function's param names to the implementation's param names. However, these binding were created per-statement (per condition), and conditions are common/re-used across multiple implementations. Thus, if different implementions used different param-names, then the mapping gets corrupted.

Therefore, instead of mapping to the implementation's param names (i.e: instead of making the binding implementations specific), bind the names to the index, so it is generic across all implementtions.
______


bbq/compiler/desugar.go
```diff
@@ -564,22 +574,48 @@ func (d *Desugar) desugarInheritedCondition(
...
+	functionParams *ast.ParameterList,
+	inheritedFunc *inheritedFunction,
+) map[string]int {
+	if functionParams == nil || len(functionParams.Parameters) == 0 {
+		return nil
+	}
+
+	paramBinding := make(map[string]int)
+
+	inheritedFunctionParams := inheritedFunc.functionDecl.ParameterList
```
turbolent: Maybe provide the capacity here:

```suggestion
	inheritedFunctionParams := inheritedFunc.functionDecl.ParameterList
	
	paramBinding := make(map[string]int, len(inheritedFunctionParams.Parameters))

```

turbolent (approved):
Good catch!

bbq/compiler/desugar.go
SupunS: Added in https://github.com/onflow/cadence/pull/4505/changes/9898cf36f122048bb6e525aa4f8129bea13c10db

PR #4520, opened 2026-07-28
[Compiler] Fix compiling inherited code from same contract

SupunS:

### Problem

Compiling inherited code (e.g: pre/post condition) always treated contract variables as references (i.e: `Foo` in an expression `Foo.hello()`, where `Foo` is a contract). This was correct if the contract `Foo` is an import.

However, this wasn't correct for scenarios where the inherited code  refers to the same contract, and those should not be treat as references.

### Fix

Treat the contract variable as a reference only if it's an imported contract. Otherwise (it's the enclosing contract of the current code) treat it as the concrete type.

______


turbolent (approved):
Good catch!

